PR #60, opened 2018-04-04
Tg bg 1294

tomshopsys:
| Q             | A
| ------------- | ---
|New feature| No
|BC breaks| No
|Fixes issues| ...
|Standards and tests pass| ?
|License| MIT

### DIFF
--- a/packages/framework/src/Resources/views/Admin/Content/Superadmin/cssDocumentation.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Content/Superadmin/cssDocumentation.html.twig
@@ -781,7 +781,7 @@
 
         <div class="in-article__preview">
             <span class="in-article__preview__info">{{ 'Live view of the component in white window'|trans }}</span>
-            <textarea class="input input--textarea-big">{{ '--textarea-big'|trans }}</textarea>
+            <textarea class="input input--textarea">{{ '--textarea'|trans }}</textarea>
             <br><br>
             <textarea class="input input--seo">{{ '--seo'|trans }}</textarea>
         </div>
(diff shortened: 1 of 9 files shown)


packages/framework/src/Resources/views/Admin/Content/Superadmin/cssDocumentation.html.twig
```diff
@@ -1074,15 +1074,15 @@
                     {{ 'Textarea may be across the whole width'|trans }}
                 </label>
                 <div class="form-line__line">
-                    <textarea class="input input--textarea-big">{{ 'Textarea spread to the whole width. In this case it is not necessary to use form-line__item, because the element is across 100 % of width.'|trans }}</textarea>
+                    <textarea class="input input--textarea">{{ 'Textarea spread to the whole width. In this case it is not necessary to use form-line__item, because the element is across 100 % of width.'|trans }}</textarea>
```
TomasLudvik: This and textarea on line 1085 seems same

TomasLudvik: Also, there is textarea on line 784, that might be different because of the label, is it right? I do not know where to open this documentation live, so I could compare it rendered.

TomasLudvik:
This issue does not pass on the CI. I cannot test if every acceptance criterium has been solved. Please fix this.

packages/framework/src/Resources/views/Admin/Content/Superadmin/cssDocumentation.html.twig
MattCzerner: The line 1085 and 1077 are not the same, the wrap is different, i dont actually want to change anything in this file, because it is used as showcase and i think that it should be reviewed as whole. I wouldnt spend time on this right now if you agree.

You can see documentation if you log into administration as superadmin, that way you can click on superadmin section where is documentation situated.

MattCzerner:
Branch now passes test on our CI so you can check it :)

packages/framework/src/Resources/views/Admin/Content/Superadmin/cssDocumentation.html.twig
TomasLudvik: Ok, just create US for that review.

TomasLudvik:
I do not think, that we should resize all textareas in the administration, my suggestion was only for those including code. The others were sized by request of product owner.

PR #72, opened 2018-04-09
[US-3559] autowiring deprecations

EdoBarnas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| ...
|New feature| No 
|BC breaks| No 
|Fixes issues| ...
|Standards and tests pass| Yes
|License| MIT
|Have you read and signed our [License Agreement for contributions](https://www.shopsys-framework.com/license-agreement)?| Yes=

PR #87, opened 2018-04-13
[US-3558] service deprecations

EdoBarnas:
| Q             | A
| ------------- | ---
|New feature| No 
|BC breaks| No 
|Fixes issues| ...
|License| MIT

### DIFF
--- a/packages/framework/src/Resources/config/services.yml
+++ b/packages/framework/src/Resources/config/services.yml
@@ -480,6 +480,8 @@ services:
     calls:
       - [setContainer, ['@service_container']]
 
+  Symfony\Component\Config\Loader\LoaderInterface: '@routing.loader'
+
   Shopsys\FrameworkBundle\Component\Router\CurrentDomainRouter:
     tags:
       - { name: router, priority: 70 }
@@ -496,6 +498,8 @@ services:
   Shopsys\FrameworkBundle\Component\Router\Security\RouteCsrfProtector:
     arguments:
       - '@annotations.reader'
+
+  Shopsys\FrameworkBundle\Component\Grid\GridFactory: ~
 
   Shopsys\FrameworkBundle\Model\Script\ScriptFacade:
     public: true
--- a/packages/framework/src/Resources/config/services.yml
+++ b/packages/framework/src/Resources/config/services.yml
@@ -43,6 +43,10 @@ services:
       - '@request_stack'
       - '%kernel.debug%'
 
+  Symfony\Component\Translation\MessageSelector: '@translator.selector'
+
+  Doctrine\Common\Annotations\DocParser: '@jms_translation.doc_parser'
+
   # disable original DefaultPhpFileExtractor
   jms_translation.extractor.file.default_php_extractor:
     alias: Shopsys\FrameworkBundle\Component\Translation\PhpFileExtractor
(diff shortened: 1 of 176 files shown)


PetrHeinz:
I am closing this PR for now as it is based on `eb-autowiring-dependencies` for which you created a separate PR #72. After #72 is accepted we can rebase this branch and jump right into it.

packages/framework/src/Resources/config/services.yml
```diff
@@ -13,6 +13,15 @@ services:
...
     public: false
 
+  Shopsys\FrameworkBundle\:
+      resource: '../../{Component,Model,DataFixtures}/**/{*Resolver.php,*Factory.php,*Service.php,*Facade.php,*Repository.php,*Listener.php,*Loader.php}'
+
+  Shopsys\FrameworkBundle\Component\:
+    resource: '../../Component/**/{*Provider.php,*Transformer.php,*Finder.php,*Parser.php,*DataCreator.php}'
+
+  Shopsys\FrameworkBundle\Model\:
+    resource: '../../Model/**/{*Scheduler.php,*Generator.php,*Calculation*.php,*Recalculator.php,*Filter.php,*Mail.php,*Setting.php,*Config.php,*DataMapper.php}'
```
PetrHeinz: `../../Model/**/*Calculation*.php` will match for example `\Shopsys\FrameworkBundle\Model\Product\Exception\InvalidPriceCalculationTypeException`, which should not be defined as a service.

I suggest auto-discovering only by a postfix, i.e. using `../../Model/**/*Calculation.php`. This would not match `Model/Product/Pricing/ProductPriceCalculationForUser`, this class should be defined explicitly (as it is an exception from the rule).

packages/framework/src/Resources/config/services.yml
```diff
@@ -102,8 +102,6 @@ services:
   Shopsys\FrameworkBundle\Component\Breadcrumb\BreadcrumbResolver:
     factory: ['@Shopsys\FrameworkBundle\Model\Breadcrumb\FrontBreadcrumbResolverFactory', create]
 
-  Shopsys\FrameworkBundle\Model\Breadcrumb\FrontBreadcrumbResolverFactory: ~
```
PetrHeinz: How did you find out which services to remove?

packages/framework/src/Resources/config/services.yml
```diff
@@ -13,6 +13,15 @@ services:
...
     public: false
 
+  Shopsys\FrameworkBundle\:
+      resource: '../../{Component,Model,DataFixtures}/**/{*Generator.php,*Resolver.php,*Provider.php,*Registry.php,*Factory.php,*Service.php,*Facade.php,*Creator.php,*Repository.php,*Mapper.php,*Listener.php,*Subscriber.php,*Loader.php,*Injector.php,*Reader.php,*Writer.php,*Setting.php,*Config.php}'
+
+  Shopsys\FrameworkBundle\Component\:
+    resource: '../../Component/**/{*Transformer.php,*Finder.php,*Parser.php,*Comparator.php,*Quoter.php,*CompilerPass.php}'
+
+  Shopsys\FrameworkBundle\Model\:
+    resource: '../../Model/**/{*Scheduler.php,*Handler.php,*Calculation.php,*Recalculator.php,*Filter.php,*Mail.php,*InlineEdit.php}'
```
PetrHeinz: I would rather auto-discover classes with these prefixes in the whole `src` directory so it's simpler to use and maintain. Why did you separate it into 3 sets?

packages/framework/src/Resources/config/services.yml
EdoBarnas: Removed services did not have configuration and each of them suited autodiscovery pattern.

packages/framework/src/Resources/config/services.yml
EdoBarnas: These sets seem clearer for me. Do you want to merge that?

EdoBarnas:
Autoloading project-base is a great idea. I think we should create new US.

PR #88, opened 2018-04-14
composer: simplify autoloading paths to make it more readable

TomasVotruba:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Cleaner reading of `composer.json`
|New feature| No
|BC breaks| No
|Standards and tests pass| Yes
|License| MIT
|Have you read and signed our [License Agreement for contributions](https://www.shopsys-framework.com/license-agreement)?| Yes

PR #96, opened 2018-04-20
[BG-2586] Consistency of namespaces and directory structures in packages

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| there were inconsistencies in namespaces and directory structures, reported here: https://github.com/shopsys/shopsys/pull/88#discussion_r181557290
|New feature| No 
|BC breaks| No, directory change should not be a BC break, namespace changed in test classes only 
|Fixes issues| ...
|Standards and tests pass| Yes, I've built shopsys/coding-standards and shopsys/migrations locally and run all Travis checks as well
|License| MIT


packages/migrations/app/phpunit.xml
```diff
@@ -1,21 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
PetrHeinz: This configuration was never used and in other packages we do not provide a custom configuration, so I've removed it.

packages/http-smoke-testing/.travis.yml
```diff
@@ -21,4 +21,3 @@ script:
   - php vendor/bin/php-cs-fixer fix --config=vendor/shopsys/coding-standards/build/phpcs-fixer.php_cs --dry-run --verbose --diff ./src ./tests
   - php vendor/bin/phpcs --standard=vendor/shopsys/coding-standards/rulesetCS.xml --extensions=php --encoding=utf-8 --tab-width=4 -sp ./src ./tests
   - php vendor/bin/phpmd "./src,./tests" text vendor/shopsys/coding-standards/rulesetMD.xml --extensions=php
-  - php vendor/bin/phpunit tests
```
simara-svatopluk: This change is not mentioned in the commit message. It doesn't belong here maybe

PetrHeinz: Thanks for noticing, I've managed to confuse http-smoke testing with migrations, as you can see in https://github.com/shopsys/shopsys/pull/96/commits/2fdcb5be82fe2766226484206716c6edd0d9e1d8

PetrHeinz: I've fixed it

PetrHeinz:
I've made the test directory consistent in all packages. There is a slight inconsistency in project-base, but such changes would be probably considered a BC break and are outside of the scope of this PR.

A test split and Travis check of all packages will be needed to test this change after accepting CR.

PetrHeinz:
Thanks for the review. I'm splitting all packages into `ph-test-namespace-consistency`, this will trigger Travis build. If everything is green, I'll merge this PR.

PetrHeinz:
Travis green, merging.

packages/coding-standards/composer.json
```diff
@@ -22,12 +22,12 @@
...
     "autoload": {
         "psr-4": {
-            "Shopsys\\CodingStandards\\": "src"
+            "Shopsys\\CodingStandards\\": "src/"
         }
     },
     "autoload-dev": {
         "psr-4": {
-            "Shopsys\\Tests\\Unit\\CodingStandards\\": "tests"
+            "Tests\\CodingStandards\\": "tests/"
```
TomasVotruba: What if there will be 3rd party with the same generic `Tests\CodingStandards` namespace? 

Standard unique naming prevents these bugs:

- Vendor\Package: src
- Vendor\Package\Tests: tests

PetrHeinz: It's in `autoload-dev`, so it shouldn't be a problem. Test classes of 3rd party packages should not be autoloaded.

We have adapted this standard while upgrading to Symfony 3.4. 
https://symfony.com/doc/3.4/testing.html

In 4.0 they switched from `Tests\Namespace` to `Vendor\Tests\Namespace`.
https://symfony.com/doc/current/testing.html

We have never run into issues with this approach but we will keep in mind that it could happen.

TomasVotruba: > It's in autoload-dev...

It's very easy to confuse those and use non-vendor-prefixed naming in `autoload`.

> We have adapted this standard while upgrading to Symfony 3.4.
https://symfony.com/doc/3.4/testing.html

Symfony documentation is not very good for testing practise. The source code, on the other hand, tells the true story. I learned unique class naming (test or not) from there. They've been doing for ages, e.g. see Symfony 2.8: https://github.com/symfony/symfony/blob/v2.8.0/src/Symfony/Component/Console/Tests/ApplicationTest.php#L12

> We have never run into issues with this approach but we will keep in mind that it could happen.

This just opens the door to such bugs.

I have run into couple similar issues in my life and it's really exhausting to find such bug, moreover when you really on standards and best practice of open-source.

TomasVotruba: This is my favorite example of "unique" approach that caused already so many troubles related to phpunit, diff and constraint invalid version r apis: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.11/composer.json#L65-L80

PetrHeinz: Good points, I have created a user story to solve this. Thanks for pointing this out.

PR #102, opened 2018-04-26
Protected visibility of all private properties and methods of facades

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When trying to implement entity extension from project-base, I found out that I need to extend behavior of `ArticleFacade::create()` method because it directly creates new instance of the framework `Article` class and I need to exchange it for my `ExtendedArticle` class in `project-base` (instantiation using factories is work-in-progress in another user-story). This could be easily accomplished by extending `ArticleFacade`, only if properties of the class were `protected`, not `private`. We discussed it with @PetrHeinz and he suggested this ad-hoc solution. Facades are used as an entry point all across the framework so we believe this step is very useful (and probably necessary - will be further explored later) towards the glassbox extensibility.
|New feature| Yes
|BC breaks| No
|Fixes issues| ...


PetrHeinz:
merged manually in https://github.com/shopsys/shopsys/commit/8325f9868a239bb5617ef1665b4240da69ef5767

PR #113, opened 2018-05-07
[BG-2618] terminology: expression "indexes" is used now instead of "i…

Miroslav-Stopka:
…ndices"

| Q             | A
| ------------- | ---
|Description, reason for the PR| BG-2618
|New feature| No 
|BC breaks| No 
|Fixes issues| ...


packages/framework/src/Resources/views/Admin/Content/Product/edit.html.twig
```diff
@@ -354,7 +354,7 @@
                             </button>
                             <span class="list-images__item__overlay js-product-images-image-overlay">{{ 'Image will be deleted after saving.'|trans }}</span>
                         </div>
-                        {# Ignore indices from form definition. Let them reflect new positions. #}
+                        {# Ignore index from form definition. Let them reflect new positions. #}
```
vitek-rostislav: "indexes"

Miroslav-Stopka: done

PR #114, opened 2018-05-10
[US-3646] docs: coding-standards: visibility of all properties and me…

Miroslav-Stopka:
…thods of entities must be protected or public to enable extensibility

| Q             | A
| ------------- | ---
|Description, reason for the PR| [US-3646] Rule about visibility of all properties and methods of entities must be protected or public is not checked automatically
|New feature| No 
|BC breaks| No 
|Fixes issues| ...

### DIFF
--- a/docs/contributing/coding-standards.md
+++ b/docs/contributing/coding-standards.md
@@ -95,3 +95,4 @@ Besides the rules that are checked by automatic tools, we have few rules for whi
     }
     ```
 - Entities have to be created by factories. Only allowed exception are `*Translation` entities that are created by their owner entity.
+- Visibility of all properties and methods of entities must be protected or public to enable extensibility. 


docs/contributing/coding-standards.md
```diff
@@ -95,3 +95,4 @@ Besides the rules that are checked by automatic tools, we have few rules for whi
     }
     ```
 - Entities have to be created by factories. Only allowed exception are `*Translation` entities that are created by their owner entity.
+- Visibility of all properties and methods of entities must be protected or public to enable extensibility.
```
simara-svatopluk: It is a bit difficult to understand for me. I suggest:

- Properties and methods of entities must be `protected` or `public`. To enable extensibility.

If You agree, please change it and also the commit message. If You disagree, it's ok and leave Your version :-)
The CR is done either way :+1:

simara-svatopluk: If You don't think two sentences are better than one, it's ok. This is just one sentence documentation, keep what You prefer better :-) You version contains everything, and is ok.

PR #122, opened 2018-05-15
[BG-1625] Scheduled immediate price recalculation option is removed

Miroslav-Stopka:
- price recalculation is called directly without scheduler
- tests of removed methods are removed too
 - ProductPriceRecalculatorTest: immediateRecalculation does not exist anymore
 - framework/../ProductServiceTest: visibilityRecalculation is not part of ProductService
 - project-base/../ProductServiceTest: ProductPriceRecalculator requires existing Product for ProductCalculatedPrice

- scheduler for immediate price recalculation was required in the past because of recalculation with the use of native query. Now recalculation is processed with the use of objects. Main problem of scheduled approach is that immediate recalculation is initiated only with kernelResponse. It means that even if ProductFacade->edit is called, prices are not recalculated, prices are only scheduled for recalculation. In the case of usage of edit method without following call of KernelResponse, in this case prices will be not recalculated - for example editing product from transfers.

| Q             | A
| ------------- | ---
|Description, reason for the PR| [BG-1625]
|New feature| No 
|BC breaks| No 
|Fixes issues| ...


packages/framework/src/Resources/config/services.yml
```diff
@@ -480,10 +480,7 @@ services:
 
   Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPriceCalculationForUser: ~
 
-  Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPriceRecalculator:
-    tags:
-      # must be run before ProductVisibilityFacade
-      - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse, priority: 20 }
+  Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPriceRecalculator: ~
```
TomasLudvik: Hello @Miroslav-Stopka,

this should be removed.

packages/framework/tests/Unit/Model/Product/ProductServiceTest.php
```diff
@@ -239,156 +237,15 @@ public function testDeleteMainVariant()
         $this->assertFalse($variant->isVariant());
     }
 
-    public function testMarkProductForVisibilityRecalculation()
```
TomasLudvik: These tests should be just updated. It is testing of Visibility recalculation, no Price recalculation

project-base/tests/ShopBundle/Database/Model/Product/ProductServiceTest.php
```diff
@@ -27,17 +31,24 @@ public function testRecalculateInputPriceForNewVatPercentWithInputPriceWithoutVa
...
         $productData = new ProductData();
         $productData->price = 1000;
         $productData->vat = $vat;
         $productData->unit = $this->getReference(UnitDataFixture::UNIT_PIECES);
+        $productData->availability = $availability;
         $product = Product::create($productData);
 
         $productManualInputPrice = new ProductManualInputPrice($product, $pricingGroup, 1000);
 
+        $em->persist($vat);
```
TomasLudvik: Why is it necessary to persist Vat and Product now? Why this was not necessary before your changes?

packages/framework/src/Model/Product/ProductService.php
```diff
@@ -33,29 +33,29 @@ class ProductService
...
+        ProductCategoryDomainFactoryInterface $productCategoryDomainFactory,
+        ProductPriceRecalculator $productPriceRecalculator
     ) {
         $this->productPriceCalculation = $productPriceCalculation;
         $this->inputPriceCalculation = $inputPriceCalculation;
         $this->basePriceCalculation = $basePriceCalculation;
         $this->pricingSetting = $pricingSetting;
-        $this->productPriceRecalculationScheduler = $productPriceRecalculationScheduler;
         $this->productCategoryDomainFactory = $productCategoryDomainFactory;
+        $this->productPriceRecalculator = $productPriceRecalculator;
```
TomasLudvik: Service should not be dependent on any Database using Class like Recalculator. It is because services should be easily tested so you do not have to mock every dependency.

Miroslav-Stopka:
Hi Tomas,
lately we found out, that this change will have greater impact then we thought on the start :/ At the same time, more extensive changes of pricing are planned for next weeks. So we decided 
to stop the solution of this problem until all planned changes in pricing are completed. 
Next we will come up with a decision on how we will solve this problem.

I will keep you informed :)

PR #136, opened 2018-05-24
add XMLHttpRequest header

slk500:
|Description, reason for the PR| Be able to test routes that have a condition req. XMLHttpRequest
|New feature| Yes 
|BC breaks| No
|Fixes issues| No


PetrHeinz:
Hello @slk500 and thanks for the contribution. It would be nice to be able to test AJAX request with the HTTP smoke tests.

But this approach means that every request by the HTTP smoke tests will have the same header and that would be misleading. A nice solution would be to add a new method `setHeader($key, $value)` to [`RequestDataSetConfig`](https://github.com/shopsys/shopsys/blob/07abd93ff3353222513ca71322359382104c65dd/packages/http-smoke-testing/src/RequestDataSetConfig.php) interface, adding the headers from `RequestDataSet` to the `$request->headers` in [`HttpSmokeTestCase::createRequest()`](https://github.com/shopsys/shopsys/blob/07abd93ff3353222513ca71322359382104c65dd/packages/http-smoke-testing/src/HttpSmokeTestCase.php), and you could then use it in the customization callback.

This would add a possibility to control the request headers in your test cases.

Issue #151, opened 2018-05-28
 Explanation --stability = alpha in installation guides


jakubhavlak:
In the installation manuals, we explain what the --no-install and --keep-vcs switches mean, it would also describe --stability = alpha, which ensures that the project-base is installed from the last tag, not the master. If someone wants the current master, they must use --stability = dev

vitek-rostislav:
Solved in #162

PR #162, opened 2018-05-28
installation guides: added explanation of composer create-project "stability" option

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| all other options are explained so there should be explanation for "stability" option as well
|New feature| No 
|BC breaks| No 
|Fixes issues| https://github.com/shopsys/shopsys/issues/151

### DIFF
--- a/docs/installation/installation-using-docker-linux.md
+++ b/docs/installation/installation-using-docker-linux.md
@@ -16,6 +16,8 @@ cd project-base
 *Note: The `--no-install` option disables installation of the vendors - this will be done later in the Docker container.*
 
 *Note: The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
+
+*Note: The `--stability=alpha` option enables you to install the project from the last alpha release. Default value for the option is `stable` but there is no stable release yet.*
  
 ### 2. Create docker-compose.yml file
 Create `docker-compose.yml` from template [`docker-compose.yml.dist`](../../project-base/docker/conf/docker-compose.yml.dist).
--- a/docs/installation/installation-using-docker-linux.md
+++ b/docs/installation/installation-using-docker-linux.md
@@ -13,11 +13,10 @@ composer create-project shopsys/project-base --stability=alpha --no-install --ke
 cd project-base
 ```
 
-*Note: The `--no-install` option disables installation of the vendors - this will be done later in the Docker container.*
-
-*Note: The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
-
-*Note: The `--stability=alpha` option enables you to install the project from the last alpha release. Default value for the option is `stable` but there is no stable release yet.*
+*Notes:* 
+- *The `--no-install` option disables installation of the vendors - this will be done later in the Docker container.*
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Looks good to me.

Just a note to "minimum-stability: alpha/dev/beta" notes. In my experience it's not much spread in the PHP community. Why? The most common use case for these is for **project maintainers** to test pre-release versoin ("alpha/dev/beta").

On the other side, I always have to explain **cotributors** what it is.

For that reasons, I'd personally use syntax that is commonly used by every composer user, not just small marginal group, and that is version tagging:

```
composer create-project symfony/demo v4.0.0
```

That can be easily changed any other version. Just copy paste the tag and it's there:

```
composer create-project symfony/demo v4.1.0-dev
composer create-project symfony/demo dev-master
```

Also, when instead of using "minimum-stability: alpha/beta/rc", you usually want to **use specific version, because these version can contain BC breaks** and you don't want to break you application on "composer update".

Take these notes just as inspiration for the future. I think this PR is fine to merge at the moment.

docs/installation/installation-using-docker-linux.md
```diff
@@ -16,6 +16,8 @@ cd project-base
 *Note: The `--no-install` option disables installation of the vendors - this will be done later in the Docker container.*
 
 *Note: The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
+
+*Note: The `--stability=alpha` option enables you to install the project from the last alpha release. Default value for the option is `stable` but there is no stable release yet.*
```
simara-svatopluk: Consider reformulating notes into one:

*Notes:*
* *The `--no-install` option disables installation of the vendors - this will be done later in the Docker container.*
* *The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
* *The `--stability=alpha` option enables you to install the project from the last alpha release. Default value for the option is `stable` but there is no stable release yet.*

Also in other files.

If you will think the current state is fine, then don't do any changes.

TomasVotruba: What exactly is the change? Diff would be nice.

simara-svatopluk: I have a feeling that three notes are easier to read when they are a list. Also the triple *Note:* is reduced to one

TomasVotruba: Ah, so only design point.

vitek-rostislav:
Hi guys, thanks for your notes :+1: 

@TomasVotruba, if we were using specific version number in installation guide, we would have to change it with every release, I do not want that. But thanks for mentioning the option, we could consider this approach when releasing stable. About these BC breaks - the instructions are only for the first installation of the project, in `composer.json` we require specific versions of course to prevent unwanted application breaks during the `composer update`

@simara-svatopluk, I do agree with you, the list is more readable so I have added new commit :+1: is it ok with you now?

LukasHeinz:
Business validation done - for me it is OK. 

However I am not quite happy that there ale several files with exact same information. If we need to change something it is more work and some places can be missed. There is no chance to put those information into one place i.e. remove step one of each installation guide into some general "first step of installation"?

vitek-rostislav:
@LukasHeinz thanks. I think if we moved 1st step into yet another page and linked it from particular guides, there would too much "clicking" back and forth through docs during the installation, I do not like that. However, we could investigate a possibility of including `md` files in each other, that would do the trick :+1:

PR #143, opened 2018-05-28
[US-3593] Shopsys framework now uses latest version of Shopsys coding standards

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| ...
|New feature| Yes
|BC breaks| Yes
|Fixes issues| -
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #165, opened 2018-05-28
[US-2652] Different approach to multidomain entities

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Review of designed solution
|New feature| No
|BC breaks| Yes 


MattCzerner:
- BrandData now has responsibility for SEO properties
- BrandFormType now has reponsibility of SEOP properties
- BrandEditFormType deleted
- BrandEditData* deleted because it is useless now
- BrandDOmainFactory deleted since you can access brandDomain through entity or BrandData

composer.json
```diff
@@ -111,7 +111,7 @@
     "phpunit/phpunit": "^7.0",
     "presta/sitemap-bundle": "^1.5.3",
     "prezent/doctrine-translatable": "^1.2",
-    "prezent/doctrine-translatable-bundle": "^1.0.3",
+    "prezent/doctrine-translatable-bundle": "1.0.3",
```
PetrHeinz: What was the error and in which version of prezent/doctrine-translatable-bundle? You should definitely elaborate in the commit message.

packages/framework/src/DataFixtures/Demo/BrandDataFixture.php
```diff
@@ -65,6 +65,14 @@ public function load(ObjectManager $manager)
...
                 'en' => 'This is description of brand ' . $brandData->name . '.',
             ];
+
+            $brandData->seoTitles[1] = null;
+            $brandData->seoTitles[2] = null;
+            $brandData->seoMetaDescriptions[1] = null;
+            $brandData->seoMetaDescriptions[2] = null;
+            $brandData->seoH1s[1] = null;
+            $brandData->seoH1s[2] = null;
+
```
PetrHeinz: Demo data fixtures are executed before second domain is created, you should add the nulls only for Domain::FIRST_DOMAIN_ID and add the second domain in DemoMultidomain (or let it be automatically created during create-demo-data phing target)

packages/framework/src/Form/Admin/Product/Brand/BrandFormType.php
```diff
@@ -3,24 +3,84 @@
...
+        SeoSettingFacade $seoSettingFacade
+    ) {
+        $this->domain = $domain;
+        $this->seoSettingFacade = $seoSettingFacade;
+    }
+
     /**
      * @param \Symfony\Component\Form\FormBuilderInterface $builder
      * @param array $options
+     * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
```
PetrHeinz: Don't forget to remove these annotations when https://github.com/shopsys/shopsys/pull/143 is merged. :)

packages/framework/src/Model/Product/Brand/BrandDataFactory.php
```diff
@@ -51,9 +65,13 @@ public function createFromBrand(Brand $brand)
             $brandData->descriptions[$translation->getLocale()] = $translation->getDescription();
         }
 
-        foreach ($brandDomains as $brandDomain) {
+        foreach ($brand->getDomains() as $brandDomain) {
```
PetrHeinz: Is it necessary to make the getter public? It might be better to keep the concept of BrandDomains inside the Brand itself. You can iterate over $this->domain->getAll(). See that you access the $brandDomain here only in order to get its domain ID.

packages/framework/src/Model/Product/Brand/BrandFacade.php
```diff
@@ -110,19 +99,16 @@ public function create(BrandEditData $brandEditData)
...
     {
         $domains = $this->domain->getAll();
         $brand = $this->brandRepository->getById($brandId);
-        $brandData = $brandEditData->getBrandData();
         $brand->edit($brandData);
         $this->imageFacade->uploadImage($brand, $brandData->image->uploadedFiles, null);
-
-        $this->refreshBrandDomains($brand, $brandEditData);
-
+        $this->em->persist($brand);
```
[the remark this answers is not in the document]
PetrHeinz: Entity cannot be responsible for persisting its secondary entities, because it doesn't have access to EntityManager. Entity is rather responsible just for creating the secondary entities, facade is responsible for persisting the created/updated entity so it starts the persistence cascade.

It's probably just semantics, I just want to be on the same page :)

packages/framework/src/Model/Transport/Transport.php
```diff
@@ -244,6 +254,44 @@ public function setPosition($position)
...
+     */
+    public function getDomains()
+    {
+        return $this->domains;
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Transport\TransportData $transportData
+     */
+    private function setDomains(TransportData $transportData)
```
PetrHeinz: There should be no private methods in entities.

packages/framework/src/Model/Transport/Transport.php
```diff
@@ -244,6 +254,44 @@ public function setPosition($position)
...
+    {
+        foreach ($transportData->domains as $domainId) {
+            try {
+                $paymentDomain = $this->getTransportDomain($domainId);
+            } catch (TransportDomainNotFoundException $e) {
+                $paymentDomain = new TransportDomain($this, $domainId);
+            }
+            $this->domains[$domainId] = $paymentDomain;
```
PetrHeinz: TransportDomain is not an entity for multidomain attributes (similar to ProductDomain), but a optional association with a domain (similar to PaymentDomain). This means, that you should be able to remove the association to a domain (when trying to switch the transport from the first domain to the second, for example), but the setDomains method doesn't allow it, it only adds new Transport domains.

packages/framework/src/Model/Product/ProductRepository.php
```diff
@@ -165,10 +157,12 @@ public function addTranslation(QueryBuilder $queryBuilder, $locale)
...
      */
-    public function addDomain(QueryBuilder $queryBuilder, $domainId)
+    private function addDomain(QueryBuilder $queryBuilder, $domainId)
     {
-        $queryBuilder->join(ProductDomain::class, 'pd', Join::WITH, 'pd.product = p AND pd.domainId = :domainId');
-        $queryBuilder->setParameter('domainId', $domainId);
+        $queryBuilder->addSelect('pd')
+            ->join('p.domains', 'pd', Join::WITH, 'pd.domainId = :domain');
+
+        $queryBuilder->setParameter('domain', $domainId);
```
TomasLudvik: I do not understand changes in this method

packages/framework/src/Resources/views/Admin/Content/Product/visibilityItems.html.twig
```diff
@@ -1,12 +1,13 @@
 <ul class="list-domain">
-    {% for productDomain in productDetail.productDomainsIndexedByDomainId %}
+    {% for product in productDetail.product %}
```
TomasLudvik: Are you now iterating products instead of domains?

packages/framework/src/DataFixtures/DemoMultidomain/CategoryRootDataFixture.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
TomasLudvik: There is no description why is this class added

TomasLudvik: Like in CM

composer.json
TomasLudvik: You do not need to use this anymore. This problem has been already fixed. Please remove this commit.

project-base/tests/ShopBundle/Database/Model/Order/OrderTransportAndPaymentTest.php
```diff
@@ -86,7 +83,7 @@ public function testVisibleTransportHiddenPayment()
         $transport = new Transport(new TransportData(['cs' => 'transportName', 'en' => 'transportName'], $vat, [], [], false));
         $transportDomain = new TransportDomain($transport, 1);
         $payment = new Payment(new PaymentData(['cs' => 'paymentName', 'en' => 'paymentName'], $vat, [], [], true));
```
TomasLudvik: You should set domain here

packages/framework/src/Model/Product/Brand/Brand.php
```diff
@@ -91,6 +102,75 @@ protected function createTranslation()
...
+        return $this->domains;
+    }
+
+    /**
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Product\Brand\BrandDomain
+     */
+    protected function getBrandDomain(int $domainId)
+    {
+        foreach ($this->domains as $domain) {
```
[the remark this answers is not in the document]
PetrHeinz: The ProductDomains are not necessarily indexed by domain ID (and it's probably unwise to index it in such a way in https://github.com/shopsys/shopsys/pull/165/files/0ff35b922475cab1d02d6b2ce9a3b452d4906834#diff-ce1e32a3065400b38b9b2cac44e743e3R116). That's because Doctrine will hydrate the property as a unordered collection - it won't even keep the order of the array, let alone the indexes. Indexes are not persisted in any way by the Doctrine.

composer.json
MattCzerner: prezent released new version with fix of that error, ill drop this commit.

packages/framework/src/DataFixtures/Demo/BrandDataFixture.php
MattCzerner: I created DemoDomainFixture. I think that it belongs into create-demo-data but there is plan which will change behaviour of this target, so I will leave it to implementation of that US

packages/framework/src/Form/Admin/Product/Brand/BrandFormType.php
MattCzerner: I will delete this with rebase on current master. I dont want my build to fail at this time.

packages/framework/src/Model/Product/Brand/BrandDataFactory.php
MattCzerner: I agree, now are all domain ids looped in this foreach

packages/framework/src/Model/Product/ProductRepository.php
MattCzerner: ProductRepository does some visibility calculations in which they control ProdudctDomain properties, thats why it is needed to be added. It is same concept as it is used for Translations.

packages/framework/src/DataFixtures/DemoMultidomain/CategoryRootDataFixture.php
MattCzerner: I will add it at the final rebase

project-base/tests/ShopBundle/Database/Model/Order/OrderTransportAndPaymentTest.php
MattCzerner: I refactored it entirely.

packages/framework/src/Model/Product/Product.php
```diff
@@ -893,6 +905,109 @@ protected function setTranslations(ProductData $productData)
         }
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\ProductData $productData
+     */
+    private function setDomains($productData)
```
PetrHeinz: Facades should have protected methods as they are supposed to be extended by inheritance.

PetrHeinz: Also, the parameter should be type-hinted.

packages/framework/src/Model/Product/Product.php
```diff
@@ -893,6 +905,109 @@ protected function setTranslations(ProductData $productData)
...
+    public function getSeoMetaDescription($domainId)
+    {
+        return $this->getProductDomain($domainId)->getSeoMetaDescription();
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @return string
+     */
+    public function getSeoTitleForHtml(Domain $domain)
```
PetrHeinz: Should this logic be in entity? In Brand, logic such as this was in the template.

packages/framework/src/Model/Product/Product.php
```diff
@@ -893,6 +905,109 @@ protected function setTranslations(ProductData $productData)
...
+    public function getSeoTitle(int $domainId)
+    {
+        return $this->getProductDomain($domainId)->getSeoTitle();
+    }
+
+    /**
+     * @param int $domainId
+     * @return string|null
+     */
+    public function getSeoMetaDescription($domainId)
```
PetrHeinz: not type-hinted

packages/framework/src/Model/Product/Product.php
```diff
@@ -893,6 +905,109 @@ protected function setTranslations(ProductData $productData)
...
+            $productDomain->setSeoMetaDescription($productData->seoMetaDescriptions[$domainId]);
+            $productDomain->setDescription($productData->descriptions[$domainId]);
+            $productDomain->setShortDescription($productData->shortDescriptions[$domainId]);
+        }
+    }
+
+    /**
+     * @return \Shopsys\FrameworkBundle\Model\Product\ProductDomain[]|\Doctrine\Common\Collections\ArrayCollection
+     */
+    public function getDomains()
```
PetrHeinz: In Brand, method such as this wasn't public. The idea was that BrandDomain should look like an implementation detail from the outside.

packages/framework/src/Model/Product/Product.php
```diff
@@ -893,6 +905,109 @@ protected function setTranslations(ProductData $productData)
...
+    {
+        return $this->domains;
+    }
+
+    /**
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Product\ProductDomain
+     */
+    public function getProductDomain(int $domainId)
```
PetrHeinz: In Brand, method such as this wasn't public. The idea was that BrandDomain should look like an implementation detail from the outside.

packages/framework/src/Model/Product/ProductDataFactory.php
```diff
@@ -124,12 +124,23 @@ public function createFromProduct(Product $product)
...
         $translations = $product->getTranslations();
+        $domains = $product->getDomains();
         $names = [];
         $variantAliases = [];
         foreach ($translations as $translation) {
             $names[$translation->getLocale()] = $translation->getName();
             $variantAliases[$translation->getLocale()] = $translation->getVariantAlias();
         }
+
+        foreach ($domains as $domain) {
```
PetrHeinz: We usually call instances of DomainConfig a $domain, these objects should be called $productDomains imo to avoid confusion.

packages/framework/src/Model/Product/ProductRepository.php
```diff
@@ -162,10 +154,12 @@ public function addTranslation(QueryBuilder $queryBuilder, $locale)
      * @param \Doctrine\ORM\QueryBuilder $queryBuilder
      * @param int $domainId
      */
-    public function addDomain(QueryBuilder $queryBuilder, $domainId)
+    private function addDomain(QueryBuilder $queryBuilder, $domainId)
```
PetrHeinz: Generally, there shouldn't be private methods in repositories as they are supposed to be extended by inheritance.

packages/framework/src/Model/Product/ProductRepository.php
```diff
@@ -162,10 +154,12 @@ public function addTranslation(QueryBuilder $queryBuilder, $locale)
...
     {
-        $queryBuilder->join(ProductDomain::class, 'pd', Join::WITH, 'pd.product = p AND pd.domainId = :domainId');
-        $queryBuilder->setParameter('domainId', $domainId);
+        $queryBuilder->addSelect('pd')
+            ->join('p.domains', 'pd', Join::WITH, 'pd.domainId = :domain');
```
PetrHeinz: Why did you rename the parameter `domainId` to `domain`?

packages/framework/src/DataFixtures/Base/CategoryRootDataFixture.php
```diff
@@ -20,36 +19,35 @@ class CategoryRootDataFixture extends AbstractReferenceFixture implements Depend
...
 
     /**
      * @param \Doctrine\Common\Persistence\ObjectManager $manager
      */
     public function load(ObjectManager $manager)
     {
-        $rootCategory = $this->categoryFactory->create(new CategoryData());
+        $categoryData = $this->categoryDataFactory->createDefault();
+        $categoryData->enabled = [
+            Domain::FIRST_DOMAIN_ID => true,
```
PetrHeinz: Shouldn't this behavior be a part of CategoryDataFactory::createDefault()?

packages/framework/src/DataFixtures/DemoMultidomain/CategoryDataFixture.php
```diff
@@ -128,6 +128,7 @@ private function editCategoryOnDomain2($referenceName, $descriptionDomain2)
         $categoryData = $this->categoryDataFacade->createFromCategory($category);
         $domainId = 2;
         $categoryData->descriptions[$domainId] = $descriptionDomain2;
+        $categoryData->enabled[$domainId] = true;
```
PetrHeinz: Is this necessary? The create-domain-data should take care of non-nullable multidomain parameters automatically.

packages/framework/src/Migrations/Version20180611215810.php
```diff
@@ -0,0 +1,24 @@
...
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20180611215810 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->sql('ALTER TABLE category_domains RENAME COLUMN hidden TO enabled');
```
PetrHeinz: This looks like it negates all current category visibility.

packages/framework/src/Model/Category/Category.php
```diff
@@ -210,11 +214,114 @@ protected function setTranslations(CategoryData $categoryData)
         }
     }
 
+    /**
+     * @param int $domainId
+     * @return string|null
+     */
+    public function getSeoTitle($domainId)
```
PetrHeinz: In previous entities these methods were always type-hinted as `int`.

packages/framework/src/Model/Category/Category.php
```diff
@@ -210,11 +214,114 @@ protected function setTranslations(CategoryData $categoryData)
...
+     */
+    public function getDescription($domainId)
+    {
+        return $this->getCategoryDomain($domainId)->getDescription();
+    }
+
+    /**
+     * @return \Doctrine\Common\Collections\ArrayCollection|\Shopsys\FrameworkBundle\Model\Category\CategoryDomain[]
+     */
+    public function getCategoryDomains()
```
PetrHeinz: In Brand, this method wasn't public.

packages/framework/src/Model/Category/Category.php
```diff
@@ -210,11 +214,114 @@ protected function setTranslations(CategoryData $categoryData)
...
+                $this->domains[$domainId] = $categoryDomain;
+            }
+        }
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Category\CategoryData $categoryData
+     * @return array
+     */
+    private function getDomainIdsForCategoryDomainData(CategoryData $categoryData)
```
PetrHeinz: Why not use just the `array_keys($categoryData->seoTitles)` as in Brand and Product? Provided data object should be valid - should have attributes set for every domain ID (for which the data object factory is responsible).

packages/framework/src/Component/Transformers/ShowOnDomainBooleanToIntegerTransformer.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Shopsys\FrameworkBundle\Component\Transformers;
+
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Symfony\Component\Form\DataTransformerInterface;
+
+class ShowOnDomainBooleanToIntegerTransformer implements DataTransformerInterface
```
PetrHeinz: This logic should be probably implemented right in the `DomainsType` - it is used in the three places only (Category, Payment, Transport). If it was just a collection of `CheckboxTypes`, the data would be in correct format without transformers.

PetrHeinz: The DomainsType can be as simple as this, without the need of a transformer or any changes to current form types or theme:

```php
<?php

namespace Shopsys\FrameworkBundle\Form;

use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
use Symfony\Component\Form\FormBuilderInterface;

class DomainsType extends AbstractType
{
    /**
     * @var \Shopsys\FrameworkBundle\Component\Domain\Domain
     */
    private $domain;

    /**
     */
    public function __construct(Domain $domain)
    {
        $this->domain = $domain;
    }

    {
        foreach ($this->domain->getAll() as $domainConfig) {
            $builder->add($domainConfig->getId(), CheckboxType::class, [
                'required' => false,
                'label' => $domainConfig->getName(),
            ]);
        }
    }
}

File: packages/framework/src/Model/Product/Brand/BrandDataFactory.php [hunk 28]
```diff
@@ -51,8 +65,11 @@ public function createFromBrand(Brand $brand)
             $brandData->descriptions[$translation->getLocale()] = $translation->getDescription();
         }
 
-        foreach ($brandDomains as $brandDomain) {
-            $domainId = $brandDomain->getDomainId();
+        $allDomainIds = $this->domain->getAllIds();
+        foreach ($allDomainIds as $domainId) {
+            $brandData->seoH1s[$domainId] = $brand->getSeoH1($domainId);
+            $brandData->seoTitles[$domainId] = $brand->getSeoTitle($domainId);
+            $brandData->seoMetaDescriptions[$domainId] = $brand->getSeoMetaDescription($domainId);
```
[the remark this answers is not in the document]
MattCzerner: I dont understand on which line they got overwrited?

packages/framework/src/Model/Product/Brand/Exception/BrandDomainNotFoundException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Shopsys\FrameworkBundle\Model\Product\Brand\Exception;
+
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+class BrandDomainNotFoundException extends NotFoundHttpException implements BrandException
```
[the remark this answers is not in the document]
MattCzerner: Okay, your right, now i throw 500 error and i added message into exception

packages/framework/src/Form/Admin/Product/Brand/BrandFormType.php
```diff
@@ -3,24 +3,83 @@
...
+        $this->seoSettingFacade = $seoSettingFacade;
+    }
+
     /**
      * @param \Symfony\Component\Form\FormBuilderInterface $builder
      * @param array $options
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $brand = $options['brand'];
```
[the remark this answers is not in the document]
MattCzerner: You are right, changed

packages/framework/src/DataFixtures/DemoMultidomain/BrandDataFixture.php
```diff
@@ -0,0 +1,115 @@
...
+
+        foreach ($brandReferencesNames as $referenceName => $brandName) {
+            $brand = $this->getReference($referenceName);
+            /* @var $brand \Shopsys\FrameworkBundle\Model\Product\Brand\Brand */
+
+            $brandData = $this->brandDataFactory->createFromBrand($brand);
+
+            $brandData->seoTitles[self::DOMAIN_ID] = null;
+            $brandData->seoMetaDescriptions[self::DOMAIN_ID] = null;
+            $brandData->seoH1s[self::DOMAIN_ID] = null;
```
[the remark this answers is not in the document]
MattCzerner: Well you are right that this is actually useless and its already done by BrandDataFactory. But i dont think that this should be used for translating the name, i think that Brand name is not translatable.

packages/framework/src/DataFixtures/Demo/BrandDataFixture.php
```diff
@@ -65,6 +65,11 @@ public function load(ObjectManager $manager)
                 'cs' => 'Toto je popis značky ' . $brandData->name . '.',
                 'en' => 'This is description of brand ' . $brandData->name . '.',
             ];
+
+            $brandData->seoTitles[1] = null;
+            $brandData->seoMetaDescriptions[1] = null;
+            $brandData->seoH1s[1] = null;
```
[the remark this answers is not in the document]
MattCzerner: These 3 lines are actually useless, because it is done by BrandDataFactory.

packages/framework/src/DataFixtures/DemoMultidomain/BrandDataFixture.php
```diff
@@ -0,0 +1,115 @@
...
+    const BRAND_HTC = 'brand_htc';
+    const BRAND_JURA = 'brand_jura';
+    const BRAND_LOGITECH = 'brand_logitech';
+    const BRAND_MICROSOFT = 'brand_microsoft';
+    const BRAND_SAMSUNG = 'brand_samsung';
+    const BRAND_SONY = 'brand_sony';
+    const BRAND_ORAVA = 'brand_orava';
+    const BRAND_OLYMPUS = 'brand_olympus';
+    const BRAND_HYUNDAI = 'brand_hyundai';
+    const BRAND_NIKON = 'brand_nikon';
```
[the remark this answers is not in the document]
MattCzerner: There is no reason, i just didnt do it. Fixed!

packages/framework/src/Migrations/Version20180611123750.php
```diff
@@ -0,0 +1,37 @@
...
+
+class Version20180611123750 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->sql('ALTER TABLE transport_domains ADD enabled BOOLEAN NOT NULL');
+        $this->sql('ALTER TABLE payment_domains ADD enabled BOOLEAN NOT NULL');
```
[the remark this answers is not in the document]
MattCzerner: i deleted NOT NULL condition and added queries that will migrate existing domain records to be enabled and all other records in translations will be inserted with enabled set on false.

packages/framework/src/Model/Product/Brand/BrandDataFactory.php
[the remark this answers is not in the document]
PetrHeinz: https://github.com/shopsys/shopsys/blob/97583067168a4a4754f1a2a0c0a528732b8678a7/packages/framework/src/Model/Product/Brand/BrandDataFactory.php#L80-L82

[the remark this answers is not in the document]
PetrHeinz: The overwrite was then removed in a15ef6c97e9440b597bc1cc7cc0f54387feaaf03, see history of the file: https://github.com/shopsys/shopsys/commits/mc-multidomain/packages/framework/src/Model/Product/Brand/BrandDataFactory.php

packages/framework/src/DataFixtures/Demo/BrandDataFixture.php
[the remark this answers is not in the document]
PetrHeinz: Yeah, deleting it would be much better :+1:

packages/framework/src/DataFixtures/DemoMultidomain/BrandDataFixture.php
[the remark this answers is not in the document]
PetrHeinz: So there is no need for this data fixture?

packages/framework/src/Model/Feed/Standard/StandardFeedItemRepository.php
```diff
@@ -51,7 +49,6 @@ public function getItems(DomainConfigInterface $domainConfig, $seekItemId, $maxR
             ->addSelect('a')->join('p.calculatedAvailability', 'a')
             ->addSelect('b')->leftJoin('p.brand', 'b')
             ->andWhere('p.variantType != :variantTypeMain')->setParameter('variantTypeMain', Product::VARIANT_TYPE_MAIN)
-            ->join(ProductDomain::class, 'pd', Join::WITH, 'pd.product = p.id AND pd.domainId = :domainId')
```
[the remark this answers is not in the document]
MattCzerner: Well since ProductDomain was not property of Product, there was need to join in query, or i dont understand the question right. But you are right, i added it there.

packages/framework/src/Model/Product/Product.php
MattCzerner: It is not facade, but i think that this rule applies to entities too. Fixed

packages/framework/src/Model/Product/Product.php
MattCzerner: Okay, moved to template and refactored the old implementation of getting seo title in brand since it was wierdly conditioned and it was incosistent with other implementations.

packages/framework/src/Model/Product/Product.php
MattCzerner: But, Brand has this method public, you need to have this public because of `ProductDataFactory` that needs these domains in `createFromBrand()` method, I thought of it and i dont see any other solution for not requiring these Domain entities

packages/framework/src/Model/Product/ProductRepository.php
MattCzerner: ups, fixed

packages/framework/src/Model/Product/Product.php
PetrHeinz: How about iterating over `Domain::getAllIds()` in the factory and accessing then accessing the attributes via the main entity - eg. `Product::getSeoTitle($domainId)`?

packages/framework/src/DataFixtures/Base/CategoryRootDataFixture.php
MattCzerner: well not really, `createDefault()` creates categoryData with enabled set to false. So this line makes sure that category will be enabled on domain

packages/framework/src/Migrations/Version20180611123750.php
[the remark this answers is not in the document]
MattCzerner: Actually i added default value into Entities

packages/framework/src/Model/Category/Category.php
MattCzerner: it was. as i said, i think that there is not other solution to make this method private

packages/framework/src/Model/Category/Category.php
MattCzerner: added type hint

packages/framework/src/Migrations/Version20180611215810.php
MattCzerner: Good point I added UPDATE migration that will solve this problem.

packages/framework/src/DataFixtures/Base/CategoryRootDataFixture.php
PetrHeinz: Why is it created with `false`s as default? Isn't more appropriate to create it with `true`s? On `master`, all domains are enabled by default when eg. creating a new category:

image

packages/framework/src/Migrations/Version20180611123750.php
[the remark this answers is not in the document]
PetrHeinz: You would still need to fix the migration - no matter what Doctrine says, these SQL queries will end up with an error...

Default values are not used very much in SSFW, are you sure about this approach?

packages/framework/src/DataFixtures/DemoMultidomain/BrandDataFixture.php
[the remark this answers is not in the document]
MattCzerner: Well, the fix of this is done via create-multidomain-data. in which i manually added all of our new domain entities, i know that this is not the best solution but i think that this is actually responsibility of other US.

packages/framework/src/DataFixtures/Base/CategoryRootDataFixture.php
MattCzerner: Okay, i think that i handled all of our enabled domain properties and it is set to be enabled as default, which made data fixtures much more easier.

packages/framework/src/DataFixtures/DemoMultidomain/CategoryDataFixture.php
MattCzerner: Not now.

packages/framework/src/Component/Transformers/ShowOnDomainBooleanToIntegerTransformer.php
MattCzerner: great

MattCzerner:
So, i think that i fixed everything that you asked me to.

I think that there is only 3 things to discuss. 

First thing is migrations, since i was developing and adding these new ORM things on the run, there was created many migrations. For example Version20180612230916 should be executed earlier then Version20180611123750, because i expect that while im migrating old Transport/Payment Domain data, there is already created unique Constraints, so i dont know what to do with it

Second is, that i added paragraph about domain entities into basics-about-model-architecture.md, so i need to review this

And at last, make these commit messages better, i dont want to do it now, since this CR is already hard so i would love to take care of this with you.

Build is okay, i think that everything is functional.

Except these commit messages, i dont care if you find something and fix it right away, i would just like to see in what i did mistake.

packages/framework/src/DataFixtures/DemoMultidomain/BrandDataFixture.php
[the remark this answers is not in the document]
PetrHeinz: Please, add a note about these changes to *US-4163* so it's not forgotten.

docs/introduction/basics-about-model-architecture.md
```diff
@@ -67,15 +67,15 @@ class Product
 }
 ```
 
-### Factory
+## Factory
```
PetrHeinz: It was indented on purpose - it's a factory for entities, so it should be a child of the *Entity* heading: 
- Basics about the model
- Entity
  - Example
  - Factory
    - Example
- Repository
  - Example
- Facade
  - Example
- Service
  - Example
- Cooperation of layers

It would make sense to either leave it indented, or at least renaming it to *Entity Factory*.

PetrHeinz:
I have rebased and squashed your commits, you can see the commits in https://github.com/shopsys/shopsys/compare/dc33388c8...mc-multidomain-original-after-squashing, so you can see any further changes. I will have to rebase it a few times more and it would be a pain to do it in fixup commits...

PetrHeinz:
I thoroughly checked the changes, fixed a few things and written proper commit messages and a changelog. Please, check the changes (I would suggest checking out mc-multidomain and soft resetting to mc-multidomain-original-after-squashing to view changes locally).

I do not think we should implement the implicit current domain ID in the scope of this change, but please, create a user story for that.

Send it straight to testing after checking the changes I made :+1:

project-base/tests/ShopBundle/Database/Model/Category/CategoryDomainTest.php
```diff
@@ -0,0 +1,119 @@
...
+use Shopsys\FrameworkBundle\Model\Category\CategoryFactory;
+use Tests\ShopBundle\Test\DatabaseTestCase;
+
+class CategoryDomainTest extends DatabaseTestCase
+{
+    const FIRST_DOMAIN_ID = 1;
+    const SECOND_DOMAIN_ID = 2;
+    const DEMONSTRATIVE_SEO_TITLE = 'Demonstrative seo title';
+    const DEMONSTRATIVE_SEO_META_DESCRIPTION = 'Demonstrative seo description';
+    const DEMONSTRATIVE_SEO_H1 = 'Demonstrative seo H1';
```
PetrHeinz: I would personally avoid using constants for these testing strings, but it's OK to do so.

project-base/tests/ShopBundle/Database/Model/Category/CategoryDomainTest.php
```diff
@@ -0,0 +1,119 @@
...
+        $categoryData->enabled[self::FIRST_DOMAIN_ID] = true;
+
+        $category = $this->categoryFactory->create($categoryData);
+
+        $refreshedCategory = $this->getRefreshedCategoryFromDatabase($category);
+
+        $this->assertTrue($refreshedCategory->isEnabled(self::FIRST_DOMAIN_ID));
+    }
+
+    public function testCreateCategoryNotEnabledOnDomain()
```
PetrHeinz: not enabled = disabled

project-base/tests/ShopBundle/Database/Model/Category/CategoryDomainTest.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public function testCreateCategoryWithDifferentVisibilityOnDomains()
+    {
+        $categoryData = $this->categoryDataFactory->createDefault();
+
+        $categoryData->enabled = [
+            self::FIRST_DOMAIN_ID => true,
+            self::SECOND_DOMAIN_ID => false,
+        ];
```
PetrHeinz: Here, you are rewriting the whole array, in the other methods you were just rewriting a single item. This inconsistency may have strange consequences when running with different number of domains.

You can achieve what you want without being inconsistent:

```php
$categoryData->enabled[self::FIRST_DOMAIN_ID] = true;
$categoryData->enabled[self::SECOND_DOMAIN_ID] = false;
```

In Payment / Transport tests you even change the number of items - you use single-item arrays.

Not sure if it's a problem that has to be solved...

project-base/tests/ShopBundle/Database/Model/Product/ProductDomainTest.php
```diff
@@ -0,0 +1,94 @@
...
+     * @var \Doctrine\ORM\EntityManager
+     */
+    private $em;
+
+    public function setUp()
+    {
+        parent::setUp();
+        $this->productDataFactory = $this->getContainer()->get(ProductDataFactory::class);
+        $this->productFactory = $this->getContainer()->get(ProductFactory::class);
+        $this->availabilityFacade = $this->getContainer()->get(AvailabilityFacade::class);
```
PetrHeinz: AvailabilityFacade is never used, persistent reference is used instead.

project-base/tests/ShopBundle/Database/Model/Category/CategoryDomainTest.php
MattCzerner: done

project-base/tests/ShopBundle/Database/Model/Category/CategoryDomainTest.php
MattCzerner: I solved it

PetrHeinz:
@MattCzerner Please, mention the merge commit (c77083c3c) when you merge PRs manually, otherwise it seem,s like it wasn't merged at all.

PR #171, opened 2018-05-29
[US-3323] Update to twig 2.x

TomasLudvik:
| Q             | A
| ------------- | ---
|New feature| No
|BC breaks| Yes
|Fixes issues| No

### DIFF
--- a/packages/framework/src/Model/Feed/FeedXmlWriter.php
+++ b/packages/framework/src/Model/Feed/FeedXmlWriter.php
@@ -4,7 +4,7 @@ namespace Shopsys\FrameworkBundle\Model\Feed;
 
 use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
 use Twig_Environment;
-use Twig_Template;
+use Twig_TemplateWrapper;
 
 class FeedXmlWriter
 {
@@ -25,7 +25,8 @@ class FeedXmlWriter
      */
     public function writeBegin(DomainConfig $domainConfig, $feedTemplatePath, $targetFilepath)
     {
-        $twigTemplate = $this->twig->loadTemplate($feedTemplatePath);
+        $twigTemplate = $this->twig->load($feedTemplatePath);
+
         $renderedBlock = $this->getRenderedBlock($twigTemplate, 'begin', ['domainConfig' => $domainConfig]);
         file_put_contents($targetFilepath, $renderedBlock);
     }
@@ -37,7 +38,7 @@ class FeedXmlWriter
      */
     public function writeEnd(DomainConfig $domainConfig, $feedTemplatePath, $targetFilepath)
     {
-        $twigTemplate = $this->twig->loadTemplate($feedTemplatePath);
+        $twigTemplate = $this->twig->load($feedTemplatePath);
         $renderedBlock = $this->getRenderedBlock($twigTemplate, 'end', ['domainConfig' => $domainConfig]);
         file_put_contents($targetFilepath, $renderedBlock, FILE_APPEND);
     }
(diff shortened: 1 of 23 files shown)


packages/framework/src/Resources/views/Admin/Content/BestsellingProduct/list.html.twig
```diff
@@ -4,10 +4,12 @@
 {% block h1 %}{{ 'Bestsellers'|trans }}{% endblock %}
 
 {% block main_content %}
+    {% import _self as self %}
```
vitek-rostislav: does this kind of change belong into this commit?

packages/framework/src/Model/Feed/FeedXmlWriter.php
```diff
@@ -68,18 +69,15 @@ public function writeItems(array $items, DomainConfig $domainConfig, $feedTempla
...
 
     /**
-     * @param \Twig_Template $twigTemplate
+     * @param \Twig_TemplateWrapper $twigTemplate
      * @param string $name
      * @param array $parameters
      * @return string
      */
-    private function getRenderedBlock(Twig_Template $twigTemplate, $name, array $parameters = [])
+    private function getRenderedBlock(Twig_TemplateWrapper $twigTemplate, $name, array $parameters = [])
```
vitek-rostislav: it would be nice to rename `$twigTemplate` to `$twigTemplateWrapper`

packages/framework/src/Resources/views/Admin/Content/BestsellingProduct/list.html.twig
```diff
@@ -1,4 +1,5 @@
 {% extends '@ShopsysFramework/Admin/Layout/layoutWithPanel.html.twig' %}
+{% import _self as self %}
```
vitek-rostislav: This is little bit weird...I would appreciate at least link to some more information about this but I could not find any relevant :slightly_frowning_face: But maybe you could write a little bit more about it in the commit message - "when there is a macro defined in a template, it can not be used via '__self' in the same template anymore, it must be imported" or something like that

packages/framework/src/Resources/config/services_test.yml
```diff
@@ -61,10 +61,20 @@ services:
 
   Shopsys\FrameworkBundle\Component\Domain\Config\DomainsConfigLoader: ~
 
+  Shopsys\FrameworkBundle\Component\Domain\Domain:
```
vitek-rostislav: Why is this commit in this branch? It seems like it has nothing to do with the Twig update, am I right? Could you do that in a separate pull request with proper description please?

packages/framework/src/Model/Feed/FeedXmlWriter.php
TomasVotruba: @vitek-rostislav ...or you can cover these situations by a fixer: https://github.com/symplify/codingstandard#property-name-should-match-its-type-if-possible

packages/framework/src/Resources/config/services_test.yml
TomasLudvik: Removed

packages/framework/src/Resources/views/Admin/Content/BestsellingProduct/list.html.twig
TomasLudvik: updated

packages/framework/src/Resources/views/Admin/Content/BestsellingProduct/list.html.twig
TomasLudvik: No it does not :(

Issue #158, opened 2018-05-28
Admin: Validation for free shipping is inconsistent


jakubhavlak:
An inline message in front of the field appears instead of an explanation icon:
![image](https://user-images.githubusercontent.com/31988844/40614458-2f3728a6-6283-11e8-8607-a0a704543246.png)

LukasHeinz:
This issue was resolved in https://github.com/shopsys/shopsys/commit/f15e42f7fc5a9aba9ff388a507bc0df46d3b5fca

PR #176, opened 2018-05-31
[BG-2820] Admin: Validation for free shipping is inconsistent

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| .
|New feature| No 
|BC breaks| No
|Fixes issues| #158

### DIFF
--- a/packages/framework/src/Resources/views/Admin/Content/TransportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Content/TransportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
@@ -35,7 +35,10 @@
                     {% set domainName = domain.getDomainConfigById(domainId).getName() %}
                     <tr id="{{ priceLimitForm.vars.id }}" class="js-free-transport-and-payment-price-limit" data-domain-id="{{ domainId }}">
                         {% if isMultidomain() %}
-                            <td>{{ domainName }}</td>
+                        <td class="in-overlay">
+                            {{ form_errors(priceLimitForm.priceLimit) }}
+                            {{ domainName }}
+                        </td>
                         {% endif %}
                         <td>
                             {{ form_widget(priceLimitForm.enabled, { attr: { class: 'js-free-transport-and-payment-price-limit-enabled' } } ) }}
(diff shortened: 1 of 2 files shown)


packages/framework/src/Resources/views/Admin/Content/TransportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
```diff
@@ -35,12 +35,15 @@
                     {% set domainName = domain.getDomainConfigById(domainId).getName() %}
                     <tr id="{{ priceLimitForm.vars.id }}" class="js-free-transport-and-payment-price-limit" data-domain-id="{{ domainId }}">
                         {% if isMultidomain() %}
-                            <td>{{ domainName }}</td>
+                            <td>
```
vitek-rostislav: this was not necessary, at least it does not belong into this commit but never mind :slightly_smiling_face:

packages/framework/src/Resources/views/Admin/Content/TransportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
```diff
@@ -35,12 +35,15 @@
...
+                                {{ domainName }}
+                            </td>
                         {% endif %}
                         <td>
                             {{ form_widget(priceLimitForm.enabled, { attr: { class: 'js-free-transport-and-payment-price-limit-enabled' } } ) }}
                         </td>
-                        <td class="form-inline">
+                        <td class="form-inline in-overlay">
```
vitek-rostislav: Why? Is this necessary? I have tested it and saw the behavior without the `in-overlay` class but I am not sure if this is a proper solution, maybe you could discuss it with @kulincz

boris-brtan: seems that without position relative it is outside of the box
image

packages/framework/src/Resources/views/Admin/Content/TransportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
boris-brtan: accepted

boris-brtan: form_error block was added into free shipping and payment form page        

- to have the same error components in admin forms

LukasHeinz:
BV done, ok for merge

PR #182, opened 2018-06-01
optimize check of modifications in cart

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Each request on FE call unnecessary flush with computing change set for all entities in identity map.
|New feature| No
|BC breaks| No
|Fixes issues| ...

I did not update CHANGELOG.md, I did not know link to this futured pull request :( I am demotivated about updating of it twice (befor pull request and after creating...)


Miroslav-Stopka:
Hi Peter,

thank you for your contributing. Usage of function flush() with argument is deprecated I think, but I agree with you that this form is used on another places in framework too and there will be propably some refactoring of usage of this function in the near future. So in this case Im ok with your solution :) 

Now we will run some tests and I will inform you about results.

LukasHeinz:
Please someone add entry to changelog. It is important for everyone to be able find everything that changed in next version. Thank you.

PR #183, opened 2018-06-04
services_test.yml: added services used in tests so they are marked as public

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| services loaded from service container directly must be public
|New feature| No
|BC breaks| No
|Fixes issues| No

### DIFF
--- a/packages/framework/src/Resources/config/services_test.yml
+++ b/packages/framework/src/Resources/config/services_test.yml
@@ -61,6 +61,16 @@ services:
 
   Shopsys\FrameworkBundle\Model\Feed\Delivery\DeliveryFeedItemRepository: ~
 
+  Shopsys\FrameworkBundle\Component\Doctrine\DatabaseSchemaFacade:
+    arguments:
+      - '%shopsys.default_db_schema_filepath%'
+
+  Shopsys\FrameworkBundle\Component\Domain\Domain:
+    factory: ['@Shopsys\FrameworkBundle\Component\Domain\DomainFactory', create]
+    arguments:
+      - '%shopsys.domain_config_filepath%'
+      - '%shopsys.domain_urls_config_filepath%'
+
   Shopsys\FrameworkBundle\Component\Domain\Config\DomainsConfigLoader: ~
 
   Shopsys\FrameworkBundle\Component\Domain\DomainDataCreator: ~


TomasVotruba:
So for every similar services that is tested it neededs to be passed to `services_tests.yml` with all the setup (arguments, tags, method calls)? What if original `services.yml` is changed for one service but sb forgets to update `services_tests.yml`?

Tip for better public services for tests: https://www.tomasvotruba.cz/blog/2018/05/17/how-to-test-private-services-in-symfony/#quot-what-about-compiler-pass-quot

TomasLudvik:
Hello @MattCzerner,

all services used in tests must be marked as public because they are requested from the service container. I thought that my explanation is sufficient. What change do you suggest? Can you be more specific please?

PetrHeinz:
I would love the solution @TomasVotruba posted. It seems like the exact solution to our problem. Of cource, the `services_test.yml` would be still there, but it would contain only services that need to be reconfigured - eg. using the `DomainFactoryOverwritingDomainUrl`. 

Also I'm not sure if I like the detection of PHPUnit execution though - could it be linked to the parameter `kernel.environment` being `test` instead?

TomasVotruba:
What needs to be done here?

The code change is too simple to deserve to be opened for 21 days.

PetrHeinz:
@TomasLudvik - do you think it's worth it implementing the compiler pass? I think it would mean a drastic reduction of our `services_test.yml` and making our test more robust.

I'm willing to do it myself in a separate PR if you don't want to do it yourself. I think it will not take more than an hour, even with a review.

TomasLudvik:
@PetrHeinz well I think it is not necessary because when Symfony 4.1 is released (and implemented in framework) we have to simply delete services_test.yml and that's it... Right now I have no problem to add it there.

PetrHeinz:
@TomasLudvik: It may take months before we upgrade to Symfony 4.1, should we really wait?

@MattCzerner is right about the commit message not being descriptive enough. Are those services really needed in tests? Why don't the tests fail without these changes?

PetrHeinz:
I did try to implement a compiler pass that would set all services as public for test environment (in a similar fashion as the way @TomasVotruba recommended), but I failed to do so in a reasonable time. I've encountered problems with services that are autoregistered as private services, but cannot be instantiated by the DIC (eg. migrations). Furthermore, there where some problems with services outside the framework bundle.

It's not as easy as I've originally thought. But maybe there is some easy way out - some hack I haven't thought of.

You can take a look at the WIP in [ph-services-test](https://github.com/shopsys/shopsys/compare/ph-services-test). It may help during resolving this PR.

MattCzerner:
Hello, I will try to make this clear and finally close this Issue, after consultation with @PetrHeinz we have come to conclussion that this PR does not make sense to us. This PR does not solve any problem and just adds two classes into `services_test.yml`. Our final statement is that, these 2 services are not needed if tests does not fall. I created US for compiler pass making services public for test which has been recommended to us by @TomasVotruba. I tried to create this implementation during some time and ran into simillar problems as @PetrHeinz and since this is not needed right now in Shopsys Framework i feel like spending time on this implementation is pointless.

TomasVotruba:
Is there a link to failing build? I'd like to see what errors do you have. I don't get it from the text

MattCzerner:
@TomasVotruba, at the moment there is no CI linked to github and we got it privatly runned on jenkins, i made screenshot of error output from jenkins. 
screenshot from 2018-07-13 13-45-46

TomasVotruba:
Oh, it looks unrelated.

This is caused by autodiscovery of exception classes as services.

Those should be excluded, e.g. https://github.com/Symplify/Symplify/blob/master/packages/EasyCodingStandard/src/config/services.yml#L11

TomasLudvik:
Hello,

sadly I have also discussed this with @PetrHeinz and there has been some misunderstanding. These two services cause Deprecation calls in Acceptance tests as can be seen on the screenshot. I am sorry that I did not mention this earlier.

screen_deprecation

PR #185, opened 2018-06-04
[US-3636] Maintaining the order of database migration

TomasLudvik:
| Q             | A
| ------------- | ---
|New feature| Yes
|BC breaks| No
|Fixes issues| No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /docker-sync.yml
 /project-base/composer.lock
 /project-base/package-lock.json
+/project-base/migrations-lock.yml
--- a/build.xml
+++ b/build.xml
@@ -37,6 +37,20 @@
     <target name="standards" depends="phplint,ecs,phpstan,twig-lint,yaml-lint,eslint-check,standards-packages" description="Checks coding standards in both project-base and packages."/>
     <target name="standards-diff" depends="phplint-diff,ecs-diff,phpstan,twig-lint-diff,yaml-lint,eslint-check-diff,standards-packages" description="Checks coding standards on changed files in project-base and all files in packages."/>
 
+    <target name="clean" description="Cleans up directories with cache and scripts which are generated on demand.">
+        <delete failonerror="false" includeemptydirs="true">
+            <fileset dir="${path.var}/cache/">
+                <exclude name="/" />
+            </fileset>
+            <fileset dir="${path.web.scripts}/">
+                <exclude name="/" />
+            </fileset>
+        </delete>
+        <delete failonerror="false" includeemptydirs="true" quiet="true">
+            <filelist dir="${path.root}" files="migrations-lock.yml"/>
+        </delete>
+    </target>
+
     <target name="standards-fix" description="Automatically fixes some coding standards violations in both project-base and packages.">
         <phingcall target="ecs-fix" />
         <phingcall target="eslint-fix" />
--- a/packages/migrations/src/Command/MigrateCommand.php
+++ b/packages/migrations/src/Command/MigrateCommand.php
@@ -2,18 +2,13 @@
 
 namespace Shopsys\MigrationBundle\Command;
 
-use Doctrine\DBAL\Migrations\Configuration\Configuration;
-use Doctrine\DBAL\Migrations\OutputWriter;
-use Doctrine\DBAL\Migrations\Tools\Console\Helper\ConfigurationHelper;
 use Doctrine\ORM\EntityManagerInterface;
-use Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsLocator;
-use Symfony\Component\Console\Command\Command;
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsLock;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
-class MigrateCommand extends Command
+class MigrateCommand extends AbstractCommand
 {
     const RETURN_CODE_OK = 0;
     const RETURN_CODE_ERROR = 1;
@@ -29,28 +24,16 @@ class MigrateCommand extends Command
     private $em;
 
     /**
-     * @var \Symfony\Component\DependencyInjection\ContainerInterface
+     * @var \Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsLock
      */
-    private $container;
+    private $migrationsLock;
 
-    /**
-     * @var \Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsLocator
-     */
-    private $migrationsLocator;
-
-    /**
-     * @param \Doctrine\ORM\EntityManagerInterface $em
-     * @param \Symfony\Component\DependencyInjection\ContainerInterface $container
-     * @param \Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsLocator $migrationsLocator
-     */
     public function __construct(
         EntityManagerInterface $em,
-        ContainerInterface $container,
-        MigrationsLocator $migrationsLocator
+        MigrationsLock $migrationsLock
     ) {
         $this->em = $em;
-        $this->container = $container;
-        $this->migrationsLocator = $migrationsLocator;
+        $this->migrationsLock = $migrationsLock;
 
         parent::__construct();
     }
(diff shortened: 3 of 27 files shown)


packages/migrations/src/Command/MigrateCommand.php
```diff
@@ -107,10 +114,14 @@ function ($message) use ($output) {
...
         );
 
-        $migrationsConfiguration = new Configuration($this->em->getConnection(), $outputWriter);
+        $migrationsConfiguration = new Configuration($this->migrationsConfig, $this->em->getConnection(), $outputWriter);
         $configurationHelper = new ConfigurationHelper($this->em->getConnection(), $migrationsConfiguration);
         $this->getApplication()->getHelperSet()->set($configurationHelper, 'configuration');
 
+        $migrationsConfiguration->setMigrationsDirectory(self::MIGRATIONS_DIRECTORY);
+        $migrationsConfiguration->setMigrationsNamespace(self::MIGRATIONS_NAMESPACE);
+        $migrationsConfiguration->setMigrationsFinder($this->migrationFinder);
```
PetrHeinz: If you don't need to change this in runtime, why don't you set this in the DIC configuration?

packages/migrations/src/Component/Doctrine/Migrations/MigrationsConfig.php
```diff
@@ -0,0 +1,81 @@
...
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->migrationsConfigPath = $container->getParameter('shopsys.root_dir') . '/' . self::MIGRATIONS_CONFIG_FILE_NAME;
+    }
+
+    /**
+     * @return string[]
+     */
+    private function getMigrationsConfig()
```
PetrHeinz: The naming isn't clear: method `MigrationsConfig::getMigrationsConfig()` returning array of strings is confusing...

packages/migrations/src/Component/Doctrine/Migrations/MigrationsConfig.php
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * @param string[] $migrationsConfig
+     */
+    private function saveMigrationsConfig(array $migrationsConfig)
+    {
+        $yamlMigrationsConfig = Yaml::dump($migrationsConfig);
+
+        file_put_contents($this->migrationsConfigPath, $yamlMigrationsConfig);
+    }
+}
```
PetrHeinz: missing EOL

packages/migrations/src/Component/Doctrine/Migrations/MigrationsConfig.php
```diff
@@ -0,0 +1,81 @@
...
+     * @param string[] $migrations
+     */
+    public function updateMigrationsConfig(array $migrations)
+    {
+        $migrationsConfig = $this->getMigrationsConfig();
+
+        foreach ($migrations as $migrationVersion => $migrationFile) {
+            if (!array_key_exists($migrationVersion, $migrationsConfig)) {
+                $migrationsConfig[$migrationVersion] = [
+                    'file' => $migrationFile,
```
PetrHeinz: Why are you callling it a file? It's supposed to be FQCN of the migration, see `MigrationsFinder::findMigrations()`.

packages/migrations/src/Component/Doctrine/Migrations/MigrationsConfig.php
```diff
@@ -0,0 +1,81 @@
...
+    const MIGRATIONS_CONFIG_FILE_NAME = 'migrations_config.yml';
+
+    /**
+     * @var string
+     */
+    private $migrationsConfigPath;
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->migrationsConfigPath = $container->getParameter('shopsys.root_dir') . '/' . self::MIGRATIONS_CONFIG_FILE_NAME;
```
PetrHeinz: This path should be set in configuration, the constructor should take it as an argument.

packages/migrations/src/Component/Doctrine/Migrations/MigrationsConfig.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public function updateMigrationsConfig(array $migrations)
+    {
+        $migrationsConfig = $this->getMigrationsConfig();
+
+        foreach ($migrations as $migrationVersion => $migrationFile) {
+            if (!array_key_exists($migrationVersion, $migrationsConfig)) {
+                $migrationsConfig[$migrationVersion] = [
+                    'file' => $migrationFile,
+                    'skip' => 0
```
PetrHeinz: Why use zero instead of a boolean value?

packages/migrations/src/Component/Doctrine/Migrations/MigrationsFinder.php
```diff
@@ -0,0 +1,61 @@
...
+
+        $migrations = [];
+
+        foreach ($this->locator->getMigrationsLocations($directory, $namespace) as $location) {
+            $migrations += $this->finder->findMigrations($location->getDirectory(), $location->getNamespace());
+        }
+
+        ksort($migrations, SORT_STRING);
+
+        $this->migrationsConfig->updateMigrationsConfig($migrations);
```
PetrHeinz: This seems too soon to me - the migrations do not have to be installed, the transaction could be rolled back, the config would stay updated...

packages/migrations/src/Command/MigrateCommand.php
```diff
@@ -107,10 +114,14 @@ function ($message) use ($output) {
             }
         );
 
-        $migrationsConfiguration = new Configuration($this->em->getConnection(), $outputWriter);
+        $migrationsConfiguration = new Configuration($this->migrationsConfig, $this->em->getConnection(), $outputWriter);
         $configurationHelper = new ConfigurationHelper($this->em->getConnection(), $migrationsConfiguration);
         $this->getApplication()->getHelperSet()->set($configurationHelper, 'configuration');
 
+        $migrationsConfiguration->setMigrationsDirectory(self::MIGRATIONS_DIRECTORY);
```
PetrHeinz: This actually creates a directory Migrations in the root directory.

packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
```diff
@@ -0,0 +1,104 @@
...
+namespace Shopsys\MigrationBundle\Component\Doctrine\Migrations;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Migrations\Configuration\Configuration as DoctrineConfiguration;
+use Doctrine\DBAL\Migrations\Finder\MigrationFinderInterface;
+use Doctrine\DBAL\Migrations\OutputWriter;
+use Doctrine\DBAL\Migrations\QueryWriter;
+use Doctrine\DBAL\Migrations\Version;
+
+class Configuration extends DoctrineConfiguration
```
PetrHeinz: This looks to me like duplication of the ordering logic... Who's responsibility is it to order the migrations? Finder's or Configuration's?

build.xml
```diff
@@ -13,6 +13,20 @@
...
+        <delete failonerror="false" includeemptydirs="true">
+            <fileset dir="${path.var}/cache/">
+                <exclude name="/" />
+            </fileset>
+            <fileset dir="${path.web.scripts}/">
+                <exclude name="/" />
+            </fileset>
+        </delete>
+        <delete failonerror="false" includeemptydirs="true" quiet="true">
+            <filelist dir="${path.root}" files="migrations_config.yml"/>
```
PetrHeinz: You don't have to copy-paste the target in order to append one more call, see https://stackoverflow.com/questions/12715855/in-a-phing-build-file-when-i-import-another-build-file-and-then-override-a-targ

Also, why do we need to remove the file?

.gitignore
```diff
@@ -8,3 +8,4 @@
 /docker-compose.yml
 /docker-sync.yml
 /project-base/composer.lock
+/project-base/migrations_config.yml
```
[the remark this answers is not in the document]
TomasLudvik: Yes, that is exact. On framework, we do not need to set order, because we want to be sure, that default order (date) is working right.

packages/migrations/src/Command/MigrateCommand.php
TomasLudvik: We create the instance of Configuration on our own, so it cannot be injected, but I have moved this to parameters passed to the constructor instead of setting it via the method.

packages/migrations/src/Command/MigrateCommand.php
TomasLudvik: I will move this back to MigrationsLocator

packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
TomasLudvik: Well, finders find migrations, but it was not supposed in multiple directories, so it orders them default by timestamp. Then Configuration orders migrations using MigrationsConfig the way it is supposed by setting in migrations_config.yml

build.xml
PetrHeinz: Do we really need to *always* install migrations in the default order? I think it would be better to test the real behavior of the app. We can always manually remove the file.

packages/migrations/src/Resources/config/services.yml
```diff
@@ -14,7 +14,21 @@ services:
       - '@Doctrine\DBAL\Schema\Comparator'
       - '@Doctrine\ORM\Tools\SchemaTool'
 
-  Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsLocator: ~
+  Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsFinder: ~
+
+  Shopsys\MigrationBundle\Component\Doctrine\Migrations\MigrationsConfig:
+    arguments:
+        $migrationsConfigPath: '%shopsys.root_dir%/migrations_config.yml'
```
PetrHeinz: This ties it to the Shopsys Framework even though it is not dependent on it,

PR #206, opened 2018-06-05
clearing Setting's cache is now done via DoctrineEventListener

PetrHeinz:
- removed EntityManagerFacade
- clearing Setting's cache does not bother developers, it is done every time Doctrine identity cache is cleared
- data fixtures do not need to clear the cache manually
- code is much simpler

| Q             | A
| ------------- | ---
|Description, reason for the PR| simplicifation and error-proofing data fixtures
|New feature| No 
|BC breaks| Yes 
|Fixes issues| ...


simara-svatopluk:
I disagree with this PR for a couple of reasons:
* You created a cycle: `EM->ClearSettingsCacheDoctrineSubscriber->Settings->EM`
* The program flow is not obvious anymore.  
Originally it was straight: `EMFacade -> clean -> (clean EM, clean settings)`.  
Now it is: `EM -> clean -> raise event -> catched by subscriber -> clean settings`. I cannot look into the code and know what is going on. I have to know about configuration that is hidden in ` packages/framework/src/Resources/config/services.yml`
* Program logic depends on infrastructure

> code is much simpler

There is less code, but the code is more difficult

PetrHeinz:
@simara-svatopluk:

Can you elaborate on the cycle? How is it created and why it's a bad thing? I am not completely sure what you mean, even though I sense the base idea.

Cleaning the Setting's cache is something that should be done EVERY time the identity map is cleared. This is something event subcribers are made for.

This approach is extensible from modules and project repository - you just add subscribers - so the application can be consistent in the future.

I believe that this really makes the code much simpler, there are less dependencies, the EM can be used much more easily.

It seems to me that you are, in principle, against using event listeners - is this something we should avoid using?


I'm hesitant to merge this when you object so strongly against this, we should discuss it.

simara-svatopluk:
My point of view is that I prefer explicit code. In the case of `EMFacade` I can see what is going on - I can see it directly in the code. When the behavior is done by events, then I don't know what will be done just by the code. The code have to run and event subscribers must be attached.... simply I cannot click in the IDE to see what is happening.

vitek-rostislav:
Hi guys, basically I do agree with Peter's arguments and I am for merging the PR. I do not see the events usage to be a problem in justified cases and IMHO this is justified case. It is true that you cannot simply click in the IDE but you can use `php bin/console debug:event-dispatcher`...also, we should probably describe all our custom subscribers in docs. What do you say @TomasVotruba?

PetrHeinz:
@vitek-rostislav: this is a Doctrine event, unrelated to Symfony event dispatcher - it won't appear in `php bin/console debug:event-dispatcher`.

We should probably look into it and document it, as there currently are 77 Symfony event listeners listening to 16 different events (and a few more Doctrine listeners, I would guess around 20). It might be a good idea to list even 3rd party listeners with a link to their documentation. It would be nice to know what's happening.

The only place in our docs that mentions events is currently the article about [Entity Extension](https://github.com/shopsys/shopsys/blob/v7.0.0-alpha2/docs/wip_glassbox/entity-extension.md#how-does-it-work).

packages/framework/src/Resources/config/services.yml
```diff
@@ -292,6 +292,9 @@ services:
 
   Shopsys\FrameworkBundle\Component\Setting\Setting: ~
 
+  Shopsys\FrameworkBundle\Component\Setting\ClearSettingsCacheDoctrineSubscriber:
+    tags: ['doctrine.event_subscriber']
```
TomasVotruba: This could be handled by `autoconfigure`, right?

PetrHeinz: Yes, it could be, but I didn't feel the need to set it up for autoconfiguration because of a single service (other services are either listeners, or the have specific priority set).

PR #227, opened 2018-06-07
Shopsys.responsive.isDesktopVersion() shoud be same as CSS media query for desktop version

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Media query breakpoint is 769px in css for dekstop version. If you have same JS functionality depended on this  breakpoint, there is `Shopsys.responsive.isDesktopVersion()`, but some browsers return different width of viewport (without scrollbars, this is problem) by used `$(window).width()`. For example some JS is executed, because `Shopsys.responsive.isDesktopVersion()` returns true, but some element is not visible, because css media query is still in tablet version.
|New feature| No
|BC breaks| No
|Fixes issues| ...

I did not update CHANGELOG.md, I did not know link to this futured pull request :( I am demotivated about updating of it twice (befor pull request and after their creating...)


project-base/src/Shopsys/ShopBundle/Resources/scripts/frontend/responsive.js
```diff
@@ -14,7 +14,7 @@
     var lastIsDesktop = null;
 
     Shopsys.responsive.isDesktopVersion = function () {
-        return $(window).width() >= Shopsys.responsive.LG;
+        return window.innerWidth >= Shopsys.responsive.VL;
```
MattCzerner: Im not sure about these changes. The `innerWidth` gets width of element with paddings, not border or margin. In description you were talking about scrollbar, so i dont know how `innerWidth` would help in this one? Also why is desktop's pixel definition `980` and not `769` as before?

Miroslav-Stopka:
Hi @pk16011990,

so basically, the application thinks it has more space for display than it actually has. This is because scrollbars are not taken into calculations of available area. As a result, there is a case where the app still displays some desktop weidgets, even though the real display dimensions already match the mobile version. Thank you for explaining this one :)

PetrHeinz:
Personally, I'm not sure how to test such changes. What browsers and OSs are we supposed to test it on?

I've looked into how we work with breakpoints such as `Shopsys.responsive.LG` elsewhere in JS and I've found [slick.js](https://github.com/shopsys/shopsys/blob/master/project-base/src/Shopsys/ShopBundle/Resources/scripts/frontend/plugins/slick.js#L619) plugin that works with the breakpoints by comparing the value of `window.innerWidth || $(window).width()`, probably because of  browser compatibility. Do we support IE 8 (it looks like it [does not support innerWidth](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth) which might have been the reason to use `$(window).width()`)?

I would have much more confidence in the solution, if we implement the same comparison, i.e.:
```js
    Shopsys.responsive.isDesktopVersion = function () {
        var windowWidth = window.innerWidth || $(window).width();

        return windowWidth >= Shopsys.responsive.LG;
    };
```

boris-brtan:
TESTS looks OK

PR #249, opened 2018-06-13
First architectonical codesniff

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| code needs to follow agreed rules
|New feature| Yes
|BC breaks| No
|Fixes issues| 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,7 +99,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### [shopsys/coding-standards]
 ##### Added
 - [#249 - First architectonical codesniff](https://github.com/shopsys/shopsys/pull/249)
-    - new sniff `ObjectIsCreatedByFactorySniff` was created
+    - new sniff `ObjectIsCreatedByFactorySniff` was created and was integrated into coding standards as service
 
 ##### Changed
 - [#143](https://github.com/shopsys/shopsys/pull/143) [EasyCodingStandard v4.3.0](https://github.com/Symplify/EasyCodingStandard/tree/4.3) is now used
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [#230 - composer-dev updates dependencies if composer.json was changed](https://github.com/shopsys/shopsys/pull/230)
  
 ### [shopsys/coding-standards]
+##### Added
+- [#249 - First architectonical codesniff](https://github.com/shopsys/shopsys/pull/249)
+    - new sniff `ObjectIsCreatedByFactorySniff` was created
+
 ##### Changed
 - [#143](https://github.com/shopsys/shopsys/pull/143) [EasyCodingStandard v4.3.0](https://github.com/Symplify/EasyCodingStandard/tree/4.3) is now used
     - rules config file changed its format from neon to yaml
--- a/packages/coding-standards/src/Sniffs/ObjectIsCreatedByFactorySniff.php
+++ b/packages/coding-standards/src/Sniffs/ObjectIsCreatedByFactorySniff.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\Sniffs;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\ClassHelper;
+use Symplify\TokenRunner\Analyzer\SnifferAnalyzer\Naming;
+
+final class ObjectIsCreatedByFactorySniff implements Sniff
+{
+    /**
+     * @var \Symplify\TokenRunner\Analyzer\SnifferAnalyzer\Naming
+     */
+    private $naming;
+
+    public function __construct(Naming $naming)
+    {
+        $this->naming = $naming;
+    }
+
+    /**
+     * @return int[]
+     */
+    public function register(): array
+    {
+        return [T_NEW];
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $position
+     */
+    public function process(File $file, $position): void
+    {
+        // get full name of the file class
+        $fileClassName = ClassHelper::getFullyQualifiedName($file, $file->findNext(T_CLASS, 2));
+
+        // get full name of the class that is instantiated inside of some method of the file class
+        $className = '\\' . $this->naming->getClassName($file, $file->findNext(T_STRING, $position));
+        $factoryName = $className . 'Factory';
+
+        // if instantiated class is instantiated inside it's factory
+        // if instantiated class doesn't have factory in the same namespace path then code is valid
+        if ($factoryName === $fileClassName || !class_exists($factoryName)) {
+            return;
+        }
+
+        $file->addError(
+            sprintf('For creation of "%s" class use its factory "%s"', $className, $factoryName),
+            $position,
+            self::class
+        );
+    }
+}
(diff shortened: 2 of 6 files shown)


docs/cookbook/creation-of-custom-code-sniff.md
```diff
@@ -0,0 +1,120 @@
+# Creation of custom code sniff
```
vitek-rostislav: I do not think we should have a cookbook for something that is already present in the code. I can imagine article "How to create a custom sniff" that would be a part of `coding-standards` package (linked from `README`, see https://github.com/shopsys/coding-standards#custom-rules - your new sniff should be described here as well). The article then could describe the implementation in more general way and for examples, link to the actual implementation. (We can discuss it personally :wink: )

packages/coding-standards/src/Sniffs/ObjectIsCreatedByFactorySniff.php
```diff
@@ -0,0 +1,52 @@
...
+        return [T_NEW];
+    }
+
+    /**
+     * @param File $file
+     * @param int $position
+     */
+    public function process(File $file, $position): void
+    {
+        $fileClassName = ClassHelper::getFullyQualifiedName($file, $file->findNext(T_CLASS, 2));
```
vitek-rostislav: Is there a difference in between ClassHelper::getFullyQualifiedName() and $this->naming->getClassName? Is not it possible to use only one of these services?

CHANGELOG.md
```diff
@@ -89,6 +89,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 #### Added
 - [#143](https://github.com/shopsys/shopsys/pull/143) [Shopsys Coding Standards dev-master](./packages/coding-standards/) is now used
     - config [easy-coding-standard.yml](./easy-coding-standard.yml]) for importing rules of new easy-coding-standard package
+- [#249 first architectonical codesniff](https://github.com/shopsys/shopsys/pull/249)
```
vitek-rostislav: I do not want to be a nitpick but all the other entries (or most of them) in the changelog have the unified format - hash and number of the PR followed with dash and name of the PR starting with capital letter...Or, it seems like you can use just the hash and number (as you can see on the lines above) and github automatically displays the PR name in the output. But in that case, you should probably rename the PR (start with capital letter). Also, when I am mentioning the PR itself, please always try to add some description/reason of the PR in one sentence at least...

packages/framework/autoload-easy-coding-standard.yml
```diff
@@ -27,3 +27,14 @@ parameters:
         PHP_CodeSniffer\Standards\Generic\Sniffs\CodeAnalysis\EmptyStatementSniff.DetectedWhile:
             - '*/src/Model/Product/Availability/ProductAvailabilityRecalculator.php'
             - '*/src/Model/Product/Pricing/ProductPriceRecalculator.php'
+
+        Shopsys\CodingStandards\Sniffs\ObjectIsCreatedByFactorySniff:
+          - '*/tests/*'
+          - '*DataFixture.php'
```
vitek-rostislav: I do not like all the other exceptions, are they justified? Would it be difficult to fix the violations of the rule? If not, please do that. Otherwise you should explain all the exceptions here by adding comments and create a new user story for resolving it.

vitek-rostislav:
Hi @boris-brtan - from testing: In `Admin\ArticleController`, I have added following line:
```
$article2 = new Article(new ArticleData());
```
and my build (using `php phing build-demo-dev`) passed with no errors even though both `Article` and  `ArticleData` have their factories. When I run the checks once again (`php phing checks-diff`), the violations were reported correctly, how so? Could there be some problem with cache?

docs/cookbook/creation-of-custom-code-sniff.md
```diff
@@ -0,0 +1,120 @@
...
+```
+For excluding some files from checking of this sniff rule there is a way by defining row in skip subsection of parameters section.
+```yml
+parameters:
+    skip:
+        Shopsys\CodingStandards\Sniffs\ObjectIsCreatedByFactorySniff:
+            - '*/tests/*'
+```
+
+## Analyze helpers
```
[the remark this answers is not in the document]
vitek-rostislav: it was not my point :laughing: I just wanted you to use the term "Analysis helpers" instead of the original "Analyze helpers" but never mind :slightly_smiling_face:

packages/coding-standards/src/Sniffs/ObjectIsCreatedByFactorySniff.php
vitek-rostislav: ok

packages/coding-standards/src/Sniffs/ObjectIsCreatedByFactorySniff.php
```diff
@@ -0,0 +1,52 @@
...
+        $className = '\\' . $this->naming->getClassName($file, $file->findNext(T_STRING, $position));
```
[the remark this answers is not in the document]
vitek-rostislav: I meant the variable name...`$instantiatedClassName`

packages/framework/autoload-easy-coding-standard.yml
vitek-rostislav: ok, but even in data fixtures, the instances should be created via factory - I have added it in the issue, please move the comment in the code right after the line with - '*/tests/*'

packages/coding-standards/src/Sniffs/ObjectIsCreatedByFactorySniff.php
```diff
@@ -0,0 +1,52 @@
...
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\ClassHelper;
+use Symplify\TokenRunner\Analyzer\SnifferAnalyzer\Naming;
+
+final class ObjectIsCreatedByFactorySniff implements Sniff
+{
+    /**
+     * @var Naming
```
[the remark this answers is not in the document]
TomasVotruba: Why not create a sniff to check this or use common non-FQCN names?

I don't think a case for people to check.

packages/framework/autoload-easy-coding-standard.yml
```diff
@@ -27,3 +27,21 @@ parameters:
...
+
+        Shopsys\CodingStandards\Sniffs\ObjectIsCreatedByFactorySniff:
+          - '*/tests/*'
+          - '*/src/DataFixtures/Base/CategoryRootDataFixture.php'
+          - '*/src/DataFixtures/Demo/ArticleDataFixture.php'
+          - '*/src/DataFixtures/Demo/CategoryDataFixture.php'
+          - '*/src/DataFixtures/Demo/PaymentDataFixture.php'
+          - '*/src/DataFixtures/Demo/PersonalDataAccessRequestDataFixture.php'
+          - '*/src/DataFixtures/Demo/TransportDataFixture.php'
+          - '*/src/DataFixtures/DemoMultidomain/ArticleDataFixture.php'
```
TomasVotruba: Here it would be easy to maintain to use `fnmatch()`:

```diff
       Shopsys\CodingStandards\Sniffs\ObjectIsCreatedByFactorySniff:
-          - '*/tests/*'
-          - '*/src/DataFixtures/Base/CategoryRootDataFixture.php'
-          - '*/src/DataFixtures/Demo/ArticleDataFixture.php'
-          - '*/src/DataFixtures/Demo/CategoryDataFixture.php'
-          - '*/src/DataFixtures/Demo/PaymentDataFixture.php'
-          - '*/src/DataFixtures/Demo/PersonalDataAccessRequestDataFixture.php'
-          - '*/src/DataFixtures/Demo/TransportDataFixture.php'
-          - '*/src/DataFixtures/DemoMultidomain/ArticleDataFixture.php'
+          - '*/src/DataFixtures/*/*Fixture.php'
```

PetrHeinz: But this would ignore all data fixtures. Only these seven (out of more than fifty fixtures) are violating the sniff.

TomasVotruba: I see. Why not fix it then if it's only just few files?

These ignores (in PHPStan, Code sniffers, ECS) usually suggest it's instentional in whole group. Like in FormTypes. 

Comment above would also help.

PetrHeinz: They should be fixed in  #252, we wanted the sniff in master as soon as possible.

TomasVotruba: I see. 

That would be great to have here in PR description. I could click on that instead of writing comments, /cc @boris-brtan

Issue #190, opened 2018-06-05
 Admin: Product detail: JS validation does not work for the newly added parameter


jakubhavlak:
**How to reproduce:**

1. Open the iPhone editing in administration
2. Add a new parameter and leave the values ​​blank
3. Get out of the box
4. Submit a form

**What will happen:**

When you leave the box, nothing happens.
The form can be sent even though the parameter is not filled in.

**What should happen:**

When leaving the box, a validation message should pop up.
The form should not be sent.

**Note:**

Once the form has been submitted, JS validation for that parameter works.

boris-brtan:
Part of the issue with unwanted saving of form was moved into #374

PR #260, opened 2018-06-18
JS validation: dynamically added form inputs are now validated

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| problem with saving forms with dynamically added parameters
|New feature| No 
|BC breaks| No 
|Fixes issues| #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### [shopsys/framework]
+#### Fixed
+- [#260 - JS validation: dynamically added form inputs are now validated](https://github.com/shopsys/shopsys/pull/260)
 
 ## [7.0.0-alpha4] - 2018-08-02
 ### [shopsys/framework]
--- a/packages/framework/src/Resources/scripts/admin/parameters.js
+++ b/packages/framework/src/Resources/scripts/admin/parameters.js
@@ -27,7 +27,7 @@
             var item = prototype
                 .replace(/__name__label__/g, index)
                 .replace(/__name__/g, index);
-            var $item = $(item);
+            var $item = $($.parseHTML(item));
             $item.data('index', index);
 
             $collection.data('index', index + 1);
--- a/packages/framework/src/Resources/scripts/admin/parameters.js
+++ b/packages/framework/src/Resources/scripts/admin/parameters.js
@@ -33,10 +33,11 @@
             $collection.data('index', index + 1);
 
             $collection.append($item);
+            Shopsys.register.registerNewContent($item);
 
+            Shopsys.validation.addNewItemToCollection('#product_edit_form_parameters', index);
             Shopsys.formChangeInfo.showInfo();
             Shopsys.parameters.refreshCount($collection);
-            Shopsys.validation.addNewItemToCollection('#product_edit_form_parameters', index);
 
             return false;
         });
--- a/packages/framework/src/Resources/scripts/common/validation/customizeBundle.js
+++ b/packages/framework/src/Resources/scripts/common/validation/customizeBundle.js
@@ -115,6 +115,23 @@
         Shopsys.timeout.setTimeoutAndClearPrevious('Shopsys.validation.validateWithParentsDelayed', executeDelayedValidators, 100);
     };
 
+    // Issue in dynamic collections validation that causes duplication of substrings in identifiers
+    // Issue described in https://github.com/formapro/JsFormValidatorBundle/issues/139
+    // PR with fix in the original package: https://github.com/formapro/JsFormValidatorBundle/pull/141
+    FpJsFormValidator._preparePrototype = FpJsFormValidator.preparePrototype;
+    FpJsFormValidator.preparePrototype = function (prototype, name) {
+        prototype.name = prototype.name.replace(/__name__/g, name);
+        prototype.id = prototype.id.replace(/__name__/g, name);
+
+        if (typeof prototype.children === 'object') {
+            for (var childName in prototype.children) {
+                prototype[childName] = this.preparePrototype(prototype.children[childName], name);
+            }
+        }
+
+        return prototype;
+    };
+
     Shopsys.validation.removeDelayedValidationWithParents = function (jsFormValidator) {
         do {
             delete delayedValidators[jsFormValidator.id];
--- a/packages/framework/src/Resources/scripts/admin/categoryTree.formItem.js
+++ b/packages/framework/src/Resources/scripts/admin/categoryTree.formItem.js
@@ -108,7 +108,7 @@
             newItemHtml = newItemHtml.replace(/__name__/g, itemData.id);
             newItemHtml = newItemHtml.replace(/__category_name__/g, itemData.categoryName);
 
-            var $newItem = $(newItemHtml);
+            var $newItem = $($.parseHTML(newItemHtml));
             $newItem.data('load-url', itemData.loadUrl);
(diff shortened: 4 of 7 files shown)


packages/framework/src/Resources/scripts/common/validation/customizeBundle.js
```diff
@@ -115,6 +115,22 @@
         Shopsys.timeout.setTimeoutAndClearPrevious('Shopsys.validation.validateWithParentsDelayed', executeDelayedValidators, 100);
     };
 
+    // Issue in dynamic collections validation
+    // https://github.com/formapro/JsFormValidatorBundle/issues/139
+    FpJsFormValidator._preparePrototype = FpJsFormValidator.preparePrototype;
+    FpJsFormValidator.preparePrototype = function (prototype, name, id) {
+        prototype.name = prototype.name.replace(/__name__/g, name);
```
vitek-rostislav: Could you please add some explanation what exactly happens here, i.e. what was the problem and how it was fixed? I see that you replace both prototype.name and prototype.id with given name but I do not understand why :slightly_frowning_face:

project-base/tests/ShopBundle/Acceptance/acceptance/PageObject/Admin/ProductEditPage.php
```diff
@@ -0,0 +1,31 @@
...
+namespace Tests\ShopBundle\Acceptance\acceptance\PageObject\Admin;
+
+use Tests\ShopBundle\Acceptance\acceptance\PageObject\AbstractPage;
+
+class ProductEditPage extends AbstractPage
+{
+    /**
+     * @param int $productId
+     */
+    public function saveInvalidForm($productId)
```
vitek-rostislav: I would name this method more specifically - it saves form with invalid parameter...

project-base/tests/ShopBundle/Acceptance/acceptance/AdminProductSaveFormWithEmptyParametersCest.php
```diff
@@ -0,0 +1,24 @@
...
+        LoginPage $loginPage,
+        ProductEditPage $productEditPage
+    ) {
+        $me->wantTo('not able to save invalid form');
+        $loginPage->loginAsAdmin();
+
+        $productEditPage->saveInvalidForm(5);
+        $productEditPage->assertSaveFormErrorBoxVisible();
+
+        $productEditPage->cleanUpWorkspaceAfterChanges();
```
vitek-rostislav: why? could not you just call `$this->tester->acceptPopup()`?

project-base/tests/ShopBundle/Acceptance/acceptance/AdminProductSaveFormWithEmptyParametersCest.php
```diff
@@ -0,0 +1,24 @@
...
+use Tests\ShopBundle\Test\Codeception\AcceptanceTester;
+
+class AdminProductSaveFormWithEmptyParametersCest
+{
+    public function testSaveForm(
+        AcceptanceTester $me,
+        LoginPage $loginPage,
+        ProductEditPage $productEditPage
+    ) {
+        $me->wantTo('not able to save invalid form');
```
vitek-rostislav: `$me->wantTo('see that I am not able to save an invalid form');`?

project-base/tests/ShopBundle/Acceptance/acceptance/PageObject/Admin/ProductEditPage.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $this->tester->amOnPage('/admin/product/edit/' . $productId);
+        // scroll to have the button visible because of the fixed bars
```
vitek-rostislav: is it necessary to scroll down? I thought that when you do just `$this->tester->clickByCss('.js-parameters-item-add');`, it does not matter that the element is not in viewport, or does it?

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [#260 - Admin: Js validation: dynamically added parameters are now validated](https://github.com/shopsys/shopsys/pull/260)
     - validation is evaluated after dynamic parameters are added
     - jQuery migration console error logging is no more
+    - support for multiple option of Flag and Brand items were added for form to be validated and not submitted if it is invalid
```
vitek-rostislav: I honestly do not know what to do with this problem, the implementation in `JsFormValidatorFactory::normalizeViewTransformers()` is (already was) really confusing and fragile :disappointed: It should probably be refactored completely but I do not know how... @PetrHeinz could you please help us here? :angel:

vitek-rostislav:
While testing, I noticed yet another bug but it seems that it is already on master so I created a separate issue - https://github.com/shopsys/shopsys/issues/268

project-base/tests/ShopBundle/Acceptance/acceptance/PageObject/Admin/ProductEditPage.php
boris-brtan: Yes, there is the problem with fixed elements. Click method will do the scroll to element like function ```javascript
document.querySelector('.js-parameters-item-add').scrollIntoView()
```
and then client will try to click on it but there is element fixed bar above it and codeception will fail on error 
```
Element \<a href="#" class="js-parameters-item-add btn btn--plus">...\</a> is not clickable at point (338, 15). Other element would receive the click: \<header class="web__header">...\</header>
```

File: project-base/tests/ShopBundle/Acceptance/acceptance/AdminProductSaveFormWithEmptyParametersCest.php (hunk 3 above)
Comment (boris-brtan) [reply project-base/tests/ShopBundle/Acceptance/acceptance/AdminProductSaveFormWithEmptyParametersCest.php]: In this test the form is changed and browser will stop execution of next tests due to modal box that wants approval. Seems that opposite order of commands doesn't work if the test is ended and next test requests for different page modal box is not canceled.

File: packages/framework/src/Resources/scripts/admin/parameters.js [hunk 7]
```diff
@@ -37,6 +37,7 @@
             Shopsys.formChangeInfo.showInfo();
             Shopsys.parameters.refreshCount($collection);
             Shopsys.validation.addNewItemToCollection('#product_edit_form_parameters', index);
+            $('#product_edit_form_parameters .js-parameters-item input').blur();
```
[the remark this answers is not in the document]
boris-brtan: thx, yes better from UX view will be to focus into input

PetrHeinz:
After discussion with @boris-brtan we've reduced scope of this PR as there are two unrelated issues atm:
- The dynamically added inputs do not have JS validation. This is what we aim to fix.
- On product edit page, the form is submitted even if the values are not valid, i.e. JS validation does not prevent the form from submitting. This is related to BG-2136 fixed in 3771de6, we will create a separate issue for this. Boris wanted to fix it in one of the last commits, which recieved most comment during this CR.

Can someone please do a code review of this fix with the reduced scope?

CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [#352 - flushes executed in loops are now executed outside of loop](https://github.com/shopsys/shopsys/pull/352)
 - [#342 - procedures are now executed only if relevant columns are changed](https://github.com/shopsys/shopsys/pull/342)
 - [#362 - guidelines-for-pull-request.md: fixed indentation of lines and code blocks](https://github.com/shopsys/shopsys/pull/362)
+- [#260 - JS validation: dynamically added form inputs are now validated](https://github.com/shopsys/shopsys/pull/260)
```
LukasHeinz: Please correct changelog entry because new version was released.

PR #266, opened 2018-06-21
ecs fix and unification in monorepo

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| problems with ObjectIsCreatedByFactorySniff in packages
|New feature| No 
|BC breaks| Yes 
|Fixes issues| 


packages/coding-standards/README.md
```diff
@@ -25,15 +25,14 @@ php composer require shopsys/coding-standards
 
 ## Usage
 
-Create `custom-coding-standard.yml` config file in your project which includes predefined rulesets. 
+Create `easy-coding-standard.yml` config file in your project which includes predefined ruleset. 
 You can also [customize the rules](./UPGRADE.md)
```
PetrHeinz: There is no shopsys/coding-standards/UPGRADE.md

packages/framework/easy-coding-standard.yml
```diff
@@ -1,5 +1,5 @@
 imports:
-    - { resource: '../**/coding-standards/easy-coding-standard.yml' }
+    - { resource: '%vendor_dir%/shopsys/coding-standards/easy-coding-standard.yml', ignore_errors: true  }
```
PetrHeinz: The reason to use `ignore_errors` should be explaind in the commit msg imho

packages/product-feed-zbozi/easy-coding-standard.yml
```diff
@@ -0,0 +1,10 @@
+imports:
+    - { resource: '%vendor_dir%/shopsys/coding-standards/easy-coding-standard.yml', ignore_errors: true  }
+
+parameters:
+    skip:
+        ObjectCalisthenics\Sniffs\Files\FunctionLengthSniff:
+            - '*/src/ZboziPluginDataFixture.php'
+            - '*/tests/Unit/ZboziFeedTest.php'
+        Shopsys\CodingStandards\Sniffs\ObjectIsCreatedByFactorySniff:
+          - '*/tests/*'
```
PetrHeinz: In some of the config files, there is a new line missing at the end. This results in diffs with worse readability, but it's not that important to fix those - you can leave it as-is.

build.xml
```diff
@@ -70,7 +69,7 @@
         </exec>
         <exec executable="${path.ecs.executable}" logoutput="true" passthru="true" checkreturn="true">
             <arg value="check" />
```
PetrHeinz: The flag `--verbose` is not used here, but it is used in other build definitions. Can you explain or fix it?

build.xml
```diff
@@ -98,8 +97,8 @@
     <target name="standards-fix-packages" description="Automatically fixes some coding standards violations in all packages.">
         <exec executable="${path.ecs.executable}" logoutput="true" passthru="true" checkreturn="true">
             <arg value="check" />
```
PetrHeinz: The flag `--verbose` is not used here, but it is used in other build definitions. Can you explain or fix it?

build.xml
```diff
@@ -70,7 +69,7 @@
         </exec>
         <exec executable="${path.ecs.executable}" logoutput="true" passthru="true" checkreturn="true">
             <arg value="check" />
-            <arg value="--config=${path.ecs.configuration}"/>
+            <arg value="--clear-cache" />
```
PetrHeinz: Clearing the cache with every execution of `standards-packages` might be a bit painful as there is no `standards-packages-diff` variant and it is called during `build-demo-dev`.

Please reconsider this behavior.

build.xml
boris-brtan: verbose was added

build.xml
boris-brtan: verbose was added

build.xml
boris-brtan: clearing of cache was removed

TomasVotruba:
I think there is one missed spot: https://github.com/shopsys/shopsys/blob/master/packages/http-smoke-testing/.travis.yml

Or is that intentional inconsistency?

Issue #145, opened 2018-05-28
Hidden call of ObjectManager::flush() without parameter


vitek-rostislav:
### What is happening
Throughout the system, there is a rule that `flush` should not be called without parameter specifying which entity (or array of entities) should be flushed. We are trying to avoid calling `flush` without parameter by cumulating entities in arrays, eg.: https://github.com/shopsys/shopsys/blob/253540e2e949b3447c3230e04318bdfdf826bf95/packages/framework/src/Model/Customer/CustomerFacade.php#L111. But sometimes the array can be empty and calling `ObjectManager::flush([])` is the same as calling it without parameter, ie. causes flushing the whole system. Someone might be unpleasantly surprised.

### Expected result
There is no hidden threat in calling `ObjectManager::flush([])`. At least the usage and behavior is documented well.

PR #267, opened 2018-06-21
Flush is not called if the array off entities intended to by flushed …

Miroslav-Stopka:
…is empty

| Q             | A
| ------------- | ---
|Description, reason for the PR| Hidden call of ObjectManager::flush() without parameter #145
|New feature| No
|BC breaks| No 
|Fixes issues| #145

### DIFF
--- a/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
+++ b/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
@@ -174,6 +174,7 @@ class FriendlyUrlFacade
             foreach ($friendlyUrls as $friendlyUrl) {
                 if (!$friendlyUrl->isMain()) {
                     $this->em->remove($friendlyUrl);
+                    $toFlush[] = $friendlyUrl;
                 }
             }
         }
@@ -186,7 +187,9 @@ class FriendlyUrlFacade
             $toFlush[] = $newFriendlyUrl;
         }
 
-        $this->em->flush($toFlush);
+        if (count($toFlush) > 0) {
+            $this->em->flush($toFlush);
+        }
     }
 
     /**
--- a/packages/framework/src/Model/Pricing/Currency/CurrencyFacade.php
+++ b/packages/framework/src/Model/Pricing/Currency/CurrencyFacade.php
@@ -251,6 +251,8 @@ class CurrencyFacade
             $toFlush[] = $transportPrice;
         }
 
-        $this->em->flush($toFlush);
+        if (count($toFlush) > 0) {
+            $this->em->flush($toFlush);
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


packages/framework/src/Model/Customer/CustomerFacade.php
```diff
@@ -130,7 +130,9 @@ public function create(CustomerData $customerData)
         $this->em->persist($user);
         $toFlush[] = $user;
 
-        $this->em->flush($toFlush);
+        if (count($toFlush) > 0) {
```
PetrHeinz: It will always be true here.

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -203,7 +203,10 @@ public function createOrder(OrderData $orderData, OrderPreview $orderPreview, Us
 
         $this->orderService->calculateTotalPrice($order);
         $this->em->persist($order);
-        $this->em->flush($toFlush);
+
+        if (count($toFlush) > 0) {
```
PetrHeinz: This will always be true.

packages/framework/src/Model/Product/ProductVariantFacade.php
```diff
@@ -78,7 +78,11 @@ public function createVariant(Product $mainProduct, array $variants)
         try {
             $toFlush = $mainVariant->getVariants();
             $toFlush[] = $mainVariant;
-            $this->em->flush($toFlush);
+
+            if (count($toFlush) > 0) {
```
PetrHeinz: This will always be true.

packages/framework/src/Model/Customer/CustomerFacade.php
TomasVotruba: These should be reported by PHPStan.
Is it broken?

Miroslav-Stopka: Hi Petr, modified. 

Hi Tomas, I'm not sure but there is default setting with allowed reports only for two kinds of warnings - CommentedOutCodeSniff and AssignmentInConditionSniff. And UnconditionalIfStatementSniff is only a warning but seems not allowed for reporting. Could you confirm this for me?

packages/framework/src/Model/Order/OrderFacade.php
Miroslav-Stopka: Modified.

packages/framework/src/Model/Product/ProductVariantFacade.php
Miroslav-Stopka: Modified.

Miroslav-Stopka:
Petr, you are right - $em->flush([]) does not flush all entities the same way as $em->flush() . This problem with flushes seems to be much more extensive than we thought so we created the user story for more deeper analysis. Plus we need to keep on mind that flush with an argument is already deprecated. 

Follow us and you will see some changes in the usage of flush() on Shopsys Framework :)

LukasHeinz:
BV ok

PR #270, opened 2018-06-22
Clean cache before run test db rebuild

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| phing target `tests` fails after change dependencies in container or change ORM mappings.  It is a extension of https://git.shopsys-framework.com/shopsys/shopsys-framework/issues/21
|New feature| No
|BC breaks| No
|Fixes issues| ...

### DIFF
--- a/project-base/build-dev.xml
+++ b/project-base/build-dev.xml
@@ -16,7 +16,7 @@
     <target name="standards" depends="phplint,ecs,phpstan,twig-lint,yaml-lint,eslint-check" description="Checks coding standards."/>
     <target name="standards-diff" depends="phplint-diff,ecs-diff,phpstan,twig-lint-diff,yaml-lint,eslint-check-diff" description="Checks coding standards on changed files."/>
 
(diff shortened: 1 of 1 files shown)


pk16011990:
Yes, I understand and I dont care about it. But my view is:

I run `db-demo` only in dev environment (It does not need `clean`) and `db-rebuild` in `dev` too or in production, where cache must be cleared after all changes (or automatically on deploy). There is no reason for cleaning cache in my opinion and in my use-cases.

And as dependency for `tests` can by annoying: When I run only one test (eg. from IDE), than sometime I can want rebuild test DB it can cause fatal. Target `tests` is only transitive depends on `clean` not directly, In fact they are `test-db-demo`, `tests-db` and `tests-smoke` depend. Exists use-cases where developer can run only `test-db-demo`

PetrHeinz:
Yeah, it makes sense to add it only to the `test-db-demo` as other targets are run mainly in `dev` env, as you say...

PR #273, opened 2018-06-25
URL for submitting newsletter subscription should not be visible for robots

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| MethodNotAllowed is very often logged in production as robots trying to access URL that is dedicated for ajax newsletter registration (taken from https://git.shopsys-framework.com/jakub.dolba/shopsys-framework/commit/d9f353819f8d311dcd621a65f706c02e1953c802)
|New feature| No 
|BC breaks| No 
|Fixes issues| ...

### DIFF
--- a/project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Newsletter/subscription.html.twig
+++ b/project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Newsletter/subscription.html.twig
@@ -1,4 +1,12 @@
-{{ form_start(form, { attr: { 'data-on-submit': 'ajaxSubmit', 'data-success': success ? 'true' : 'false' }}) }}
+{{ form_start(form, {
+    attr: {
+        'data-on-submit': 'ajaxSubmit',
+        'method': 'GET',
+        'action': '',
+        'data-action': url('front_newsletter_send'),
+        'data-success': success ? 'true' : 'false'
+    }
+}) }}
     <div class="footer__newsletter">
         <div class="footer__newsletter__text">
             <p>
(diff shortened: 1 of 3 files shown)


project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Newsletter/subscription.html.twig
```diff
@@ -1,4 +1,12 @@
-{{ form_start(form, { attr: { 'data-on-submit': 'ajaxSubmit', 'data-success': success ? 'true' : 'false' }}) }}
+{{ form_start(form, {
+    attr: {
+        'data-on-submit': 'ajaxSubmit',
+        'method': 'GET',
```
PetrHeinz: Isn't that the default value?

PetrHeinz: Also, does it even matter? I mean, it is only sent in JS and there the method is hard-coded.

project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Newsletter/subscription.html.twig
```diff
@@ -1,4 +1,12 @@
...
+        'action': '',
+        'data-action': url('front_newsletter_send'),
```
PetrHeinz: I would consider a name that points to the fact that this is fully intentional. Something along the way of `data-ajax-only-action`... I'm just a bit afraid about readability and the possibility that someone will "simplify it" to the previous version in the future.

PetrHeinz:
Wouldn't it be even better to just allow POSTing it via a standard form? It would only require a condition on `$request->isXmlHttpRequest()` with a redirect to either a referer or the homepage or something.

A form that's not rendered by JS, but can be submitted only via AJAX may be an obstacle in some cases.

Something similar can be found in `CustomerLoginHandler`: https://github.com/shopsys/shopsys/blob/07abd93ff3353222513ca71322359382104c65dd/packages/framework/src/Model/Security/CustomerLoginHandler.php#L35-L48

vitek-rostislav:
**Summary of the current state:** there are few notes from @PetrHeinz that could be addressed but it is not necessary. However, at least considering them would be nice. The only essential task is **testing**.

boris-brtan:
is this branch still needed and relevant?

boris-brtan:
then, branch deleted

Issue #170, opened 2018-05-29
Admin: Products: Behavior of sort by name is platform specific


jakubhavlak:
When I view the / admin / product / list / page on my computer (Czech 64bit Linux) or on CI, the first products in the series starting with A (Apple iPhone 5S 64GB, gold).

When I do the same with Peter Heinz on his laptop (Windows, maybe 32 bit), the first products in the series starting with alphabet (something like "100 Czech crowns ticket").

We have confirmed that sorting is different on different platforms. This is a known PostgreSQL issue because sorting currently depends on the implementation of collations in the OS.

This should be solved in PostgreSQL v10, when the sorting will be solved by the independent ICU library platform https://blog.2ndquadrant.com/icu-support-postgresql-10/ 

Another problem is that the grid does not even call ORDER BY xxx COLLATE. This will then have to be added to sort the language in which the names appear.

Miroslav-Stopka:
Solved in #278

PR #278, opened 2018-06-27
docs: sorting is different on different platforms

Miroslav-Stopka:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Behavior of sort by name is platform specific 
|New feature| No
|BC breaks| No
|Fixes issues| #170

### DIFF
--- a/docs/installation/native-installation.md
+++ b/docs/installation/native-installation.md
@@ -183,5 +183,14 @@ Timezone for PHP can be set in your `php.ini` (usually located in `/etc/php.ini`
 date.timezone = "UTC"
 ```
 
+### Sorting is different on different platforms
+This is known PostgreSQL issue, PostgreSQL uses the collation implementation from the OS.
+As a result, sorting can be different on different platforms.
+For example, results sorted by name can begin with a product starting with “A” (Apple iPhone 5S) on iOS and with “1” (100 Czech crowns) on Linux.
+
+This problem only occurs if the installation is native and it should be solved in PostgreSQL v10, when the sorting will be solved by the independent ICU library. 
+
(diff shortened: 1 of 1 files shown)


docs/installation/native-installation.md
```diff
@@ -183,5 +183,14 @@ Timezone for PHP can be set in your `php.ini` (usually located in `/etc/php.ini`
...
 ```
 
+### Sorting is different on different platforms
+This is known PostgreSQL issue, PostgreSQL uses the collation implementation from the OS. As a result,
+sorting can be different on different platforms. For example, results sorted by name can begin with a product
+starting with “A” (Apple iPhone 5S) on iOS and with “1” (100 Czech crowns) on Linux.
+
+This problem only occurs if the installation is native and it should be solved in PostgreSQL v10, when the sorting will be solved by the independent ICU library. 
+
+Installation using docker provide the unified environment for all platforms as a result of which the sorting is the same.
```
PetrHeinz: "...provide**s a** unified environment..."

docs/installation/native-installation.md
```diff
@@ -183,5 +183,14 @@ Timezone for PHP can be set in your `php.ini` (usually located in `/etc/php.ini`
 date.timezone = "UTC"
 ```
 
+### Sorting is different on different platforms
+This is known PostgreSQL issue, PostgreSQL uses the collation implementation from the OS. As a result,
+sorting can be different on different platforms. For example, results sorted by name can begin with a product
+starting with “A” (Apple iPhone 5S) on iOS and with “1” (100 Czech crowns) on Linux.
```
PetrHeinz: Docs are much easier to maintain if you put every sentence on a different line.

Issue #281, opened 2018-06-27
monorepo-tools: rewrite_history_into.sh has a non-portable implementation of tab


lukaso:
### What is happening
On OS X, the `sed` command's tab declaration is interpreted as a `t`.

File names unexpectedly have the sub directory substituted after the first `t`

### Expected result
The tab should be interpreted as a tab.

I will submit a pull request shortly.

PetrHeinz:
#282 fixed this issue along with other problem with OS X compatibility

Thanks, @lukaso!

PR #282, opened 2018-06-27
monorepo-tools: Fix scripts to work on OS X

lukaso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixes non-portable tabs so they work on OS X
|New feature| No
|BC breaks| No
|Fixes issues| #281
|Standards and tests pass| Not tested - no changes to main framework

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     - `PaymentDomain` and `TransportDomain` are now created even for domains on which the entity should not be visible (to allow for other multi-domain entities and in the sake of consistency)
 
 #### Fixed
+- [#281 - monorepo-tools: rewrite_history_into.sh has a non-portable implementation of tab](https://github.com/shopsys/shopsys/pull/xxx)
+    - monorepo-tools: fixes non-portable use of tab to function on OS X
 - [#246 - docker-sync.yml.dist: fixed not existing relative paths](https://github.com/shopsys/shopsys/pull/246) [@DavidKuna]
 - [#132 - Admin: brand edit page: URLs setting rendering](https://github.com/shopsys/shopsys/pull/132):
     - admin: brand detail page: rendering of URLs setting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     - `PaymentDomain` and `TransportDomain` are now created even for domains on which the entity should not be visible (to allow for other multi-domain entities and in the sake of consistency)
 
 #### Fixed
-- [#281 - monorepo-tools: rewrite_history_into.sh has a non-portable implementation of tab](https://github.com/shopsys/shopsys/pull/xxx)
+- [#281 - monorepo-tools: rewrite_history_into.sh has a non-portable implementation of tab](https://github.com/shopsys/shopsys/pull/282)
     - monorepo-tools: fixes non-portable use of tab to function on OS X
 - [#246 - docker-sync.yml.dist: fixed not existing relative paths](https://github.com/shopsys/shopsys/pull/246) [@DavidKuna]
 - [#132 - Admin: brand edit page: URLs setting rendering](https://github.com/shopsys/shopsys/pull/132):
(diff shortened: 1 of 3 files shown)


lukaso:
Here is the link to the solution: https://stackoverflow.com/questions/2610115/why-is-sed-not-recognizing-t-as-a-tab/2623007#2623007

PetrHeinz:
Hello @lukaso and thanks for your contribution! :+1: 

I've noticed that you fixed this issue only in `rewrite_history_into.sh` but the same thing (`\t`) is used in `rewrite_history_from.sh`, would you mind fixing it up even there and testing the script on OS X?

Otherwise, please indicate in the PR that only `rewrite_history_into.sh` was fixed - I think it's useful to have it fixed in a single script than have it broken everywhere. I tried to make it work on OS X originally, but then encountered a few problems that prevented me to do it well.

And feel free to link the contribution to your GitHub profile in the changelog (similarly as *DavidKuna* did) :slightly_smiling_face:

lukaso:
@PetrHeinz will do both. I've only been building the repos so haven't had a chance to check the other way... Some of our repos are quite large so takes a while!

lukaso:
OK, this all now works on OS X. @PetrHeinz do take a look.

PetrHeinz:
Hello @lukaso, thanks for fixing both the scripts on OS X! :+1: 

As we currently do not have any automatic tests we will have to test it manually both on Linux and Mac, it might take a while, since we have a national holiday till the end of the week.

I've briefly reviewed the changes and there are a few things I'd like to address (in order of importance):

1) The variable `XARGS` is not used in the index-filter, there has to be `XARGS=$XARGS` before the `git filter-branch`. Also not sure about the variable's name - how about `XARGS_FLAG_NO_EMPTY`?
2) The comment above the code still says *Piping through "echo -e"...* even though `echo` is not used anymore.
3) Why is there a space in `printf "%s "`? Is there a reason for that, or is that a typo?
4) In the rest of the scripts the shorter version of variables is used where possible (eg. `$TAB` instead of `${TAB}`). Not that it's important, just a bit confusing.
5) I haven't found the flag `sed -E` in the manual even though it works... (see https://linux.die.net/man/1/sed) That's a bit weird, but I suppose it's a feature implemented to be compatible with OS X.
6) In CHANGELOG.md, you have to add a link to your profile at the end of the file for it to be properly linked :slightly_smiling_face: 

After the fixes and testing I would like to squash and merge the PR, so don't worry much about Git history...

lukaso:
> 1. The variable XARGS is not used in the index-filter, there has to be XARGS=$XARGS before the git filter-branch. Also not sure about the variable's name - how about XARGS_FLAG_NO_EMPTY?

Fixed and named `$XARGS_OPTS`. Given the long line of variables, does it make sense to do one of these other options? https://stackoverflow.com/questions/9772036/pass-all-variables-from-one-shellscript-to-another

> 2. The comment above the code still says Piping through "echo -e"... even though echo is not used anymore.

Changed.

> 3. Why is there a space in printf "%s "? Is there a reason for that, or is that a typo?

This is required. `echo -e` separates the values with a space, but terminates with `\n`. `printf` adds this space (if included) but does not terminate the same. However, the result for git command is the same.

> 4. In the rest of the scripts the shorter version of variables is used where possible (eg. $TAB instead of ${TAB}). Not that it's important, just a bit confusing.

Fixed.

> 5. I haven't found the flag sed -E in the manual even though it works... (see https://linux.die.net/man/1/sed) That's a bit weird, but I suppose it's a feature implemented to be compatible with OS X.

I've now tested on Ubuntu and it didn't work. Fixed by creating `$SED_OPTS`.

> 6. In CHANGELOG.md, you have to add a link to your profile at the end of the file for it to be properly linked 🙂

Thanks for looking out for my kudos!

I've tested the fixed versions on ubuntu and Mac and no errors occurred, and the resulting repos have the form I would expect.

I've also written some scripts based on our internal repos which do the end to end run, if they would be helpful. Linux tests are then run inside of docker on ubuntu.

lukaso:
OK, phew, I've now tested the unicode variant. I didn't realise what the problem was before, so didn't have a good way to test. Hopefully this solves all the problems. Let me know if this works.

lukaso:
I think 70bc43e improves the unicode handling.

PetrHeinz:
@lukaso: That's great! I didn't know about the `-c` flag for temporary config changes, it's much better than piping through `echo` or something like that... :slightly_smiling_face: I've retested it on Ubuntu and it's OK, thank you.

@Miroslav-Stopka: Can you please test it on Mac? Thanks.

lukaso:
@PetrHeinz Awesome. Happy to have helped! Didn't know about the flag either, but somehow found it :) (Google is our friend)

PR #286, opened 2018-06-28
US-3985 Instantiate entity data objects by factories

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| US-3985 
|New feature| Yes 
|BC breaks| No 
|Fixes issues| ...

### DIFF
--- a/docs/cookbook/adding-new-attribute-to-an-entity.md
+++ b/docs/cookbook/adding-new-attribute-to-an-entity.md
@@ -176,27 +176,39 @@ Shopsys\FrameworkBundle\Model\Product\ProductFactoryInterface: '@Shopsys\ShopBun
 
 9. Create new `ProductDataFactory` in the same namespace as your entity
 by extending [`Shopsys\FrameworkBundle\Model\Product\ProductDataFactory`](../../packages/framework/src/Model/Product/ProductDataFactory.php)
-and overwrite the `createDefault()` method. 
+and overwrite `create()` and `createFromProduct()` methods. Or you can create an independent class by implementing
+[`Shopsys\FrameworkBundle\Model\Product\ProductDataFactoryInterface`](../../packages/framework/src/Model/Product/ProductDataFactoryInterface.php).
 
-*Note: There is no interface for `ProductDataFactory` like it was in the previous step for `ProductFactory`. 
-This issue will be addressed in near future.*
 ```php
 <?php
 
 namespace Shopsys\ShopBundle\Model\Product;
 
+use Shopsys\FrameworkBundle\Model\Product\Product as BaseProduct;
+use Shopsys\FrameworkBundle\Model\Product\ProductData as BaseProductData;
 use Shopsys\FrameworkBundle\Model\Product\ProductDataFactory as BaseProductDataFactory;
 
 class ProductDataFactory extends BaseProductDataFactory
 {
     /**
+     * @param \Shopsys\ShopBundle\Model\Product\Product $product
      * @return \Shopsys\ShopBundle\Model\Product\ProductData
      */
-    public function createDefault()
+    public function createFromProduct(BaseProduct $product): BaseProductData
     {
         $productData = new ProductData();
+        $this->fillFromProduct($productData, $product);
 
-        // ... (copy paste from BaseProductDataFactory)
+        return $productData;
+    }
+
+    /**
+     * @return \Shopsys\ShopBundle\Model\Product\ProductData
+     */
+    public function create(): BaseProductData
+    {
+        $productData = new ProductData();
+        $this->fillNew($productData);
 
         return $productData;
     }
@@ -204,9 +216,9 @@ class ProductDataFactory extends BaseProductDataFactory
 ```
 
 Register your `ProductDataFactory` in [`services.yml`](../../project-base/src/Shopsys/ShopBundle/Resources/config/services.yml) 
-as an alias for the original service.
+as an alias for the original interface.
 ```
-Shopsys\FrameworkBundle\Model\Product\ProductDataFactory: '@Shopsys\ShopBundle\Model\Product\ProductDataFactory'
+Shopsys\FrameworkBundle\Model\Product\ProductDataFactoryInterface: '@Shopsys\ShopBundle\Model\Product\ProductDataFactory'
 ```
 
 ## Enable administrator to edit the `extId` field
@@ -258,6 +270,8 @@ Shopsys\ShopBundle\Form\Admin\ProductFormTypeExtension:
 
 3. In your `Product` class, overwrite the `edit()` method.
 ```php
+<?php
+
 namespace Shopsys\ShopBundle\Model\Product;
 
 use Shopsys\FrameworkBundle\Model\Product\ProductCategoryDomainFactoryInterface;
@@ -276,27 +290,34 @@ public function edit(ProductCategoryDomainFactoryInterface $productCategoryDomai
 }
 ```
 
-4. In your `ProductDataFactory` class, overwrite the `createFromProduct()` method.
-You need to copy paste the parent's method content and on top of it, set your new `extId` field.
+4. In your `ProductDataFactory` class, update the `createFromProduct()` method so it sets your new `extId` field.
+
 ```php
+<?php
+
 namespace Shopsys\ShopBundle\Model\Product;
 
-use Shopsys\FrameworkBundle\Model\Product\Product as BaseProduct;
+use Shopsys\FrameworkBundle\Model\Product\Product;
+use Shopsys\FrameworkBundle\Model\Product\ProductData as BaseProductData;
 
 // ...
 
-/**
- * @param \Shopsys\ShopBundle\Model\Product\Product $product
- * @return \Shopsys\ShopBundle\Model\Product\ProductData
- */
-public function createFromProduct(BaseProduct $product)
+class ProductDataFactory extends BaseProductDataFactory
 {
-    $productData = $this->createDefault();
+    /**
+     * @param \Shopsys\ShopBundle\Model\Product\Product $product
+     * @return \Shopsys\ShopBundle\Model\Product\ProductData
+     */
+    public function createFromProduct(BaseProduct $product): BaseProductData
+    {
+        $productData = new ProductData();
+        $this->fillFromProduct($productData, $product);
+        $productData->extId = $product->getExtId();
 
-    // ... (copy paste from BaseProductDataFactory)
-    $productData->extId = $product->getExtId();
+        return $productData;
+    }
 
-    return $productData;
+    // ...
 }
 ```
 
@@ -343,4 +364,4 @@ $form['product_edit_form[productData][extId]'] = 123456;
 ```
 
 ## Data fixtures
-Currently, it is not possible to modify data fixtures from your project, this issue will be addressed in near future. 
+Currently, it is not possible to modify data fixtures from your project, this issue will be addressed in near future.
--- a/docs/introduction/basics-about-model-architecture.md
+++ b/docs/introduction/basics-about-model-architecture.md
@@ -8,7 +8,7 @@ Model architecture of Shopsys Framework is inspired by Domain Driven Design (DDD
 - **Domain** in DDD is a sphere of knowledge or activity we build application logic around. The domain of Shopsys Framework is e-commerce.
 - **Domain model** is a system of abstractions that describes selected aspects of the domain.
 - **Domain logic** or **business logic** is the higher level rules for how objects of the domain model interact with one another.
- 
+
 Domain model of Shopsys Framework is located in [`FrameworkBundle/Model`](https://github.com/shopsys/framework/tree/master/src/Model). Its concept is to separate behavior and properties of objects from its persistence. This separation is suitable for code reusability, easier testing and it fulfills the Single Responsibility Principle.
 
 Code belonging to the same feature is grouped together (eg. `Cart` and `CartItem`). Names of classes and methods are based on real world vocabulary to be more intuitive (eg. `OrderHashGenerator` or `getSellableProductsInCategory()`).
@@ -23,6 +23,8 @@ Is class encapsulating data. All entities are persisted by Doctrine ORM. One ent
 Entities are inspired by Rich Domain Model. That means entity contains not only getters and setters, but also some use case methods with basic domain logic (e.g. `Product::changeVat()` sets vat and marks product for price recalculation). The entity cannot depend on any other class.
 
 Entities can be used by all layers of the model and even outside of model (eg. controller or templates).
+
+You'll find more about our entities specialities in a [detailed article](entities.md).
 
 ### Example
 ```php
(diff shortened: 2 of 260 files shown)


packages/framework/src/Form/Admin/Product/ProductEditFormType.php
```diff
@@ -50,16 +50,23 @@ class ProductEditFormType extends AbstractType
      */
     private $pluginDataFormExtensionFacade;
 
+    /**
```
vitek-rostislav: If I were a nitpick, I would say that this not belong into this commit... but I won't :smile:

packages/framework/src/Model/Product/ProductDataFactory.php
```diff
@@ -118,18 +127,27 @@ public function createDefault()
...
     /**
      * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
      * @return \Shopsys\FrameworkBundle\Model\Product\ProductData
      */
-    public function createFromProduct(Product $product)
+    public function createFromProduct(Product $product): ProductData
     {
-        $productData = $this->createDefault();
+        $productData = new ProductData();
+        $this->fillNew($productData);
```
vitek-rostislav: why? is this necessary? You do not do that anywhere else

vitek-rostislav: btw, generally, I was thinking, why `createFrom*` methods do not use `create()` methods in factories? Would not it be better? e.g.:
```
public function createFromProduct(Product $product): ProductData
{
        $productData = $this->create();
        $this->fillFromProduct($productData, $product);

        return $productData;
}
```

docs/cookbook/adding-new-attribute-to-an-entity.md
```diff
@@ -343,4 +343,4 @@ $form['product_edit_form[productData][extId]'] = 123456;
 ```
 
 ## Data fixtures
-Currently, it is not possible to modify data fixtures from your project, this issue will be addressed in near future. 
\ No newline at end of file
+Currently, it is not possible to modify data fixtures from your project, this issue will be addressed in near future.
```
vitek-rostislav: What happened here? I do not see any change :smile: empty line or something like that?

docs/cookbook/adding-new-attribute-to-an-entity.md
```diff
@@ -176,37 +176,49 @@ Shopsys\FrameworkBundle\Model\Product\ProductFactoryInterface: '@Shopsys\ShopBun
...
 use Shopsys\FrameworkBundle\Model\Product\ProductDataFactory as BaseProductDataFactory;
 
 class ProductDataFactory extends BaseProductDataFactory
 {
     /**
+     * @param \Shopsys\ShopBundle\Model\Product\Product $product
      * @return \Shopsys\ShopBundle\Model\Product\ProductData
      */
-    public function createDefault()
+    public function createFromProduct(Product $product): BaseProductData
```
vitek-rostislav: as mentioned before, using `BaseProduct` alias would be more readable in my opinion

docs/introduction/entities.md
```diff
@@ -0,0 +1,285 @@
...
+        $this->seoMetaDescriptions = [];
+        $this->seoH1s = [];
+    }
+}
+```
+
+... and example of entity construction
+
+```php
+$brandData = new BrandData();
```
vitek-rostislav: here should be used factory, am I right?

docs/introduction/entities.md
```diff
@@ -0,0 +1,285 @@
...
+
+    // ...
+
+}
+```
+
+## Entity data
+
+Is a data object that is used to transfer data through application and also to create an entity.
+The entity data can be created in the UI (or other data source), then propagated via facade and finally used to create the entity.
```
vitek-rostislav: I am not sure if I understand the "UI (or other data source)" correctly, by UI you mean controllers?

docs/introduction/entities.md
```diff
@@ -0,0 +1,285 @@
...
+... and example of entity construction
+
+```php
+$brandData = new BrandData();
+// $brandData->name = ...
+// ...
+$brand = $this->brandFactory->create($brandData);
+```
+
+## Entity data factory
```
vitek-rostislav: It would be nice to mention `ObjectIsCreatedByFactorySniff` somewhere

docs/wip_glassbox/entity-extension.md
```diff
@@ -85,6 +85,16 @@ Only exception are `*Translation` entities.
...
 If it is needed to extend the translation, it is also necessary to extend the owner entity and override the `createTranslation` method to produce the extended translation.
 
+### Data and DataFactories
+
+Entity data are extended by inheritance.
+Since they are not persisted, there is no need to do anything like in case of entities.
+
+Entity data are created by factories only.
+These factories work as same as entity factories.
+If any part of framework creates an entity data, it uses a factory.
```
vitek-rostislav: again, I would mention `ObjectIsCreatedByFactorySniff` :slightly_smiling_face:

docs/wip_glassbox/entity-extension.md
```diff
@@ -112,9 +122,21 @@ DiscriminatorMap must always contain descendants' FQN because LoadORMMetadataSub
...
+      ```php
+        Shopsys\FrameworkBundle\Model\Product\ProductDataFactoryInterface:
+          alias: DreamProject\Model\Product\ProductDataFactory
+      ```
 * Now your extended entity is automatically used instead of the parent entity:
   * in hydrated Doctrine references
   * in the EntityManager, Repositories and QueryBuilders
   * in newly created entities
+  * in data objects
+    * *but still not in data objects created by forms*
```
vitek-rostislav: could you be more descriptive here please? What does it mean?

project-base/tests/ShopBundle/Database/Model/Order/OrderTransportAndPaymentTest.php
```diff
@@ -419,7 +419,10 @@ public function getDefaultTransport(Vat $vat, $enabledForDomains, $hidden)
      */
     private function getDefaultVat()
```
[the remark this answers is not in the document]
simara-svatopluk: So... I won't do anything with it

packages/framework/src/Model/Product/Filter/ParameterFilterData.php
```diff
@@ -2,29 +2,15 @@
...
     /**
-     * @var \Shopsys\FrameworkBundle\Model\Product\Parameter\Parameter
+     * @var \Shopsys\FrameworkBundle\Model\Product\Parameter\Parameter|null
      */
     public $parameter;
 
     /**
      * @var \Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterValue[]
      */
     public $values = [];
```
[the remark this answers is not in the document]
simara-svatopluk: The goal was to clear constructors if possible (and it was).
In my opinion, the goal wasn't to have all parameters initialized (we would need them as parameters in constructors)

packages/framework/src/Model/Product/Filter/ProductFilterData.php
```diff
@@ -20,7 +20,7 @@ class ProductFilterData
     public $parameters = [];
```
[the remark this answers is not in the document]
simara-svatopluk: I updated entity data objects only. Only those data objects that have its' entity.
ProductFilterData is just a class named data, but it is not an entity data object

packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlToGenerateRepository.php
```diff
@@ -19,11 +19,14 @@ class FriendlyUrlToGenerateRepository
     private $em;
 
     /**
-     * @param \Doctrine\ORM\EntityManagerInterface $em
+     * @var \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlDataFactoryInterface
      */
-    public function __construct(EntityManagerInterface $em)
+    protected $friendlyUrlDataFactory;
+
+    public function __construct(EntityManagerInterface $em, FriendlyUrlDataFactoryInterface $friendlyUrlDataFactory)
```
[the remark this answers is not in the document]
simara-svatopluk: Thanks, it will be a great fun :coffin:

packages/framework/src/Form/Admin/Customer/CustomerFormType.php
```diff
@@ -57,7 +67,7 @@ public function configureOptions(OptionsResolver $resolver)
             ->setAllowedTypes('user', [User::class, 'null'])
             ->setAllowedTypes('domain_id', 'int')
             ->setDefaults([
-                'data_class' => CustomerData::class,
+                'empty_data' => $this->customerDataFactory->create(),
```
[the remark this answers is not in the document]
simara-svatopluk: Yes, unfortunatelly, it is intended. `CustomerData` must have parameters in constructor, because User, Billing and Delivery Data are created by factory (right after this commit).
Is it enought to write this sentence into CM?

packages/framework/src/Model/Customer/CustomerFacade.php
```diff
@@ -81,7 +89,8 @@ public function register(UserData $userData)
     {
         $userByEmailAndDomain = $this->findUserByEmailAndDomain($userData->email, $userData->domainId);
 
-        $billingAddress = $this->billingAddressFactory->create(new BillingAddressData());
+        $billingAddressData = $this->billingAddressDataFactory->create();
+        $billingAddress = $this->billingAddressFactory->create($billingAddressData);
```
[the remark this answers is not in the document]
simara-svatopluk: This approach is used in about 5-10 places.
Your suggestion will influence entity factory interfaces - they will be more complicated.
I think that simple interface is more important than 2 lines of code (times 5-10). What do You think?

packages/framework/src/Model/Customer/UserDataFactory.php
```diff
@@ -19,15 +19,58 @@ public function __construct(PricingGroupSettingFacade $pricingGroupSettingFacade
...
+    {
+        return new UserData();
+    }
+
     /**
      * @param int $domainId
      * @return \Shopsys\FrameworkBundle\Model\Customer\UserData
      */
-    public function createDefault($domainId)
+    public function createFromDomainId(int $domainId): UserData
```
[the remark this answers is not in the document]
simara-svatopluk: I wanted to keep the naming `createFrom` and also `fillFrom`.

You may say that the data are filled from some internal dependency, so it doesn't make sense. But also when data object is created from an entity, it is usually filled by an internal dependency.

So... Do You still think this method should be named `createForDomainId`? If so, I'll change it

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Shopsys\FrameworkBundle\Model\Order\Item;
+
+class OrderItemDataFactory implements OrderItemDataFactoryInterface
```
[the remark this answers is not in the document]
simara-svatopluk: So You don't think they should be related?

Because for me, I don't see any reason - every factory implements an interface...
I had another idea - one `OrderItemDataFactoryInterface` for all three types - there would be `createItem`, `createTransportItem`, ... but in the end I decided to keep consistency - one data object -> one factory.

packages/framework/src/Form/Admin/Product/ProductEditFormType.php
simara-svatopluk: thanks :nail_care:

packages/framework/src/Model/Product/ProductDataFactory.php
simara-svatopluk: Good point.
If method `createFromProduct` depends on `create` method, and I need to add something into `create`, then I have to think how it influences the dependent method. So I decided to separate them, so we are able to add something into `create` without influencing the `createFromProduct`.

Couple years ago, I did similar approach in one project, and in the end I had troubles, because changing one method influenced the second one, I started to create new method like `createRaw`, `createDefault`, and it was... ununderstable

docs/cookbook/adding-new-attribute-to-an-entity.md
simara-svatopluk: There wasn't new line in the end of file, and phpstorm created it. I'll change it to original, so this file is not changed in this commit

docs/introduction/entities.md
simara-svatopluk: Ok, I moved this example under data factory, so I can use them ;-)

docs/introduction/entities.md
simara-svatopluk: changed to "controllers"

docs/introduction/entities.md
simara-svatopluk: Hm... the same is valid also for entity factory. Shouldn't I mention it in both? But it can be strange to have the same sentence twice in text... what do You think?

docs/wip_glassbox/entity-extension.md
simara-svatopluk: This is a horrible mistake - this part is about entities. Removed

simara-svatopluk:
After the CR I have to:
* add to commit message of 7a8278f5f1 information why did I change ArticleDataFixture

packages/framework/src/Model/Customer/CustomerFacade.php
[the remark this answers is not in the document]
vitek-rostislav: I do agree with you, it is not important :slightly_smiling_face:

packages/framework/src/Model/Customer/UserDataFactory.php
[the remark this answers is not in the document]
vitek-rostislav: for readability sake, I would prefer renaming the method to `createForDomainId` as the former seems quite confusing to me even though it will break the `createFrom*` naming consistency. I believe it is justified here :slightly_smiling_face:

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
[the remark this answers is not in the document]
vitek-rostislav: Yes, I do agree with you, it was just some gut feeling ("objects are in relation -> should their factories be in relation, too?") but otherwise I do not see any reason for that, too. I just wanted to discuss that feeling :smile:

packages/framework/src/Model/Product/ProductDataFactory.php
vitek-rostislav: I see, ok, that makes sense :+1: And what about my question on the `$this->fillNew($productData);` above?

simara-svatopluk: I thought that `fillNew` method contains something that `fillFromProduct` doesnt. But it is not true, so I'm removing the calling. Good point :100:

docs/introduction/basics-about-model-architecture.md
```diff
@@ -67,174 +67,7 @@ class Product
...
-        Product $product,
-        int $quantity,
-        string $watchedPrice
-    ): CartItem;
-}
-```
-
-The factory has an implementation in the framework and can be overwritten in your project when you need to work with an extended entity.
-You can read about entity extension in a [separate article](../wip_glassbox/entity-extension.md).
+You'll find more about our entities specialities in a [detailed article](entities.md).
```
LukasHeinz: Please do not place important information below big image and next heading. It is hard to find thi link to this separate article. I would suggest to place it above example

Issue #291, opened 2018-06-28
Unnecessary SQL queries on category detail in admin


vitek-rostislav:
### What is happening

When accessing page with category editing or creating in admin, as many SQL queries are executed, as many categories I have in my eshop. This is caused by selectbox for setting parent (ancestor) category - it contains all categories loaded by `$this->categoryFacade->getAllWithoutBranch()`, resp. by `$this->categoryFacade->getAll()` which loads categories but without translations which are loaded for each category in separate query.

### Expected result

Category translations are loaded in the same query as categories themselves so there are no more unnecessary queries on admin category detail pages.

Miroslav-Stopka:
Solved in #304

PR #304, opened 2018-07-03
admin: category translations for ancestor category are loaded in the …

Miroslav-Stopka:
…same query as categories

-  before this change, there was executed the query for every possible ancestor category, which means hundreds of queries for hundreds of categories

| Q             | A
| ------------- | ---
|New feature| No 
|BC breaks| No
|Fixes issues| #291


LukasHeinz:
BV done

Issue #149, opened 2018-05-28
Coding standards - sniff for forgotten dumps


vitek-rostislav:
### What is happening
Sometimes a programmer forgets a dump in the code, which can be accidentally displayed in production.
### Expected result
There should be a sniff watching for forgotten dump calls (at least for PHP and Twig calls - `d()`, `dump()`, `var_dump()`, `var_export()`, maybe for JS `conosle.log()`)

PR #308, opened 2018-07-09
Bb coding standards dump sniff

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| needed for clean environment without forgotten log outputs
|New feature| Yes 
|BC breaks| No 
|Fixes issues| #149 


packages/coding-standards/src/Finder/FileFinder.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\Finder;
+
+use IteratorAggregate;
+use SplFileInfo;
+use Symfony\Component\Finder\Finder;
```
simara-svatopluk: Please add `symfony/finder` dependency into package `composer.json`

CHANGELOG.md
```diff
@@ -39,6 +39,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     - all Base Data Fixtures were removed
     - the data are created either in database migrations or in Demo Data Fixtures
 
+### [shopsys/coding-standards]
+#### Added
+- [#308 - Sniff for forgotten dumps](https://github.com/shopsys/shopsys/pull/308)
+    - added support for checking standards of file types twig, html, js
```
simara-svatopluk: I'm not sure... did we add support for html? Because I see only `.twig`, `.html.twig`, `.js` in source code

packages/coding-standards/src/Finder/FileFinder.php
```diff
@@ -0,0 +1,41 @@
...
+            if (is_file($singleSource)) {
+                $fileInfo = new SplFileInfo($singleSource);
+                $files[$fileInfo->getPath()] = new SymfonySplFileInfo($singleSource, $fileInfo->getPath(), $fileInfo->getPathname());
+            } else {
+                $directories[] = $singleSource;
+            }
+        }
+
+        $finder = Finder::create()->files()
+            ->name('#\.(twig|html\.twig|php|js)$#')
```
simara-svatopluk: I thought php files are found automatically. If so, could You remove `php`? If it works of course

composer.json
```diff
@@ -137,6 +137,7 @@
     "symfony/monolog-bundle": "^3.1.2",
     "symfony/swiftmailer-bundle": "^3.2.2",
     "symfony/symfony": "^3.4.8",
+    "symplify/easy-coding-standard-tester": "^4.4",
```
simara-svatopluk: I think this should be a dev dependency

packages/coding-standards/easy-coding-standard.yml
```diff
@@ -128,6 +128,8 @@ services:
         maxCount: 30
 
 parameters:
+    exclude_files:
+        - '*/tests/Unit/**/wrong/*'
```
simara-svatopluk: I have a question - is this file used by other packages? If so, then this line would ignore files in other packages, and that is not prefered in my opinion. Am I right? If so, is it possible to solve excluding wrong tests in a different way? Because I don't know

packages/coding-standards/tests/Unit/CsFixer/MissingButtonTypeFixer/MissingButtonTypeFixerTest.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Tests\CodingStandards\CsFixer\OrmJoinColumnRequireNullableFixer;
+
+use Symplify\EasyCodingStandardTester\Testing\AbstractCheckerTestCase;
+
+final class MissingButtonTypeFixerTest extends AbstractCheckerTestCase
+{
+    public function testFix(): void
+    {
+        $this->doTestWrongToFixedFile(__DIR__ . '/wrong/wrong.html.twig', __DIR__ . '/fixed//fixed.html.twig');
```
simara-svatopluk: typo - double `//`

packages/coding-standards/tests/Unit/CsFixer/OrmJoinColumnRequireNullableFixer/fixed/fixed.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
simara-svatopluk: When we have here more wrong/fixed files, please rename them so they express what is wrong in the filename
This can be for example `fixed_many_to_one_missing_join_column`

And the same can be done with correct cases... Because correct3 doesn't tell anything

packages/coding-standards/tests/Unit/Sniffs/ForbiddenExitSniff/correct/correct.php.inc
```diff
@@ -0,0 +1,3 @@
+<?php
+
+echo 'Hi';
```
simara-svatopluk: Other tests do no have such "empty" test case. Please remove it.

packages/coding-standards/tests/Unit/Sniffs/ForbiddenSuperGlobalSniff/wrong/wrong.php.inc
```diff
@@ -0,0 +1,3 @@
+<?php
+
+$password = $_POST['password'];
```
simara-svatopluk: also here, please rename files to more descriptive names like `post.php.inc`, `env.php.inc`

packages/coding-standards/tests/Unit/Sniffs/ObjectIsCreatedByFactorySniff/ObjectIsCreatedByFactorySniffTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Symplify\EasyCodingStandardTester\Testing\AbstractCheckerTestCase;
+
+final class ObjectIsCreatedByFactorySniffTest extends AbstractCheckerTestCase
+{
+    public function testCorrect(): void
+    {
+        $this->doTestCorrectFile(__DIR__ . '/Correct/PostFactory.php');
+    }
+
```
simara-svatopluk: We are missing calling the `wrong` case

packages/coding-standards/tests/Unit/Sniffs/ObjectIsCreatedByFactorySniff/Wrong/PostFactory.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\CodingStandards\Sniffs\ObjectIsCreatedByFactorySniff\Wrong;
+
+final class PostFactory
+{
```
simara-svatopluk: this factory have to contain also the `create` method

packages/coding-standards/easy-coding-standard.yml
boris-brtan: this file is used only by monorepo and coding-standards package in other packages.
this kind of solution is used in other packages, path is almost specific so it should not be problem in most cases as general and no problem in out monorepo.

packages/coding-standards/src/Finder/FileFinder.php
boris-brtan: '#\.(twig|html(\.twig)?|php|js)$#'

composer.json
```diff
@@ -147,6 +147,7 @@
     "codeception/codeception": "^2.4.0",
     "codeception/phpunit-wrapper": "^7.0",
     "phpstan/phpstan": "^0.7"
+    "symplify/easy-coding-standard-tester": "^4.4",
```
simara-svatopluk: please move comma on the on of the line to the previous line - json is invalid

packages/coding-standards/src/CsFixer/ForbiddenDumpFixer.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $tokens->setCode($code);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName(): string
+    {
+        return 'Shopsys/missing_button_type';
```
simara-svatopluk: The name isn't correct ;-)

packages/coding-standards/src/CsFixer/ForbiddenDumpFixer.php
```diff
@@ -0,0 +1,86 @@
...
+        return false;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(SplFileInfo $file, Tokens $tokens): void
+    {
+        $code = preg_replace_callback(
+            '@[{]{2}\s*(d|dump)\s*[(]\s*.+?\s*[)]\s*[}]{2}@imsu',
```
simara-svatopluk: I found `@{{\s*(d|dump)\s*\(\s*.+?\s*\)\s*}}@imsu` more readable. What do You think?

PR #311, opened 2018-07-10
[US-3861] Monorepo split allow adding new package when monorepo is already tagged

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| US-3861
|New feature| Yes
|BC breaks| No
|Fixes issues| ...

### DIFF
--- a/packages/monorepo-tools/README.md
+++ b/packages/monorepo-tools/README.md
@@ -76,6 +76,21 @@ Other branches are not pushed.
 
 It may again take a while, depending on the size of your monorepo.
 
+### Add a new package into the monorepo
+
+When you have the monorepo, you may find a reason for adding a new package after some time you already use the monorepo.
+In this case, don't use `monorepo_build.sh`, but do following steps:
+
+* Create a new repository, for example, *vendor/gamma*
+* Add remote into the monorepo `git remote add package-gamma http://github.com/vendor/gamma.git`
+* Create a new directory in the monorepo **packages/gamma**
+* Add the code and commit it
+* Use split tool with the new package
+    ```
+    ~/monorepo-tools/monorepo_split.sh \
+        main-repository package-alpha:packages/alpha package-beta:packages/beta package-gamma:packages/gamma
+    ```
+
 ## Reference
 
 This is just a short description and usage of all the tools in the package.
@@ -122,6 +137,18 @@ Usage: `original_refs_wipe.sh`
 Delete all local branches and create all non-remote-tracking branches of a specified remote.
 
 Usage: `load_branches_from_remote.sh <remote-name>`
+
+### [tag_refs_backup.sh](./tag_refs_backup.sh)
+
+Backup tag refs into `refs/original-tags/`
+
+Usage: `tag_refs_backup.sh`
+
+### [tag_refs_move_to_original.sh](./tag_refs_move_to_original.sh)
+
+Move tag refs from `refs/original-tags/` into `refs/original/`
+
+Usage: `tag_refs_move_to_original.sh`
 
 ## Contributing
 Thank you for your contributions to Shopsys Monorepo Tools package.
--- a/packages/monorepo-tools/rewrite_history_from.sh
+++ b/packages/monorepo-tools/rewrite_history_from.sh
@@ -10,8 +10,25 @@
 # Example: rewrite_history_from.sh main-repository --branches
 
 SUBDIRECTORY=$1
-REV_LIST_PARAMS=${@:2}
+REV_LIST_PARAMS=`echo ${@:2} | tr " " "\n"`
 echo "Rewriting history from a subdirectory '$SUBDIRECTORY'"
+
+# Get directory of the other scripts
+MONOREPO_SCRIPT_DIR=$(dirname "$0")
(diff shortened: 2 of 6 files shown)


Miroslav-Stopka:
Hi @simara-svatopluk,

thank you for your contribution :) Im ready for review and tests.

packages/monorepo-tools/rewrite_history_from.sh
```diff
@@ -10,8 +10,25 @@
...
-REV_LIST_PARAMS=${@:2}
+REV_LIST_PARAMS=`echo ${@:2} | tr " " "\n"`
 echo "Rewriting history from a subdirectory '$SUBDIRECTORY'"
+
+# Get directory of the other scripts
+MONOREPO_SCRIPT_DIR=$(dirname "$0")
+# Backup tag refs because `git filter-branch` doesn't backup tags
+$MONOREPO_SCRIPT_DIR/tag_refs_backup.sh
+
+# Remove tags that don't contain any files in $SUBDIRECTORY
```
Miroslav-Stopka: Im am kind of dumb user Svatopluk. So I like clear and describing comments. This comment means that tag can contain some files?

packages/monorepo-tools/rewrite_history_from.sh
```diff
@@ -10,8 +10,25 @@
...
 
 SUBDIRECTORY=$1
-REV_LIST_PARAMS=${@:2}
+REV_LIST_PARAMS=`echo ${@:2} | tr " " "\n"`
 echo "Rewriting history from a subdirectory '$SUBDIRECTORY'"
+
+# Get directory of the other scripts
+MONOREPO_SCRIPT_DIR=$(dirname "$0")
+# Backup tag refs because `git filter-branch` doesn't backup tags
+$MONOREPO_SCRIPT_DIR/tag_refs_backup.sh
```
Miroslav-Stopka: First you backup some tags. Next, you remove some tags. Why?

packages/monorepo-tools/rewrite_history_from.sh
```diff
@@ -35,3 +52,5 @@ SUBDIRECTORY=$SUBDIRECTORY SUBDIRECTORY_SED=${SUBDIRECTORY//-/\\-} TAB=$'\t' XAR
     --tag-name-filter 'cat' \
     -- $REV_LIST_PARAMS
 
+# Moves backup tag refs to the same group with automatic backup from `git filter-branch`
```
Miroslav-Stopka: So you backup all tags. Next, you are removing some tags - tags that are unnecessary for an actual split of a package, finally, you restore all tags in monorepo. Am I right?

packages/monorepo-tools/rewrite_history_into.sh
```diff
@@ -12,6 +12,12 @@
 SUBDIRECTORY=$1
 REV_LIST_PARAMS=${@:2}
 echo "Rewriting history into a subdirectory '$SUBDIRECTORY'"
+
```
Miroslav-Stopka: Was this change necessary? rewrite_history_into.sh is used only during the build of monorepo. Why you backup and restore tags during the build?

packages/monorepo-tools/README.md
```diff
@@ -123,6 +123,18 @@ Delete all local branches and create all non-remote-tracking branches of a speci
```
Miroslav-Stopka: We have new function :) New package is allowed by split of monorepo. I would like to see a brief tutorial on how to split the new package. I need only two or three sentences in the new section before the Reference section

Miroslav-Stopka:
@simara-svatopluk I have already tested this change (it was not an easy one, because I have tried to make it functional with git version 2.10 :D ) But in the end, everything worked fine. I have just some questions. Please, answer them :)

packages/monorepo-tools/rewrite_history_from.sh
simara-svatopluk: Almost exactly!
1. Backup monorepo tags :heavy_check_mark: 
2. Remove tags that are unnecessary for split :heavy_check_mark: 
3. Restoring is not the right word, because monorepo tags are not restored, they are "restored"/moved into refs/original section, so they can be restored by git later.

packages/monorepo-tools/rewrite_history_into.sh
simara-svatopluk: We agreed on this during PIA.
Since git do not backup tags during `git filter-branch`, we decided to do so even during creating monorepo.
Behavior before this change is strange - git does `filter-branch`, moves tags. Then git is reseted to original state, but tags are still on commits after rewriting history into. Then second package is processed, but tags still aims to first package commits. And in the end `filter-branch` deals with this strange situation "somehow".
This change makes this strange situation more straightforward - backup all tags, filter-branch, restore all tags.

packages/monorepo-tools/rewrite_history_from.sh
simara-svatopluk: contains means affected

packages/monorepo-tools/rewrite_history_into.sh
simara-svatopluk: this comment is invalid. Change reverted from rewrite_history_into.sh

PR #313, opened 2018-07-11
Streamed logging

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| for scalable environment is common to use centralized logging system
|New feature| Yes 
|BC breaks| No 
|Fixes issues| 


Miroslav-Stopka:
Hi @boris-brtan,
thank you for this PR. I'll do the review :)

Miroslav-Stopka:
First question - US-4235 says "They are explored the differences between STDERR and STDOUT". Is this criterium done?

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Php-fpm
+In php-fpm docker conainer configuration is set by default that php-fpm service sends error and access log messages into `STDERR` stream.
+`/usr/local/etc/php-fpm.d/docker.conf`
+```
+[global]
+error_log = /proc/self/fd/2
+
+[www]
+; if we send this to /proc/self/fd/1, it never appears
```
Miroslav-Stopka: Why? Why it never appears?

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+access.log = /proc/self/fd/2
+
+clear_env = no
```
Miroslav-Stopka: Why is this parameter set to "no"?

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+
+; Ensure worker stdout and stderr are sent to the main error log.
+catch_workers_output = yes
+```
+For other docker containers we can easily setup logging of any service inside container like it is for php-fpm.
```
Miroslav-Stopka: I don't understand. First, you say that php-fpm is set by default -I dont see where I can change the configuration of this container. Here you say that we can easily setup logging - how, how we can set up logging? Where? If I would need to set up configurations for logging into the stream I do not know what I should do.

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+      type: stream
+      path: "php://stdout"
+      channels: cron
+      level: debug
+    slow:
+      type: stream
+      path: "php://stdout"
+      channels: slow
+      level: debug
+```
```
Miroslav-Stopka: So we need to change the configuration only for slow and cron logs? What about main monolog log or nested log?

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+    postgres:
+        image: postgres:9.5-alpine
+        container_name: shopsys-framework-postgres
+        logging:
+          driver: syslog
+          options:
+            syslog-address: udp://host_with_syslog_daemon:514
+```
+
+### Json-file driver
```
Miroslav-Stopka: I think that we do not want to use json-file driver. Am I right? If so, why is this driver described here?

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+as a default, syslog stores all incoming messages into `/var/log/syslog` but there is a way how to filter logs that come
+from docker containers and store them into different path by modifying rsyslog configuration file so before
+line `*.*;auth,authpriv.none -/var/log/syslog` we can put
+```
+if $syslogtag startswith "docker" and $msg contains "cron." then /path_to_logs/docker-php-fpm.cron.log
+& stop # stop processing of matched message 
+if $syslogtag startswith "docker" and $msg contains "slow." then /path_to_logs/docker-php-fpm.slow.log
+& stop
+if $syslogtag startswith "docker" then /path_to_logs/logs.log
+& stop
```
Miroslav-Stopka: This code will be put where? Is it possible to format this code in this documentation? The current version is very hard to read. Thank you

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+## Conclusion
+
+Since we need another container for the centralized logging system and additional setup, we will postpone it until 
+implementation of orchestration system like Kubernetes that will handle all streams from docker containers unified way.
```
Miroslav-Stopka: The last acceptance criterium is not done. I'm not sure why is not possible to configure new container for logs even before orchestration. I was expecting that after this US there will be logging problem solved.

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
+# Logging with docker containers
+If there is need to debug some problem in application that is deployed within many docker containers,
```
Miroslav-Stopka: in the application

Miroslav-Stopka: is need -> is a need

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+it is better to have logs from different sources centralized.
+For docker, there is easy way how to do it by sending all logs into streams `STDOUT` and `STDERR`.
```
Miroslav-Stopka: is easy way -> is an easy way

docs/introduction/docker-stream-logging.md
Miroslav-Stopka: docker containers we -> docker containers, we

Miroslav-Stopka: inside container -> inside a container

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+clear_env = no
+
+; Ensure worker stdout and stderr are sent to the main error log.
+catch_workers_output = yes
+```
+For other docker containers we can easily setup logging of any service inside container like it is for php-fpm.
+
+## Monolog
+Php application uses [symfony/monolog-bundle](https://github.com/symfony/monolog-bundle) for logging. It is easy to setup
+monolog to send logs into streams by modifiying `app/config/config.yml`
```
Miroslav-Stopka: modifiying -> modifying

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+          options:
+            syslog-address: udp://host_with_syslog_daemon:514
+```
+
+### Json-file driver
+Default setting of logging driver for each docker container is [json-file](https://docs.docker.com/config/containers/logging/json-file/)
+This driver stores logs as json files in `/var/lib/docker/containers/container_id/` thus the log files are divided
+can be rotated by size and limited by files count.
+
+### Syslog driver
```
Miroslav-Stopka: The Syslog driver?

docs/introduction/docker-stream-logging.md
```diff
@@ -0,0 +1,104 @@
...
+To have centralized logs from multiple containers is better to use syslog driver that send logs in its specific form to
```
Miroslav-Stopka: use syslog driver -> use the syslog driver

Miroslav-Stopka:
First iteration of reviewing is done. Please check the results.

CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - cron modules use the logger for debug information
   - DailyFeedCronModule is responsible for continuation of the correct feed after waking up
 - [#182 - Cart: flush() is called only if there are really some changes in cart items](https://github.com/shopsys/shopsys/pull/182)
+- [#313 - Streamed logging](https://github.com/shopsys/shopsys/pull/313)
+    - monolog is logging into streams
```
Miroslav-Stopka: monolog is logging -> monolog logs

docs/introduction/logging.md
```diff
@@ -0,0 +1,48 @@
+# Logging
+Shopsys Framework is Symfony based application that uses [Monolog](https://github.com/Seldaek/monolog)
+with [symfony/monolog-bundle](https://github.com/symfony/monolog-bundle) as logging tool.
+Now it stores cron, slow and other logs into `project-base/var/logs/` directory in which the log files are rotated based
+on date and count of each log file.
+
+Shopsys Framework runs in environment that is built of [docker containers](https://www.docker.com/what-container)
+for php, database, web server, mailing server, testing server and cache server.
```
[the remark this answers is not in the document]
boris-brtan: yes, it is redis container.

Miroslav-Stopka:
First, there is a lot of warnings when I use docker-compose logs. It is a bit difficult to navigate in such a log in the terminal, find some manually logged info message is almost impossible. Can we do something about it? For example, most of the log entries are from event.DEBUG. 

Second, every log entry is marked as WARNING. Even my manually logged info message. Is this state correct?

Third, right now build is outputting every executed query into the terminal. The smooth and hard-to-keep output of the build looks kind of broken right now. Probably do something with doctrine.DEBUG log entries.

Fourth, event.DEBUG is really messing terminal output - try for example php phing cron-list

boris-brtan:
we can use grep function for filtering like
```
docker-compose logs php-fpm -f | grep "php.INFO"
```
Now it is correct but I found out that also others are complaining about it  in https://bugs.php.net/bug.php?id=71880 now the solution is to do some hack. I am not sure whether it is good solution. 

for now it can be hacked with something like 
```
./phing cron-list 2> /proc/1/fd/2
```
monolog sends logs into error stream of the process that is not linked to master process or php-fpm pool processes so hacky solution is to send error stream of the process into the master process and it will redirect it into docker logging system.

docs/index.md
```diff
@@ -12,6 +12,7 @@
 * [Components](./introduction/components.md)
 * [Abstract Filesystem](./introduction/abstract-filesystem.md)
 * [Product Feeds](./introduction/product-feeds.md)
+* [Logging](./introduction/logging)
```
PetrHeinz: The link 404's, the `.md` is missing...

docs/installation/native-installation.md
```diff
@@ -36,7 +36,17 @@ cd project-base
...
-### 2. Install dependencies and configure parameters
+### 2. Set up configuration for native installation
+- Monolog is configured for logging into streams in `app/config/config.yml`.
+If you want to log into a file, change the configuration of handlers. For example 
+
+`cron`:
+    
+    `type`: rotating_file 
+    
+    `path`: "%kernel.logs_dir%/%kernel.environment%.log"
```
PetrHeinz: The formatting seems broken, also it's not clear where exactly to put the config...

I would try:
```markdown
```yml
monolog:
  # ...
  handlers:
    # ...
    cron:
      type: rotating_file
      max_files: 7
      path: "%kernel.logs_dir%/%kernel.environment%.cron.log"
` ` ` (without the spaces... haven't managed to escape it...)
```

Which will look like:

```yml
monolog:
  # ...
  handlers:
    # ...
    cron:
      type: rotating_file
      max_files: 7
      path: "%kernel.logs_dir%/%kernel.environment%.cron.log"
```

project-base/docker/php-fpm/Dockerfile
```diff
@@ -57,6 +57,8 @@ RUN apk add --update --no-cache --repository http://dl-cdn.alpinelinux.org/alpin
 ENV LD_PRELOAD /usr/lib/preloadable_libiconv.so php
 ENV LC_ALL=en_US.utf8 LANG=en_US.utf8 LANGUAGE=en_US.utf8
 
+COPY docker-php-entrypoint /usr/local/bin/
```
PetrHeinz: This overwrites the original docker-php-entrypoint (https://github.com/docker-library/php/blob/master/docker-php-entrypoint)

An inline comment would be helpful...

PetrHeinz:
Merged manually in 23954edd9

PR #322, opened 2018-07-17
Monorepo tools partial build using `-p`

lukaso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Adds a feature that allows a monorepo to be built in a number of steps, rather than all at once.
|New feature| Yes 
|BC breaks| No
|Fixes issues| ...
|Standards and tests pass| I have tested, but there are no standard tests for these tools.


PetrHeinz:
Hi @lukaso, if I understood you correctly, the partial build is a way to add existing repositories to an already built monorepo, right? I'm not entirely sure I would recommend adding more functionality by flags, it could end up in mess after a couple more features. Maybe making it a separate script would make sense, something like `monorepo_add.sh <remote-name>[:<subdirectory>] <remote-name>[:<subdirectory>] ...` What do you think?

I think we can do it with an option as well, but I'd like to consider other ways as well.

lukaso:
I'll take a look.

PetrHeinz:
Hi @lukaso, I thought the different approach would be only to add the new `monorepo_add.sh` without the changes to `monorepo_build.sh` or the commonly used script.

The `monorepo_*.sh` scripts are basically just simple wrappers, without much logic. This would keep it as simple as it is now. What do you think?

lukaso:
> Hi @lukaso, I thought the different approach would be only to add the new monorepo_add.sh without the changes to monorepo_build.sh or the commonly used script.

> The monorepo_*.sh scripts are basically just simple wrappers, without much logic. This would keep it as simple as it is now. What do you think?

When I went to do that, most of the logic was duplicated. That didn't look that great, so I went with this shared logic approach. Basically, if there were any bugs or changes with one script, the change would have been needed with the other (other than very trivial changes like documentation). There's actually more logic in these scripts than meets the eye.

If you feel the copy and modify version is better, I can do that quite easily, just let me know the preferred direction.

PetrHeinz:
Hi @lukaso, I have discussed it with @MattCzerner and we agreed that it would be best to implement the new use case as a different script at the moment, even though it will lead to copy-pasted code.

The main reason are the automatic tests that we haven't implemented yet. This leads to difficult testing on our part. If the original script `monorepo_build.sh` is not modified in any way, it requires no tests.

In the future, we'd like to refactor the scripts and extract the common parts into separate scripts or functions. We don't write much scripts in bash, so we'd like to keep away from complicating things with options etc.

Could you please modify the solution so only the `monorepo_add.sh` is implemented (along with a documenting comment) and brief documentation is added (just add a new section in the Reference part of the README)? There should be no changes to the other scripts. Even though there will be some code duplication (maybe a lot), for the reasons above we feel it will be better in the end. We can of course discuss it further, if you believe this is a wrong way to do it :slightly_smiling_face:

Thanks a lot for your enthusiasm :1st_place_medal: 

P.S.: please ignore the scrutinizer and codacy builds, we're just testing the possibilites of integration - they do not have to be green atm.

lukaso:
No problem. I'll open a new PR and link this when I'm ready.

PR #365, opened 2018-07-27
fix phing target descriptions

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The descriptions were just wrong and misleading.
|New feature| No 
|BC breaks| No 
|Fixes issues| ...

### DIFF
--- a/project-base/build.xml
+++ b/project-base/build.xml
@@ -54,9 +54,9 @@
     <target name="build" depends="build-deploy-part-1-db-independent, build-deploy-part-2-db-dependent" description="Builds application for production preserving your DB."/>
     <target name="build-deploy-part-1-db-independent" depends="clean,composer,npm,dirs-create,domains-urls-check,assets" description="First part of application build for production preserving your DB (can be run without maintenance page)."/>
     <target name="build-deploy-part-2-db-dependent" depends="db-migrations,create-domains-data,generate-friendly-urls,replace-domains-urls,grunt,error-pages-generate,warmup" description="Second part of application build for production preserving your DB (must be run with maintenance page when containing DB migrations)."/>
-    <target name="build-new" depends="wipe,composer,npm,dirs-create,domains-urls-check,assets,db-rebuild,grunt,error-pages-generate,warmup" description="Builds application for production preserving your DB."/>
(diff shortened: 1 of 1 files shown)


project-base/build.xml
```diff
@@ -54,9 +54,9 @@
...
-    <target name="build-new" depends="wipe,composer,npm,dirs-create,domains-urls-check,assets,db-rebuild,grunt,error-pages-generate,warmup" description="Builds application for production preserving your DB."/>
-    <target name="build-demo" depends="wipe,composer,npm,dirs-create,domains-urls-check,assets,db-demo,grunt,error-pages-generate,warmup" description="Builds application for production with clean DB."/>
-    <target name="db-demo" depends="db-wipe-public-schema,db-import-basic-structure,db-migrations,db-fixtures-demo-singledomain,create-domains-data,db-fixtures-demo-multidomain,load-plugin-demo-data,generate-friendly-urls,replace-domains-urls" description="Builds application for production with clean demo DB."/>
+    <target name="build-new" depends="wipe,composer,npm,dirs-create,domains-urls-check,assets,db-rebuild,grunt,error-pages-generate,warmup" description="Builds application for production with clean DB (with base data only)."/>
+    <target name="build-demo" depends="wipe,composer,npm,dirs-create,domains-urls-check,assets,db-demo,grunt,error-pages-generate,warmup" description="Builds application for production and fills DB with demo data."/>
```
PetrHeinz: Sound like the DB is preserved and demo data are added. I would suggest "Builds application for production **with clean demo DB**"

vitek-rostislav: Are you sure? The target depends on `db-demo` which depends on `db-wipe-public-schema` that actually drops the public schema so I believe no data are preserved. Or am I missing something?

PetrHeinz: I mean the new description sounds like the DB is preserved and filled with demo data:

> Builds application for production and fills DB with demo data.

That's why I suggested using this message instead:

> Builds application for production with clean demo DB

vitek-rostislav: Oh, I see :laughing: thx, fixed :+1:

PR #366, opened 2018-07-30
fix missing translation on constrain bad combination transport and payment

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| On constraint with bad combination transport and payment missing "t" function for translation
|New feature| No 
|BC breaks| No 
|Fixes issues| ...

### DIFF
--- a/project-base/src/Shopsys/ShopBundle/Form/Front/Order/TransportAndPaymentFormType.php
+++ b/project-base/src/Shopsys/ShopBundle/Form/Front/Order/TransportAndPaymentFormType.php
@@ -101,7 +101,7 @@ class TransportAndPaymentFormType extends AbstractType
         }
 
         if (!$relationExists) {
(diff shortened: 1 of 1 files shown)


PetrHeinz:
Hi @sspooky13, thanks for the PR. The violations are always translated with a translation domain "validators" (not "messages"), meaning you should add the translation to `validators.<locale>.po`. The use of `t()` is not necessary and it would lead to double translation. For this reason I'm closing the PR.

The bad thing is that these translation are not exported via `dump-translations` phing target, that is something that need improvement on our part. Maybe you'd like to help with that?

sspooky13:
Hello @PetrHeinz, I tried to solve this problem for one hour but it is more complicated as I thought.
I create issue with this problem and if i have time, i try again fix this problem but I do not promise to do it

PR #371, opened 2018-07-31
Remove unused dependencies + Add UnusedPrivateElementsSniff

TomasVotruba:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Removing forgotten dependencies that are never called
|New feature| No
|BC breaks| No
|Fixes issues| No

Found with following sniff:

```yaml
# easy-coding-standard.yml
services:
    SlevomatCodingStandard\Sniffs\Classes\UnusedPrivateElementsSniff: ~
```

Must have to keep your code clean from this

### DIFF
--- a/packages/framework/src/Twig/ImageExtension.php
+++ b/packages/framework/src/Twig/ImageExtension.php
@@ -7,7 +7,6 @@ use Shopsys\FrameworkBundle\Component\Image\ImageFacade;
 use Shopsys\FrameworkBundle\Component\Image\ImageLocator;
 use Shopsys\FrameworkBundle\Component\Utils\Utils;
 use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 use Twig_Extension;
 use Twig_SimpleFunction;
 
@@ -19,11 +18,6 @@ class ImageExtension extends Twig_Extension
      * @var string
      */
     private $frontDesignImageUrlPrefix;
-
-    /**
-     * @var \Symfony\Component\DependencyInjection\ContainerInterface
-     */
-    private $container;
 
     /**
      * @var \Shopsys\FrameworkBundle\Component\Domain\Domain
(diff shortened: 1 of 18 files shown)


packages/framework/src/Twig/ImageExtension.php
```diff
@@ -47,21 +41,18 @@ class ImageExtension extends Twig_Extension
 
     /**
      * @param string $frontDesignImageUrlPrefix
-     * @param \Symfony\Component\DependencyInjection\ContainerInterface $container
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
      * @param \Shopsys\FrameworkBundle\Component\Image\ImageLocator $imageLocator
      * @param \Shopsys\FrameworkBundle\Component\Image\ImageFacade $imageFacade
```
TomasVotruba: Btw, these annotations are really weird. Do you still see value in them? :)

They are also often missed, as I noticed on many places even only in this PR, which makes code look randomly developed.

TomasVotruba:
@simara-svatopluk 

> Could You please add notes into CHANGELOG.md about this change to sections "framework", "product-feed-heureka" and "project-base"?

I don't see value in doing changelog manually, moreover over multiple files. Who would read all 15 files to find out what has changed?

Also, why would somebody look for such a change? Changelog should be bring value to the readers, not dozens of cluttered pages they have to read to find out. Seing such changelog I'd rather go to pull-requests instead.

> And could you also explain removing blank line in the commit message?

I'm not sure what you mean. Should I split that into standalone commit? What message do you want there and why?

TomasVotruba:
@MattCzerner I'm on it. 

> I just wonder if we can add this sniff into our application to watch over these things, I tried to add it into sniffs but it is controling moch more than I want.

Well, sniff doesn't have to find ONLY cases you want. It find few false positives from time to time, but if it brings value like this one, it's worth it. Also, you need to add `skip` for false positive just once. Setup & forget.

I'll look on it.

Also next time some list/link of these false positives would be more helpful. Now I don't know what exactly are you talking about.

TomasVotruba:
@simara-svatopluk I've updated the changelog with your notes just to make this pass.

Please don't make more people suffer like me and use [Generated Changelog](https://www.tomasvotruba.cz/blog/2018/06/25/let-changelog-linker-generate-changelog-for-you/#category-monorepo-support) instead (in any form).

TomasVotruba:
:tada: Thanks guys

PR #386, opened 2018-08-07
FileUpload wrong second parameter for UploadFailedException

Maybeway:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When upload of file contains errors, application will faill. UploadFailedException will not be thrown because wrong second parameter expected instance of Exception or null -  integer given. Method getError() of class UploadedFile returns int only.
|New feature| No 
|BC breaks| No 
|Fixes issues| ...


vitek-rostislav:
Hi @Maybeway, thank you for your contribution :+1: :slightly_smiling_face:

PR #387, opened 2018-08-08
OrderInterface is created for extensibility of OrderItem

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| after Order entity is extended, all entities that have targetEntity set onto OrderEntity needs to be remapped on the extended one.
|New feature| Yes 
|BC breaks| No 
|Fixes issues| 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### [shopsys/framework]
+#### Changed
+- [#387 - OrderInterface is created for extensibility of OrderItem](https://github.com/shopsys/shopsys/pull/387)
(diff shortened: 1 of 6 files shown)


PetrHeinz:
I have added a new PR #388 that demonstrates this is not needed. I have successfully extended the Order entity along with OrderItem and all its descendants that were already extended in the test without using resolve target entity.

boris-brtan:
Problem was with [EntityExtensionTest](https://github.com/shopsys/shopsys/blob/master/project-base/tests/ShopBundle/Database/EntityExtension/EntityExtensionTest.php)
solution is to add alias into $entityExtensionMap

```
use Shopsys\FrameworkBundle\Model\Order\Order;
use Shopsys\ShopBundle\Model\Order\Order as ExtendedOrder;
....
$entityExtensionMap[Order::class] = ExtendedOrder::class;
```

Issue #200, opened 2018-06-05
[cs] remove docs that are already typehinted, pre-PHP7 prepare


jakubhavlak:
I'm upgrading packages on Github to PHP 7/PHP7.1 with automated tools:

- https://github.com/Symplify/EasyCodingStandard
- https://github.com/RectorPHP/Rector

It's very fast upgrade that does 80 % work for you.

I'd like to try it here when decide to do so, starting with this dummy PR.

Let me know what you think :)

(I didn't run tests yet. I'd do it so and fix false positives after/if this gets a go).

Merge request available here:
https://git.shopsys-framework.com/shopsys/shopsys-framework/merge_requests/17

TomasVotruba:
Especially this 1 fixer: 
https://www.tomasvotruba.cz/blog/2017/12/17/new-in-symplify-3-doc-block-cleaner-fixer/

Miroslav-Stopka:
Hi,
we discussed it with several developers, and we decided to annotate everything. This is the most standard way, and at the same time makes the annotation clear and consistent. Problem will be solved in #417. I'm closing this issue.

PR #390, opened 2018-08-08
Turn param types from docblock to typehints

TomasVotruba:
| Q             | A
| ------------- | ---
|New feature| No
|BC breaks| No
|Fixes issues| Closes #376
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

As response to an email by @Miroslav-Stopka  :)

Inspiration: https://github.com/kukulich/php-type-hints-convertor

**This is just prototype to discuss over. It needs to run tests and fix the some missing typehints due to inconsistent parent classes/interfaces.**

### How to Replicate?

```yaml
# easy-coding-standard.yml
services:
    PhpCsFixer\Fixer\FunctionNotation\PhpdocToReturnTypeFixer:

    SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff:
        usefulAnnotations:
            - '@deprecated'
            - '@see'
            - '@dataProvider'
            - '@group'

    Symplify\CodingStandard\Fixer\Commenting\RemoveUselessDocBlockFixer:


parameters:
    skip:
        SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff.MissingTraversableParameterTypeHintSpecification: ~

        # remove, run then add
        SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff.MissingParameterTypeHint: ~

        SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff.MissingTraversableReturnTypeHintSpecification: ~
        
        # remove, run then add
        SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff.MissingReturnTypeHint: ~

        SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff.MissingTraversablePropertyTypeHintSpecification: ~

        SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff.MissingPropertyTypeHint: ~
```
	
**dry-run**

```bash
vendor/bin/ecs check packages project-base 
```

**fix**

```bash
vendor/bin/ecs check packages project-base --fix
```

PR #391, opened 2018-08-08
WIP Added fixers to remove redundant annotations

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| redundant info in annotations (takes space and requires maintenance)
|New feature| No 
|BC breaks| No 
|Fixes issues| #200 


packages/framework/src/Command/Exception/MissingLocaleException.php
```diff
@@ -25,9 +25,6 @@ public function __construct($missingLocale, Throwable $previous = null)
         parent::__construct($message, 0, $previous);
     }
 
-    /**
-     * @return string
```
PetrHeinz: This should not have been removed!

packages/framework/src/Component/Grid/GridView.php
```diff
@@ -118,9 +112,7 @@ public function renderBlock($name, array $parameters = [], $echo = true)
     }
 
     /**
-     * @param \Shopsys\FrameworkBundle\Component\Grid\Column $column
      * @param array|null $row
```
PetrHeinz: This should have been removed as well (as the default value is clearly `null`). Seems like an issue in the fixers themselves, it shouldn't influence us as it isn't something critical...

packages/framework/src/Component/Grid/InlineEdit/AbstractGridInlineEdit.php
```diff
@@ -57,9 +56,6 @@ public function getGrid()
         return $grid;
     }
 
-    /**
-     * @return bool
```
PetrHeinz: This should not have been removed.

TomasVotruba:
I think we had the same idea :)
There is more inspiration #390

packages/framework/src/Component/Grid/GridView.php
TomasVotruba: It' not the same: `?array` != `array`

packages/framework/src/Command/Exception/MissingLocaleException.php
TomasVotruba: Please report these issues at Symplify so they can be fixed

packages/framework/src/Component/Grid/GridView.php
PetrHeinz: But argument declared as `array $var = null` is of type `array|null`, not `array`.

TomasVotruba: I see. In that case some NullableFixer/Sniff might handle it

PetrHeinz: You mean `NullableTypeForNullDefaultValueSniff` from Slevomat? That would change the typehint to `?array $var = null`, I think, and that's not intended here.

TomasVotruba: Probably. Personally I prefer `?array $var` without default value, due to clean code

Miroslav-Stopka:
Hi,
we discussed it with several developers, and we decided to annotate everything. This is the most standard way, and at the same time makes the annotation clear and consistent. Problem will be solved in #417. I'm closing this issue.

PR #399, opened 2018-08-11
monorepo-tools: Handle rewriting of empty commits

drekbour:
Repos created via 'git svn' always start with an empty commit that generates no rewritten index file and fails with `mv: cannot stat  .... index.new`. This is commonly reported issue on SO with a simple fix.

| Q             | A
| ------------- | ---
|Description, reason for the PR| monorepo_build/add doesn't work for reasonably common enterprise scenario of upgrading many trunks from a single svn repo into a git monorepo
|New feature| No
|BC breaks| No
|Fixes issues| 
|Standards and tests pass| N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [#260 - JS validation: dynamically added form inputs are now validated](https://github.com/shopsys/shopsys/pull/260)
 - [#397 - classes excluded from ObjectIsCreatedByFactorySniff rule are fixed](https://github.com/shopsys/shopsys/pull/397)
 
+### [shopsys/monorepo-tools]
+#### Fixed
+- [#399 - monorepo-tools: Handle rewriting of empty commits](https://github.com/shopsys/shopsys/pull/399) [@drekbour]
+
 ### [shopsys/project-base]
 #### Fixed
 - [#359 - fix wrong variable usage in url to reset product filter](https://github.com/shopsys/shopsys/pull/359)
@@ -1393,3 +1397,4 @@ That's why is this section formatted differently.
 [@DavidKuna]: https://github.com/DavidKuna
 [@lukaso]: https://github.com/lukaso
 [@TomasVotruba]: https://github.com/TomasVotruba
+[@drekbour]: https://github.com/drekbour
--- a/packages/monorepo-tools/rewrite_history_into.sh
+++ b/packages/monorepo-tools/rewrite_history_into.sh
@@ -17,6 +17,6 @@ echo "Rewriting history into a subdirectory '$SUBDIRECTORY'"
 # The tags are rewritten as well as commits (the "cat" command will use original name without any change)
 SUBDIRECTORY_SED=${SUBDIRECTORY//-/\\-} TAB=$'\t' git filter-branch \
     --index-filter '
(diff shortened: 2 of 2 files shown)


PetrHeinz:
Hi @drekbour, thanks for the fix!

I've notiíced that you haven't actually filled out the [License Agreement for contributions](https://www.shopsys.com/license-agreement), without which we cannot proceed with the PR for licencing reasons... Please, fill it out.

Would you consider mentioning this change in our CHANGELOG.md (under section shopsys/monorepo-tools > Fixed, preferably with your username linked to your profile - search for `@lukaso` for inspiration)? It may help others see that the `git svn` scenario is fixed.

Thanks a lot!

---

Also, I am a bit worried that after splitting the monorepo the commits will not be the same in such a case, because the first, empty commit will be discarded (because of the [`git_commit_non_empty_tree`](https://github.com/shopsys/shopsys/blob/master/packages/monorepo-tools/rewrite_history_from.sh#L51)). I don't have any idea how to address the issue at the moment but it's not blocking merging this PR, still it's something to be aware of when working with such repos.

Have you tried building and splitting monorepo from such repositories? How did it work for you?

drekbour:
Done and done.

* git svn also writes empty commits for metadata updates (most likely being a standalone change to `svn:ignore`)
* all empty commits _are_ retained in the monorepo master
* split isn't really part of the migration plan but I did try a test and it did fail though I didn't spend any time debugging.

It might be worth adding a specific section for `git svn` although, being only at the beginning of a migration, I'm not going to PR that myself yet

PetrHeinz:
Thanks a lot, @drekbour :+1: I've tested your change and verified it works (sadly, we do not have automatic tests for the monorepo tools at this point).

Unfortunately, I've noticed you've added the changelog entry under already released version `7.0.0-alpha4` instead of `Unreleased`. You should have added the whole section to the bottom of `Unreleased`, at [the line 50](https://github.com/shopsys/shopsys/pull/399/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR50) like so:

```markdown
...
#### Fixed
- [#260 - JS validation: dynamically added form inputs are now validated](https://github.com/shopsys/shopsys/pull/260)

### [shopsys/monorepo-tools]
#### Fixed
- [#399 - monorepo-tools: Handle rewriting of empty commits](https://github.com/shopsys/shopsys/pull/399) [@drekbour]
 
## [7.0.0-alpha4] - 2018-08-02
### [shopsys/framework]
...
```

I guess it is a bit hard to see as the `7.0.0-alpha4` was pretty large... We're working on doing smaller releases more often so it should be easier in the future.

Also, in CHANGELOG.md, you have to add a link to your profile at the end of the file for it to be properly linked :slightly_smiling_face:

Can you please fix it up so we can merge it in time for the `7.0.0-alpha5` release? Thanks a lot!

drekbour:
Oops. Will correct. 
Shouldn't you be _generating_ change logs from the content instead of this way around? Add some kind of `noteworthy` label if you want to thin down what get's reported etc...

PetrHeinz:
Thank you @drekbour :+1: You're right that modifying the changelog is a pain at the moment. We'll try to fix it, hopefully it will require almost no effort from the implementators in a few weeks time.

If you have any tips or experience on tools to generate changelogs in a monorepo with some sort of manual control via tagging, please, feel free to share them :slightly_smiling_face:

drekbour:
No concrete "this works" experience to give but your choice is whether to generate from Git or GitHub. With that out of he way I suspect the biggest problem is picking from the hundreds of utilities people have churned out for this (it seems to be a popular side project).
Good Luck.

Issue #419, opened 2018-08-21
Order flow fix - function call on string


jDolba:
In OrderFlow (\Shopsys\ShopBundle\Form\Front\Order\OrderFlow::changeRequestToStep) is called function on string

this is wrong method usage since upgrade from Symfony 2

alse see:  https://github.com/symfony/form/commit/08f76324a8e2c19b140dbdda9bce6cea328ff38a#diff-b828a8e3d29bb649e85781718cb9fbe4
 - method getName does not exists in FormTypeInterface anymore

### What is happening

in some circumstances, fatal error (Call to a member function getName() on string) might occur 

I was not able to reproduce this problem myself, I noticed this only in error log few times

### Expected result

 - do not throw fatal error and 500 error page in order process

TomasVotruba:
This should be reproted by PHPStan. Is the build broken or PHPStan missed it?

jDolba:
this "bug" was there since SSFW upgrade from symfony 2.8 to symfony 3+

so PHPStan missed it

jDolba:
this "bug" was there since SSFW upgrade from symfony 2.8 to symfony 3+

so PHPStan missed it

vitek-rostislav:
I'll take care of it :wink:

PR #420, opened 2018-08-21
Order flow fix - function call on string (#419)

jDolba:
| Q             | A
| ------------- | ---
|Description, reason for the PR|.see #419 
|New feature| No
|BC breaks| No
|Fixes issues| #419 

### DIFF
--- a/project-base/src/Shopsys/ShopBundle/Form/Front/Order/OrderFlow.php
+++ b/project-base/src/Shopsys/ShopBundle/Form/Front/Order/OrderFlow.php
@@ -164,7 +164,7 @@ class OrderFlow extends FormFlow
         $request = $this->getRequest()->request;
         $requestParameters = $request->all();
         $requestParameters['flow_order_step'] = $step->getNumber();
-        $requestParameters[$step->getFormType()->getName()] = $stepData;
+        $requestParameters[$step->getFormType()] = $stepData;
         $request->replace($requestParameters);
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [#401 - Microservice Product Search](https://github.com/shopsys/shopsys/pull/401)
     - added [Microservice Product Search](https://github.com/shopsys/microservice-product-search), microservice is used for the searching of products on Shopsys Framework 
     - added MicroserviceClient component
+- [#420 - Order flow fix](https://github.com/shopsys/shopsys/pull/420)
+    - fix fatal error in OrderFlow (issue #419): function call on string, also method getName does not exists since upgrade to Symfony 3
 
 #### Changed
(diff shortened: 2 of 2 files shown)


vitek-rostislav:
Hi @jDolba, thanks for the pull request, the changes are fine with me but I need yet another one team member to approve and test your changes.

project-base/src/Shopsys/ShopBundle/Form/Front/Order/OrderFlow.php
```diff
@@ -164,7 +164,7 @@ private function changeRequestToStep(StepInterface $step)
         $request = $this->getRequest()->request;
         $requestParameters = $request->all();
         $requestParameters['flow_order_step'] = $step->getNumber();
-        $requestParameters[$step->getFormType()->getName()] = $stepData;
+        $requestParameters[$step->getFormType()] = $stepData;
```
TomasLudvik: Hello @jDolba,

I have reviewed your PR and I found one problem. `$step->getFormType()` can return `FormTypeInterface`. You cannot use interface as key for an array. Can you find some better way to fix this?

Thank you.

jDolba: `FormTypeInterface` does not have any method to get "name" since symfony3 


also look at the
`\Shopsys\ShopBundle\Form\Front\Order\OrderFlow::loadStepsConfig`
which is used to set `$step::formType` - you will find that configuration of `form_type` as always string


also look at the issue and watch my link to symfony component commit where is also note in `@deprecation` note that you should always use class name (string)

``` Deprecated since Symfony 2.8, to be removed in Symfony 3.0.  Use the fully-qualified class name of the type instead. ```

vitek-rostislav:
After discussion with @jDolba I am sure that `$step->getFormType()` can not return `FormTypeInterface`. This is just type hinted as an option because `CraueFormFlowBundle` still supports `symfony/form` in `~2.7` version (see https://github.com/craue/CraueFormFlowBundle/blob/13a2bac29ed7dc721506f089cafa3156af475e6a/composer.json#L20).
As @jDolba  mentioned above, the `Step::$formType` property is set from configuration (`form_type`) where it is always string: https://github.com/shopsys/shopsys/blob/353eb791df97d32c9100bc885e0e46bbc6b39c99/project-base/src/Shopsys/ShopBundle/Form/Front/Order/OrderFlow.php#L47-L52
This value is then used for creating new form type in https://github.com/craue/CraueFormFlowBundle/blob/13a2bac29ed7dc721506f089cafa3156af475e6a/Form/FormFlow.php#L971 where it is passed to `Symfony\Component\Form\FormFactoryInterface::create()`. Passing type instances into the method is deprecated since https://github.com/symfony/symfony/commit/3d9e5de2b9dfec8d39b57486064b6e881548a86b#diff-be514034dbf6559a8ea89b9a4b8f159f and removed since Symfony 3 (https://github.com/symfony/symfony/commit/3d9e5de2b9dfec8d39b57486064b6e881548a86b#diff-be514034dbf6559a8ea89b9a4b8f159f). As we use Symfony in version `"^3.4.8"` now, we can rely on that :slightly_smiling_face:

PR #455, opened 2018-09-09
Improved description where to place GitHub token in installation guide

LukasHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In installation guide is info, that "Token is located in services -> php-fpm -> build -> args -> github_oauth_token". But token is not there yet when you are doing clean install and I spent lot of time looking for it. 
|New feature| No 
|BC breaks| No 
|Fixes issues| 
|Standards and tests pass| -


PetrHeinz:
It seems you're using installation guide from `dev-master` to install a project in version `7.0.0-alpha5`.

When working with a specific version, you should use a corresponding docs, see https://github.com/shopsys/shopsys/tree/v7.0.0-alpha5/docs

PetrHeinz:
We've added a note about this and recommend installing current `dev-master` until next release (#456). This should avoid such issues for now.

Issue #435, opened 2018-08-29
monorepo-tools: splitting to an existing repository with signed commits


dominikkaluza:
### What is happening
It's not possible to split to an existing repository if the repository contains commits signed by GitHub (typically init commit or merge commits from pull requests).

#### How to reproduce:
- Crate a new repository on GitHub with `Initialize this repository with a README` option checked
- Build a monorepo from this repository
- Try splitting the monorepo

#### Result:
`monorepo_split.sh` fails when trying to push to this repository

#### Workaround:
- edit line 36 in `monorepo_split.sh` and change `git push --tags $REMOTE master` to `git push --tags --force $REMOTE master` (this is needed only for the first split, `--force` flag can be removed after that)

### Expected result
`monorepo_split.sh` re-signs the commits and pushes them to the existing repository without using the `--force` flag

vitek-rostislav:
Hi @dominikkaluza, thanks for reporting the issue. We had the same problem in the past and solved it exactly the same way as you describe - by using `push --force`. At least, we should probably document this behavior :thinking: 

`monorepo_split.sh` could sign the commit using the same GPG key (https://help.github.com/articles/signing-commits-using-gpg/) that is used in the original github repository but this would work only if all the commits in the original repository were signed by the key.

PetrHeinz:
I suppose it is theoretically possible to store all data about the signature into the commit msg during building the monorepo and then use this data to recreate exactly the same signed commit during the splitting. I haven't looked into it yet, but it should be possible to store all raw data of the commit encoded in the commit msg ([it's possible to store 100MiB there](https://stackoverflow.com/questions/9733757/maximum-commit-message-size)) and during splitting restore the exact commit...

This could be a messy and time-consuming workaround, I would suggest to just document this behavior instead. We could leave this issue open if someone from the community want to take a look at it.

PR #461, opened 2018-09-12
monorepo-tools: add note about splitting signed commits

### LINKED ISSUE #445: monorepo-tools: undocumented behavior with signed commits
### What is happening

As mentioned in #435, repos that are converted into a monorepo and split back are not identical if the original repos contained signed commits.

Github signs commits by default, so it's not unusual to get into trouble and our documentation does not cover this scenario.

### Expected result

This behavior is documented (in the [README.md](https://github.com/shopsys/shopsys/blob/master/packages/monorepo-tools/README.md)) along with a description of what to do in such a scenario (i.e. force push). 

The possibility of re-signing the commits is mentioned in the issue #435.

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We should be more explicit about the behavior of monorepo-tools in respect to keeping the history intact. GitHub signs all commits made via its web UI (even the init commit during repository creation) by default, so it's likely that people will run into issues at this step. It would be nice to explore the possibility of re-signing the commits (see #435), but until then we should document the current state and the commonly used workaround.
|New feature| No 
|BC breaks| No 
|Fixes issues| fixes #445 

### DIFF
--- a/packages/monorepo-tools/README.md
+++ b/packages/monorepo-tools/README.md
@@ -76,6 +76,14 @@ Other branches are not pushed.
 
 It may again take a while, depending on the size of your monorepo.
 
+***Note:***  
+*The commits in the split repositories should be identical to those from the original repo, keeping the git history intact.*
+*Thus, if you have checked out the original `master` previously, you should be able to fast-forward to the new version after splitting.*  
+*The only known exception is a signed commit (note that GitHub signs commits made via its web UI by default).*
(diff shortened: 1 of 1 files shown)


PetrHeinz:
@vitek-rostislav I've noticed we do that for multiline notes in Markdown, see /docs/introduction/running-acceptance-tests.md:11-13 or search for "*Notes:*" - there is a list with every item individually italicized.

Personally, I think it's more readable and more diff-friendly this way.

Issue #440, opened 2018-09-02
Use global package hirak/prestissimo in composer


MattCzerner:
Our application has many dependencies and requierements, whole composer install/update takes too much time.

### Expected result
Use global composer package for parallel composer installation. It speeds up installation about 60-70%.

https://github.com/hirak/prestissimo

PR #496, opened 2018-09-20
Require global package hirak/prestisimo in docker images

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Our composer install takes too long because of number of dependencies. hirak/prestisimo package allows paralel installation of dependencies. This cause dramatic improvement of composer install performance
|New feature| No
|BC breaks| No
|Fixes issues| #440

I did performance tests of our composer install after few optimalizations.

I tested 4 scenarios and used this process to make sure that installation is clear without any cache:

```
git checkout <branch-name>
docker-compose up -d --build --force-recreate
docker-exec -it shopsys-framework-php-fpm sh
    rm -rf vendor/
    rm composer.lock
    composer clearcache
    composer install
```

Master - using forks
9m12s

Master - using patches (one patch does not apply)
4m 39s

Master with hirak/prestisimo package for paralel installetion of composer dependencies - using forks
7m 31s

Master with hirak/prestisimo package for paralel installetion of composer dependencies - using patches (one patch does not apply)
1m 17s

### DIFF
--- a/project-base/docker/php-fpm/Dockerfile
+++ b/project-base/docker/php-fpm/Dockerfile
@@ -70,3 +70,6 @@ USER www-data
 # allow configuring the GitHub OAuth token
 ARG github_oauth_token
 RUN if [ -n "$github_oauth_token" ]; then composer config -g github-oauth.github.com $github_oauth_token; fi
+
+# hirak/prestissimo makes the install of Composer dependencies faster by parallel downloading
+RUN composer global require hirak/prestissimo
--- a/microservices/product-search-export/docker/Dockerfile
+++ b/microservices/product-search-export/docker/Dockerfile
@@ -38,6 +38,9 @@ USER www-data
 ARG github_oauth_token
 RUN composer config -g github-oauth.github.com $github_oauth_token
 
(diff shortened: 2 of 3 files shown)


project-base/docker/php-fpm/Dockerfile
```diff
@@ -70,3 +70,6 @@ USER www-data
 # allow configuring the GitHub OAuth token
 ARG github_oauth_token
 RUN if [ -n "$github_oauth_token" ]; then composer config -g github-oauth.github.com $github_oauth_token; fi
+
+# hirak/prestissimo makes the install of Composer dependencies faster by parallel downloading
+RUN composer global require hirak/prestissimo
```
Miroslav-Stopka: Can we move this package into composer.json? :)

Miroslav-Stopka: But I'm thinking now, If this package will be placed into composer.json, that will be too late for faster composer install. Am I right? I need to be sure.

MattCzerner: It is not default composer package. it is required globally, thats why it is added during dockerfile build :)

PR #497, opened 2018-09-21
[project-base] adding php.ini to image is now done only in dockerfiles

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Mounting of php.ini using volumes does not apply the changes without manually restarting the php-fpm process. Because of that, the mounting is not used anymore and instead it is added using COPY during Docker image build. That way, if anything changes in php.ini, it is added to the image during build and loaded when the container starts.
|New feature| No
|BC breaks| No
|Fixes issues| ...

### DIFF
--- a/docker/conf/docker-compose-mac.yml.dist
+++ b/docker/conf/docker-compose-mac.yml.dist
@@ -36,7 +36,6 @@ services:
             - shopsys-framework-sync:/var/www/html
             - shopsys-framework-vendor-sync:/var/www/html/vendor
             - shopsys-framework-web-sync:/var/www/html/project-base/web
-            - ./project-base/docker/php-fpm/php-ini-overrides.ini:/usr/local/etc/php/php.ini:delegated
         links:
             - microservice-product-search
             - microservice-product-search-export
@@ -57,7 +56,6 @@ services:
         container_name: shopsys-framework-microservice-product-search
         volumes:
             - shopsys-framework-microservice-product-search-sync:/var/www/html
-            - ./microservices/product-search/docker/php-ini-overrides.ini:/usr/local/etc/php/php.ini
         links:
             - elasticsearch
         depends_on:
@@ -72,7 +70,6 @@ services:
         container_name: shopsys-framework-microservice-product-search-export
         volumes:
             - shopsys-framework-microservice-product-search-export-sync:/var/www/html
-            - ./microservices/product-search-export/docker/php-ini-overrides.ini:/usr/local/etc/php/php.ini
         links:
             - elasticsearch
         depends_on:
--- a/project-base/docker/conf/docker-compose-mac.yml.dist
+++ b/project-base/docker/conf/docker-compose-mac.yml.dist
@@ -36,7 +36,6 @@ services:
             - shopsys-framework-sync:/var/www/html
             - shopsys-framework-vendor-sync:/var/www/html/vendor
             - shopsys-framework-web-sync:/var/www/html/web
-            - ./docker/php-fpm/php-ini-overrides.ini:/usr/local/etc/php/php.ini:delegated
         links:
             - postgres
             - redis
--- a/project-base/docker/php-fpm/Dockerfile
+++ b/project-base/docker/php-fpm/Dockerfile
@@ -60,6 +60,9 @@ ENV LC_ALL=en_US.utf8 LANG=en_US.utf8 LANGUAGE=en_US.utf8
 # overwrite the original entry-point from the PHP Docker image with our own
 COPY docker-php-entrypoint /usr/local/bin/
 
+# copy php.ini configuration
+COPY php-ini-overrides.ini /usr/local/etc/php/php.ini
+
 # allow overwriting UID and GID o the user "www-data" to help solve issues with permissions in mounted volumes
 # if the GID is already in use, we will assign GID 82 instead (82 is the standard uid/gid for "www-data" in Alpine)
 ARG www_data_uid
--- a/project-base/docker/php-fpm/ci/Dockerfile
+++ b/project-base/docker/php-fpm/ci/Dockerfile
@@ -63,6 +63,9 @@ ENV LC_ALL=en_US.utf8 LANG=en_US.utf8 LANGUAGE=en_US.utf8
 # overwrite the original entry-point from the PHP Docker image with our own
 COPY project-base/docker/php-fpm/docker-php-entrypoint /usr/local/bin/
 
+# copy php.ini configuration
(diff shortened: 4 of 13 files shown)


MattCzerner:
Php.ini is mounted there but it does not apply changes if you change something. I dont think that php.ini configuration isnt something that should be mounted, it is changed really rarely and it could improve performance for docker-sync. 

In kubernetes, there is not any reason to make it volume.

PetrHeinz:
I agree, that there is no reason to mount the config files in Kubernetes manifests. I would remove it from there, just as you suggest.

---

I would argue that the config files should behave consistently in Docker Compose during development - look at https://github.com/shopsys/shopsys/blob/v7.0.0-beta1/docker/conf/docker-compose.yml.dist and notice that there are `postgres.conf`, `nginx.conf`, `php-ini-overrides.ini` and  another `php-ini-overrides.ini` for each microservice.

Some of these mounts are necessary because official Docker images are used (and we don't want to create our own Dockerfile atm). Some of these mounts are redundant - they overwrite already copied files during build. Mounting of `php-ini-overrides.ini` inside the `php-fpm` container is necessary as we don't copy the file into the image directly, even though we build the image ourselves.

Personally, I think that from the point of view of a developer using monorepo, all config files should behave the same, or at least it should be clear what happens when.

---

This is a valuable insight for me in context of our planned mutli-stage Docker image building - we should probably make this behavior consistent and documented. Maybe use mounts for local development (with both the source codes and the configs NOT copied into the image), no mounting on our CI (as they are copied into the image) and no mounting to the product images? We might even need to build our own Docker images of services such as nginx, just because of having more control over tagging and having the config files copied inside them - or is the config-mounting in Kubernetes OK for production images?

MattCzerner:
Okay, ur right in case of consistency. I dont see really the problem in having 2 scenarios. 

1st - our dockerimages, we add it using dockerfile and not mount it at all, i think that there is no point, even if it is mounted it is not automatic to load this configuration into running php process.

2nd - 3rd party dockerimages, it would be wierd to create dockerfile just to add configurations. 

But i guess thats different problem then just removing volumes. Should i just delete mounts of kubernetes and mount of configs in docker-compose let be, since it will be decided in multistage US?

PetrHeinz:
I would be perfectly fine with both solutions: either just remove [the redundant Kubernetes mounts](https://github.com/shopsys/shopsys/blob/v7.0.0-beta1/.ci/build_kubernetes.sh#L23-L24), or remove all config mounts from our docker images and add the COPY instruction to both our `php-fpm` Dockerfiles (which would make [its Kubernetes mount](https://github.com/shopsys/shopsys/blob/v7.0.0-beta1/.ci/build_kubernetes.sh#L22) redundant as well).

project-base/docker/php-fpm/ci/Dockerfile
```diff
@@ -63,13 +63,17 @@ ENV LC_ALL=en_US.utf8 LANG=en_US.utf8 LANGUAGE=en_US.utf8
...
+COPY project-base/docker/php-fpm/php-ini-overrides.ini /usr/local/etc/php/php.ini
+
 # allow configuring the GitHub OAuth token
 ARG github_oauth_token
 RUN composer config -g github-oauth.github.com $github_oauth_token
 
 # hirak/prestissimo makes the install of Composer dependencies faster by parallel downloading
 RUN composer global require hirak/prestissimo
 
+
```
PetrHeinz: probably a mistake

MattCzerner:
Okay i deleted mounting of php-ini of php-fpm container too. 

Now it is only possible to reload php.ini configuration by build of dockerimage.

docker/conf/docker-compose-mac.yml.dist
```diff
@@ -35,8 +35,7 @@ services:
         volumes:
             - shopsys-framework-sync:/var/www/html
             - shopsys-framework-vendor-sync:/var/www/html/vendor
-            - shopsys-framework-web-sync:/var/www/html/project-base/web
-            - ./project-base/docker/php-fpm/php-ini-overrides.ini:/usr/local/etc/php/php.ini:delegated
+            - shopsys-framework-web-sync:/var/www/html/project-base/web:delegated
```
PetrHeinz: the delegated is probably a typo, please fix it or explain it

docker/conf/docker-compose-mac.yml.dist
```diff
@@ -4,7 +4,7 @@ services:
         image: postgres:10.5-alpine
         container_name: shopsys-framework-postgres
         volumes:
-            - ./project-base/docker/postgres/postgres.conf:/var/lib/postgresql/data/postgresql.conf:delegated
+            - ./project-base/docker/postgres/postgres.conf:/var/lib/postgresql/data/postgresql.conf
```
PetrHeinz: the removal of delegated is probably a mistake

project-base/docker/conf/docker-compose-mac.yml.dist
```diff
@@ -4,7 +4,7 @@ services:
         image: postgres:10.5-alpine
         container_name: shopsys-framework-postgres
         volumes:
-            - ./docker/postgres/postgres.conf:/var/lib/postgresql/data/postgresql.conf:delegated
+            - ./docker/postgres/postgres.conf:/var/lib/postgresql/data/postgresql.conf
```
PetrHeinz: this makes me think there is a reason for the changes - why are you doing this?

project-base/docker/php-fpm/Dockerfile
```diff
@@ -60,6 +60,9 @@ ENV LC_ALL=en_US.utf8 LANG=en_US.utf8 LANGUAGE=en_US.utf8
 # overwrite the original entry-point from the PHP Docker image with our own
 COPY docker-php-entrypoint /usr/local/bin/
 
+# copy php.ini configuration
+COPY project-base/docker/php-fpm/php-ini-overrides.ini /usr/local/etc/php/php.ini
```
PetrHeinz: this assumes that the Docker build context is the root of the monorepo, right? But the COPY above assumes the actual context is the `docker` directory. Both can't be true.

PetrHeinz:
Please, rename the PR to reflect the new changes and our rules about package-tagging.

MattCzerner:
I did it by global find and replace function and i didnt see the delegated options. 

I rebased it and squashed it into one lucid commit. Also I got a question about upgrade notes. Because you should copy new version of  `docker-compose.yml` but is not essential for application to work.

So do we write it into `UPGRADE.md` or not?

MattCzerner:
@PetrHeinz I updated name of PR and description, also i have added record to `UPGRADE.md` if you would find any gramatical mistake, you can override it.

Issue #477, opened 2018-09-17
Running on Docker and Windows 10 Pro is slow


LukasHeinz:
### What is happening
My setup:
- Win 10 Pro 64bit
- CPU Intel i5 2,5 GHz
- 4 GB RAM
- SSD disc
- Docker settings: 2 CPU, 1792 MB RAM, 1024 Swap

My experience SSFW performance:
- build-demo-dev (on 4th try): 50 minutes
- first load of homepage: 46 sec
- second (and next) load of homepage: 20 sec (but about 7 sec it is loading Symfony bar)

This is not suitable or real development on Docker and Win 10 Pro.

### Expected result

Load speed of pages will be faster so developer can work with it... or support Docker on Win is retracted

boris-brtan:
My setup:
- Win 10 Pro 64bit
- CPU intel i7-8550 2GHZ
- 16 GB RAM DDR4
- SSD 512GB
- Docker settings: 4 CPU, 4GB, 3072 Swap
- windows defender turned off
My experience SSFW performance:
- build-demo-dev-quick (on 1 try): 8 min
- build-demo-dev (first try): 40 min
  - tests: 28 min
    - tests-db: 10 min
    - tests-smoke: 15min
- first load of homepage: 54 sec
- second (and next) load of homepage: 13 sec

PR #499, opened 2018-09-24
[shopsys] Docker-sync for Windows development is added

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| development on windows os is very slow. for instance build-demo-dev lasts 45 minutes instead of 6 minutes and load of the page 20 seconds.
|New feature| No 
|BC breaks| No 
|Fixes issues| #477 


boris-brtan:
changed

boris-brtan:
sorry for the late response @simara-svatopluk but this is very hard to find, i think you are right we miss records in upgrade log, even these are optional i think but during demoshop upgrade we have problem with this thing.

Issue #508, opened 2018-10-08
Fix problems in settings with base data for production


boris-brtan:
If we want to setup shop for production version with phing target `build-new` there are created invalid records in table setting_values.  

### What is happening

for example administration page /admin/legal-conditions/setting/ doesn't work because of that.
problematic is method for transforming table records into its original values. 
if record has value NULL and type `integer` null is wrongly transformed into 0.

https://github.com/shopsys/demoshop/issues/17
### Expected result

Null value needs to have type `none` so it will be correctly transformed into NULL.

PR #509, opened 2018-10-08
[framework] Records of SettingValues table need to be valid with application

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| if the shop is created with build-new phing target, which should be state when there are basic data needed to create new shop, there are some pages that don't work because of inconsistency with the records of SettingsValue table when null value is converted into number 0.
|New feature| No 
|BC breaks| No 
|Fixes issues| #508 [Demoshop_17](https://github.com/shopsys/demoshop/issues/17)

### DIFF
--- a/packages/framework/src/Component/Setting/SettingValue.php
+++ b/packages/framework/src/Component/Setting/SettingValue.php
@@ -89,6 +89,11 @@ class SettingValue
      */
     public function getValue()
     {
+        if ($this->value === null && $this->type !== self::TYPE_NULL) {
+            $message = 'Setting value type "' . $this->type . '" does not allow null value.';
+            throw new \Shopsys\FrameworkBundle\Component\Setting\Exception\SettingValueTypeNotMatchValueException($message);
+        }
+
         switch ($this->type) {
             case self::TYPE_INTEGER:
                 return (int)$this->value;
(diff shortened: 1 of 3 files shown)


packages/framework/src/Component/Setting/SettingValue.php
```diff
@@ -89,6 +89,11 @@ public function getName()
      */
     public function getValue()
     {
+        if ($this->value === null && $this->type !== self::TYPE_NULL) {
+            $message = 'Correct setting value type for null value is "' . self::TYPE_NULL . '".';
```
TomasLudvik: I would change this message to: Setting value type "' . $this->type . "' does not allow null value.

packages/framework/src/Migrations/Version20181008000001.php
```diff
@@ -3,6 +3,8 @@
 namespace Shopsys\FrameworkBundle\Migrations;
 
 use Doctrine\DBAL\Schema\Schema;
+use Shopsys\FrameworkBundle\Component\Setting\Setting;
+use Shopsys\FrameworkBundle\Model\Product\Availability\Availability;
```
vitek-rostislav: this is unused :wink:

Issue #469, opened 2018-09-14
Migrations: `db-migrations-count` command is failing


boris-brtan:
### What is happening

We presume that after these implementations
https://github.com/shopsys/shopsys/pull/39
https://github.com/shopsys/shopsys/pull/185

commands like
```bash
php phing db-migrations-count
php bin/console shopsys:migrations:count
```
started to fail
### Expected result

commands will not fail and we should get the result of the migrations that are not applied.

PR #515, opened 2018-10-10
[migrations] migration count command is now fixed and initialized with correct configuration

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| migration command shopsys:migrations:count is not initialized with migration configuration and some of the methods that depends on validate() method are failing.
|New feature| No 
|BC breaks| No 
|Fixes issues| #469 


boris-brtan:
thX

Issue #516, opened 2018-10-11
Impossible to display product detail of new product (after clean installation - build-new)


vitek-rostislav:
### What is happening

After the clean installation by `php phing build-new`, I created a default unit and availability (as suggested on the dashboard) and then tried to create a new product - it worked but I was redirected to the detail of my new product where I got an error because there is no `ProductVisibility` for the second domain and therefore rendering of the visibilities box fails.

![image](https://user-images.githubusercontent.com/10401898/46793090-2a3edb00-cd45-11e8-8d1b-69f27b68bf7d.png)

The reason why there is no `ProductVisibility` is that there is no pricing group for the second domain without installing data fixtures - this issue should be solved in #513.

### Expected result
I believe that when rendering the visibility box (using `ProductVisibilityExtension::isVisibleForDefaultPricingGroupOnDomain()`), missing entry in `product_visibilities` DB table should not cause an error - when the entry does not exist, the app should simply assume that the product is not visible

PR #513, opened 2018-10-09
[framework] Manipulation with domains is modified and documented now

Miroslav-Stopka:
- now the CreateDomainsDataCommand creates new default pricing group for every new domain
- for pricing groups that are not filled in demo data, the manual price is set to 0 while importing demo data
- data fixtures of products for single domain are imported in locale according to the domains locale

| Q             | A
| ------------- | ---
|Description, reason for the PR| Testing of phing targets for manipulation with domains to confirm consistency database after some modifications of domains. Documenting manipulation with domains for project developers.
|New feature| No
|BC breaks| No
|Fixes issues| ...
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #517, opened 2018-10-11
[framework] fixed rendering of product visibilities box

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Rendering of product visibilities box in admin ends up with an error when the entry is not found in `product_visibilities` DB table
|New feature| No 
|BC breaks| No 
|Fixes issues| #516 

### DIFF
--- a/packages/framework/src/Twig/ProductVisibilityExtension.php
+++ b/packages/framework/src/Twig/ProductVisibilityExtension.php
@@ -70,7 +70,11 @@ class ProductVisibilityExtension extends \Twig_Extension
     public function isVisibleForDefaultPricingGroupOnDomain(Product $product, $domainId)
     {
         $pricingGroup = $this->pricingGroupSettingFacade->getDefaultPricingGroupByDomainId($domainId);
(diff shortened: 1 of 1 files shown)


PetrHeinz:
The ProductVisibilities should be created for all products at all times. Supressing errors coming from database inconsistency may lead to even more problems... What is the use case in which this error happens?

Wouldn't it make more sense to enforce the consistency of the database - i.e. make sure the records are properly created?

PetrHeinz:
Oh, I've checked the #516 and you're explaing the scenario pretty well. I still don't understand one thing - you're saying the reason is that the second domain has no pricing group, but you're able to get the pricing group via `PricingGroupSettingFacade::getDefaultPricingGroupByDomainId()` and that's pretty weird...

boris-brtan:
is this still needed after #513 implementation?

vitek-rostislav:
Hi guys, thanks for your comments, I have checked it once more and the issue reported in #516 is not occurring anymore after #513 was merged :+1:

PR #519, opened 2018-10-11
[framework] feeds generation now works with performance data

TomasLudvik:
- feeds are now generated on local filesystem and moved to abstract filesystem after generation is complete
- feeds files are moved also to the abstract filesystem in case that generation of feed is put to sleep and moved back on feed wakeup

| Q             | A
| ------------- | ---
|Description, reason for the PR| Generating of feeds was problematic around 50k+ products
|New feature| No 
|BC breaks| No 
|Fixes issues| ...

### DIFF
--- a/packages/framework/src/Model/Feed/FeedExport.php
+++ b/packages/framework/src/Model/Feed/FeedExport.php
@@ -4,7 +4,9 @@ namespace Shopsys\FrameworkBundle\Model\Feed;
 
 use Doctrine\ORM\EntityManagerInterface;
 use League\Flysystem\FilesystemInterface;
+use League\Flysystem\MountManager;
 use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
+use Symfony\Component\Filesystem\Filesystem;
 
 class FeedExport
 {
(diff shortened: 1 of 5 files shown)


packages/framework/src/Model/Feed/FeedExport.php
```diff
@@ -29,7 +31,17 @@ class FeedExport
     /**
      * @var \League\Flysystem\FilesystemInterface
      */
-    protected $filesystem;
+    private $filesystem;
```
boris-brtan: was this change somehow connected to the implementation of `private $localFileSystem` in `packages/framework/src/Model/Feed/FeedExportFactory.php` or what is the reason for changing the scope to private  ?

packages/framework/src/Model/Feed/FeedExportFactory.php
```diff
@@ -31,16 +33,30 @@ class FeedExportFactory
...
+     * @var \Symfony\Component\Filesystem\Filesystem
+     */
+    private $localFilesystem;
+
+    /**
+     * @var \League\Flysystem\MountManager
+     */
+    private $mountManager;
+
     public function __construct(
```
boris-brtan: does it make sense to create annotations for constructors like it is in document about decision about annotations?

PR #533, opened 2018-10-17
[shopsys/shopsys] main php-fpm container now uses multi-stage build feature

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Multiple Dockerfiles has been squashed into one
|New feature| Yes 
|BC breaks| Yes 
|Fixes issues| ...
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #535, opened 2018-10-18
[shopsys] added .dockerignore files

PetrHeinz:
- makes the Docker image build quicker by reducing time needed to send the context to Docker daemon and for the `COPY /` instruction
- allows for better cache reuse (irrelevant changes, eg. to the docs, do not invalidate the cache)
- not sure what exactly should be ignored, so consider this just a suggestion (especially not sure about the vendor directory, as it makes sense to ignore it, but it slows down development builds if you've already installed your dependencies)

| Q             | A
| ------------- | ---
|Description, reason for the PR| #533 slows down building the images, because it uses much larger context and copies it into the image, which takes time and has to be redone for even irrelevant changes to the files (eg. docs change)
|New feature| No 
|BC breaks| No 
|Fixes issues| none


PetrHeinz:
Hi @TomasLudvik and thanks for the review. Altering `docker-sync.yml`s or volume mounts seem a bit out of scope of this PR. I wanted to focus on reducing the size of just the Docker build context. Furthermore, as I use linux, I have no option to thoroughly test the docker-sync.

Seeing that to add the directories there is not much work, I've added them as you requested in 
fdead55. If there are any doubts or issues with the docker-sync changes, I will separate the commit into its own PR for the reasons above.

I've also noticed three slight inconsistencies in the docker-sync files:
- `docker/conf/docker-sync-win.yml.dist` does not exclude `microservices`
- `project-base/docker/conf/docker-sync.yml.dist` contains extra record for `docker/` (is it different from `docker` anyhow?)
- `project-base/docker/conf/docker-sync.yml.dist` is not ordered alphabetically (`var/postgres-data` is before `var/elasticsearch-data`)

Should the files be consistent, or do some of these have any meaning? If they're not meant to be copies, why mention `.DS_Store` in Windows configuration?

---

I'm a bit surprised by the lack of comments on the actual list of ignored files. Does the suggested `.dockerignore` content make sense to you as-is? What do you think about ignoring the `vendor` dir and its influence on dev build as mentioned in the desc? Can you think of other directories that should definitely be ignored here?

TomasLudvik:
I have rebased this branch for testing purposes on master so it includes multi-stage build. I have tested build before and after changes.

Before it took 08:45, after it took 06:42.

boris-brtan:
this could be removed 
https://github.com/shopsys/shopsys/blame/ph-dockerignore/docs/installation/installation-using-docker-on-production-server.md#L220-L223

PetrHeinz:
I've added the upgrade notes and updated the docs as @boris-brtan suggested. I've also rebased the branch onto current master to avoid conflicts in the `UPGRADE.md`. No other changes were done.

Issue #532, opened 2018-10-17
Missing links to production installation


LukasHeinz:
### What is happening
In documention there are no links to productiuon installation of SSFW

### Expected result
There should be in readme and docekr install page.

vitek-rostislav:
Is this still relevant? We have links to our prod. installation guide in here - https://github.com/shopsys/shopsys/blob/master/docs/index.md

LukasHeinz:
In other files like https://github.com/shopsys/shopsys/blob/master/docs/installation/installation-using-docker.md or readme.md the guide is reffered as "the installation". But since production installation there are two types of installation - developement and production. These two should be more clearly separated and linked together.

LukasHeinz:
This issue has been resolved in previous months.

PR #543, opened 2018-10-23
[project-base] the environment setting during "composer install" is not interactive

PetrHeinz:
- "composer install" previously created `PRODUCTION`/`DEVELOPMENT` file only in interactive mode, based on user's input
- now the file is created even in non-interactive mode, base on whether `composer install` or `composer install --no-dev` was called
- the file is still created only if there is no environment file yet

| Q             | A
| ------------- | ---
|Description, reason for the PR| current behavior needs user interaction, even when unnecessary (user already used either dev build or production build), also it leads to unspecified environment when installed non-interactively (`prod` is used by default)
|New feature| No 
|BC breaks| default non-interactive installation changed from `prod` to `dev` 
|Fixes issues| ...


LukasHeinz:
@PetrHeinz I do not feel good about the fact that we remove for user the ability to choose environment. The best solution for me is to document how environment works in install guide. If you explain that these installs are for devs so dev environment and those are for production so PROD env I am OK with that (it is connected with issue #532). There would be nice to include also info to change environment.

PR #557, opened 2018-10-30
[project-base] php-fpm image has standard workdir (/var/www/html) in ci stage

PetrHeinz:
- this commit just removes an inconsistency between the stages
- previously, the workdir was set to /source-code, so it can be copied into a mounted volume in an init container in Kubernetes
- this can be also achieved by mounting the volume into different directory of the init container (/tmp/source-codes)
- in the final container, the same volume should be of course mounted to its final destination (/var/www/html)
- because the path is not changed, the autoload cache cleaning should not be necessary

| Q             | A
| ------------- | ---
|Description, reason for the PR| removal of a hacky solution
|New feature| No 
|BC breaks| Yes, but only for a CI stage 
|Fixes issues| ...

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -65,6 +65,8 @@ There is a list of all the repositories maintained by monorepo, changes in log b
 - *(optional)* [#535 added .dockerignore files](https://github.com/shopsys/shopsys/pull/535)
     - to make your Docker image build faster, copy the `.dockerignore` file to the root of you project
     - if you're using Docker-sync, add the directories mentioned in the PR into `sync_exclude` section of your `docker-sync.yml` to make the synchronization faster as well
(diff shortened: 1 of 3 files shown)


MattCzerner:
Hi @PetrHeinz,
I like it. It simplifies that temporary workdir workaround to more understandable level and it feels much more clean and magicless :) 

Please add the UPGRADE.md section, and I would actually let someone less interested in Kubernetes read the PR descriptions, because I understand what happened and why, but I wonder if it is understandable to someone else.

Good Job!!

PetrHeinz:
I've added the optional upgrade note. It's ready for CR.

boris-brtan:
TESTS ARE OK

PetrHeinz:
Rebased just to resolve conflicts with #558

PR #566, opened 2018-10-31
[project-base] Set development docker build target before production and CI targets

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| current stable version of docker doesn't allow building targeted stage but it builds stages from top to bottom until targeted stage occurs, so we can speed up build for development target because development target do not need to build images with the source code because all the code is mounted as a volume.
|New feature| No 
|BC breaks| No 
|Fixes issues| -

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -116,6 +116,8 @@ There is a list of all the repositories maintained by monorepo, changes in log b
       as command `shopsys:fixtures:load` doesn't exist anymore and remove `--fixtures` argument
     - rename `db-fixtures-demo-singledomain` to `db-fixtures-demo`
     - rename `test-db-fixtures-demo-singledomain` to `test-db-fixtures-demo`
+- *(optional)* [#566 - Set development docker build target before production and CI targets](https://github.com/shopsys/shopsys/pull/566)
+    - move `development` stage build before `production` stage in `docker/php-fpm/Dockerfile` to make your dev build faster
 
 ### [shopsys/shopsys]
 - *(MacOS only)* [#503 updated docker-sync configuration](https://github.com/shopsys/shopsys/pull/503/)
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -81,6 +81,8 @@ There is a list of all the repositories maintained by monorepo, changes in log b
         - run `docker-sync start` to create volumes
     - run `docker-compose up -d --build --force-recreate` to start application
     - phing target for checking and fixing standards has changed, update `build-dev.xml` according to the changes
+- *(optional)* [#566 - Set development docker build target before production and CI targets](https://github.com/shopsys/shopsys/pull/566)
+    - move development stage build before production stage `docker/php-fpm/Dockerfile` and set default UID and GID to 82
```
PetrHeinz: I would suggest providing info about the reasons directly in the upgrade note:
```suggestion
    - move `development` stage build before `production` stage in `docker/php-fpm/Dockerfile` to make your dev build faster
    - set default UID and GID to 82 to make the build arguments optional (for building `ci` and `production` stage)
```

project-base/docker/php-fpm/Dockerfile
```diff
@@ -79,6 +79,23 @@ RUN composer global require hirak/prestissimo
...
 
+FROM base as development
+
+USER root
+
+# allow overwriting UID and GID o the user "www-data" to help solve issues with permissions in mounted volumes
+# if the GID is already in use, we will assign GID 82 instead (82 is the standard uid/gid for "www-data" in Alpine)
+ARG www_data_uid=82
+ARG www_data_gid=82
+RUN deluser www-data && (addgroup -g $www_data_gid www-data || addgroup -g 82 www-data) && adduser -u $www_data_uid -D -S -G www-data www-data
```
PetrHeinz: In sake of consistency, I would suggest making the arguments optional in the same way we do it in microservices: https://github.com/shopsys/shopsys/blob/65c68a133161f0bf3848db2795359e5412fc263d/microservices/product-search/docker/Dockerfile#L198-L202

This would of course require rewriting the upgrade note to something like "_make the `www_data_*id` arguments optional using an `if` condition (for building `ci` and `production` stage)_".

boris-brtan:
THX, fixups created

PR #538, opened 2018-10-19
[shopsys] phing targets: create-domains-data is now dependent on create-domains-db-functions

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This is a fix of https://github.com/shopsys/shopsys/pull/513/commits/36717a3c75180865c30923e5a58232950418ed7f. In the linked commit, creation of db functions was extracted from `create-domains-data` target but new target was not used in all places where the original target was used
|New feature| No 
|BC breaks| Yes 
|Fixes issues| ...
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #570, opened 2018-11-02
[shopsys] improved upgrading instructions for #538

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| https://github.com/shopsys/shopsys/pull/538#discussion_r230324919
|New feature| No 
|BC breaks| No 
|Fixes issues| ...


simara-svatopluk:
Well, this is better.

Now imagine I have more instructions for phing targets. I would like to add these instruction next to your instructions, so the user can do all changes in one file. But I cannot do this because of the PR introduction...

This is one of reasons I don't prefer to have upgrade.md structured in PR-instructions, PR-instructions, but rather instructions (see PR), instructions (see PR). Because if you had it in this fashion, I could add instructions, more instructions (see PR1, PR2). But I don't like links to PRs in upgrade.md anyway because they only confuses me.

vitek-rostislav:
I see your point :slightly_smiling_face: It seems that we need to discuss the style of writing upgrading notes anyway with the whole team... @PetrHeinz had some notes as well

simara-svatopluk:
yes. Anyway I'm approving this PR as it improved upgrade.md

PR #581, opened 2018-11-08
[framework] Product creation: calculated availability is set additionally if necessary

### LINKED ISSUE #189: Converting to variants ends with an error of 500 if the main product is not available
When setting up a product, I do not have to choose availability if I use the "stock stock". This is a created product that has calculatedAvailability but not availibility.

When I want to create variants from this product, the system shouts to me that the product does not have the availability and the system drops into the database.

The option is the raw availability of the main product.

This is a business issue that will be followed by programming.

Analysis

- We are considering removing SSFW variants
- But how could it be solved:
     - The product in the administration should always include availability
         - Change forms, logic
         - It does not require any further interventions
     - The Product Designer should get the product into a consistent state
         - The Product Designer would also include default accessibility
         - The Factory on Products could cater for this dependence
         - Construction is swelling
         - This constructor parameter will almost never be used
     - The Product Factory will contain a creation logic
         - We will not dig into the Product constructor
         - The logic will have to be copied by the factory implementor in the project-base
               - We break forward compatibility

Closed Partial PR: https://github.com/shopsys/shopsys/pull/120

vitek-rostislav:
- `ProductFactory` is now responsible for setting calculated availability to the new product (if necessary)

| Q             | A
| ------------- | ---
|Description, reason for the PR| When a developer is trying to create the main variant from a product that has nullable availability, the null value is then propagated to `$calculatedAvailability` attribute of the new product. This is a problem as the property is not nullable - therefore application fails when trying to persist the new product. It was already (hot)fixed in `ProductFacade::create()` but not in `ProductVariantFacade::createVariant()`
|New feature| No 
|BC breaks| No 
|Fixes issues| fixes #189

### DIFF
--- a/packages/framework/src/Model/Product/ProductFactory.php
+++ b/packages/framework/src/Model/Product/ProductFactory.php
@@ -3,6 +3,7 @@
 namespace Shopsys\FrameworkBundle\Model\Product;
 
 use Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver;
+use Shopsys\FrameworkBundle\Model\Product\Availability\ProductAvailabilityCalculation;
 
 class ProductFactory implements ProductFactoryInterface
 {
@@ -12,11 +13,20 @@ class ProductFactory implements ProductFactoryInterface
     protected $entityNameResolver;
 
     /**
-     * @param \Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver $entityNameResolver
+     * @var \Shopsys\FrameworkBundle\Model\Product\Availability\ProductAvailabilityCalculation
      */
-    public function __construct(EntityNameResolver $entityNameResolver)
-    {
+    protected $productAvailabilityCalculation;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver $entityNameResolver
+     * @param \Shopsys\FrameworkBundle\Model\Product\Availability\ProductAvailabilityCalculation $productAvailabilityCalculation
+     */
+    public function __construct(
+        EntityNameResolver $entityNameResolver,
+        ProductAvailabilityCalculation $productAvailabilityCalculation
+    ) {
         $this->entityNameResolver = $entityNameResolver;
+        $this->productAvailabilityCalculation = $productAvailabilityCalculation;
     }
 
     /**
@@ -27,7 +37,10 @@ class ProductFactory implements ProductFactoryInterface
     {
         $classData = $this->entityNameResolver->resolve(Product::class);
 
-        return $classData::create($data);
+        $product = $classData::create($data);
+        $this->setCalculatedAvailabilityIfMissing($product);
+
+        return $product;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


packages/framework/src/Model/Product/ProductFactory.php
```diff
@@ -39,6 +50,25 @@ public function createMainVariant(ProductData $data, array $variants): Product
...
-        return $classData::createMainVariant($data, $variants);
+        $mainVariant = $classData::createMainVariant($data, $variants);
+        $this->setMissingCalculatedAvailability($mainVariant);
+
+        return $mainVariant;
+    }
+
+    /**
+     * When creating new product, ProductData::$availability is set to Product::$calculatedAvailability property as well.
+     * This is a problem when ProductData::$availability === null
```
PetrHeinz: I would put the sentence on one line:
```suggestion
     * This is a problem when ProductData::$availability === null as $calculatedAvailability is not nullable (unlike $availability).
```

packages/framework/src/Model/Product/ProductFactory.php
```diff
@@ -39,6 +50,25 @@ public function createMainVariant(ProductData $data, array $variants): Product
...
+        $mainVariant = $classData::createMainVariant($data, $variants);
+        $this->setMissingCalculatedAvailability($mainVariant);
+
+        return $mainVariant;
+    }
+
+    /**
+     * When creating new product, ProductData::$availability is set to Product::$calculatedAvailability property as well.
+     * This is a problem when ProductData::$availability === null 
+     * as "$calculatedAvailability" is not nullable (unlike $availability)
```
PetrHeinz:
```suggestion
```

packages/framework/src/Model/Product/ProductFactory.php
```diff
@@ -39,6 +50,25 @@ public function createMainVariant(ProductData $data, array $variants): Product
...
+     * @see \Shopsys\FrameworkBundle\Model\Product\Product::__construct()
+     *
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     */
+    protected function setMissingCalculatedAvailability(Product $product) {
```
PetrHeinz: I would consider renaming it to `setCalculatedAvailabilityIfMissing` as current name could be slightly misleading.

packages/framework/tests/Unit/Model/Product/ProductVariantServiceTest.php
```diff
@@ -13,7 +16,7 @@ class ProductVariantServiceTest extends TestCase
 {
     public function testCheckProductIsNotMainVariantException()
     {
-        $productVariantService = new ProductVariantService(new ProductFactory(new EntityNameResolver([])));
+        $productVariantService = new ProductVariantService(new ProductFactory(new EntityNameResolver([]), $this->getAvailabilityFacadeMock()));
```
PetrHeinz: Seeing how many lines you had to modify, wouldn't it make more sense to extract this into a `setUp` method?

Just assign `$this->productVariantService` and let all test methods use it.

boris-brtan:
Seems to be working now. i set `availability_id = NULL` then i tried to create variant on master branch and it failed on constraint violation and when i switched into this branch vatiant was created.
TESTS OK.

PR #597, opened 2018-11-13
[shopsys] update firewalld paths to create services

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| firewalld custom services should be created in /etc/firewalld/*, because firewalld can rewrite services in /usr/lib/ after update package
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 


vitek-rostislav:
Hi @henzigo, thanks for the PR :+1: We are finishing our current sprint the following Tuesday (20.11.) so in the worst case scenario we will take a look into it at the beginning of the upcoming sprint next week :slightly_smiling_face:

PR #528, opened 2018-10-16
[shopsys] Removed depends_on and links from docker-compose.yml files

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Links are deprecated in docker and depends_on is not necessary. To keep docker-compose files simple and consistent, deleted `links` and `depends_on` from docker-compose files
|New feature| No
|BC breaks| No
|Fixes issues| resolves #505 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #598, opened 2018-11-13
[project-base] fix folder sharing for production docker-compose

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Volumes on production are not shared well. Webserver container mounts empty folders, but we need data from php-fpm container.
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 


LukasHeinz:
Thanks for pull request. Our developers will look into it soon. 

Just for further reference: the depends_on was removed from docekr files recently - for more see https://github.com/shopsys/shopsys/pull/528

boris-brtan:
Thx, it has something to do with the creation of named volumes with the content from the container and sharing it between others.

boris-brtan:
TESTS:
without depends_on launching of containers is random, sometimes is webserver before php-fpm but no matter the order in console there is `empty` named volume in web folder instead of files from php-fpm `web` and `default.conf` in nginx named volume instead of `nginx.conf`.

With `depends_on: php-fpm` setting there is still php-fpm container created before webserver and volumes contain correct files.

henzigo:
@PetrHeinz I'm ok with that. It's nice to see that you can easily update what you need to.

PR #603, opened 2018-11-15
[shopsys] docs: describe the conflict of firewalld with docker

henzigo:
taken from https://docs.docker.com/v1.6/installation/centos/#installing-docker-centos-7

| Q             | A
| ------------- | ---
|Description, reason for the PR| In Installation guide for production should be information about conflict firewalld with docker after restart firewall.
|New feature| No 
|BC breaks| No 
|Fixes issues| 

### DIFF
--- a/docs/installation/installation-using-docker-on-production-server.md
+++ b/docs/installation/installation-using-docker-on-production-server.md
@@ -26,6 +26,10 @@ firewall-cmd --permanent --direct --add-chain ipv4 filter DOCKER
 firewall-cmd --permanent --direct --add-rule ipv4 filter DOCKER 0 ! -s 127.0.0.1 -j RETURN
 ```
 
+When `firewalld` is started or restarted it will remove the `DOCKER` chain from iptables, preventing Docker from working properly.
+
+When using Systemd, `firewalld` is started before Docker, but if you start or restart `firewalld` after Docker, you will have to restart the Docker daemon.
+
 ### Nginx
 
 Let's presume that we want to have our site running on `HTTPS` protocol and everything that concerns domain and its `certificates` is already setup.


henzigo:
Information was taken over https://docs.docker.com/v1.6/installation/centos/#installing-docker-centos-7

boris-brtan:
THX, for the info.
if it is about docker chain, in the cofiguration there is set
`firewall-cmd --permanent --direct --add-chain ipv4 filter DOCKER` so after firewall restart it will recreate this chain.
I tried this on server and after firewall restart everything seemed to work. But maybe there could be some configuration requirements which can have problems.
Can you provide us with some use cases? or can we arrange some quick meeting about this issue?

henzigo:
Sorry, I am confused. I gave you link to official documentation where they highlight the problem. So i do not understand why I have to provide use case with this problem. 

If you do not believe in official documentation, you can try `systemctl status firewalld -l` on your production server. You should see something like this:

```
firewalld[497]: WARNING: COMMAND_FAILED: '/usr/sbin/iptables -w2 -t nat -C POSTROUTING -p tcp -s 192.168.0.4 -d 192.168.0.4 --dport 8080 -j MASQUERADE' failed: iptables: No chain/target/match by that name.
firewalld[497]: WARNING: COMMAND_FAILED: '/usr/sbin/iptables -w2 -t nat -C DOCKER -p tcp -d 0/0 --dport 8000 -j DNAT --to-destination 192.168.0.4:8080 ! -i br-710051d83031' failed: iptables: No chain/target/match by that name.
firewalld[497]: WARNING: COMMAND_FAILED: '/usr/sbin/iptables -w2 -t filter -C DOCKER ! -i br-710051d83031 -o br-710051d83031 -p tcp -d 192.168.0.4 --dport 8080 -j ACCEPT' failed: iptables: Bad rule (does a matching rule exist in that chain?).
```

boris-brtan:
Some update when we implented 2 CIs and we found that without the part from documentation docker doesnt work due to firewall but if we apply the part from documentation it is OK so i think we have it OK and the logs are just logs. 
The main issue is that we do not have to restart `docker` service. If you have more info for this please provide us with it and we will test it, whether the restart of docker will be needed but it will cause only the problems to restart the whole docker service and its nodes.

PR #610, opened 2018-11-19
[shopsys] ProductOnCurrentDomainFacade: rename getPaginatedProductDetails* methods

PetrHeinz:
Created because of feedback from @Miroslav-Stopka.

I could've kept the original methods annotated with `@deprecated`, triggered an `E_USER_DEPRECATED` error and called the new method. This would avoid the BC break, but I'm not sure if we wanna do that during beta releases.

| Q             | A
| ------------- | ---
|Description, reason for the PR| `ProductDetail` class was removed along with the whole concept of `*Detail` classes during #276, the method names should reflect that for clarity.
|New feature| No 
|BC breaks| Yes 
|Fixes issues| ... 


boris-brtan:
any progress on this PR?

PetrHeinz:
I've rebased it on current master and rewritten the upgrade notes after #699 was merged.

Issue #591, opened 2018-11-12
DX - prepared classes for framework extending


vitek-rostislav:
### What is happening
When extending the framework behavior, there are some steps that could be prepared beforehand.
Example scenario - I need to find brands by their name so I have to:
- extend `BrandRpository` (i.e. create a new class)- add new method `findByName()`
- register my new class as an alias for the one from the framework in `services.yml`
- extend `BrandFacade` (i.e. create a new class) - add new method `findByName()`
- again, register my new class in `services.yml`

In https://github.com/shopsys/shopsys/pull/409, we prepared following empty extended classes in `project-base` for the most commonly extended entities:
- entities themselves
- data objects
- data object factories
- form extensions

**I believe we should prepare yet two more empty classes:**
**- facades**
**- repositories**

### Expected result

There are less necessary steps when a developer wants to extend the framework bahavior.

Miroslav-Stopka:
Hi @vitek-rostislav,
1) one of the nice features of the glassbox approach is that you dont have to maintain and a lot of unmodified files (old/core files). In my opinion, if you will prepare all facades and all repositories, this nice feature will be gone :/

2) register FACADE in services.yml
- are you sure this is necessary? I think that all facades are already registered.

3) register REPOSITORY in services.yml
- this is the only thing of your suggestions that I agree with

vitek-rostislav:
Hi @Miroslav-Stopka, thanks for the reply. 

ad 1) This is a good point - it should not be a problem for most of the time - you would just have simple empty classes in your project ready for your use if necessary. However, it is true that it might make the upgrading harder in some scenarios.

ad 2) oh, sure, you are right :+1: 

What do the others think? :slightly_smiling_face:

vitek-rostislav:
I would just add that this problem was reported by developers from Barton studio. 

For them, it seemed like too many steps for achieving such a simple modification.

PetrHeinz:
Creating an empty Facade and Repository layer for the basic entities (those we extended in the project-base) seems like a good idea to me.

Ad 2) and 3) The manual registering of the facades and repositories is necessary if you want to use your extended services instead of the original ones. Eg. when you create `MyProductFacade extends ProductFacade` and want it to be autowired to all services that require `ProductFacade` in the constructor. This requires a service alias to be created. The autoregistration would register both services independently.

I think that having those classes prepared in the project-base would serve as an explanation of how Shopsys Framework is intended to be used.

[...]

PR #616, opened 2018-11-20
[project-base] services.yml: automatic registration of classes with suffix "Repository" in namespace ShopBundle\Model\

Miroslav-Stopka:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This modification is needed on every implemented project with at least one repository class in ShopBundle. Mentioned in #591 as well.
|New feature| No
|BC breaks| No
|Fixes issues| ...


PetrHeinz:
In the `FrameworkBundle` we autoregister all classes in the Model namespace (also Component and DataFixture) with these suffixes: `Calculation`, `Facade`, `Factory`, `Generator`, `Handler`, `InlineEdit`, `Listener`, `Loader`, `Mapper`, `Parser`, `Provider`, `Recalculator`, `Registry`, `Repository`, `Resolver`, `Service`, `Scheduler`, `Subscriber` and `Transformer`.

Are you sure that `Repository` is the only one we should add?

Miroslav-Stopka:
1) This is the only one I know is needed for sure. 

2) Truth is Im not happy with this random group of suffixes - Calculation, Facade, Factory, Generator, Handler ... Why? For example, I created new Loaded in ShopBundle and this one has not registered automatically. So I ve searched where is registered DataFixtureLoader in FrameworkBundle - found nothing. And then, after a longer period, I have found out that there exists some **unspecified**  group of classes, that is registered automatically because of suffixes :/

PetrHeinz:
My point is that we shouldn't add these auto-discovered types of classes on one-by-one basis because every change generates an upgrade note - some work for the users to do.

We should probably make a list of all suffixes we want to set up the auto-discovery for, implement it and document it.

Maybe we should auto-discover `Facade`, `Factory`, `Repository` and `Service` because those class types are covered in [Basics About Model Architecture](https://github.com/shopsys/shopsys/blob/v7.0.0-beta2/docs/introduction/basics-about-model-architecture.md) (the article could be extended with a note about the auto-discovery).

vitek-rostislav:
I do agree with your last comment, @PetrHeinz, however, should we really block this PR from merging because of that? :slightly_smiling_face:

PetrHeinz:
@vitek-rostislav: Probably not.

vitek-rostislav:
Hi @Miroslav-Stopka, could you please rebase your branch onto current master?

vitek-rostislav:
Hi @Miroslav-Stopka, I have resolved the conflicts using Github web UI (it created the new commit that merged the master branch into your branch.) Now I will wait for tests on CI and validation with @LukasHeinz. If everything is ok, I will merge it using "Squash and merge" :wink:

Miroslav-Stopka:
Hi @vitek-rostislav, @PetrHeinz,

thank you both of you :)

PR #618, opened 2018-11-20
maintenance page fix

### LINKED ISSUE #167: Broken styling of the maintenance page
The maintenance page has been broken since the execution of some redesigned graphics (there is an immense 6000x4000px image).

Note: on / off maintenance - phin maintenance-on and phing maintenance-off

tomshopsys:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Set responsive visual of maintenance page
|New feature| No 
|BC breaks| No 
|Fixes issues| Broken styling of the maintenance page closes #167 

maintenance_page_fixed


project-base/app/maintenance.html
```diff
@@ -39,6 +37,129 @@
 <body>
 <div class="web">
     <h1>Site Maintenance</h1>
+    <p>
+      <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
PetrHeinz: The XML declaration should probably not be included when inlining the svg.

> If the XML declaration is included, it must be situated at the first position of the first line in the XML documentwell-formedness constraint. (Source: https://xmlwriter.net/xml_guide/xml_declaration.shtml)

Inlining SVGs into Twig templates should be avoided, but here it makes sense, as the `maintenance.html` should be without external dependencies.

Furthermore, is it necessary to include the `<p>` tag? It's not a paragraph.

project-base/app/maintenance.html
```diff
@@ -39,6 +37,129 @@
...
+      <!-- Generator: Adobe Illustrator 16.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
boris-brtan: this comment could be also removed

boris-brtan:
commit will be squashed

project-base/app/maintenance.html
tomshopsys: Yes, thats true, it is on the top of svg file, but when it is inline included has to be removed. Fixed.

PR #619, opened 2018-11-20
[framework] Simplify adding new form groups

### LINKED ISSUE #147: Simplification of .form-group separation
When using multiple .form-group elements we always had to add separators manually via .wrap-border
This is tiresome, error-prone and leads to complicated ifs when .form-groups are added dynamically (see plugins on product detail)
Some pages even had extra divs with .wrap-border (see order detail)

I have tried to simplify it using CSS adjacent sibling combinator in **ph-form-group-separation**, please check it out and:

- validate that changes to group.less make sense
- update CHANGELOG.md, UPGRADE.md and CSS documentation
- fix excessive gap after Images section on product detail (issue is also in master, separate commit)
- manually check as much pages in admin to see that forms are OK (especially if they contain mix of .form-groups and other elements)
- break changes Order/edit.html.twig into separate commits to make the changes clear
- check that .wrap-border makes sense as it creates the separation AFTER the element, not BEFORE it - is that best practice?

topolanekmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Created simplified method for adding new form groups. By this pull request there won't be needed checking for last form group by condition in twig template no more.
|New feature| No
|BC breaks| Yes
|Fixes issues| closes #147 


packages/framework/src/Resources/views/Admin/Content/Pricing/Groups/pricingGroupSettings.html.twig
```diff
@@ -1,14 +1,16 @@
...
-        {% endblock %}
-    {% endembed %}
-{{ form_end(form) }}
\ No newline at end of file
+        {% embed '@ShopsysFramework/Admin/Inline/FixedBar/fixedBar.html.twig' %}
+            {% block fixed_bar_content %}
+                {{ form_widget(form.save, { label: 'Save changes'|trans}) }}
+            {% endblock %}
+        {% endembed %}
+    </div>
```
boris-brtan: is it important to have the div after embed ? Wouldn't be better to move it before `{% embed` block ?

packages/framework/src/Resources/views/Admin/Form/productParameterValue.html.twig
```diff
@@ -25,7 +25,7 @@
 
 {% block product_parameter_value_widget %}
     {% import _self as self %}
-    <div class="js-form-group wrap-border form-group">
+    <div class="js-form-group form-group">
```
boris-brtan: here was removed `wrap-border` but no `border-divider` div element was added, is this intentional?

packages/framework/src/Resources/styles/admin/components/wrap/divider.less
```diff
@@ -0,0 +1,17 @@
+@wrap-border-gap: @item-gap;
+@wrap-border-inner-gap-small: 15px;
+
+.wrap-divider {
+    padding-bottom: @wrap-border-gap;
+    margin-bottom: @wrap-border-gap;
+
+    border-bottom: 1px solid @color-border-dark;
+
+    &:last-child {
```
boris-brtan: this could be changed into `last-of-type` so only div elements would be considered. 
example:
url: /admin/category/edit/2
in the form is last element hidden input tag and the last visible element shows the bottom border
image

boris-brtan:
TESTS:
*missing bottom border* 
- /admin/customer/list/
- /admin/vat/list/
- /admin/newsletter/list/
maybe can be solved with something like
```
/* .wrap-divider ~ .wrap-divider:last-of-type { */
.wrap-divider + .wrap-divider:last-of-type {
    margin-bottom: 0;
    border-bottom: none;
}
```
or 
```
.wrap-divider:last-of-type {
    margin-bottom: 0;
    border-bottom: none;
}
.wrap-divider:first-of-type {
    margin-bottom: 25px;
    border-bottom: 1px solid grey;
}
```

*some components could be wrapped into wrap-divider*
- /admin/order/edit/4 - order items
- /admin/customer/edit/2 - customer orders

*other problems*
- /admin/advert/edit/1 - there seems to be redundant class .wrap-border
- `/personal-data/setting/` `/admin/mail/template` - could be solved by renaming `div.window-fixed-bar` into some different element like `section` or `footer` or in different forms like `/admin/advert/edit/1` this fixed block is located before the form
- /admin/customer-communication/order-submitted/ empty space could be solved with the use of solution from mail templates

*after the changes there could be problem with*
- /admin/cookies/setting/ maybe we could somehow remove wrap-divider

boris-brtan:
TESTS: OK
build for kubernetes is working, kustomize is contained in the image via shell

boris-brtan:
Hi, the last only thing `/admin/administrator/edit/2`

PR #620, opened 2018-11-20
[shopsys] generate error pages in restart_kubernetes.sh

PetrHeinz:
- the error pages are not a part of neither the built Docker images nor build-demo-dev-quick target
- generation takes just 2-4 s so it doesn't delay the task

| Q             | A
| ------------- | ---
|Description, reason for the PR| Application does not handle errors correctly when restarted on CI using .ci/restart_kubernetes.sh
|New feature| No 
|BC breaks| No 
|Fixes issues|  

### DIFF
--- a/.ci/restart_kubernetes.sh
+++ b/.ci/restart_kubernetes.sh
@@ -25,4 +25,4 @@ kubectl rollout status --namespace=${JOB_NAME} deployment/microservice-product-s
 PHP_FPM_POD=$(kubectl get pods -n ${JOB_NAME} -l app=webserver-php-fpm -o=jsonpath='{.items[0].metadata.name}')
 
 # Run phing build in the pod
-kubectl exec --namespace=${JOB_NAME} ${PHP_FPM_POD} ./phing db-create test-db-create build-demo-dev-quick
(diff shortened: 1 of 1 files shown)


.ci/restart_kubernetes.sh
```diff
@@ -25,4 +25,4 @@ kubectl rollout status --namespace=${JOB_NAME} deployment/microservice-product-s
 PHP_FPM_POD=$(kubectl get pods -n ${JOB_NAME} -l app=webserver-php-fpm -o=jsonpath='{.items[0].metadata.name}')
 
 # Run phing build in the pod
-kubectl exec --namespace=${JOB_NAME} ${PHP_FPM_POD} ./phing db-create test-db-create build-demo-dev-quick
+kubectl exec --namespace=${JOB_NAME} ${PHP_FPM_POD} ./phing db-create test-db-create error-pages-generate build-demo-dev-quick
```
boris-brtan:
```suggestion
kubectl exec --namespace=${JOB_NAME} ${PHP_FPM_POD} ./phing db-create test-db-create build-demo-dev-quick error-pages-generate 
```
there is problem with empty database after task delete namespaces is executed

PetrHeinz: That makes sense :+1:

PR #626, opened 2018-11-23
[shopsys] UPGRADE.md: diff links point to shopsys/project-base

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR|  it makes more sense to link the commits in the split repository as they contain only the relevant changes, maybe we should do something like this automatically
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 


PetrHeinz:
@boris-brtan: eg. we could export all commits in project-base that were a part of some PR in monorepo during release, so they can be added as a link to either CHANGELOG.md or UPGRADE.md
Or, we could automatically replace all links to specific commits in shopsys/project-base section of UPGRADE.md with their split project-base counterparts.

This could be done hopefully using GitHub APIs, as the commit metadata (datetime & commit message) are identical in the split commits.

But first, we should find out what kind of information will help our users the most during the upgrade. Then we can automate the process of providing the info.

PR #629, opened 2018-11-26
remove symfony/flex from microservices

### LINKED ISSUE #450: Repeated Symfony Flex installation in microservices
### What is happening

`symfony.lock` is git-ignored in our microservices (along with `composer.lock`). The original motivation stemmed from our work with shopsys/project-base and `composer.lock` - we wanted to keep the dependencies up-to-date without manual updates.

Missing `symfony.lock` leads to repeated Symfony Flex installation when `composer installed` is executed for the first time, and it creates some uncommited changes (eg. adds some `.gitignore` files).

We should definetely reconsider the way we work with lock files or fix this situation some other way.

### Expected result

User has no uncommited changes after installation of microservices.

vitek-rostislav:
- we are not developing them at the moment so the flex is not needed
- prevents microservices from repeated Flex installation, solves #450

| Q             | A
| ------------- | ---
|Description, reason for the PR| `microservice-product-search-export` - [build on Travis](https://travis-ci.org/shopsys/microservice-product-search-export/jobs/455425565) fails - as a result to missing (gitignored) `/symofny.lock` file, Symfony flex creates `.tests/bootstrap.php` file that does not pass our coding standards. `microservice-product-search` might have the same problem but the package is not added to Travis.
|New feature| No 
|BC breaks| No 
|Fixes issues| closes #450 

### DIFF
--- a/microservices/product-search-export/composer.json
+++ b/microservices/product-search-export/composer.json
@@ -8,7 +8,6 @@
         "ext-iconv": "*",
         "elasticsearch/elasticsearch": "^6.0.1",
         "symfony/console": "^4.1",
-        "symfony/flex": "^1.0",
         "symfony/framework-bundle": "^4.1",
         "symfony/lts": "^4@dev",
         "symfony/monolog-bundle": "^3.3",
--- a/microservices/product-search/composer.json
+++ b/microservices/product-search/composer.json
@@ -8,7 +8,6 @@
         "ext-iconv": "*",
         "elasticsearch/elasticsearch": "^6.0.1",
         "symfony/console": "^4.1",
-        "symfony/flex": "^1.0",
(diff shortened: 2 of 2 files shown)


boris-brtan:
Now there is no need for fix, thus automatization of force build master branch on travis would be big plus.

vitek-rostislav:
Removal of `symfony/flex` is not as easy as it seemed - there are other parts of app dependent on that, e.g. composer scripts using `symfony-cmd`, see https://travis-ci.org/shopsys/microservice-product-search-export/jobs/459693097. I am closing this PR now as the master build on Travis is ok at the moment. Removing flex from microservices might be a possible solution of #450 but it might be solved some other way as well.

Issue #523, opened 2018-10-12
Documentation about Elasticsearch is outdated


vitek-rostislav:
### What is happening

https://github.com/shopsys/shopsys/blob/master/docs/introduction/product-search-via-elasticsearch.md  is outdated:
- it describes `shopsys.framework.elasticsearch_sources_dir` parameter but it does not exist anymore
- phing targets are not renamed, should be:
        - `elasticsearch-indexes-create` -> `microservice-product-search-create-structure`
        - `elasticsearch-indexes-delete` -> `microservice-product-search-delete-structure`
        - `elasticsearch-indexes-recreate` -> `microservice-product-search-recreate-structure`
        - `elasticsearch-products-export` -> `microservice-product-search-export-products`
- `ElasticsearchExportCronModule` does not exist anymore
- `ElasticsearchProductRepository` does not exist anymore
- `ElasticsearchProductDataConverter` does not exist anymore
- ???

### Expected result
The document should be up-to-date with the current state of the application

vitek-rostislav:
Partially solved in #504

vitek-rostislav:
solved in #793

PR #504, opened 2018-09-26
[shopsys] Update docs with new phing targets for elasticsearch

shastik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| phing commands has changed in https://github.com/shopsys/shopsys/commit/8fdf0e31d43e1f7aee05ff8e1edb0d07914ec37f#diff-e22ff16d006c03464bffffa8462c123a, but the documentation is still old
|New feature| No 
|BC breaks| No 
|Fixes issues| ...
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes/No

PR #637, opened 2018-11-28
[microservice-product-search-export] fixed search in EAN, partno, catnum

### LINKED ISSUE #398: Improvement of FE product search results
### What is happening
When I search "8800" I get as result for example product with EAN 8845781243207. It is because it fing match in n-gram of search entry and n-gram of atribute value. It is wierd.

Same with name - when I look for "apple" it also look into "appliance" (because of the same n-gram of both expressions ("app")

### Expected result
When I search "8800" I should get only results with EAN , partno or catalogue number starting with searched entry e.g. "8800123" or "880000258". Suggested change - do not create n-gram frem search entry but only on product attributes. 

Generaly - the functionality of search should match the documentation - https://github.com/shopsys/shopsys/blob/master/docs/introduction/product-search-via-elasticsearch.md

simara-svatopluk:
You'll need to recreate the structure in Elasticsearch and export the products for the change to take place in production environment (otherwise the issue described in #398 will still occur):
```sh
# execute it in the php-fpm container when using Docker
php phing microservice-product-search-recreate-structure \
          microservice-product-search-export-products
```

| Q             | A
| ------------- | ---
|Description, reason for the PR| #398
|New feature| No 
|BC breaks| No 
|Fixes issues| closes #398

### DIFF
--- a/docs/introduction/product-search-via-elasticsearch.md
+++ b/docs/introduction/product-search-via-elasticsearch.md
@@ -53,9 +53,9 @@ But if we simplify, we can say that the search term is searched in attributes an
 * partno - exact match
 * name - match in first couple of letters of any word
 * name - match in first couple of letters of any word ignoring diacritics
-* ean - match in first couple of letters
-* catnum - match in first couple of letters
-* partno - match in first couple of letters
+* ean - exact match in first characters
+* catnum - exact match in first characters
+* partno - exact match in first characters
 * short description - match anywhere
 * description - match anywhere
 
--- a/microservices/product-search-export/src/Resources/definition/1.json
+++ b/microservices/product-search-export/src/Resources/definition/1.json
@@ -113,7 +113,8 @@
           "fields": {
             "edge_ngram": {
               "type": "text",
-              "analyzer": "edge_ngram_unanalyzed"
+              "analyzer": "edge_ngram_unanalyzed",
+              "search_analyzer": "keyword"
             }
           }
         },
@@ -123,7 +124,8 @@
           "fields": {
             "edge_ngram": {
               "type": "text",
-              "analyzer": "edge_ngram_unanalyzed"
+              "analyzer": "edge_ngram_unanalyzed",
+              "search_analyzer": "keyword"
             }
           }
         },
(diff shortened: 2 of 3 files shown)


LukasHeinz:
Thank you for fast PR.

simara-svatopluk:
@LukasHeinz I tried to improve the documentation a bit. But I have a feeling that I can explain searching fully - that would take at least page or more, or I can explain searching inaccurately - that is current state.

boris-brtan:
hello, can you check wheter this part is now OK or not since you are updating docs? #504

simara-svatopluk:
@boris-brtan I'm not sure what do you want to know, hopefully I'll answer. #504 is still relevant, current documentation about phing targets is wrong.

PetrHeinz:
@simara-svatopluk: The documentation on Elasticsearch is not up-to-date since the indtroduction of microservices, see #523. It is unrelated to this PR imo.

I've closed the PR you've linked, as we sadly cannot merge it for licencing reasons.

simara-svatopluk:
There are always multiple ways to do almost everything in Elasticsearch. My thinking was - there is an issue that have to be quickly solved. What is the behavior we want to achieve? Do I have an Idea? Yes, and I have a quick solution.
When we go further, can you easily combine suggesting query with multi match? I'm not sure and I think you cannot use suggestion in multi match. It would probably mean that we will have to do changes in querying... and so on.
In the end, I don't want to use competition suggester for this fix.

As I said, I tried something in documentation, "exact match in first characters" seems fine... fixing

Upgrade notes... I'm not sure if it belongs here. Is this change back compatible? If you don't recreate your indexes, nothing changes, so it is. If you recreate your indexes, searching works differently, but code works the same, so it is not BC break in my opinion.
After thinking about BC, I would say no, I don't want to mention it in upgrade notes.

simara-svatopluk:
And final note, you can also use [prefix query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html). We have 3 options to solve prefix searching, I decided for the shortest one - no need to change querying. And I'm pretty sure that Elasticsearch will provide 4th way to search in prefixes. But what is more important I don't find value in discussing what way is better when since this is just a fix.

PR #648, opened 2018-12-03
WIP [framework] JS validation of form collections

### LINKED ISSUE #213: JS validation of the uniqueness of the parameters on the product card does not work
If I select 2 same parameters, the JS validation will show you "Each parameter can be used only once" (UniqueProductParameters)

In PHP UniqueProductParametersValidator is only dummy implementation, it relies on JS. PHP validation will not happen at all in this case because
ProductParameterValueToProductParameterValuesLocalizedTransformer :: transform () drops to the exception ("Duplicate parameter").

Now, therefore, only the Transformer Fault message is displayed: "This value is not valid."

JS validation does not work because storesysShopBundleComponentConstraintsUniqueProductParameters.validate will always be a blank field value.

image

I was looking at history when the functionality was broken:

now (ad03bf1 407f625fce0a5bf293941f8efb0cec8c4 27.9.2017) it does not work
before installing ESLint (069aff5 d6ed198d10e71131c3cf4cc0fb58448da 5.9.2017) it did not work ...
and it did not work until 2 months ago (4d70b8 c9fb1b73f5646849467e128c8cc2885d89 1.8.2017)

before updating to Symfony 3 (1bf11d30 b4bc238f797b44a1eac02184ac1b0f06 24.2.2017) The JS Validator of Parameters Unique Works
after updating to Symphony 3 still works (5c91ff22 d92716031bd17af110d108b0a4fd39db 30.3.2017)

The breakdown occurred sometime in April, May, June or July (I do not see it as a very specific time window ...)

In addition, it seems to have never worked well - fail only if the value of the already inserted parameter is modified (ie after adding a new duplicate parameter using the JS buttons never infinitely)

### LINKED ISSUE #268: Admin: Product edit: unclear message for non-unique parameters
### What is happening - steps to reproduce
1. Go to admin edit page of a product with ID 76
2. Add parameter (keep the default "Screen size") and fill in it's translations
3. Submit form

You see that form was submitted but there is an error flash message with the text "Please check the correctness of all data filled.". However, when you scroll down the page, there is no error highlighted so you have no clue what is wrong. Then, when you try to submit the form once more, you get a JS validation error that tells you that your parameters are not unique. (see https://prnt.sc/ouj0k3)

### Expected result
JS validation for new parameters is performed right away.

PetrHeinz:
Our implementation of JS constraint validators has been broken ever since moving Constraint classes from Components to Form namespace (9f96cd22ead5a091485119981b1cc905d3a86ff9, `v7.0.0-alpha3`).

First two commits fix just that. I'd like to also fix the behavior of JS validation collection before it's merged, as it will require a lot of manual testing.

---

`ProductParameterValueType` renders its errors rather ugly when it doesn't specifically render them inside the Twig theme with `{{ form_errors(form) }}`:
image
Now:
image
*Note: There are more form types with this problem (eg. the pure `CollectionType`). Maybe a systematic solution should be implemented instead.*

---

I had created a pure (not extended) `CollectionType` for testing, JS validation works nicely there.
This shows that the main problem with JS validation is that we extend the form type with our own types (eg. `ProductParameterValueType`). This is a problem for `FpJsFormValidator.js`, which has collection-specific behavior (which works rather nicely, as shown by the prototype), but it's inside a condition `'Symfony\\Component\\Form\\Extension\\Core\\Type\\CollectionType' == element.type` which doesn't work for the extended types.

image

*Note: At the moment, it is quite hard to create a dynamic form collections. See fbe468d for details.*

---

At the very least for this PR to be merged, JS validation has to be either fixed for `CollectionType` descendants (i.e. fix all the conditions for `element.type` to check whether it extends the `CollectionType` as well), or we can remove our extended collection types (This might be even harder - how would we enforce the fields to be rendered as they are rendered now? Parameters, order items and friendly URLs are all rendered quite differently).

Then, the JS validation has to be tested thoroughly.

---

In my opinion, our work with forms should be documented better and JS validation should be automatically tested.

| Q             | A
| ------------- | ---
|Description, reason for the PR| JS validation is problematic in dynamic form collections
|New feature| No 
|BC breaks| No 
|Fixes issues| fixes #213, fixes #268 (maybe more) 


PetrHeinz:
Portion of this was merged in #829

vitek-rostislav:
Stalebot, do not close this, please!

vitek-rostislav:
Stalebot, do not close this, please!

PetrHeinz:
@vitek-rostislav We could maybe move 69b962740047a3026767efe99471403b91618151 into a separate PR (it solves the problem that `ProductParameterValueType` renders its errors rather ugly when it doesn't specifically render them inside the Twig theme with `{{ form_errors(form) }}`, see description above, should be targeted to `7.3`, probably). This would be pretty simple - it only needs retesting and before-after screenshot after the rebase.

Then the only thing left here will be the prototype of a collection form rendered by default in fbe468ddf6b23d8d71379cd9d50c1884ba887e90 and the WIP fix of the behavior of collection validation using `Shopsys.validation.isCollectionType`. This prototype can be rebased when we will finally want to tackle the issues described in the description above...

Maybe we could rebase the branch and groom the description a bit so it's not so scary 🙂

PR #596, opened 2018-11-13
[project-base] Trusted proxies are now configurable in parameters.yml file

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| To be able to configure trusted proxies used for production in parameters.yml file
|New feature| No
|BC breaks| No
|Fixes issues| ...
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #660, opened 2018-12-07
[project-base] Bootstrap: do not prematurely boot Kernel when running in console

### LINKED ISSUE #658: Unwanted logs in console during bin/console command
### What is happening

there are some strange unwanted logs and it seems that it is happening after kernel boot was implemented
https://github.com/shopsys/shopsys/blame/master/project-base/app/Bootstrap.php#L57-L57

### Expected result

To have no unwanted logs in console

PetrHeinz:
- the Kernel boot was done to build DIC so `%trusted_proxies%` parameter is available
- setting trusted proxies is not required when running in console, as there are no Requests involved
- this stops excessive logging in console environment

| Q             | A
| ------------- | ---
|Description, reason for the PR| console commands output excessive log messages after #596 was merged
|New feature| No 
|BC breaks| No, but it's a project-base change, so users will have to replicate it 
|Fixes issues| fixes #658  

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -59,7 +59,7 @@ There is a list of all the repositories maintained by monorepo, changes in log b
 ### [shopsys/project-base]
 - *(optional)* [#596 Trusted proxies are now configurable in parameters.yml file](https://github.com/shopsys/shopsys/pull/596)
     - for easier deployment to production, make the trusted proxies in `Shopsys\Boostrap` class loaded from DIC parameter `trusted_proxies` instead of being hard-coded
-    - in your `Shopsys\Boostrap`, move the `Request::setTrustedProxies(...)` call along with `Kernel::boot()` so it's not run in console environment, like in [the PR #660 diff](https://github.com/shopsys/shopsys/pull/660/files), otherwise console commands will trigger excessive logging
+    - in your `Shopsys\Boostrap`, move the `Request::setTrustedProxies(...)` call along with `Kernel::boot()` so it's not run in console environment, like in [the PR #660](https://github.com/shopsys/shopsys/pull/660/files), otherwise console commands will trigger excessive logging
 - [#579 - ajaxMoreLoader is generalized](https://github.com/shopsys/shopsys/pull/579)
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -63,6 +63,8 @@ There is a list of all the repositories maintained by monorepo, changes in log b
     - new macro `loadMoreButton` is integrated into `@ShopsysShop/Front/Inline/Paginator/paginator.html.twig`, update files based on commit from [`ajaxMoreLoader is updated and generalized`](https://github.com/shopsys/shopsys/pull/579/files)
 - *(optional)* [#645 SVG icons in generated document](https://github.com/shopsys/shopsys/pull/645)
     - to display svg icons collection correctly in grunt generated document for all browsers please add `src/Shopsys/ShopBundle/Resources/views/Grunt/htmlDocumentTemplate.html` file and update `src/Shopsys/ShopBundle/Resources/views/Grunt/gruntfile.js.twig` based on changes in this pull request
+- [#660 - Bootstrap: do not prematurely boot Kernel when running in console](https://github.com/shopsys/shopsys/pull/660)
```
simara-svatopluk: I think this is optional - the project works without this change

simara-svatopluk:
By the way - great commit message! I'm going to test it

UPGRADE.md
PetrHeinz: I was thinking about it for a bit when I wrote that upgrade note - yes, it works, but #658 is tagged as a bug and I don't feel good about marking bugfixes optional. Are you sure we should mark this change as optional?

simara-svatopluk: Ok... so we have one idea why it should be optional and one why it shouldn't.
For a project that was created before #596 this is optional, for project that was created after #596, this is required.

So again - two ideas against two ideas... 

After some consideration, I don't know if there should be optional or not. If you think there shouldn't be, leave it as it is :+1:

PetrHeinz: You are right! It's tied to whether you have #596 merged in your repo. Fortunately it's merged only in `master`, so we can append note about this to the original optional upgrade note.

I've done that in 43fe6b307 - what do you think? I think it might be the best solution, as this fixes a bug introduced in the #596.

simara-svatopluk: Nice solution. I would prefer `[the PR #660]` over ~`[the PR #660 diff]`~

PetrHeinz: Done :slightly_smiling_face: Can I merge?

simara-svatopluk: I cannot approve it more :smiling_imp:

PR #669, opened 2018-12-12
admin: category editation: main URL is not recreated from the category name anymore

### LINKED ISSUE #668: [framework] Custom URL for Category
### What is happening
It is not possible to create own URL for a category

### What is happening
I am not able to create new URL for category and delete the one generated from category name.

Steps to reproduce:
1. Create a new URL for category and save.
2. Set the new URL as main and save.
3. Delete the original main URL that was generated from category name.

Once again, URL is generated from the category name and set as the main one.
Should be sufficient to remove one line from `CategoryFacade::edit()` where the new URL is generated.

### Expected result
Do not create an unwanted URL from category name.

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Problem with URL recreation described in #668. If the solution is accepted, the same approach should be used for products as there is a similar problem when editing a product.
|New feature| No 
|BC breaks| No 
|Fixes issues| closes #668 

### DIFF
--- a/packages/framework/src/Model/Category/CategoryFacade.php
+++ b/packages/framework/src/Model/Category/CategoryFacade.php
@@ -140,7 +140,6 @@ class CategoryFacade
         $category = $this->categoryService->edit($category, $categoryData, $rootCategory);
         $this->em->flush();
         $this->friendlyUrlFacade->saveUrlListFormData('front_product_list', $category->getId(), $categoryData->urls);
-        $this->friendlyUrlFacade->createFriendlyUrls('front_product_list', $category->getId(), $category->getNames());
         $this->imageFacade->uploadImage($category, $categoryData->image->uploadedFiles, null);
 
         $this->pluginCrudExtensionFacade->saveAllData('category', $category->getId(), $categoryData->pluginData);


PetrHeinz:
This would mean that the URL of a category won't change during a rename (unless changed manually, of course). I'd guess that it would be better if the URL changed during a rename when the URL was matching the original name (with any numeric suffix).

Example:
- Category *"Foo bar"* with default URL *"foo-bar/"* is renamed to *"Foo baz"*. A new default URL *"foo-baz/"* is created.
- Category *"Foo bar"* with default URL *"my-custom-url/"* is renamed to *"Foo baz"*. URLs are not changed.
- Category *"Foo bar"* with default URL *"foo-bar-2/"* (eg. there was already a product called *"Foo bar"*) is renamed to *"Foo baz"*. A new default URL *"foo-baz/"* is created.

The fact that names are multilocale and URLs are multidomain (and that their creation is a responsibility of the `FriendlyUrl` component) could complicate the implementation a bit.

Swejzi:
Not fixed in 9.0 :-( 🤦 

https://github.com/shopsys/shopsys/pull/1886

PR #679, opened 2018-12-13
[shopsys] webserver container starts after php-fpm is started

boris-brtan:
- all docker-compose files are fixed with depends_on attribute for webserver
- webserver doesn't fail on "host not found in upstream php-fpm:9000"

| Q             | A
| ------------- | ---
|Description, reason for the PR| during testing of special jobs there emerged error once per 30 minutes on 32 core CPU and 130GB server that webserver didn't start on error nginx: host not found in upstream php-fpm:9000
|New feature| No 
|BC breaks| No 
|Fixes issues| - 


boris-brtan:
UPDATE: now during upgrade of our [demoshop project](/shopsys/demoshop) we found out the webserver fails when it is running and --force-recreated so nginx is launched before php-fpm.

boris-brtan:
Hi, production docker-compose file was solved before because of the named volume https://github.com/shopsys/shopsys/commit/9908050173ffea48a6ed13ffdeaf791e20f82c09#diff-c9ce1b9ac4c90ebde520b156e5cc2e29 .

MattCzerner:
as @boris-brtan mentioned, production docker-compose was modified earlier.. So i guess its fine.

PR #680, opened 2018-12-13
[shopsys] docker infrastructure is simplified

boris-brtan:
- development infrastructure is proxied via webserver container and subdomains adminer, elastic, ...
- exposed ports are reduced to one for just webserver container
- urls for livereload tool are updated in twig templates

| Q             | A
| ------------- | ---
|Description, reason for the PR| we have many opened ports in dev environment and CI and ...., this could be managed by subdomains and proxying all opened ports via webserver container port so opened ports would reduce to just one per project
|New feature| No 
|BC breaks| No 
|Fixes issues|  


MattCzerner:
Hi @boris-brtan,
I will close this PR because of few reasons.
1) It is not simplified, instead of having defined port in docker-compose to connect to some service, you now need to know, that there exists some nginx conf file that defines these as a host names. 

It does not even simplify code in terms of code number, because you added 140 lines of code

2) This will not work out of the box as current solution does. You need to modify `/etc/hosts` file to make it run, for Linux it will sometimes work out of the box, but not for Mac and Windows.

PR #691, opened 2018-12-17
[shopsys] model service layer removal (phase 7)

simara-svatopluk:
Model service layer doesn't make sense too much. Should I put business logic into entities, services or both? We decided that business logic belongs to entities, we can get rid of model services and simplify the model.

This is a quite big change, so we split it into several phases.

| Q             | A
| ------------- | ---
|Description, reason for the PR| 
|New feature| No
|BC breaks| Yes
|Fixes issues|  


simara-svatopluk:
The only *service mentioned in code are:
* `InvalidServiceException` - not related to model services
* `$inlineEditService` variable and method names like this used in `Grid`, not related to model services, more to symfony services

PR #694, opened 2018-12-19
[shopsys] Updated PHP to version 7.3

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Support new version of PHP
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/microservices/product-search-export/.travis.yml
+++ b/microservices/product-search-export/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
     - 7.1
     - 7.2
+    - 7.3
 
 cache:
     directories:
--- a/composer.json
+++ b/composer.json
@@ -85,7 +85,7 @@
     "doctrine/migrations": "^1.6.0",
     "egeloen/ckeditor-bundle": "^4.0.6",
     "elasticsearch/elasticsearch": "^6.0.1",
-    "friendsofphp/php-cs-fixer": "^2.11.1",
+    "friendsofphp/php-cs-fixer": "^2.14.0",
     "fp/jsformvalidator-bundle": "^1.5.1",
     "fzaninotto/faker": "^1.7.1",
     "gedmo/doctrine-extensions": "^2.4.34",
--- a/docs/contributing/upgrading-monorepo.md
+++ b/docs/contributing/upgrading-monorepo.md
@@ -7,6 +7,8 @@ Typical upgrade sequence should be:
 * if you're experiencing some errors, you can always rebuild application and load demo data with `php phing build-demo-dev`
 
 ## [From v7.0.0-beta5 to Unreleased]
+- [#694 PHP 7.3 support](https://github.com/shopsys/shopsys/pull/694)
+    - rebuild your Docker images with `docker-compose up -d --build`
 
 ## [From v7.0.0-beta4 to v7.0.0-beta5]
 - [#651 It's possible to add index prefix to elastic search](https://github.com/shopsys/shopsys/pull/651)
--- a/docs/installation/native-installation.md
+++ b/docs/installation/native-installation.md
@@ -16,7 +16,7 @@ First of all, you need to install the following software on your system:
 ### Linux / MacOS
 * [GIT](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 * [PostgreSQL 9.5](https://wiki.postgresql.org/wiki/Detailed_installation_guides)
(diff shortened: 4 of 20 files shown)


grossmannmartin:
Currently, we have to install composer dependencies with `--ignore-platform-reqs`, because FriendsOfPHP/PHP-CS-Fixer does not support PHP 7.3. 
There is an [issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3697) about this topic.

build.xml
```diff
@@ -155,7 +156,10 @@
     </target>
 
     <target name="phpstan" depends="tests-acceptance-build" description="Performs static analysis of PHP files using PHPStan on all packages.">
-        <exec executable="${path.phpstan.executable}" logoutput="true" passthru="true" checkreturn="true">
+        <exec executable="${path.php.executable}" logoutput="true" passthru="true" checkreturn="true">
```
vitek-rostislav: as @boris-brtan pointed out, the proper solution might be using `--memory-limit` option, see phpstan/phpstan/issues/1148

vitek-rostislav:
>Currently, we have to install composer dependencies with `--ignore-platform-reqs`, because FriendsOfPHP/PHP-CS-Fixer does not support PHP 7.3.
There is an issue about this topic.

@TomasVotruba, what do you think about this, please?

It seems a little bit dangerous to me - when we want to say that we support php 7.3, we should not use tools that lack the support. And by ignoring the platform requirements, we could overlook some incompatibility problems, right?

Btw, should not the dependency on `symplify/easy-coding-standard` defined in `require-dev`?

build.xml
boris-brtan: nevertheless, i couldnt find it in options of our phpstan

TomasVotruba:
It's has nothing to do with code. php-cs-fixer runs on PHP 7.3 with no problem.  
It's ugly missuse of composer to force people to work, see https://stackoverflow.com/a/53727287/1348344

> Btw, should not the dependency on symplify/easy-coding-standard be defined in require-dev?

Why?

microservices/product-search-export/.travis.yml
```diff
@@ -3,6 +3,7 @@ language: php
 php:
     - 7.1
     - 7.2
+    - 7.3
```
boris-brtan: seems like in all travis scripts there is need to put composer install parameter `--ignore-platform-reqs`
see the splitted versions results from travis https://travis-ci.org/shopsys/microservice-product-search-export
i am also no fan of ignoring platform requirements, if the problem is with only one package it would be best to solve problem with that `friendsofphp/php-cs-fixer` rather than to use this kind of param.

boris-brtan: but is see in their requirements this bug
https://github.com/boris-brtan/PHP-CS-Fixer/blob/8e146bcca71112075ea9bc96791bf975b4873373/composer.json#L17

vitek-rostislav:
Hi @TomasVotruba, thanx for your reply :+1:
 
>> btw, should not the dependency on symplify/easy-coding-standard be defined in require-dev?

> What line do you talk about? Why?

My bad...I was thinking about `shopsys/coding-standards` which is in `require-dev` indeed. The `symplify/easy-coding-standard` is defined in `require` section in monorepo only so everything is ok :slightly_smiling_face:

grossmannmartin:
I simplified this PR, so it covers only one thing - Upgrade to PHP 7.3.
We decided, that we don't want to install dependencies with `--ignore-platform-reqs` option, so I'm putting this on hold until https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3697 (PR https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4184) will be solved.

Memory limit of PhpStan or possible upgrade of it should be solved in an independent PR, thank you for your insights on this topic :+1:

keradus:
FYI @grossmannmartin : PHP CS Fixer v2.14 supports PHP 7.3

MattCzerner:
Seems like its done bois. https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3697#issuecomment-451654988

build.xml
grossmannmartin: Solved differently in #716

microservices/product-search-export/.travis.yml
grossmannmartin: Library updated to a version that has support for 7.3.

PR #702, opened 2018-12-21
[project-base] switched to Debian php-fpm image

MattCzerner:
- debian commands are easier to understand
- alpine cannot connect to MSSQL databases

| Q             | A
| ------------- | ---
|Description, reason for the PR| debian commands are easier to understand, alpine cannot connect to MSSQL databases
|New feature| No
|BC breaks| Yes - I will upgrade `UPGRADE.md` if we will decide to go on with debian
|Fixes issues| 

### DIFF
--- a/docs/introduction/shopsys-framework-on-docker.md
+++ b/docs/introduction/shopsys-framework-on-docker.md
@@ -65,7 +65,7 @@ kind of recipe by which final image is cooked.
 
 Dockerfile example command:
 ```dockerfile
-FROM php:7.2-fpm-alpine
+FROM php:7.2-fpm-stretch
 ```
 * The `FROM` instruction specifies the base image, from which you are building
 
@@ -76,3 +76,39 @@ The official list of Dockerfile commands can be found on [Dockerfile reference](
 
 #### 3.5 docker/nginx/nginx.conf
 [Nginx.conf](/project-base/docker/nginx/nginx.conf) file contains Nginx configuration for new webserver container.
+
+#### 3.6 Images Distribution
+While running Shopsys Framework on docker we needed to decide which distribution will be our images running on. We use 2 types of distributions for our images.
+
+* Debian
+* Alpine Linux
+
+Both have their advantages and drawbacks
+
+##### Debian
+###### Advantages
+Debian is one of the oldest distributions, so even users without Docker experience are able to modify these images because installing tools in this distributions is easy, common and very easy to find on internet if some problem occurs.
+
+Also, Debian supports big amount of Databases and tools used by developers and are easy to install.
+###### Drawbacks
+Debian image is pretty big, which is not ideal for pulling these images on Cloud Solutions because there is more resources used to pull these.
+
+##### Alpine
+###### Advantages
+Alpine Linux is really small and minimalistic which means that clean installation of Alpine linux takes 6MB of memory which is great for pulling these images on Cloud where resource used for pull matters.
+
+###### Disadvantages
+Dockerfiles based from Alpine are harder to modify because it is so slim that with every technology, database or tool you want to install, you also need to install all its dependencies, making it a bit hard to find right solutions.
+
+Alpine Linux also has longtime problem with connection to MSSQL databases, due to missing support by Microsoft drivers.
+
+We divide images used by us into 3 types.
+
+##### Not extended, only used (PostgreSQL, Redis, Elasticsearch)
+Those images are only used as it is, we are not extending them and we are only using them as they are. In this case we choose Alpine Distribution thanks to its size.
+
+##### Microservices
+Microservices should be small as possible, even we are extending php-fpm images used for them, we take advantage of small size and suppose that users of these images will not extend them and more likely will be implementing their own microservices.
+
+##### Application PHP-FPM
+In this case, we use Debian mainly because we suppose users to modify these images often(adding php-extensions, implementing connections to Databases). As Debian is much more easier to modify, we decided that it will be best for new users to start on Debian, and if they care about size, they can always rewrite their Dockerfiles to be use alpine image if needed. Also many of clients using Shopsys Framework often connects to MSSQL databases and we want to make it easy for them.
--- a/project-base/docker/php-fpm/Dockerfile
+++ b/project-base/docker/php-fpm/Dockerfile
@@ -1,38 +1,43 @@
-FROM php:7.2-fpm-alpine as base
+FROM php:7.2-fpm-stretch as base
 
 ARG project_root=.
 ENV REDIS_PREFIX=''
 
-# install git for computing diffs
-RUN apk add --update git
+# install required tools
+# git for computing diffs
+# wget for installation of other tools
+# gnupg and g++ for gd extension
+# locales for locale-gen command
+# apt-utils so package configuartion does not get delayed
+# unzip to ommit composer zip packages corruption
+# dialog for apt-get to be
+# git for computing diffs and for npm to download packages
+RUN apt-get update && apt-get install -y wget gnupg g++ locales unzip dialog apt-utils git && apt-get clean
+
+# Install NodeJS
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash
+RUN apt-get update && apt-get install -y nodejs && apt-get clean
 
 # install Composer
 COPY ${project_root}/docker/php-fpm/docker-install-composer /usr/local/bin/docker-install-composer
-RUN chmod +x /usr/local/bin/docker-install-composer && docker-install-composer
+
+RUN chmod +x /usr/local/bin/docker-install-composer && \
+    docker-install-composer
 
 # libpng-dev needed by "gd" extension
-# icu-dev needed by "intl" extension
-# postgresql-dev needed by "pgsql" extension
 # libzip-dev needed by "zip" extension
+# libicu-dev for intl extension
+# libpg-dev for connection to postgres database
 # autoconf needed by "redis" extension
-# freetype-dev needed by "gd" extension
-# libjpeg-turbo-dev needed by "gd" extension
-RUN apk add --update \
+RUN apt-get install -y \
     libpng-dev \
-    icu-dev \
-    postgresql-dev \
     libzip-dev \
-    autoconf \
-    freetype-dev \
-    libjpeg-turbo-dev
+    libicu-dev \
+    libpq-dev \
+    autoconf && \
+    apt-get clean
 
-# "zip" extension warns about deprecation if we do not use a system library
-RUN docker-php-ext-configure zip --with-libzip
-
-# "gd" extension needs to have specified jpeg and freetype dir for jpg/jpeg images support
-RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/
-
-# install necessary PHP extensions requested by Composer
+# install necessary tools for running application
 RUN docker-php-ext-install \
     bcmath \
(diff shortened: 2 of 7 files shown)


grossmannmartin:
Hi, I just checked pros and cons of changing image and here are some of my thoughts about it.
Official PHP images are only Debian or Alpine, so even if it's possible to use a different distribution and install PHP itself, I don't think it has much of added value. 
Only distribution I have in mind is RHEL, but I'm not sure of licensing policy, but it's corporate-tested and some type of customers may want to use it.
But again, I don't think it's worth it.

I've checked the performance of Debian image and I'm pleased that there is no performance degradation :+1:

We want to consider several points of view with images:
- image size - alpine is a clear winner with almost 200MB difference
- ease of editing and management of Dockerfile - to me it seems similar. Maybe different package management, but as we stay on official PHP image, there is no considerable difference.
- image build speed - more thoughts about this follows
- performance - no difference (maybe slightly better on Debian)
- access to other services (MSSQL, Oracle) - more follows in separate paragraph
- license terms - I don't have the necessary knowledge of the law for evaluation

**Speed build image**
One thing I like to consider is to create the image based on one of official images, so there is no need to build an image. Speed of a clean build will increase dramatically and also we cut down more of repetitive work. 
Also running SSFW in a EKS/ECS (Kubernetes cluster or container service in AWS) will be much easier. 
It will be little more work with pushing and maintaining an image, but it may worth it. 
But if we stay on building an image from Dockerfile, and we stay with official PHP images, there should not be a marginal difference. 
Installing PHP to other distribution manually (RedHat) surely increase build time.

**Access to services**
One of the demands I've read in an internal document is to allow to access different services (MSSQL, Oracle...) from SSFW. 
I don't think we want to support other databases as our main storage, so I skip this scenario.
Programmer may want to connect to a different database with, for example, inventory management. 
In a perfect world, no application should touch database of the others. They should exchange data through API. But we're not living in a perfect world.
I know there is a problem with MSSQL drivers under Alpine. Someone managed to get this working with PDO-ODBC and FreeTDS, so maybe this can be a way (https://breeto.id.au/2017/05/docker-alpine-linux-php-sql-server/). 
But where is the line? Do we want to support every possible database/service?


TLDR:
- I would like to see built SSFW PHP image on dockerhub. 
- I would not change Alpine for Debian
- If it has business value, I would consider two different images (light/pro)

MattCzerner:
As this problem was discussed personally I will now write the conclusion that we came to with @grossmannmartin.

We will want to adopt Debian php-fpm image. 

Size of the image does not matter too much because mainly tools and our application is the main problem here.

Application itself has 600MB of size and distribution adds just a little to it.

Debian 120MB
Alpine 6MB

That means that our application will be at 720MB in size, that is not so bad in comparison to other images, the avarage image size is between 200-600MB.

Debian also solves problem with licenses, where there is only last bad license with libpng-dev, which can be taken out of dockerfile and described in documentation how to get it into image if you want.

Debian is also much more easier to rewrite and add your own functionality, Debian is one of oldest distributions and even not docker users know how to install some tools using apt-get.

**Conclusion**
Lets adopt Debian docker image and focus more on how to reduce size of image. For example we can install grunt only in dev environment, saving image size in production environment, where size really matters. Also I would check on Distroless images, if it is possible to adpot somehow for php applications.

@PetrHeinz

docs/introduction/shopsys-framework-on-docker.md
```diff
@@ -76,3 +76,39 @@ The official list of Dockerfile commands can be found on [Dockerfile reference](
...
 #### 3.5 docker/nginx/nginx.conf
 [Nginx.conf](/project-base/docker/nginx/nginx.conf) file contains Nginx configuration for new webserver container.
+
+#### 3.6 Images Distribution
+While running Shopsys Framework on docker we needed to decide which distribution will be our images running on. We use 2 types of distributions for our images.
+
+* Debian
+* Alpine Linux
+
+Both has their advantages and drawbacks
```
grossmannmartin:
```suggestion
Both have their advantages and drawbacks
```

docs/introduction/shopsys-framework-on-docker.md
```diff
@@ -76,3 +76,39 @@ The official list of Dockerfile commands can be found on [Dockerfile reference](
...
+
+##### Debian
+###### Advantages
+Debian is one of the oldest distributions, so even users without Docker experience are able to modify these images, because installing tools in this distributions is easy, common and very easy to find on internet if some problem occurs.
```
grossmannmartin:
```suggestion
Debian is one of the oldest distributions, so even users without Docker experience are able to modify these images because installing tools in this distributions is easy, common and very easy to find on internet if some problem occurs.
```

docs/introduction/shopsys-framework-on-docker.md
```diff
@@ -76,3 +76,39 @@ The official list of Dockerfile commands can be found on [Dockerfile reference](
...
+Also, Debian supports big amount of Databases and tools used by developers and are easy to install.
+###### Drawbacks
+Debian image is pretty big, which is not ideal for pulling these images on Cloud Solutions because there is more resources used to pull these.
+
+##### Alpine
+###### Advantages
+Alpine Linux is really small and minimalistic which means that clean installation of Alpine linux takes 6MB of memory which is great for pulling these images on Cloud where resource used for pull matters.
+
+###### Disadvantages
+Dockerfiles based from Alpine are harder to modify, because it is so slim that with every technology, database or tool you want to install, you also need to install all its dependencies, making it a bit hard to find right solutions.
```
grossmannmartin:
```suggestion
Dockerfiles based from Alpine are harder to modify because it is so slim that with every technology, database or tool you want to install, you also need to install all its dependencies, making it a bit hard to find right solutions.
```

docs/introduction/shopsys-framework-on-docker.md
```diff
@@ -76,3 +76,39 @@ The official list of Dockerfile commands can be found on [Dockerfile reference](
...
+We divide images used by us into 3 types.
+
+##### Not extended, only used (PostgreSQL, Redis, Elasticsearch)
+Those images are only used as it is, we are not extending them and we are only using them as they are. In this case we choose Alpine Distribution thanks to its size.
+
+##### Microservices
+Microservices should be small as possible, even we are extending php-fpm images used for them, we take advantage of small size and suppose that users of these images will not extend them and more likely will be implementing their own microservices.
+
+##### Application PHP-FPM
+In this case we use Debian, mainly because we suppose users to modify these images often(adding php-extensions, implementing connections to Databases). As Debian is much more easier to modify, we decided that it will be best for new users to start on Debian, and if they care about size, they can always rewrite their Dockerfiles to be use alpine image if needed. Also many of clients using Shopsys Framework often connects to MSSQL databases and we want to make it easy for them.
```
grossmannmartin:
```suggestion
In this case, we use Debian, mainly because we suppose users to modify these images often (adding php-extensions, implementing connections to Databases). As Debian is much easier to modify, we decided that it will be best for new users to start on Debian, and if they care about size, they can always rewrite their Dockerfiles to use alpine image if needed. Also, many of clients using Shopsys Framework often connects to MSSQL databases and we want to make it easy for them.
```

project-base/docker/php-fpm/Dockerfile
```diff
@@ -42,33 +47,40 @@ RUN docker-php-ext-install \
...
-# install grunt-cli using npm to be able to run grunt watch
+# install grunt cli used by frontend developers for continuous generating of css files
 RUN npm install -g grunt-cli
 
-# install postgresql to allow execution of pg_dump for acceptance tests
-RUN apk add --update --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.7/main postgresql
+# install PostgreSQl client for dumping database
+RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
+    sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -sc)-pgdg main" > /etc/apt/sources.list.d/PostgreSQL.list' && \
+    apt-get update && apt-get install -y postgresql-10 postgresql-client-10
```
grossmannmartin:
```suggestion
    apt-get update && apt-get install -y postgresql-10 postgresql-client-10 && apt-get clean
```

project-base/docker/php-fpm/Dockerfile
```diff
@@ -84,10 +96,10 @@ FROM base as development
 USER root
 
 # allow overwriting UID and GID o the user "www-data" to help solve issues with permissions in mounted volumes
-# if the GID is already in use, we will assign GID 82 instead (82 is the standard uid/gid for "www-data" in Alpine)
+# if the GID is already in use, we will assign GID 33 instead (33 is the standard uid/gid for "www-data" in Debian)
 ARG www_data_uid
 ARG www_data_gid
-RUN if [ -n "$www_data_uid" ]; then deluser www-data && (addgroup -g $www_data_gid www-data || addgroup -g 82 www-data) && adduser -u $www_data_uid -D -S -G www-data www-data; fi;
+RUN if [ -n "$www_data_uid" ]; then deluser www-data && (addgroup --gid $www_data_gid www-data || addgroup --gid 33 www-data) && adduser --system --home /home/www-data --uid $www_data_uid --disabled-password --gid $www_data_gid www-data; fi;
```
grossmannmartin: When gid is in use and www-data group is created with default 33 gid, you still create user with gid from argument. On Mac, the default gid of user is 20 (staff), but this gid is already in use by debian group dialout. Resulting in user www-data with group dialout and group www-data is unused.

MattCzerner: If I understand it correctly, the solution should be to add user using usermod to an existing group by name, not by ID. Can you please test it? Im not entirely sure.

project-base/docker/php-fpm/Dockerfile
```diff
@@ -84,10 +96,10 @@ FROM base as development
 USER root
 
 # allow overwriting UID and GID o the user "www-data" to help solve issues with permissions in mounted volumes
-# if the GID is already in use, we will assign GID 82 instead (82 is the standard uid/gid for "www-data" in Alpine)
+# if the GID is already in use, we will assign GID 33 instead (33 is the standard uid/gid for "www-data" in Debian)
 ARG www_data_uid
 ARG www_data_gid
-RUN if [ -n "$www_data_uid" ]; then deluser www-data && (addgroup -g $www_data_gid www-data || addgroup -g 82 www-data) && adduser -u $www_data_uid -D -S -G www-data www-data; fi;
+RUN if [ -n "$www_data_uid" ]; then deluser www-data && (addgroup --gid $www_data_gid www-data || addgroup --gid 33 www-data) && adduser --system --home /home/www-data --uid $www_data_uid --disabled-password www-data && usermod -a -G www-data www-data; fi;
```
grossmannmartin: I tried it and it seems it's sufficient to just create the user with option `--group` (according to https://manpages.debian.org/jessie/adduser/adduser.8.en.html)
```suggestion
RUN if [ -n "$www_data_uid" ]; then deluser www-data && (addgroup --gid $www_data_gid www-data || addgroup --gid 33 www-data) && adduser --system --home /home/www-data --uid $www_data_uid --disabled-password --group www-data; fi;
```

project-base/docker/php-fpm/Dockerfile
grossmannmartin: I've tried and user has group www-data, along with somw weird ones. It should be enough to create user with `group` option and do not modify at all. 
Suggestion is in files.

PR #714, opened 2019-01-06
[shopsys] Google Cloud Storage Bucket

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Store files and images into Google Cloud Storage Bucket, so files are not destroyed by pod deletion and we move closer to scalable application
|New feature| Yes
|BC breaks| No
|Fixes issues| 


composer.json
```diff
@@ -130,7 +130,9 @@
     "tracy/tracy": "^2.4.13",
     "twig/extensions": "^1.5.1",
     "twig/twig": "^2.4.8",
-    "vasek-purchart/console-errors-bundle": "^1.0.1"
+    "vasek-purchart/console-errors-bundle": "^1.0.1",
+    "superbalist/flysystem-google-storage": "^7.1",
+    "google/cloud-storage": "^1.10"
```
PetrHeinz: These should be alphabetically sorted.

packages/framework/src/Component/Environment/EnvironmentFileSetting.php
```diff
@@ -8,10 +8,21 @@ class EnvironmentFileSetting
         EnvironmentType::DEVELOPMENT => 'DEVELOPMENT',
         EnvironmentType::PRODUCTION => 'PRODUCTION',
         EnvironmentType::TEST => 'TEST',
+        EnvironmentType::GOOGLE_CLOUD => 'GOOGLE_CLOUD',
```
PetrHeinz: I don't think an extra environment is a good solution to this. In my opinion, setting up the GCS bucket as a file system adapter should be orthogonal - i.e. it should be independent of the other settings. It shouldn't be impossible to run a dev, test or production environment with files stored in GCS.

Eg. an ENV variable GOOGLE_CLOUD_STORAGE_BUCKET_NAME could be introduced. If empty, the application would behave just as it behaves right now, otherwise it would use the GCS adapter. This should work in all three environments.

Can we please discuss the configuration options futher?

project-base/app/AppKernel.php
```diff
@@ -50,11 +50,15 @@ public function registerBundles()
...
-            $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
+            $developmentBundles = [
+                new Sensio\Bundle\DistributionBundle\SensioDistributionBundle(),
+                new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle(),
+                new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle(),
+                new Symfony\Bundle\DebugBundle\DebugBundle(),
+                new Symfony\Bundle\WebServerBundle\WebServerBundle(),
+            ];
+
+            $bundles = array_merge($bundles, $developmentBundles);
```
PetrHeinz: Not sure whether this is an improvement over the previous state. Why do you prefer `array_merge`?

project-base/app/config/parameters_google_cloud.yml.dist
```diff
@@ -0,0 +1,3 @@
+parameters:
+    google_cloud_project_id: ~
+    google_cloud_storage_bucket_name: ~
```
PetrHeinz: Config file templates (`parameters.yml.dist` or `parameters_<ENV>.yml.dist`) are mentioned in `composer.json` option `extra.incenteev-parameters`, so users are prompted to fill them out during `composer install`. Is it because it's an unusual use-case?

project-base/docker/php-fpm/Dockerfile
```diff
@@ -115,3 +115,15 @@ RUN composer install --optimize-autoloader --no-interaction --no-progress --dev
 RUN php phing composer-dev npm dirs-create test-dirs-create assets standards tests-static tests-acceptance-build
 
 RUN ./bin/console shopsys:environment:change prod
+
+########################################################################################################################
+
+FROM base as google_cloud
```
PetrHeinz: It seems to me that ideally the Docker image should be the same, only its configuration (via ENVs or configmaps) should change.

Alternatively, the GCS adapter could require a change in the codebase, similar to installing a plugin. This way we could have Shopsys Framework independent of `superbalist/flysystem-google-storage` and `google/cloud-storage` (which might be important when we support more cloud providers, now it's not necessary imo).

packages/framework/composer.json
```diff
@@ -51,6 +51,7 @@
         "fp/jsformvalidator-bundle": "^1.5.1",
         "fzaninotto/faker": "^1.7.1",
         "gedmo/doctrine-extensions": "^2.4.34",
+        "google/cloud-storage": "^1.10",
```
PetrHeinz: Nice catch, using `Google\Cloud\Storage` in our service definition means that we are directly dependent on this.

project-base/infrastructure/google-cloud/gke-cluster.tf
```diff
@@ -12,9 +12,10 @@ resource "google_container_cluster" "primary" {
   node_config {
     oauth_scopes = [
       "https://www.googleapis.com/auth/compute",
-      "https://www.googleapis.com/auth/devstorage.read_only",
+      "https://www.googleapis.com/auth/devstorage.full_control",
```
PetrHeinz: Is managing the permissions necessary as well? Souldn't `read_write` be enough?

project-base/infrastructure/google-cloud/gke-cluster.tf
```diff
@@ -12,9 +12,10 @@ resource "google_container_cluster" "primary" {
...
       "https://www.googleapis.com/auth/logging.write",
       "https://www.googleapis.com/auth/monitoring",
+      "https://www.googleapis.com/auth/cloud-platform",
```
PetrHeinz: So this is permitting the application to access the bucket without `service-account.json`? If so, maybe you can remove the `kubernetes_secret.gcs-service-account`, `google_service_account_key.gcs-service-account`, `google_service_account.gcs-service-account` and the output variable `gcs-service-account-json-key` from Terraform, as the secret is not used anywhere. I think it was prepared to be used to access the GCS. If not, I might have misunderstood the way deployed app is authorized.

project-base/docker/php-fpm/Dockerfile
boris-brtan: seems like dirs-create is not needed also for production so we can use production target with different `RUN ./bin/console shopsys:environment:change `

project-base/.ci/deploy-to-google-cloud.sh
```diff
@@ -6,12 +6,19 @@ echo ${DOCKER_PASSWORD} | docker login --username ${DOCKER_USERNAME} --password-
 # Create unique docker image tag with commit hash
 DOCKER_IMAGE_TAG=production-commit-${GIT_COMMIT}
 
+cp app/config/parameters_google_cloud.yml.dist app/config/parameters_google_cloud.yml
```
boris-brtan: Upgrade log is needed.

packages/framework/composer.json
boris-brtan: what about `project-base/composer.json` ???

project-base/infrastructure/google-cloud/stateful-services.tf
```diff
@@ -1,6 +1,7 @@
 resource "google_storage_bucket" "file-store" {
   name     = "${var.GOOGLE_CLOUD_STORAGE_BUCKET_NAME}"
   location = "EU"
+  force_destroy = true
```
boris-brtan: could we put in documentation that if someone would like to keep the data, it needs to be backed up ?

PetrHeinz:
Replaced by #730

PR #715, opened 2019-01-09
[framework] OrderItem is easily extendable

### LINKED ISSUE #621: Extending OrderPayment and OrderTransport Entity results in "It is not supported to define inheritance information on a mapped superclass" exception
I need to extend Model\Order\Item\OrderPayment and OrderTransport classes. I followed Adding new attribute article. Everything seem all right unless I try to submit an order. Then it throws It is not supported to define inheritance information on a mapped superclass ‘Shopsys\FrameworkBundle\Model\Order\Item\OrderItem’ exception.

### What is happening
I need to be able to set catnum attribute even for OrderTransport and OrderPayment classes due to synchronization with Pohoda (~ foreign key).

I extended both entities according to Adding new attribute article and added:
Shopsys\FrameworkBundle\Model\Order\Item\OrderPayment: Shopsys\ShopBundle\Model\Order\Item\OrderPayment

Shopsys\FrameworkBundle\Model\Order\Item\OrderTransport: Shopsys\ShopBundle\Model\Order\Item\OrderTransport

into parameters_common.yml

All is working as expected, eg. if I dump Order entity, it contains OrderPayment and OrderTransport objects of correct namespace (ShopBundle) within items and these objects are filled correctly. The problem arises when Doctrine tries to store data into DB; it throws

It is not supported to define inheritance information on a mapped superclass ‘Shopsys\FrameworkBundle\Model\Order\Item\OrderItem’

any hints? 

### Expected result

I would like the filled Order object to be stored into DB 🥇

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We need to make working with `OrderItem` easier for implementators and also to make it extendable
|New feature| Yes
|BC breaks| Yes
|Fixes issues| closes #621

# Resolution of implementation

We discussed 3 possible solutions removing inheritance.

## One class OrderItem

We remove all descendants and everything will be solved in one class based on its "type".

**Pros**
- easily extendable class
- less code
- easy access to all the data in listings
- everything stays in one "OrderItem" collection, with which I can work the same way as now
- easier form rendering in administration

**Cons**
- big class with mixed responsibilities

## Multiple classes with associations

Instead of inheritance, dependent classes will have an association with `OrderItem`, ie. OrderTransport will contain `private OrderItem`, `private Transport`.

**Pros**
- same classes as now with  the same responsibilities
- extending is possible
- clear responsibilities

**Cons**
- problematic data access in listing because data are stored in multiple entities/tables
- may be difficult to implement because `OrderTransport`, etc. will not inherit from the same parent

## Pull off OrderTransport and OrderPayment outside OrderItem

The main thought is to separate `OrderPayment` and `OrderTransport` to special fields of `Order`. `OrderItem` will be responsible only for products.

**Pros**
- clarified responsibilities
- easily extended class

**Cons**
- a different view to `OrderItem`, which brings big BC break for implementation teams
- new type (for example `PromoCode`) means, either way, need to add a type to `OrderItem` entity, so solution no. 1 will still need to be applied

**In the end, we decide to go with one class `OrderItem`, responsible for all the types.**


packages/framework/src/Model/Order/Item/OrderItemFactory.php
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @param string $name
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Price $price
+     * @param string $vatPercent
+     * @param int $quantity
+     * @param string|null $unitName
+     * @param string|null $catnum
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product|null $product
+     *
```
simara-svatopluk: @grossmannmartin I see here an empty line.... Did you see this style somewhere in the framework already or it is a novelty? It seems readable, but I want to keep our code consistent :-)

packages/framework/src/Model/Order/Item/OrderItemFacade.php
```diff
@@ -42,9 +42,9 @@ class OrderItemFacade
     protected $orderPriceCalculation;
 
     /**
-     * @var \Shopsys\FrameworkBundle\Model\Order\Item\OrderProductFactoryInterface
+     * @var \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemFactoryInterface
      */
-    protected $orderProductFactory;
+    private $orderItemFactory;
```
simara-svatopluk: @grossmannmartin We use `protected` for properties in Facades for easier extendability

packages/framework/src/Model/Order/Item/OrderItemFactory.php
grossmannmartin: You're right. I fixed it.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -45,6 +45,29 @@ There you can find links to upgrade notes for other versions too.
...
+        - replace usages of method `create()` with `createPayment()`
+    - replace usages of `OrderTransportFactoryInterface` with common `OrderItemFactoryInterface`
+        - replace usages of `OrderTransportFactory` with `OrderItemFactory`
+        - replace usages of method `create()` with `createTransport()`
+    - replace usages of `OrderPaymentDataFactoryInterface` with common `OrderItemDataFactoryInterface`
+        - replace usages of `OrderPaymentDataFactory` with common `OrderItemDataFactory`
+        - replaces usages of method `createFromOrderPayment()` with `createFromOrderItem()`
+    - replace usages of `OrderTransportDataFactoryInterface` with common `OrderItemDataFactoryInterface`
+        - replace usages of `OrderTransportDataFactory` with common `OrderItemDataFactory`
+        - replaces usages of method `createFromOrderTransport()` with `createFromOrderItem()`
```
PetrHeinz: I would suggest grouping all the above required steps together like this:

```
    - replace usages of `OrderProductFactoryInterface`, `OrderPaymentFactoryInterface` and `OrderTransportFactoryInterface` with common `OrderItemFactoryInterface`
        - replace usages of `OrderProductFactory`, `OrderPaymentFactory` and `OrderTransportFactory` with `OrderItemFactory`
        - replace usages of method `create()` with `createProduct()`, `createPayment()` or `createTransport()`, respectively
    - replace usages of `OrderPaymentDataFactoryInterface` and `OrderTransportDataFactoryInterface` with common `OrderItemDataFactoryInterface`
        - replace usages of `OrderPaymentDataFactory` and `OrderTransportDataFactory` with common `OrderItemDataFactory`
        - replaces usages of method `createFromOrderPayment()` and `createFromOrderTransport()` with `createFromOrderItem()`
```

It seems clearer to me and it's much shorter.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -45,6 +45,29 @@ There you can find links to upgrade notes for other versions too.
...
+    - following classes changed constructors, if you extend them, change them appropriately:
+        - `Order`
+        - `OrderDataFactory`
+        - `OrderItemFacade`
+        - `OrderFacade`
```
PetrHeinz: I can think of two neccessary steps that are not included in the upgrade notes:
- using the `isType<type>()` method instead of `instanceof`
- removal of the non-existing test cases from `EntityExtensionTest.php`

packages/framework/src/Model/Order/Item/OrderItemFactory.php
```diff
@@ -0,0 +1,121 @@
...
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\OrderProduct
+     */
+    public function createProduct(
```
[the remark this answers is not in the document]
simara-svatopluk: You are right... I didn't see it, thanks for noticing. Anyway we'll not fix it as it would take some time and You don't find it necessary

docs/upgrade/UPGRADE-unreleased.md
simara-svatopluk: I don't care, much... so why not :+1:

packages/framework/src/Model/Order/Item/OrderItemFactoryInterface.php
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @param string $name
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Price $price
+     * @param string $vatPercent
+     * @param int $quantity
+     * @param \Shopsys\FrameworkBundle\Model\Payment\Payment $payment
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\OrderPayment
+     */
+    public function createPayment(
```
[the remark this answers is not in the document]
simara-svatopluk: At first I was 100% for changing the interface to Data objects.

When I went through implementation, I realized, that `OrderItemFactory` is used only inside of `Order` and that it is used in a creative way for example without data https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Order/Order.php#L1128  for example in this particular case I would be surprised if I had to pass also `OrderItemDataFactory` into this method. 

All in all I'm not even sure if we want to change this interface... So I prefer to not doing it in this PR. Should I create u user story or issue for "decision what to do with not standard OrderItemFactoryInterface"?

simara-svatopluk:
* Ok, we'll not fix the history to pass standards in all commits
* squashing that two commits about removing factory is for me counter-intuitive. I like small commits during refactoring and they fits for me. I would rather keep them as they are :-)

packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+    public function testPaymentCannotBeSettedForWrongType(): void
+    {
+        $orderItem = $this->createOrderProduct();
+
+        $this->expectException(WrongItemTypeException::class);
+        $orderItem->setPayment($this->createPaymentMock());
+    }
+
+    public function testPaymentCannotBeGettedFromWrongType(): void
```
PetrHeinz: The past participle of `get` is `got` or `gotten`, not `getted` (`taken` can be used as well and is more common in such context).

packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+    public function testProductCannotBeSettedForWrongType(): void
+    {
+        $orderItem = $this->createOrderTransport();
+
+        $this->expectException(WrongItemTypeException::class);
+        $orderItem->setProduct($this->createProductMock());
+    }
+
+    public function testProductCannotBeGettedFromWrongType(): void
```
PetrHeinz: The past participle of `get` is `got` or `gotten`, not `getted` (`taken` can be used as well and is more common in such context).

packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+        $orderItemData = new OrderItemData();
+        $transport = $this->createTransportMock();
+        $orderItemData->transport = $transport;
+        $orderItem->edit($orderItemData);
+
+        $this->assertSame($transport, $orderItem->getTransport());
+    }
+
+    public function testPaymentCannotBeSettedForWrongType(): void
```
PetrHeinz: The past participle of `set` is `set`, not `setted`.

packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+        $orderItemData = new OrderItemData();
+        $payment = $this->createPaymentMock();
+        $orderItemData->payment = $payment;
+        $orderItem->edit($orderItemData);
+
+        $this->assertSame($payment, $orderItem->getPayment());
+    }
+
+    public function testProductCannotBeSettedForWrongType(): void
```
PetrHeinz: The past participle of `set` is `set`, not `setted`.

docs/introduction/framework-extensibility.md
```diff
@@ -7,6 +7,7 @@ as well as a list of customizations that are not (and will not be) possible at a
 ## What is achievable easily
 * [Extending an entity](/docs/wip_glassbox/entity-extension.md)
     * [Adding a new attribute](/docs/cookbook/adding-new-attribute-to-an-entity.md)
+    * *Beware the [OrderItem has some limitations](/docs/wip_glassbox/entity-extension.md#orderitem)*
```
TomasLudvik: I would replace beware with note like this:
`*Note: There are some limitations when extending OrderItem, for more ....*

docs/wip_glassbox/entity-extension.md
```diff
@@ -133,3 +133,18 @@ So in the project, we can change the factory to produce extended entity data ins
   * add your extended entity into `$entityExtensionMap` in the `setUp()` method
 
 *Tip: to see how it works in practice check out `\Tests\ShopBundle\Functional\EntityExtension\EntityExtensionTest` that tests end-to-end extensibility of `Product`, `Category` and `OrderItem`.*
+
+## Limitations
+
+### OrderItem
+
+`OrderItem` is a bit special entity, because it is not created from `OrderItemData`, it is created from different sources like from product itself.
```
TomasLudvik: There should not be a comma before because

docs/wip_glassbox/entity-extension.md
```diff
@@ -133,3 +133,18 @@ So in the project, we can change the factory to produce extended entity data ins
...
+All creations are done by [`Order`](/packages/framework/src/Model/Order/Order.php) entity only, where you can check that it really make sense to not create the `OrderItem` from a data object.
+
+If you need to extend the `OrderItem` by a new field, for example, an ID from an external system, you'll have to fill this field after the `Order` is created by a new public method on `Order` class.
+And then if you'll need to be able to edit this field from the administration, you'll have to override `edit` method of `Order` entity and solve setting this new field there.
+
+The other way of data from - from `OrderItem` to `OrderItemData` is standard.
```
TomasLudvik: there is word from two times

docs/wip_glassbox/entity-extension.md
```diff
@@ -133,3 +133,18 @@ So in the project, we can change the factory to produce extended entity data ins
...
+So if you extend `OrderItem` in a standard fashion, add setting a new field in extended `OrderItemDataFactory` then `OrderItemData` object will contain the correct value from the `OrderItem` object.
```
TomasLudvik: add ~~setting~~ a new field

docs/wip_glassbox/entity-extension.md
```diff
@@ -133,3 +133,18 @@ So in the project, we can change the factory to produce extended entity data ins
...
+
+Creating a new type of `OrderItem` is possible and doesn't cause problems because the new type is completely in your hands.
```
TomasLudvik: doesn't -> does not

project-base/src/Shopsys/ShopBundle/Model/Order/Item/OrderItem.php
```diff
@@ -0,0 +1,49 @@
...
+     * @param \Shopsys\ShopBundle\Model\Order\Order $order
+     * @param string $name
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Price $price
+     * @param string $vatPercent
+     * @param int $quantity
+     * @param string $type
+     * @param null|string $unitName
+     * @param null|string $catnum
+     */
+    public function __construct(
```
TomasLudvik: This is not necessary here

docs/wip_glassbox/entity-extension.md
grossmannmartin: Why? Do we have some rule to not use contractions? For me, it sounds more natural, but I'm eager to hear different opinions

project-base/src/Shopsys/ShopBundle/Model/Order/Item/OrderItem.php
grossmannmartin: Like a whole constructor?
True, but we have constructors like this in every other extended entity.

Or?

TomasLudvik: Oh, it is true, it is ok then

docs/wip_glassbox/entity-extension.md
TomasLudvik: It is more formal, it is only reason

docs/wip_glassbox/entity-extension.md
TomasLudvik: so can you update it appropriately?

Issue #686, opened 2018-12-17
Images for the main variant are not displayed


DavidOstrozlik:
### What is happening
If you create variant and select a product with images as main variant, resulting main variant has images that are not visible or the creating results in error 500.

eg.
- Select product "22" Sencor SLE 22F46DM4 HELLO KITTY " as main variant -> error 500
- Add images to some product and select it as main variant -> the images are not displayed (maybe because of route to the image "eg. http://127.0.0.1:8000/content/images/product/original/106.")

### Expected result
Main variant should have visible images and the creation of variant should not end up with error 500

boris-brtan:
still confused but it seems to me that there are 2 different initialisations of flysystem configs during `ImageFacade::copyImages` and `DoctrineListener::postPersist---fileUpload->postFlushEntity`
![image](https://user-images.githubusercontent.com/39240194/53021552-c0a0ad00-3459-11e9-84cc-1cf175463022.png)

PR #723, opened 2019-01-10
[framework] Copy categories to variant if main variant is edited or variant is added

### LINKED ISSUE #586: We cannot work with variants and categories
### What is happening
We cannot define variant categories, so it happens that variant is not visible on a domain.
* we cannot change variants' categories
* categories don't take over categories of main product

### Expected result
*Product owner will define*

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Variants don't have categories of the main variant which can lead to variant being not sellable or other problems #586 
|New feature| No 
|BC breaks| Yes 
|Fixes issues| closes #586  


boris-brtan:
## TESTS
with 100 crowns it worked but when i tried to create variant with image i think something got wrong with flysystem and images were not copied. If this is some other bug please create issue

DavidOstrozlik:
Hi, thanks for review and tests. I've already created an issue for that problem #686.

PR #698, opened 2018-12-20
[shopsys] manageable continuous integration via .ci folders

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| we will have scripts for building custom special jobs for docker-compose and kubernetes projects
|New feature| No 
|BC breaks| No 
|Fixes issues| - 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

Issue #408, opened 2018-08-16
describe parallel splitting on our CI in monorepo-tools' docs


PetrHeinz:
### What is happening

The `monorepo_split.sh` can take a while on large repostories. We have developed a better way of splitting for our CI, but haven't described it in the docs, even though it could be very helpful to others:

```bash
#!/bin/bash +x

# ANSI color codes
RED="\e[31m"
GREEN="\e[32m"
BLUE="\e[34m"
NC="\e[0m"

# Relatively new version of git must be installed
printf "\n${BLUE}Using $(git --version). The package shopsys/monorepo-tools was tested on 2.16.1.${NC}\n\n"

# List of packages to be split
PACKAGES="\
    framework \
    product-feed-zbozi \
    product-feed-google \
    product-feed-heureka \
    product-feed-heureka-delivery \
    plugin-interface \
    coding-standards \
    http-smoke-testing \
    form-types-bundle \
    migrations \
    monorepo-tools \
    project-base"

for PACKAGE in $PACKAGES; do
    # Preparing the variables to be used for splitting
    LOG_FILE="$WORKSPACE/split/$PACKAGE.log"
    if [ "$PACKAGE" == "project-base" ]; then
        SUBDIRECTORY=$PACKAGE
    else
        SUBDIRECTORY="packages/$PACKAGE"
    fi
    
    # Cloning the repository as bare into separate directory, so it can be split in a parallel process
    cd $WORKSPACE
    git clone --bare .git $WORKSPACE/split/$PACKAGE
    cd $WORKSPACE/split/$PACKAGE
    
    printf "${BLUE}$(date +%T) > Splitting package '$PACKAGE' from directory '$SUBDIRECTORY'...${NC}\n"
    printf "The progress will be logged into '$LOG_FILE'.\n\n"
    
    # Running the splitting processes in parallel
    $WORKSPACE/packages/monorepo-tools/rewrite_history_from.sh $SUBDIRECTORY master $(git tag) > $LOG_FILE 2>&1 &&
        printf "${GREEN}$(date +%T) > Splitting package '$PACKAGE' from directory '$SUBDIRECTORY' finished!${NC}\n" ||
        printf "${RED}$(date +%T) > Splitting package '$PACKAGE' from directory '$SUBDIRECTORY' failed!${NC}\n" &
done

wait

# Checking the status of the split repositories
EXIT_STATUS=0
[...]

PR #729, opened 2019-01-11
[shopsys] added scripts for monorepo maintenance into /.ci directory

PetrHeinz:
- basically copy-pasted from our Jenkins configuration so we can change it with the codebase (eg. add a package inside a PR as a modification of `get_all_packages()` in `monorepo_functions.sh` and let it be split automatically after the PR is merged)
- `monorepo_split.sh` splits master into all subrepositories
  - should be executed automatically after `master` is successfully built
- `monorepo_force_split_branch.sh` splits and force pushes selected branch into all subrepositories
  - should be executed manually as a parameterized build
- `monorepo_remove_branch.sh` removes selected branch from all subrepositories
  - should be executed manually as a parameterized build
  - cleans up after `monorepo_force_split_branch.sh`
- `monorepo_functions.sh` contains reusable functions
- related to #698 and #408
- allows us to simplify our Jenkins configuration for monorepo maintenance to something like:
image

| Q             | A
| ------------- | ---
|Description, reason for the PR| adding a package requires changes to Jenkins configuration for it to be split (or stop being split for package removal)
|New feature| Yes 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/.ci/monorepo_force_split_branch.sh
+++ b/.ci/monorepo_force_split_branch.sh
@@ -44,10 +44,10 @@ for PACKAGE in $(get_all_packages); do
     cd $WORKSPACE
     git clone --bare .git $WORKSPACE/split/$PACKAGE
     cd $WORKSPACE/split/$PACKAGE
-    
+
     printf "${BLUE}$(date +%T) > Splitting package '$PACKAGE' from directory '$SUBDIRECTORY'...${NC}\n"
     printf "The progress will be logged into '$LOG_FILE'.\n\n"
-    
+
     # Running the splitting processes in parallel
     $WORKSPACE/packages/monorepo-tools/rewrite_history_from.sh $SUBDIRECTORY $SPLIT_BRANCH > $LOG_FILE 2>&1 &&
         printf "${GREEN}$(date +%T) > Splitting package '$PACKAGE' from directory '$SUBDIRECTORY' finished!${NC}\n" ||
@@ -62,9 +62,9 @@ printf "\n${BLUE}$(date +%T) > Splitting of all packages finished. Checking the 
 for PACKAGE in $(get_all_packages); do
     cd $WORKSPACE/split/$PACKAGE
     REMOTE=$(get_package_remote "$PACKAGE")
-    
+
     git push $REMOTE $SPLIT_BRANCH --dry-run --force
-    
+
     if [[ $? -eq 0 ]]; then
         printf "${GREEN}The split package '$PACKAGE' can be pushed into '$REMOTE'!${NC}\n"
     else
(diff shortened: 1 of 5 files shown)


.ci/monorepo_force_split_branch.sh
```diff
@@ -0,0 +1,85 @@
...
+# Example: monorepo_force_split_branch.sh my-feature-branch
+
+# ANSI color codes
+RED="\e[31m"
+GREEN="\e[32m"
+BLUE="\e[34m"
+NC="\e[0m"
+
+SPLIT_BRANCH=$1
+WORKSPACE=${WORKSPACE:-.}
```
boris-brtan:
```suggestion
WORKSPACE=${WORKSPACE:-$PWD}
```
if default value for $workspace is relative path and we will execute 
```sh
cd $WORKSPACE
git clone --bare .git $WORKSPACE/split/$PACKAGE
cd $WORKSPACE/split/$PACKAGE
```
in the next iteration we will not be in the original $WORKSPACE but in $WORKSPACE/split/$PACKAGE

.ci/monorepo_split.sh
```diff
@@ -0,0 +1,72 @@
...
+done
+
+wait
+
+# Checking the status of the split repositories
+EXIT_STATUS=0
+printf "\n${BLUE}$(date +%T) > Splitting of all packages finished. Checking the ability to push the split repositories...${NC}\n\n"
+for PACKAGE in $(get_all_packages); do
+    cd $WORKSPACE/split/$PACKAGE
+    REMOTE="git@github.com:shopsys/$PACKAGE.git"
```
boris-brtan: REMOTE variable is used here for dry-run but it is not used in real push

.ci/monorepo_force_split_branch.sh
boris-brtan: `monorepo_force_split_branch.sh` and `monorepo_split.sh` differ on just a few places.

PetrHeinz: Thanks for noticing, I've fixed that in 11a620a

.ci/monorepo_split.sh
PetrHeinz: It's in a for-cycle so it cannot be reused below. The value should be the same though (`"git@github.com:shopsys/$PACKAGE.git"`).

It could be re-assigned and used the same way (`REMOTE="git@github.com:shopsys/$PACKAGE.git"`) so it's obvious that they are the same, or it could be moved to `monorepo_functions.sh` in a new function `get_package_remote` so it's actually located in one place.

PetrHeinz: I have done the extraction in b83fd6b

PR #730, opened 2019-01-13
[shopsys] Google Cloud Storage Bucket

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We need to use storage for our files provided by Google Cloud to be able to stop/remove pods without losing files because they would be in running pods, this way it is saved onto Google Cloud Storage Bucket.
|New feature| Yes
|BC breaks| No
|Fixes issues| 


packages/google-cloud-bundle/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+This bundle is used to allow Shopsys Framework](https://www.shopsys-framework.com) integration with Google Cloud.
+
+This repository is maintained by [shopsys/shopsys] monorepo, information about changes are in [monorepo CHANGELOG.md](https://github.com/shopsys/shopsys/blob/master/CHANGELOG.md).
+
+## Installation
+The plugin is a Symfony bundle and is installed in the same way:
+
+### Download
+First, you download the package using [Composer](https://getcomposer.org/):
```
PetrHeinz: It sound like a second section, perhaps *Bundle registration*, is missing. It should cover adding the bundle into the list in `AppKernel.php`.

packages/google-cloud-bundle/.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+
+php:
+    - 7.1
+    - 7.2
+
+install:
+    - composer install
+
+script:
+    - php vendor/bin/parallel-lint ./src
```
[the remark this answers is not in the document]
MattCzerner: I wouldnt necessary add composer memory limit if its not needed. And its not.

project-base/infrastructure/google-cloud/stateful-services.tf
```diff
@@ -1,6 +1,7 @@
 resource "google_storage_bucket" "file-store" {
   name     = "${var.GOOGLE_CLOUD_STORAGE_BUCKET_NAME}"
   location = "EU"
+  force_destroy = true
```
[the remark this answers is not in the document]
MattCzerner: I dont hink its necessary, I dont think that deleting your whole infrastructure will be the case in production environment, therefore there will not be need to backup these data. We are still not close to production like environment so I dont think thats necessary.

packages/form-types-bundle/composer.json
```diff
@@ -22,6 +22,7 @@
     },
     "require": {
         "php": "^7.0",
+        "ext-redis": "*",
```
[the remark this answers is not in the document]
MattCzerner: it was a mistake and this was fixed by fixup commit, this change does not exist anymore.

packages/google-cloud-bundle/composer.json
```diff
@@ -18,7 +18,10 @@
     "require": {
         "php": "^7.1",
         "google/cloud-storage": "^1.10",
-        "shopsys/framework": "dev-master",
+        "shopsys/form-types-bundle": "@dev",
```
boris-brtan: change it to dev-master for all shopsys/* because of `utils/releaser` package

boris-brtan:
## Tests
/content/images/product/original/1.jpg urls got 404
wysiwyg doesn't seem to upload files into bucket
feeds are somehow not generated into bucket nor the local

packages/google-cloud-bundle/composer.json
PetrHeinz: Why? What would happen using `@dev`?

According to [Composer docs](https://getcomposer.org/doc/04-schema.md#package-links), this is the correct solution (as it's not a dependency, rather a allowed expection to the minimal-stability rule).

docs/introduction/abstract-filesystem.md
```diff
@@ -1,7 +1,7 @@
 # Abstract Filesystem
 One of the goals of the Shopsys Framework is to give you tools to make your e-commerce platform scalable.
 One of the requirements for scalable application is separated file storage that is accessible from all application instances.
-We use abstract filesystem - [Flysystem](https://github.com/thephpleague/flysystem) via [OneUpFlysystemBundle](https://github.com/1up-lab/OneupFlysystemBundle) to fulfill this requirement.
+We use abstract filesystem - [Flysystem](https://github.com/thephpleague/flysystem).
```
boris-brtan: i am not sure whether this should not be as part of html comments `<!--  blablabla change during release blablabla -->` that will be updated after new release is done but .... i think this should be solved during BV.

project-base/kubernetes/kustomize/overlays/production/kustomization.yaml
```diff
@@ -19,4 +19,4 @@ secretGenerator:
 configMapGenerator:
     -   name: nginx-configuration
         files:
-        -   ../../../../docker/nginx/nginx.conf
+        -   ../../../../docker/nginx/nginx-google-cloud.conf
```
boris-brtan: is this OK from the view of the `key` value?
```

                -   name: nginx-configuration
                    configMap:
                        name: nginx-configuration
                        items:
                        -   key: nginx.conf
                            path: default.conf
```

project-base/docker/nginx/nginx-google-cloud.conf
```diff
@@ -0,0 +1,66 @@
...
+            expires 1w;
+
+            # (try_files directive is not inherited from the parent location so we have to repeat it)
+            try_files $uri @app;
+        }
+
+
+        location ~ ^/content/ {
+            resolver 8.8.8.8;
+            proxy_pass https://storage.googleapis.com/{{GOOGLE_CLOUD_STORAGE_BUCKET_NAME}}/web$request_uri;
```
boris-brtan: is there something to do to have google cloud bucket publicly accessible? if yes please write something about it in docs.

boris-brtan: https://erikzaadi.com/2014/06/23/handle-proxy-404-in-nginx/

project-base/.ci/deploy-to-google-cloud.sh
```diff
@@ -17,6 +20,9 @@ docker image pull ${DOCKER_USERNAME}/php-fpm:${DOCKER_IMAGE_TAG} || (
     docker image push ${DOCKER_USERNAME}/php-fpm:${DOCKER_IMAGE_TAG}
 )
 
+# Set proxy url address to google cloud storage bucket API
+sed "s/{{GOOGLE_CLOUD_STORAGE_BUCKET_NAME}}/${GOOGLE_CLOUD_STORAGE_BUCKET_NAME}/g" docker/nginx/nginx-google-cloud.conf
```
boris-brtan:
```suggestion
sed -i "s/{{GOOGLE_CLOUD_STORAGE_BUCKET_NAME}}/${GOOGLE_CLOUD_STORAGE_BUCKET_NAME}/g" docker/nginx/nginx-google-cloud.conf
```
replace of file needs to be with -i parameter

project-base/docker/nginx/nginx-google-cloud.conf
```diff
@@ -0,0 +1,66 @@
...
+
+                # (try_files directive is not inherited from the parent location so we have to repeat it)
+                try_files $uri @app;
+            }
+
+            # (try_files directive is not inherited from the parent location so we have to repeat it)
+            try_files $uri @app;
+        }
+
+        location ~ ^/content(-test)?/images/ {
```
boris-brtan: not sure whether it makes sense

project-base/docker/nginx/nginx-google-cloud.conf
```diff
@@ -0,0 +1,66 @@
...
+        # hide dotfiles (send to @app)
+        try_files @app @app;
+    }
+    location ~ ^/index.php$ {
+        # hide index.php (send to @app)
+        try_files @app @app;
+    }
+    location / {
+        # try to serve existing files directly, fallback to @app
+        try_files $uri @app;
```
boris-brtan: make sure that this will have lower priority than proxy_pass for /content for GCS-Bucket

project-base/.ci/deploy-to-google-cloud.sh
MattCzerner: I dont think so, i searched on how to do it but it seems like you need to use some kind of plugins.. Which even though they would work, meant that we will ned to extend kubernetes-buildpack.

project-base/docker/nginx/nginx-google-cloud.conf
MattCzerner: i would leave it to another US or issue?

packages/google-cloud-bundle/composer.json
simara-svatopluk: @PetrHeinz `@dev` is really just allowing to install transitive `dev-*` dependencies.

But this package depends on `shopsys/framework`, so the `shopsys/framework` must be required in `dev-master` dependency (as we are not allowing installation, but it is requirement).

What do you think?

PetrHeinz: Well, there is a service definition that decorates `Shopsys\FrameworkBundle\Component\Filesystem\MainFilesystemFactory` in `services.yml`, which is a dependency on `shopsys/framework`. I agree :+1:

simara-svatopluk: This conversation is closed by https://github.com/shopsys/shopsys/pull/878

PR #732, opened 2019-01-14
[framework] avoid 404 in new product page on an empty shop

### LINKED ISSUE #615: Error 404 in admin if default unit is not set
### What is happening

Page admin_product_new throws error 404 if default unit is not setted. Maybe that problem is on another pages. 

### Expected result

There should be redirect to main page with flash message that tells something like "You need to set default unit" or maybe better redirect directly to page where i could set that unit.

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Page admin_product_new throws error 404 if default unit is not setted. 
|New feature| No 
|BC breaks| No 
|Fixes issues| closes #615


simara-svatopluk:
use `./phing build-new` for testing

packages/framework/src/Controller/Admin/ProductController.php
```diff
@@ -167,7 +168,12 @@ public function editAction(Request $request, $id)
      */
     public function newAction(Request $request)
     {
-        $productData = $this->productDataFactory->create();
+        try {
+            $productData = $this->productDataFactory->create();
+        } catch (HttpException $e) {
```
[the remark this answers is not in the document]
simara-svatopluk: We (me+David) decided to catch general Http Not Found Exception because it is used in framework for not founds. When user extends the factory and add a new default field (and if have a clean database), this will catch it and redirect to dashboard where should be warnings about not-setted default values.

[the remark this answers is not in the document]
PetrHeinz: OK, that makes sense, so how about catching a bit more specific `\Symfony\Component\HttpKernel\Exception\NotFoundHttpException`?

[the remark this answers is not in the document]
simara-svatopluk: Yep, it was a mistake, chaning...

grossmannmartin:
Hi, I tested this PR.

I built the application for production with base data only (with `php phing build-new`) and when I try to create a product, I get a bunch of warnings, what I need to set (please notice the sentence *fill all default values*).

snimek obrazovky 2019-01-17 v 13 51 57

After creating some unit and set the default unit, I can proceed to a page with a product creation without a problem, although I didn't solve any other warning shown.

And here it's even more confusing because I'm still unable to create the product because I don't have any "availability" created (required field, without any possible value).

I'm aware, that these warnings are still on the dashboard page. 
I'm not even sure if this should be solved as a part of this PR, but I had the urge to say it.

simara-svatopluk:
Thanks for testing @grossmannmartin 

Your observation that you are not able to create product because there is no availability is correct and I think it have to be solved in the same fashion.

I made a fix for this and sending it back to CR.

simara-svatopluk:
sorry... I decided that this have to be discussed during BV because it is a strange behavoir and maybe it is so edge-case that we don't want to solve it

simara-svatopluk:
This PR needs major improvements:
* redirecting to `product list` instead of dashboard
* if there some data that will not allow create a new product, the button "new product" is disabled
* there have to be an explanation next to disabled button (tooltip/text) that user have to fill some data, and link to dashboard where they are named

PetrHeinz:
@simara-svatopluk That is much harder to implement. In the suggested implementation user is redirected to dashboard, because there are warning messages displayed there.

In my opinion, this implementation is easy, introduces no BC-break and the result is better than the current state. **Do we really want to block merging this in favor of implementing something else?**

packages/framework/src/Controller/Admin/ProductController.php
```diff
@@ -372,4 +408,18 @@ public function visibilityAction($productId)
...
+
+    /**
+     * @return bool
+     */
+    protected function productCanBeCreated()
+    {
+        $productCanBeCreated = true;
+
+        if (empty($this->unitFacade->getAll()) || $this->setting->get(Setting::DEFAULT_UNIT) === 0 || empty($this->availabilityFacade->getAll())) {
+            $productCanBeCreated = false;
```
simara-svatopluk: unnecessary variable, please change to `return false` and after `if` statement `return true`

packages/framework/src/Controller/Admin/ProductController.php
```diff
@@ -372,4 +408,18 @@ public function visibilityAction($productId)
             'domains' => $this->domain->getAll(),
         ]);
     }
+
+    /**
+     * @return bool
+     */
+    protected function productCanBeCreated()
```
simara-svatopluk: please consider return type `protected function productCanBeCreated(): bool`

packages/framework/src/Resources/views/Admin/Content/Product/list.html.twig
```diff
@@ -3,12 +3,24 @@
...
+    {% if productCanBeCreated %}
+        <a href="{{ url('admin_product_new') }}" class="btn btn--plus wrap-bar__btn" target="_blank">
+            <i class="btn__icon">+</i>{{ 'Create new product'|trans }}
+        </a>
+        <a href="{{ url('admin_product_createvariant') }}" class="btn btn--plus wrap-bar__btn">
+            <i class="btn__icon">+</i>{{ 'Create variant'|trans }}
+        </a>
+    {% else %}
+        <a href="{{ url('admin_default_dashboard') }}" class="btn-link-style wrap-bar__btn">
+            {{ 'Some settings are missing. Go to Dashboard for more information.'|trans }}
```
simara-svatopluk:
```suggestion
            {{ 'You cannot create a product because some settings are missing. Go to Dashboard for more information.'|trans }}
```

simara-svatopluk: Please change also translations

packages/framework/src/Controller/Admin/ProductController.php
TomasLudvik: updated

packages/framework/src/Controller/Admin/ProductController.php
TomasLudvik: that class does not use strict types so I will not use it here

simara-svatopluk:
@TomasLudvik :heavy_check_mark:

PR #733, opened 2019-01-14
[shopsys] twig/twig v2.6.1 is now ignored

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| due to https://github.com/twigphp/Twig/issues/2810 we have to ignore version 2.6.1 of twig/twig package
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -145,7 +145,8 @@
     "symplify/monorepo-builder": "^5.2.17"
   },
   "conflict": {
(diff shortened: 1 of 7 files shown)


grossmannmartin:
There is already PR twigphp/Twig#2813 that should fix this error and it is assumed it will be fixed in another release. 
I don't think its necessary to mark as conflicting any version >= 2.6.1

PR #736, opened 2019-01-14
[framework] added phing target for cleaning redis cache: clean-redis

### LINKED ISSUE #442: Redis cache is not cleared after deploy with native installation
### What is happening

First of all SSFW doesn't have any command to clear Redis cache. Clearing of redis should be as part of phing target `clean`.

During the deploy, new release will be created next to the currently running release and new release need to work with a cache. If new release will read from current cache it may cause an error. I could still delete that cache but there is another problem when someone will accede to the currently running release and he will write to redis cache again. 

Another problem will be if i will have two instances on same server (development and production instances). 

### Expected result

Redis cache should be cleared after deploy new release. 
Redis key should have some prefix for every SSFW instance and unique ID for every build so application can easily clear old cache for his instance after deploy. Maybe unique ID would be generated after build container.

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Redis cache can be cleared after application deploy
|New feature| Yes
|BC breaks| No
|Fixes issues|


PetrHeinz:
I've noticed that upgrade notes are missing in the PR.

PetrHeinz:
@simara-svatopluk To me, this sounds like a thing that should be included in the optional upgrade notes. Otherwise, people will not be able to use this new documented feature even though they have upgraded according to our advice.

packages/framework/src/Command/RedisCleanCacheCommand.php
```diff
@@ -0,0 +1,64 @@
...
+            ->setDescription('Cleans up redis cache');
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Input\InputInterface $input
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        foreach ($this->redisClients as $redis) {
```
henzigo: I have an idea. What if you have some default client with redis prefix only. So you dont need foreach for all clients and you will be able to remove all cache.

simara-svatopluk:
@PetrHeinz You are right, optional upgrade note will be fine

packages/framework/src/Command/RedisCleanCacheCommand.php
```diff
@@ -0,0 +1,64 @@
...
+            $prefix = $redis->getOption(Redis::OPT_PREFIX);
+            $keys = $redis->keys('*');
+            $plainKeys = $this->removePrefixes($keys, $prefix);
+            $redis->del($plainKeys);
```
henzigo: This implementation can take a long time. I found better solution for me. If you use some default client, you dont need to find all keys and remove prefixes to be able to delete them. 

This is the method from facade in my project. You can specific custom pattern (bestselling_products, twig, etc.) after prefix and Id for example if you want to remove only one section in twig cache.
```
public function clearCacheByPattern(string $pattern = '', string $id = null): void
{
    $pattern = $this->cachePrefix . $pattern . '*';
    if ($id !== null) {
        $pattern .= $id . '*';
    }
    $this->redisClient->eval('return redis.call(\'del\', unpack(redis.call(\'keys\', ARGV[1])))', [$pattern]);
}
```

project-base/build.xml
```diff
@@ -89,6 +89,13 @@
         </delete>
     </target>
 
+    <target name="clean-redis" description="Cleans up redis cache">
```
henzigo: People would be confused that `clean` command deletes only var folder. What about create command `clean-all` or use `clean` (and rename old `clean` to `clean-var`) which will clean var folder and redis cache together

project-base/build.xml
```diff
@@ -52,10 +52,10 @@
     </if>
 
     <target name="build" depends="build-deploy-part-1-db-independent, build-deploy-part-2-db-dependent" description="Builds application for production preserving your DB."/>
-    <target name="build-deploy-part-1-db-independent" depends="clean,composer,npm,dirs-create,assets" description="First part of application build for production preserving your DB (can be run without maintenance page)."/>
+    <target name="build-deploy-part-1-db-independent" depends="clean,clean-redis,composer,npm,dirs-create,assets" description="First part of application build for production preserving your DB (can be run without maintenance page)."/>
```
henzigo: This can be an issue if you will be deploying new release and someone will approach to the portal. At that moment he writes to redis cache informations that may not be consistent with new release. After deploy application can throws error.

henzigo: We preceed this by creating a prefix that is generated during a build. It's working like generating css files. So we have file with that prefix and with phing target, we change that version
```
parameters:
    build-version: %%version%%
```

```
<target name="generate-build-version">
    <exec executable="${path.php.executable}" checkreturn="true" outputProperty="version">
        <arg value="-r" />
        <arg value="echo uniqid(true);" />
    </exec>
    <copy file="${path.app}/config/parameters_version.yml.dist" tofile="${path.app}/config/parameters_version.yml" overwrite="true">
        <filterchain>
            <replacetokens begintoken="%%" endtoken="%%">
                <token key="version" value="${version}" />
            </replacetokens>
        </filterchain>
    </copy>
</target>
```

simara-svatopluk:
@henzigo I'm sorry you are sad. You expected that we'll consult this solution with you, but how did you let us know that you have a solution for this problem? I'm just wondering, because there is no open PR that solves this issue and in your issue #442 there is no mentioning about proposed solution, or that you are offering a consultation.

I don't want to be rude, I'm surprised by your reaction and trying to explain my point of view. Well next when you'll expect a consultation, please write it into the issue, and I'll :heart: to consult the solution. Because confrontation with a different opinion leads to better results.

project-base/build.xml
DavidOstrozlik: I agree with this because we don't want any cache errors after deployment

docs/introduction/console-commands-for-application-management-phing-targets.md
```diff
@@ -84,6 +84,11 @@ Cleans up directories with cache and scripts which are generated on demand.
 
 Your go-to command when you feel something should work but does not. Especially useful in the test environment in which cache is not automatically invalidated.
 
+#### clean-redis
+Cleans up cache in Redis database.
```
DavidOstrozlik:
```suggestion
Cleans up cache in Redis database except for sessions.
```

packages/framework/src/Command/RedisCleanCacheCommand.php
DavidOstrozlik: This is a good idea but it would require complete refactoring of the way we use Redis and I think that should not be part of this PR, so please create Issue or new PR.

packages/framework/src/Command/RedisCleanCacheCommand.php
DavidOstrozlik: As I said above, I think that should not be part of this PR, so it would be great if you could create Issue or new PR.

project-base/build.xml
DavidOstrozlik: Now I don't agree with this and I think it should be part of the PR or issue as I mentioned above

packages/framework/src/Command/RedisCleanCacheCommand.php
simara-svatopluk: I used this `eval` solution, so there is no need to move it to the issue

simara-svatopluk:
We need to create an issue where we'll describe unique ID proposal together will all unsolved points by henzigo.

henzigo:
I appreciate that my points are useful, but i don't like how you work with them. There is an issue (https://github.com/shopsys/shopsys/pull/736#discussion_r247842225) in this PR and i don't agree with merging this PR without fix this issue. 

First of all i do not like that you will merge new feature with bug where there may be a problem for users who use SSFW. I just want to remind you that this product aren't using only Shopsys programmers. Yes you want to create issue, that will solve that problem, but i don't think that you will solve that before you will release new version. So here's the problem you'll know about, that could be solved during implementation.

Secondly i don't think that without solving that bug, you can close that issue, because in this condition you don't have fixed whole issue, because that bug is mentioned there.

simara-svatopluk:
@henzigo I'm writing to clarify communication.

This PR doesn't close your issue #442, it was my mistake. Your issue will not be solved when this issue is done (I just changed the description).

Anyway this PR have value - there is a new phing target that allows users to clear redis cache.
This is a way we work - bringing a value step by step. PRs are easier to track, review and test in this fashion. And we prefer to bring some value instead of waiting to solve all issues.

And I promise you that your issue will stay on our scrumboard so we keep working on it.

PR #737, opened 2019-01-15
[shopsys] docs: use "sh" instead of "bash" in docker exec commands

vitek-rostislav:
- Debian version with bash is not released yet
- added TODOs for the next release

| Q             | A
| ------------- | ---
|Description, reason for the PR| When someone tries to install project-base by the current installation guide, he will encounter an error - `OCI runtime exec failed: exec failed: container_linux.go:344: starting container process caused "exec: \"bash\": executable file not found in $PATH": unknown`
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/docs/docker/docker-troubleshooting.md
+++ b/docs/docker/docker-troubleshooting.md
@@ -84,8 +84,9 @@ Remember that after changing these you need to do few things differently.
 * You changed `container_name` of php-fpm which means that in order to get inside the php-fpm container you must now use this name.
   for instance, if your new container name is `my-new-project-name-php-fpm` you need to execute
 
+<!--- TODO exhchange "sh" with "bash" when releasing a new version -->
 ```
-docker exec -it my-new-project-name-php-fpm bash
+docker exec -it my-new-project-name-php-fpm sh
 ```
 
 ## Update of Dockerfile is not Reflected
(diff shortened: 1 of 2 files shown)


docs/docker/docker-troubleshooting.md
```diff
@@ -84,8 +84,9 @@ Remember that after changing these you need to do few things differently.
 * You changed `container_name` of php-fpm which means that in order to get inside the php-fpm container you must now use this name.
   for instance, if your new container name is `my-new-project-name-php-fpm` you need to execute
 
+<!--- TODO exhchange "sh" with "bash" when releasing a new version -->
 ```
-docker exec -it my-new-project-name-php-fpm bash
+docker exec -it my-new-project-name-php-fpm sh
```
boris-brtan: for info, we can use also sh in debian so there is no need for that change, but debian sh desn't contain autocompleting and other stuff for now.

grossmannmartin: User should always work with documentation for his installed version. This problem can occur in every PR altering documentation. 
I tried to use sh instead of bash and its really annoying to work with.

vitek-rostislav: yes, it is true that he should work with documentation of his version, however, it is not quite intuitive here on Github so we decided to perform changes in the docs this way - by adding TODO comments for next release so, in the current master, there is documentation compatible with the last release

Issue #739, opened 2019-01-15
Image thumbnails are not generated due to missing JPEG support in PHP


grossmannmartin:
### What is happening

Version from master, probably since Debian PHP-FPM image was introduced.

Steps:
Perform a clean installation of the latest version and check the front page.
Image thumbnails are missing. 

On direct access to the URL of image thumbnail (for example http://127.0.0.1:8000/content/images/product/list/9.jpg) the following exception is thrown

`Generate image for entity "product" (type=, size=list, imageId=9) failed.`

With warning: **Warning: imagecreatefromstring(): No JPEG support in this PHP build**

### Expected result
Images are properly generated.

boris-brtan:
missing support for [jpeg and freetype](https://github.com/exozet/docker-php-fpm/pull/4/files) that was in alpine
https://github.com/shopsys/shopsys/pull/702/files#diff-50a0e02c146dc64c2a172b42022589faL33

grossmannmartin:
Hi @boris-brtan,
thanks for quick response. This solution seems good to me. 
Please don't forget to update upgrade file.

PR #740, opened 2019-01-15
generation of jpeg previews for php-fpm debian container

boris-brtan:
- generation of image previews for brands, products, ... is fixed

| Q             | A
| ------------- | ---
|New feature| No 
|BC breaks| No 
|Fixes issues| #739  

### DIFF
--- a/project-base/docker/php-fpm/Dockerfile
+++ b/project-base/docker/php-fpm/Dockerfile
@@ -31,11 +31,16 @@ RUN chmod +x /usr/local/bin/docker-install-composer && \
 # autoconf needed by "redis" extension
 RUN apt-get install -y \
     libpng-dev \
+    libjpeg-dev \
+    libfreetype6-dev \
     libzip-dev \
     libicu-dev \
     libpq-dev \
     autoconf && \
     apt-get clean
+
+# "gd" extension needs to have specified jpeg and freetype dir for jpg/jpeg images support
+RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/
 
 # install necessary tools for running application
 RUN docker-php-ext-install \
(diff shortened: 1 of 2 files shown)


project-base/docker/php-fpm/Dockerfile
```diff
@@ -31,12 +31,17 @@ RUN chmod +x /usr/local/bin/docker-install-composer && \
 # autoconf needed by "redis" extension
 RUN apt-get install -y \
     libpng-dev \
+    libjpeg-dev \
+    libfreetype6-dev \
```
simara-svatopluk: do we really need the free type library?

simara-svatopluk: or... why do we need the free type library?

project-base/docker/php-fpm/Dockerfile
```diff
@@ -31,12 +31,17 @@ RUN chmod +x /usr/local/bin/docker-install-composer && \
...
     libzip-dev \
     libicu-dev \
     libpq-dev \
     autoconf && \
     apt-get clean
 
+# "gd" extension needs to have specified jpeg and freetype dir for jpg/jpeg images support
+RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/
```
simara-svatopluk: and please why do we need to specify directories and why is it `/usr/include`? because this directory contains more than just jpeg support

project-base/docker/php-fpm/Dockerfile
boris-brtan: i do not know. it was just copied from alpine version that we had before.

project-base/docker/php-fpm/Dockerfile
boris-brtan: http://php.net/manual/en/image.installation.php based on manual

project-base/docker/php-fpm/Dockerfile
boris-brtan: https://github.com/shopsys/shopsys/issues/741

PR #750, opened 2019-01-18
[shopsys] monorepo force splitting using a RAM-disk

PetrHeinz:
- requires permission to sudo "mount" and "umount" commands (set in "/etc/sudoers")
- if the permission is not provided, RAM-disk will not be mounted
- at the end of the script it will back-up the data and unmount the volume (this will not be done if the script is interrupted; signal handling using "trap" would need to be implemented)

| Q             | A
| ------------- | ---
|Description, reason for the PR| using a tmpfs mount should increase performance and decrease disk usage; this is just a prototype (considering only `monorepo_force_split_branch.sh`)
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/.ci/monorepo_force_split_branch.sh
+++ b/.ci/monorepo_force_split_branch.sh
@@ -35,6 +35,11 @@ printf "\n${BLUE}Using $(git --version). The package shopsys/monorepo-tools was 
 # Import functions
 . $(dirname "$0")/monorepo_functions.sh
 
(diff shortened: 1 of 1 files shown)


PetrHeinz:
There was no significant performance boost during testing.

Splitting of this branch took 25 minutes, splitting of a testing branch created on `master` took 26 minutes, which is not much of an improvement. The added complexity and requirement of a record in `/etc/sudoers` is not worth it.

PR #758, opened 2019-01-22
fix and enhance configuration of docker-compose and docker-sync for windows development

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| seems that web volume is mounted the same way as in project-base so it needs to be corrected and docker-sync volumes for microservices could be used in php-fpm container 
|New feature| No 
|BC breaks| No 
|Fixes issues| - 


boris-brtan:
rebased to contain install script

Issue #495, opened 2018-09-20
Use shell files for installation


MattCzerner:
To install the application there is several commands you must execute to do that. These commands are written in documentation, but they could be stored into shell files to make installation really easy.

### How it could look like 

To set up default and simple development infrastructure you can do it by:

```
cp app/config/parameters.yml.dist app/config/parameters.yml
cp app/config/parameters_test.yml.dist app/config/parameters_test.yml
cp app/config/domains_urls.yml.dist app/config/domains_urls.yml
cp docker/conf/docker-compose.yml.dist docker-compose.yml
docker-compose up -d
```

To install the application you can do:

```
docker exec -it shopsys-framework composer install
docker exec -it shopsys-framework ./phing build-demo-dev
```
*How about to copy these common commands to shell files and allow users to install the application on once click?*

I would like to see some opinions on that.

vitek-rostislav:
I love it :heart: Of course, you would have to detect the correct `docker-compose` template by your operating system in the script.

MattCzerner:
Or just leave it as an option to fill.

something like 
```
Choose your operating system:

1) Linux
2) Windows
3) Mac
```

MattCzerner:
Install script implemented, closing

PR #759, opened 2019-01-22
[project-base] Created install script for mac and linux distribution

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| To ease installlation process, we can extract commands required to execute in order to install the Shopsys Framework into shell script that will install fully functional demo version of Shopsys Framework in container.
|New feature| Yes
|BC breaks| No
|Fixes issues| #495

### DIFF
--- a/docs/installation/installation-using-docker-linux.md
+++ b/docs/installation/installation-using-docker-linux.md
@@ -24,7 +24,16 @@ cd project-base
 - *The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
 - *The `--stability=beta` option enables you to install the project from the last beta release. Default value for the option is `stable` but there is no stable release yet.*
 
-### 2. Create docker-compose.yml file
+### 2.1 Use install script
+In case you want to start demo of the app as fast as possible, you can now execute install script.
+
+```
+./scripts/install.sh
+```
+
+If you want to know more about what is happening during installation, continue with next step.
+
+### 2.2 Create docker-compose.yml file
 Create `docker-compose.yml` from template [`docker-compose.yml.dist`](../../project-base/docker/conf/docker-compose.yml.dist).
 ```
 cp docker/conf/docker-compose.yml.dist docker-compose.yml
--- a/docs/installation/installation-using-docker-macos.md
+++ b/docs/installation/installation-using-docker-macos.md
@@ -28,7 +28,16 @@ cd project-base
 - *The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
 - *The `--stability=beta` option enables you to install the project from the last beta release. Default value for the option is `stable` but there is no stable release yet.*
 
-### 1.1 Enable second domain (optional)
+### 1.1 Use install script
+In case you want to start demo of the app as fast as possible, you can now execute install script.
+
+```
+./scripts/install.sh
+```
+
+If you want to know more about what is happening during installation, continue with next step.
+
+### 1.2 Enable second domain (optional)
 There are two domains each for different language in default installation. First one is available via IP adress `127.0.O.1` and second one via `127.0.0.2`.
 `127.0.0.2` is not alias of `127.0.0.1` on Mac by default. To create this alias in network interface run:
 ```
--- a/project-base/scripts/install.sh
+++ b/project-base/scripts/install.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+set -e
+numberRegex="^[0-9]+([.][0-9]+)?$"
+operatingSystem=""
+allowedValues=(1 2 3)
+projectPathPrefix=""
+echo This is installation script that will install demo Shopsys Framework application on docker with all required containers and with demo database created.
+
+echo "Start with specifying your operating system: \
+
+    1) Linux
+    2) Mac
+    3) Windows
+    "
+
+while [[ 1 -eq 1 ]]
+do
+    read -p "Enter OS number: " operatingSystem
+    if [[ ${operatingSystem} =~ $numberRegex ]] ; then
+        if [[ " ${allowedValues[@]} " =~ " ${operatingSystem} " ]]; then
+            break;
+        fi
+        echo "Not existing value, please enter one of existing values"
+    else
+        echo "Please enter a number"
+    fi
+done
+
+if [[ -d "project-base" ]]; then
+    projectPathPrefix="project-base/"
+    echo "You are in monorepo, prefixing paths app paths with ${projectPathPrefix}"
+fi
+
+echo "Creating config files.."
+cp -f "${projectPathPrefix}app/config/parameters.yml.dist" "${projectPathPrefix}app/config/parameters.yml"
+cp -f "${projectPathPrefix}app/config/parameters_test.yml.dist" "${projectPathPrefix}app/config/parameters_test.yml"
+cp -f "${projectPathPrefix}app/config/domains_urls.yml.dist" "${projectPathPrefix}app/config/domains_urls.yml"
+
+echo "Creating docker configuration.."
+case "$operatingSystem" in
+    "1")
+        cp -f docker/conf/docker-compose.yml.dist docker-compose.yml
+        ;;
+    "2")
+        cp -f docker/conf/docker-compose-mac.yml.dist docker-compose.yml
+        cp -f docker/conf/docker-sync.yml.dist docker-sync.yml
+
+        echo "You will be asked to enter sudo password in case to allow second domain alias in your system config.."
+        sudo ifconfig lo0 alias 127.0.0.2 up
+
+        mkdir -p ${projectPathPrefix}var/postgres-data ${projectPathPrefix}var/elasticsearch-data vendor
+        docker-sync start
+        ;;
+    "3")
+        cp -f docker/conf/docker-compose-win.yml.dist docker-compose.yml
+        cp -f docker/conf/docker-sync-win.yml.dist docker-sync.yml
+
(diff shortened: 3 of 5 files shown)


project-base/scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+echo This is installation script that will install demo Shopsys Framework application on docker with all required containers and with demo database created.
+
+echo "Start with specifying your operating system: \
+    1) Linux
+    2) Mac
+    "
+
+while [[ 1 -eq 1 ]]
+do
+    read -p "Enter OS number: " operatingSystem
```
PetrHeinz: This step can be fully automated using `uname` command, which outputs either `"Darwin"` on a Mac or `"Linux"` on Linux.

We have implemented a similar thing in monorepo-tools where used flags differ between the OSs (introduced in #282):
https://github.com/shopsys/shopsys/blob/55e785de1ba7861ebab066acb1d14e595b336b15/packages/monorepo-tools/rewrite_history_from.sh#L38-L44

docs/installation/installation-using-docker-linux.md
```diff
@@ -23,8 +23,16 @@ cd project-base
...
 - *The `--stability=beta` option enables you to install the project from the last beta release. Default value for the option is `stable` but there is no stable release yet.*
+### 2.1 Use install script
+In case you want to start demo of app as fast as possible, you can now execute install script.
 
-### 2. Create docker-compose.yml file
+```
+./sripts/install.sh
+```
+
+If you want to know more about what is happening during installation, continue with step `2.2`
```
boris-brtan: with `next section` or with `next step` ?

docs/installation/installation-using-docker-linux.md
```diff
@@ -23,8 +23,16 @@ cd project-base
 - *The `--no-install` option disables installation of the vendors - this will be done later in the Docker container.*
 - *The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
 - *The `--stability=beta` option enables you to install the project from the last beta release. Default value for the option is `stable` but there is no stable release yet.*
+### 2.1 Use install script
+In case you want to start demo of app as fast as possible, you can now execute install script.
```
boris-brtan: SSFW app | an app | the app

docs/installation/installation-using-docker-linux.md
```diff
@@ -23,8 +23,16 @@ cd project-base
 - *The `--no-install` option disables installation of the vendors - this will be done later in the Docker container.*
 - *The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
 - *The `--stability=beta` option enables you to install the project from the last beta release. Default value for the option is `stable` but there is no stable release yet.*
+### 2.1 Use install script
```
boris-brtan:
```suggestion

### 2.1 Use install script
```

docs/installation/installation-using-docker-macos.md
```diff
@@ -28,7 +28,16 @@ cd project-base
 - *The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
 - *The `--stability=beta` option enables you to install the project from the last beta release. Default value for the option is `stable` but there is no stable release yet.*
 
-### 1.1 Enable second domain (optional)
+### 1.1 Use install script
+In case you want to start demo of app as fast as possible, you can now execute install script.
```
boris-brtan:
```suggestion
In case you want to start demo of an app as fast as possible, you can now execute install script.
```

project-base/scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+
+operatingSystemsByIndex[1]='Linux'
+operatingSystemsByIndex[2]='Mac'
+
+echo This is installation script that will install demo Shopsys Framework application on docker with all required containers and with demo database created.
+
+echo "Start with specifying your operating system: \
+    1) Linux
+    2) Mac
+    "
```
boris-brtan: Windows can be added because we develop on windows via WSL and docker-sync where we have files `docker-sync-win.yml` and `docker-compose-win.yml` so almost the same principle like for Mac

project-base/scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+        echo "Not existing value, please enter one of existing values"
+    else
+        echo "Please enter a number"
+    fi
+done
+
+echo "You chosed ${operatingSystemsByIndex[$operatingSystem]}. Now your app will be installed.."
+
+echo "Creating config files.."
+cp -f app/config/parameters.yml.dist app/config/parameters.yml
```
boris-brtan: is this needed? can you check whether during composer install is that copied automatically from `.dist` files ?

project-base/scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+case "$operatingSystem" in
+    "1")
+        cp -f docker/conf/docker-compose.yml.dist docker-compose.yml
+        ;;
+    "2")
+        cp -f docker/conf/docker-compose-mac.yml.dist docker-compose.yml
+        cp -f docker/conf/docker-sync.yml.dist docker-sync.yml
+
+        echo "You will be asked to enter sudo password in case to allow second domain alias in your system config.."
+        sudo ifconfig lo0 alias 127.0.0.2 up
```
boris-brtan: this will be not needed for windows WSL

project-base/scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+
+        mkdir -p var/postgres-data var/elasticsearch-data vendor
+        docker-sync start
+        ;;
+esac
+
+echo "Starting docker-compose.."
+
+docker-compose up -d --build
```
boris-brtan: maybe --force-recreate and --remove-orphans would be OK when there is also --build param

boris-brtan: is there also some guide or check whether user is able to launch docker-compose or docker ?
because on linux by default we need to launch it with `sudo`

scripts/install.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
boris-brtan: is here some big difference that it couldn't use the same script as it is in project-base with some additional relative path ?
so it would be maintained mainly from one place ?

project-base/scripts/install.sh
boris-brtan: be aware that for windows we recommend to develop in WSL and uname for it is also Linux https://travis-ci.org/boris-brtan/test-travis-os/builds/483328068 but it should behave osx-like

MattCzerner: Yeah I totally forgot about WSL. Ok but anyway, I would like to stay with filling the number of OS since windows will output "Linux" during `uname`..

Feels like not big of a deal? Or do you have any idea on how to make this fully automatic, discovering the windows OS on WSL?

boris-brtan: Ok, for me but to answer the question:
```
uname
-r, --kernel-release     print the kernel release
-v, --kernel-version     print the kernel version
```
returns Microsoft word https://travis-ci.org/boris-brtan/test-travis-os/jobs/483478869
```
4.4.0-17134-Microsoft
#523-Microsoft Mon Dec 31 17:49:00 PST 2018
```
Or there exists `/etc/wsl.conf` file.

project-base/scripts/install.sh
```diff
@@ -0,0 +1,81 @@
...
+
+docker-compose exec php-fpm composer install
+
+echo "Installing application inside a php-fpm container"
+
+docker-compose exec php-fpm ./phing build-demo-dev
```
boris-brtan:
```suggestion
docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev
```

project-base/scripts/install.sh
```diff
@@ -0,0 +1,81 @@
...
+
+echo "Your application is now ready under http://127.0.0.1:8000"
```
boris-brtan: and http://127.0.0.2:8000 ?

boris-brtan: adminer http://127.0.0.1:1100
redis-admin ...
..... ???

docs/installation/installation-using-docker-macos.md
```diff
@@ -28,7 +28,16 @@ cd project-base
 - *The `--keep-vcs` option initializes GIT repository in your project folder that is needed for diff commands of the application build and keeps the GIT history of `shopsys/project-base`.*
 - *The `--stability=beta` option enables you to install the project from the last beta release. Default value for the option is `stable` but there is no stable release yet.*
 
-### 1.1 Enable second domain (optional)
+### 1.1 Use install script
+In case you want to start demo of the app as fast as possible, you can now execute install script.
+
+```
+./sripts/install.sh
```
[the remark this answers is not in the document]
boris-brtan: maybe we can update also `installation-using-docker-windows.md`

project-base/scripts/install.sh
boris-brtan: make sense

project-base/scripts/install.sh
MattCzerner: I would omit the adminer, since this will in my opinion be used only for first installation and for first review of Shopsys Framework.

MattCzerner:
Maybe, we could create a special job to assert that script is functional..

project-base/scripts/install.sh
MattCzerner: That seems little bit sketchy for me. I would prefer to let fill out the OS number.

PetrHeinz:
@MattCzerner can we move the script into another already existing directory? I think we have a lot of root directories already, maybe we can use `.ci`, `build` or `bin` for this. Alternatively, we could remove `.ci` dirs and move the contents into more general `scripts` directories to avoid bloating the root.

What do you think?

Issue #761, opened 2019-01-22
Installation guide is not working


mbohal:
The installation guide https://github.com/shopsys/shopsys/blob/master/docs/installation/installation-using-docker-linux.md is not working

### What is happening
```sh
mb@mb-laptop ~/Projects/PB/shopsys FAILED ~> composer create-project shopsys/project-base --stability=beta --no-install --keep-vcs                                                               17:06:14.680296782
Installing shopsys/project-base (v7.0.0-beta5)
  - Installing shopsys/project-base (v7.0.0-beta5): Cloning 443fcacad4 from cache
Created project in /home/mb/Projects/PB/shopsys/project-base
mb@mb-laptop ~/Projects/PB/shopsys ~> cd project-base                                                                                                                                            17:07:01.489425237
mb@mb-laptop ~/Projects/PB/shopsys/project-base ((v7.0.0-beta5) ✔) ~> cp docker/conf/docker-compose.yml.dist docker-compose.yml                                                                  17:07:06.849002616
mb@mb-laptop ~/Projects/PB/shopsys/project-base ((v7.0.0-beta5) ✔) ~> docker-compose up -d --build                                                                                               17:07:17.478530255
Building php-fpm
Step 1/41 : FROM php:7.2-fpm-stretch as base
ERROR: Service 'php-fpm' failed to build: Error parsing reference: "php:7.2-fpm-stretch as base" is not a valid repository/tag: invalid reference format
mb@mb-laptop ~/Projects/PB/shopsys/project-base ((v7.0.0-beta5) ✔) FAILED ~>
```
### Expected result
The docker container starts up with no errors.

mbohal:
@PetrHeinz I take whatever fedora repo sends my way. Having to install docker version not provided by my distro is not something that I would be overly interested in doing.

What is the minimal docker version? I have 1.13.1.

PetrHeinz:
@mbohal: According to the [docs](https://docs.docker.com/develop/develop-images/multistage-build/), multi-stage builds are supported in **Docker 17.05** or higher.

PetrHeinz:
I've added mention of this to the installation guide and our troubleshooting sections so other users don't encounter the same problem, hope it'll help :slightly_smiling_face:

mbohal:
Thanks for quick reactions.

[...]

PR #764, opened 2019-01-23
[shopsys] explain minimal version of Docker in docs

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| we use multi-stage Docker builds which are supported only in Docker 17.05 or higher, which was not mentioned in our docs (and lead to problems during installation for some people, eg. #761)
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/docs/docker/docker-troubleshooting.md
+++ b/docs/docker/docker-troubleshooting.md
@@ -151,3 +151,14 @@ When `composer install` or `composer update` fails on an error with exceeding th
 
 *Note: Since `v7.0.0-beta4` we have set the Composer memory limit to `-1` (which means unlimited) in the php-fpm's `Dockerfile`.*
 *If you still encounter memory issues while using Docker for Windows (or Mac), try increasing the limits in `Docker -> Preferences… -> Advanced`.*
+
+## Starting up the Docker containers fails due to invalid reference format
+Docker images may fail to build during `docker-compose up -d` due to invalid reference format, eg.:
+```
+Building php-fpm
+Step 1/41 : FROM php:7.2-fpm-stretch as base
+ERROR: Service 'php-fpm' failed to build: Error parsing reference: "php:7.2-fpm-stretch as base" is not a valid repository/tag: invalid reference format
+```
+This is because you have a version of Docker which does not support [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/).
+
+Upgrade your Docker to version **17.05 or higher** and try running the command again.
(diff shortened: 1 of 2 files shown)


docs/docker/docker-troubleshooting.md
```diff
@@ -151,3 +151,14 @@ When `composer install` or `composer update` fails on an error with exceeding th
...
+## Starting up the Docker containers fails due to invalid reference format
+Docker images may fail to build during `docker-compose up -d` due to invalid reference format, eg.:
+```
+Building php-fpm
+Step 1/41 : FROM php:7.2-fpm-stretch as base
+ERROR: Service 'php-fpm' failed to build: Error parsing reference: "php:7.2-fpm-stretch as base" is not a valid repository/tag: invalid reference format
+```
+This is because you have a version of Docker which does not support [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/).
+
+Upgrade your Docker to version **17.05 or higher** and try running the command again.
```
boris-brtan: in guide for osX there is 17.09 and for Win there is mentioned no version, we can make it consistent to 17.05 or in some problems with osX to 17.09.

boris-brtan: Maybe we should highlight that for Linux it needs to be docker-ce distribution

PetrHeinz: Hi, not sure I understand you. The troubleshooting section is just for

docs/installation/installation-using-docker-linux.md
```diff
@@ -9,7 +9,7 @@ Take a look at the article about [Monorepo](../introduction/monorepo.md) for mor
 * [GIT](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 * [PHP](http://php.net/manual/en/install.unix.php)
 * [Composer](https://getcomposer.org/doc/00-intro.md#installation-linux-unix-osx)
-* [Docker](https://docs.docker.com/engine/installation/)
+* [Docker 17.05 or higher](https://docs.docker.com/engine/installation/)
```
boris-brtan:
```suggestion
* [Docker 17.05 or higher](https://docs.docker.com/engine/installation/)
```
osX got 17.09 and win is without version.

docs/docker/docker-troubleshooting.md
boris-brtan: OK, and what about the  consistency for other OS

PR #766, opened 2019-01-24
[shopsys] added support for multiple image sizes (picture element)

### LINKED ISSUE #159: Multi dimensional picture element
### What is happening
We have usually request to have more than one dimension of image. Now we are using only one e.g. in product detail, which is set to maximum size of parent element and centered. 

But now we have new possibility: use picture element

```
<picture>
  <source media="(min-width: 650px)" srcset="img_pink_flowers.jpg">
  <source media="(min-width: 465px)" srcset="img_white_flower.jpg">
  <img src="img_orange_flowers.jpg" alt="Flowers" style="width:auto;">
</picture>
```

Source with description:
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_picture

We have to just think about picture dimensions for desktop, tablet, mobile, retina etc.

### Expected result
For frontend developers there will be no change - bcs this element can be generated by twig function `image`

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Adding support for multiple image sizes to be able to work better with pictures on frontend.
|New feature| Yes 
|BC breaks| No 
|Fixes issues| closes #159 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -45,6 +45,26 @@ There you can find links to upgrade notes for other versions too.
         - change `ExtendedOrderItem` to standard class - remove `abstract` and inheritance annotations
         - change `doTestExtendedOrderItemsPersistence` to test only `OrderItem`
         - please find inspiration in [#715](https://github.com/shopsys/shopsys/pull/715/files)
+- *(optional)* to allow [support for multiple image sizes #766](https://github.com/shopsys/shopsys/pull/766) you have to
+    - implement action `getAdditionalImageAction()` in `Front/ImageController.php` (or copy it from [ImageController.php](https://github.com/shopsys/project-base/blob/master/src/Shopsys/ShopBundle/Controller/Front/ImageController.php))
+    - add routes into your frontend router
+      ```yml
+        front_additional_image:
+            path: /%shopsys.content_dir_name%/images/{entityName}/{type}/{sizeName}/additional_{additionalIndex}_{imageId}.{extension}
+            defaults: { _controller: ShopsysShopBundle:Front\Image:getAdditionalImage }
+            requirements:
+                imageId: \d+
+                additionalIndex: \d+
+
+        front_additional_image_without_type:
+            path: /%shopsys.content_dir_name%/images/{entityName}/{sizeName}/additional_{additionalIndex}_{imageId}.{extension}
+            defaults:
+                _controller: ShopsysShopBundle:Front\Image:getAdditionalImage
+                type: ~
+            requirements:
+                imageId: \d+
+                additionalIndex: \d+
+      ```
 
 [Upgrade from v7.0.0-beta5 to Unreleased]: https://github.com/shopsys/shopsys/compare/v7.0.0-beta5...HEAD
 [shopsys/shopsys]: https://github.com/shopsys/shopsys
--- a/packages/framework/src/Component/Image/Config/ImageConfigLoader.php
+++ b/packages/framework/src/Component/Image/Config/ImageConfigLoader.php
@@ -2,6 +2,8 @@
 
 namespace Shopsys\FrameworkBundle\Component\Image\Config;
 
+use Shopsys\FrameworkBundle\Component\Image\Config\Exception\DuplicateMediaException;
+use Shopsys\FrameworkBundle\Component\Image\Config\Exception\WidthAndHeightMissingException;
 use Shopsys\FrameworkBundle\Component\Utils\Utils;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Filesystem\Filesystem;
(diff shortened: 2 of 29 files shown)


project-base/src/Shopsys/ShopBundle/Resources/config/images.yml
```diff
@@ -10,6 +10,7 @@
 #                    height: 300
 #                    crop: false
 #                    occurrence: 'Front-end: Product detail, when selected'
+#                    additionalSizes: [{width: 1100, height: ~, media: "(min-width: 1200px)"}, {width: 275, height: ~, media: "(max-width: 480px)"}]
```
PetrHeinz: Expanding the array would make it more readable and consistent imo
```suggestion
#                    additionalSizes:
#                        -   {width: 1100, height: ~, media: "(min-width: 1200px)"}
#                        -   {width: 275, height: ~, media: "(max-width: 480px)"}
```

packages/framework/src/Component/Image/Config/ImageConfigLoader.php
```diff
@@ -113,25 +113,44 @@ private function prepareSizes($sizesConfig)
...
         }
 
         return $result;
     }
 
+    /**
+     * @param array $additionalSizesConfig
+     * @return \Shopsys\FrameworkBundle\Component\Image\Config\ImageAdditionalSizeConfig[]
+     */
+    private function prepareAdditionalSizes(array $additionalSizesConfig): array
```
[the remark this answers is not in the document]
simara-svatopluk: Well the current name is consistent with the rest of the class - `prepareSizes`, `prepareTypes`. So I'm not going to rename the method to keep consistency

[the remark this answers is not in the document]
TomasLudvik: oh you are right, I did not like the word "prepare", but we are already using it, so it is ok with me

grossmannmartin:
Hi @simara-svatopluk, I'm sending a few insights from testing. I really like this idea 👍 

- I really miss any documentation about this feature. It's hard to figure out how to configure it and what exactly does it do.
- after change in images.yml (when miniatures were already created) is possible to recreate files miniatures? Didn't find a way how to do it (probably connected with documentation)
- what relation is between parameters width and height? With try&fail approach, I think it defines the upper boundary for one of them – neither width nor height can exceed specified size, but I'm not sure (documentation again?)
- missing mandatory `media` parameter throws a nice error 👍, but missing `width` or `height` is very hard to spot because the error is thrown only on direct access to an image – element is rendered properly with `source`, but will not be used because the image cannot exist. And user will never know.
- duplicate width/height definition for different media cause making two same miniatures instead of re-use one
- duplicate media definition is not warned out, but I'm not considering it a bug
- should there be some upgrade note? I'm not sure, so I'm asking here.

simara-svatopluk:
@grossmannmartin Thanks for testing. I wonder - what is a scope of this US? I really wonder, because the scope could answer at least some of questions.

1-3. It's all about documentation. As far as I know we don't have documentation for images currently. So there is no place to document this new feature. Your points lead to questions that should have been already documented.
4. I just tested this with following configuration, and it works. Do you think that both `width` and `height` must miss?
```php
            additionalSizes:
                - {width: 1100, media: "(min-width: 1200px)"}
                - {height: 500, media: "(max-width: 480px)"}
```
5. duplicate - I don't consider this to be an issue. It has as same behavior as images with different sizes.
6. duplicate media... hm.... Is it really a bug? Should we disallow user to have 2 images for the same media query? Because I don't understand HTML this much.
7. upgrade note.. This is a new feature and if you want to enable it, you have to do some changes in project-base... So yes, I'm writing optional upgrade notes now

As you can see, the main issue is that I don't know if documentation is in scope of this US. Let's discuss it during standup or before it! Please answer 4-6. Maybe we should ask coders about 6. 7 is definitive :-)

grossmannmartin:
Hi @simara-svatopluk,
1-3. In user story documentation was not mentioned. But without documentation, it feels a little incomplete to me.
4. Yeah, both `width` **and** `height` missing. I made a mistake when typing.
5. Okay.
6. I pointed it out, to think about because I don't understand HTML this much either. Therefore I said I don't think it is a bug.
 7. :+1:

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -43,6 +43,26 @@ There you can find links to upgrade notes for other versions too.
         - change `ExtendedOrderItem` to standard class - remove `abstract` and inheritance annotations
         - change `doTestExtendedOrderItemsPersistence` to test only `OrderItem`
         - please find inspiration in [#715](https://github.com/shopsys/shopsys/pull/715/files)
+- *(optional)* to allow [support for multiple image sizes](https://github.com/shopsys/shopsys/pull/766) you have to
+    - implement this feature in `Front/ImageController.php` (or copy it from [ImageController.php](https://github.com/shopsys/project-base/blob/master/src/Shopsys/ShopBundle/Controller/Front/ImageController.php))
```
grossmannmartin: Will be the name of the action helpful? Or it will be clear from the link?

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -43,6 +43,26 @@ There you can find links to upgrade notes for other versions too.
         - change `ExtendedOrderItem` to standard class - remove `abstract` and inheritance annotations
         - change `doTestExtendedOrderItemsPersistence` to test only `OrderItem`
         - please find inspiration in [#715](https://github.com/shopsys/shopsys/pull/715/files)
+- *(optional)* to allow [support for multiple image sizes](https://github.com/shopsys/shopsys/pull/766) you have to
```
grossmannmartin: missing viewable number of pull request

simara-svatopluk:
4. Both width and height missing - new exception is thrown in this case - implemented
6. Duplicate media - together with HTML expert Tom Gottvald we decided that duplicit media have to be checked and exception thrown when occured - implemented

Issue #744, opened 2019-01-15
Error page generate throws confusing error when redis is not running


henzigo:
### What is happening

If redis server is not running and i run `error-page-generate`, command throws confusing error "call get() on null" in [CartMigrationFacade](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Cart/CartMigrationFacade.php#L79) 
 
```
Shopsys Framework > error-pages-generate:
13:22:38 ERROR     [app] Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function get() on null (uncaught exception) at /var/www/html/vendor/shopsys/framework/src/Model/Cart/C
artMigrationFacade.php line 78 while running console command `shopsys:error-page:generate-all` ["exception" => Symfony\Component\Debug\Exception\FatalThrowableError { …}]
13:22:38 ERROR     [console] Error thrown while running command "shopsys:error-page:generate-all". Message: "Call to a member function get() on null" ["exception" => Error { …},"command" => "shopsys:
error-page:generate-all","message" => "Call to a member function get() on null"]
13:22:38 ERROR     [app] Command `shopsys:error-page:generate-all` exited with status code 1
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function get() on null in /var/www/html/vendor/shopsys/framework/src/Model/Cart/CartMigrationFacade.
php:78
Stack trace:
#0 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php(212): Shopsys\FrameworkBundle\Model\Cart\CartMigrationFacade->onKernelController(Object(Symfony\Compo
nent\HttpKernel\Event\FilterControllerEvent), 'kernel.controll...', Object(Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher))
#1 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php(44): Symfony\Component\EventDispatcher\EventDispatcher->doDispatch(Array, 'kernel.controll...', Objec
t(Symfony\Component\HttpKernel\Event\FilterControllerEvent))
#2 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(139): Symfony\Component\EventDispatcher\EventDispatcher->dispatch('kernel.controll...', Object(Symfony\Componen
[...]

PR #770, opened 2019-01-25
Cart migration: call method on null

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In cart migration facade is not correct call the method because it's called on null
|New feature| No
|BC breaks| No
|Fixes issues| ...

On one SS6 project, this row https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Cart/CartMigrationFacade.php#L79 often throw error because `session` is null

```
[2019-01-25 15:17:21] request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalThrowableError: "Call to a member function get() on null" at .../src/Shopsys/ShopBundle/Model/Cart/CartMigrationFacade.php line 79 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function get() on null at .../src/Shopsys/ShopBundle/Model/Cart/CartMigrationFacade.php:79)"} []
```
I know that method migrate cart after login (i hope). I did a "fix" but i am not really sure whether is correct. Can you check this "fix" is correct?

Thanks

### DIFF
--- a/packages/framework/src/Model/Cart/CartMigrationFacade.php
+++ b/packages/framework/src/Model/Cart/CartMigrationFacade.php
@@ -3,6 +3,7 @@
 namespace Shopsys\FrameworkBundle\Model\Cart;
 
 use Doctrine\ORM\EntityManagerInterface;
+use Shopsys\FrameworkBundle\Component\String\TransformString;
 use Shopsys\FrameworkBundle\Model\Cart\Item\CartItemFactoryInterface;
 use Shopsys\FrameworkBundle\Model\Customer\CustomerIdentifierFactory;
 use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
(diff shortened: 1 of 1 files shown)


henzigo:
This PR closes #744, but i think that we should tell programmers that something is wrong with session.

vitek-rostislav:
Hi @sspooky13, thanks for the PR :+1: 

So, which solution do you guys prefer? I do agree with the expected result of #744 mentioned by @henzigo - i.e. better description of the exception. 

So we will close this PR probably, once yet another development team member confirms that.

sspooky13:
It's only migrating cart items from non-registered user to registered user. I think it's not necessary to study why session doesn't create but as you wish.
I think, in end the result will look similar to this solution.

Ok, when you decide, then you can close this PR.

henzigo:
@sspooky13 Yes, its only migrating cart items. But that method is calling in `error-page-generate` command and people are asking why session is null while deploying. So throw exception or just log that error is better, because people will know what to do and how to fix. We should probably say something like "Hey check your redis server maybe its not running"

sspooky13:
@henzigo, ok then this function is badly implemented but it makes no sense to me why it's called during deploy and generate error page...it should be called only after login user.
I agree that i would suppress error because i think it's login and migrate cart work but if you say, that function is called in deploy, generate error page,... in functions which have nothing to do with functionality then i think it should be called **only** after login

boris-brtan:
hi @vitek-rostislav, what is the final result then?

vitek-rostislav:
Hi, I feel that the problem is still not fixed, even though #815 was merged. I liked the solution in #814 but the PR was closed and not merged. I have written a message to @DavidOstrozlik to find out why :slightly_smiling_face:

vitek-rostislav:
After discussions, I described the problem in a new issue (#955) and I am closing this PR.

PR #779, opened 2019-01-29
[shopsys] upgrade instructions of cart as entity has been updated to be more format consistent

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Wrong formatting of UPGRADE instructions
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -106,54 +106,45 @@ for instance:
                 imageId: \d+
                 additionalIndex: \d+
       ```
-- if you extended `EntityExtensionParentMetadataCleanerEventSubscriber`, review whether you need to implement [fix for translation entities #748](https://github.com/shopsys/shopsys/pull/748) or whether you still need the extension at all
-
-`Cart` has been slightly refactored ((#765)[https://github.com/shopsys/shopsys/pull/765/]), so change your usages appropriately:
- - property `cartItems` has been renamed to `items`
- - method `getCartItemById` has been renamed to `getItemById`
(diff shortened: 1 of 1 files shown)


simara-svatopluk:
We usually put changes from one PR into a single bullet, so I suggest to do the same.

This is not an instruction:
`CartItemRepository class has been removed and all its logic has been moved to new CartRepository`

Also by reading this instruction I don't know what should I do:
```
CartMigrationFacade has been refactored, so change your usages appropriately:
    implementation of methods mergeCurrentCartWithCart and onKernelController has been changed, so revise them if you extended them
```

And there are more instructions that confuses me...

@vitek-rostislav may I ask you for a second review of these instructions? Because you already reviewed them and I have feeling that I would suggest major changes and I don't know if that is what I'm supposed to do.

vitek-rostislav:
Hi @TomasLudvik and @simara-svatopluk. This PR was about formatting only, in such a case I agree with Svaťa's note that it would be nice to have one 1st level bullet that introduces the PR and all the other instructions below as 2nd and deeper levels bullets. Something like:
```
- Cart has been slightly refactored ((#765)[https://github.com/shopsys/shopsys/pull/765/]):
    - `Cart` is now entity, so change your usages appropriately:
        - property `cartItems` has been renamed to `items`
        - method `getCartItemById` has been renamed to `getItemById`
...
```

TBH I am not sure what to do about the other notes @simara-svatopluk - if you have particular objections, please discuss them with @TomasLudvik, suggestions would be great.

TomasLudvik:
Hello @vitek-rostislav and @simara-svatopluk,

only thing I could do was to updated indentation, so I did it. Can you please accept it so it can be merged before release? Thank you

PR #782, opened 2019-02-01
[shopsys] change forms in popup so they work with js validation

### LINKED ISSUE #361: Javascript validation of form in popup window
Reported by @swejzi and @jDolba.
### What is happening
When there are some javascript validation errors while submitting a form, a popup window with list of the errors is displayed. This is standard feature for all forms. But when the form is displayed in popup window, then the popup with errors overwrites the original window with the form. Normally, this is also wanted feature as we do not want to open multiple popups. 

### Expected result
Javascript validation of form in popup is resolved somehow nicely from UX point of view. The solution should be discussed with @LukasHeinz beforehand.

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| JS validation doesn't work with forms in popups, because it closes the original popup to show errors
|New feature| No 
|BC breaks| Yes 
|Fixes issues| closes #361  
|Standards and tests pass| No

When solving #361 I found out that there are 3 problems, that should be solved.

1. We don't have any form in popup with working JS validation
Solution: I changed login form in popup so it is loaded by AJAX and therefore works with JS validation

2. JS validation opens popup with error messages that closes popup with form
Solution: I Added hidden div to popup and changed `showFormErrorsWindow` function in a way that it show errors in the hidden div if popup is open

3. The errors in popup should be under the heading
Solution: I added heading option to window.js so you can have heading at the top of the popup, then you have error messages and then you have the content of the popup

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -172,6 +172,44 @@ for instance:
     - if you have extended `CountryController` revise your changes – `new` and `edit` actions were added
 - if you have extended `Localization` class, you have to add type-hints to extended methods because they were added in the parent class ([#806](https://github.com/shopsys/shopsys/pull/806))
     - if you have extended method `Localization::getAdminLocale()` only to have administration in a different language than english, you can delete it and set parameter `shopsys.admin_locale` in your `parameters.yml` file instead
+- fixed JS validation of forms in popup windows ([#782](https://github.com/shopsys/shopsys/pull/782))
+    - login form in popup is now loaded via AJAX
+    - in `window.js` add options `textHeading = ''` and `cssClassHeading: ''` to `var defaults` like this:
+        ```diff
+            var defaults = {
+                content: '',
+                buttonClose: true,
+                buttonCancel: false,
+                buttonContinue: false,
+                textContinue: Shopsys.translator.trans('Yes'),
+                textCancel: Shopsys.translator.trans('No'),
+        +       textHeading: '',
+                urlContinue: '#',
+                cssClass: 'window-popup--standard',
+                cssClassContinue: '',
+                cssClassCancel: '',
+        +       cssClassHeading: '',
+                closeOnBgClick: true,
+                eventClose: function () {},
+                eventContinue: function () {},
+                eventCancel: function () {},
+                eventOnLoad: function () {}
+            };
+        ```
+        - then add the heading to `$windowContent` and add div with `js-validation-errors` class to every popup.:
+        ```diff
+        -   var $windowContent = $('<div class="js-window-content window-popup__in"></div>').html(options.content);
+        +   var $windowContent = $('<div class="js-window-content window-popup__in"></div>');
+        +   if (options.textHeading !== '') {
+        +       $windowContent.append('<h2 class="' + options.cssClassHeading + '">' + options.textHeading + '</h2>');
+        +   }
+        +   $windowContent.append(
+        +       '<div class="display-none in-message in-message--alert js-window-validation-errors"></div>'
+        +       + options.content
+        +   );
+        ```
+    - *(optional)* change login form so it is loaded by AJAX and works with JS validation and change `Login/windowForm.html.twig`, `Login/loginForm.html.twig` and `header.html.twig` templates
+        - you can change it as it was done in this [commit](https://github.com/shopsys/shopsys/pull/782/commits/0e7cf1e615563e804657d7b96ef335617461236c)
 
 - if you have extended classes from `Shopsys\FrameworkBundle\Model`, `Shopsys\FrameworkBundle\Component` or `Shopsys\FrameworkBundle\DataFixtures\Demo` namespace ([#788](https://github.com/shopsys/shopsys/pull/788))
     - you need to adjust extended methods and fields to `protected` visibility because all `private` visibilities from these namespaces were changed to `protected`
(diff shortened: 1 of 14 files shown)


project-base/src/Shopsys/ShopBundle/Resources/scripts/frontend/login.js
```diff
@@ -3,9 +3,30 @@
...
+        };
+
+        function showWindow (event) {
+            Shopsys.ajax({
+                url: $(this).data('url'),
+                type: 'POST',
+                success: function (data) {
+                    var $window = Shopsys.window({
+                        content: data,
+                        textHeading: Shopsys.translator.trans('Login')
```
PetrHeinz: The feature you're using here is not yet implemented (it belongs to the next commit).

project-base/src/Shopsys/ShopBundle/Form/Front/Login/LoginFormType.php
```diff
@@ -51,8 +53,9 @@ public function getBlockPrefix()
      */
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefaults([
-            'attr' => ['novalidate' => 'novalidate'],
-        ]);
+        $resolver
+            ->setDefaults([
+                'attr' => ['novalidate' => 'novalidate'],
+            ]);
```
PetrHeinz: No change in this file was actually necessary.

project-base/src/Shopsys/ShopBundle/Resources/scripts/frontend/window.js
```diff
@@ -165,6 +171,7 @@
         }
 
         Shopsys.register.registerNewContent($window);
+        $window.load();
```
PetrHeinz: To be honest, I'm not sure why are you calling the method. I'm not even sure which method it actually is, I found 2 `.load()` methods in jQuery (https://api.jquery.com/load/ and deprecated https://api.jquery.com/load-event/), both require at least one argument.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -159,6 +159,27 @@ for instance:
     - if you have extended `CountryController` revise your changes – `new` and `edit` actions were added
 - if you have extended `Localization` class, you have to add type-hints to extended methods because they were added in the parent class ([#806](https://github.com/shopsys/shopsys/pull/806))
     - if you have extended method `Localization::getAdminLocale()` only to have administration in a different language than english, you can delete it and set parameter `shopsys.admin_locale` in your `parameters.yml` file instead
+- changed JS validation of forms in popup and login form in popup is now get by AJAX ([#782](https://github.com/shopsys/shopsys/pull/782))
```
TomasLudvik:
```suggestion
- fixed JS validation of forms in popup windows ([#782](https://github.com/shopsys/shopsys/pull/782))
    - login form in popup is now loaded via AJAX
```

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -159,6 +159,27 @@ for instance:
...
+    - in `window.js` add options `textHeading = ''` and `cssClassHeading: ''` to `var defaults` and then add it to `$windowContent` and add div with `js-validation-errors` class to every popup.
```
TomasLudvik: I would change this to: 
This javascript files has been changed, please update them appropriately:
- src/Shopsys/ShopBundle/Resources/scripts/frontend/login.js
- src/Shopsys/ShopBundle/Resources/scripts/frontend/window.js

For me is better to see diff on Github than this instructions

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -159,6 +159,27 @@ for instance:
...
+                      $windowContent.append('<h2 class="' + options.cssClassHeading + '">' + options.textHeading + '</h2>');
+                  }
+
+                  $windowContent.append(
+                      '<div class="display-none in-message in-message--alert js-window-validation-errors"></div>'
+                      + options.content
+                  );
+            ```
+- *(optional)* change login form so it is loaded by AJAX and works with JS validation
+    - you can change it as it was done in this [commit](https://github.com/shopsys/shopsys/pull/782/commits/bb6dadfce1d588aa3e06a8ba7b117d3de7afc1ce)
```
TomasLudvik: You did not mention changes in templates

packages/framework/src/Resources/scripts/common/validation/customizeBundle.js
```diff
@@ -441,14 +442,26 @@
 
     Shopsys.validation.showFormErrorsWindow = function (container) {
         var $formattedFormErrors = Shopsys.validation.getFormattedFormErrors(container);
+        var $window = $('#js-window');
+
+        if ($window.length === 0) {
+            Shopsys.window({
+                content:
+                    '<div class="text-left">'
```
TomasLudvik: I would refactor this part of code, because it is identical as code few lines bellow and it violation with DRY principle

project-base/src/Shopsys/ShopBundle/Resources/views/Front/Content/Login/loginForm.html.twig
```diff
@@ -15,9 +15,24 @@
 
         <h1>{{ 'Login'|trans }}</h1>
 
-        {{ form_row(form.email, {label: 'E-mail'|trans} ) }}
 
-        {{ form_row(form.password, {label: 'Password'|trans} ) }}
+        <dl class="form-line">
```
TomasLudvik: Why are these changes necessary

DavidOstrozlik: Because form attribute `multipleForm: true` changes ID's only for form labels and form rows but not for form_errors so we need remove the form errors so JS validation in login popup works on /login page

docs/upgrade/UPGRADE-unreleased.md
DavidOstrozlik: The change or login.js is mentioned in the optional point below and It is only few lines so I think it's more readable here

TomasLudvik: We have discussed it and you want to make it more straightforward

PR #788, opened 2019-02-04
[shopsys] added new coding standard: forbidden private visibility in namespace

### LINKED ISSUE #664: automatic fixer for visibility in extendable classes
### What is happening
According to our [coding standards](https://github.com/shopsys/shopsys/blob/7998b139f2abe8ab7d3c94cb1208192ba327509f/docs/contributing/coding-standards.md), there should be only public or protected functions, properties, and constants in Entities, Factories, Controllers, Facades, and Repositories in framework package and plugins.

There is no automatic checker for this, and we sometimes create private methods or properties that make problems when extending the classes in projects. Fixing it manually is sometimes necessary, see #650.

### Expected result

There should be an automatic fixer (only relevant in monorepo), that would overwrite `T_PRIVATE` tokens by `T_PROTECTED` in all classes with suffix `Factory`, `Controller`, `Facade`, or `Repository`, or classes annotated with `@ORM\Entity`. This should be applied to all classes in `packages/framework/src` and `packages/product-feed-*/src`.

Due to the specificity of this rule, I'm not sure whether this fixer should be even a part of our `shopsys/coding-standards` repository... If so, it should be either disabled by default, or configurable (so the paths are not hard-coded).

### LINKED ISSUE #726: check relevance of private visibility for parameters and methods in Product\Pricing classes
### What is happening

during implementation of discount for customer we needed to extend classes with private methods and properties https://github.com/shopsys/demoshop/pull/31

### Expected result

decision or result whether it is possible to have it at least protected

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Enabled and forces not having any private fields and methods in certain namespaces.
|New feature| No 
|BC breaks| Yes 
|Fixes issues| closes #726, closes #664 

According to our coding standards, there should be only public or protected functions, properties, and constants in Entities, Factories, Controllers, Facades, and Repositories in framework package and plugins.

Following namespaces are checked:
- `Shopsys\FrameworkBundle\Model`
- `Shopsys\FrameworkBundle\Component`
- `Shopsys\FrameworkBundle\Controller`
- `Shopsys\FrameworkBundle\DataFixtures\Demo`
- `Shopsys\ProductFeed\GoogleBundle\Model`
- `Shopsys\ProductFeed\HeurekaBundle\Model`
- `Shopsys\ProductFeed\HeurekaDeliveryBundle\Model`
- `Shopsys\ProductFeed\ZboziBundle\Model`

### DIFF
--- a/docs/contributing/coding-standards.md
+++ b/docs/contributing/coding-standards.md
@@ -87,12 +87,5 @@ Besides the rules that are checked by automatic tools, we have few rules for whi
     }
     ```
 - Entities have to be created by factories. Only allowed exception are `*Translation` entities that are created by their owner entity.
-- Visibility of all properties, methods and constants of following classes must be protected or public to enable extensibility.
-  This is valid for packages in which we plan extensibility - Framework, ProductFeed\*
-    - Entities
-    - Factories
-    - Controllers
-    - Facades
-    - Repositories
 - Framework entities must not use [Doctrine inheritance mapping](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/inheritance-mapping.html)
   because it causes problems during entity extension. Such problem with `OrderItem` was resolved during [making OrderItem extendable #715](https://github.com/shopsys/shopsys/pull/715).
--- a/packages/coding-standards/src/CsFixer/ForbiddenPrivateVisibilityFixer.php
+++ b/packages/coding-standards/src/CsFixer/ForbiddenPrivateVisibilityFixer.php
@@ -1,0 +1,183 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\CsFixer;
+
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Fixer\ConfigurableFixerInterface;
+use PhpCsFixer\Fixer\DefinedFixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\FixerDefinitionInterface;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use Shopsys\CodingStandards\Exception\NamespaceNotFoundException;
+use SplFileInfo;
+
+final class ForbiddenPrivateVisibilityFixer implements DefinedFixerInterface, ConfigurableFixerInterface
+{
+    private const OPTION_ANALYZED_NAMESPACE = 'analyzed_namespaces';
+
+    private $analyzedNamespaces = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null): void
+    {
+        if ($configuration !== null) {
+            $this->analyzedNamespaces = $this->extractNamespaces($configuration);
+        }
+    }
+
+    /**
+     * @param array $configuration
+     * @return array
+     */
+    private function extractNamespaces(array $configuration): array
+    {
+        if (!array_key_exists(self::OPTION_ANALYZED_NAMESPACE, $configuration)) {
+            return [];
+        }
+
+        if (!is_array($configuration[self::OPTION_ANALYZED_NAMESPACE])) {
+            throw new InvalidFixerConfigurationException($this->getName(), 'Namespace configuration has to be an array');
+        }
+
+        return $configuration[self::OPTION_ANALYZED_NAMESPACE];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition(): FixerDefinitionInterface
+    {
+        return new FixerDefinition(
+            'Properties and methods should be public or protected in defined namespace',
+            [
+                new CodeSample(
+                    '<?php
+namespace Some\Namespace;
+class SomeClass
+{
+private $property;
+}'
+                ),
+                new CodeSample(
+                    '<?php
+namespace Some\Namespace;
+class SomeClass
+{
+private function method()
+{
+    ...
+}
+}'
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens): bool
+    {
+        return $tokens->isAllTokenKindsFound([T_PRIVATE]) && $this->checkNamespace($tokens);
+    }
+
+    /**
+     * @param \PhpCsFixer\Tokenizer\Tokens $tokens
+     * @return bool
+     */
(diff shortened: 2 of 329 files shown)


packages/coding-standards/src/CsFixer/ForbiddenPrivateVisibilityFixer.php
```diff
@@ -0,0 +1,141 @@
...
+    private const OPTION_ANALYZED_NAMESPACE = 'analyzed_namespaces';
+
+    private $analyzedNamespaces = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null): void
+    {
+        if (array_key_exists(self::OPTION_ANALYZED_NAMESPACE, $configuration) && is_array($configuration[self::OPTION_ANALYZED_NAMESPACE])) {
```
PetrHeinz: This opens up to misconfiguration, if the array key exists it should always be an array, I would rather throw an exception is such a case.

simara-svatopluk:
Good job with the fixer :-)

simara-svatopluk:
One last question - who was involved in decision what namespaces have to be `protected` only?

docs/contributing/coding-standards.md
```diff
@@ -94,5 +94,10 @@ Besides the rules that are checked by automatic tools, we have few rules for whi
     - Controllers
     - Facades
     - Repositories
+    - whole namespaces
```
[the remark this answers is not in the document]
grossmannmartin: But one of the acceptance criteria is "we have the rule documented" and this seems to me like proper documentation.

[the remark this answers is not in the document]
simara-svatopluk: You are right.. confusing situation for me. I suggest to contact our product owner because when we have a sniffer/fixer, then we don't document it in "coding standards"

Another issue is that the documentation is located in section "Standards that are not checked automatically" - and that is not true ;-)

grossmannmartin:
> One last question - who was involved in decision what namespaces have to be `protected` only?

I started from task definition in user story (= `Shopsys\FrameworkBundle\Model` and `Shopsys\FrameworkBundle\Component`), I believe it was consulted internally. 

Then I added DataFixtures (because without them, we have to have exception for ie. AbstractReferenceFixture in Components)

And I added Model namespace from product-feed-* packages thanks to #664 and because it makes sense to me like this (and this rule is already applied there)

packages/coding-standards/src/CsFixer/ForbiddenPrivateVisibilityFixer.php
```diff
@@ -0,0 +1,141 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens): bool
+    {
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_PRIVATE]) && $this->checkNamespace($tokens);
```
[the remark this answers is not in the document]
grossmannmartin: No, we don't need to 👍

packages/coding-standards/easy-coding-standard.yml
```diff
@@ -18,6 +18,12 @@ services:
 
     # Shopsys Checkers
     Shopsys\CodingStandards\CsFixer\ForbiddenDumpFixer: ~
+    Shopsys\CodingStandards\CsFixer\ForbiddenPrivateVisibilityFixer:
+        analyzed_namespaces:
+            - Shopsys\FrameworkBundle\Model
```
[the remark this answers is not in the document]
grossmannmartin: You're right. I moved it, but in monorepo have to be still a whole list of namespaces. But there I believe it's ok.

docs/contributing/coding-standards.md
```diff
@@ -94,5 +94,10 @@ Besides the rules that are checked by automatic tools, we have few rules for whi
...
+        - `Shopsys\FrameworkBundle\Model`
+        - `Shopsys\FrameworkBundle\Component`
+        - `Shopsys\FrameworkBundle\Controller`
+        - `Shopsys\FrameworkBundle\DataFixtures\Demo`
```
PetrHeinz: This whole section is under the heading *Standards that are not checked automatically*. As it is checked automatically, I would rather remove this bullet altogether (starting by `- Visibility of all properties, methods and constants of`).

packages/coding-standards/easy-coding-standard.yml
```diff
@@ -18,6 +18,7 @@ services:
 
     # Shopsys Checkers
     Shopsys\CodingStandards\CsFixer\ForbiddenDumpFixer: ~
+    Shopsys\CodingStandards\CsFixer\ForbiddenPrivateVisibilityFixer: ~
```
simara-svatopluk: Yep, the last nitpick note - if I understand correctly, the Fixer does nothing because it is not configured, right? If so, I would prefer to remove this line

docs/contributing/coding-standards.md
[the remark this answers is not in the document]
grossmannmartin: Yes, I didn't notice it's in that section (Standards that are not checked automatically).
It doesn't make sense this way. For now, I removed the note and I will discuss it with product owner.

docs/contributing/coding-standards.md
grossmannmartin: Right, I didn't notice that. As I said in https://github.com/shopsys/shopsys/pull/788#discussion_r254157257, I deleted it and I will discuss this topic with product owner.

packages/coding-standards/easy-coding-standard.yml
grossmannmartin: You're right. Removed.

simara-svatopluk:
@grossmannmartin Please review commit messages - 2. and 3. commits have wrong formatting in sub bullets.
Also please add note about removed documentation to 2. commit

I'll keep testing :-)

packages/product-feed-google/easy-coding-standard.yml
```diff
@@ -1,6 +1,15 @@
 imports:
     - { resource: '%vendor_dir%/shopsys/coding-standards/easy-coding-standard.yml', ignore_errors: true  }
 
+services:
+    forbidden_private_visibility_fixer.google:
```
simara-svatopluk: I would like to have here a full package name to set some kind of standard (also in other product-feed-* packages)

simara-svatopluk:
Reviewed again, tested. Ok!

simara-svatopluk:
Reviewed again, tested, works!

simara-svatopluk:
Reviewed again, tested, works!

I have only one comment left - naming private services.

packages/product-feed-google/easy-coding-standard.yml
grossmannmartin: seems reasonable 👍

PR #792, opened 2019-02-05
[framework] added ability to use distinct templates for the other domains

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to enable project developers implementation of different designs on multiple domains
|New feature| Yes 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -42,12 +42,14 @@ Add:
 * [Modifying a Template in Administration](./cookbook/modifying-a-template-in-administration.md)
 * [Adding Ajax Load More Button into Pagination](./cookbook/adding-ajax-load-more-button-into-pagination.md)
 * [Implementing a basic data import](./cookbook/basic-data-import.md)
+* [Creating a Multidomain Design](./cookbook/creating-a-multidomain-design.md)
 
 ## Functional documentation
 * [Behavior of Product Variants](./functional/behavior-of-product-variants.md)
 * [Product Visibility and Exclude from Sale](./functional/product-visibility-and-exclude-from-sale.md)
 
 ## Frontend
+* [Design implementation and Customization](./frontend/design-implementation-and-customization.md)
 * [Introduction to LESS](./frontend/introduction-to-less.md)
 * [Frontend Troubleshooting](./frontend/frontend-troubleshooting.md)
 
--- a/packages/framework/src/Component/Domain/Multidomain/Twig/FilesystemLoader.php
+++ b/packages/framework/src/Component/Domain/Multidomain/Twig/FilesystemLoader.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace Shopsys\FrameworkBundle\Component\Domain\Multidomain\Twig;
+
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Symfony\Bundle\TwigBundle\Loader\FilesystemLoader as BaseFilesystemLoader;
+use Symfony\Component\Config\FileLocatorInterface;
+use Symfony\Component\Templating\TemplateNameParserInterface;
+
+class FilesystemLoader extends BaseFilesystemLoader
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Domain\Domain|null
+     */
+    protected $domain;
+
+    /**
+     * @param \Symfony\Component\Config\FileLocatorInterface $locator
+     * @param \Symfony\Component\Templating\TemplateNameParserInterface $parser
+     * @param string|null $rootPath
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain|null $domain
+     */
+    public function __construct(
+        FileLocatorInterface $locator,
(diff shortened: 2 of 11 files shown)


vitek-rostislav:
Hi @DavidOstrozlik, thanks :+1: 

> Shouldn't we put the parameter to `domains.yml` so User know that it can be there?

I would say that the user knows it from the docs. Or by looking at `DomainConfig` class. It seems weird to me setting the attribute to null there. Should I put it there for both domains? You can see that we do set `styles_directory` for the first domain as well.

> Is `creating-a-multidomain-design.md` really a cookbook, shouldn't it be somewhere else?

I would say that it is a cookbook, it contains step by step instructions.

packages/framework/src/Component/Domain/Multidomain/Twig/FilesystemLoader.php
```diff
@@ -0,0 +1,80 @@
...
+                return parent::findTemplate($multidesignTemplateName);
+            } catch (\Twig_Error_Loader $loaderException) {
+                if (strpos($loaderException->getMessage(), 'Unable to find template') !== 0) {
+                    $message = sprintf('Unexpected exception when trying to load multidesign template `%s`', $multidesignTemplateName);
+                    throw new \Twig_Error_Loader($message, -1, null, $loaderException);
+                }
+            }
+        }
+
+        return parent::findTemplate($templateName);
```
TomasLudvik: findMultidesignTemplate at the end finds not mutlidesignTemplate :-/ this should not happen in my opinion

packages/framework/src/Component/Domain/Multidomain/Twig/FilesystemLoader.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @param string $templateName
+     * @return string
+     */
+    protected function findMultidesignTemplate($templateName)
+    {
+        $designId = $this->domain->getDesignId();
+        if ($designId !== null) {
+            $multidesignTemplateName = preg_replace('/^(.*)(\.[^\.]*\.twig)/', '$1.' . $designId . '$2', $templateName);
```
TomasLudvik:
```suggestion
            $multidesignTemplateName = preg_replace('/^(.*)(\.[^\.]*\.twig)$/', '$1.' . $designId . '$2', $templateName);
```

this said Petr

packages/framework/src/Component/Domain/Multidomain/Twig/FilesystemLoader.php
vitek-rostislav: Hi, what do you suggest? I wanted to name the method "tryToFindMultidesign..." but it seemed too complicated. It is just a fallback, I do not see it as much important here. More on that, it is just a protected helper method...

packages/framework/src/Component/Domain/Multidomain/Twig/FilesystemLoader.php
```diff
@@ -0,0 +1,80 @@
...
+     * When exists a template with filename.{designId}.html.twig, then it is automatically used
+     * on domain with this {designId} whenever template named filename.html.twig is on input
+     *
+     * @inheritdoc
+     */
+    protected function findTemplate($template, $throw = true)
+    {
+        $templateName = (string)$template;
+        if (strpos($templateName, '@ShopsysShop/Front/') === 0) {
+            $this->assertDomainDependency();
```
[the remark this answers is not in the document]
vitek-rostislav: ok, the assertion moved to constructor

docs/index.md
```diff
@@ -42,12 +42,14 @@ Add:
...
 * [Adding Ajax Load More Button into Pagination](./cookbook/adding-ajax-load-more-button-into-pagination.md)
 * [Implementing a basic data import](./cookbook/basic-data-import.md)
+* [Creating a Multidomain Design](./cookbook/creating-a-multidomain-design.md)
 
 ## Functional documentation
 * [Behavior of Product Variants](./functional/behavior-of-product-variants.md)
 * [Product Visibility and Exclude from Sale](./functional/product-visibility-and-exclude-from-sale.md)
 
 ## Frontend
+* [Design implementation and Customization](frontend/design-implementation-and-customization.md)
```
DavidOstrozlik:
```suggestion
* [Design implementation and Customization](./frontend/design-implementation-and-customization.md)
```

packages/framework/src/Component/Domain/Multidomain/Twig/FilesystemLoader.php
DavidOstrozlik: I think this method would be better if it would return MultidesignTemplate|null because of the naming find*
Or if you don't want to refactor it, I would suggest renaming the method

vitek-rostislav: ok guys, you convinced me :laughing: :+1:

PR #803, opened 2019-02-07
Upgrade notes are extended and fixed

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| we have some feedback from upgraders about upgrade instructions and about need for upgrade notes improvement and we decided to make upgrade notes better.
|New feature| Yes 
|BC breaks| Yes 
|Fixes issues|  

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -14,9 +14,8 @@ Follow instructions in a [monorepo upgrade guide](docs/contributing/upgrading-mo
 * upgrade only your composer dependencies and follow instructions
 * check all instructions in all sections, any of them could be relevant for you
 * if you want update your project with the changes from [shopsys/project-base],
-    you can follow the *(optional)* instructions or cherry-pick from the repository whatever is relevant for you but we do not recommend rebasing or merging everything because the changes might not be compatible with your project as it probably evolves in time
-* instructions marked as *(optional)* are not vital, but could be helpful,
-    so we recommend to perform them as well during upgrading as it might ease your work in the future
+    you should follow the *(low priority)* instructions or cherry-pick from the repository whatever is relevant for you but we do not recommend rebasing or merging everything because the changes might not be compatible with your project as it probably evolves in time
+* instructions marked as *(low priority)* are not vital and can be done last and we recommend to perform them as well during upgrading as it might ease your work in the future
 * upgrade locally first. After you fix all issues caused by the upgrade, commit your changes and then continue with upgrading application on a server
 * upgrade one version at a time:
     * Start with a working application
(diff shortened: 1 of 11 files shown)


PetrHeinz:
I don't thing that removing the *(optional)* tag is a good idea, as it may help developers upgrading their projects at least in prioritizing the step during the upgrade. When upgrading, I'd definitely do all the non-optional steps first so that the project is functional, merge, and then do the rest.

docs/upgrade/UPGRADE-v7.0.0-beta3.md
```diff
@@ -30,16 +26,22 @@ There you can find links to upgrade notes for other versions too.
                 - shopsys-network
         ```
     - remove smtp row from `extra_hosts` ([#530](https://github.com/shopsys/shopsys/pull/530))
+    - check whether there exists `depends_on` property between websercer and php-fpm services since this is needed for functional volume mounting (see [#598](https://github.com/shopsys/shopsys/pull/598))
+- update `docker-compose.yml` file
+    - update php-fpm build configuration ([#533](https://github.com/shopsys/shopsys/pull/533))
+        - change `context` to current directory (`.`)
+        - add `dockerfile` directive with value `docker/php-fpm/Dockerfile`
+        - add `target` environment (`development`)
```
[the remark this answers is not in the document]
grossmannmartin: Oh, nevermind. I thought `docker-compose-prod.yml` is `docker-compose.yml.dist`. My mistake

UPGRADE.md
```diff
@@ -12,8 +12,8 @@ Follow instructions in a [monorepo upgrade guide](docs/contributing/upgrading-mo
 * upgrade only your composer dependencies and follow instructions
 * check all instructions in all sections, any of them could be relevant for you
 * if you want update your project with the changes from [shopsys/project-base],
-    you can follow the *(optional)* instructions or cherry-pick from the repository whatever is relevant for you but we do not recommend rebasing or merging everything because the changes might not be compatible with your project as it probably evolves in time
-* instructions marked as *(optional)* are not vital, but could be helpful,
+    you can follow the *(low priority)* instructions or cherry-pick from the repository whatever is relevant for you but we do not recommend rebasing or merging everything because the changes might not be compatible with your project as it probably evolves in time
```
grossmannmartin: I would say `you should follow`

UPGRADE.md
```diff
@@ -12,8 +12,8 @@ Follow instructions in a [monorepo upgrade guide](docs/contributing/upgrading-mo
...
+* instructions marked as *(low priority)* are not vital, but could be helpful,
```
grossmannmartin: * instructions marked as *(low priority)* are not vital and can be done last,
  and we recommend to perform them as well during upgrading as it might ease your work in the future

or something like that. To be clear, we really recommending doing them.

boris-brtan:
> I don't thing that removing the _(optional)_ tag is a good idea, as it may help developers upgrading their projects at least in prioritizing the step during the upgrade. When upgrading, I'd definitely do all the non-optional steps first so that the project is functional, merge, and then do the rest.
> 
> Maybe a clarification on this topic would suffice, as our [Upgrade notes](https://github.com/shopsys/shopsys/blob/v7.0.0-beta5/UPGRADE.md) mention the tag only briefly:
> 
> > * instructions marked as _(optional)_ are not vital, but could be helpful, so we recommend to perform them as well during upgrading as it might ease your work in the future

changed to low priority

docs/upgrade/UPGRADE-v7.0.0-beta3.md
```diff
@@ -226,7 +228,7 @@ There you can find links to upgrade notes for other versions too.
             'route_label' => $product->getMainVariant()->getName(),
         ]);
         ```
-- *(optional)* rename Database tests to Functional tests
+- rename Database tests to Functional tests
```
grossmannmartin: It should be `low priority`, shouldn't?

It was probably removed along with phing modifications where functional tests are mentioned. 
This exactly is, from my point of view, non-vital instruction and we made it vital just because of the phing target name.

Wouldn't be better to mention this rename along with instructions for phing targets near line 120 and keep this `low priority`?

PR #806, opened 2019-02-07
[framework] now it is possible to set administration locale in parameters

### LINKED ISSUE #589: It is quite hard to change admin locale
### What is happening
When I want to change admin locale, I need to overwrite `Localization::getAdminLocale()` (see https://github.com/shopsys/shopsys/blob/master/docs/introduction/how-to-set-up-domains-and-locales.md#36-locale-in-administration). To me, it means many unnecessary steps for such a tiny change.

### Expected result

It should be possible to set administration locale without the need of extending a framework class. E.g. using `parameters.yml` or any other config file for that purpose would be nice

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Administration locale was possible to change only by inherit Localization class and override method getAdminLocale(). That is too many steps for such an easy change. So now the developer can set parameter `shopsys.admin_locale` to the desired locale. If selected locale is not one of domains registered locales, an exception is thrown because in that case application behaved funny.
|New feature| Yes 
|BC breaks| No 
|Fixes issues| closes #589 

### DIFF
--- a/docs/introduction/faq-and-common-issues.md
+++ b/docs/introduction/faq-and-common-issues.md
@@ -67,7 +67,8 @@ See how to install Shopsys Framework in production and how to proceed when deplo
 
 ## How to set up the administration with a different locale/language (e.g. Czech)?
 The administration uses `en` locale by default.
-If you want to switch it to the another locale, override the method `getAdminLocale()` of the class `Shopsys\FrameworkBundle\Model\Localization\Localization`.
+If you want to switch it to the another locale, set a parameter `shopsys.admin_locale` in your `parameters.yml` configuration.
+However, the selected locale has to be one of registered domains locale.
 This scenario is described in more detail in the tutorial [How to Set Up Domains and Locales (Languages)](./how-to-set-up-domains-and-locales.md#36-locale-in-administration).
 
 ## What are the differences between "listable", "sellable", "offered" and "visible" products?
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -157,6 +157,7 @@ for instance:
     - if you have extended `CountryGridFactory`, revise your changes because class changed its namespace
     - if you have extended `CountryFormType`, revise your changes – new fields are available
     - if you have extended `CountryController` revise your changes – `new` and `edit` actions were added
+- if you have extended `Localization` class, you have to add type-hints to extended methods because they were added in the parent class ([#806](https://github.com/shopsys/shopsys/pull/806))
 
 - if you have extended classes from `Shopsys\FrameworkBundle\Model`, `Shopsys\FrameworkBundle\Component` or `Shopsys\FrameworkBundle\DataFixtures\Demo` namespace ([#788](https://github.com/shopsys/shopsys/pull/788))
     - you need to adjust extended methods and fields to `protected` visibility because all `private` visibilities from these namespaces were changed to `protected`
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -158,6 +158,7 @@ for instance:
     - if you have extended `CountryFormType`, revise your changes – new fields are available
     - if you have extended `CountryController` revise your changes – `new` and `edit` actions were added
 - if you have extended `Localization` class, you have to add type-hints to extended methods because they were added in the parent class ([#806](https://github.com/shopsys/shopsys/pull/806))
+    - if you have extended method `Localization::getAdminLocale()` only to have administration in a different language than english, you can delete it and set parameter `shopsys.admin_locale` in your `parameters.yml` file instead
 
 - if you have extended classes from `Shopsys\FrameworkBundle\Model`, `Shopsys\FrameworkBundle\Component` or `Shopsys\FrameworkBundle\DataFixtures\Demo` namespace ([#788](https://github.com/shopsys/shopsys/pull/788))
     - you need to adjust extended methods and fields to `protected` visibility because all `private` visibilities from these namespaces were changed to `protected`
--- a/packages/framework/src/Model/Localization/Localization.php
+++ b/packages/framework/src/Model/Localization/Localization.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Shopsys\FrameworkBundle\Model\Localization;
 
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
@@ -38,7 +40,7 @@ class Localization
     protected $domain;
 
     /**
-     * @var array
+     * @var string[]|null
      */
     protected $allLocales;
 
@@ -53,7 +55,7 @@ class Localization
     /**
      * @return string
      */
-    public function getLocale()
+    public function getLocale(): string
     {
         return $this->domain->getLocale();
     }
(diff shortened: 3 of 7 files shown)


docs/introduction/faq-and-common-issues.md
```diff
@@ -67,7 +67,8 @@ See how to install Shopsys Framework in production and how to proceed when deplo
 
 ## How to set up the administration with a different locale/language (e.g. Czech)?
 The administration uses `en` locale by default.
-If you want to switch it to the another locale, override the method `getAdminLocale()` of the class `Shopsys\FrameworkBundle\Model\Localization\Localization`.
+If you want to switch it to the another locale, override a parameter `shopsys.admin_locale` in your `parameters.yml` configuration.
```
PetrHeinz: Just a grammar tweak: I think you can either "override the parameter" (as the parameter must be there to be overridden) or "set a parameter" (as it's a new parameter in that file):
```suggestion
If you want to switch it to the another locale, set a parameter `shopsys.admin_locale` in your `parameters.yml` configuration.
```

packages/framework/src/Model/Localization/Localization.php
```diff
@@ -63,7 +71,15 @@ public function getLocale()
      */
     public function getAdminLocale()
     {
-        return 'en';
+        if ($this->allLocales === null) {
+            $this->loadAllLocalesFromAllDomains();
```
PetrHeinz: A better approach imho would be:

```php
$allLocales = $this->getLocalesOfAllDomains();

 if (!in_array($this->adminLocale, $allLocales, true)) {
    throw new AdministrationLocaleNotFoundException($this->adminLocale, $allLocales);
}

return $this->adminLocale;
```

as it doesn't force you to assume that the new method will set values to `$this->allLocales`

packages/framework/src/Model/Localization/Localization.php
```diff
@@ -72,15 +88,22 @@ public function getAdminLocale()
...
-            foreach ($this->domain->getAll() as $domainConfig) {
-                $this->allLocales[$domainConfig->getLocale()] = $domainConfig->getLocale();
-            }
+            $this->loadAllLocalesFromAllDomains();
         }
 
         return $this->allLocales;
     }
 
+    private function loadAllLocalesFromAllDomains(): void
```
PetrHeinz: should be protected to avoid possible merge problems with #788

packages/framework/src/Model/Localization/Localization.php
```diff
@@ -42,12 +43,19 @@ class Localization
...
+     * @var string
+     */
+    protected $adminLocale;
+
     /**
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @param string $adminLocale
      */
-    public function __construct(Domain $domain)
+    public function __construct(Domain $domain, $adminLocale = 'en')
```
PetrHeinz: Consider type-hinting it with `string`. Why having a default value duplicate with setting in `services.yml`?

packages/framework/src/Model/Localization/Exception/AdministrationLocaleNotFoundException.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function __construct(string $adminLocale, array $possibleLocales, Exception $previous = null)
+    {
+        $message = sprintf(
+            'You tried to use administration in "%s" locale, but you have registered only ["%s"].'
+            . ' Either register "%s" as a locale with some domain or use one of ["%s"] as administration locale.',
+            $adminLocale,
+            implode('","', $possibleLocales),
+            $adminLocale,
+            implode('","', $possibleLocales)
```
PetrHeinz: You can avoid the duplication of parameters using argument swapping (see [docs](http://php.net/manual/en/function.sprintf.php)):
```php
sprintf(
  `You tried to use administration in "%1$s" locale, but you have registered only ["%2$s"]. Either register "%1$s" as a locale with some domain or use one of ["%2$s"] as administration locale.',
  $adminLocale,
  implode('","', $possibleLocales)
)
```

packages/framework/src/Model/Localization/Exception/AdministrationLocaleNotFoundException.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Shopsys\FrameworkBundle\Model\Localization\Exception;
+
+use Exception;
+use RuntimeException;
+
+class AdministrationLocaleNotFoundException extends RuntimeException implements LocalizationException
+{
+    /**
+     * @param string $adminLocale
+     * @param array $possibleLocales
```
PetrHeinz:
```suggestion
     * @param string[] $possibleLocales
```

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -154,6 +154,7 @@ for instance:
     - if you have extended `CountryGridFactory`, revise your changes because class changed its namespace
     - if you have extended `CountryFormType`, revise your changes – new fields are available
     - if you have extended `CountryController` revise your changes – `new` and `edit` actions were added
+- if you override method `Localization::getAdminLocale()` to have administration in a different language than english, you can delete it and set parameter `shopsys.admin_locale` in your `parameters.yml` file instead ([#806](https://github.com/shopsys/shopsys/pull/806))
```
PetrHeinz: Without doing this change, the project would still work as intended, so I'd use the *(optional)* tag.

If would have to required if you'd remove the default value of the second argument in the constructor. (As the extending service with overridden method would not inherit the `arguments` configuration from framework's `services.yml`.)

PetrHeinz:
@grossmannmartin BTW I think this PR should be titled *[framework] now it is possible*, as all the code changes were made in shopsys/framework (and were only documented outside of it)

packages/framework/src/Model/Localization/Exception/AdministrationLocaleNotFoundException.php
grossmannmartin: Oh, that's nice. Thanks

packages/framework/src/Model/Localization/Localization.php
grossmannmartin: I thought about it but didn't want to assign into the local variable value that already is in the property.
But it will be much clearer in a way you suggest.

packages/framework/src/Model/Localization/Localization.php
grossmannmartin: Because I wanted to avoid a bc break for people that might already extend Localization class and call `parent::__construct`.

docs/upgrade/UPGRADE-unreleased.md
grossmannmartin: According to #803, I didn't want to introduce new optional instruction and assume, that "if you ... you can" phrasing is sufficient.

packages/framework/src/Model/Localization/Localization.php
grossmannmartin: Method removed while solving the previous comment.

packages/framework/src/Model/Localization/Localization.php
PetrHeinz: The way I see it is that `$this->allLocales` is in fact `string[]|null` but you need `string[]` in your `$allLocales` in your local variable, and `$this->getLocalesOfAllDomains()` can provide it to you.

packages/framework/src/Model/Localization/Localization.php
PetrHeinz: If you're trying to avoid BC breaks you must not add type-hints for return values.

grossmannmartin: Right. As we consult, I added type-hints everywhere in this class and removed this default value as we're still in beta and we want to make our codebase cleaner as much as possible before stable. 
Upgrade notes were updated accordingly.

PR #755, opened 2019-01-21
[shopsys] update npm packages to latest version

topolanekmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This pull request update npm packages to its latest version. 
|New feature| No
|BC breaks| No
|Fixes issues| ...
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #809, opened 2019-02-08
[project-base] Reload page after promocode is set

topolanekmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Implemented reload function after the promo code is set as a fix of npm upgrade implementation in #755 
|New feature| No
|BC breaks| No
|Fixes issues| ...


boris-brtan:
this would need some prelinks and maybe some check whether the removal is not already in UPGRADE.md

Issue #742, opened 2019-01-15
There are no rules for phing targets


DavidOstrozlik:
### What is happening
We don't have any rules for phing targets so there are lots of them and they need some revision.
Some targets do almost the same thing but they have different names/nomenclature.
We don't know what targets are used and when you should use/create a phing target and when or if you should use/create symfony command.  

### Expected result

Phing targets have been revised so we know what phing targets we have and why we have them, we have some structure of phing targets (every part of bigger target should have its own target or not eg. `clean` should run `clean-dirs`, `clean-redis`,....).
We have some rules so all phing targets follow the same rules about nomenclature (`tests` runs `tests-unit` and other test targets vs `clean` that cleans up directories with cache and scripts but doesn't run other cleaning targets eg. `clean-redis`).

TomasVotruba:
Great to open this! I get lost all the time :) 

Inspiration for some simple targets https://blog.martinhujer.cz/have-you-tried-composer-scripts/
Also Travis might help to clarify what is "simple" target.

vitek-rostislav:
I think we need a user story for that, could you please create one, @DavidOstrozlik?

boris-brtan:
THX

vitek-rostislav:
I have created US-5760 in our internal backlog.

PR #816, opened 2019-02-11
[project-base] warm up production cache before generating error pages

TomasLudvik:
- The `ErrorPagesFacade::generateAllErrorPagesForProduction()` method uses production DIC for generating the pages.
- After #788 was merged, generating the cache for production consumes more memory (because protected and public methods of services are inspected using reflection).
- When the production cache was not generated, `./phing error-pages-generate` started to fail on memory usage.
- Phing targets with specified environment are prefixed with the environemnt (see `test-*` targets), that's why the new target is named `prod-warmup`.

| Q             | A
| ------------- | ---
|Description, reason for the PR| Execution of `./phing error-pages-generate` fails on memory limit.
|New feature| No 
|BC breaks| No 
|Fixes issues| ... 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -190,8 +190,22 @@ for instance:
     - `ProductSearchExportRepositoryTest` move from `Tests\ShopBundle\Functional\Model\Product\ProductSearchExport` to `Tests\ShopBundle\Functional\Model\Product\Search`
     - in production you will need to run `product-search-recreate-structure` Phing target while next build to create indexes again with new name
         - after that remove previous indexes used for Product search, so they do not consume any memory ([link to Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-index.html))
-- force generating of error pages in production environment ([#816](https://github.com/shopsys/shopsys/pull/816))
-    - update your phing target `error-pages-generate` in `build.xml` to include `<arg value="--env=prod" />`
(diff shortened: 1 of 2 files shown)


PetrHeinz:
@TomasLudvik This is just a hotfix and it may cause unforeseen issues when using some custom configuration for `prod` env on a project. Why has the memory consumption increased?

PetrHeinz:
After discussion with @TomasLudvik, I've changed the implementation to warming up the DIC before generating the error pages. Can anyone review this?

grossmannmartin:
Shouldn't be `prod-warmup` also registered as a dependency of some other targets? 
For example `build-deploy-part-2-db-dependent` or `build-new` as they are meant to be run on the production server and warming cache for production make much more sense to me.

Also, when we would solve #742, we should rename `warmup` command to `dev-warmup` because right now at a first sight I would run `warmup` for production.

PetrHeinz:
@grossmannmartin `warmup` runs in current environment (no `--env` specified). You could run `warmup` for production after you switch the env using `./bin/console shopsys:environment:change` (see [FAQ](https://github.com/shopsys/shopsys/blob/v7.0.0-beta5/docs/introduction/faq-and-common-issues.md#how-do-i-change-the-environment-productiondevelopmenttest)).

Eg. `build-new` might be meant to be run in production environement, but it'll work in `dev` as well. In [`ErrorPagesFacade::generateAllErrorPagesForProduction()`](https://github.com/shopsys/shopsys/blob/v7.0.0-beta5/packages/framework/src/Component/Error/ErrorPagesFacade.php#L57) the production environment is explicitly specified (because we don't want to render Tracy pages) and because of this I suggest warming up the cache for `production` explicitly as well.

Issue #745, opened 2019-01-15
Choosing a type for price representation and calculation


PetrHeinz:
### What is happening

At the moment, we are using a mix of `string`, `int` and `float` types when dealing with monetary values. This leads to problems with type-hinting (see #683 and #288), comparing two values (eg. `0.1 + 0.2 === 0.3` produces `false` in PHP) and rounding errors, which may get out of hand when a lot of different calculations are compounded (eg. applying a lot of coefficients to a price).

This may lead to prices being off by a cent or to hard-to-debug issues.

I have found 6 different approaches that we could take and I've described them below with their pros and cons. I'd like to hear your opinion on what should get implemented.

1) We will just use floats.
2) We will cast numbers to strings when passing the values to another method.
3) We will implement a custom value-object.
4) We will use a library for Decimal type.
5) We will use [a PHP extension](http://php-decimal.io/) for Decimal type.
6) We will use the [Money pattern](https://martinfowler.com/eaaCatalog/money.html) (using a library).

**The options are described below in separate comments - please use reactions :+1: and :-1: to express your opinion and add a comment to share your thoughts. Thanks!**

**We would like to come to a final decision on Thursday 24th, so we have a week to discuss this.**

If you find another possible approach you'd like to propose, feel free to add one in a separate comment, preferably in the same format as one of the six. :slightly_smiling_face: 

### Expected result

Decision is made about how we will represent and calculate prices (at least in the first release after we leave beta). After the changes are implemented, type-hints are consistent (will close #683, #288). Users will have a documented way to upgrade to this state.

Moreover, all no non-strict comparisons like these will be replaced by their strict counterparts:

https://github.com/shopsys/shopsys/blob/2abf7e410a322f97a909f5c4fc03728bd49e64d5/packages/framework/src/Model/Cart/Watcher/CartWatcher.php#L52
https://github.com/shopsys/shopsys/blob/2abf7e410a322f97a909f5c4fc03728bd49e64d5/packages/framework/src/Model/Order/Watcher/TransportAndPaymentWatcher.php#L117
https://github.com/shopsys/shopsys/blob/2abf7e410a322f97a909f5c4fc03728bd49e64d5/packages/framework/src/Twig/PriceExtension.php#L155
https://github.com/shopsys/shopsys/blob/2abf7e410a322f97a909f5c4fc03728bd49e64d5/packages/framework/src/Twig/PriceExtension.php#L170

PetrHeinz:
[...]

PR #821, opened 2019-02-14
[shopsys] new class for representing monetary values

### LINKED ISSUE #683: Price: annotation does not match reality
Hi,

### What is happening
According to the annotation of `getPriceWithVat()` method of the`Price` class, the return type of this method is `string`. But in some cases the return value is of type `float`.

On our project we are using `strict types` so when we are passing `getPriceWithVat()`  value to the method which expects `string`, the result is `Type error`.

Right now we are retyping this value into `string` before passing it to the next method, so we can continue without problem.

It is not propably a big issue right now, Im just noticing you that this behaviour can be a problem in the future.

Example code:
```
$orderPreview = $this->orderPreviewFactory->createForCurrentUser();
$productsPrice = $orderPreview->getProductsPrice();
$productPriceWithVat = $productsPrice->getPriceWithVat()

//$productPriceWithVat should be of type string according to the annotation but it is of type float  
```

### Expected result
Trusted annotations in the class `Price`

Thank you :)

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| consistent typing of monetary values as described in #745
|New feature| Yes 
|Fixes issues| resolves #683 


packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+        yield [1, 0, '1'];
+        yield [-1, 0, '-1'];
+        yield [10, 0, '10'];
+        yield [0.05, 1, '0.1'];
+        yield [0.5, 0, '1'];
+        yield [0.0001, 3, '0.000'];
+        yield [1.0001, 3, '1.000'];
+    }
+
+    public function testAddIsImmutable(): void
```
simara-svatopluk: What about mergin all immutable tests into one `testImmutability`?
1. Create Money Object
2. Perform all operations
3. Test not changed

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+        $this->assertEquals($expectedResult, $result->toString());
+    }
+
+    /**
+     * @return \Iterator
+     */
+    public function subtractProvider(): Iterator
+    {
+        yield ['2', '1', '1'];
+        yield ['12.15', '34.965', '-22.815'];
```
simara-svatopluk: please add also test case `higher presision, lower precision, higher precision`

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+     * @param int $scale
+     * @param string $expectedResult
+     */
+    public function testDivide(string $a, string $b, int $scale, string $expectedResult): void
+    {
+        $moneyA = Money::fromString($a);
+
+        $result = $moneyA->divide($b, $scale);
+
+        $this->assertEquals($expectedResult, $result->toString());
```
simara-svatopluk: Please use `assertSame` in all tests

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+    /**
+     * @return \Iterator
+     */
+    public function divideProvider(): Iterator
+    {
+        yield ['1', '2', 0, '1'];
+        yield ['1', '2', 1, '0.5'];
+        yield ['1', '2', 2, '0.50'];
+        yield ['1', '3', 3, '0.333'];
+        yield ['2', '3', 3, '0.667'];
```
simara-svatopluk: Here it is more obvious that our `Money` divides with rounding. Why do we have this behavior? Why we don't have division with truncation?

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+
+    /**
+     * @return \Iterator
+     */
+    public function modProvider(): Iterator
+    {
+        yield ['100', '100', '0'];
+        yield ['120', '100', '20'];
+        yield ['530', '100', '30'];
+        yield ['-20', '100', '80'];
```
simara-svatopluk: After seeing this result I would prefer to remove `mod` operation from our `Money` because this is behavior of numbers, not money

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+    /**
+     * @return \Iterator
+     */
+    public function fromStringProvider(): Iterator
+    {
+        yield ['0', '0'];
+        yield ['-0', '-0'];
+        yield ['-0.0', '-0.0'];
+        yield ['1', '1'];
+        yield ['-1', '-1'];
```
simara-svatopluk: I wonder - can money be negative? Can You have negative amount of money? It sort of doesn't make sense for me.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+
+        $this->assertEquals($expectedResult, $result->toString());
+    }
+
+    /**
+     * @dataProvider roundProvider
+     * @param string $amount
+     * @param int $scale
+     */
+    public function testRoundingIsSameAsInPhp(string $amount, int $scale): void
```
simara-svatopluk: What is the purpose of this test? I don't see it's value

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+
+        $this->assertEquals($expectedResult, $result);
+    }
+
+    /**
+     * @return \Iterator
+     */
+    public function equalsProvider(): Iterator
+    {
+        foreach ($this->compareProvider() as $dataset) {
```
simara-svatopluk: Please no. Please duplicate it here in form `a, b, expected`. This is unreadable and difficult to understand

simara-svatopluk: The same is also valid for tests below

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+        $this->assertEquals($expected, $money->toString());
+    }
+
+    /**
+     * @return \Iterator
+     */
+    public function fromStringProvider(): Iterator
+    {
+        yield ['0', '0'];
+        yield ['-0', '-0'];
```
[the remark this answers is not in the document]
PetrHeinz: I've added the test case because it surprised me (but is somewhat understandable) and I wanted to know if the behavior changed in the future. Maybe we'd like to change this behavior. I just wanted to be explicit about it.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+        yield ['1e1', '10'];
+    }
+
+    /**
+     * @dataProvider invalidValuesFromStringProvider
+     * @param string $string
+     */
+    public function testInvalidValuesInFromString(string $string): void
+    {
+        $this->expectException(BigNumbersError::class);
```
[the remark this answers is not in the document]
PetrHeinz: Good idea

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: To minimalize the error. Also, I think it's closer to our current usage (eg. form type `MoneyType` round when `scale` is defined).

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: Well, it may represent a debt. Negative money amounts are commonly used imo.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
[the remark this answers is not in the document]
simara-svatopluk: After the weekend without using my brain, I'm refreshed and back :smiling_imp:

packages/framework/tests/Unit/Component/Money/MoneyTest.php
simara-svatopluk: Thanks for the response, that was what I needed. Please add this reason also to documentation :-)

packages/framework/tests/Unit/Component/Money/MoneyTest.php
simara-svatopluk: I have 200 crowns, I'll give you 300 crowns (subtraction). Will You have 300 and I -100 or this is impossible operation?

The answer is "Depends on domain". So what makes sense in our shop domain and why?

packages/framework/src/Component/Money/Money.php
```diff
@@ -0,0 +1,213 @@
...
+
+namespace Shopsys\FrameworkBundle\Component\Money;
+
+use Litipk\BigNumbers\Decimal;
+
+/**
+ * Immutable object for representing a monetary value
+ * Uses decimal numbers with specifiable number of decimal places
+ */
+class Money
```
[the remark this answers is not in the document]
TomasLudvik: We have discussed this with @PetrHeinz and we want to postpone this decision until end of implementation of whole US. We might implement factory for this + we do not use final in framework yet and it might be in conflict with some sniffs.

packages/framework/src/Component/Money/Money.php
```diff
@@ -0,0 +1,213 @@
...
+     * @return bool
+     */
+    public function equals(self $money): bool
+    {
+        return $this->decimal->equals($money->decimal);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money $money
+     * @return int
```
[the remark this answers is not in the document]
TomasLudvik: We want to keep this method because it is also implemented in MoneyType and Decimal and it is useful for easy sorting in arrays. We have added some info in the annotation

packages/product-feed-heureka/src/Model/Product/HeurekaProductDomain.php
```diff
@@ -58,7 +59,7 @@ public function __construct(HeurekaProductDomainData $heurekaProductDomainData)
     public function edit(HeurekaProductDomainData $heurekaProductDomainData)
     {
         $this->product = $heurekaProductDomainData->product;
-        $this->cpc = $heurekaProductDomainData->cpc;
+        $this->cpc = $heurekaProductDomainData->cpc !== null ? Money::fromValue($heurekaProductDomainData->cpc) : null;
```
simara-svatopluk: I would appreciate to have `$this->setCpc()` protected method as entities above have. But it is not a blocking point, just to avoid duplication ;-)

simara-svatopluk: The same applies for the `ZboziProductDomain` below

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
simara-svatopluk: :+1: fine reason

packages/framework/tests/Unit/Component/Money/MoneyTest.php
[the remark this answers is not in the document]
TomasLudvik: We have added some new test cases to test this scenario and we have updated Money object to handle this. We have found some unexpected behavior we had to resolve. More info: c0673b8f0082a79babb3ae01eaa95f403008ba7a

packages/framework/tests/Unit/Component/Money/MoneyTest.php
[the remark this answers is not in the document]
TomasLudvik: We have encapsulated the exceptions by our own and added some test cases, more info: 44035ab9983b260198f7fb71ae8e30d3333e46c0

packages/framework/tests/Unit/Component/Money/MoneyTest.php
TomasLudvik: We do not see any advantage of such solution, but we see one disadvantage and that is that you would not see which test method failed.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
TomasLudvik: We have added one test case:  `yield ['1.000', '1.00', '0.000'];`

packages/framework/tests/Unit/Component/Money/MoneyTest.php
TomasLudvik: We have used this in all cases except one (`testRoundingIsConsistentWithPhpRound` formerly `testRoundingIsSameAsInPhp`)

packages/framework/tests/Unit/Component/Money/MoneyTest.php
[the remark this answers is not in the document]
PetrHeinz: I'd like to clarify that zero still has multiple (meaningful) representations due to its scale: `0`, `0.0`, `0.00`, ...

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: This might still be useful for example when rounding to the nearest fifty-cent. I would feel much more comfortable with using modulo in such case than multiplying, rounding and dividing the value.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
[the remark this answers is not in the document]
simara-svatopluk: Does it really have? Is `0.000 == 0`? If so, then the representation is not meaningful in my opinion. Because `0.00` only expresses that it has `scale 2`, but it is still the same value.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
simara-svatopluk: If we would have need rounding to the nearest 50-cent, then this rounding has to be part of the `Money` interface. So I would use `roundTo50` method, not `mod`.

So still, the `mod` operation is numbers behavior, not money behavior.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: I'm not sure here. In the current scope of the modification, I'm comfortable with keeping the behavior unchanged which means giving negative result (as would working with float|string).

simara-svatopluk: I'm really uncomfortable with such solution. We are coming with `Money` concept, not with `Number` concept. I need to understand our money domain before I can say whether negative money is valid. We should be serious when working with money. And I want to introduce as much restrictions as possible to avoid unintended behavior or errors.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: I see this as a fairly decent way of showcasing how exactly rounding works (in addition to `testRound` method itself). It asserts that the behavior hasn't changed, which might be useful eg. during business validation.

simara-svatopluk: As you said, "showing how rounding works" is already written in rounding use-case.

This test says "rounding works as same as in pure PHP", but I don't see value in such claim. I don't expect our users to use pure php rounding instead of ours. I would even suggest to use `Money` rounding exclusively, not php rounding ever.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: Fixed in 6cd43983e

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+     */
+    public function addProvider(): Iterator
+    {
+        yield ['1', '1', '2'];
+        yield ['12.15', '34.965', '47.115'];
+        yield ['10', '-2', '8'];
+        yield ['1', '0.01', '1.01'];
+        yield ['0.5', '0.5', '1.0'];
+        yield ['1.525', '0.475', '2.000'];
+        yield ['1.00', '1.000', '2.000'];
```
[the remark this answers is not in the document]
PetrHeinz: We've added test cases covering this:
```
        yield ['-1', '1', '0'];
        yield ['1', '-1', '0'];
        yield ['-0', '0', '0'];
        yield ['-0.0', '0', '0.0'];

File: packages/framework/tests/Unit/Component/Money/MoneyTest.php [hunk 15]
```diff
@@ -0,0 +1,582 @@
...
+     */
+    public function subtractProvider(): Iterator
+    {
+        yield ['2', '1', '1'];
+        yield ['12.15', '34.965', '-22.815'];
+        yield ['10', '-2', '12'];
+        yield ['1', '0.01', '0.99'];
+        yield ['0.5', '0.5', '0.0'];
+        yield ['1.525', '0.475', '1.050'];
+        yield ['1.00', '1.000', '0.000'];
```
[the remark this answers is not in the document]
PetrHeinz: I understand you concern but these questions are out-of-scope of this PR imo. For me, this question is similar as https://github.com/shopsys/shopsys/pull/821#discussion_r257194968

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+    public function divideProvider(): Iterator
+    {
+        yield ['1', '2', 0, '1'];
+        yield ['1', '2', 1, '0.5'];
+        yield ['1', '2', 2, '0.50'];
+        yield ['1', '3', 3, '0.333'];
+        yield ['2', '3', 3, '0.667'];
+        yield ['1000', '3', 3, '333.333'];
+        yield ['3.33', '3', 2, '1.11'];
+        yield ['0.1', '0.1', 0, '1'];
```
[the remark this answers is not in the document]
PetrHeinz: See `testCannotDivideByZero`

packages/product-feed-heureka/src/Model/Product/HeurekaProductDomain.php
PetrHeinz: This DRY violation will be removed when `Money::fromValue` gets removed. No worries :slightly_smiling_face:

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: It seems reasonable to remove `mod`. We can use some uglier method for the rounding to fifties...

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: Let's keep negative money and document this decision.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
PetrHeinz: OK

packages/framework/src/Component/Doctrine/MoneyType.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+namespace Shopsys\FrameworkBundle\Component\Doctrine;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\ConversionException;
+use Doctrine\DBAL\Types\Type;
+use Shopsys\FrameworkBundle\Component\Money\Money;
+
+class MoneyType extends Type
```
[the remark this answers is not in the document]
PetrHeinz: It seems reasonable to write it due to the `NULL` handling. We consider this a nice-to-have.

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -0,0 +1,582 @@
...
+     */
+    public function multiplyProvider(): Iterator
+    {
+        yield ['2', '1', '2'];
+        yield ['12.15', '34.965', '424.82475'];
+        yield ['10', '-2', '-20'];
+        yield ['1', '0.01', '0.01'];
+        yield ['0.5', '0.5', '0.25'];
+        yield ['1.525', '0.475', '0.724375'];
+        yield ['1.00', '1.000', '1.00000'];
```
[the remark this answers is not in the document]
PetrHeinz: We've discussed it thoroughly and there are use cases where it makes sense. Let's keep the implementation close to mathematical point-of-view. Do you see this issue as resolved?

PetrHeinz:
Now we have reviewed and fixed phases 1 (implementation of the class + unit testing) and 2 (persistence of data in entities). Thanks @simara-svatopluk. 

There are some unresolved discussions which we would like to include in the documentation, a nice-to-have unit test that we might write (https://github.com/shopsys/shopsys/pull/821#pullrequestreview-204675847) and a decision about `final`izing the class that we want to postpone until we have all usages implemented (https://github.com/shopsys/shopsys/pull/821#discussion_r257182230).

We want to resolve the next phases in separate PRs targeting this branch:

3. user input of prices (forms and data objects)
4. templating (twig extension)
5. implementing the rest of usages (mainly calculator classes)
6. documentation and upgrade notes

simara-svatopluk:
There are still unresolved comments because we are planing to document, or decide later what to do. CR done for phases 1 and 2.

simara-svatopluk:
@PetrHeinz @TomasLudvik I just realized that the build is failing on coding standards, please fix them ;-)

PetrHeinz:
@simara-svatopluk Thanks for noticing, I've fixed the standards.

PetrHeinz:
Maybe we should rename `Money::toString()` to `Money::getDecimalAmount()` so it's obvious, doesn't get confused with sub-unit amount and is future-proof (we want to add `Currency` to the properties and then it would be hard to convert to string).

PetrHeinz:
Phase 3 was merged.

The only unresolved suggestions were two missing unit tests - for `MoneyRange` constraint and `NumericToMoneyTransformer`. Writing these tests was postponed.

TomasLudvik:
Phase 4 has been merged into this PR. We will document MoneyExtension as @simara-svatopluk requested in 
https://github.com/shopsys/shopsys/pull/856#discussion_r259232102

PetrHeinz:
Phase 5 (#867) merged.

PetrHeinz:
Rebased on current `master`.

packages/framework/src/Model/Order/Watcher/TransportAndPaymentWatcher.php
```diff
@@ -283,4 +289,21 @@ protected function getRememberedPaymentPrices(): array
...
+    protected function getUnexpectedTypeOfPriceInSessionDeprecationMessage($value, string $methodName): string
+    {
+        $message = sprintf(
+            'Method "%s" expected the price to be saved as "%s" in session, got "%s" instead.',
+            $methodName,
+            Money::class,
+            \is_object($value) ? \get_class($value) : \gettype($value)
+        );
+
+        return $message . ' The method warned about the possibility of changed price, but it would fail in v8.0.0.';
```
simara-svatopluk: I'm not sure about mentioning concrete version... If this sentence isn't here, it would be still valuable and we'll decide later when exactly we want to remove the deprecation

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
PetrHeinz: Money class marked as final in cc04c5c54

packages/framework/src/Component/Doctrine/MoneyType.php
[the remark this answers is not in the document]
PetrHeinz: MoneyType class marked as final in cc04c5c

PetrHeinz:
@simara-svatopluk `NumericToMoneyTrasformer` is final, `NotMoneyTypeException` is not - that's because it is an exception thrown inside form validators. The validators are not final and neither are the constraints (none of our constraints are final).

packages/framework/src/Model/Order/Watcher/TransportAndPaymentWatcher.php
PetrHeinz: I think we'd want to remove this as soon as possible, which is `v8.0.0`. Do you think we could decide otherwise later?

PetrHeinz:
All changes to application code should be finalized, this PR is mergable to `master` as-is.

**Documentation for the new class, the concepts behind it and upgrade notes will be added in #870.**

PR #826, opened 2019-02-14
[shopsys] phpstan updated and resolved dependencies

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Updated PHPStan due to old version and persistent travis build fails.
|New feature| No 
|Fixes issues| ... 

This PR contains 5 almost independent commits, but they all changes needs to be applied before update of PHPStan is even possible and creating separate pull requests only raise the difficulty of update.

- new version of PHPStan reveal incorrect letter case in stdClass
- new version of PHPStan reveal different placeholders and values count
- due to update of nikic/php-parser I added test for ConstraintViolationExtractorClass (to be sure it's safe to update)

I need to add ignored error for project-base phpstan because in ErrorHandlerController is intentionally `undefined` error. 
And because we have phing target for phpstan for project-base and for packages separated, I had to change configuration file for phpstan project-base.
PHPStan in project-base is now run with ./project-base/phpstan.neon configuration even when you're working in monorepo.

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -47,6 +47,10 @@ for instance:
               </exec>
           </target>
         ```
+- update PHPStan to the 0.11 version ([#826](https://github.com/shopsys/shopsys/pull/826))
+    - change version of phpstan/phpstan to ^0.11 in your composer.json and update dependencies with `composer update`
+    - add `- '#Undefined variable: \$undefined#'` as ignored error to `phpstan.neon` configuration file
+    - you may need to change `StdClass` to `stdClass` in `tests/ShopBundle/Functional/Component/Grid/Ordering/GridOrderingFacadeTest.php` to pass PHPStan check
 
 ### Database migrations
 - after running database migrations, all your countries across domains will be merged together and original names will be added as translations
--- a/packages/framework/src/Component/Translation/ConstraintViolationExtractor.php
+++ b/packages/framework/src/Component/Translation/ConstraintViolationExtractor.php
@@ -96,7 +96,7 @@ class ConstraintViolationExtractor implements FileVisitorInterface, NodeVisitor
         $this->currentExecutionContextVariableNames = [];
         foreach ($node->getParams() as $parameter) {
             if ($this->isParameterExecutionContextInterfaceSubclass($parameter)) {
-                $this->currentExecutionContextVariableNames[] = $parameter->name;
+                $this->currentExecutionContextVariableNames[] = $parameter->var->name;
             }
         }
     }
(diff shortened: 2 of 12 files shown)


packages/framework/tests/Unit/Component/Translation/ConstraintViolationExtractorTest.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    public function testMessagesAreExtractedFromConstraintViolationClass(): void
+    {
+        $file = new SplFileInfo(__DIR__ . '/Resources/ConstraintViolationClass.php');
+
+        $actualCatalogue = $this->extract($file);
+
+        $expectedCatalogue = new MessageCatalogue();
+
+        $message = new Message('This message will be extracted into "validators" translation domain', 'validators');
```
vitek-rostislav: Maybe it would make sense to put the sentence into some constant so it is obvious that it is the same message as the one defined `ConstraintViolationClass`. However, I think it will not work then, right?

packages/framework/src/Component/Translation/ConstraintViolationExtractor.php
```diff
@@ -123,9 +123,10 @@ protected function isParameterExecutionContextInterfaceSubclass(Node\Param $para
      */
     protected function isAddViolationMethodCall(Node $node): bool
     {
+        /** @var \PhpParser\Node\Expr\MethodCall $node */
         return $node->var instanceof Variable
             && in_array($node->var->name, $this->currentExecutionContextVariableNames, true)
-            && $node->name === 'addViolation';
+            && $node->name->name === 'addViolation';
```
boris-brtan: ???

packages/framework/tests/Unit/Component/Translation/ConstraintViolationExtractorTest.php
grossmannmartin: Yeah, It's mentioned in `ConstraintViolationClass` annotation. Violations cannot be variables (and constants).

packages/framework/src/Component/Translation/ConstraintViolationExtractor.php
grossmannmartin: In version 3, `$node->name` returns `string`. 

Since version 4,`Identifier` object is returned and to get the string I need to read property name of the `Identifier` object or retype to string (`(string)$node->name`).

grossmannmartin: As example: 
If I parse `$context->addViolation('message')` then
`$node->name->name` returns `addViolation` and `$node->var->name` returns `context`.

That's why it's like this.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -47,6 +47,10 @@ for instance:
               </exec>
           </target>
         ```
+- update PHPStan to the 0.11 version ([#826](https://github.com/shopsys/shopsys/pull/826))
+    - change version of phpstan/phpstan to ^0.11 in your composer.json and update dependencies with `composer update`
```
vitek-rostislav: what about nikic/php-parser?

grossmannmartin: well, it's not a dependency in project-base. Only in framework, so I don't think it should be mentioned.

vitek-rostislav: oh, of course, you are right :slightly_smiling_face:

PR #828, opened 2019-02-14
[framework] implementations of GridFactoryInterface are more modular now

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Slight change of behavior of GridFactory required to fork whole class, this PR modularize GridFactory so you only change/extend the part that you are interested in changing.
|New feature| No
|Fixes issues| 

### DIFF
--- a/packages/framework/src/Model/Product/Flag/FlagGridFactory.php
+++ b/packages/framework/src/Model/Product/Flag/FlagGridFactory.php
@@ -4,6 +4,8 @@ namespace Shopsys\FrameworkBundle\Model\Product\Flag;
 
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Query\Expr\Join;
+use Shopsys\FrameworkBundle\Component\Grid\DataSourceInterface;
+use Shopsys\FrameworkBundle\Component\Grid\Grid;
 use Shopsys\FrameworkBundle\Component\Grid\GridFactory;
 use Shopsys\FrameworkBundle\Component\Grid\GridFactoryInterface;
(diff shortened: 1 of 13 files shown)


packages/framework/src/Model/Product/Flag/FlagGridFactory.php
```diff
@@ -45,15 +47,32 @@ public function __construct(
...
+    protected function getDataSource(): DataSourceInterface
     {
         $queryBuilder = $this->em->createQueryBuilder();
         $queryBuilder
             ->select('a, at')
             ->from(Flag::class, 'a')
             ->join('a.translations', 'at', Join::WITH, 'at.locale = :locale')
             ->setParameter('locale', $this->localization->getAdminLocale());
-        $dataSource = new QueryBuilderDataSource($queryBuilder, 'a.id');
+        return new QueryBuilderDataSource($queryBuilder, 'a.id');
```
vitek-rostislav:
```suggestion

        return new QueryBuilderDataSource($queryBuilder, 'a.id');
```

vitek-rostislav:
Hi @MattCzerner, after some testing and prototyping, it seems to me, that it would be sufficient to extract just the data source creation to a protected method, see https://github.com/shopsys/shopsys/commit/d7e4b172dd85f5b64c0f84e3a12cea0ad01bb88c.

I would like to discuss this, so lets call each other :phone: :slightly_smiling_face:

vitek-rostislav:
We decided not to continue with this PR now. First, we will document the Grid component in general, investigate use cases how the grids are modified on real projects a then return to the refactoring which will be based on the collected information.

boris-brtan:
hi @vitek-rostislav, is this branch still needed.

PR #835, opened 2019-02-18
required version of PHP is mentioned in installation guides

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Composer requires version of PHP 7.1 or newer when creating project
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/docs/installation/installation-using-docker-linux.md
+++ b/docs/installation/installation-using-docker-linux.md
@@ -10,6 +10,7 @@ Take a look at the article about [Monorepo](../introduction/monorepo.md) for mor
 ## Requirements
 * [GIT](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 * [PHP](http://php.net/manual/en/install.unix.php)
+    * At least version **7.1 or higher**
 * [Composer](https://getcomposer.org/doc/00-intro.md#installation-linux-unix-osx)
 * [Docker](https://docs.docker.com/engine/installation/)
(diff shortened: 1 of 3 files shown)


docs/installation/installation-using-docker-linux.md
```diff
@@ -9,7 +9,7 @@ Take a look at the article about [Monorepo](../introduction/monorepo.md) for mor
 
 ## Requirements
 * [GIT](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
-* [PHP](http://php.net/manual/en/install.unix.php)
+* [PHP 7.1 - 7.3](http://php.net/manual/en/install.unix.php)
```
vitek-rostislav: Hi, I would not mention the upper limit, we would need to change it every time we add support of new PHP :slightly_smiling_face:

TomasLudvik: Hello @vitek-rostislav,
I got inspiration from native-installation.md where it is wrote same way :-) do you insist on updating it?

vitek-rostislav: yes please :innocent: I would rather get inspired by the line below in the same doc where we mention minimum version for Docker, so I would put it this way:

```
* [PHP](http://php.net/manual/en/install.unix.php)
	* At least version **7.1 or higher**
```

TomasLudvik: Hello @vitek-rostislav,

I have updated it as requested and also rebased on actual master.

PR #838, opened 2019-02-18
[shopsys] new class for representing monetary values PHASE 3

PetrHeinz:
In this PR, we're changing mainly user input of prices.

`Money` will be used in forms and data objects.

| Q             | A
| ------------- | ---
|Description, reason for the PR| PR for easier review and management of #821
|New feature| Yes 
|Fixes issues| see #821 

### DIFF
--- a/packages/framework/src/Component/Money/Money.php
+++ b/packages/framework/src/Component/Money/Money.php
@@ -59,6 +59,14 @@ class Money implements JsonSerializable
     }
 
     /**
+     * @return \Shopsys\FrameworkBundle\Component\Money\Money
+     */
+    public static function zero(): self
+    {
+        return static::fromInteger(0);
+    }
+
+    /**
      * @return string
      */
     public function toString(): string
@@ -196,6 +204,30 @@ class Money implements JsonSerializable
     public function isLessThanOrEqualTo(self $money): bool
     {
         return $this->decimal->isLessOrEqualTo($money->decimal);
+    }
+
+    /**
+     * @return bool
+     */
+    public function isNegative(): bool
+    {
+        return $this->decimal->isNegative() && !$this->decimal->isZero();
+    }
+
+    /**
+     * @return bool
+     */
+    public function isPositive(): bool
+    {
+        return $this->decimal->isPositive();
+    }
+
+    /**
+     * @return bool
+     */
+    public function isZero(): bool
+    {
+        return $this->decimal->isZero();
     }
 
     /**
--- a/packages/framework/src/Component/Money/Money.php
+++ b/packages/framework/src/Component/Money/Money.php
@@ -4,9 +4,10 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Component\Money;
 
+use JsonSerializable;
 use Litipk\BigNumbers\Decimal;
 
-class Money
+class Money implements JsonSerializable
 {
     /**
      * @var \Litipk\BigNumbers\Decimal
@@ -67,6 +68,16 @@ class Money
         }
 
         return (string)$this->decimal;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function jsonSerialize(): array
+    {
+        return [
+            'amount' => $this->toString(),
+        ];
     }
 
     /**
--- a/packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
+++ b/packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
@@ -2,10 +2,12 @@
 
 namespace Shopsys\FrameworkBundle\Form\Admin\Order;
 
+use Shopsys\FrameworkBundle\Form\Transformers\NumericToMoneyTransformer;
 use Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\MoneyType;
+use Symfony\Component\Form\Extension\Core\Type\NumberType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Validator\Constraints;
@@ -28,18 +30,20 @@ class OrderPaymentFormType extends AbstractType
             ])
             ->add('priceWithVat', MoneyType::class, [
                 'currency' => false,
+                'scale' => 6,
                 'constraints' => [
                     new Constraints\NotBlank(['message' => 'Please enter unit price with VAT']),
                 ],
                 'error_bubbling' => true,
             ])
-            ->add('vatPercent', MoneyType::class, [
-                'currency' => false,
+            ->add('vatPercent', NumberType::class, [
                 'constraints' => [
                     new Constraints\NotBlank(['message' => 'Please enter VAT rate']),
                 ],
                 'error_bubbling' => true,
             ]);
+
+        $builder->get('priceWithVat')->addModelTransformer(new NumericToMoneyTransformer(6));
     }
 
     /**
--- a/packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
+++ b/packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
@@ -2,7 +2,6 @@
 
 namespace Shopsys\FrameworkBundle\Form\Admin\Order;
 
-use Shopsys\FrameworkBundle\Form\Transformers\NumericToMoneyTransformer;
 use Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
@@ -42,8 +41,6 @@ class OrderPaymentFormType extends AbstractType
                 ],
                 'error_bubbling' => true,
             ]);
-
-        $builder->get('priceWithVat')->addModelTransformer(new NumericToMoneyTransformer(6));
     }
 
     /**
--- a/packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
+++ b/packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
@@ -28,7 +28,6 @@ class OrderPaymentFormType extends AbstractType
                 'error_bubbling' => true,
             ])
             ->add('priceWithVat', MoneyType::class, [
-                'currency' => false,
                 'scale' => 6,
                 'constraints' => [
                     new Constraints\NotBlank(['message' => 'Please enter unit price with VAT']),
--- a/packages/framework/src/Form/Constraints/MoneyRange.php
+++ b/packages/framework/src/Form/Constraints/MoneyRange.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Form\Constraints;
+
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ */
+class MoneyRange extends Constraint
+{
+    /**
+     * @var string
+     */
+    public $minMessage = 'The amount of money should be {{ limit }} or more.';
+
+    /**
+     * @var string
+     */
+    public $maxMessage = 'The amount of money should be {{ limit }} or less.';
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
+     */
+    public $min;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
+     */
+    public $max;
+
+    /**
+     * @param array $options
+     */
+    public function __construct(array $options)
+    {
+        $this->validateMoneyOrNullOption('min', $options);
+        $this->validateMoneyOrNullOption('max', $options);
+
+        parent::__construct($options);
+
+        if ($this->min === null && $this->max === null) {
+            $message = sprintf('Either option "min" or "max" must be given for constraint "%s".', __CLASS__);
+
+            throw new \Symfony\Component\Validator\Exception\MissingOptionsException($message, ['min', 'max']);
+        }
+    }
+
+    /**
+     * @param string $optionName
+     * @param array $options
+     */
+    protected function validateMoneyOrNullOption(string $optionName, array $options): void
+    {
+        if (!isset($options[$optionName])) {
+            return;
+        }
+
+        $value = $options[$optionName];
+        if ($value !== null && !($value instanceof Money)) {
+            $message = sprintf('The "%s" constraint requires the "%s" options to be either "%s" or null', __CLASS__, $optionName, Money::class);
+            $message .= sprintf(', "%s" given.', \is_object($value) ? \get_class($value) : \gettype($value));
+
+            throw new \Symfony\Component\Validator\Exception\ConstraintDefinitionException($message);
+        }
+    }
+}
--- a/packages/framework/src/Form/Constraints/MoneyRangeValidator.php
+++ b/packages/framework/src/Form/Constraints/MoneyRangeValidator.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Form\Constraints;
+
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+class MoneyRangeValidator extends ConstraintValidator
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money|null $value
+     * @param \Symfony\Component\Validator\Constraint $constraint
+     */
+    public function validate($value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof MoneyRange) {
+            throw new \Symfony\Component\Validator\Exception\UnexpectedTypeException($constraint, MoneyRange::class);
+        }
+
+        if ($value === null) {
+            return;
+        }
+
+        if (!($value instanceof Money)) {
+            throw new \Shopsys\FrameworkBundle\Form\Exception\NotMoneyTypeException($value);
+        }
+
+        if ($constraint->max !== null && $value->isGreaterThan($constraint->max)) {
+            $this->context->addViolation($constraint->maxMessage, [
+                '{{ limit }}' => $constraint->max->toString(),
+            ]);
+        }
+        if ($constraint->min !== null && $value->isLessThan($constraint->min)) {
+            $this->context->addViolation($constraint->minMessage, [
+                '{{ limit }}' => $constraint->min->toString(),
+            ]);
+        }
+    }
+}
--- a/packages/framework/src/Form/Constraints/NotNegativeMoneyAmount.php
+++ b/packages/framework/src/Form/Constraints/NotNegativeMoneyAmount.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Form\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ */
+class NotNegativeMoneyAmount extends Constraint
+{
+    /**
+     * @var string
+     */
+    public $message = 'Amount of money should be greater than or equal to zero.';
+}
--- a/packages/framework/src/Form/Constraints/NotNegativeMoneyAmountValidator.php
+++ b/packages/framework/src/Form/Constraints/NotNegativeMoneyAmountValidator.php
@@ -28,8 +28,7 @@ class NotNegativeMoneyAmountValidator extends ConstraintValidator
             throw new \Shopsys\FrameworkBundle\Form\Exception\NotMoneyTypeException($value);
         }
 
-        $zeroMoney = Money::fromInteger(0);
-        if ($value->isLessThan($zeroMoney)) {
+        if ($value->isNegative()) {
             $this->context->addViolation($constraint->message);
         }
     }
--- a/packages/framework/src/Form/Constraints/NotNegativeMoneyAmountValidator.php
+++ b/packages/framework/src/Form/Constraints/NotNegativeMoneyAmountValidator.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Form\Constraints;
+
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+class NotNegativeMoneyAmountValidator extends ConstraintValidator
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money|null $value
+     * @param \Symfony\Component\Validator\Constraint $constraint
+     */
+    public function validate($value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof NotNegativeMoneyAmount) {
+            throw new \Symfony\Component\Validator\Exception\UnexpectedTypeException($constraint, NotNegativeMoneyAmount::class);
+        }
+
+        if ($value === null) {
+            return;
+        }
+
+        if (!($value instanceof Money)) {
+            throw new \Shopsys\FrameworkBundle\Form\Exception\NotMoneyTypeException($value);
+        }
+
+        $zeroMoney = Money::fromInteger(0);
+        if ($value->isLessThan($zeroMoney)) {
+            $this->context->addViolation($constraint->message);
+        }
+    }
+}
--- a/packages/framework/src/Form/JsFormValidatorFactory.php
+++ b/packages/framework/src/Form/JsFormValidatorFactory.php
@@ -3,11 +3,14 @@
 namespace Shopsys\FrameworkBundle\Form;
 
 use Fp\JsFormValidatorBundle\Factory\JsFormValidatorFactory as BaseJsFormValidatorFactory;
+use Fp\JsFormValidatorBundle\Model\JsFormElement;
+use JsonSerializable;
 use Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterValue;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Form;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Routing\Exception\RouteNotFoundException;
+use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints;
 
 class JsFormValidatorFactory extends BaseJsFormValidatorFactory
@@ -79,7 +82,13 @@ class JsFormValidatorFactory extends BaseJsFormValidatorFactory
             $prototype->setParent($form);
         }
 
-        return parent::createJsModel($form);
+        $model = parent::createJsModel($form);
+
+        if ($model !== null) {
+            $this->jsonSerializeValuesInAllConstraints($model);
+        }
+
+        return $model;
     }
 
     /**
@@ -94,5 +103,42 @@ class JsFormValidatorFactory extends BaseJsFormValidatorFactory
         }
 
         return parent::generateUrl($route);
+    }
+
+    /**
+     * Method searches for all Constraints and serializes all their values implementing JsonSerializable
+     * (eg. $min and $max in {@see \Shopsys\FrameworkBundle\Form\Constraints\MoneyRange}) because
+     * {@see \Fp\JsFormValidatorBundle\Model\JsModelAbstract::phpValueToJs()} does not support JsonSerializable objects
+     *
+     * Method does not modify the original Constraint objects (it clones all the Constraints)
+     *
+     * @param \Fp\JsFormValidatorBundle\Model\JsFormElement $model
+     */
+    protected function jsonSerializeValuesInAllConstraints(JsFormElement $model): void
+    {
+        if (isset($model->data['form']['constraints'])) {
+            foreach ($model->data['form']['constraints'] as $constraintName => $constraintSet) {
+                foreach ($constraintSet as $key => $constraint) {
+                    $model->data['form']['constraints'][$constraintName][$key] = $this->jsonSerializeConstraintValues($constraint);
+                }
+            }
+        }
+    }
+
+    /**
+     * @param \Symfony\Component\Validator\Constraint $constraint
+     * @return \Symfony\Component\Validator\Constraint
+     */
+    protected function jsonSerializeConstraintValues(Constraint $constraint): Constraint
+    {
+        $constraint = clone $constraint;
+
+        foreach ($constraint as $name => $value) {
+            if ($value instanceof JsonSerializable) {
+                $constraint->$name = $value->jsonSerialize();
+            }
+        }
+
+        return $constraint;
     }
 }
--- a/packages/framework/src/Form/PriceTableType.php
+++ b/packages/framework/src/Form/PriceTableType.php
@@ -3,7 +3,6 @@
 namespace Shopsys\FrameworkBundle\Form;
 
 use Shopsys\FrameworkBundle\Form\Constraints\NotNegativeMoneyAmount;
-use Shopsys\FrameworkBundle\Form\Transformers\NumericToMoneyTransformer;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\FormType;
 use Symfony\Component\Form\Extension\Core\Type\MoneyType;
@@ -24,7 +23,7 @@ class PriceTableType extends AbstractType
         foreach ($options['currencies'] as $key => $currency) {
             /* @var $currency \Shopsys\FrameworkBundle\Model\Pricing\Currency\Currency */
 
-            $priceBuilder = $builder->create($key, MoneyType::class, [
+            $builder->add($key, MoneyType::class, [
                 'currency' => false,
                 'scale' => 6,
                 'required' => true,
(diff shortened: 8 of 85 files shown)


packages/product-feed-heureka/src/Form/HeurekaProductFormType.php
```diff
@@ -38,12 +40,17 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'currency' => 'CZK',
                 'scale' => 2,
                 'constraints' => [
-                    new Range([
-                        'min' => 0,
-                        'max' => 100,
+                    new MoneyRange([
+                        'min' => Money::fromInteger(1),
```
simara-svatopluk: There used to be `'min' => 0`. Is this change intentional or is it a mistake?

packages/framework/src/Form/Constraints/MoneyRange.php
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Shopsys\FrameworkBundle\Form\Constraints;
+
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ */
+class MoneyRange extends Constraint
```
simara-svatopluk: please test this constraint

packages/framework/src/Form/Constraints/NotNegativeMoneyAmount.php
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * @Annotation
+ */
+class NotNegativeMoneyAmount extends Constraint
+{
+    /**
+     * @var string
+     */
+    public $message = 'Amount of money should be greater than or equal to zero.';
```
simara-svatopluk: ou... I expect protected... why public?

packages/framework/src/Form/Constraints/MoneyRange.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @var string
+     */
+    public $maxMessage = 'The amount of money should be {{ limit }} or less.';
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
+     */
+    public $min;
```
simara-svatopluk: I definitelly exptect `protected` here

packages/framework/src/Form/Constraints/MoneyRange.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
+     */
+    public $max;
+
+    /**
+     * @param array|null $options
+     */
+    public function __construct($options = null)
```
simara-svatopluk:
```suggestion
    public function __construct($options)
```
There is no way `$options` can be nullable and this class instantiated because at least min or max must be given

packages/framework/src/Form/Constraints/MoneyRange.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        parent::__construct($options);
+
+        if ($this->min === null && $this->max === null) {
+            $message = sprintf('Either option "min" or "max" must be given for constraint "%s".', __CLASS__);
+
+            throw new \Symfony\Component\Validator\Exception\MissingOptionsException($message, ['min', 'max']);
+        }
+
+        $this->validateMoneyOrNullOption('min', $this->min);
```
simara-svatopluk: This is strange in combination with typehint `mixed $value` in `validateMoneyOrNullOption` because we are passing `$this->min` that is `?Money`

I would prefer to validate options before calling parent constructor - local variable will be passed and `$this->min/max` will be always `?Money`

packages/framework/src/Form/Constraints/MoneyRangeValidator.php
```diff
@@ -0,0 +1,37 @@
...
+                    '{{ limit }}' => $constraint->max->toString(),
+                ]);
+            }
+
+            if ($constraint->min !== null && $value->isLessThan($constraint->min)) {
+                $this->context->addViolation($constraint->minMessage, [
+                    '{{ limit }}' => $constraint->min->toString(),
+                ]);
+            }
+        } elseif ($value !== null) {
```
simara-svatopluk: Please rearrange IF - as same as with `NotNegative...` - move `if ($value === null) {return;}` above chechking `Money`, and then throwing an exception can be without if again

packages/product-feed-heureka/src/Form/HeurekaProductCrudExtension.php
```diff
@@ -67,7 +68,7 @@ public function getData($productId)
             'cpc' => [],
         ];
         foreach ($heurekaProductDomains as $heurekaProductDomain) {
-            $pluginData['cpc'][$heurekaProductDomain->getDomainId()] = $heurekaProductDomain->getCpc();
+            $pluginData['cpc'][$heurekaProductDomain->getDomainId()] = $heurekaProductDomain->getCpc() !== null ? Money::fromString($heurekaProductDomain->getCpc()) : null;
```
PetrHeinz:
```suggestion
            $pluginData['cpc'][$heurekaProductDomain->getDomainId()] = $heurekaProductDomain->getCpc() !== null ? Money::fromValue($heurekaProductDomain->getCpc()) : null;
```

packages/framework/src/Component/Money/Money.php
```diff
@@ -4,9 +4,10 @@
 
 namespace Shopsys\FrameworkBundle\Component\Money;
 
+use JsonSerializable;
 use Litipk\BigNumbers\Decimal;
 
-class Money
+class Money implements JsonSerializable
```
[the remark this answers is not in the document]
PetrHeinz: It helped us solve the issue with JS validation, it was pretty easy to do and it is [consistent with the Money type](https://github.com/moneyphp/money/blob/master/src/Money.php#L512-L523). What would you suggest to do instead?

packages/framework/src/Form/PriceTableType.php
```diff
@@ -21,20 +23,21 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                'scale' => 6,
+                'required' => true,
+                'invalid_message' => 'Please enter price in correct format (positive number with decimal separator)',
+                'constraints' => [
+                    new Constraints\NotBlank(['message' => 'Please enter price']),
+                    new NotNegativeMoneyAmount(['message' => 'Price must be greater or equal to zero']),
+                ],
+            ]);
+
+            $priceBuilder->addModelTransformer(new NumericToMoneyTransformer(6));
```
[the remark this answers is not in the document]
PetrHeinz: It is the same `6` as in `MoneyType`'s option `scale`, which is also hardcoded in the form definition. We plan to add the model transformer automatically using `MoneyTypeExtension` and have its scale dependent on the option.

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
simara-svatopluk: Doing serialization out of the `Money` eg.:

```php
class MoneyJsonSerializer
{
  public function serialize(Money $money): array {
    return ['amount' => $money->toString()];
  }
}

File: packages/framework/src/Form/Transformers/NumericToMoneyTransformer.php [hunk 11]
```diff
@@ -0,0 +1,65 @@
...
+     * @return \Shopsys\FrameworkBundle\Component\Money\Money|null $value
+     */
+    public function reverseTransform($value): ?Money
+    {
+        if ($value === null) {
+            return null;
+        } elseif (is_string($value)) {
+            try {
+                return Money::fromString($value);
+            } catch (\Exception $e) {
```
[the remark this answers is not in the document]
PetrHeinz: Good catch, we've written this before we had the exceptions wrapped in our own...

packages/framework/src/Form/PriceTableType.php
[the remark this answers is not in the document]
simara-svatopluk: Ok, please extract the `6` into a meaningful constant.

packages/framework/src/Resources/scripts/common/validation/ShopsysConstraintNotNegativeMoneyAmount.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+
+    ShopsysFrameworkBundleFormConstraintsNotNegativeMoneyAmount = function () {
+        var self = this;
+        this.message = '';
+
+        this.validate = function (value) {
+            if (value < 0) {
+                return [self.message];
```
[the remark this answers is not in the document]
PetrHeinz: The value is set during the JS model creation in the code block rendered by `js_validation_init` Twig function. Basically, these properties are copied from the Constraint PHP object. Take a look at other JS validators in the same directory, the usage is pretty consistent.

packages/framework/src/Resources/scripts/common/validation/ShopsysConstraintNotNegativeMoneyAmount.js
```diff
@@ -0,0 +1,17 @@
...
+            } else {
```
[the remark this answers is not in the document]
PetrHeinz: You're right

packages/framework/src/Form/Constraints/NotNegativeMoneyAmount.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Form\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ */
+class NotNegativeMoneyAmount extends Constraint
```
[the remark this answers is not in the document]
PetrHeinz: This constraint will be used in `\Shopsys\FrameworkBundle\Form\Admin\TransportAndPayment\FreeTransportAndPaymentPriceLimitsFormType` where it constrains the price limit, not the price itself. Also, it may be used for other general money amounts - amount of VAT paid, a tip, etc.

packages/product-feed-heureka/src/Form/HeurekaProductFormType.php
PetrHeinz: I'd say it was a mistake
```suggestion
                        'min' => Money::fromInteger(0),
```

packages/framework/src/Form/Constraints/NotNegativeMoneyAmount.php
PetrHeinz: Constraints have their properties public. I'd say it has something to do with how they are used in Symfony forms internally...

packages/product-feed-heureka/src/Form/HeurekaProductFormType.php
simara-svatopluk: I really don't know.. one more information - in other feeds it is `1`

packages/framework/src/Form/Constraints/MoneyRange.php
PetrHeinz: Constraints must have their properties public. For example, take a look at `\Symfony\Component\Validator\Constraints\Range`.

simara-svatopluk: My eyes bleeds https://github.com/symfony/symfony/blob/v4.2.3/src/Symfony/Component/Validator/Constraint.php#L126

packages/framework/src/Form/Constraints/MoneyRange.php
PetrHeinz: I get your point, rewritten in 893142f1aad654362e0ef7d8edffacb84acfa544

packages/framework/src/Form/Constraints/MoneyRange.php
PetrHeinz: Makes sense, `$options` has to be `array`.

packages/product-feed-heureka/src/Form/HeurekaProductFormType.php
```diff
@@ -38,12 +40,17 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                        'max' => Money::fromInteger(500),
                     ]),
                 ],
             ],
         ]);
+
+        foreach ($builder->get('cpc')->all() as $price) {
```
simara-svatopluk: I think you can assign result of `->add` above to a variable like
`$cpc = $builder->add('cpc', MultidomainType::class, [`

and here use `foreach ($cpc->all()...`

packages/product-feed-heureka/src/Form/HeurekaProductCrudExtension.php
simara-svatopluk: I just found the same issue :-D

project-base/src/Shopsys/ShopBundle/Form/Front/Product/ProductFilterFormType.php
```diff
@@ -32,26 +34,20 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'currency' => false,
                 'scale' => $priceScale,
                 'required' => false,
-                'attr' => ['placeholder' => $priceTransformer->transform($config->getPriceRange()->getMinimalPrice())],
+                'attr' => ['placeholder' => $viewTransformer->transform($modelTransformer->transform($config->getPriceRange()->getMinimalPrice()))],
```
simara-svatopluk: There lines started to be difficult to read and understand (as they are doing a lot of tranformation) please consider extracting into a function or variable

packages/framework/src/Model/Product/Filter/PriceRangeRepository.php
```diff
@@ -83,6 +84,9 @@ protected function getPriceRangeByProductsQueryBuilder(QueryBuilder $productsQue
         $priceRangeData = $queryBuilder->getQuery()->execute();
         $priceRangeDataRow = reset($priceRangeData);
 
-        return new PriceRange($priceRangeDataRow['minimalPrice'], $priceRangeDataRow['maximalPrice']);
+        return new PriceRange(
+            Money::fromValue($priceRangeDataRow['minimalPrice'] ?? 0),
```
simara-svatopluk: you can you `fromString()` in my opinion because in the query, there are used aggregation functions, so the Doctrine have no idea how to map these results. I think it is right that Doctrine returns string in this case (not our Money).

project-base/src/Shopsys/ShopBundle/Resources/views/Front/Content/Product/filterFormMacro.html.twig
```diff
@@ -18,9 +18,9 @@
             <div
                 class="js-range-slider"
                 data-minimum-input-id="{{ filterForm.minimalPrice.vars.id }}"
-                data-minimal-value="{{ priceRange.minimalPrice }}"
+                data-minimal-value="{{ priceRange.minimalPrice.toValue }}"
```
simara-svatopluk:
```suggestion
                data-minimal-value="{{ priceRange.minimalPrice.toString }}"
```

I think we should use `toString` right now, not later. The same is valid for maximalPrice

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
PetrHeinz: It could be rewritten like this:

```diff
diff --git a/packages/framework/src/Component/Money/Money.php b/packages/framework/src/Component/Money/Money.php
index ef3150a29..e0bc937ef 100644
--- a/packages/framework/src/Component/Money/Money.php
+++ b/packages/framework/src/Component/Money/Money.php
@@ -4,10 +4,9 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Component\Money;
 
-use JsonSerializable;
 use Litipk\BigNumbers\Decimal;
 
-class Money implements JsonSerializable
+class Money
 {
     /**
      * @var \Litipk\BigNumbers\Decimal
@@ -70,16 +69,6 @@ class Money implements JsonSerializable
         return (string)$this->decimal;
     }
 
-    /**
-     * @return string[]
-     */
-    public function jsonSerialize(): array
-    {
-        return [
-            'amount' => $this->toString(),
-        ];
-    }
-
     /**
      * @param \Shopsys\FrameworkBundle\Component\Money\Money $money
      * @return \Shopsys\FrameworkBundle\Component\Money\Money
diff --git a/packages/framework/src/Form/JsFormValidatorFactory.php b/packages/framework/src/Form/JsFormValidatorFactory.php
index 83bcdddb9..336a55468 100644
--- a/packages/framework/src/Form/JsFormValidatorFactory.php
+++ b/packages/framework/src/Form/JsFormValidatorFactory.php
@@ -5,6 +5,7 @@ namespace Shopsys\FrameworonekBundle\Form;
 use Fp\JsFormValidatorBundle\Factory\JsFormValidatorFactory as BaseJsFormValidatorFactory;
 use Fp\JsFormValidatorBundle\Model\JsFormElement;
 use JsonSerializable;
+use Shopsys\FrameworkBundle\Component\Money\Money;
 use Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterValue;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Form;
@@ -85,7 +86,7 @@ class JsFormValidatorFactory extends BaseJsFormValidatorFactory
         $model = parent::createJsModel($form);
 
         if ($model !== null) {
-            $this->jsonSerializeValuesInAllConstraints($model);
+            $this->jsonSerializeMoneyValuesInAllConstraints($model);
         }
 
         return $model;
@@ -106,21 +107,20 @@ class JsFormValidatorFactory extends BaseJsFormValidatorFactory
     }
 
     /**
-     * Method searches for all Constraints and serializes all their values implementing JsonSerializable
-     * (eg. $min and $max in {@see \Shopsys\FrameworkBundle\Form\Constraints\MoneyRange}) because
-     * {@see \Fp\JsFormValidatorBundle\Model\JsModelAbstract::phpValueToJs()} does not support JsonSerializable objects
+     * Method searches for all Constraints and serializes all their values of Money type
+     * (eg. $min and $max in {@see \Shopsys\FrameworkBundle\Form\Constraints\MoneyRange})
      *
      * Method does not modify the original Constraint objects (it clones all the Constraints)
      *
      * @param \Fp\JsFormValidatorBundle\Model\JsFormElement $model
      */
-    protected function jsonSerializeValuesInAllConstraints(JsFormElement $model): void
+    protected function jsonSerializeMoneyValuesInAllConstraints(JsFormElement $model): void
     {
         if (isset($model->data['form']['constraints'])) {
             foreach ($model->data['form']['constraints'] as $constraintName => $constraintSet) {
                 foreach ($constraintSet as $key => $constraint) {
                     $constraint = clone $constraint;
-                    $this->jsonSerializeConstraintValues($constraint);
+                    $this->jsonSerializeConstraintMoneyValues($constraint);
 
                     $model->data['form']['constraints'][$constraintName][$key] = $constraint;
                 }
@@ -131,11 +131,13 @@ class JsFormValidatorFactory extends BaseJsFormValidatorFactory
     /**
      * @param \Symfony\Component\Validator\Constraint $constraint
      */
-    protected function jsonSerializeConstraintValues(Constraint $constraint): void
+    protected function jsonSerializeConstraintMoneyValues(Constraint $constraint): void
     {
         foreach ($constraint as $name => $value) {
-            if ($value instanceof JsonSerializable) {
-                $constraint->$name = $value->jsonSerialize();
+            if ($value instanceof Money) {
+                $constraint->$name = [
+                    'amount' => $value->toString(),
+                ];
             }
         }
     }

```

But we've discussed it with @TomasLudvik and there is definitely added value in `JsFormValidatorFactory` understanding the concept of `JsonSerializable` objects in `Constraint`s, if we want to use value objects in our code more. Furthermore there might be added value in `Money` being able to serialize to a normalized format and JSON seems quite suitable.

Money type in `moneyphp/money` uses this approach too.

I'm undecided here, to be honest. What do others think?

packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
```diff
@@ -27,14 +28,13 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'error_bubbling' => true,
             ])
             ->add('priceWithVat', MoneyType::class, [
-                'currency' => false,
+                'scale' => 6,
```
simara-svatopluk: I think the constraint below should be `NotNegativeMoneyAmount` the same is valid for other `OrderXXXFormType`

packages/framework/src/Form/PriceTableType.php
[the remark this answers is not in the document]
PetrHeinz: We automated adding the transformer in e79a2b737, it looks much better now :slightly_smiling_face:

packages/product-feed-heureka/src/Form/HeurekaProductFormType.php
PetrHeinz: According to @TomasLudvik it was just a copy-paste mistake

packages/framework/src/Form/JsFormValidatorFactory.php
```diff
@@ -95,4 +104,39 @@ protected function generateUrl($route)
...
+    }
+
+    /**
+     * @param \Symfony\Component\Validator\Constraint $constraint
+     */
+    protected function jsonSerializeConstraintValues(Constraint $constraint): void
+    {
+        foreach ($constraint as $name => $value) {
+            if ($value instanceof JsonSerializable) {
+                $constraint->$name = $value->jsonSerialize();
```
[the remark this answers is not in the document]
PetrHeinz: Good point, rewritten in ab5ed618c

packages/framework/src/Form/Constraints/NotNegativeMoneyAmountValidator.php
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money|null $value
+     * @param \Symfony\Component\Validator\Constraint $constraint
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        if ($value instanceof Money) {
+            $zeroMoney = Money::fromInteger(0);
+
+            if ($value->isLessThan($zeroMoney)) {
```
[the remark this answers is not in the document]
PetrHeinz: I guess we should implement `isPositive`, `isNegative` and `isZero` then. Maybe a static method for zero money should be also implemented (`Money::zero()`), because these use-cases are quite common.

All new methods should be unit-tested.

packages/framework/src/Form/JsFormValidatorFactory.php
```diff
@@ -95,4 +104,39 @@ protected function generateUrl($route)
...
+    /**
+     * Method searches for all Constraints and serializes all their values implementing JsonSerializable
+     * (eg. $min and $max in {@see \Shopsys\FrameworkBundle\Form\Constraints\MoneyRange}) because
+     * {@see \Fp\JsFormValidatorBundle\Model\JsModelAbstract::phpValueToJs()} does not support JsonSerializable objects
+     *
+     * Method does not modify the original Constraint objects (it clones all the Constraints)
+     *
+     * @param \Fp\JsFormValidatorBundle\Model\JsFormElement $model
+     */
+    protected function jsonSerializeValuesInAllConstraints(JsFormElement $model): void
```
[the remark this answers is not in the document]
PetrHeinz: In the context of JS validation it seems to me sufficiently documented by the comment. The JS validation is a complex issue and shouild be properly documented, tested and fixed. It it definitely out of the scope of this PR. See #648

packages/product-feed-heureka/src/Form/HeurekaProductFormType.php
PetrHeinz: No, the `add` method returns the parent builder, not the currently added one (so you can call multiple `add`s).

packages/framework/src/Model/Product/Filter/ProductFilterRepository.php
```diff
@@ -72,25 +73,25 @@ public function applyFiltering(
...
+        ?Money $minimalPrice,
+        ?Money $maximalPrice,
         PricingGroup $pricingGroup
     ) {
         if ($maximalPrice !== null || $minimalPrice !== null) {
             $priceLimits = 'pcp.product = p AND pcp.pricingGroup = :pricingGroup';
             if ($minimalPrice !== null) {
                 $priceLimits .= ' AND pcp.priceWithVat >= :minimalPrice';
-                $productsQueryBuilder->setParameter('minimalPrice', $minimalPrice);
+                $productsQueryBuilder->setParameter('minimalPrice', $minimalPrice->toValue());
```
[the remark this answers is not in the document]
PetrHeinz: I was hoping there was a way to make Doctrine understand the type and convert it in DQL. The same problem is in `PriceRangeRepository` where we fetch `MIN(pcp.priceWithVat)` and the Money is not returned by Doctrine.

I wanted to read more about custom types in Doctrine later. Maybe we can just use `fromString()` / `toString()` in such edge cases and be done with it.

project-base/src/Shopsys/ShopBundle/Form/Front/Product/ProductFilterFormType.php
PetrHeinz: We extracted it into a method after the model transformer was automatically added in MoneyTypeExtension

packages/framework/src/Model/Product/Filter/PriceRangeRepository.php
PetrHeinz: It might be that case. Thanks for mentioning that, I was trying to postpone the decision here (same as in https://github.com/shopsys/shopsys/pull/838#discussion_r258407036).

project-base/src/Shopsys/ShopBundle/Resources/views/Front/Content/Product/filterFormMacro.html.twig
PetrHeinz: We might want to use a Twig filter instead. Let's wait with this into the phase 4.

packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
PetrHeinz: I don't want to change the behavior. This is only an admin form type and OrderItems may have negative prices (eg. promo codes, rounding items, ...).

packages/framework/src/Form/Constraints/NotNegativeMoneyAmountValidator.php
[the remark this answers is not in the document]
TomasLudvik: Added this in bff9560fade13d1f75c80c8546f1177cbe395c24

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
simara-svatopluk: As I already said couple of times - it is mixing responsibilities for me; and if we are able to separate those responsibilities (and we are, pretty easily), then I'm always for SRP.

TomasLudvik:
@simara-svatopluk said:
> PetrHeinz TomasLudvik I think we should use `fromString` in tests right now (as we do in DataFixtures).
> 
> In tests there are (at least what I saw) data from data providers and there are strings (even floats or int won't cause problems). I think we can call `fromString` in tests right now without any problem.
> 
> If you agree, please fix your tests :-)
I have reviewed all current tests using `fromValue` and all those tests will be updated in next phases.

project-base/src/Shopsys/ShopBundle/Form/Front/Product/ProductFilterFormType.php
```diff
@@ -94,4 +89,23 @@ public function configureOptions(OptionsResolver $resolver)
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money $money
+     * @param \Symfony\Component\Form\FormBuilderInterface $moneyBuilder
+     * @return string
+     */
+    protected function transformMoneyToView(Money $money, FormBuilderInterface $moneyBuilder): string
+    {
+        foreach ($moneyBuilder->getModelTransformers() as $modelTransformer) {
+            /** @var \Symfony\Component\Form\DataTransformerInterface $modelTransformer */
+            $money = $modelTransformer->transform($money);
```
simara-svatopluk: This seems to me quite hacky... But I don't have better solution

packages/framework/src/Component/Money/Money.php
```diff
@@ -187,6 +198,34 @@ public function isLessThanOrEqualTo(self $money): bool
         return $this->decimal->isLessOrEqualTo($money->decimal);
     }
 
+    /**
+     * @return bool
+     */
+    public function isNegative(): bool
+    {
+        if ($this->decimal->isZero() && $this->decimal->isNegative()) {
```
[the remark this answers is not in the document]
TomasLudvik: Yeah, but decimal says it is for `-0` value

packages/framework/src/Component/Money/Money.php
```diff
@@ -219,4 +258,12 @@ protected static function createDecimalFromString(string $string, int $scale = n
             throw new \Shopsys\FrameworkBundle\Component\Money\Exception\InvalidNumericArgumentException($string, $e);
         }
     }
+
+    /**
+     * @return \Shopsys\FrameworkBundle\Component\Money\Money
+     */
+    public static function getZeroMoney(): self
```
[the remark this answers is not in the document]
TomasLudvik: We have renamed it to `Money::zero`

packages/framework/tests/Unit/Component/Money/MoneyTest.php
```diff
@@ -672,4 +672,90 @@ public function lessThanOrEqualToProvider(): Iterator
...
+        yield ['-1', false];
+        yield ['-0.55', false];
+    }
+
+    public function testGetZeroMoney(): void
+    {
+        $this->assertEquals(Money::fromInteger(0), Money::getZeroMoney());
+        $this->assertEquals(Money::fromString('0'), Money::getZeroMoney());
+        $this->assertEquals(Money::fromFloat(0.00, 0), Money::getZeroMoney());
+        $this->assertSame(true, Money::getZeroMoney()->isZero());
```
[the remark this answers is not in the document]
TomasLudvik: We have updated that test:
```
$zeroMoney = Money::zero();

$this->assertSame('0', $zeroMoney->toString());
```

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
simara-svatopluk: It doesn't matter. Zero => false, always. and then return isNegative.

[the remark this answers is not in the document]
PetrHeinz: So it can be simplified to `return $this->decimal->isNegative() && !$this->decimal->isZero()`, right?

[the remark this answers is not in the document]
simara-svatopluk: Yep

packages/framework/src/Form/Transformers/NumericToMoneyTransformer.php
```diff
@@ -0,0 +1,65 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Form\Transformers;
+
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Symfony\Component\Form\DataTransformerInterface;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+
+class NumericToMoneyTransformer implements DataTransformerInterface
```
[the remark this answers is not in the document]
PetrHeinz: I'm setting writing extra tests aside.

packages/framework/src/Form/Constraints/MoneyRange.php
PetrHeinz: I'm setting writing extra tests aside.

PetrHeinz:
Phase 3 was successfully reviewed and tested :heavy_check_mark: 

The only unresolved suggestions were two missing unit tests - for MoneyRange constraint and NumericToMoneyTransformer. Writing these tests was postponed.

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
grossmannmartin: From my point of view (correct me if I'm wrong), serialization is currently used mainly for constraints. But that doesn't mean it cannot be used anywhere else in the future.

At first, it seemed logical to me for `Money` to implement `JsonSerializable` but the more I think about it the more it looks like a bad idea. Not just because of the violation of SRP (and here I really don't know if it is the violation or not. The line seems a little blurry), but we can set the path for potential god-object. 

If I want to "transform" Money in some different way (pretty print, to XML, ...) I always create a separate class with separate responsibility. And thinking about that, I believe it should be done the same for JSONs.

And I would really appreciate the class responsible for serialization (even if it would be used in just one place) instead of creating the array in place. But that's just my opinion and I'm not forcing anything :)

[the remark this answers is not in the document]
DavidOstrozlik: I don't see any problem with Money being `JsonSerializable`. JSON serialization is mainly used when you need to work with forms on frontend (JS validation, sending with AJAX, ...) or when you need to get some data from backend (again with AJAX). I see that the case of Money is almost the same. Some transformer would work but I don't see the added value of it, because for pretty printing and XML's we mostly use twig filters.
But i'm okay with both outcomes.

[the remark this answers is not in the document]
simara-svatopluk: The current `Money` responsibility is not simple - it is to represent money in the shop. It has to be always consistent, it must greatly cover edge cases, it must handle money operations. The `Money` becomes a stable piece of the SSFW, it is something we rely on.

The `JsonSerializable` suggest to add a method for JSON serialization. JSON serialization is not mentioned in the Money responsibility explained earlier. JSON serialization is not part of "working with money" domain.

Also JSON serialization is used only out of the model - in forms; maybe ajax; maybe admin - these use cases are not model use cases, therefore methods for these use cases should not be part of the model.

PetrHeinz:
Rebased on `ph-tl-money` before merging.

packages/framework/src/Component/Money/Money.php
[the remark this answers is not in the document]
PetrHeinz: Advantages of the two approaches:

Using `JsonSerializable`:
- easiest usage of the serialization
- changes in `JsFormValidatorFactory` can be immediately reused for any `JsonSerializable` constraint parameter
- same approach as in `moneyphp/money`

Using separate class for serializing Money to JSON:
- avoids mixing responsibilities
- usage can be found more easily in IDE (Money can be passed to any `json_encode` call)
- can implement deserialization as well as serialization (not used in Shopsys Framework atm)

The serialization is used at two places in the application: `JsFormValidatorFactory` (constraint parameters) and `Admin/Content/Order/orderItem.html.twig` (transport and payment prices in array).

PR #851, opened 2019-02-19
flysystem for ImageFacade::copyImages and FileUpload::postFlushEntity is unified

### LINKED ISSUE #686: Images for the main variant are not displayed

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| ImageFacade::copyImages is used during creation of the main variant from nonVariant product and all images need to be copied into newly created entity and there was bug when first step of copying copied files into cache on external filesystem instead into local one so the second step could use it and copied it back with the new filenames for newly created entities
|New feature| No 
|Fixes issues| resolves #686   

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -13,6 +13,8 @@ There you can find links to upgrade notes for other versions too.
         - `src/Shopsys/ShopBundle/Resources/scripts/frontend/promoCode.js`
         - `src/Shopsys/ShopBundle/Resources/views/Front/Content/Order/PromoCode/index.html.twig`
     - dump translations using `php phing dump-translations` and fill in the translations based on the changes from pull request
+- check whether you extended class or method `ImageFacade::copyImages` or used it in your project and make sure it works like you intended ([#851](https://github.com/shopsys/shopsys/pull/851))
+    - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, set as local filesystem storage in path `%kernel.project_dir%` by default
 
 ### Tools
 - *(low priority)* add `product-search-export-products` as a dependency of `build-demo` phing target in your `build.xml`
--- a/packages/framework/src/Component/Image/ImageFacade.php
+++ b/packages/framework/src/Component/Image/ImageFacade.php
@@ -4,6 +4,7 @@ namespace Shopsys\FrameworkBundle\Component\Image;
 
 use Doctrine\ORM\EntityManagerInterface;
 use League\Flysystem\FilesystemInterface;
+use League\Flysystem\MountManager;
 use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
 use Shopsys\FrameworkBundle\Component\FileUpload\FileUpload;
 use Shopsys\FrameworkBundle\Component\Image\Config\ImageConfig;
@@ -29,6 +30,11 @@ class ImageFacade
      * @var \League\Flysystem\FilesystemInterface
      */
     protected $filesystem;
+
+    /**
+     * @var \League\Flysystem\MountManager
+     */
+    protected $mountManager;
 
     /**
      * @var \Shopsys\FrameworkBundle\Component\FileUpload\FileUpload
@@ -59,6 +65,8 @@ class ImageFacade
      * @param \Shopsys\FrameworkBundle\Component\FileUpload\FileUpload $fileUpload
      * @param \Shopsys\FrameworkBundle\Component\Image\ImageLocator $imageLocator
      * @param \Shopsys\FrameworkBundle\Component\Image\ImageFactoryInterface $imageFactory
+     * @param \League\Flysystem\MountManager $mountManager
+     * @param \League\Flysystem\MountManager;
      */
     public function __construct(
         $imageUrlPrefix,
@@ -68,7 +76,8 @@ class ImageFacade
         FilesystemInterface $filesystem,
         FileUpload $fileUpload,
         ImageLocator $imageLocator,
-        ImageFactoryInterface $imageFactory
+        ImageFactoryInterface $imageFactory,
+        MountManager $mountManager
     ) {
         $this->imageUrlPrefix = $imageUrlPrefix;
         $this->em = $em;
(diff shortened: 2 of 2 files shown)


packages/framework/src/Component/Image/ImageFacade.php
```diff
@@ -329,9 +339,9 @@ public function copyImages($sourceEntity, $targetEntity)
         $sourceImages = $this->getAllImagesByEntity($sourceEntity);
         $targetImages = [];
         foreach ($sourceImages as $sourceImage) {
-            $this->filesystem->copy(
-                $this->imageLocator->getAbsoluteImageFilepath($sourceImage, ImageConfig::ORIGINAL_SIZE_NAME),
-                $this->fileUpload->getTemporaryFilepath($sourceImage->getFilename())
+            $this->mountManager->copy(
+                'main://' . $this->imageLocator->getAbsoluteImageFilepath($sourceImage, ImageConfig::ORIGINAL_SIZE_NAME),
+                'local://' . $this->fileUpload->getTemporaryFilepath($sourceImage->getFilename())
```
grossmannmartin: I'm really confused about this solution. It seems inappropriate to set here wrapper directly. 
Do you need to copy the images from the main filesystem (i believe this could be for example Google Cloud Bucket) to the local filesystem (disk)?

And how does the inconsistency between `ImageFacade` and `FileUpload` happened?

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -6,6 +6,11 @@ This guide contains instructions to upgrade from version v7.0.0-beta6 to Unrelea
 There you can find links to upgrade notes for other versions too.
 
 ## [shopsys/framework]
+### Application
+- `ImageFacade::copyImages` was fixed to use correct filesystem endpoint for storing temporary image files ([#851](https://github.com/shopsys/shopsys/pull/851))
```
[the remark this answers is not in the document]
boris-brtan: it is like introduction to the instructions

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -6,6 +6,11 @@ This guide contains instructions to upgrade from version v7.0.0-beta6 to Unrelea
...
+    - check the use or extensions of this class
```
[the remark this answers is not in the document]
boris-brtan: i dont know. would it be better somethin like ? `check whether you extended this class or use it in your project and make sure it is OK from the view of this Pull Request` ???

packages/framework/src/Component/Image/ImageFacade.php
boris-brtan: inconsistency happened via implementation of `Flysystem` 
as we can see here the first step of the copying 
```
$this->filesystem->copy(
                $this->imageLocator->getAbsoluteImageFilepath($sourceImage, ImageConfig::ORIGINAL_SIZE_NAME),
                $this->fileUpload->getTemporaryFilepath($sourceImage->getFilename())
``` tries to copy for example gc://product/1.jpg into gc://%kernel.root_dir%/var/cache/dev/fileUploads and the second `FileUpload.php` class
```
$this->mountManager->move('local://' . $sourceFilepath, 'main://' . $targetFilename);
```
tries to copy it from %kernel.root_dir%/var/cache/dev/fileUploads so this PR fixes the first part where the copy is from GC into local container filesystem and the second part could copy it into GC with the new id.png or id.jpg that it will get from Doctrine eventlistener postPersist.

Comment (boris-brtan) [reply packages/framework/src/Component/Image/ImageFacade.php]: so old buggy way was GC -> GC Local(could not be here since it was copied into GC) -> GC
new way is GC -> Local Local -> GC

**Comment by boris-brtan** [comment]
rebased onto master due to conflicts in `UPGRADE-unreleased.md`

File: packages/framework/src/Component/Image/ImageFacade.php (hunk 1 above)
Comment (grossmannmartin) [reply packages/framework/src/Component/Image/ImageFacade.php]: Okay, I'm still not convinced, using wrappers directly is a good idea, but it's consistent with other usages (in `FileUpload` or `FeedExport`).
So it's fine by me.

Comment (boris-brtan) [reply packages/framework/src/Component/Image/ImageFacade.php]: ooooo, Feed export, daaamn. Another temporary path in external storage. @TomasLudvik is that right? I will create some issue for this

Comment (boris-brtan) [reply packages/framework/src/Component/Image/ImageFacade.php]: oh it seems to be OK. $feed->getTemporaryFilepath() just adds suffix `.tmp` so no var/www/html/var/cache is created

**Comment by boris-brtan** [comment]
@grossmannmartin new CR please.

**Comment by boris-brtan** [comment]
issue that should cover strange thing with cache files https://github.com/shopsys/shopsys/issues/863

File: docs/upgrade/UPGRADE-unreleased.md [hunk 4]
```diff
@@ -13,6 +13,8 @@ There you can find links to upgrade notes for other versions too.
         - `src/Shopsys/ShopBundle/Resources/scripts/frontend/promoCode.js`
         - `src/Shopsys/ShopBundle/Resources/views/Front/Content/Order/PromoCode/index.html.twig`
     - dump translations using `php phing dump-translations` and fill in the translations based on the changes from pull request
+- check whether you extended class or method `ImageFacade::copyImages` or used it in your project and make sure it works like you intended ([#851](https://github.com/shopsys/shopsys/pull/851))
+    - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, defaultly set as local filesystem storage in path `%kernel.project_dir%`
```
VitMickal:
```suggestion
    - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, set as local filesystem storage in path `%kernel.project_dir%` by default
```

PR #869, opened 2019-02-25
[framework] fixed files uploading

### LINKED ISSUE #863: check validity of convertToShopFormatAndGetNewFilename, ImageFactory::create and classes that interfere with cache folders storage
### What is happening

We found some strange behaviour that during uploading files (at least from beta4 {before services removal it was Imageservice->createImage}) and also for creating main variant with images({moved after service removal into ImageFactory::create}) there is duplicitly created file into `%kernel.cache_dir%/var/html/var/cache/%ENV%/fileUploads` and it is not removed after persisting the new Image entity.
Another thing is the definition of ImageFactory::create where it contains nullable parameters that are in the method used as non-nullable parameters and so on....

### Expected result

to know what TH is happening and create some issue or issues for fixing current strange state.
- check validity/BV of `ImageFacade::copyImages`
- gain know-how about ImageFactory::create and its reponsibilities
- create issues/docs for fixing/explaining the current state

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| weird behavior of file uploads (see #863), also renaming from jpeg to jpg was not working
|New feature| Yes 
|Fixes issues| closes #863 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -17,6 +17,8 @@ There you can find links to upgrade notes for other versions too.
     - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, set as local filesystem storage in path `%kernel.project_dir%` by default
 - use `\Shopsys\FrameworkBundle\Component\Money\Money` class for representing monetary values in the whole application ([#821](https://github.com/shopsys/shopsys/pull/821))
     - detailed upgrade instruction will be added shortly, see the PR for more information
+- you need to provide `$temporaryFilenames` parameter anywhere you use `ImageFactoryInterface::create()` and `ImageFacade::uploadImage()` functions ([#869](https://github.com/shopsys/shopsys/pull/869))
+    - the parameter is not nullable now
 
 ### Tools
 - *(low priority)* add `product-search-export-products` as a dependency of `build-demo` phing target in your `build.xml`
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -19,6 +19,7 @@ There you can find links to upgrade notes for other versions too.
     - detailed upgrade instruction will be added shortly, see the PR for more information
 - you need to provide `$temporaryFilenames` parameter anywhere you use `ImageFactoryInterface::create()` and `ImageFacade::uploadImage()` functions ([#869](https://github.com/shopsys/shopsys/pull/869))
     - the parameter is not nullable now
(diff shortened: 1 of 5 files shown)


docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -15,6 +15,9 @@ There you can find links to upgrade notes for other versions too.
     - dump translations using `php phing dump-translations` and fill in the translations based on the changes from pull request
 - check whether you extended class or method `ImageFacade::copyImages` or used it in your project and make sure it works like you intended ([#851](https://github.com/shopsys/shopsys/pull/851))
     - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, set as local filesystem storage in path `%kernel.project_dir%` by default
+- fixed files upload ([#869](https://github.com/shopsys/shopsys/pull/869))
```
grossmannmartin: it's not an instruction I would drop this whole line and put PR link to next line.

packages/framework/src/Component/Image/Processing/ImageProcessor.php
```diff
@@ -162,17 +160,4 @@ public function getSupportedImageExtensions()
     {
         return $this->supportedImageExtensions;
     }
-
-    /**
-     * @param string $filepath
-     * @param string $newFilepath
-     */
-    protected function removeFileIfRenamed($filepath, $newFilepath)
```
grossmannmartin: Removed method is not mentioned in upgrade notes and probably should be.

vitek-rostislav: I was thinking about it as well, however, I decided not to mention it as this is a protected method. So the upgrade note would look like "If you are extending ImageProcessor class and if you are using its method removeFileIfRenamed(), you need to implement it yourself now as the method was removed"...seems like too many ifs in one instruction to me...what do you think?

grossmannmartin: Since it is a little edge-case to me (method was private a month ago), I would simplify it just to "if you're using protected method ImageProcessor::removeFileIfRenamed in your code, remove the usage due to the method was removed and the code was moved to `convertToShopFormatAndGetNewFilename`".

grossmannmartin: but I can live without this upgrade note (private month ago)

vitek-rostislav: ok, I like your formulation so I added it there :slightly_smiling_face:

vitek-rostislav:
Thanks, the annotation fixed :+1:

PR #854, opened 2019-02-20
[shopsys] move data fixtures to project-base

DavidOstrozlik:
This is only a temporary solution as data fixtures will be moved back to framework when they'll be prepared for extension. We made the decision in #738 so you can head there see why we made it.

| Q             | A
| ------------- | ---
|Description, reason for the PR| this is a temporary solution, because we want users to have possibility to change data fixtures but they are not yet extendable.
|New feature| No 
|[BC breaks](/docs/contributing/backward-compatibility-promise.md)| Yes 
|Fixes issues| ... 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #870, opened 2019-02-25
[shopsys] new class for representing monetary values PHASE 6

PetrHeinz:
In this PR, we're adding documentation and upgrade notes related to the `Money` object (#821).

| Q             | A
| ------------- | ---
|Description, reason for the PR| PR for easier review and management of #821
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/docs/introduction/how-to-work-with-money.md
+++ b/docs/introduction/how-to-work-with-money.md
@@ -1,0 +1,438 @@
+# How to Work with Money
+
+Money is a very important concept for every e-commerce project.
+In Shopsys Framework, all monetary values (*prices, account balances, discount amounts, price limits etc.*) are represented by an instance of [the `Money` class](#money-class).
+
+This approach has several advantages:
+- it avoids problems with floating point number calculations and comparisons (see [official PHP documentation](http://php.net/manual/en/language.types.float.php) for details)
+- allows easy-to-use interfaces with consistent type-hinting so you can be sure what type of value you should be using
+- prevents accidental conversion to unexpected types (which may be problematic eg. when using the `===` operator)
+- makes the application design clearer and future changes easier
+
+**Table of Contents:**
+- [General Concept](#general-concept)
+- [Money Class](#money-class)
+- [Money in Forms](#money-in-forms)
+- [Money in Twig Templates](#money-in-twig-templates)
+- [Money in Javascript](#money-in-javascript)
+- [Money in Doctrine](#money-in-doctrine)
+- [Unit and Functional Tests](#unit-and-functional-tests)
+- [Price Class](#price-class)
+
+## General Concept
+The money concept in Shopsys Framework represents and encapsulates monetary values with a decimal part, like `100`, `0.50`, `10.99`, `0.0005`, ...
+Money is represented without currency.
+
+### Scale
+Scale defines the precision of the decimal part and it can be a bit tricky.
+
+Imagine you want to represent `1/3` (one third) in your application. In a `float`, it would be actually represented as `0.333333333333333314829616256247390992939472198486328125` because of [the floating-point precision](http://php.net/manual/en/language.types.float.php).
+
+When you want to work with one third in terms of money, you have to specify the scale - the number of places after the decimal point that should be taken into account.
+So you can create a monetary value from `1/3` in the scale of 2 (`0.33`), or in the scale of 8 (`0.33333333`).
+But it will never be exactly one third as it is inexpressible using a finite decimal.
+
+The money concept keeps the computation and comparisons precise up to the defined scale.
+Eg. if you have `1/3` with the scale of 4 (`0.3333`) and multiply it by `3` you'll get `0.9999`, not `1`.
+You can get around it using [rounding](#rounding).
+
+The scale has to be specified during [rounding](#rounding), [creating from floats](#construction) and [division](#computation).
+
+## Money Class
+
+[`Money`](/packages/framework/src/Component/Money/Money.php) is an immutable [value object](https://codete.com/blog/value-objects/).
+
+It uses a decimal representation of the money amount and it does not contain any reference to the used currency.
+You can get the decimal representation as a `string` via the `getAmount` method.
+
+*Note: If in doubt about the results of any method, you can take a look at [its unit tests](/packages/framework/tests/Unit/Component/Money/MoneyTest.php) which contain many examples of the class' behavior.*
+
+### Construction
+
+`Money` can be constructed directly from integers and numeric strings as they are able to represent decimal numbers precisely: `$tenDollars = Money::create(10)`, `$oneAndHalfEuro = Money::create('1.50')`.
+
+It may be also constructed from floating point numbers, but the scale (number of decimal places) must be explicitly specified: `Money::createFromFloat($floatFromExternalSource, 2)`.
+The created value will be [rounded](#rounding) to the provided scale.
+
+Zero amount of money can be constructed by calling `Money::zero()`.
+
+### Computation
+
+To compute with monetary values you have to use the object's methods instead of arithmetic operators (`+`, `-`, `*`, `/`):
+
+- `Money::add(Money $addend) : Money`
+- `Money::subtract(Money $subtrahend) : Money`
+- `Money::multiply(int|string $multiplier) : Money`
+- `Money::divide(int|string $divisor, int $scale) : Money`
+
+*Note: `Money` is immutable, which means that all these methods create a new object and the original is never modified.*
+
+For addition and subtraction, the other parameter has to be also a `Money` instance.
+For multiplication and division, the other parameter has to be an integer or a numeric string (as they are able to represent decimal numbers precisely), not a float.
+
+The scale (number of decimal places) of the result is assigned automatically to all operations except division, keeping the results as precise as possible.
+Results of a division may be inexpressible with a finite decimal (eg. 1 / 3 = 0.3333...), so it's up to the user to specify the requested scale.
+- scale of the result of `add` and `subtract` is the *maximal scale* of both money values
+- scale of the result of `multiply` is the *sum of scales* of both money values
+- scale of the result of `divide` must be *explicitly specified*, the last decimal place will be rounded to minimize the error
+
+*Note: The scale of the money amount is always preserved - `getAmount` will use all decimal places of its scale (eg. zero money with scale 6 would return `0.000000`).*
+
+### Rounding
+
+You may use `Money::round(int $scale) : Money` method that rounds the amount of money up to `$scale` decimal places, it rounds 0.5 away from zero (making 1.5 into 2 and -1.5 into -2).
+This behavior is consistent with `PHP_ROUND_HALF_UP` rounding mode, which is the default mode for [the `round` function](http://php.net/manual/en/function.round.php).
+
+The scale of the result will always be equal to the provided `$scale`.
+
+### Comparing
+
+To compare two monetary values you have to use the object's methods instead of [comparison operators](http://php.net/manual/en/language.operators.comparison.php):
+
+- `Money::equals(Money $other) : bool` instead of `===` and `==`
+- `Money::isLessThan(Money $other) : bool` instead of `<`
+- `Money::isGreaterThan(Money $other) : bool` instead of `>`
+- `Money::isLessThanOrEqualTo(Money $other) : bool` instead of `<=`
+- `Money::isGreaterThanOrEqualTo(Money $other) : bool` instead of `>=`
+- `Money::compare(Money $other) : int` instead of the Spaceship operator `<=>`
+
+To compare a value with zero you may use the short-hand methods`isPositive`, `isNegative` or `isZero`.
+A zero is considered neither positive nor negative.
+
+All methods compare the amounts of both objects counting all decimal places.
+
+## Money in Forms
+
+For the user input of monetary values use the `MoneyType` (see [Symfony docs](https://symfony.com/doc/3.4/reference/forms/types/money.html) for details).
+
+```php
+use Symfony\Component\Form\Extension\Core\Type\MoneyType;
+use Symfony\Component\Validator\Constraints\NotBlank;
+
+// ...
+
+$orderItemFormBuilder->add('priceWithVat', MoneyType::class, [
+    'scale' => 6,
+    'constraints' => [
+        new NotBlank(['message' => 'Please enter unit price with VAT']),
+    ],
+]);
+```
+
+The form type is configured with a model data transformer that converts the value into a `Money` object automatically ([`NumericToMoneyTransformer`](/packages/framework/src/Form/Transformers/NumericToMoneyTransformer.php)).
+Thanks to this approach you can use `Money` in your [data objects](/docs/introduction/entities.md#entity-data) directly.
+
+In Shopsys Framework, the default value of the `currency` option is `false` instead of `EUR`, hiding the currency symbol by default.
+
+*Note: For non-monetary numeric values use `NumberType` (see [Symfony docs](https://symfony.com/doc/3.4/reference/forms/types/number.html) for details).*
+
+### Form Constraints
+
+There are two form constraints to be used specifically with the `MoneyType` form fields.
+Because of model data transformation, you cannot use constraints that are validating scalar values (eg. `GreaterThan`).
+
+#### NotNegativeMoneyAmount
+
+Validates that the amount of money is greater or equal to zero.
+
+It has only the `message` option specifying the validation error message in case the entered value is negative.
+Specifying the validation message is optional, there is a default value.
+
+```php
+use Shopsys\FrameworkBundle\Form\Constraints\NotNegativeMoneyAmount;
+use Symfony\Component\Form\Extension\Core\Type\MoneyType;
+use Symfony\Component\Validator\Constraints\NotBlank;
+
+// ...
+
+$priceTableFormBuilder->add($key, MoneyType::class, [
+    'scale' => 6,
+    'required' => true,
+    'invalid_message' => 'Please enter price in correct format (a number with decimal separator)',
+    'constraints' => [
+        new NotBlank(['message' => 'Please enter price']),
+        new NotNegativeMoneyAmount(['message' => 'Price must be greater or equal to zero']),
+    ],
+]);
+```
+
+#### MoneyRange
+
+Similarly to [the Symfony `Range` constraint](https://symfony.com/doc/3.4/reference/constraints/Range.html), it validates that the amount of money is between some minimum and maximum.
+
+It has four options:
+- `min` specifies the minimum value, has to be an instance of `Money` or `null`
+- `max` specifies the maximum value, has to be an instance of `Money` or `null`
+- `minMessage` specifies the validation error message in case the entered value is less than the `min` value
+- `maxMessage` specifies the validation error message in case the entered value is greater than the `max` value
+
+At least one of the `min` and `max` options has to be provided for the constraint to make sense.
+Specifying the validation messages is optional, there are default values.
+
+```php
+use Shopsys\FormTypesBundle\MultidomainType;
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Shopsys\FrameworkBundle\Form\Constraints\MoneyRange;
+use Symfony\Component\Form\Extension\Core\Type\MoneyType;
+
+// ...
+
+$zboziFeedProductFormBuilder->add('cpc', MultidomainType::class, [
+    'label' => $this->translator->trans('Maximum price per click'),
+    'entry_type' => MoneyType::class,
+    'required' => false,
+    'entry_options' => [
+        'currency' => 'CZK',
+        'constraints' => [
+            new MoneyRange([
+                'min' => Money::create(1),
+                'max' => Money::create(500),
+            ]),
+        ],
+    ],
+]);
+```
+
+## Money in Twig Templates
+
+Instances of the `Money` class cannot be directly converted to strings.
+To be able to display a monetary value in a template you should use one of the prepared Twig filters.
+
+These filters can be used only with an instance of `Money`.
+
+### price
+
+Filter `price` formats the amount of money in a localized manner, including the currency symbol.
+It uses basic frontend currency on the current domain and the current locale (language) to format it.
+
+For example, one thousand Czech crowns would be rendered as *"CZK1,000.00"* on an English domain and as *"1 000,00 Kč"* on a Czech one.
+
+The filter expects the value to be already rounded.
+If not, it will render the extra decimal places.
+
+All `price*` Twig filters use this format.
+They usually differ only in the currency and locale they use.
+
+### priceText
+
+Filter `priceText` formats the amount of money in a localized manner, similarly to the `price` filter.
+The only difference is that it outputs the text *"Free"* (or the corresponding [translation](/docs/introduction/translations.md)) when zero amount of money is provided.
+
+### priceTextWithCurrencyByCurrencyIdAndLocale
+
+Filter `priceTextWithCurrencyByCurrencyIdAndLocale` can be used to format the amount of money in any currency and locale.
+It outputs the text *"Free"* when zero amount of money is provided.
+
+The *currency ID* (`int`) and the *locale* (`string`) must be provided as parameters.
+
+### priceWithCurrency
+
+Filter `priceWithCurrency` can be used to format the amount of money in any currency.
+It uses the current locale (the locale of current domain or administration) to format it.
+
+The *currency* (`Currency` entity) must be provided as a parameter.
+
+### priceWithCurrencyAdmin
+
+It works similarly to `priceWithCurrency`, but it uses the default administration currency.
+
+It has no parameters.
+
+### priceWithCurrencyByCurrencyId
+
+It works similarly to `priceWithCurrency`, but it uses currency ID instead of the `Currency` entity.
+
+The *currency ID* (`int`) must be provided as a parameter.
+
+### priceWithCurrencyByDomainId
+
+It works similarly to `priceWithCurrency`, but it uses the basic frontend currency of the provided domain.
+
+The *domain ID* (`int`) must be provided as a parameter.
+
+### moneyFormat
+
+Formats the amount of money as a decimal number without any currency symbol.
+
+Three optional parameters can be provided:
+- *number of decimal places* - `null` by default (meaning all), it will [round](#rounding) the value if necessary
+- *decimal point character* - `"."` by default
+- *separator of thousands* - `""` by default
+
+```twig
+{# the "money" variable contains Money::create('1234.5670') #}
+
+{{ money|moneyFormat }}                  {# renders "1234.5670" #}
+{{ money|moneyFormat(0) }}               {# renders "1235" #}
+{{ money|moneyFormat(2, ',') }}          {# renders "1234,57" #}
+{{ money|moneyFormat(null, '.', ',') }}  {# renders "1,234.5670" #}
+```
+
+## Money in Javascript
+
+`Money` implements the `JsonSerializable` interface, which means it can be serialized into JSON format with `json_encode` into an object with the following structure:
+
+```json
+{
+  "amount": "1234.5670"
+}
+```
+
+This format can be readily used in Javascript when JSON is rendered in Twig:
+
+```twig
+{# the "money" variable contains Money::create('1234.5670') #}
+
+<div id="my-money" data-money="{{ money|json_encode }}"/>
+```
+
+```js
+(function ($) {
+
+    $(document).ready(function () {
+        var money = $('#my-money').data('money');
+
+        // Displays "1234.5670"
+        alert(money.amount);
+    });
+
+})(jQuery);
+```
+
+See the native Javascript method [`.toFixed()`](https://www.w3schools.com/jsref/jsref_tofixed.asp) and the function [`parseFloat()`](https://www.w3schools.com/jsref/jsref_parsefloat.asp) to see how to convert between decimal strings and numbers.
+
+## Money in Doctrine
+
+In Shopsys Framework, there is a custom Doctrine type `money` which can be used in similar fashion as `decimal` type.
+The entity property value will be automatically hydrated to an instance of `Money` if it's configured to use the type:
+
+```php
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ */
+class MyEntity
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
+     *
+     * @ORM\Column(type="money", precision=20, scale=6, nullable=true)
+     */
+    protected $price;
+
+    // ...
+}
+```
+
+### In Parameters
+
+When you want to use a `Money` instance as a parameter in DQL, use the `getAmount` method in your [repository class](/docs/introduction/basics-about-model-architecture.md#repository):
+
+```php
+use Shopsys\FrameworkBundle\Component\Money\Money;
+
+// ...
+
+$priceLimit = Money::create(1000);
+
+/** @var \Doctrine\ORM\QueryBuilder $builder */
+$builder->setParameter('priceLimit', $priceLimit->getAmount());
+```
+
+### In Function Results
+
+Doctrine hydrates values to an instance of `Money` when selecting an entity property:
+
+```php
+use Shopsys\FrameworkBundle\Model\Order\Order;
+
+/** @var \Doctrine\ORM\EntityManager $em */
+$result = $em->createQuery('SELECT o.totalPriceWithVat FROM ' . Order::class . ' o WHERE o.id = :id')
+    ->setParameter('id', 1)
+    ->getSingleResult();
+
+// The result is automatically hydrated into Money
+$money = $result['totalPriceWithVat'];
+```
+
+But when you're working with aggregate functions (such as `MIN()`, `MAX()`, `SUM()`, etc.) Doctrine cannot infer the type.
+In such a case, you have to convert the fetched decimal string into a `Money` instance yourself using `Money::create()`:
+
+```php
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Shopsys\FrameworkBundle\Model\Order\Order;
+
+/** @var \Doctrine\ORM\EntityManager $em */
+$result = $em->createQuery('SELECT AVG(o.totalPriceWithVat) AS averagePriceWithVat FROM ' . Order::class . ' o')
+    ->getSingleResult();
+
+// The result of a function is a decimal string (eg. '19590.772727272727') and it must be converted manually
+$money = Money::create($result['averagePriceWithVat']);
+```
+
+## Unit and Functional Tests
+
+You should use `Money` for all monetary values even in tests.
+
+Create the instances via `Money::create(int|string)` in your data providers or directly in your test methods if you don't use providers.
+
+You can use a custom PHPUnit constraint `IsMoneyEqual` for an assertion that two monetary values are equal using the `assertThat($value, Constraint $constraint)` method.
+
+Example:
+
+```php
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Tests\FrameworkBundle\Test\IsMoneyEqual;
+use Tests\ShopBundle\Test\FunctionalTestCase;
+
+class MyTest extends FunctionalTestCase
+{
+    // ...
+
+    /**
+     * @return array
+     */
+    public function customerLoyaltyCreditAmountProvider(): array
+    {
+        return [
+            [self::CUSTOMER_ID_WITHOUT_LOYALTY_CREDIT, Money::zero()],
+            [self::CUSTOMER_ID_WITH_LOW_LOYALTY_CREDIT, Money::create('12.5')],
+            [self::CUSTOMER_ID_WITH_HIGH_LOYALTY_CREDIT, Money::create(1000)],
+        ];
+    }
+
+    /**
+     * @dataProvider customerLoyaltyCreditAmountProvider
+     * @param int $customerId
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money $expectedCreditAmount
+     */
+    public function testCustomerLoyaltyCreditAmount(int $customerId, Money $expectedCreditAmount) {
+        $customer = $this->getCustomerFromDatabase($customerId);
+
+        $creditAmount = $this->customerLoyaltyFacade->calculateCreditAmount($customer);
+
+        $this->assertThat($creditAmount, new IsMoneyEqual($expectedCreditAmount));
+    }
+}
+```
+
+## Price Class
+
+[`Price`](/packages/framework/src/Model/Pricing/Price.php) is also an immutable [value object](https://codete.com/blog/value-objects/) used in pricing.
+
+It represents a price with and without VAT and is used in many parts of Shopsys Framework.
+Price calculation classes usually output instances of `Price`.
+
+It can be constructed by calling `new Price(Money $priceWithoutVat, Money $priceWithVat)`.
+For a zero price, you can use a short-hand method `Price::zero()`.
+
+The class has three getters you can use to retrieve the prices or the VAT amount:
+- `Price::getPriceWithoutVat() : Money`
+- `Price::getPriceWithVat() : Money`
+- `Price::getVatAmount() : Money`
+
+And you can calculate with prices using its methods:
+- `Price::add(Price $addend) : Price`
+- `Price::subtract(Price $subtrahend) : Price`
+- `Price::inverse() : Price`
--- a/docs/upgrade/UPGRADE-v7.0.0.md
+++ b/docs/upgrade/UPGRADE-v7.0.0.md
@@ -20,8 +20,10 @@ if you want to have products data exported to Elasticsearch after `build-demo` t
     - dump translations using `php phing dump-translations` and fill in the translations based on the changes from pull request
 - check whether you extended class or method `ImageFacade::copyImages` or used it in your project and make sure it works like you intended ([#851](https://github.com/shopsys/shopsys/pull/851))
     - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, set as local filesystem storage in path `%kernel.project_dir%` by default
-- use `\Shopsys\FrameworkBundle\Component\Money\Money` class for representing monetary values in the whole application ([#821](https://github.com/shopsys/shopsys/pull/821))
-    - detailed upgrade instruction will be added shortly, see the PR for more information
+- use `Money` class for representing monetary values in the whole application ([#821](https://github.com/shopsys/shopsys/pull/821))
+    - we recommend first reading the article [How to Work with Money](/docs/introduction/how-to-work-with-money.md) which explains the concept
+    - you can take a look at [what was modified in `shopsys/project-base` during this change](https://github.com/shopsys/project-base/compare/cb6d02f335819aeff575dec01bda5b228263a2eb...c08cac7b55ebc46b43c2e988d36e2f122cbb4598#files_bucket) (24 files)
+    - you'll find detailed instructions in separate article [Upgrade Instructions for Money Class](/docs/upgrade/money-class.md)
 - you need to provide `$temporaryFilenames` parameter anywhere you use `ImageFactoryInterface::create()` and `ImageFacade::uploadImage()` functions ([#869](https://github.com/shopsys/shopsys/pull/869))
     - the parameter is not nullable now
 - if you're using protected method `ImageProcessor::removeFileIfRenamed()` in your code, remove the usage due to the method was removed and the code was moved to `convertToShopFormatAndGetNewFilename()` ([#869](https://github.com/shopsys/shopsys/pull/869))
--- a/docs/upgrade/money-class.md
+++ b/docs/upgrade/money-class.md
@@ -1,0 +1,467 @@
+# Upgrade Instructions for Money Class
+
+This article describes upgrade instructions for [#821 new class for representing monetary values](https://github.com/shopsys/shopsys/pull/821).
+Upgrade instructions are in a separate article because there is a lot of instructions and we don't want to jam the [UPGRADE-v7.0.0.md](/docs/upgrade/UPGRADE-v7.0.0.md).
+Follow these instructions only if you upgrade from `v7.0.0-beta6` to `v7.0.0`.
+
+All monetary values (*prices, account balances, discount amounts, price limits etc.*) should be represented by an instance of the class `\Shopsys\FrameworkBundle\Component\Money\Money`.
+
+Before the upgrade, we recommend reading the article [How to Work with Money](/docs/introduction/how-to-work-with-money.md) which explains the concept in detail.
+
+**When in doubt, you can take a look at [what was modified in `shopsys/project-base` during this change](https://github.com/shopsys/project-base/compare/cb6d02f335819aeff575dec01bda5b228263a2eb...c08cac7b55ebc46b43c2e988d36e2f122cbb4598#files_bucket).**
+
+## Configuration
+Register the new Doctrine type in `doctrine.dbal.types` in your `app/config/packages/doctrine.yml`:
+```diff
+         password: "%database_password%"
+         charset: UTF8
+         server_version: "%database_server_version%"
++        types:
++            money: \Shopsys\FrameworkBundle\Component\Doctrine\MoneyType
+
+     orm:
+         auto_generate_proxy_classes: false
+```
+
+## Extended and Custom Entities
+If you have created your custom entities that store monetary value or added a new such column [by entity extension](/docs/extensibility/entity-extension.md), change your `decimal` Doctrine type to the new `money` type:
+
+```diff
+  /**
+-  * @var string
+-  * @ORM\Column(type="decimal", precision=20, scale=6)
++  * @var \Shopsys\FrameworkBundle\Component\Money\Money
++  * @ORM\Column(type="money", precision=20, scale=6)
+   */
+  protected $myMonetaryValue;
+```
+
+If you do so, you'll have to create [a database migration](/docs/introduction/database-migrations.md) which will add a `(DC2Type:money)` comment on the columns, informing Doctrine about the specified type (similarly to the [`Version20190215092226`](/packages/framework/src/Migrations/Version20190215092226.php) in `shopsys/framework`).
+
+Because the entity property now changed the type, its usages (along with the related [entity data class](/docs/introduction/entities.md#entity-data)) have to be reviewed and changed as well.
+We recommend using type-hinting in the changed methods as explained in [Working With Money in PHP](#working-with-money-in-php) below.
+
+If you need to set a value to the property in these custom entities or data objects directly, you should use the `Money::create()` method that accepts a string or an integer (or another [construction method](/docs/introduction/how-to-work-with-money.md#construction)).
+
+For more details read [the Money in Doctrine section](/docs/introduction/how-to-work-with-money.md#money-in-doctrine) of the documentation.
+
+## Working With Money in PHP
+You should use `Money` for all monetary values across your project.
+We strongly recommend declaring the `Money` type for both [the method arguments](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration) and [the return values](http://php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration) in your methods.
+
+If you need to compute money, use its methods `add`, `subtract`, `multiply` and `divide` instead of the operators `+`, `-`, `*` and `/`.
+
+If you need to compare two instances of `Money`, use its methods `equals`, `isGreaterThan`, `isGreaterThanOrEqualTo`, `isLessThan`, `isLessThanOrEqualTo` and `compare` instead of the operators `==`/`===`, `>`, `>=`, `<`, `<=` and `<=>`.
+If you want to compare `Money` with zero, you can use the methods `isZero`, `isPositive` and `isNegative`.
+
+For more details read [the Money Class section](/docs/introduction/how-to-work-with-money.md#money-class) of the documentation.
+
+There were no changes needed in `shopsys/project-base` as most PHP implementation is in `shopsys/framework`.
+You might need to update your custom classes working with prices and overridden services, because of the changed types.
+
+*Note: Running the `phpstan` [Phing target](/docs/introduction/console-commands-for-application-management-phing-targets.md) will help you find out problems with incompatible method declarations mentioned in this section and in [Strict Typing](#strict-typing) below.*
+*In case of problems, it will fail on a `PHP Fatal error` with a reference to the incompatible method (unfortunately it is unable to continue with the checks as it cannot recover from this error).*
+
+## Strict Typing
+In the process of spreading the use of `Money` across the methods, we enabled [strict typing](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.strict) in some classes.
+A type declaration was added into all method parameters and return values when possible, so you can be sure about the accepted and expected variable types.
+Strict typing was also enabled in all newly created classes.
+
+Check how you use or extend these classes, into which the statement `declare(strict_types=1);` was added:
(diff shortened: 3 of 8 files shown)


docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
 - check whether you extended class or method `ImageFacade::copyImages` or used it in your project and make sure it works like you intended ([#851](https://github.com/shopsys/shopsys/pull/851))
     - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, set as local filesystem storage in path `%kernel.project_dir%` by default
 - use `\Shopsys\FrameworkBundle\Component\Money\Money` class for representing monetary values in the whole application ([#821](https://github.com/shopsys/shopsys/pull/821))
-    - detailed upgrade instruction will be added shortly, see the PR for more information
+    - if you have created your custom entities that store monetary value, change your `decimal` ORM type to `money` type
```
simara-svatopluk: There is huge amount of classes that changed interfaces. These classes can be extended or used and you have to provide instructions what classes and how to do it.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
...
+    ```diff
+    - * @var string
+    - * @ORM\Column(type="decimal", precision=20, scale=6)
+    + * @var \Shopsys\FrameworkBundle\Component\Money\Money
+    + * @ORM\Column(type="money", precision=20, scale=6)
+    ```
+        - update your custom data objects annotation type to `\Shopsys\FrameworkBundle\Component\Money\Money`
```
simara-svatopluk: formatting is strange here, but it would be great if you added `diff` example here as well

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
...
+    - if you are creating value for these custom entities or data objects directly, you have to use `Money::create` method that accepts string or integer or `Money::createFromFloat` method that accepts float and has second mandatory parameter scale
```
simara-svatopluk: > if you are creating value for these custom entities

I don't understand what is a "value", please rephrase this sentence

simara-svatopluk: please explain what is the "scale", maybe by the link to the documentation

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
...
+    - if you need to compute mathematical operations use `Money::add`, `Money::substract`, `Money::multiply` or `Money::divide` method
```
simara-svatopluk: not mathematical operations... maybe "if you need to compute money, use ..."

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
...
+    - if you need to compare two `Money` use `compare`, `equals`, `isGreaterThan`, `isGreaterThanOrEqualTo`, `isLessThan` or `isLessThanOrEqualTo` method
+    - if you need to use `Money` in javascript, it is automatically serialized to JSON object with `amount` parameter
```
simara-svatopluk: please rephrase this upgrade note to "what the implementator must do in JS" - there will be more work because the result of serialization to JS is not a single value but JSON object. And you should explain what to do (not what is done)

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
...
+    - `MoneyType` form type has been updated, it now works with `Money` class and uses `NumericToMoneyTransformer`, if you are using it you have to:
+        - use one of these two new constraints:
+            - `Shopsys\FrameworkBundle\Form\Constraints\MoneyRange` for setting acceptable range (only one side of interval is required)
+            - `Shopsys\FrameworkBundle\Form\Constraints\NotNegativeMoneyAmount` for setting that entered amount of money has to be zero or higher
+            - if non of these constraints is sufficient for your use case, create your own ([see commit with ne MoneyRange constraint](https://github.com/shopsys/shopsys/pull/821/commits/390b21416e54ca8bf34a7f7deec4e5c3eed5cd51))
```
simara-svatopluk:
```suggestion
            - if non of these constraints is sufficient for your use case, create your own ([see commit with MoneyRange constraint](https://github.com/shopsys/shopsys/pull/821/commits/390b21416e54ca8bf34a7f7deec4e5c3eed5cd51))
```

simara-svatopluk: Please don't make links to a specific commit. The commit hash could change once the commit is rebased (luckily not this case, but in general it is a true)

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
...
+        - if you are using `MoneyType` for non monetary values, change it to `NumberType` instead
+        - currency is no longer mandatory option, default value is `false`
+    - many classes working with `Money` now use strict types, run `php phing phplint` to find out if there are any problems
+    - there is new twig filter `moneyFormat` that accepts values for count of decimal places, format of decimal point etc, use it in cases that you need to display `Money` value without currency symbol
```
simara-svatopluk:
```suggestion
    - there is new twig filter `moneyFormat`, use it in cases that you need to display `Money` value without currency symbol
```
when there is a parameter description, the sentence doesn't make sense for me

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -16,7 +16,47 @@ There you can find links to upgrade notes for other versions too.
...
+    use PHPUnit\Framework\TestCase;
+    use Shopsys\FrameworkBundle\Component\Money\Money;
+    use Tests\FrameworkBundle\Test\IsMoneyEqual;
+
+    class SomeModelTest extends TestCase
+    {
+
+        public function testSomeFunctionality()
+        {
+            $moneyA = Money::create(1);
```
simara-svatopluk:
```suggestion
            $expected = Money::create(2);
            $money = Money::create(1);
            $actual = $money->multiply(2);
            
            $this->assertThat($actual, new IsMoneyEqual($moneyB));
```
Maybe a better example

docs/upgrade/UPGRADE-unreleased.md
simara-svatopluk: Please add note about changed twig filters `price*`

docs/upgrade/UPGRADE-unreleased.md
PetrHeinz: Why not link the specific commit in this case then? It won't ever change as it is merged to master already and it is a perfect working example of how such a constraint is implemented (with JS validation as well).

TomasLudvik:
> Please add also instructions for upgrading all changed files in `project-base`
> eg. doctrine.yml, DataFixtures, ProductFilterFormType, twigs, and tests (but check their list by yourself too)
> 
> There is huge amount of classes that changed interfaces. These classes can be extended or used so please provide instructions how to deal with changes for all classes. To clarify my expectation, I would like to see something similar to https://github.com/shopsys/shopsys/blob/master/docs/upgrade/services-removal.md
> And as there will be a lots of instructions, I think you can create a separate article for them

I have mentioned all important with upgrade information, tests and datafixtures are mentioned to be reviewed, methods that changed input parameters have been metioned, that they have been changed and all other methods whose typehints has been changed should be covered with section run php phing phplint.

docs/upgrade/UPGRADE-unreleased.md
TomasLudvik: I have fixed formatting, I do not want to add diff there because it can be from string and decimal...

docs/upgrade/UPGRADE-unreleased.md
TomasLudvik: I have updated it

docs/upgrade/UPGRADE-unreleased.md
TomasLudvik: updated

docs/upgrade/UPGRADE-unreleased.md
TomasLudvik: I do not want to rephrase it, because it is not necessary... but I have added usage from one of our js files so it is more clear

docs/upgrade/UPGRADE-unreleased.md
TomasLudvik: updated

docs/upgrade/UPGRADE-unreleased.md
TomasLudvik: updated

docs/upgrade/UPGRADE-unreleased.md
TomasLudvik: added

PetrHeinz:
I have rebased the branch on current `master` which was needed after the `7.0.0` release.

The documentation part is ready for review, the upgrade notes still need revision and some changes.

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+For example, one thousand Czech crowns would be rendered as *"CZK1,000.00"* on an English domain and as *"1 000,00 Kč"* on a Czech one.
+
+All `price*` Twig filters use this format.
+They usually differ only in the currency and locale they use.
+
+### priceText
+
+Filter `priceText` formats the amount of money in a localized manner, similarly to the `price` filter.
+The only difference is that it outputs the text *"Free"* when zero amount of money is provided.
```
vitek-rostislav:
```suggestion
The only difference is that it outputs the text *"Free"* (or the corresponding translation) when zero amount of money is provided.
```

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+### priceWithCurrencyAdmin
+
+It works similarly to `priceWithCurrency`, but it uses the default administration currency.
+
+It has no parameters.
+
+### priceWithCurrencyByCurrencyId
+
+It works similarly to `priceWithCurrency`, but it uses currency ID instead of the `Currency` entity.
```
vitek-rostislav: I think we should create some US for simplification (reduction of the amount of these filters) :thinking:

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+     *
+     * @ORM\Column(type="money", precision=20, scale=6, nullable=true)
+     */
+    protected $price;
+
+    // ...
+}
+```
+
+When you want to use a `Money` instance as a parameter in DQL, use the `getAmount` method in your [repository class](/docs/introduction/basics-about-model-architecture.md#repository):
```
vitek-rostislav: I do not want to be a nitpick, however, I would expect this to work automatically provided that `ProductCalculatedPrice::$priceWithVat` is `Money` type :slightly_smiling_face:

docs/upgrade/UPGRADE-unreleased.md
simara-svatopluk: I expresses myself wrongly... It is a bad idea to link a specific commit because it caused us already couple of problems because we rebase a lot.

In this particular case it is ok

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+It uses a decimal representation of the money amount and it does not contain any reference to the used currency.
+You can get the decimal representation as a `string` via the `getAmount` method.
+
+*Note: If in doubt about the results of any method, you can take a look at [its unit tests](/packages/framework/tests/Unit/Component/Money/MoneyTest.php) which contain many examples of the class' behavior.*
+
+### Construction
+
+`Money` can be constructed directly from integers and numeric strings as they are able to represent decimal numbers precisely: `$tenDollars = Money::create(10)`, `$oneAndHalfEuro = Money::create('1.20')`.
+
+It may be also constructed from floating point numbers, but the scale (number of decimal places) must be explicitly specified: `Money::createFromFloat($floatFromExternalSource, 2)`.
```
simara-svatopluk: Please change the example to eg. `Money::createFromFloat(1.88888, 2)` and explain that the `scale` will truncate (or round?) to `1.88` (or `1.89`).

I see that there is explained what is `scale` in the article, but it is after the first usage. Please explain first what is the `scale` and then use it. Or you can explain the `scale` directly under this sentence.

The explanaion of the `scale` should contain more examples and should be also focused on the `concept`, not so much on methods and implementation. I think it could be also useful to separate a section `Scale` because it is pretty tricky concept.

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+### Computation
+
+To compute with monetary values you have to use the object's methods instead of arithmetic operators (`+`, `-`, `*`, `/`):
+
+- `Money::add(Money $addend) : Money`
+- `Money::subtract(Money $subtrahend) : Money`
+- `Money::multiply(int|string $multiplier) : Money`
+- `Money::divide(int|string $divisor, int $scale) : Money`
+
```
simara-svatopluk: Please explain here that all these operation creates a new object and don't modify the original one. The immutability is not commonly spread knowledge and readers can misunderstand the usage.

Also you can link the `value object` article again here :-)

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+The scale (number of decimal places) of the result is assigned automatically to all operations except division, keeping the results as precise as possible.
+Results of a division may be inexpressible with a finite decimal (eg. 1 / 3 = 0.3333...), so it's up to the user to specify the requested scale.
+- scale of the result of `add` and `subtract` is the *maximal scale* of both money values
+- scale of the result of `multiply` is the *sum of scales* of both money values
+- scale of the result of `divide` must be *explicitly specified*, the last decimal place will be rounded to minimize the error
+
+*Note: The scale of the money amount is always preserved - `getAmount` will use all decimal places of its scale (eg. zero money with scale 6 would return `0.000000`).*
+
+### Rounding
```
simara-svatopluk: Please add 4 examples of rounding here.
* 1.494
* 1.055
* and both with 2 different scales - 1 and 2

simara-svatopluk: scale is really a tricky concept :-/

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+To compare two monetary values you have to use the object's methods instead of [comparison operators](http://php.net/manual/en/language.operators.comparison.php):
+
+- `Money::equals(Money $other) : bool` instead of `===` and `==`
+- `Money::isLessThan(Money $other) : bool` instead of `<`
+- `Money::isGreaterThan(Money $other) : bool` instead of `>`
+- `Money::isLessThanOrEqualTo(Money $other) : bool` instead of `<=`
+- `Money::isGreaterThanOrEqualTo(Money $other) : bool` instead of `>=`
+- `Money::compare(Money $other) : int` instead of the Spaceship operator `<=>`
+
+To compare a value with zero you may use the short-hand methods`isPositive`, `isNegative` or `isZero`.
```
simara-svatopluk: Maybe a note "beware zero itself is not positive nor negative"

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+`Money` is an immutable [value object](https://codete.com/blog/value-objects/).
+
+It uses a decimal representation of the money amount and it does not contain any reference to the used currency.
+You can get the decimal representation as a `string` via the `getAmount` method.
+
+*Note: If in doubt about the results of any method, you can take a look at [its unit tests](/packages/framework/tests/Unit/Component/Money/MoneyTest.php) which contain many examples of the class' behavior.*
+
+### Construction
+
+`Money` can be constructed directly from integers and numeric strings as they are able to represent decimal numbers precisely: `$tenDollars = Money::create(10)`, `$oneAndHalfEuro = Money::create('1.20')`.
```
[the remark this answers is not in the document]
PetrHeinz: I did want to communicate that $10 may be represented by `Money::create(10)` instead of `Money::create(1000)` (sometimes money is created by the smallest subunit - eg. cents - and I wanted to clear that out with an example).

You're definitely onto something with that `1.5 != 1.2` issue you mention though :smile:

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+Instances of the `Money` class cannot be directly converted to strings.
+To be able to display a monetary value in a template you should use one of the prepared Twig filters.
+
+These filters can be used only with an instance of `Money`.
+
+### moneyFormat
+
+Formats the amount of money as a decimal number without any currency symbol.
+Optional parameters can be provided to specify the *number of decimal places* (`null` by default, meaning all), the *decimal point character* (`"."` by default) and the *separator of thousands* (`""` by default).
```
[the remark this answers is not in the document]
PetrHeinz: A list might help here :slightly_smiling_face:

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+$orderItemFormBuilder->add('priceWithVat', MoneyType::class, [
+    'scale' => 6,
+    'constraints' => [
+        new NotBlank(['message' => 'Please enter unit price with VAT']),
+    ],
+]);
+```
+
+The form type is configured with a model data transformer that converts the value into a `Money` object automatically (`\Shopsys\FrameworkBundle\Form\Transformers\NumericToMoneyTransformer`).
```
simara-svatopluk: I think that the `Tranformer` class takes too much space here. Please change it to `NumericToMoneyTransformer` with link to the code.

It would be helpful in my opition to show that the form type really returns the `Money` instance and takes as an input the `Money` instance by an example (but I have no idea how the example should look like :-D).

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+#### MoneyRange
+
+Similarly to [the Symfony `Range` constraint](https://symfony.com/doc/3.4/reference/constraints/Range.html), it validates that the amount of money is between some minimum and maximum.
+
+It has four options:
+- `min` specifies the minimum value, has to be an instance of `Money` or `null`
+- `max` specifies the maximum value, has to be an instance of `Money` or `null`
+- `minMessage` specifies the validation error message in case the entered value is less than the `min` value
```
simara-svatopluk: Please clarify these options are optional and they have default value

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+use Shopsys\FrameworkBundle\Form\Constraints\MoneyRange;
+use Symfony\Component\Form\Extension\Core\Type\MoneyType;
+
+// ...
+
+$zboziFeedProductFormBuilder->add('cpc', MultidomainType::class, [
+    'label' => $this->translator->trans('Maximum price per click'),
+    'entry_type' => MoneyType::class,
+    'required' => false,
+    'entry_options' => [
```
simara-svatopluk: I think it would be nice to have here one of `{min/max}Message` to ilustrate where the option belongs

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+## Money in Twig Templates
+
+Instances of the `Money` class cannot be directly converted to strings.
+To be able to display a monetary value in a template you should use one of the prepared Twig filters.
+
+These filters can be used only with an instance of `Money`.
+
+### moneyFormat
```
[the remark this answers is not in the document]
simara-svatopluk: Please move this filter under `price*` filters because it is special - doesn't use currency. And also it is less used than `price*` filters

docs/introduction/how-to-work-with-money.md
PetrHeinz: We've expected that too, that's why I've included it in the docs. There are possibilities how to do that in Doctrine, but you either have to implement `__toString` in your value object (which we wanted to avoid), or you have to call `
Type::getType('money')->convertToDatabaseValue($money, $platform)`, but that requires having an instance of `\Doctrine\DBAL\Platforms\AbstractPlatform` and is rather complicated...

Setting it as a parameter and getting it out of aggregate functions (such as `MAX()` or `SUM()`) are two issues with Doctrine we run into. Converting it to/from `Money` in a repository is a decent solution though.

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+{{ money|moneyFormat }}                  {# renders "1234.5670" #}
+{{ money|moneyFormat(0) }}               {# renders "1235" #}
+{{ money|moneyFormat(2, ',') }}          {# renders "1234,57" #}
+{{ money|moneyFormat(null, '.', ',') }}  {# renders "1,234.5670" #}
+```
+
+### price
+
+Filter `price` formats the amount of money in a localized manner.
```
simara-svatopluk: please add examples into all filters with rendered outputs. As same as you do with `moneyFormat`
It is really helpful for reader to imagine what is going on.

docs/introduction/how-to-work-with-money.md
PetrHeinz: Maybe we should document (or at least mention) working with aggregate functions as well...

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+    });
+
+})(jQuery);
+```
+
+See the native Javascript method [`.toFixed()`](https://www.w3schools.com/jsref/jsref_tofixed.asp) and the function [`parseFloat()`](https://www.w3schools.com/jsref/jsref_parsefloat.asp) to see how to convert between decimal strings and numbers.
+
+## Money in Doctrine
+
+In Shopsys Framework, there is a custom Doctrine type `money` which can be used in similar fashion as `decimal` type.
```
simara-svatopluk:
```suggestion
We use a custom Doctrine type `money` which can be used in similar fashion as `decimal` type.
```

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+/**
+ * @ORM\Table(name="my_entities")
+ * @ORM\Entity
+ */
+class MyEntity
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
+     *
+     * @ORM\Column(type="money", precision=20, scale=6, nullable=true)
```
simara-svatopluk: The example is quite long and the most important part `type="money"` looses in the context. Please remove class, uses and keep here only the field, so it will be highlighted.
I know it won't be 100% correct, but I prefer simplicity over correctness ;-)

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+- [Money in Javascript](#money-in-javascript)
+- [Money in Doctrine](#money-in-doctrine)
+- [Unit and Functional Tests](#unit-and-functional-tests)
+- [Price class](#price-class)
+
+## Money class
+
+`Money` is an immutable [value object](https://codete.com/blog/value-objects/).
+
+It uses a decimal representation of the money amount and it does not contain any reference to the used currency.
```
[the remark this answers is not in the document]
PetrHeinz: I didn't want to get into it in the documentation, but basically currency and monetary values are different concepts in Shopsys Framework at the moment. Changing that would be pretty hard and there are some technical hurdles in our way, but we want to do it in the future.

I tried explaining that for a bit, it sounded quite confusing so I've just erased it.

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * @param \Doctrine\ORM\QueryBuilder $builder
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money $limit
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup $group
+     */
+    protected function filterByMinimalPrice(QueryBuilder $builder, Money $limit, PricingGroup $group)
+    {
+        $condition = 'pcp.product = p AND pcp.pricingGroup = :group AND pcp.priceWithVat >= :limit';
+
+        $productsQueryBuilder->join(ProductCalculatedPrice::class, 'pcp', Join::WITH, $condition)
```
simara-svatopluk: the example is too long. Please use only the `filterByMinimalPrice` method and remove `$groupParameter`.
You can maybe add link to actual usage for correctness

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+        $creditAmount = $this->customerLoyaltyFacade->calculateCreditAmount($customer);
+
+        $this->assertThat($creditAmount, new IsMoneyEqual($expectedCreditAmount));
+    }
+}
+```
+
+## Price class
+
+`\Shopsys\FrameworkBundle\Model\Pricing\Price` is also an immutable [value object](https://codete.com/blog/value-objects/) used in pricing.
```
simara-svatopluk: Please add link to code

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: Oh, I've missed that, thanks!

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+It represents a price with and without VAT and is used in many parts of Shopsys Framework.
+Price calculation classes (except `InputPriceCalculation`) output instances of `Price`.
```
simara-svatopluk: You can maybe remove the sentence... I don't see added value and it also contains an exception that forces me to think about calculators while I should focus on `Price`

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+It can be constructed by calling `new Price(Money $priceWithoutVat, Money $priceWithVat)`.
+For a zero price, you can use a short-hand method `Price::zero()`.
+
+The class has three getters:
```
simara-svatopluk: I don't see the point of explaining the interface. I'm missing something like ".... so you can...." after the method explanation

docs/introduction/how-to-work-with-money.md
PetrHeinz: I've also added a link to the Translations article

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+
+**Table of Contents:**
+- [Money class](#money-class)
+- [Money in Forms](#money-in-forms)
+- [Money in Twig Templates](#money-in-twig-templates)
+- [Money in Javascript](#money-in-javascript)
+- [Money in Doctrine](#money-in-doctrine)
+- [Unit and Functional Tests](#unit-and-functional-tests)
+- [Price class](#price-class)
+
```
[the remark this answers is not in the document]
simara-svatopluk: After a short break I think the `scale` concept should be explained at the top of the article. It it a tricky concept and I would appreciate conceptual explanation with examples. Like if you divide 1/3, you have to specify the scale and if you pick 2, the result will be `1.33`. You have to specify scale for money creating and if you create money from 1.494 it has different results for scale 1 - `1.5` and 2 - `1.49`. Also explain here that scale rounds the number and how

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: Added in 6926788dd.

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: Added in 6926788dd, it looks much better imo.

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: @simara-svatopluk: I've mentioned it first because it's simpler and it might be overlooked when at the end.

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
+# How to Work with Money
+
+Money is a very important concept for every e-commerce project.
```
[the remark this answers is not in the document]
PetrHeinz: I'm convinced an article must be there and Grammarly agrees with me:
image

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,388 @@
...
+In Shopsys Framework, all monetary values (*prices, account balances, discount amounts, price limits etc.*) are represented by an instance of [the `Money` class](/packages/framework/src/Component/Money/Money.php).
+
+This approach has several advantages:
+- it avoids problems with floating point number calculations and comparisons (see [official PHP documentation](http://php.net/manual/en/language.types.float.php) for details)
+- allows easy-to-use interfaces with consistent type-hinting so you can be sure what type of value you should be using
+- prevents accidental conversion to unexpected types (which may be problematic eg. when using the `===` operator)
+- makes the application design clearer and future changes easier
+
+**Table of Contents:**
```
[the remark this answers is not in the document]
PetrHeinz: I was afraid it is pretty long and I've found the short index helpful. Originally I was considering a complete index, but it was definitely too long and complicated (especially with the Twig filters).

docs/introduction/how-to-work-with-money.md
PetrHeinz: Addressed in 199b37562

docs/introduction/how-to-work-with-money.md
PetrHeinz: Addressed in 4664094d0

docs/introduction/how-to-work-with-money.md
PetrHeinz: This is an example straight from the actual code. It wouldn't make much sense to specify only one message (as both `min` and `max` are used). I could use an entirely different example to show how to use it (eg. specify only `min` and `minValue`) but I don't see much value.

In my opinion, people are used to the `Range` constraint and the usage is pretty straight-forward.

docs/introduction/how-to-work-with-money.md
PetrHeinz: Added a note 53dca3b68

docs/introduction/how-to-work-with-money.md
PetrHeinz: Why?

docs/introduction/how-to-work-with-money.md
PetrHeinz: I consider this out of scope of this PR. I've already incuded an example in the text:

> For example, one thousand Czech crowns would be rendered as *"CZK1,000.00"* on an English domain and as *"1 000,00 Kč"* on a Czech one.

There is a lot of price filters and the output depends on global configuration, current domain and whether you are in administration or not.

docs/introduction/how-to-work-with-money.md
PetrHeinz: There is no "we" or "us" in this article and I'd like to keep it consistent.

docs/introduction/how-to-work-with-money.md
PetrHeinz: I've removed the `@ORM\Table` in 504fc6b94 which wasn't required. I think it's important to note the alias `use Doctrine\ORM\Mapping as ORM;` even though it's common in our code and the fact that I'm talking about entities.

docs/introduction/how-to-work-with-money.md
PetrHeinz: I've already severely shortened the example. Having it shorter may hurt actual understandability in my opinion. I'm not sure in which context it might work without `PricingGroup`.

Would this be better?

```php
use Shopsys\FrameworkBundle\Component\Money\Money;

// ...

$priceLimit = Money::create(1000);

/** @var \Doctrine\ORM\QueryBuilder $builder */
$builder->setParameter('priceLimit', $priceLimit->getAmount());
```

docs/introduction/how-to-work-with-money.md
PetrHeinz: Fixed in 2f053a2a0

docs/introduction/how-to-work-with-money.md
PetrHeinz: Well, it's probably the most important usage, I think the mention is deserved.

docs/introduction/how-to-work-with-money.md
PetrHeinz: Fixed in d366d8659

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: The scale might deserve its own section, but I wouldn't want to explain the WHATs and WHYs any further. I think that the introductory section covers it sufficiently in the scope of this PR:

> Money is a very important concept for every e-commerce project. In Shopsys Framework, all monetary values (prices, account balances, discount amounts, price limits etc.) are represented by an instance of the Money class.
>
> This approach has several advantages:
>
> - it avoids problems with floating point number calculations and comparisons (see official PHP documentation for details)
> - allows easy-to-use interfaces with consistent type-hinting so you can be sure what type of value you should be using
> - prevents accidental conversion to unexpected types (which may be problematic eg. when using the === operator)
> - makes the application design clearer and future changes easier

That currency is not included in the value object is explained in "Money class" section:

> It uses a decimal representation of the money amount and it does not contain any reference to the used currency.

I had a hard time explaining the resons in a clear way because in the end, we'd actually like to include it but we could not implement it in the possible timeframe (https://github.com/shopsys/shopsys/issues/745#issuecomment-460635377). We've discussed this with @vitek-rostislav in https://github.com/shopsys/shopsys/pull/870#discussion_r261502306 as well. Do you have any ideas how to explain it without breaking the flow or getting into nitty gritty details?

docs/introduction/how-to-work-with-money.md
PetrHeinz: The Scale section might be a good addition to the article.

docs/introduction/how-to-work-with-money.md
PetrHeinz: Do you really think that rounding needs more explanation?

I've added a note with a link to more examples:
```diff
- This behavior is consistent with `PHP_ROUND_HALF_UP` rounding mode.
+ This behavior is consistent with `PHP_ROUND_HALF_UP` rounding mode, which is the default mode for [the `round` function](http://php.net/manual/en/function.round.php).
```

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
simara-svatopluk: **Suggestion:**

## Concept

The money concept in SSFW represents and encapsulates money values with decimal part, like `100`, `0.5`, `10.99`, `0.0005`. Money are representation without currency (if you need to pass also information about currency through the system, you need a separated concept).

### Scale
Scale is a precision after decimal part and it is a bit tricky concept.

Imagine you want to represent `1/3`, in float it is actually repsesented as `0.3333333432674407958984375` (because of precision (link)). When you want to work with the `1/3` in terms of money, you have to specify "amount of places after decimal point". So you can create a money from `1/3` in scale of `2` -> `0.33`, or in scale of `8` -> `0.33333334`

 Scale have to be specified during rounding (link), creating from floats (link) or dividing (link).

docs/introduction/how-to-work-with-money.md
simara-svatopluk: This is better, but there is problem with terms. We are using term `scale`, but in php documentation there is used term `precision`, so it is difficult to understand what do you want to express.
I would really appreciate as same examples as in the php documentation http://php.net/manual/en/function.round.php#refsect1-function.round-examples

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
simara-svatopluk: By mentioning it first you are also telling that it is the most important.
`price*` filter are more common. `moneyFormat` filter is used in feeds, for JS. `price*` filters are used for humans. So I think `price*` filters are more important

docs/introduction/how-to-work-with-money.md
simara-svatopluk: yep, you mentioned it... but without examples i don't know how to use filters and what will the the output

docs/introduction/how-to-work-with-money.md
simara-svatopluk: see the difference
```php

/**
 * @ORM\Entity
 */
class MyEntity
{
    /**
     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
     *
     * @ORM\Column(type="money", precision=20, scale=6, nullable=true)
     */
    protected $price;

    // ...
}
```

```php
// class MyEntity

    /**
     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
     *
     * @ORM\Column(type="money", precision=20, scale=6, nullable=true)
     */
    protected $price;
```

shorter (half the length), maybe not so correct, but highlighting the most important part

docs/introduction/how-to-work-with-money.md
simara-svatopluk: much better :+1:

docs/introduction/how-to-work-with-money.md
simara-svatopluk: Ok, then I would suggest to remove the exception. As I said, it forces me to think about calculators while I should focus on `Price`... the exception interupts flow

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: Yeah, I think think you're right. Changed in 39cfd6f7a.

docs/introduction/how-to-work-with-money.md
PetrHeinz: I've tried to explain it in f446f4082

docs/introduction/how-to-work-with-money.md
PetrHeinz: Replaced in 1b497aec1

docs/introduction/how-to-work-with-money.md
PetrHeinz: That's a fair point, I've removed the specific exception in 96d46fb74

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: I've added the section in aa1e1900c

docs/introduction/how-to-work-with-money.md
PetrHeinz: I think it's sufficient. I don't want to include that much examples, if you don't mind.

docs/introduction/how-to-work-with-money.md
PetrHeinz: Due to the dependency on configuration, domain and whatnot I'd rather keep it as-is.,

docs/introduction/how-to-work-with-money.md
PetrHeinz: I think the verbosity is not such a big deal here and I'd like to keep it as it is. It better communicates the concept of the entity class and it explains the `@ORM\Column` annotation.

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
PetrHeinz: I think that the list of advantages and the new General Concept section answer it sufficiently.

docs/introduction/how-to-work-with-money.md
simara-svatopluk: I'm sorry but I cannot find the link... In which file is it?

docs/introduction/how-to-work-with-money.md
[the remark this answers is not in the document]
simara-svatopluk: I was just wondering why there is the article `a` in the sentence `... is a very ...`. Nevermind, not important

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,428 @@
...
+- [Money Class](#money-class)
+- [Money in Forms](#money-in-forms)
+- [Money in Twig Templates](#money-in-twig-templates)
+- [Money in Javascript](#money-in-javascript)
+- [Money in Doctrine](#money-in-doctrine)
+- [Unit and Functional Tests](#unit-and-functional-tests)
+- [Price Class](#price-class)
+
+## General Concept
+The money concept in Shopsys Framework represents and encapsulates monetary values with a decimal part, like `100`, `0.50`, `10.99`, `0.0005`...
```
simara-svatopluk:
```suggestion
The money concept in Shopsys Framework represents and encapsulates monetary values with a decimal part, like `100`, `0.50`, `10.99`, `0.0005`, ...
```

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,428 @@
...
+Money is represented without currency.
+
+### Scale
+Scale defines the precision of the decimal part and it can be a bit tricky.
+
+Imagine you want to represent `1/3` (one third) in your application. In a `float`, it would be actually represented as `0.333333333333333314829616256247390992939472198486328125` because of [the floating-point precision](http://php.net/manual/en/language.types.float.php).
```
simara-svatopluk: Please add a newline after the sentence with `0.33333333333333333....` to visually separate "problem with floats" from "money scale"

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,428 @@
...
+When you want to work with one third in terms of money, you have to specify the scale - the number of places after the decimal point that should be taken into account.
+So you can create a monetary value from `1/3` in the scale of 2 (`0.33`), or in the scale of 8 (`0.33333333`).
+But it will never be exactly one third as it is inexpressible using a finite decimal.
+
+The money concept keeps the computation and comparisons precise up to the defined scale.
+Eg. if you multiply `1/3` with the scale of 4 (`0.3333`) by `3` you'll get `0.9999`, not `1`.
```
simara-svatopluk:
```suggestion
Eg. if you have `1/3` with the scale of 4 (`0.3333`) and multiply it by `3` you'll get `0.9999`, not `1`.
```

docs/introduction/how-to-work-with-money.md
```diff
@@ -0,0 +1,428 @@
...
+// ...
+
+$priceLimit = Money::create(1000);
+
+/** @var \Doctrine\ORM\QueryBuilder $builder */
+$builder->setParameter('priceLimit', $priceLimit->getAmount());
+```
+
+### In Function Results
+
```
simara-svatopluk: I suggest to split the text into 2 parts together with examples (quick suggestion, please review text and usages):

Doctrine hydrates values to an instance of `Money`.
```php
/** @var \Doctrine\ORM\EntityManager $em */
$result = $em->createQuery('SELECT o.totalPriceWithVat FROM ' . Order::class . ' o WHERE o.id = :id')
    ->setParameter('id', 1)
    ->getSingleResult();
// The result is automatically hydrated into Money
$money = $result['totalPriceWithVat'];
```

But once you're working with aggregate functions such as `MIN()`, `MAX()`, `SUM()`, etc., Doctrine have no idea about the type.
When fetching a monetary value from a function, you have to convert the fetched decimal string into a `Money` instance yourself using `Money::create()`:
```php
$result = $em->createQuery('SELECT AVG(o.totalPriceWithVat) AS averagePriceWithVat FROM ' . Order::class . ' o')
    ->getSingleResult();
// The result of a function is a decimal string (eg. '19590.772727272727') and it must be converted manually
$money = Money::create($result['averagePriceWithVat']);
```

docs/upgrade/UPGRADE-v7.0.0.md
```diff
@@ -20,8 +20,10 @@ if you want to have products data exported to Elasticsearch after `build-demo` t
     - dump translations using `php phing dump-translations` and fill in the translations based on the changes from pull request
 - check whether you extended class or method `ImageFacade::copyImages` or used it in your project and make sure it works like you intended ([#851](https://github.com/shopsys/shopsys/pull/851))
     - *(low priority)* remove `/var/www/html/var/cache` folder from your `@main_filesystem` filesystem storage if exists, set as local filesystem storage in path `%kernel.project_dir%` by default
-- use `\Shopsys\FrameworkBundle\Component\Money\Money` class for representing monetary values in the whole application ([#821](https://github.com/shopsys/shopsys/pull/821))
-    - detailed upgrade instruction will be added shortly, see the PR for more information
+- use `Money` class for representing monetary values in the whole application ([#821](https://github.com/shopsys/shopsys/pull/821))
+    - we recommend first reading the article [How to Work with Money](/docs/introduction/how-to-work-with-money.md) which explains the concept
+    - you'll find detailed instructions in separate article [Upgrade Instructions for Money Class](/docs/upgrade/money-class.md)
```
simara-svatopluk: separating to article is great and link to changes in project-base also!

docs/introduction/how-to-work-with-money.md
PetrHeinz: Added in ce5370814

docs/introduction/how-to-work-with-money.md
PetrHeinz: Even though it's up to personal preference, your suggestion is more consistent with current style.

docs/introduction/how-to-work-with-money.md
PetrHeinz: Nice idea

docs/introduction/how-to-work-with-money.md
PetrHeinz: I've split it in 4880c8fcf

docs/introduction/how-to-work-with-money.md
PetrHeinz: @simara-svatopluk search for "or the corresponding translation"  in https://github.com/shopsys/shopsys/blob/ph-tl-money-6/docs/introduction/how-to-work-with-money.md

simara-svatopluk: great thanks

docs/upgrade/money-class.md
```diff
@@ -0,0 +1,430 @@
...
+         server_version: "%database_server_version%"
++        types:
++            money: \Shopsys\FrameworkBundle\Component\Doctrine\MoneyType
+
+     orm:
+         auto_generate_proxy_classes: false
+```
+
+## Extended and Custom Entities
+If you have created your custom entities that store monetary value or added a new such column [by entity extension](/docs/extensibility/entity-extension.md), change your `decimal` Doctrine type to the new `money` type:
```
simara-svatopluk: Extremely important note™ ;-)

Double dots in the end of the line is quite disturbing for me, I see you are using it consistently, but I would prefer to omit it everywhere

docs/upgrade/money-class.md
```diff
@@ -0,0 +1,430 @@
...
+
+If you need to compute money, use its methods `add`, `subtract`, `multiply` and `divide` instead of the operators `+`, `-`, `*` and `/`.
+
+If you need to compare two instances of `Money`, use its methods `equals`, `isGreaterThan`, `isGreaterThanOrEqualTo`, `isLessThan`, `isLessThanOrEqualTo` and `compare` instead of the operators `==`/`===`, `>`, `>=`, `<`, `<=` and `<=>`.
+If you want to compare `Money` with zero, you can use the methods `isZero`, `isPositive` and `isNegative`.
+
+For more details read [the Money Class section](/docs/introduction/how-to-work-with-money.md#money-class) of the documentation.
+
+There were no changes needed in `shopsys/project-base` as most PHP implementation is in `shopsys/framework`.
+You might need to update your custom classes working with prices and overridden services, because of the changed types (and [strict typing](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.strict) in some cases).
```
simara-svatopluk: The note about stricted types deserves a dedicated section. It is important - it can cause fatal errors (or exceptions?) and here it looks like a subnote. And also stricted types are not directly related to `Money` class itself

docs/upgrade/money-class.md
```diff
@@ -0,0 +1,430 @@
...
+
+*Note: Running the `phpstan` [Phing target](/docs/introduction/console-commands-for-application-management-phing-targets.md) will help you find out problems with incompatible method declarations.*
+*In case of problems, it will fail on a `PHP Fatal error` with a reference to the incompatible method (unfortunately it is unable to continue with the checks as it cannot recover from this error).*
+
+## Twig Templates
+Templates mostly use [the `price` Twig filter](/docs/introduction/how-to-work-with-money.md#price) (or one of its `price*` variants) for rendering monetary values.
+The filters work the same way, but they now require an instance of `Money` to be provided.
+If you pass a value of a different type into one of the `price*` filters (it would fail on a `TypeError`) you should check where it came from and fix it there - all monetary values have to be represented by the new `Money` value object.
+Because the monetary value is mostly taken from a `Price` object and its getters use `Money` as a return type now, the templates usually don't need any changes.
+
```
simara-svatopluk: Maybe smoke tests can help finding occurences? Not all of them, but it could help a bit

docs/upgrade/money-class.md
```diff
@@ -0,0 +1,430 @@
...
+    $this->sql('UPDATE setting_values SET type = \'money\' WHERE name = \'freeTransportAndPaymentPriceLimit\' AND type != \'none\'');
+}
+```
+
+*Note: If you didn't add any new `SettingValue` entities that hold monetary values you can safely skip this step.*
+*The migration in `shopsys/framework` mentioned above will take care of changing the type of the limits for free transport and payment in your DB.*
+
+## Data Fixtures
+All monetary values in data fixtures should use `Money` as well.
+If you haven't altered the data fixtures in any way, you should be ready to go when you copy the data fixtures from `shopsys/project-base` in `v7.0.0` (as explained in the upgrade notes of [PR #854](https://github.com/shopsys/shopsys/pull/854)).
```
simara-svatopluk: Please change the link to upgrade instructions instead of PR. When I click into the PR, I don't see upgrade notes and I don't know where are they.

docs/upgrade/money-class.md
PetrHeinz: Well, it's supposed to mean "...like in this example right here:"

I can remove the colons from such imperatives if it you find it disturbing, but there are some cases in which ending the line with colon is definitely appropriate (*I'm always talking about the first line*):

>There are some cases in which you'll have to modify your templates:
>
>### Manipulation with money before rendering
>Instead of arithmetic operators, ...

>See all the changes that were needed in shopsys/project-base:
>
>- in @ShopsysShop/Front/Content/Product/detail.html.twig:
>  ```diff
>        content="{{ currencyCode(domain.id) }}"
>    ...
>  ```

>These 3 data fixture classes had to be changed in `shopsys/project-base`:
>- `\Shopsys\ShopBundle\DataFixtures\Demo\PaymentDataFixture.php`
>- `\Shopsys\ShopBundle\DataFixtures\Demo\ProductDataFixtureLoader.php`
>- `\Shopsys\ShopBundle\DataFixtures\Demo\TransportDataFixture.php`

docs/upgrade/money-class.md
```diff
@@ -0,0 +1,430 @@
...
+                <tr class="table-cart__row js-cart-item">
+                    <td class="table-cart__cell table-cart__cell--image">
+```
+
+If you extended `QuantifiedProductPriceCalculation`, check your usage of the protected methods `getTotalPriceWithoutVat()`, `getTotalPriceWithVat()` and `getTotalPriceVatAmount()`.
+The signatures were changed and they now require two parameters to be provided.
+
+This was done to prevent the protected parameters `$quantifiedProduct`, `$product`, and `$productPrice` from being used to pass values between methods.
+These protected parameters were removed.
+Refactor your code if you have used them (you can take a look at [the original commit for inspiration](https://github.com/shopsys/shopsys/pull/821/commits/deb643e3d9d3ee077488250d50cc10936086c6a5)).
```
simara-svatopluk: Since instructions are really long, it would be nice to end it with some kind of conclusion...

"The change is huge, we know about it, but we want to keep improving the framework for You. If you find anything we missed or if you'll need to explain anything more, please create an issue or contact us on our slack"

docs/upgrade/money-class.md
PetrHeinz: It might help, good idea. I've added a note in e71c5b592

docs/upgrade/money-class.md
simara-svatopluk: ok, keep them ;-)

docs/upgrade/money-class.md
```diff
@@ -0,0 +1,433 @@
...
+*In case of problems, it will fail on a `PHP Fatal error` with a reference to the incompatible method (unfortunately it is unable to continue with the checks as it cannot recover from this error).*
+
+## Twig Templates
+Templates mostly use [the `price` Twig filter](/docs/introduction/how-to-work-with-money.md#price) (or one of its `price*` variants) for rendering monetary values.
+The filters work the same way, but they now require an instance of `Money` to be provided.
+If you pass a value of a different type into one of the `price*` filters (it would fail on a `TypeError`) you should check where it came from and fix it there - all monetary values have to be represented by the new `Money` value object.
+Because the monetary value is mostly taken from a `Price` object and its getters use `Money` as a return type now, the templates usually don't need any changes.
+
+*Note: [HTTP smoke tests](/docs/introduction/automated-testing.md#http-smoke-tests) can be very helpful when finding issues in your templates.*
+*You can run then by executing the `tests-smoke` [Phing target](/docs/introduction/console-commands-for-application-management-phing-targets.md).*
```
simara-svatopluk:
```suggestion
*You can run them by executing the `tests-smoke` [Phing target](/docs/introduction/console-commands-for-application-management-phing-targets.md).*
```

docs/upgrade/money-class.md
PetrHeinz: You mean a link to https://github.com/shopsys/shopsys/pull/854/files#diff-da18024d2b6b8b4b2fafe94d18cb2866 or to https://github.com/shopsys/shopsys/blob/master/docs/upgrade/UPGRADE-v7.0.0.md#application ?

docs/upgrade/money-class.md
PetrHeinz: A conclusion is a good idea, I've rewritten your suggestion a bit and added links in aa2108474

docs/upgrade/money-class.md
simara-svatopluk: Both are better than just link to PR.... The first is the best in my opinion (even it is just a diff, it shows exactly what to to)

docs/upgrade/money-class.md
PetrHeinz: You're right, added in `declare(strict_types=1);`

docs/upgrade/money-class.md
PetrHeinz: Changed in 86c645ccc

docs/upgrade/money-class.md
PetrHeinz: Thx

docs/upgrade/money-class.md
simara-svatopluk: Please add a also note that phpstan will help as same as in "money in php"

docs/introduction/how-to-work-with-money.md
simara-svatopluk: @PetrHeinz did you create a US? Or an issue?

I'd like to resolve this conversation, but I don't know the output

docs/upgrade/money-class.md
PetrHeinz: I didn't want to copy and I've realized that if you run phpstan, you'll get also errors for Strict typing, so you'll have to fix it together. I've tried to link it in 9182a1a97

docs/introduction/how-to-work-with-money.md
PetrHeinz: I've created US-5461 in our internal backlog.

PetrHeinz:
Rebased on current `master`

PR #875, opened 2019-02-28
[framework] mail spooling

### LINKED ISSUE #728: problem with sending emails in CLI
### What is happening

In CLI mode like cron etc. there could be issue with swiftmailer spool if cron crash, because emails are send only onTerminate event and that event is not called on exception (https://github.com/symfony/swiftmailer-bundle/blob/master/EventListener/EmailSenderListener.php#L46).

Example: 
1. I run command php phing cron
2. First module sends new registered users to IS and sends email to user
2. Cron say that it sent emails to new users
3. Next cron module throws exception for some unknown reason
4. All emails that should be send are away 

### Expected result

If i call $mailer->send($message), that email should be sent immediately.

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| based on implementation of project for targeted group we decided to document behavior of mail spooling and make some changes for CronModules 
|New feature| Yes 
|Fixes issues| closes #728 

### DIFF
--- a/docs/introduction/faq-and-common-issues.md
+++ b/docs/introduction/faq-and-common-issues.md
@@ -19,7 +19,7 @@ For more detailed information about the Shopsys Framework, please see [Shopsys F
 - [Are some periodic tasks part of the Shopsys Framework (cron)?](#are-some-periodic-tasks-part-of-the-shopsys-framework-cron)
 - [Why are you using entity data instead of entities for Symfony forms?](#why-are-you-using-entity-data-instead-of-entities-for-symfony-forms)
 - [What is the configuration file `services_test.yml` good for?](#what-is-the-configuration-file-services_testyml-good-for)
-- [Why are e-mails not sent immediately but at the end of the script](#why-are-e-mails-not-sent-immediately-but-at-the-end-of-the-script)
+- [Why are e-mails not sent immediately but at the end of the script](#why-are-e-mails-sent-before-end-of-the-script-and-not-immediately)
 
 ## What are the phing targets?
 Every phing target is a task that can be executed simply by `php phing <target-name>` command.
--- a/docs/introduction/faq-and-common-issues.md
+++ b/docs/introduction/faq-and-common-issues.md
@@ -106,4 +106,4 @@ E.g., by default, all our services are defined as private. However, in tests, we
 ## Why are e-mails sent before end of the script and not immediately
 Project uses `SwiftMailer` package for sending e-mails and defaultly it has set `spool queue` that stores all mails into `memory` until the script execution is at the end.  
 However, based on implementations of many projects, project gained functionality that releases spool after the `*cronModule` run is ended in case of `cron` phing target that runs all cron modules at once.
-It is also possible to turn the spool off by removing it from [swiftmailer.yml](../../project-base/app/config/packages/swiftmailer.yml) or changing the behavior of storing e-mails based on [symfony docs](https://symfony.com/doc/3.4/email/spool.html).
+It is also possible to turn the spool off by removing it from [swiftmailer.yml](../../project-base/app/config/packages/swiftmailer.yml) or changing the behavior of storing e-mails based on [symfony docs](https://symfony.com/doc/3.4/email/spool.html) or [snc_redis docs](https://github.com/snc/SncRedisBundle/blob/master/Resources/doc/index.md#swiftmailer-spooling).
--- a/docs/introduction/faq-and-common-issues.md
+++ b/docs/introduction/faq-and-common-issues.md
@@ -19,6 +19,7 @@ For more detailed information about the Shopsys Framework, please see [Shopsys F
 - [Are some periodic tasks part of the Shopsys Framework (cron)?](#are-some-periodic-tasks-part-of-the-shopsys-framework-cron)
 - [Why are you using entity data instead of entities for Symfony forms?](#why-are-you-using-entity-data-instead-of-entities-for-symfony-forms)
 - [What is the configuration file `services_test.yml` good for?](#what-is-the-configuration-file-services_testyml-good-for)
+- [Why are e-mails not sent immediately but at the end of the script](#why-are-e-mails-not-sent-immediately-but-at-the-end-of-the-script)
 
 ## What are the phing targets?
 Every phing target is a task that can be executed simply by `php phing <target-name>` command.
@@ -101,3 +102,8 @@ because Symfony forms need setters for all fields and we don't want to mess enti
 is a service configuration file that is loaded in TEST environment in addition to
 the standard configuration defined in [`services.yml`](../../project-base/src/Shopsys/ShopBundle/Resources/config/services.yml) as sometimes the configuration differs from the standard one and we need to override it.
 E.g., by default, all our services are defined as private. However, in tests, we are retrieving some services directly from the container hence we need to have them public in TEST environment.
+
+## Why are e-mails sent before end of the script and not immediately
+Project uses `SwiftMailer` package for sending e-mails and defaultly it has set `spool queue` that stores all mails into `memory` until the script execution is at the end.  
+However, based on implementations of many projects, project gained functionality that releases spool after the `*cronModule` run is ended in case of `cron` phing target that runs all cron modules at once.
+It is also possible to turn the spool off by removing it from [swiftmailer.yml](../../project-base/app/config/packages/swiftmailer.yml) or changing the behavior of storing e-mails based on [symfony docs](https://symfony.com/doc/3.4/email/spool.html).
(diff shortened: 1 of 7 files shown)


PetrHeinz:
:warning: This change is considered a BC-break according to our [BC Promise](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md#php-code) - specifically because of adding a new parameter to a public method (the constructors of `Mailer` and `CronFacade`). See [the table of allowed changes in classes](https://symfony.com/doc/3.4/contributing/code/bc.html#changing-classes).

docs/introduction/faq-and-common-issues.md
```diff
@@ -19,6 +19,7 @@ For more detailed information about the Shopsys Framework, please see [Shopsys F
 - [Are some periodic tasks part of the Shopsys Framework (cron)?](#are-some-periodic-tasks-part-of-the-shopsys-framework-cron)
 - [Why are you using entity data instead of entities for Symfony forms?](#why-are-you-using-entity-data-instead-of-entities-for-symfony-forms)
 - [What is the configuration file `services_test.yml` good for?](#what-is-the-configuration-file-services_testyml-good-for)
+- [Why are e-mails not sent immediately but at the end of the script](#why-are-e-mails-not-sent-immediately-but-at-the-end-of-the-script)
```
DavidOstrozlik: This link is not working

packages/framework/src/Component/Cron/CronFacade.php
```diff
@@ -26,19 +27,27 @@ class CronFacade
...
+     */
+    protected $mailer;
+
     /**
      * @param \Symfony\Bridge\Monolog\Logger $logger
      * @param \Shopsys\FrameworkBundle\Component\Cron\Config\CronConfig $cronConfig
      * @param \Shopsys\FrameworkBundle\Component\Cron\CronModuleFacade $cronModuleFacade
+     * @param \Shopsys\FrameworkBundle\Model\Mail\Mailer $mailer
      */
     public function __construct(
```
DavidOstrozlik: Shouldn't there be upgrade notes for this change?

packages/framework/src/Model/Mail/Mailer.php
```diff
@@ -5,20 +5,54 @@
...
      * @var \Swift_Mailer
      */
     protected $swiftMailer;
 
     /**
      * @param \Swift_Mailer $swiftMailer
+     * @param \Symfony\Component\HttpKernel\KernelInterface $kernel
      */
-    public function __construct(Swift_Mailer $swiftMailer)
+    public function __construct(Swift_Mailer $swiftMailer, KernelInterface $kernel)
```
DavidOstrozlik: I don't think that you should get the whole KernelInterface. Can't you pass the mailers as parameter?

DavidOstrozlik: + missing upgrade notes

packages/framework/src/Model/Mail/Mailer.php
```diff
@@ -5,20 +5,54 @@
...
+        $this->kernel = $kernel;
+    }
+
+    public function flushSpoolQueue()
+    {
+        $container = $this->kernel->getContainer();
+        if (!$container->has('mailer')) {
+            return;
+        }
+        $mailers = array_keys($container->getParameter('swiftmailer.mailers'));
```
DavidOstrozlik: do you need to get only mailer names? isn't there a way to get the whole mailers? so you don't need to get them later on?

packages/framework/src/Model/Mail/Mailer.php
```diff
@@ -5,20 +5,54 @@
...
+        foreach ($mailers as $name) {
+            if (method_exists($container, 'initialized') ? $container->initialized(sprintf('swiftmailer.mailer.%s', $name)) : true) {
```
DavidOstrozlik: I think this could be more understandable if it was in its own protected function/functions (maybe with the second `if` as well)

packages/framework/src/Model/Mail/Mailer.php
```diff
@@ -5,20 +5,54 @@
...
+                if ($container->getParameter(sprintf('swiftmailer.mailer.%s.spool.enabled', $name))) {
+                    $mailer = $container->get(sprintf('swiftmailer.mailer.%s', $name));
+                    $transport = $mailer->getTransport();
+                    if ($transport instanceof \Swift_Transport_SpoolTransport) {
```
DavidOstrozlik: Is there a way that this won't be SpoolTransport? Because you have a condition that spool needs to be enabled for the mailer and maybe this line is not necessary. I don't know the whole logic behind it it would just make sense to me.

packages/framework/src/Model/Mail/Mailer.php
```diff
@@ -5,20 +5,54 @@
...
+                        $spool = $transport->getSpool();
+                        if ($spool instanceof \Swift_MemorySpool) {
+                            try {
+                                $spool->flushQueue($container->get(sprintf('swiftmailer.mailer.%s.transport.real', $name)));
```
DavidOstrozlik: do you need to get the transport from container? Because you have the whole mailer already so isn't there a way you could get it from there?

packages/framework/src/Model/Mail/Mailer.php
```diff
@@ -5,20 +5,54 @@
...
+                            } catch (\Swift_TransportException $exception) {
```
DavidOstrozlik: Why do you only catch the Exception but don't throw any error?

packages/framework/src/Model/Mail/Mailer.php
boris-brtan: hmm this is good idea, how could it look like to pass mailers as parameters?

boris-brtan: upgrade notes about ? 
`fix constructor of FrameworkBundle/Model/Mail/Mailer if you extended it somehow.` ???

docs/introduction/faq-and-common-issues.md
boris-brtan: fixing with `#why-are-e-mails-sent-before-end-of-the-script-and-not-immediately`

packages/framework/src/Model/Mail/Mailer.php
boris-brtan: i tried to call the same thing that is called on the releasing of spool. please look into the `vendor/symfony/swiftmailer-bundle/EventListener/EmailSenderListener.php`

packages/framework/src/Model/Mail/Mailer.php
boris-brtan: hold-on EmailSenderListener.php::onTerminate

packages/framework/src/Model/Mail/Mailer.php
boris-brtan: hold-on EmailSenderListener.php::onTerminate

packages/framework/src/Model/Mail/Mailer.php
boris-brtan: hold-on EmailSenderListener.php::onTerminate

packages/framework/src/Model/Mail/Mailer.php
boris-brtan: hold-on EmailSenderListener.php::onTerminate

boris-brtan: there is no logger in the Mailer

packages/framework/src/Component/Cron/CronFacade.php
boris-brtan: `fix constructor of FrameworkBundle/Model/Mail/Mailer if you extended it somehow`. ???

boris-brtan:
> adding
hmmm, i found this rule
Class >> public METHOD >> Add argument without a default value >> No

packages/framework/src/Component/Cron/CronFacade.php
```diff
@@ -130,6 +140,12 @@ protected function runModule(CronModuleExecutor $cronModuleExecutor, CronModuleC
             $this->cronModuleFacade->isModuleSuspended($cronModuleConfig)
         );
 
+        try {
+            $this->mailer->flushSpoolQueue();
+        } catch (Swift_TransportException $exception) {
+            $this->logger->error(sprintf('Exception occurred while flushing email queue: %s', $exception->getMessage()));
```
PetrHeinz: See: vendor/symfony/symfony/src/Symfony/Component/Console/EventListener/ErrorListener.php:43 for better usage:
`return $this->logger->error('An error occurred while using the console. Message: "{message}"', ['exception' => $error, 'message' => $error->getMessage()]);`

PR #878, opened 2019-03-04
[google-cloud-bundle] unified shopsys requirements to dev-master

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| https://github.com/shopsys/shopsys/pull/730#discussion_r261583258 . Also our automatic release process will be easier with this change.
|New feature| No
|Fixes issues| ... 


PetrHeinz:
Shouldn't the only dependency be `shopsys/framework`? Other packages should be just `@dev` as [it only allows using it with lower stability](https://getcomposer.org/doc/articles/versions.md#stability-constraints). Or did I get it wrong?

simara-svatopluk:
@PetrHeinz you are right, but see the commit message https://github.com/shopsys/shopsys/pull/878/commits/3838e61916df5e2af47e7c0a8e77c02367a9bf69
.... the goal is consistency as we do it "wrongish" in all our packages

boris-brtan:
try to solve also the correctness of values in releases 
https://github.com/shopsys/google-cloud-bundle/blob/d72df259c46488b767e15183587546a66c778690/composer.json#L21-L24

simara-svatopluk:
@boris-brtan ou.. what do you mean? What is wrong with `7.0.0`? And BTW should it even be solved during this PR?

boris-brtan:
Hi @simara-svatopluk, as i see it there is some problem during releases that @dev is changed into release tag but the other way it needs to be changed into @dev manually.

boris-brtan:
### TEST
No difference between @dev and dev-master, so OK.

simara-svatopluk:
@boris-brtan So conflicts are a transcendent dependency. No need for mentioning them manually :-)

PR #881, opened 2019-03-05
[framework] Firing opening subcategories menu animation only once

### LINKED ISSUE #833: Administration: flashing menu
### What is happening
Left menu flashs when the admin switching subsection.

### Expected result
Please fix flashing.

topolanekmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| By this pull request is added changes that fire opening subcategories menu animation only once. It will not be fired by clicking on every subcategory item menu.
|New feature| No 
|Fixes issues| closes #833

### DIFF
--- a/packages/framework/src/Resources/scripts/admin/sideMenu.js
+++ b/packages/framework/src/Resources/scripts/admin/sideMenu.js
@@ -9,7 +9,7 @@
 
             $items.click(function () {
                 if (!$(this).hasClass('open')) {
-                    $(this).addClass('open');
+                    $(this).addClass('open animated');
                 }
 
                 $(this).filterAllNodes('.js-side-menu-submenu').show();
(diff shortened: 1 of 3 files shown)


PetrHeinz:
Hi @topolanekmartin, thanks for the fix! What kind of BC break do you see in here?

topolanekmartin:
Hello @PetrHeinz, I think it is because of changing the behavior of the javascript function used in administration. What do you think?

grossmannmartin:
Hi, @topolanekmartin, I tried it too and I see the problem as well.
With this commit, you fixed flashing "after click", but after page load menu still flashes.

topolanekmartin:
Hello, @vitek-rostislav , @grossmannmartin, I thought this function is desirable - firing animation when page is loaded or main category was opened. I follow the description of issue where is written _"Administration: The left menu flashes when the admin switching subsection."_

But I understand that the description of the issue can be misleading. Should I fix it even when the page is reloaded? :)

PetrHeinz:
@vitek-rostislav @topolanekmartin Personally, I wouldn't consider this a BC break because the behavior is just fixed, not changed in any meaningful way (I cannot imagine anyone would rely on `.open` being added *after* the element is shown, not before).

PR #886, opened 2019-03-11
[shopsys] added version prefix to redis caches so the application can be safely deployed

### LINKED ISSUE #442: Redis cache is not cleared after deploy with native installation

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| #442
|New feature| Yes 
|Fixes issues| resolves #442
|Standards and tests pass| No


packages/framework/src/Component/Redis/RedisVersionsFacade.php
```diff
@@ -0,0 +1,49 @@
...
+    public function cleanOldCache(): void
+    {
+        $versionPattern = '[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]*'; // redis pattern is a glob style and doesn't support repetitions
+
+        $all = $this->globalClient->keys($versionPattern);
+        $toRemove = [];
+
+        $prefix = (string)$this->globalClient->getOption(Redis::OPT_PREFIX);
+        $currentVersionPrefix = $prefix . $this->currentVersion;
+        foreach ($all as $key) {
```
henzigo: Probably that may be a problem for really big e-shops. For example, if a million records are to be deleted that foreach will be very slow.

project-base/build.xml
```diff
@@ -351,6 +358,20 @@
         </exec>
     </target>
 
+    <target name="generate-build-version">
+        <exec executable="${path.php.executable}" checkreturn="true" outputProperty="version">
+            <arg value="-r" />
+            <arg value="echo date('YmdHis');" />
```
henzigo: Isn't that unique version overkill? It's not possible to deploy every second and you will have only one previous version in the same time. So it's not necessary to have that long version.

docs/introduction/console-commands-for-application-management-phing-targets.md
```diff
@@ -92,6 +92,21 @@ Cleans up cache in Redis database except for sessions.
 
 Useful in development environment and during deploying to production.
 
+#### clean-redis-old
+Cleans up cache in Redis database for old build versions (but keeps sessions).
```
boris-brtan:
```suggestion
Cleans up cache in Redis database for old built versions (but keeps sessions).
```

project-base/app/config/packages/snc_redis.yml
```diff
@@ -5,19 +5,25 @@ snc_redis:
             alias: 'bestselling_products'
             dsn: 'redis://%redis_host%'
             options:
-                prefix: '%env(REDIS_PREFIX)%bestselling_products_'
+                prefix: '%env(REDIS_PREFIX)%%build-version%bestselling_products_'
```
boris-brtan: would some additional `_` as delimiters make sense or not?

packages/framework/src/Component/Redis/RedisVersionsFacade.php
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public function __construct(Redis $globalClient, string $currentVersion)
+    {
+        $this->globalClient = $globalClient;
+        $this->currentVersion = $currentVersion;
+    }
+
+    public function cleanOldCache(): void
+    {
+        $versionPattern = '[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]*'; // redis pattern is a glob style and doesn't support repetitions
```
boris-brtan: should not be part of $versionPattern also prefix value ?

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -10,6 +10,81 @@ There you can find links to upgrade notes for other versions too.
...
+        And add a `global` client (it is also without the `build-version` prefix)
+        ```diff
+        snc_redis:
+            clients:
+        +       global:
+        +           type: 'phpredis'
+        +           alias: 'global'
+        +           dsn: 'redis://%redis_host%'
+        +           options:
+        +               prefix: '%env(REDIS_PREFIX)%'
```
boris-brtan: should not be here also `%build-version%` ???

packages/framework/src/Component/Redis/RedisVersionsFacade.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $all = $this->globalClient->keys($versionPattern);
+        $toRemove = [];
+
+        $prefix = (string)$this->globalClient->getOption(Redis::OPT_PREFIX);
+        $currentVersionPrefix = $prefix . $this->currentVersion;
+        foreach ($all as $key) {
+            if (strpos($key, $currentVersionPrefix) === false) {
+                $keyWithoutPrefix = substr($key, strlen($prefix)); // redis returns keys including prefix but needs them without prefix during removing
```
boris-brtan: what about to put comments `//` above the code line?

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -10,6 +10,81 @@ There you can find links to upgrade notes for other versions too.
...
+                        <token key="version" value="${version}" />
+                    </replacetokens>
+                </filterchain>
+            </copy>
+        </target>
+        ```
+    - run `php phing generate-build-version`
+    - and include `generate-build-version` and `clean-redis-old` to your build phing targets. Please find inspiration in [#886](https://github.com/shopsys/shopsys/pull/886/files)
+    - once you finish this change (include the `build-version` into caches), you still should deal with older redis cache keys that don't use `build-version` prefix (16 digits).
+      Such keys are not removed even by `clean-redis-old`.
```
boris-brtan: this line and padding formating will not have any impact for viewer of md file.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -10,6 +10,81 @@ There you can find links to upgrade notes for other versions too.
...
+      Please find and remove them manually (via console or UI)
```
boris-brtan: please join this line into previous or add dash `-`

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -10,6 +10,81 @@ There you can find links to upgrade notes for other versions too.
...
+                <filterchain>
+                    <replacetokens begintoken="%%" endtoken="%%">
+                        <token key="version" value="${version}" />
+                    </replacetokens>
+                </filterchain>
+            </copy>
+        </target>
+        ```
+    - run `php phing generate-build-version`
+    - and include `generate-build-version` and `clean-redis-old` to your build phing targets. Please find inspiration in [#886](https://github.com/shopsys/shopsys/pull/886/files)
```
boris-brtan: maybe some instruction about functional test implementation would be great.

project-base/app/config/parameters.yml.dist
```diff
@@ -22,3 +22,4 @@ parameters:
     trusted_proxies:
         - "127.0.0.1"
     env(REDIS_PREFIX): ''
+    build-version: "0000000000000000"
```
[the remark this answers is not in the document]
simara-svatopluk: Maybe the default parameter `build-version` in `project-base` is really not needed as the Symfony should use the parameter from `framework` package. During testing I didn't realize that I used old framework in project-base, so the build failed during `composer install` (the parameter was obviously missing).... During tests it seems reasonable to branch-split the monorepo and try it without default parameter in project-base.

packages/framework/src/Component/Redis/RedisVersionsFacade.php
boris-brtan: https://redis.io/commands/unlink

docs/upgrade/UPGRADE-unreleased.md
grossmannmartin: It's used only in `RedisVersionFacade` and this facade has to be aware of all entries (to be able to delete the old one).

docs/introduction/console-commands-for-application-management-phing-targets.md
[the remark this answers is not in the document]
grossmannmartin: I used different phrasing. Is it ok?

packages/framework/src/Resources/config/services.yml
```diff
@@ -8,6 +8,7 @@ parameters:
     faker.seed: 1234
     shopsys.entity_extension.map: {}
     shopsys.admin_locale: en
+    build-version: "0000000000000000"
```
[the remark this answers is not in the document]
grossmannmartin: I think these default values are better here because we don't need to force users to make unnecessary changes in their parameters file as `parameters_common.yml` is a part of project-base (open-box)

packages/framework/src/Component/Redis/RedisVersionsFacade.php
grossmannmartin: You're right. `Keys` is a performance killer.

project-base/app/config/packages/snc_redis.yml
grossmannmartin: I don't see much added value in that

packages/framework/src/Component/Redis/RedisVersionsFacade.php
grossmannmartin: I would like to keep it like that and Symfony writes comments on the same line too.

docs/upgrade/UPGRADE-unreleased.md
grossmannmartin: Yeah, but its easier to read the raw file

packages/framework/src/Component/Redis/RedisVersionsFacade.php
```diff
@@ -32,18 +32,21 @@ public function cleanOldCache(): void
...
-            if (strpos($key, $currentVersionPrefix) === false) {
-                $keyWithoutPrefix = substr($key, strlen($prefix)); // redis returns keys including prefix but needs them without prefix during removing
-                $toRemove[] = $keyWithoutPrefix;
+
+        $this->globalClient->setOption(Redis::OPT_SCAN, Redis::SCAN_RETRY);
+        $iterator = null;
+
+        while ($keys = $this->globalClient->scan($iterator, $versionPattern, 10000)) {
+            $toRemove = [];
+            foreach ($keys as $key) {
```
boris-brtan: hmm what about to use rather array_map then?

grossmannmartin: Well, neither of these don't fit the need. I could somehow rewrite it with a combination of `array_filter` and `array_map`, but for me, it's hard to read then.
Or can you write an example of your idea?

packages/framework/src/Component/Redis/RedisVersionsFacade.php
grossmannmartin: I rewrote it to scan the keys, and it seems fine. I was able to delete about 1M of records in a minute.

project-base/build.xml
grossmannmartin: Maybe you're right, but I don't feel the need to rewrite it. 
Can it cause some problems?

project-base/build.xml
```diff
@@ -52,10 +52,10 @@
...
+    <target name="build-deploy-part-1-db-independent" depends="generate-build-version,clean,redis-check,composer,npm,dirs-create,assets" description="First part of application build for production preserving your DB (can be run without maintenance page)."/>
+    <target name="build-deploy-part-2-db-dependent" depends="db-migrations,create-domains-data,generate-friendly-urls,replace-domains-urls,grunt,error-pages-generate,warmup,clean-redis-old" description="Second part of application build for production preserving your DB (must be run with maintenance page when containing DB migrations)."/>
+    <target name="build-new" depends="wipe,generate-build-version,redis-check,composer,npm,dirs-create,assets,db-rebuild,grunt,error-pages-generate,warmup,product-search-create-structure,clean-redis-old" description="Builds application for production with clean DB (with base data only)."/>
+    <target name="build-demo" depends="wipe,generate-build-version,redis-check,composer,npm,dirs-create,assets,db-demo,grunt,error-pages-generate,warmup,product-search-recreate-structure,product-search-export-products,clean-redis-old" description="Builds application for production with clean demo DB."/>
```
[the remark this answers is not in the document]
grossmannmartin: Why it should be in db-demo target?

docs/upgrade/UPGRADE-unreleased.md
grossmannmartin: Instruction to download and use `RedisVersionsFacadeTest` seems sufficient to me, don't you think? Exact implementation is available and it's a few lines of code.

project-base/app/config/parameters.yml.dist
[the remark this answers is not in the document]
grossmannmartin: Okay, I did some digging and we need to have a build-version specified in both files. Composer install fails when build-version is not in parameters file, because it needs to be known before registering the snc_redis bundle. 
And for `RedisVersionFacade` we need a default value in a framework (what if someone don't do this upgrade).

project-base/build.xml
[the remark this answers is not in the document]
boris-brtan: after you rebuilt database, your bestselling product doesnþt exists anymore and doctrine metadata will change, or?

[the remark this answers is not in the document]
grossmannmartin: Also elastic search index should be wiped and exported.
What about `db-rebuild`? I'm really not sure with these two targets (like I really don't know)

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -148,6 +148,7 @@ There you can find links to upgrade notes for other versions too.
             </copy>
         </target>
         ```
+    - download [`RedisVersionsFacadeTest.php`](https://github.com/shopsys/project-base/tree/master/tests/ShopBundle/Functional/Component/Redis/RedisVersionsFacadeTest.php) to your tests directory `tests/ShopBundle/Functional/Component/Redis/`
```
boris-brtan: consider the use of master branch....

docs/upgrade/UPGRADE-unreleased.md
boris-brtan: OK just master branch is somehow strange.

grossmannmartin: What do you mean by "somehow strange"?

docs/upgrade/UPGRADE-unreleased.md
grossmannmartin: it's part of release process to change it. Right now I don't have any different target to link to

project-base/build.xml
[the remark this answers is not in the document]
boris-brtan: so we can agree with current state now.

docs/upgrade/UPGRADE-unreleased.md
boris-brtan: so master will be changed into v7.0.1 during release right?

grossmannmartin: Exactly

boris-brtan:
### TEST
looked it worked also with REDIS_PREFIX env variable setted to some value.
i see in snc_redis config doctrine_metadata doctrine_query, but i can see no records from doctrine in redis-admin maybe some issue could be created for this or reopened if it is closed

docs/introduction/console-commands-for-application-management-phing-targets.md
```diff
@@ -92,6 +92,21 @@ Cleans up cache in Redis database except for sessions.
 
 Useful in development environment and during deploying to production.
 
+#### clean-redis-old
+Cleans up cache in Redis database for previously built versions (but keeps sessions).
+Previously built version is any different from current `build-version` generated by [generate-build-version](#generate-build-version).
```
VitMickal:
```suggestion
Previously built version is different from current `build-version` generated by [generate-build-version](#generate-build-version).
```

PR #890, opened 2019-03-12
[shopsys] marked twig/twig conflicting in version >=2.7.0

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| the assetic bundle is not compatible with newer version of twig
|New feature| No 
|Fixes issues| ... 


simara-svatopluk:
`twig/twig` released fixing version, so we don't have to mark conflicting version (we could do it but it is not critical and we don't want to spent time on it)

PR #904, opened 2019-03-22
[shopsys] constants are now extendable

### LINKED ISSUE #827: Constants are not extendable
There is no possible solution to changing value of constants from framework which is relevant use case.

### What is happening
I think best to describe is an example.

This is a real class from framework
```
<?php

namespace Shopsys\FrameworkBundle\Model\Product\BestsellingProduct;

class BestsellingProductFacade
{
    const MAX_RESULTS = 10;

    ...
}
```
If I would like to change value of `MAX_RESULTS` to  for example 20, from Shopbundle, i got no way to do that in a nice way.

### Expected result
I think of 2 solutions that could be implemented.

1) Use static constants, in which case we can enable chaning of values from ShopBundle by standard overriding of constant in your implementation.

2) Use `get()` for constants. In which case you can simply inject your own constant into getter

I would prefer first solution.

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We need rules for class constants in framework and other packages to allow their extensibility from project. More description follows.
|New feature| No 
|Fixes issues| closes #827 

In this PR is done several changes, all with the same goal to set some rules for constants.

# 1. Useless access annotation
Sniff disallowing access annotation in phpdoc was changed a little bit, so now it disallows only redundant access annotations. (`Shopsys\CodingStandards\CsFixer\Phpdoc\NoUselessAccessFixer`)
Meaning this is wrong
```php
/** @access protected */
protected const A = 'value';
```

but this is valid 
```php
/** @access protected */
const A = 'value';
```

This is done to keep backward compatibility with a possibility to mark somehow constants, that are meant to be "extendible".

# 2. Mandatory visibility for constants
Every constant have to have visibility defined with the access modifier or with `@access` annotation.
The first is recommended and in the future, the access annotation will not be present in the code at all.

# 3. Force late static binding for protected constants in packages
Constants in packages (framework, etc.) have to use protected constants with `static` keyword instead of self, so it's possible to, adjust the value only with extending the class and redefine the constant needed (with `self` is mandatory to extend all methods with the usages too).

### DIFF
--- a/packages/coding-standards/src/CsFixer/Phpdoc/NoUselessAccessFixer.php
+++ b/packages/coding-standards/src/CsFixer/Phpdoc/NoUselessAccessFixer.php
@@ -1,0 +1,216 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\CsFixer\Phpdoc;
+
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\Fixer\DefinedFixerInterface;
+use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+final class NoUselessAccessFixer implements FixerInterface, DefinedFixerInterface
+{
+    private const ACCESS_TOKENS = [
+        \T_PUBLIC,
+        \T_PROTECTED,
+        \T_PRIVATE,
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            '`@access` annotations should be omitted from PHPDoc when it is useless',
+            [
+                new CodeSample(
+                    '<?php
+class Foo
+{
+    /**
+     * @internal
+     * @access private
+     */
+    private $bar;
+}
+'
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens): bool
+    {
+        return $tokens->isTokenKindFound(\T_DOC_COMMENT) && $tokens->isAnyTokenKindsFound(self::ACCESS_TOKENS);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens): void
+    {
+        $allDocBlocks = $this->findAllDocBlocks($tokens);
+
+        foreach ($allDocBlocks as $position => $docBlock) {
+            if (!$this->areThereAnyAccessAnnotation($docBlock)) {
+                continue;
+            }
+
+            if ($this->followingTokenIsAccessModifier($tokens, $position)) {
+                $this->removeAllAccessAnnotations($docBlock);
+                $this->rewriteDocBlock($tokens, $docBlock, $position);
+            }
+
+            if ($this->areThereAnyEmptyAccessAnnotation($docBlock)) {
+                $this->removeEmptyAccessAnnotations($docBlock);
+                $this->rewriteDocBlock($tokens, $docBlock, $position);
+            }
+        }
+    }
+
+    /**
+     * @param \PhpCsFixer\Tokenizer\Tokens $tokens
+     * @return \PhpCsFixer\DocBlock\DocBlock[]
+     */
+    private function findAllDocBlocks(Tokens $tokens): array
+    {
+        return \array_map(function (Token $token) {
+            return new DocBlock($token->getContent());
+        }, $tokens->findGivenKind(\T_DOC_COMMENT));
+    }
+
+    /**
+     * @param \PhpCsFixer\DocBlock\DocBlock $docBlock
+     * @return bool
+     */
+    private function areThereAnyAccessAnnotation(DocBlock $docBlock): bool
+    {
+        $annotations = $docBlock->getAnnotationsOfType('access');
+
+        return !empty($annotations);
+    }
+
+    /**
+     * @param \PhpCsFixer\DocBlock\DocBlock $docBlock
+     * @return bool
+     */
+    private function areThereAnyEmptyAccessAnnotation(DocBlock $docBlock): bool
+    {
+        $annotations = $docBlock->getAnnotationsOfType('access');
+
+        foreach ($annotations as $annotation) {
+            if (\preg_match('~(public|protected|private)~', $annotation->getContent()) === 0) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @param \PhpCsFixer\DocBlock\DocBlock $docBlock
+     */
+    private function removeEmptyAccessAnnotations(DocBlock $docBlock): void
+    {
+        $annotations = $docBlock->getAnnotationsOfType('access');
+
+        foreach ($annotations as $annotation) {
+            if (\preg_match('~(public|protected|private)~', $annotation->getContent()) === 0) {
+                $annotation->remove();
+            }
+        }
+    }
+
+    /**
+     * @param \PhpCsFixer\Tokenizer\Tokens $tokens
+     * @param int $position
+     * @return bool
+     */
+    private function followingTokenIsAccessModifier(Tokens $tokens, int $position): bool
+    {
+        $nextMeaningfulTokenId = $tokens[$tokens->getNextMeaningfulToken($position)]->getId();
+
+        return \in_array($nextMeaningfulTokenId, self::ACCESS_TOKENS, true);
+    }
+
+    /**
+     * @param \PhpCsFixer\DocBlock\DocBlock $docBlock
+     */
+    private function removeAllAccessAnnotations(DocBlock $docBlock): void
+    {
+        $annotations = $docBlock->getAnnotationsOfType('access');
+
+        foreach ($annotations as $annotation) {
+            $annotation->remove();
+        }
+    }
+
+    /**
+     * @param \PhpCsFixer\Tokenizer\Tokens $tokens
+     * @param \PhpCsFixer\DocBlock\DocBlock $docBlock
+     * @param int $position
+     */
+    private function rewriteDocBlock(Tokens $tokens, DocBlock $docBlock, int $position): void
+    {
+        if ($docBlock->getContent() === '' || $this->isDocBlockEmpty($docBlock)) {
+            $tokens->clearTokenAndMergeSurroundingWhitespace($position);
+        } else {
+            $tokens[$position] = new Token([T_DOC_COMMENT, $docBlock->getContent()]);
+        }
+    }
+
+    /**
+     * @param \PhpCsFixer\DocBlock\DocBlock $docBlock
+     * @return bool
+     */
+    private function isDocBlockEmpty(DocBlock $docBlock): bool
+    {
+        foreach ($docBlock->getLines() as $line) {
+            if ($line->containsUsefulContent()) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky(): bool
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName(): string
+    {
+        return 'Shopsys/phpdoc_no_useless_access';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority(): int
+    {
+        return 0;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(\SplFileInfo $file): bool
+    {
+        return true;
+    }
+}
--- a/packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
+++ b/packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\Sniffs;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+class ConstantVisibilityRequiredSniff implements Sniff
+{
+    /**
+     * @return int[]
+     */
+    public function register(): array
+    {
+        return [\T_CONST];
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param $constPosition
+     */
+    public function process(File $file, $constPosition): void
+    {
+        if (!$this->isConstInsideClass($file, $constPosition)) {
+            return;
+        }
+
+        if ($this->isConstWithAccessModifier($file, $constPosition)) {
+            return;
+        }
+
+        if ($this->isConstWithAccessAnnotation($file, $constPosition)) {
+            return;
+        }
+
+        $file->addError('Constant must have access modifier', $constPosition, self::class);
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $constPosition
+     * @return bool
+     */
+    private function isConstInsideClass(File $file, int $constPosition): bool
+    {
+        $classStartPosition = $file->findPrevious(\T_CLASS, $constPosition);
+        if ($classStartPosition === false) {
+            return false;
+        }
+
+        $tokens = $file->getTokens();
+        $classEndPosition = $tokens[$classStartPosition]['scope_closer'];
+
+        return $constPosition > $classStartPosition && $constPosition < $classEndPosition;
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $constPosition
+     * @return bool
+     */
+    private function isConstWithAccessModifier(File $file, int $constPosition): bool
+    {
+        $previousTokenEndPosition = $this->findScopeSearchEndPosition($file, $constPosition);
+
+        $accessModifierStartPosition = $file->findPrevious(Tokens::$scopeModifiers, $constPosition, $previousTokenEndPosition);
+
+        return (bool)$accessModifierStartPosition;
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $constPosition
+     * @return bool
+     */
+    private function isConstWithAccessAnnotation(File $file, int $constPosition): bool
+    {
+        $previousTokenEndPosition = $this->findScopeSearchEndPosition($file, $constPosition);
+
+        $phpDocStartPosition = $file->findPrevious(\T_DOC_COMMENT_OPEN_TAG, $constPosition, $previousTokenEndPosition ?: 0);
+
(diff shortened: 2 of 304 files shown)


docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -153,6 +153,13 @@ There you can find links to upgrade notes for other versions too.
     - and include `generate-build-version` and `clean-redis-old` to your build phing targets. Please find inspiration in [#886](https://github.com/shopsys/shopsys/pull/886/files)
     - once you finish this change (include the `build-version` into caches), you still should deal with older redis cache keys that don't use `build-version` prefix (16 digits).
       Such keys are not removed even by `clean-redis-old`, please find and remove them manually (via console or UI)
+- skip `Shopsys\CodingStandards\Sniffs\ForceLateStaticBindingForProtectedConstantsSniff` sniff in your `easy-coding-standard.yml`, so you won't be forced to use `static` on protected constants ([#904](https://github.com/shopsys/shopsys/pull/904))
+    - ```yaml
+        parameters:
+            skip:
+                Shopsys\CodingStandards\Sniffs\ForceLateStaticBindingForProtectedConstantsSniff:
+                    - '*'
+      ```
```
PetrHeinz: Do we want to make the rule default in our coding-standards? Maybe we can add it only to our extensible packages (or better yet, make an `extensible_easy_coding_standards.yml` ruleset for extensible packages) so it requires no change from the project developers.

packages/coding-standards/src/Sniffs/ForceLateStaticBindingForProtectedConstantsSniff.php
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Shopsys\CodingStandards\Sniffs;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\AnnotationHelper;
+use SlevomatCodingStandard\Helpers\ConstantHelper;
+use SlevomatCodingStandard\Helpers\TokenHelper;
+
+class ForceLateStaticBindingForProtectedConstantsSniff implements Sniff
```
PetrHeinz: Was it too hard to implement as a fixer?

packages/framework/src/Command/DatabaseDumpCommand.php
```diff
@@ -11,7 +11,9 @@
 
 class DatabaseDumpCommand extends Command
 {
+    /** @access private */
     const ARG_OUTPUT_FILE = 'outputFile';
```
PetrHeinz: Shouldn't argument names be public? I'd guess they might come handy when calling the command programatically or during an extension. As an example, take a look at `ServerStartWithCustomRouterCommand` which would definitely use a `ServerStartCommand::ARG_ROUTER` constant if it were declared (now it uses only the literal string `'router'`).

Or did you have some other reason for the private visibility in mind?

packages/coding-standards/src/Sniffs/ForceLateStaticBindingForProtectedConstantsSniff.php
grossmannmartin: It is a fixer. But it uses a PHP_CodeSniffer where everything is a sniff (some are fixable though).

docs/upgrade/UPGRADE-unreleased.md
grossmannmartin: That would be nice, but I didn't find a way how to achieve this from a view of a monorepo. 
When I run ecs with a monorepo configuration, it uses all the configurations and if I don't skip the sniff in a project base, I end up with different rules when installing monorepo and project-base itself.

PetrHeinz: Yeah, that makes sense. It would mean we have to do separate runs of the checking in monorepo or configure monorepo to skip it in the whole project-base directory (or namespace).

@vitek-rostislav: This might be a reason why the `ForbiddenDoctrineInheritanceSniff` was configured the way it was as you've mentioned the other day on a team meeting.

simara-svatopluk: I think that all sniffs/fixers created during this PR should be included in standard configuration.
The coding-standards package is
> Tools for automated checks of coding standards in PHP projects. Predefined rulesets included.

The rulesset is predefined for the Shopsys usage. I think all these sniffs/fixes fits for the SSFW and also for the Shopsys agency. And if there is a user that doesn't want to use such sniff, he can freely skip the sniff/fixer.

simara-svatopluk: Please move the instruction to section `coding-standards`. Once you move it, you don't have to use the example with `yaml` configuration

packages/coding-standards/tests/Unit/Sniffs/ConstantVisibilityRequiredSniff/ConstantVisibilityRequiredSniffTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+declare(strict_types=1);
+
+namespace Tests\CodingStandards\Sniffs\ConstantVisibilityRequiredSniff;
+
+use Symplify\EasyCodingStandardTester\Testing\AbstractCheckerTestCase;
+
+final class ConstantVisibilityRequiredSniffTest extends AbstractCheckerTestCase
+{
+    public function testCorrect(): void
```
simara-svatopluk: Please add correct test case that has mixed annotation and visibility
```php
class...
/**
 * @access public
 */
private const A;
```

packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
```diff
@@ -0,0 +1,113 @@
...
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param $constPosition
+     */
+    public function process(File $file, $constPosition): void
+    {
+        if (!$this->isConstInsideClass($file, $constPosition)) {
+            return;
+        }
+
+        $previousTokenEndPosition = $file->findPrevious([\T_SEMICOLON, \T_CLOSE_CURLY_BRACKET], $constPosition) ?: 0;
```
simara-svatopluk: I don't understand what expresses the variable and why it is passed into methods below. Could you please extract the line into a private method and use the new method inside of methods below so you don't have to pass the variable?

packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
```diff
@@ -0,0 +1,113 @@
...
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $constPosition
+     * @param int $previousTokenEndPosition
+     * @return bool
+     */
+    private function isConstWithAccessAnnotation(File $file, int $constPosition, int $previousTokenEndPosition): bool
+    {
+        $phpDocStartPosition = $file->findPrevious(\T_DOC_COMMENT_OPEN_TAG, $constPosition, $previousTokenEndPosition ?: 0);
+
+        return $phpDocStartPosition !== false && $this->phpDocContainsAccessTag($file, $phpDocStartPosition);
```
simara-svatopluk: It's difficult to read for me

```suggestion
        if ($phpDocStartPosition === false) {
            return false;
        }
        return $this->phpDocContainsAccessTag($file, $phpDocStartPosition);
```

packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $phpDocStartPosition
+     * @return bool
+     */
+    private function phpDocContainsAccessTag(File $file, int $phpDocStartPosition): bool
```
simara-svatopluk: Wow, I don't understand insides of the function. But the functionality is well encapsulated and it is well tested... So no problem :-)

packages/coding-standards/tests/Unit/CsFixer/Phpdoc/NoUselessAccessFixer/NoUselessAccessFixerTest.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\CodingStandards\CsFixer\OrmJoinColumnRequireNullableFixer;
+
+use Symplify\EasyCodingStandardTester\Testing\AbstractCheckerTestCase;
+
+final class NoUselessAccessFixerTest extends AbstractCheckerTestCase
```
simara-svatopluk: The name evokes in me that this fixer will also fix (as the access annotation is useless)
```php
class ...
/**
 * @access
 */
const X = 10;
```

But it won't... Am I right? If not maybe think about a better name

simara-svatopluk: If you decide that this fixer will also remove such empty `@access`, I'll be happy and you don't have to think about renaming

packages/coding-standards/src/CsFixer/Phpdoc/NoUselessAccessFixer.php
```diff
@@ -0,0 +1,131 @@
...
+    {
+        return $tokens->isTokenKindFound(\T_DOC_COMMENT) && $tokens->isAnyTokenKindsFound(self::ACCESS_TOKENS);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens): void
+    {
+        foreach ($tokens->findGivenKind(\T_DOC_COMMENT) as $index => $token) {
```
simara-svatopluk: This function is quite nested, could you extract inside of the foreach into a separated function?

packages/coding-standards/src/CsFixer/Phpdoc/NoUselessAccessFixer.php
```diff
@@ -0,0 +1,131 @@
...
+            $docBlock = new DocBlock($token->getContent());
+
+            $annotations = $docBlock->getAnnotationsOfType('access');
+
+            if (empty($annotations)) {
```
simara-svatopluk: Please extract the `if` into a function like `areThereAnyAnnotation($docBlock)`

packages/coding-standards/src/CsFixer/Phpdoc/NoUselessAccessFixer.php
```diff
@@ -0,0 +1,131 @@
...
+                continue;
+            }
+
+            $nextMeaningfulTokenId = $tokens[$tokens->getNextMeaningfulToken($index)]->getId();
+
+            if (\in_array($nextMeaningfulTokenId, self::ACCESS_TOKENS, true)) {
+                foreach ($annotations as $annotation) {
```
simara-svatopluk: Please extract this part into function `removeUselessAcessAnnotation($docBlock)`
And then you can remove temporary variable `$annotations` above

packages/coding-standards/src/Sniffs/ForceLateStaticBindingForProtectedConstantsSniff.php
```diff
@@ -0,0 +1,100 @@
...
+    public function process(File $file, $classPosition)
+    {
+        $tokens = $file->getTokens();
+
+        $protectedConstants = $this->getAllProtectedConstantsInClass($file);
+
+        $selfPositions = TokenHelper::findNextAll($file, \T_SELF, $classPosition);
+
+        foreach ($selfPositions as $selfPosition) {
+            $doubleColonPosition = TokenHelper::findNextEffective($file, $selfPosition + 1);
```
simara-svatopluk: I think about extracting the long chain of IFs to function `isSelfAndShouldBeStatic`

and later in code `$constantName = $tokens[$stringPosition]['content'];` change to function `$constantName = $this->findConstatntName(...)`. It would maybe contain duplicit logic with `isSelf`, but this duplicit could be maybe also extracted

By this change you'll remove temporary variables `$doubleColonPosition, $stringPosition, $positionAfterString`

packages/coding-standards/tests/Unit/CsFixer/Phpdoc/NoUselessAccessFixer/wrong/Simple.php
```diff
@@ -0,0 +1,15 @@
...
+declare(strict_types=1);
+
+namespace Tests\CodingStandards\Unit\CsFixer\Phpdoc\NoUselessAccessFixer;
+
+class Simple
+{
+    /**
+     * @access private
+     */
+    public const A = 'value';
```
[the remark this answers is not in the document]
grossmannmartin: That's right and it's intentional (that's why the test is like this).
In my opinion, visibility should always have a higher priority because a code directly depends on it (bad visibility break application, bad annotation don't do anything).

Access annotation is merely a hack, to achieve backward compatibility and it's said in PR description:

> Every constant have to have visibility defined with the access modifier or with @access annotation.
The first is recommended and in the future, the access annotation will not be present in the code at all.

packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
grossmannmartin: It goes through all annotations in one Phpdoc block in reverse order, and for each checks if it's `@access` annotation followed with public, protected, or private.

It goes in reverse order, so it's easier to keep the scope of searching a string value of access annotation.

packages/coding-standards/tests/Unit/CsFixer/Phpdoc/NoUselessAccessFixer/wrong/Simple.php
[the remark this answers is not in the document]
simara-svatopluk: @access is a user, check that link :-D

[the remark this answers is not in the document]
simara-svatopluk: If it's intentional, I would :heart: to have it expresses somewhere in the code. Maybe in the test scenario name? Maybe in commit message?

packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
simara-svatopluk: Interesting, thanks :-)

packages/coding-standards/tests/Unit/CsFixer/Phpdoc/NoUselessAccessFixer/NoUselessAccessFixerTest.php
grossmannmartin: Empty access annotations are now removed too

packages/coding-standards/src/CsFixer/Phpdoc/NoUselessAccessFixer.php
grossmannmartin: Hope it's better now :)

packages/coding-standards/src/CsFixer/Phpdoc/NoUselessAccessFixer.php
grossmannmartin: I extracted it, just used a little different naming.

packages/coding-standards/src/Sniffs/ForceLateStaticBindingForProtectedConstantsSniff.php
```diff
@@ -0,0 +1,100 @@
...
+                $file->fixer->beginChangeset();
+                $file->fixer->replaceToken($selfPosition, 'static');
+                $file->fixer->endChangeset();
+            }
+        }
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @return array
```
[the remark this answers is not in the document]
grossmannmartin: strings :D. Fixed, thanks 👍

packages/coding-standards/src/Sniffs/ForceLateStaticBindingForProtectedConstantsSniff.php
grossmannmartin: I rewrote it, but not exactly as you suggested, because I didn't want to do the same job twice (going through constant candidates, check and get the position of constant name).

Hope it's ok. 

Found a bug in the proccess, by the way, so I added another test case to cover the situations, where `self` is used not with constant.

packages/coding-standards/tests/Unit/CsFixer/Phpdoc/NoUselessAccessFixer/wrong/Simple.php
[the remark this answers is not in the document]
grossmannmartin: I created another test file, so it's covered a simple use-case and mixed access and visibility.
I add the note to commit message on the rebase

packages/framework/src/Command/DatabaseDumpCommand.php
grossmannmartin: Well, this seems to me like an edge case because they're inherited classes. I don't think it's commonly needed to inherit commands.
I don't want to use argument and option names anywhere else, because I don't see it as a need and don't know why I should need it (except extending command).

That's why I choose private :)

docs/upgrade/UPGRADE-unreleased.md
grossmannmartin: I'm not really sure if I want to force this specific fixer.
For me, it lies in a very blurry area. We want this fixer in the framework to allow users who extend the classes to change the behavior to match their needs. 
It helps with our specific use-case. 

But in a project, I don't think it's a necessary rule and I don't believe it should be forced. 
And I don't see any added value of forcing this rule in a project, do you?

packages/coding-standards/src/Sniffs/ForceLateStaticBindingForProtectedConstantsSniff.php
simara-svatopluk: There can even occur a static variable! After seeing static function I think it would be a good idea to check also static varibles...

docs/upgrade/UPGRADE-unreleased.md
simara-svatopluk: What about moving instructions to `coding-standards` section?

simara-svatopluk: I said my point of view. If you will agree with someone else that a sniff won't be part of our default coding-standards, it's ok for me... Please let me know ;-)

simara-svatopluk:
@grossmannmartin We have still couple of unresolved comments in this PR, please resolve them or answer

packages/coding-standards/src/Sniffs/ForceLateStaticBindingForProtectedConstantsSniff.php
grossmannmartin: I added a test for variables.

vitek-rostislav:
## Testing
Hi, great job there :+1: 

:heavy_check_mark: I have tested all the sniffs, and fixed `Shopsys\CodingStandards\Sniffs\ForceLateStaticBindingForProtectedConstantsSniff` setting in fixup commit so it is now applied correctly

:heavy_check_mark: I confirm that in `project-base`, I am able to change the value of a constant by overwritting the parent class from `framework` and it works :tada: (see https://github.com/shopsys/shopsys/commit/0124d46c83635cc2444aab6f9baa52e95cdf10c6)

PR #910, opened 2019-03-26
[project-base] availability of docker is now checked before installation

### LINKED ISSUE #908: Installation (linux) using install.sh script fails when the user is not in docker group
### What is happening
Reported by Adam Tyroň.

When installing the framework on Linux using `./scripts/install.sh`, there is a problem when the user is not in `docker` group (in such a case, `docker-compose` command must be executed with `sudo` privileges) - the whole installation fails with confusing error message.

### Expected result
The script works even in such a situation or the behavior is documented somehow.

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| If docker is not running or user is not allowed to run docker, installation fail on more descriptive message.
|New feature| No 
|Fixes issues| closes #908 

### DIFF
--- a/project-base/scripts/install.sh
+++ b/project-base/scripts/install.sh
@@ -1,10 +1,17 @@
 #!/bin/bash
-set -e
 numberRegex="^[0-9]+([.][0-9]+)?$"
 operatingSystem=""
 allowedValues=(1 2 3)
 projectPathPrefix=""
 echo This is installation script that will install demo Shopsys Framework application on docker with all required containers and with demo database created.
+
+docker ps -q > /dev/null 2> /dev/null
+if [[ "$?" != 0 ]]; then
+    1>&2 echo -e "\e[31mERROR:\e[0m Unable to connect to docker. Either the docker service is not running, or the current user is not allowed to run docker."
+    exit 1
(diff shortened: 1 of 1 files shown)


project-base/scripts/install.sh
```diff
@@ -1,11 +1,18 @@
 #!/bin/bash
-set -e
 numberRegex="^[0-9]+([.][0-9]+)?$"
 operatingSystem=""
 allowedValues=(1 2 3)
 projectPathPrefix=""
 echo This is installation script that will install demo Shopsys Framework application on docker with all required containers and with demo database created.
 
+docker ps -q > /dev/null 2> /dev/null
```
boris-brtan: does `&>` work also for osX shell?

boris-brtan: will this cover also `docker-compose` binary ? because only docker-compose and docker-sync are used in this script.

grossmannmartin: Yes, &> works in osx. But I'm not sure about windows.
docker-compose and docker-sync just calls Docker if I'm not mistaken, so I believe it should cover all situations.

PR #923, opened 2019-04-02
Revert "only failing versions of better-phpdoc-parser are set as conflicting"

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| During release of 7.1.0 were only versions 5.4.14 and 5.4.15 of better-phpdoc-parser set as conflicting, but after split the original problem reappeared on the Travis
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -149,7 +149,7 @@
   },
   "conflict": {
     "symfony/dependency-injection": "3.4.15|3.4.16",
-    "symplify/better-phpdoc-parser": "5.4.14|5.4.15",
+    "symplify/better-phpdoc-parser": ">=5.4.14",
     "twig/twig": "2.6.1"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


boris-brtan:
can we update Releaser with the notice `check whether actual version of resolved package is higher than the conflicted one via composer info ?`

boris-brtan:
maybe try different approach of including better-php-doc-parser config into `easy-coding-standard.yml`
```diff
imports:
    - { resource: '%vendor_dir%/shopsys/coding-standards/easy-coding-standard.yml', ignore_errors: true  }
+    - { resource: 'vendor/symplify/better-phpdoc-parser/config/config.yaml' }
```

boris-brtan:
in fact do we use functionality from `better-phpdoc-parser` ?

PR #932, opened 2019-04-08
[shopsys] fm_elfinder main filesystem volume driver

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| there was configuration for fm_elfinder that uses only local_filesystem and it is important to use main_filesystem for functional external storage adapter volumes like Google Cloud, AWS, Local, ...... after some discussion we have made a compromise to move barryvdh/elfinder-flysystem-driver from project-base into framework package and not use the barryvdh/elfinder-flysystem-driver extension method but to use `class_alias` functionality to overwrite class `elFinderVolumeFlysystem` because we needed to implement also custom functionality for GETURL plugin and missing thumbnail creating functionality
|New feature| Yes 
|Fixes issues| ... 


composer.json
```diff
@@ -25,7 +25,8 @@
       "project-base/app/AppCache.php",
       "project-base/app/AppKernel.php",
       "project-base/app/Bootstrap.php",
-      "project-base/app/Environment.php"
+      "project-base/app/Environment.php",
+      "project-base/app/elFinderVolumeFlysystem.php"
```
MattCzerner: I dont understand purpose of this file existence.

project-base/app/elFinderVolumeFlysystem.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+declare(strict_types=1);
+
+use Shopsys\ShopBundle\Component\Flysystem\VolumeDriver;
+
+class elFinderVolumeFlysystem extends VolumeDriver {}
```
MattCzerner: again, why?

MattCzerner: Why is it not possible to just extend classses from framework package and overwrite methods? why does it need to be registered as classmap? 

I would maybe have these files inside of framework since i dont see any reason why this kind of stuff should somoene have in a project-base.

project-base/src/Shopsys/ShopBundle/Component/Flysystem/VolumeDriver.php
```diff
@@ -0,0 +1,151 @@
...
+use elFinder;
+use Shopsys\ShopBundle\Component\Flysystem\Plugin\GetUrl;
+
+class VolumeDriver extends Driver
+{
+    protected function configure()
+    {
+        @parent::configure();
+
+        // set thumbnails path
```
MattCzerner: do we really need this comment?

project-base/src/Shopsys/ShopBundle/Component/Flysystem/VolumeDriver.php
```diff
@@ -0,0 +1,151 @@
...
+        $path = $this->options['tmbPath'];
```
MattCzerner: I know that this elfinder really does not care about variableNames but can atleast from our side make it as understandable as possible? this is not a $path for me it is $thumbnailPath..

Maybe you could create some methods with something like `createThumbnailPath()`

project-base/src/Shopsys/ShopBundle/Component/Flysystem/VolumeDriver.php
```diff
@@ -0,0 +1,151 @@
...
+                if ($this->_mkdir($path, '')) {
+                    $this->_chmod($path, $this->options['tmbPathMode']);
+                } else {
+                    $path = '';
+                }
+            }
+
+            $stat = $this->_stat($path);
+
+            if ($this->_dirExists($path) && $stat['read']) {
```
MattCzerner: What does that mean in a human language? if is dir readable then is tmpPathWritable = $stat['write'].

project-base/src/Shopsys/ShopBundle/Component/Flysystem/VolumeDriver.php
```diff
@@ -0,0 +1,151 @@
...
+            return $res;
+        }
+        return false;
+    }
+
+    /**
+     * @param string $path
+     * @param array $stat
+     */
+    protected function gettmb($path, $stat)
```
MattCzerner: In these you just use filesystem instead of methods such as copy() and so on right? Or is there something else that im Missing

project-base/src/Shopsys/ShopBundle/Component/Flysystem/VolumeDriver.php
```diff
@@ -0,0 +1,165 @@
+<?php
+
+declare(strict_types=1);
```
henzigo: You are declaring strict_types, but you are not using type hints

project-base/app/elFinderVolumeFlysystem.php
henzigo: > Even more this file violates PSR-4 standards

My mistake sorry

project-base/src/Shopsys/ShopBundle/Component/Flysystem/VolumeDriver.php
boris-brtan: `GetUrl::getFromMethod(string $path): string should be compatible with Barryvdh\elFinderFlysystemDriver\Plugin\GetUrl::getFromMethod($path)`
we cant redeclare extended methods so i will remove strict_type declarations then.

project-base/app/elFinderVolumeFlysystem.php
boris-brtan: i am for any better solution

henzigo:
Tests seems Ok

boris-brtan:
rebased on master and autosquashed

PR #942, opened 2019-04-12
absolute paths used via local filesystem flysystem now uses paths without drive letters

### LINKED ISSUE #913: Error with filesystem while using native installation on Windows
### What is happening
When installing application with demodata or uploading image in administration there is error File not found when moving file from local filesystem to abstract one.

Pasted_image_at_2019-03-27__2_24_PM

### Expected result
It should work.

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| methods of flysystem should not use absolute paths. Absolute path should be set in the process of creating the adapter. The problem that occurs for example windows file system uses drive letters and LocalAdapter needs to be initialised with path like `/` or `C:\` but if there is used absolute path like `C:\data\project-base\var\cache` via `LocalAdapter::has` method or others there is appended `/C:\data\project-base\var\cache` or 'C:\C:\data\project-base\var\cache' and used fopen function so the path is not correctly generated.
|New feature| No 
|Fixes issues| resolves #913 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -16,7 +16,7 @@ There you can find links to upgrade notes for other versions too.
             $finder->files()->in($origin);
             foreach ($finder as $file) {
         -        $filepath = $file->getPathname();
-        +        $filepath = TransformString::removeDriveLetterFromPath           ($file->getPathname());
+        +        $filepath = TransformString::removeDriveLetterFromPath($file->getPathname());
         ```
 
 
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -6,6 +6,20 @@ This guide contains instructions to upgrade from version v7.1.0 to Unreleased.
 There you can find links to upgrade notes for other versions too.
 
 ## [shopsys/framework]
+### Application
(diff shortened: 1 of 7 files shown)


docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -6,6 +6,20 @@ This guide contains instructions to upgrade from version v7.1.0 to Unreleased.
...
+- add `TransformString::removeDriveLetterFromPath` transformer for all absolute paths that could be based with drive letter file systems and used by `local_filesystem` service ([#942](https://github.com/shopsys/shopsys/pull/942))
+    - add TransformString::removeDriveLetterFromPath into `src/Shopsys/ShopBundle/DataFixtures/Demo/ImageDataFixture.php`
+        ```diff
+        protected function moveFilesFromLocalFilesystemToFilesystem(string $origin, string $target)
+        {
+            $finder = new Finder();
+            $finder->files()->in($origin);
+            foreach ($finder as $file) {
+        -        $filepath = $file->getPathname();
+        +        $filepath = TransformString::removeDriveLetterFromPath           ($file->getPathname());
```
DavidOstrozlik:
```suggestion
        +        $filepath = TransformString::removeDriveLetterFromPath($file->getPathname());
```

PR #949, opened 2019-04-16
[framework] fixed wrong behaviour while extending ProductFormType and CategoryFormType

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There was no dividing line between new Group added at the bottom of the form and plugin extensions. This PR simplifies the from structure which helps with the rendering of the separators.
|New feature| No 
|Fixes issues| ... 

Screenshot at Apr 16 12-52-19

This PR also deprecates the option `is_plugin_data_group` as it is no longer used and can be removed in the next major version.

### DIFF
--- a/packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
+++ b/packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
@@ -3,7 +3,6 @@
 namespace Shopsys\FrameworkBundle\Component\Plugin;
 
 use Shopsys\FrameworkBundle\Form\GroupType;
-use Symfony\Component\Form\Extension\Core\Type\FormType;
 use Symfony\Component\Form\FormBuilderInterface;
 
 class PluginCrudExtensionFacade
--- a/packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
+++ b/packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
@@ -35,13 +35,9 @@ class PluginCrudExtensionFacade
                 'label' => $crudExtension->getFormLabel(),
             ]);
 
-            $builderExtensionGroup->add($name, FormType::class, [
-                'compound' => true,
+            $builderExtensionGroup->add($key, $crudExtension->getFormTypeClass(), [
                 'render_form_row' => false,
-            ]);
-
-            $builderExtensionGroup->get($name)->add($key, $crudExtension->getFormTypeClass(), [
-                'render_form_row' => false,
+                'property_path' => sprintf('%s[%s]', $name, $key),
             ]);
 
             $builder->add($builderExtensionGroup);
--- a/packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
+++ b/packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
@@ -28,22 +28,23 @@ class PluginCrudExtensionFacade
      */
     public function extendForm(FormBuilderInterface $builder, $type, $name)
     {
-        $builder->add($name, FormType::class, [
-            'compound' => true,
-            'render_form_row' => false,
-        ]);
-
         $crudExtensions = $this->pluginCrudExtensionRegistry->getCrudExtensions($type);
+
         foreach ($crudExtensions as $key => $crudExtension) {
             $builderExtensionGroup = $builder->create($key . 'Group', GroupType::class, [
                 'label' => $crudExtension->getFormLabel(),
             ]);
 
-            $builderExtensionGroup->add($key, $crudExtension->getFormTypeClass(), [
+            $builderExtensionGroup->add($name, FormType::class, [
+                'compound' => true,
                 'render_form_row' => false,
             ]);
 
-            $builder->get($name)->add($builderExtensionGroup);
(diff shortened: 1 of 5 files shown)


vitek-rostislav:
Hi Tom, thanks for the PR, I am on it :+1:

packages/framework/src/Form/Admin/Category/CategoryFormType.php
```diff
@@ -234,20 +233,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'info_text' => t('You can upload following formats: PNG, JPG, GIF'),
             ]);
 
-        $builderPluginGroup = $builder->create('plugin', FormType::class, [
-            'inherit_data' => true,
-            'is_plugin_data_group' => true,
```
vitek-rostislav: is this option relevant yet? It was used only here (and it was a bad usage if I get it correctly) so I think it would be nice to remove `is_plugin_data_group` option completely from code and docs, what do you think?

TomasLudvik: Hello @vitek-rostislav,
yes you are right, this option is no longer used.

PetrHeinz: Are you sure, guys? See 
https://github.com/shopsys/shopsys/blob/master/docs/introduction/using-form-types.md#is_plugin_data_group

You might be thinking of `is_group_container_to_render_as_the_last_one`...?

packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
```diff
@@ -28,22 +28,23 @@ public function __construct(PluginCrudExtensionRegistry $pluginCrudExtensionRegi
...
         foreach ($crudExtensions as $key => $crudExtension) {
             $builderExtensionGroup = $builder->create($key . 'Group', GroupType::class, [
                 'label' => $crudExtension->getFormLabel(),
             ]);
 
-            $builderExtensionGroup->add($key, $crudExtension->getFormTypeClass(), [
+            $builderExtensionGroup->add($name, FormType::class, [
+                'compound' => true,
+                'render_form_row' => false,
+            ]);
```
PetrHeinz: The compound `FormType` has no other meaning here but to include the `$name` in the property path so it is mapped eg. to `CategoryData::$pluginData['shopsysProductFeedHeurekaBundleFormCategoryCrudExtension']`. You can achieve this much easily without the extra form using [the `property_path` option](https://symfony.com/doc/current/reference/forms/types/form.html#property-path) like this:

```php
$builderExtensionGroup->add($key, $crudExtension->getFormTypeClass(), [
    'render_form_row' => false,
    'property_path' => sprintf('%s[%s]', $name, $key),
]);
```

*(this code would replace the two form fields defined in lines 38-45, it behaved fine during testing)*

I can create a fixup commit for this if you want to - I didn't want to push it before I discussed it with you.

packages/framework/src/Form/Admin/Category/CategoryFormType.php
PetrHeinz: OK, my bad - you remove the last usage of the option. It is used in the template, but the boolean option is always `false` now. Thanks @TomasLudvik for clarification.

Removal of the option should be considered a BC break imo (even though we never defined BC in forms as described in #963). Maybe we can mark the option as deprecated and remove it in `8.0`?

packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
TomasLudvik: @PetrHeinz fixup would be nice, thanks.

packages/framework/src/Form/Admin/Category/CategoryFormType.php
TomasLudvik: I do not know how to mark option in form type as deprecated :(

packages/framework/src/Component/Plugin/PluginCrudExtensionFacade.php
TomasLudvik: @PetrHeinz thank you for your fixup, I like it :)

packages/framework/src/Form/Admin/Category/CategoryFormType.php
PetrHeinz: @TomasLudvik I've tried to mark it as deprecated in c2c3ad2af in similar fashion as in Symfony code. When I've configured a form type with `is_plugin_data_group = true`, a deprecation notice appeared in the debug toolbar as expected.

PetrHeinz: It might be even better to trigger the deprecation error when `false` is explicitly provided (which means the default option had to be removed, otherwise it would always trigger the error). See 5d49dc7e4.

Now, when no such deprecations are triggered on a project, the option may be in fact safely removed.

PetrHeinz:
@vitek-rostislav Hi, can you please take another look at it (review and test)? With @TomasLudvik we removed the redundant compound `FormType` by using [the `property_path` option](https://symfony.com/doc/current/reference/forms/types/form.html#property-path) and deprecated the now useless `is_plugin_data_group` option (you suggested to remove it, but it feels like a BC breaking change to remove form options).

PR #954, opened 2019-04-17
[framework] Current domain router match route for POST and ajax condition

stanoMilan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Previously for domain-specific front route with ajax condition and POST method, the 404 exception was thrown, even if the route should be matched. This is fixed now.
|New feature| No 
|Fixes issues| ... 

**If I declare route in routing_%locale%.yml:**
```
front_contact_form_send:
    path: /contactForm/
    defaults: { _controller: ShopsysShopBundle:Front\ContactForm:send }
    methods: [POST]
    condition: "request.isXmlHttpRequest()"
```

This route not found in routing. In Symfony profiler this route [matched](https://ibb.co/PcCz5HT) but [return 404 exception](https://ibb.co/txmWCX3).

### DIFF
--- a/packages/framework/src/Component/Router/DomainRouterFactory.php
+++ b/packages/framework/src/Component/Router/DomainRouterFactory.php
@@ -6,6 +6,7 @@ use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRouterFactory;
 use Symfony\Component\Config\Loader\LoaderInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Routing\RequestContext;
 use Symfony\Component\Routing\Router;
 
@@ -42,6 +43,11 @@ class DomainRouterFactory
     protected $routersByDomainId = [];
 
     /**
+     * @var \Symfony\Component\HttpFoundation\RequestStack
+     */
+    protected $requestStack;
+
+    /**
      * @param mixed $routerConfiguration
      * @param \Symfony\Component\Config\Loader\LoaderInterface $configLoader
      * @param \Shopsys\FrameworkBundle\Component\Router\LocalizedRouterFactory $localizedRouterFactory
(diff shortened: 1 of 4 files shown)


vitek-rostislav:
Hi @stanoMilan, thank you for the PR :+1: 

I will make sure it is processed during this or the upcoming sprint (ie. within 2 weeks)

grossmannmartin:
Hi @stanoMilan,
thanks for the PR.

Unfortunately, this fix is a BC break according to our [BC promise](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md#php-code) because in `DomainRouterFactory.php` you're adding a new argument to the constructor.

So we cannot accept it for the master branch.

I can think of two possible solutions.

1. Target this pull request to the `8.0` branch and this fix will be present in the upcoming major release (most likely in July).
In the end, you can always extend the class and fix it in your project, if you need the change immediately.

2. Solve it without BC break. One option that comes in my mind is to use setter injection.
- the constructor doesn't change at all
- a new method `setRequestStack(RequestStack $requestStack)`  will be added. Should be already deprecated, as we should still solve it with constructor injection in the next major.
- in service definition (services.yml) a method call has to take place.
  ```
    Shopsys\FrameworkBundle\Component\Router\DomainRouterFactory:
        calls:
           - method: setRequestStack
  ```
then we can merge it and release in the next minor version in two weeks or so.

How much this issue bugs you? Is it some common need, or more like edge case?

stanoMilan:
Hi @grossmannmartin 

I solve BC break. Created new method for set RequestStack.

Is it this ok?

Thank you

packages/framework/src/Component/Router/DomainRouterFactory.php
```diff
@@ -41,12 +42,18 @@ class DomainRouterFactory
...
+     */
+    protected $requestStack;
+
     /**
      * @param mixed $routerConfiguration
      * @param \Symfony\Component\Config\Loader\LoaderInterface $configLoader
      * @param \Shopsys\FrameworkBundle\Component\Router\LocalizedRouterFactory $localizedRouterFactory
      * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRouterFactory $friendlyUrlRouterFactory
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
```
grossmannmartin:
```suggestion
```

PR #958, opened 2019-04-18
[framework] cron works correctly with time config of "00"

### LINKED ISSUE #953: cron does not working with 00
### What is happening

If I declare cron in config as:
- { name: shopsys.cron, hours: '23', minutes: '00' }

This cron will never start

### Expected result

This cron runing in 23:00

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| fixes wrong behavior by padding the string to an equal length with zeros; a unit test dataset has been provided to be sure about the fix
|New feature| No 
|Fixes issues| fixes #953 


packages/framework/src/Component/Cron/CronTimeResolver.php
```diff
@@ -31,7 +31,7 @@ protected function isMatchWithTimeString($value, $timeString)
         $matches = null;
         foreach ($timeValues as $timeValue) {
             if ($timeValue === '*'
-                || $timeValue === (string)$value
+                || (int)$timeValue === $value
```
simara-svatopluk: This implementation seems to me a bit risky as it suddenly allows malformed input times to be "0"

```php
(int)'0' === 0; // true
(int)'00' === 0; // true
(int)'a' === 0; // true
(int)'000' === 0; // true
(int)'01' === 0; // false only
(int)'0a1' === 0; // true
(int)'0e1' === 0; // true
(int)'*/4' === 0; // true -- this case should be handled differently, but is caught by "0"
```

The string comparison was more reasonable and safe in my opinion...
```diff
  $timeValue === '*'
  || $timeValue === (string)$value
+ || $timeValue === str_pad((string) $value, 2, '0', STR_PAD_LEFT)
  || preg_match....
```

PetrHeinz: Fair point, I've used the `str_pad`ding of the value so both strings are of equal length.

PetrHeinz:
@simara-svatopluk I've squashed the commits and redone the implementation :slightly_smiling_face:

packages/framework/src/Component/Cron/CronTimeResolver.php
```diff
@@ -31,7 +31,7 @@ protected function isMatchWithTimeString($value, $timeString)
         $matches = null;
         foreach ($timeValues as $timeValue) {
             if ($timeValue === '*'
-                || $timeValue === (string)$value
+                || $timeValue === str_pad($value, strlen($timeValue), '0', STR_PAD_LEFT)
```
simara-svatopluk: This whole `if` is difficult to understand for me, I would be pleased if you refactored it into 3 functions:
`if ($this->matchesSingleValue($timeValue, $value)`

```php
protected function matchesSingleValue($timeValue, $value) {
return $timeValue === '*' || $this->matchesStringValue($timeValue, $value) || $this->matchesHourPart($timeValue, $value);
}

protected function matchesStringValue($timeValue, $value)...

protected function matchesHourPart($timeValue, $value) ...
```

It would greatly improve readability and maintainability, but it is not "must to be" to pass CR

And of course, you are welcomed to come with better naming

simara-svatopluk:
CR :heavy_check_mark: , going to test

packages/framework/src/Component/Cron/CronTimeResolver.php
PetrHeinz: I would rather skip the extraction, I don't think it's worth the effort - the result wouldn't be much more readable imo...

simara-svatopluk:
tested :heavy_check_mark:

packages/framework/src/Component/Cron/CronTimeResolver.php
simara-svatopluk: Just by looking at this line
`preg_match('@^\*/(\d+)$@', $timeValue, $matches) && $value % $matches[1] === 0 // syntax */[int]`

I have no idea what is going on, even with the comment :-D

PR #969, opened 2019-04-23
[shopsys] unified phing targets "phpstan" and "tests-unit" in monorepo

PetrHeinz:
It can be confusing that we have target `tests-unit` in monorepo which runs the unit tests in project-base only (and having to run `tests-packages` and `tests-utils` as well to unit test everything).

When an already existing phing target is defined a copy is made by prefixing it with a project name (eg. `shopsys_framework.clean` when a `clean` phing target in project named "Shopsys Framework" is redefined).
This can be used for extending a phing target and adding some functionality.

Using this approach, phing targets `phpstan-packages`, `tests-packages` and `tests-utils` were removed and their functionality was added into `phpstan` and `tests-unit`.
The same could be done to phing targets `phplint`, `phplint-diff`, `ecs`, `ecs-diff`, `twig-lint`, `twig-lint-diff`, `yaml-lint`, `eslint-check`, `eslint-check-diff` and maybe `dump-translations`, avoiding the need for all the `*-packages` and `*-utils` phing targets. I didn't want to go the extra mile while we're not sure this is the way we want to go...

Also, the project name in monorepo's build.xml was changed to "Shopsys Framework (monorepo)" to avoid the ambiguity.

| Q             | A
| ------------- | ---
|Description, reason for the PR| avoid confusing phing targets in monorepo while keeping things DRY
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No (only affects the monorepo) 
|Fixes issues| ... 

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project name="Shopsys Framework" default="list">
+<project name="Shopsys Framework (monorepo)" default="list">
 
     <property name="path.root" value="./project-base"/>
     <property name="path.vendor" value="./vendor"/>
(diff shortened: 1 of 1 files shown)


boris-brtan:
Hi @PetrHeinz ,
seems like you would need to update also CI stage in `project-base/docker/php-fpm/Dockerfile` with appropriate upgrade notes.

PR #972, opened 2019-04-26
[framework] getCollationsByLocale now gets collations by get method

MattCzerner:
- 
| Q             | A
| ------------- | ---
|Description, reason for the PR| If user uses their own collations by inheritance, this method still returned value from old collation array. 
|New feature| No
|Fixes issues|


boris-brtan:
why was this closed and branch remained? Is there any future plan with this?

PR #978, opened 2019-04-30
Configure travis to check standards and unit tests in monorepo

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Travis will check every push to PR. It will not allow to merge PR without  successful check.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+language: php
+
+php:
+  - 7.1
+
+cache:
+  directories:
+    - ~/.composer/cache
+    - ~/.npm
+
+env:
+  - ELASTIC_SEARCH_INDEX_PREFIX=""
+
+before_install:
+  - phpenv config-add project-base/docker/php-fpm/php-ini-overrides.ini
+  - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - phpenv config-rm xdebug.ini || return 0
+  - composer global require hirak/prestissimo
+
+install:
+  - COMPOSER_MEMORY_LIMIT=-1 travis_retry composer install --optimize-autoloader --no-interaction --no-progress --dev
+  - php phing npm
+
+script:
+  - php phing standards
+  - php phing tests-static
+
--- a/composer.json
+++ b/composer.json
@@ -120,7 +120,7 @@
     "shopsys/jparser": "^0.1",
     "shopsys/postgres-search-bundle": "^0.1",
     "slevomat/coding-standard": "^4.6.0",
-    "snc/redis-bundle": "^2.1.8",
+    "snc/redis-bundle": "^2.1.7",
     "squizlabs/php_codesniffer": "^3.2.0",
     "stof/doctrine-extensions-bundle": "^1.3.0",
     "superbalist/flysystem-google-storage": "^7.1",
@@ -152,7 +152,8 @@
   "conflict": {
     "symfony/dependency-injection": "3.4.15|3.4.16",
     "symplify/better-phpdoc-parser": "5.4.14|5.4.15",
-    "twig/twig": "2.6.1"
+    "twig/twig": "2.6.1",
+    "snc/redis-bundle": "2.1.8|2.1.9"
   },
   "scripts": {
     "post-install-cmd": [
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
 # Shopsys Framework
+![travis-badge](https://travis-ci.org/shopsys/shopsys.svg?branch=master)
+
 Shopsys Framework is a **fully functional ecommerce platform for businesses transitioning into tech-companies with their own software development team**. 
 It contains the most common B2C and B2B features for online stores, and its infrastructure is prepared for high scalability.
 
--- a/packages/framework/composer.json
+++ b/packages/framework/composer.json
@@ -81,7 +81,7 @@
         "shopsys/migrations": "dev-master",
         "shopsys/form-types-bundle": "dev-master",
         "shopsys/plugin-interface": "dev-master",
-        "snc/redis-bundle": "^2.1.8",
+        "snc/redis-bundle": "^2.1.7",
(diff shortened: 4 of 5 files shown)


.travis.yml
```diff
@@ -0,0 +1,26 @@
+language: php
+
+php:
+  - 7.1
```
MattCzerner: why do we use only 7.1? It should be really easy to add all versions we support, is there any reason not to?

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+
+cache:
+  directories:
+    - ~/.composer/cache
+    - $HOME/.npm
```
MattCzerner: you use `$HOME` once and then `~`. Why? Arent they same?

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+env:
+  - ELASTIC_SEARCH_INDEX_PREFIX=""
+
+before_install:
+  - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - phpenv config-rm xdebug.ini || return 0
+
+install:
+  - composer global require hirak/prestissimo
+  - COMPOSER_MEMORY_LIMIT=-1 travis_retry composer install --optimize-autoloader --no-interaction --no-progress --dev
```
MattCzerner: `COMPOSER_MEMORY_LIMIT` is an environment variable too, why not define it next to the `ELASTIC_SEARCH_INDEX_PREFIX`

.travis.yml
MattCzerner: also, you can cache things just using keyword npm, which will cache `$HOME/.npm` or `node_modules`. Have you considered it?

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+
+env:
+  - ELASTIC_SEARCH_INDEX_PREFIX=""
+
+before_install:
+  - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - phpenv config-rm xdebug.ini || return 0
+
+install:
+  - composer global require hirak/prestissimo
```
MattCzerner: Wonder if this shouldnt be part of the before_install section. What do you think? Because this is not really an install thing in my opinion, more like setting a helper for installation

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - COMPOSER_MEMORY_LIMIT=-1 travis_retry composer install --optimize-autoloader --no-interaction --no-progress --dev
+  - php phing npm
+
+script:
+  - php phing standards
+  - php phing tests-static
```
MattCzerner: Maybe im stupid, but `tests-static` is wierd, it only depends on unit-tests alone, why not just use that instead of alias phing target that is just something else.

.travis.yml
MattCzerner: Also, Im just thinking here and maybe it is not good idea but:

We could cache a vendor folder too, we do not keep composer.lock file anyway so it will update dependencies everytime right? And we got few seconds faster travis build

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - phpenv config-rm xdebug.ini || return 0
+
+install:
+  - composer global require hirak/prestissimo
+  - COMPOSER_MEMORY_LIMIT=-1 travis_retry composer install --optimize-autoloader --no-interaction --no-progress --dev
+  - php phing npm
+
+script:
+  - php phing standards
```
MattCzerner: Is there any advantage in having these 2 tasks on their own lines? Does it better profile in travis output for example?

.travis.yml
henzigo: I used `~`. I didn't use keyword npm, because then i could not cache composer.

.travis.yml
henzigo: I moved that

.travis.yml
henzigo: Ok i moved that. It was because i found that command on internet :D

.travis.yml
henzigo: Yes, thats bug in our application. But we are always using `tests-static` instead of `tests-unit` so i prefer use it here too.

.travis.yml
henzigo: You can see in output console that there are times for every command. So i prefer that, because it says how long does it takes.

.travis.yml
henzigo: I do not like cache final results of `vendor` or `node_modules` folders, because it can make a problems. I found related issue https://github.com/travis-ci/travis-ci/issues/4579

henzigo: Oh after i moved `COMPOSER_MEMORY_LIMIT` to env parameters travis started to fail on composer, so i moved that back

composer.json
```diff
@@ -120,7 +120,7 @@
     "shopsys/jparser": "^0.1",
     "shopsys/postgres-search-bundle": "^0.1",
     "slevomat/coding-standard": "^4.6.0",
-    "snc/redis-bundle": "^2.1.8",
+    "snc/redis-bundle": "2.1.7",
```
MattCzerner: Why?

.travis.yml
MattCzerner: OK

composer.json
```diff
@@ -152,7 +152,8 @@
   "conflict": {
     "symfony/dependency-injection": "3.4.15|3.4.16",
     "symplify/better-phpdoc-parser": "5.4.14|5.4.15",
-    "twig/twig": "2.6.1"
+    "twig/twig": "2.6.1",
+    "snc/redis-bundle": "2.1.8|2.1.9"
```
boris-brtan: 2.1.10 is not a problem?

boris-brtan: will be fixed in #1013

PR #981, opened 2019-05-02
[framework] CategoryDomainNotFoundException $categoryId is now mandatory

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There was optional parameter when it should be mandatory
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -18,6 +18,7 @@ There you can find links to upgrade notes for other versions too.
         -        $filepath = $file->getPathname();
         +        $filepath = TransformString::removeDriveLetterFromPath($file->getPathname());
         ```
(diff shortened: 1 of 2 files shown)


PetrHeinz:
This is the same for all `*DomainNotFoundException` classes (so `Brand`, `Category`, `Country`, `Payment`, `Product` and `Transport`).

I think the reason is that with Doctrine autoincremental ID, the property is actually of the type `int|null` as it is `null` before the entity is successfully persisted. The original authors in https://github.com/shopsys/shopsys/pull/165 probably saw more value with reliable throwing of a single exception in such cases instead of triggering a type error in this special case. There is even extra logic handling this case (the `"without ID"` string).

If you just want to make the argument mandatory instead of optional, I guess you should use `?int`.

---

Also, the upgrade notes for changes in the `8.0` branch belong to [UPGRADE-v8.0.0.md](https://github.com/shopsys/shopsys/blob/8.0/docs/upgrade/UPGRADE-v8.0.0.md), so we don't have to resolve conflicts on upgrade notes all the time.

PR #982, opened 2019-05-03
[framework] parameters moved from services to parameters_common

### LINKED ISSUE #976: Parameters should not be in services.yml
### What is happening

In frameworkBundle there are parameters in `services.yml` file. It's really confusing and bad practice. Someone can simply miss that, because programmers will be not searching parameters in `services.yml`.

### Expected result

In frameworkBundle could be `parameters_common.yml` file, where will be default parameters (e.g. `shopsys.admin_locale`) and `parameters_common.yml` in project-base should be cleaned.

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Now it's easier to be aware of possible parameters that can be overriden from framework in a project, because now they're in separate file.
|New feature| No 
|Fixes issues| closes #976 

### DIFF
--- a/packages/framework/src/DependencyInjection/ShopsysFrameworkExtension.php
+++ b/packages/framework/src/DependencyInjection/ShopsysFrameworkExtension.php
@@ -20,6 +20,7 @@ class ShopsysFrameworkExtension extends Extension
     {
         $loader = new YamlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('directories.yml');
+        $loader->load('parameters_common.yml');
         $loader->load('services.yml');
         $loader->load('paths.yml');
 
--- a/packages/framework/src/Resources/config/parameters_common.yml
+++ b/packages/framework/src/Resources/config/parameters_common.yml
@@ -5,15 +5,6 @@ parameters:
     # Seed for demo data generation
     faker.seed: 1234
 
-    # Use Shopsys validator factory for fp/jsformvalidator-bundle
-    fp_js_form_validator.factory.class: Shopsys\FrameworkBundle\Form\JsFormValidatorFactory
-
-    # Use Shopsys routing for knplabs/knp-menu-bundle that ignores missing parameters in routes
-    knp_menu.factory_extension.routing.class: Shopsys\FrameworkBundle\Model\AdminNavigation\RoutingExtension
-
-    # Use Shopsys Translatable listener for prezent/doctrine-translatable to allow set fallback language on postPersist event
-    prezent_doctrine_translatable.listener.class: Shopsys\FrameworkBundle\Model\Localization\TranslatableListener
-
     # Administration locale. Must be registered in domains.yml
     shopsys.admin_locale: en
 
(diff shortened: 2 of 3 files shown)


packages/framework/src/Resources/config/parameters_common.yml
```diff
@@ -0,0 +1,21 @@
+parameters:
+    # Default Redis keys prefix
+    build-version: '0000000000000000'
+
+    # Seed for demo data generation
+    faker.seed: 1234
+
+    # Use Shopsys validator factory for fp/jsformvalidator-bundle
+    fp_js_form_validator.factory.class: Shopsys\FrameworkBundle\Form\JsFormValidatorFactory
```
pk16011990: I think that overriding of class parameters like this should stay in services.yml, because It is part for building DI container and their services...

It is my point of view...

packages/framework/src/DependencyInjection/ShopsysFrameworkExtension.php
```diff
@@ -18,6 +18,7 @@ public function load(array $configs, ContainerBuilder $container)
     {
         $loader = new YamlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('directories.yml');
+        $loader->load('parameters_common.yml');
```
pk16011990: `parameters_common.yml` is imported in `AppKernel` yet, why is it here?

pk16011990: Oh, I see... it is in different directory...  What do you think about merging this two parameters_common.yml files into one?

grossmannmartin: Well, I don't think it's possible. The only way how to do it is to have all configuration in project itself because it may be needed to adjust the values.
And then I would like to have some default value in a framework.

packages/framework/src/Resources/config/parameters_common.yml
grossmannmartin: A fair point. You're right it's more like a "service" thing. 
I'm not sure why it needs to be adjusted via parameters... seems a little smelly to me, but we still can't do anything about it. 

I give it a second thought. Thanks for the opinion 👍

packages/framework/src/DependencyInjection/ShopsysFrameworkExtension.php
pk16011990: That is fair point from you. I didnt think in context what is a framework and what is a project. I undestand it now.

packages/framework/src/Resources/config/parameters_common.yml
grossmannmartin: I moved all three parameters with overriding of class back to the `services.yml` as it really makes much more sense.

packages/framework/src/DependencyInjection/ShopsysFrameworkExtension.php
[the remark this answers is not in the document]
grossmannmartin: I would prefer to keep the naming consistent.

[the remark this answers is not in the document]
PetrHeinz: Fine then

PR #999, opened 2019-05-06
[framework] used ICU collations to assure correct sorting throughout OSs

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| According to https://www.2ndquadrant.com/en/blog/icu-support-postgresql-10/, it is much more better to use ICU collations. These colllations are always available, not like ones we are using now. Also for example, for current collations letter "Ú" is sorted behind letter "Z"
|New feature| No
|Fixes issues| 

### DIFF
--- a/packages/framework/src/Command/CreateDatabaseCommand.php
+++ b/packages/framework/src/Command/CreateDatabaseCommand.php
@@ -186,7 +186,7 @@ class CreateDatabaseCommand extends Command
             'SELECT 1
             FROM pg_collation
             WHERE collnamespace = (SELECT oid FROM pg_namespace WHERE nspname = \'pg_catalog\')
-            AND collencoding = pg_char_to_encoding(\'UTF8\')
+            AND (collencoding = pg_char_to_encoding(\'UTF8\') OR collencoding = -1)
             AND collname = ?',
             [$collation]
         );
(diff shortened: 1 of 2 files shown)


packages/framework/src/Command/CreateDatabaseCommand.php
```diff
@@ -186,7 +186,7 @@ private function createCollationIfNotExists($collation, $locale)
             'SELECT 1
             FROM pg_collation
             WHERE collnamespace = (SELECT oid FROM pg_namespace WHERE nspname = \'pg_catalog\')
-            AND collencoding = pg_char_to_encoding(\'UTF8\')
+            AND (collencoding = pg_char_to_encoding(\'UTF8\') OR collencoding = -1)
```
simara-svatopluk: Sorry I don't understand this line, what does it mean?

MattCzerner: This creates a condition in which only encoding compatible with UTF-8 (first part of condition),or -1 if it works for any encoding

pk16011990: https://www.postgresql.org/docs/10/catalog-pg-collation.html - 
collencoding - Encoding in which the collation is applicable, or -1 if it works for any encoding

MattCzerner:
I dont think that upgrade note for a collations are necessary, there are 733 records of encodings in database and in my opinion it does not matter if there are 8 more that are not used because we are not using like 710 of them.

PR #1005, opened 2019-05-09
code-quality-principles.md: fixed typos

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| typos
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/docs/contributing/code-quality-principles.md
+++ b/docs/contributing/code-quality-principles.md
@@ -59,7 +59,7 @@ It may be difficult to create the dependency - it may have also dependencies or 
 Once creating the dependency is not the class responsibility, the class code is much more straight-forward.
 
 We inject only dependencies that the class needs, it is forbidden to pass the service locator to the class.
-With service locator injection it is difficult to understand what the class really do (or need) and such class is difficult to test.
+With service locator injection it is difficult to understand what does the class really do (or need) and such class is difficult to test.
 
 We use constructor DI so the class is always in a consistent state.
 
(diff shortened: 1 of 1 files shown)


docs/contributing/code-quality-principles.md
```diff
@@ -91,8 +91,8 @@ This leads to safer code.
 
 ### Reasons
 
-We use scalar types (`int`, `float`, `string`, `bool`) as much as possible in new code.
-Using scalar types without strict mode leads to automatic type cast that can be even harmful (eg. input string `""` is cast to integer `0`).
+We use scalar type hints (`int`, `float`, `string`, `bool`) as much as possible in new code.
```
simara-svatopluk: I prefer to keep term `scalar types` as it is used in https://wiki.php.net/rfc/scalar_type_hints_v5

Anyway if you thing `scalar type hints` is more obvious to reader, please fix the term in the whole article

PetrHeinz: The term that is used there is *scalar type declarations*.

I think that saying we use *scalar types* as much as possible means that we try to avoid arrays and objects.

vitek-rostislav: Yes, that was my feeling too - I believe there is a difference between "scalar types" (ie. integers, booleans, ...) and "scalar type hints"

PetrHeinz: However, I don't see any meaningful difference between *"scalar type declarations"* and *"scalar type hints"*, so we can keep the terminology from [wiki.php.net](https://wiki.php.net/rfc/scalar_type_hints_v5).

vitek-rostislav: ...there are two more occurrences of "scalar types" in the article, both of them are fine by me...

PR #962, opened 2019-04-23
[shopsys] upgrade symplify/easy-coding-standard to ^5.4.16

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| https://github.com/shopsys/shopsys/issues/957
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| closes #957  
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1008, opened 2019-05-10
[coding-standards] include config of better-phpdoc-parser in ECS config

### LINKED ISSUE #957: Travis builds for shopsys packages are failing
### What is happening
After the release of `7.1.0`, Travis builds started to failing - it is caused by [this commit](https://github.com/shopsys/shopsys/commit/073324a5ea257a85122fbc97ea08bfb978ddbeab) - Some Symplify services can not be autowired (see eg. https://travis-ci.org/shopsys/framework/jobs/521133575).
@TomasLudvik prepared [the fix by reverting the commit](https://github.com/shopsys/shopsys/pull/923), however, @boris-brtan has a valid point that it could be solved by including better-php-doc-parser config into our `easy-coding-standard.yml`.

I am wondering how could this happen when upgrading to the next minor of `symplify/better-php-doc-parser`, @TomasVotruba do you have any information on that, please?

### Expected result
All Shopsys Travis builds are passing.

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Travis builds fail on broken autowiring of services (see #957) due to using different versions of `symplify/better-phpdoc-parser`. This PR fixes the builds of the isolated packages by providing a link to DI service configuration.
|New feature| No 
|Fixes issues| fixes #957, closes #923 and closes #962 

:heavy_check_mark: `shopsys/framework` package passes the Travis build when using the `dev-ph-travis-ecs-fix` version of `shopsys/coding-standards`, see https://github.com/shopsys/framework/compare/ph-travis-ecs-fix

:heavy_check_mark:  `shopsys/coding-standards` package passes the Travis build during unit tests (which were misconfigured previously), see https://github.com/shopsys/coding-standards/compare/ph-travis-ecs-fix

:hourglass_flowing_sand: cannot easily test Travis build after splitting the monorepo as the packages use `shopsys/coding-standards` in the `dev-master` version, due to the two checks above I am confident the build will pass

### DIFF
--- a/packages/coding-standards/easy-coding-standard.yml
+++ b/packages/coding-standards/easy-coding-standard.yml
@@ -1,6 +1,9 @@
 imports:
     # for PSR-2 description, see: http://www.php-fig.org/psr/psr-2/
     - { resource: '%vendor_dir%/symplify/easy-coding-standard/config/psr2.yml' }
+    # some checkers use BetterPhpdocParser services which must be configured (file renamed in symplify/better-phpdoc-parser v5.4.14)
+    - { resource: '%vendor_dir%/symplify/better-phpdoc-parser/config/config.yml', ignore_errors: true }
+    - { resource: '%vendor_dir%/symplify/better-phpdoc-parser/config/config.yaml', ignore_errors: true }
 
 services:
     _defaults:
--- a/packages/coding-standards/tests/Unit/CsFixer/ForbiddenPrivateVisibilityFixer/config.yml
+++ b/packages/coding-standards/tests/Unit/CsFixer/ForbiddenPrivateVisibilityFixer/config.yml
@@ -1,3 +1,6 @@
+imports:
+    - { resource: '../../../config.yml' }
+
 services:
     Shopsys\CodingStandards\CsFixer\ForbiddenPrivateVisibilityFixer:
         analyzed_namespaces:
--- a/packages/coding-standards/tests/Unit/CsFixer/ForbiddenDumpFixer/config.yml
+++ b/packages/coding-standards/tests/Unit/CsFixer/ForbiddenDumpFixer/config.yml
@@ -1,2 +1,5 @@
+imports:
+    - { resource: '../../../config.yml' }
+
 services:
     Shopsys\CodingStandards\CsFixer\ForbiddenDumpFixer:
--- a/packages/coding-standards/tests/Unit/CsFixer/MissingButtonTypeFixer/config.yml
+++ b/packages/coding-standards/tests/Unit/CsFixer/MissingButtonTypeFixer/config.yml
@@ -1,2 +1,5 @@
+imports:
+    - { resource: '../../../config.yml' }
+
 services:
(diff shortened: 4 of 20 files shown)


packages/coding-standards/easy-coding-standard.yml
```diff
@@ -1,6 +1,9 @@
 imports:
     # for PSR-2 description, see: http://www.php-fig.org/psr/psr-2/
     - { resource: '%vendor_dir%/symplify/easy-coding-standard/config/psr2.yml' }
+    # some checkers use BetterPhpdocParser services which must be configured (file renamed in symplify/better-phpdoc-parser v5.4.14)
+    - { resource: '%vendor_dir%/symplify/better-phpdoc-parser/config/config.yml', ignore_errors: true }
+    - { resource: '%vendor_dir%/symplify/better-phpdoc-parser/config/config.yaml', ignore_errors: true }
```
boris-brtan: what about just one line with `config.y*ml` and maybe no ignore_error would be needed

packages/coding-standards/tests/Unit/CsFixer/ForbiddenPrivateVisibilityFixer/config.yml
```diff
@@ -1,3 +1,6 @@
+imports:
+    - { resource: '../../../config.yml' }
```
boris-brtan: is it not possible to do this via https://github.com/shopsys/shopsys/blob/ph-travis-ecs-fix/packages/coding-standards/tests/autoload.php ?
so no repetition will be needed ?
will that be a problem when testing via monorepo ?

boris-brtan:
### TESTS
tested via [shopsys/migrations] where there is dev-dependency on coding-standards package
https://travis-ci.org/shopsys/migrations/builds/530720706

packages/coding-standards/tests/Unit/CsFixer/ForbiddenPrivateVisibilityFixer/config.yml
PetrHeinz: The autoload.php is on a lower level then Symfony DIC build, so I believe it!s not possible to resolve that there.

In monorepo, everything is fine (you can see on our CI that tests pass).

packages/coding-standards/easy-coding-standard.yml
PetrHeinz: I haven't thought of that, that seem slightly less readable to me, but maybe it's better... I would rather use `config.y{,a}ml` or `config.{yml,yaml}` than `config.y*ml` to be more explicit.

To be honest, I'm slightly more inclined to leave it as-is in sake of readability. What do you think?

PetrHeinz:
Thanks for the review and testing, @boris-brtan. The reason #962 exists is to fix the Travis problem, so I suppose we can close it with this PR (@vitek-rostislav, as the author, is ok with that).

You're right that we might wanna change the PHP versions we use, but I would say it's outside of the scope of this PR.

boris-brtan:
OK then, after merge we could remove splitted branches from packagist and github repositories

Issue #872, opened 2019-02-26
BC promise is lacking section about dependencies


PetrHeinz:
### What is happening

Our [BC promise](https://github.com/shopsys/shopsys/blob/v7.0.0-beta6/docs/contributing/backward-compatibility-promise.md) is currently lacking sections about dependencies (Composer and NPM), which might be useful to address.

### Expected result

Backward compatibility of dependencies and reasonable expectations about it is explained in our BC promise in similar fashion and scope as other subsections of *The BC Promise in Detail*.

boris-brtan:
could be good to solve this problem on this commit https://github.com/shopsys/shopsys/pull/978/commits/f94aef8723d02ed45f69b5b071b76e9cbe915391

PetrHeinz:
I've separated the question of Phing targets into its own issue #1012

PR #1007, opened 2019-05-10
[shopsys] simplify travis

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| after fixes of env parameters for elastic (in #961) travis script can be simplified.
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1013, opened 2019-05-10
[shopsys] composer.json: revert BC breaking change of snc/redis-bundle dependency

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #978 the supported versions of `snc/redis-bundle` package were changed which breaks backward compatibility. Even though we don't define BC in terms of dependencies precisely (see #872), you cannot use `shopsys/framework` in version `dev-master` using the `shopsys/project-base` in the `7.0.0` version as a base repository without further changes (see below).
|New feature| No 
|Fixes issues| ... 

When using `shopsys/project-base` in version `7.0.0` and replacing all `shopsys/*` dependencies to `dev-master` (current master is on b3128c4a7) version, following error is thrown during `composer install`:

```
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - shopsys/framework dev-master requires snc/redis-bundle ^2.1.7 -> satisfiable by snc/redis-bundle[2.1.8, 2.1.9].
    - snc/redis-bundle 2.1.8 conflicts with shopsys/framework[dev-master].
    - snc/redis-bundle 2.1.9 conflicts with shopsys/framework[dev-master].
    - Installation request for shopsys/framework dev-master -> satisfiable by shopsys/framework[dev-master].
```

The modification of `composer.json` was originally done in #978 so that Travis build succeeds (not sure about the exact reason). This goal of this PR is to pass on Travis while still being backward-compatible with `shopsys/project-base` in `7.0.0`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
 
 before_install:
   - phpenv config-add project-base/docker/php-fpm/php-ini-overrides.ini
+  - echo '' | pecl install redis-4.1.1
   - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
   - phpenv config-rm xdebug.ini || return 0
   - composer global require hirak/prestissimo
--- a/project-base/composer.json
+++ b/project-base/composer.json
@@ -75,7 +75,7 @@
         "shopsys/product-feed-heureka-delivery": "dev-master",
         "shopsys/product-feed-zbozi": "dev-master",
         "shopsys/product-feed-google": "dev-master",
-        "snc/redis-bundle": "^2.1.7",
+        "snc/redis-bundle": "^2.1.8",
         "stof/doctrine-extensions-bundle": "^1.3.0",
         "symfony/assetic-bundle": "^2.8.2",
         "symfony/monolog-bundle": "^3.1.2",
@@ -99,8 +99,7 @@
     },
     "conflict": {
         "symfony/dependency-injection": "3.4.15|3.4.16",
-        "twig/twig": "2.6.1",
-        "snc/redis-bundle": "2.1.8|2.1.9"
+        "twig/twig": "2.6.1"
     },
     "scripts": {
         "post-install-cmd": [
--- a/composer.json
+++ b/composer.json
@@ -120,7 +120,7 @@
     "shopsys/jparser": "^0.1",
     "shopsys/postgres-search-bundle": "^0.1",
     "slevomat/coding-standard": "^4.6.0",
-    "snc/redis-bundle": "^2.1.7",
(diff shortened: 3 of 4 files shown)


boris-brtan:
maybe we could compare version of php module for redis  on travis and in our Dockerfile 
```
RUN pecl install redis-4.1.1 && \
    docker-php-ext-enable redis
```

PetrHeinz:
@boris-brtan You're right about the PECL extension version. There was 3.1.4 installed on Travis, we need 4.1.1. See https://travis-ci.org/shopsys/shopsys/builds/530758153 for details.

.travis.yml
```diff
@@ -13,6 +13,7 @@ env:
 
 before_install:
   - phpenv config-add project-base/docker/php-fpm/php-ini-overrides.ini
+  - echo '' | pecl install redis-4.1.1
```
boris-brtan: what about .travis.yml for package framework and all the packages that uses framework ?

boris-brtan: what does this do? i think it seems that it just use the default value for some question in interactive mode yes ?
i would maybe put the thing that we skip or install additionaly in comment

project-base/composer.json
```diff
@@ -75,7 +75,7 @@
         "shopsys/product-feed-heureka-delivery": "dev-master",
         "shopsys/product-feed-zbozi": "dev-master",
         "shopsys/product-feed-google": "dev-master",
-        "snc/redis-bundle": "^2.1.7",
+        "snc/redis-bundle": "^2.1.8",
```
boris-brtan: nice, there is no need for UPGRADE notes :+1:  from #978

boris-brtan:
i just noticed another thing that in README.md there is not the same link for OK build to travis at it is for packages

.travis.yml
PetrHeinz: I don't think it related to PHP 7.1 != PHP 7.1.0.
This is the only Travis script where the Redis extension is relevant, I think. It's the only one where the full DIC of the application is built (and lazy loading is working only in the newer version).
The `echo '' |` is for non-interactive PECL installation.

PetrHeinz:
@boris-brtan Yes, there is a different alt text and it's not a link to the Travis page, but to the image itself. Could you create a PR to fix that please?

.travis.yml
PetrHeinz: What do you suggest to do?

boris-brtan: hmmm, there are skipped 2 thing 
```
enable igbinary serializer support? [no] 
enable lzf compression support? [no]
```
i have already [PR](#1007) for improving travis check for monorepo so i am OK with this right now then.

Issue #194, opened 2018-06-05
Upgrade coding standards in http-smoke-testing package


boris-brtan:
Coding standards in http-smoke-testing are not up to date because there is need to have this package compatible with PHP-5.6 and PHP-7.0 and new coding standards are compatible with PHP-7.1 and up.

### What is happening

travises jobs call old coding standards because the support for older PHP versions.
### Expected result

to have it solved somehow

PR #1014, opened 2019-05-10
[http-smoke-testing] use current shopsys/coding-standards package

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Travis build failed because of coding standard conflict (it uses legacy 3.0.2 version of shopsys/coding-standards because of support for PHP 5.6, see #194)
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No (only dev dependencies were changed) 
|Fixes issues| finally fixes the stale #194 

After this PR:
- the shopsys/http-smoke-testing package still supports PHP 5.6
- Travis now runs PHP lint and unit tests in all supported versions
- Travis now runs coding standards check only in PHP 7.1

### DIFF
--- a/packages/http-smoke-testing/.travis.yml
+++ b/packages/http-smoke-testing/.travis.yml
@@ -21,8 +21,8 @@ matrix:
             env: CHECK_STANDARDS="true"
 
 install:
-    - travis_retry composer update $DEPENDENCIES
     - if [[ "$CHECK_STANDARDS" = "true" ]]; then composer require --dev shopsys/coding-standards:dev-master; fi
+    - travis_retry composer update $DEPENDENCIES
 
 script:
     - php vendor/bin/parallel-lint ./src ./tests
--- a/packages/http-smoke-testing/.travis.yml
+++ b/packages/http-smoke-testing/.travis.yml
@@ -17,13 +17,14 @@ matrix:
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
         -   php: 7.1
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
+        -   php: 7.1
+            env: CHECK_STANDARDS="true"
 
 install:
     - travis_retry composer update $DEPENDENCIES
+    - if [[ "$CHECK_STANDARDS" = "true" ]]; then composer require --dev shopsys/coding-standards:dev-master; fi
 
 script:
     - php vendor/bin/parallel-lint ./src ./tests
-    - php vendor/bin/php-cs-fixer fix --config=vendor/shopsys/coding-standards/build/phpcs-fixer.php_cs --dry-run --verbose --diff ./src ./tests
-    - php vendor/bin/phpcs --standard=vendor/shopsys/coding-standards/rulesetCS.xml --extensions=php --encoding=utf-8 --tab-width=4 -sp ./src ./tests
-    - php vendor/bin/phpmd "./src,./tests" text vendor/shopsys/coding-standards/rulesetMD.xml --extensions=php
     - php vendor/bin/phpunit tests
+    - if [[ "$CHECK_STANDARDS" = "true" ]]; then php vendor/bin/ecs check --verbose ./src ./tests; fi
--- a/packages/http-smoke-testing/.travis.yml
+++ b/packages/http-smoke-testing/.travis.yml
@@ -15,10 +15,10 @@ matrix:
         - { php: 7.3 }
 
 install:
-    - if [[ "$CHECK_STANDARDS" = "true" ]]; then composer require --dev shopsys/coding-standards:dev-master; fi
+    - if [[ "$CHECK_STANDARDS" == "true" ]]; then composer require --dev shopsys/coding-standards:dev-master; fi
     - travis_retry composer update $DEPENDENCIES
 
 script:
     - php vendor/bin/parallel-lint ./src ./tests
     - php vendor/bin/phpunit tests
-    - if [[ "$CHECK_STANDARDS" = "true" ]]; then php vendor/bin/ecs check --verbose ./src ./tests; fi
+    - if [[ "$CHECK_STANDARDS" == "true" ]]; then php vendor/bin/ecs check --verbose ./src ./tests; fi
--- a/packages/http-smoke-testing/.travis.yml
+++ b/packages/http-smoke-testing/.travis.yml
@@ -1,24 +1,18 @@
 language: php
 
-php:
-    - 5.6
-    - 7.0
-    - 7.1
-    - 7.2
-    - 7.3
-
 cache:
     directories:
         - ~/.composer/cache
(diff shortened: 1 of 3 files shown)


packages/http-smoke-testing/.travis.yml
```diff
@@ -17,13 +17,14 @@ matrix:
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
         -   php: 7.1
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
+        -   php: 7.1
+            env: CHECK_STANDARDS="true"
 
 install:
     - travis_retry composer update $DEPENDENCIES
+    - if [[ "$CHECK_STANDARDS" = "true" ]]; then composer require --dev shopsys/coding-standards:dev-master; fi
```
boris-brtan: why dev-master ????
all branches and tags that are locaten in repo shopsys/http-smoke-testing are distributed into packagist
https://docs.travis-ci.com/user/environment-variables#default-environment-variables

packages/http-smoke-testing/.travis.yml
```diff
@@ -17,13 +17,14 @@ matrix:
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
         -   php: 7.1
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
+        -   php: 7.1
```
boris-brtan: so there will be 3 tests for 7.1 now ?
so if the `php` section will be removed and distributed into `matrix include` it will be easier and more understandable

packages/http-smoke-testing/.travis.yml
```diff
@@ -17,13 +17,14 @@ matrix:
...
     - travis_retry composer update $DEPENDENCIES
+    - if [[ "$CHECK_STANDARDS" = "true" ]]; then composer require --dev shopsys/coding-standards:dev-master; fi
 
 script:
     - php vendor/bin/parallel-lint ./src ./tests
-    - php vendor/bin/php-cs-fixer fix --config=vendor/shopsys/coding-standards/build/phpcs-fixer.php_cs --dry-run --verbose --diff ./src ./tests
-    - php vendor/bin/phpcs --standard=vendor/shopsys/coding-standards/rulesetCS.xml --extensions=php --encoding=utf-8 --tab-width=4 -sp ./src ./tests
-    - php vendor/bin/phpmd "./src,./tests" text vendor/shopsys/coding-standards/rulesetMD.xml --extensions=php
     - php vendor/bin/phpunit tests
+    - if [[ "$CHECK_STANDARDS" = "true" ]]; then php vendor/bin/ecs check --verbose ./src ./tests; fi
```
boris-brtan: does `[[` allow one `=` shouldnt it be doubled ? `==`

boris-brtan: there seems to exist some solution also with https://docs.travis-ci.com/user/customizing-the-build#fast-finishing and https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail

PetrHeinz:
The `shopsys/coding-standards:dev-master` is uninstallable in the current state, see https://travis-ci.org/shopsys/http-smoke-testing/jobs/530802299

I'll try to come up with a way to fix that on Monday

packages/http-smoke-testing/easy-coding-standard.yml
```diff
@@ -0,0 +1,12 @@
+imports:
+    - { resource: '%vendor_dir%/shopsys/coding-standards/easy-coding-standard.yml', ignore_errors: true  }
+
+services:
+    Shopsys\CodingStandards\Sniffs\ConstantVisibilityRequiredSniff: ~
+
```
boris-brtan: empty row ?

packages/http-smoke-testing/.travis.yml
```diff
@@ -17,13 +17,14 @@ matrix:
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
         -   php: 7.1
             env: DEPENDENCIES="--prefer-lowest --prefer-stable"
+        -   php: 7.1
+            env: CHECK_STANDARDS="true"
```
boris-brtan:
```suggestion
            env: CHECK_STANDARDS=true
```

packages/http-smoke-testing/.travis.yml
boris-brtan: https://github.com/shopsys/http-smoke-testing/commit/ef64668575e51c350c9219cbfe0997ecf2adc207

boris-brtan: it seems also that some dependencies are missing.

PetrHeinz: all packages are dependent on `shopsys/coding-standards:dev-master` (except in tagged commits, see `composer.json`)

packages/http-smoke-testing/.travis.yml
PetrHeinz: Maybe, I'll try that :+1:

packages/http-smoke-testing/.travis.yml
PetrHeinz: The problems with symfony/finder installation were fixed by 6224ef1f0cab934c9eee11f347cfcf7c35e3f1fb, see https://travis-ci.org/shopsys/http-smoke-testing/builds/531622657

For some reason, when the symfony/finder v4 was already installed, Composer didn't downgrade it to v3 so that shopsys/coding-standards can be installed - it just rejected the package requirement. I just switched the order of `composer` calls - I required the package before dependencies were installed which ended up just fine.

packages/http-smoke-testing/.travis.yml
PetrHeinz: Done in 36a2c6b4b, I've inlined the objects because it was hard to read when in some jobs the `env` was missing. I think it's better now

packages/http-smoke-testing/.travis.yml
PetrHeinz: Both seem to work, using == in sake of consistency with monorepo-tools (72b124db4).

packages/http-smoke-testing/easy-coding-standard.yml
PetrHeinz: Yes, packages with few sniffs have empty lines between them. See migrations, google-cloud-bundle, form-types-bundle...

PetrHeinz:
Please take another look at it, @boris-brtan I think we can merge this :slightly_smiling_face: 

See https://travis-ci.org/shopsys/http-smoke-testing/builds/531625930 for the Travis package build (it's not a split package - I did the changes manually to save time, but the contents should be exactly the same)

boris-brtan:
seems ok to me

PR #1015, opened 2019-05-11
[framework] show information about project in debug toolbar

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| After I started work with more projects, I was asking my colleagues to tell me which version of framework project has. So I created toolbar with version and basic information about project. I also upgraded releaser which will update version automatically.
|New feature| Yes 
|Fixes issues| ... 

Screenshot 2019-05-11 at 23 50 44

### DIFF
--- a/packages/framework/src/Resources/views/Components/Collector/shopsysFramework.html.twig
+++ b/packages/framework/src/Resources/views/Components/Collector/shopsysFramework.html.twig
@@ -1,0 +1,46 @@
+{% extends '@WebProfiler/Profiler/layout.html.twig' %}
+
+{% block toolbar %}
+    {% set icon %}
+        <img width="24" height="24" alt="Shopsys" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAMAAADXqc3KAAAAAXNSR0IB2cksfwAAAAlwSFlzAAAuIwAALiMBeKU/dgAAAXFQTFRFAAAA////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////iIJ5mQAAAHt0Uk5TAAnGsIhGFAEO//zbdQwHe5O5+P7aNGlyE3Hh9mQEWezmYAVKvUUK8v2nJBK8zzflNaBssmP7HVXYA15qyxYPfKGij/mUAlbvX3jE87OL+u7wgFepF6qX05mlwI6SuEyuGXm0hy2twtIfnWvDPVj0Wm0aS8rZqH9dYZvJigjPOQAAAVJJREFUeJxtUmdXwkAQPCzgBkRNFKMRghEJgliioQgCirGLLfaGgl2w119v7hIkPJwPt5O5ncztvUOoCktDY1Oz1Ybq0AIAlN3RWrfhbGvvoIHp7KpRXd1OXNieXuD63FXdw3v7CREGfOAdrMh+jxgYChofoWEIj1T6RyEc/HOP0dS4ziQRJiymvElZjuAa5ZmY0W+L4zVihylcExyTnCZ6SkoLeDsDM+Qgs1mYUzCbX+AWsWcJlhFr9SPFsQKr2MOuQS6tzRNdT6INZhMh91YWtlXs2QF6l+SgPdjXVsXB8S4yeYbOkRx0AIck5+g4RBrZE9BzTuWcar6z1BlkJWzJx+DcrF8AJRGmFALeYlW/BL4g6Dx4Bdc3hq7egnwnVLqKItw/6LRU1v9v4LEM8tMzHt1dkswHib+8ytTb+8enRgVUg3yC0x6CiP6B+vXt+zELv8X1L+4rtB6IAAAAAElFTkSuQmCC" />
+        <span class="sf-toolbar-value">{{ collector.version }}</span>
+    {% endset %}
+    {% set text %}
+        <div class="sf-toolbar-info-group">
+            <div class="sf-toolbar-info-piece">
+                <b>Version</b>
+                <span>{{ collector.version }}</span>
+            </div>
+            <div class="sf-toolbar-info-piece">
+                <b>Current domain</b>
+                <span>{{ collector.currentDomainName }}</span>
+            </div>
+            <div class="sf-toolbar-info-piece">
+                <b>Locale</b>
+                <span class="sf-toolbar-status">{{ collector.currentDomainLocale }}</span>
+            </div>
+        </div>
+        <div class="sf-toolbar-info-group">
+            {% for domain in collector.domains %}
+                <div class="sf-toolbar-info-piece">
+                    <b>{{ domain.name }}</b>
+                    <span>
+                        {% if domain.id is same as(collector.currentDomainId) %}
+                            {{ domain.url }}
+                        {% else %}
+                            <a href="{{ domain.url }}">{{ domain.url }}</a>
+                        {% endif %}
+                    </span>
+                </div>
+            {% endfor %}
+        </div>
+        <div class="sf-toolbar-info-group">
+            <div class="sf-toolbar-info-piece">
+                <b>Resources</b>
+                <span><a href="https://github.com/shopsys/shopsys/blob/{{collector.docsVersion}}/docs/index.md" target="_blank" rel="help">Documentation</a></span>
+            </div>
+        </div>
+    {% endset %}
+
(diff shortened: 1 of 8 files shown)


PetrHeinz:
Nice idea! :+1: 

There are few things that need improvement before merging imo:

1) The logo in debug bar should be white, ideally in svg instead of png.
2) I would use the name "Shopsys Framework Knowledge Base" or "SSFW Knowledge Base" instead of "Framework Documentation".
3) Not sure about the "dev-master" version. There is no guarantee that it's dev-master, just that it's some unreleased version... Maybe it should be called "7.2.0-dev" (when the last released version was 7.1.0), similarly as in the [Symfony Kernel class](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Kernel.php). 
4) The release worker should not only change it to the released version during release-candidate phase, but also back to "dev-master" (or to "<next-version-number>-dev" as suggested above).
5) The classes `VersionUpgradeFileManipulator` and `UpdateUpgradeReleaseWorker` seem to be responsible for updating the upgrade notes. I think that this functionality should be separated into its own worker.
6) The `ShopsysCollector` (maybe rather `ShopsysFrameworkDataCollector`?) should fill the `data` property in the `collect` method, not in constructor. See eg. `RedisDataCollector`, `ConfigDataCollector` or `EventDataCollector`.
7) The `reset` method should set the `data` property to an empty array.

henzigo:
Hello, here is my answers to your comment:

1.  Logo is white now
2. "Shopsys Framework Knowledge Base" is too long for me and maybe "documentation" word is better to understand than "knowledge base". Secondly, we don't use SSFW often. I'm not changing it for now, but you can change it if you want.
3. I did it by you, so now its version 7.2.0-dev
4. Dev version will be now set after release (7.2.0 -> 7.3.0-dev)
5. I created new workers and file manipulator
6. updated
7. updated

I would like to see that function in next version, so please look at it and try to merge it :) @PetrHeinz

henzigo:
It looks awesome now 👍 
Screenshot 2019-05-13 at 23 21 31

henzigo:
Guys, it would be nice to have it in next minor release. What if programmer will have to write next version number on releaser request. However, the programmer can always adjust release version manually.

packages/framework/src/Resources/views/Components/Collector/shopsysFramework.html.twig
```diff
@@ -0,0 +1,46 @@
...
+                <b>Locale</b>
+                <span class="sf-toolbar-status">{{ collector.currentDomainLocale }}</span>
+            </div>
+        </div>
+        <div class="sf-toolbar-info-group">
+            {% for domain in collector.domains %}
+                <div class="sf-toolbar-info-piece">
+                    <b>{{ domain.name }}</b>
+                    <span>
+                        {% if domain.id == collector.currentDomainId %}
```
TomasLudvik:
```suggestion
                        {% if domain.id is same as(collector.currentDomainId) %}
```

packages/framework/src/Resources/views/Components/Collector/shopsysFramework.html.twig
```diff
@@ -0,0 +1,46 @@
...
+                            <a href="{{ domain.url }}">{{ domain.url }}</a>
+                        {% endif %}
+                    </span>
+                </div>
+            {% endfor %}
+        </div>
+        <div class="sf-toolbar-info-group">
+            <div class="sf-toolbar-info-piece">
+                <b>Resources</b>
+                <span><a href="https://github.com/shopsys/shopsys/blob/master/docs/index.md" target="_blank" rel="help">Framework Documentation</a></span>
```
TomasLudvik:
```suggestion
                <span><a href="https://github.com/shopsys/shopsys/blob/master/docs/index.md" target="_blank" rel="help">Documentation</a></span>
```

utils/releaser/src/FileManipulator/FrameworkVersionFileManipulator.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * @var string
+     */
+    public const FRAMEWORK_VERSION_FILE_PATH = '/packages/framework/src/ShopsysFrameworkBundle.php';
+
+    /**
+     * @var string
+     */
+    private const FRAMEWORK_VERSION_PATTEN = "/public const VERSION = '(.+)';/";
```
TomasLudvik:
```suggestion
    private const FRAMEWORK_VERSION_PATTERN = "/public const VERSION = '(.+)';/";
```

packages/framework/src/ShopsysFrameworkBundle.php
```diff
@@ -13,6 +13,11 @@
 
 class ShopsysFrameworkBundle extends Bundle
 {
+    /**
+     * @var string
+     */
+    public const VERSION = '7.2.0-dev';
```
TomasLudvik: The really cool would be to have link to documentation here. 

UpdateFrameworkKernelVersionReleaseWorker would update it to something like `https://github.com/shopsys/shopsys/blob/v7.2.0/docs/index.md` and `SetDevFrameworkKernelVersionReleaseWorker` back to `https://github.com/shopsys/shopsys/blob/master/docs/index.md`

grossmannmartin: Hi, I solved it differently to not need to update another thing in release worker and it's bounded to the version constant now.

packages/framework/src/Resources/views/Components/Collector/shopsysFramework.html.twig
grossmannmartin: Applied manually as I updated the file in the meantime

packages/framework/src/ShopsysFrameworkBundle.php
TomasLudvik: Your solution is much better than mine, good job!

simara-svatopluk:
The UI looks great!

I'm unfortunatelly unable to test the release process becasue step 9 `php phing translations-dump` fails.
Please fix it first (maybe rebase to current working branch may help).

PetrHeinz:
@simara-svatopluk Please, fix the typo in the annotation directly in vendor yourself. Currently we can either use the fix in #1169 or wait for new release of `studio-42/elfinder`.

simara-svatopluk:
I tried to release version `8.0.0-alpha1`. The RC worker works fine, turns the version to `8.0.0-alpha1`.

But the after release worker turns the version to `8.1.0-dev`, but that's wrong in this case. In this case I would expected `8.0.0-alpha2` or `8.0.0-beta` , `8.0.0` or `8.0.0-dev`, but none of them automatically... I expect to have a possibility to choose desired version.

So this "after release" behavior is wrong in my opinion and I suggest to change the worker to just a message `[Manually] Change the version in ShopsysFrameworkBundle::VERSION, suggesting "8.0.0-dev", "8.1.0-dev" and commit the change`

PetrHeinz:
@simara-svatopluk I think that after the release of `8.0.0-alpha1` you should be reverted to `8.0.0-dev`. But, considering the releaser is not prepared for working on different dev branches, I understand where the assumption that a new minor release is being prepared came from.

However, I like your idea of a manual release worker to avoid the problem altogether. Maybe with a suggestion of a new `MAJOR`, `MINOR` and `PATCH` dev version (or if the `PreReleaseSuffix` is present, suggest the dev version with same version number).

---

Seeing the output and commit messages I'd like to ask whether we really want to call it "shopsys framework kernel version" and if so, whether we shouldn't do it more consistently (see "shopsys kernel" / "framework kernel"). I'd rather just call it "SSFW version" or "Shopsys Framework version".

grossmannmartin:
@PetrHeinz hmm... You're right about inconsistency. 
the thing that bugs me most is a commit message (release worker make a commit)
Right now it's like `shopsys kernel version is now updated to version "8.0.0-dev"`, but changing it to Shopsys Framework change the meaning... (`shopsys framework is now updated....`)

PetrHeinz:
@grossmannmartin How about `ShopsysFrameworkBundle: version updated to "8.0.0-dev"`?

grossmannmartin:
@PetrHeinz Like.
I updated the names accordingly and should be unified (didn't touch the commits as they will be squashed)

utils/releaser/src/ReleaseWorker/AfterRelease/SetFrameworkBundleVersionToDevReleaseWorker.php
```diff
@@ -0,0 +1,112 @@
...
+            if (!$version->hasPreReleaseSuffix()) {
+                throw new RuntimeException(
+                    'Development version should be suffixed (with \'-dev\', \'-alpha1\', ...)'
+                );
+            }
+
+            return $version;
+        });
+
+        $developmentVersion = $this->symfonyStyle->askQuestion($question);
```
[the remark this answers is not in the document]
grossmannmartin: Applied manually as more work had to be done.

PR #979, opened 2019-05-02
[framework] read model introduction

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Better separated frontend from backend and preparation for usage of another source of frontend data
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| ... 
|Standards and tests pass| No
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

Issue #880, opened 2019-03-05
drop --verbose also for ecs in .travis.yml of all packages


boris-brtan:
### What is happening

verbosity on travis is not as nice as before because in newer version of ecs there was change of bahaviour. Fixes were done on phing targets but it was not done for other scripts like .travis.yml's
https://github.com/shopsys/shopsys/commit/f2eaf4944a9a18b6c00a35c3eed858b3492f32ab#diff-ae23427cd8e4dae17850f08f56308c3f

### Expected result

to see progressbar of ecs in all tools by default, instead of file path list.

boris-brtan:
another thing, i found out that google-cloud-bundle package is not tested for php 7.3
https://github.com/shopsys/shopsys/blob/master/packages/google-cloud-bundle/.travis.yml

PR #1018, opened 2019-05-13
read model for FE product lists

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Better separated frontend from backend and preparation for usage of another source of frontend data. Replaces #979. The functionality is implemented as a standalone package to separate the new layer from the existing domain model. The read model concept is used for frontend product lists only.
|New feature| Yes 
|Fixes issues| ... 

## TODO:
- [x] code itself ready for review
- [x] update docs (WIP in the last commit)
- [x] add tests
- [x] create a new repository for the read model package
- [x] edit the monorepo splitting script
- [x] add upgrade notes


packages/read-model/src/Image/ImageViewRepository.php
```diff
@@ -0,0 +1,56 @@
...
+     * @param array $entityIds
+     * @return array
+     */
+    public function getForEntityIds(string $entityClass, array $entityIds): array
+    {
+        $imagesIndexedByEntityIds = $this->imageFacade->getImagesOrNullsByEntitiesIndexedByEntityId($entityIds, $entityClass);
+
+        $imageViewsOrNullsIndexedByEntityIds = [];
+        foreach ($entityIds as $entityId) {
+            if (!isset($imagesIndexedByEntityIds[$entityId])) {
```
simara-svatopluk:
```suggestion
            $imageViewsOrNullsIndexedByEntityIds[$entityId] = $this->createView($imagesIndexedByEntityIds[$entityId]);
            ...
            protected function createView(?Image $image): ?ImageView {
                if ($image === null) {
                    return null;
                }
                return $this->imageViewFactory->createFromImage($image);
            }
```

the whole `if` is difficult to understand for me, so I suggest to change it to a single call of function and deal with details inside

packages/read-model/src/Product/Listed/ListedProductViewInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\ReadModelBundle\Product\Listed;
+
+/**
+ * @experimental
+ */
+interface ListedProductViewInterface
```
simara-svatopluk: Please remove the concept of interface for views as there is no added value.

```php
$topProducts = $this->listedProductsFacade->getTop();
foreach($topProducts as $product) {
    $product-> // nothing!
}
```

According to the interface I cannot do anything with the object. If you change it to actual `ListedProductView`, I'll know what can I get from it.

I understand you wanted to introduce interfaces for view objects because once the view is extended, the factory returns still the correct interface. But it also still returns the correct class anyway... So there is no added value.

It doesn't make sense generally to create an interface for read model objects, as same as for domain objects. The interface makes sense when you have a changeable behavior, eg. getting data.

packages/framework/src/Component/Image/ImageFacade.php
```diff
@@ -268,6 +268,28 @@ public function getImageUrl(DomainConfig $domainConfig, $imageOrEntity, $sizeNam
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig $domainConfig
+     * @param int $id
+     * @param string $extension
+     * @param string $entityName
+     * @param string|null $type
+     * @param string|null $sizeName
+     * @return string
+     */
+    public function getImageUrlNotUsingImageOrEntity(
```
simara-svatopluk: I know you already changed the method name, but it is still strange for me... I don't have any better unfortunately :-( And right now it is inconsistent with `ImageLocator`

maybe `getImageUrlFromParameters`? maybe the original `getImageUrlFromAttributes` was ok too

packages/read-model/LICENSE
```diff
@@ -0,0 +1,14 @@
+MIT License
```
simara-svatopluk: A new package... should it be automatically MIT? Maybe a question for the Product Owner

packages/read-model/easy-coding-standard.yml
```diff
@@ -0,0 +1,21 @@
+imports:
+    - { resource: '%vendor_dir%/shopsys/coding-standards/easy-coding-standard.yml', ignore_errors: true }
+
+services:
+    # this package is meant to be extensible using class inheritance, so we want to avoid private visibilities in the model namespace
```
simara-svatopluk: maybe this comment can be written in commit message

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -156,7 +156,7 @@ public function getNoimageHtml(array $attributes = [])
     /**
      * @return string
      */
-    private function getEmptyImageUrl()
+    public function getEmptyImageUrl(): string
```
simara-svatopluk: please explain the change in the commit message or extract the change into standalone commit

project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Product/productFlags.html.twig
```diff
@@ -0,0 +1,3 @@
+{% import '@ShopsysShop/Front/Inline/Product/productFlagsMacro.html.twig' as productFlags %}
```
simara-svatopluk: please explain this new macro in commit message or extract it into a standalone commit

vitek-rostislav:
Hi, I was discussing the design with @Miroslav-Stopka, and then with @grossmannmartin  and here comes one idea that could help to keep the whole framework architecture clearer - let's rename all the 3 `*Repository` classes in the package to `*Facade` :thinking:

simara-svatopluk:
@vitek-rostislav in fact I don't know what is the best suffix for a service that returns objects... (maybe no suffix, like `ListedProductViews`? *plular*).
I have no problem with `Facade`.

packages/read-model/tests/Unit/Image/ImageViewFactoryTest.php
```diff
@@ -0,0 +1,28 @@
...
+use PHPUnit\Framework\TestCase;
+use Shopsys\FrameworkBundle\Component\Image\Image;
+use Shopsys\ReadModelBundle\Image\ImageView;
+use Shopsys\ReadModelBundle\Image\ImageViewFactory;
+
+class ImageViewFactoryTest extends TestCase
+{
+    public function testCreateFromImage(): void
+    {
+        $imageMock = $this->createMock(Image::class);
```
simara-svatopluk: It's sad we have to use mocks in unit tests :-( ... but I understand the reason. Right now I don't have a good feeling about the `Image` entity but it's not your fault

packages/read-model/tests/Unit/Image/ImageViewRepositoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+        parent::setUp();
+
+        $this->imageFacadeMock = $this->createMock(ImageFacade::class);
+        $this->imageFacadeMock->method('getImagesOrNullsByEntitiesIndexedByEntityId')
+            ->willReturnCallback(function ($entityIds, string $entityClass) {
+                $images = [];
+
+                foreach ($entityIds as $entityId) {
+                    // id 800 represents not existing product
+                    if ($entityId === 800) {
```
simara-svatopluk:
```suggestion
                    if ($entityId === self::NOT_EXISTING_PRODUCT_ID) {
```
I suggest to name the ID and use it everywhere in the test. You can remove the comment then

packages/read-model/tests/Unit/Image/ImageViewRepositoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /** @var \Shopsys\FrameworkBundle\Component\Image\ImageFacade */
+    private $imageFacadeMock;
+
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->imageFacadeMock = $this->createMock(ImageFacade::class);
+        $this->imageFacadeMock->method('getImagesOrNullsByEntitiesIndexedByEntityId')
```
simara-svatopluk: This construction deserves an explanation... Please extract it into a function that will express what is the mock doing... maybe the whole mock creating could be in the function

packages/read-model/tests/Unit/Image/ImageViewRepositoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public function testGetForEntityIds(): void
+    {
+        $imageFactory = new ImageViewFactory();
+        $imageViewRepository = new ImageViewRepository($this->imageFacadeMock, $imageFactory);
+
+        $imageViews = $imageViewRepository->getForEntityIds('product', [1, 3, 5]);
+
+        $this->assertEquals(new ImageView(1, self::IMAGE_EXTENSION, 'product', null), $imageViews[1]);
```
simara-svatopluk: you can create the whole expected array...
```php
$expected = [
1 => new ImageView(1, self::IMAGE_EXTENSION, 'product', null),
3 => new ImageView(3, self::IMAGE_EXTENSION, 'product', null),
5 => new ImageView(5, self::IMAGE_EXTENSION, 'product', null),
];
$this->assertEquals($expected, $imaveViews);
```
... what do You think? What is more readable? (also this option checks number of elements);

packages/read-model/tests/Unit/Product/Action/ProductActionViewRepositoryTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+class ProductActionViewRepositoryTest extends TestCase
+{
+    public function testGetForProducts(): void
+    {
+        $productActionViewFactory = new ProductActionViewFactory();
+
+        $domainConfig = new DomainConfig(1, 'http://webserver:8080/', 'shopsys', 'en');
+
+        $domain = $this->createMock(Domain::class);
```
simara-svatopluk: I suggest to change the `$domain` to class property and setup it in `setup`

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
```diff
@@ -0,0 +1,162 @@
...
+                    $this->createFlagMock(4),
+                    $this->createFlagMock(3),
+                ])
+            ),
+        ];
+
+        /** @var \Shopsys\ReadModelBundle\Product\Listed\ListedProductView[] $listedProductViews */
+        $listedProductViews = $listedProductViewFactory->createFromProducts($productsMock);
+
+        $this->assertEquals(1, $listedProductViews[1]->getId());
```
simara-svatopluk: This asserting is unfortunately unreadable for me... I have a problem to understand what are we comparing. I hope you can improve it:
```php
$expected = [
1 => new ListedProductView(
 1,
 $productName1,
 'short description,
 ....
),
...
];
$this->assertEqueals($expected, $listedProductViews);
```

What do you think? Will be the new code more readable?

project-base/tests/ReadModelBundle/Functional/Image/ImageViewRepositoryTest.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    public function testGetForSingleEntityId(): void
+    {
+        $testedProductId = 1;
+
+        $imageViewRepository = $this->getContainer()->get(ImageViewRepository::class);
+
+        $imageViews = $imageViewRepository->getForEntityIds(Product::class, [$testedProductId]);
+
+        $this->assertCount(1, $imageViews);
```
simara-svatopluk: this would be definitelly more readable if you changed it to the `$expected = ....; $this->assertEquals($expected, $imageViews)` format

project-base/tests/ReadModelBundle/Functional/Image/ImageViewRepositoryTest.php
```diff
@@ -0,0 +1,54 @@
...
+use Shopsys\FrameworkBundle\Model\Product\Product;
+use Shopsys\ReadModelBundle\Image\ImageViewInterface;
+use Shopsys\ReadModelBundle\Image\ImageViewRepository;
+use Tests\ShopBundle\Test\FunctionalTestCase;
+
+class ImageViewRepositoryTest extends FunctionalTestCase
+{
+    public function testGetForSingleEntityId(): void
+    {
+        $testedProductId = 1;
```
simara-svatopluk: I'm not sure, but I think you should use

```php
$product = $this->getReference(ProductDataFixture::PRODUCT_PREFIX . '1');
$productId = $product->getId();
```

this deserves consultation with someone else @vitek-rostislav ?

project-base/tests/ReadModelBundle/Functional/Image/ImageViewRepositoryTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public function testGetForInvalidEntityId(): void
+    {
+        $testedProductId = 99999;
+
+        $imageViewRepository = $this->getContainer()->get(ImageViewRepository::class);
+
+        $imageViews = $imageViewRepository->getForEntityIds(Product::class, [$testedProductId]);
+
+        $this->assertCount(1, $imageViews);
```
simara-svatopluk:
```suggestion
        $this->assertSame([null], $imageViews);
```

project-base/tests/ReadModelBundle/Functional/Image/ImageViewRepositoryTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+        /** @var \Shopsys\ReadModelBundle\Image\ImageView[] $imageViews */
+        $this->assertSame('product', $imageViews[1]->getEntityName());
+        $this->assertSame(1, $imageViews[1]->getId());
+        $this->assertSame('jpg', $imageViews[1]->getExtension());
+    }
+
+    public function testGetForInvalidEntityId(): void
+    {
+        $testedProductId = 99999;
```
simara-svatopluk:
```suggestion
        $notExistingProductId = 99999;
```

project-base/tests/ReadModelBundle/Functional/Product/Action/ProductActionViewRepositoryTest.php
```diff
@@ -0,0 +1,35 @@
...
+            $this->getReference(ProductDataFixture::PRODUCT_PREFIX . '3'),
+        ];
+
+        $productActionViews = $productActionViewRepository->getForProducts($products);
+
+        $this->assertCount(3, $productActionViews);
+        $this->assertInstanceOf(ProductActionView::class, $productActionViews[1]);
+
+        /** @var \Shopsys\ReadModelBundle\Product\Action\ProductActionView[] $productActionViews */
+        $this->assertSame(1, $productActionViews[1]->getId());
```
simara-svatopluk: please use the `$expected... asset` format

project-base/tests/ReadModelBundle/Functional/Product/Listed/ListedProductViewRepositoryTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $this->assertEquals('Kabel HDMI A - HDMI A M/M 2m gold-plated connectors High Speed HD', $listedProductViews[24]->getName());
+        $this->assertEquals('Defender 2.0 SPK-480', $listedProductViews[13]->getName());
+    }
+
+    public function testGetPaginatedForBrand(): void
+    {
+        /** @var \Shopsys\ReadModelBundle\Product\Listed\ListedProductViewRepository $listedProductViewRepository */
+        $listedProductViewRepository = $this->getContainer()->get(ListedProductViewRepository::class);
+
+        $paginationResults = $listedProductViewRepository->getPaginatedForBrand(1, ProductListOrderingConfig::ORDER_BY_NAME_ASC, 1, 10);
```
simara-svatopluk: If you'll use the "persistent reference" for product, use it also for the brand

project-base/tests/ReadModelBundle/Functional/Product/Listed/ListedProductViewRepositoryTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $emptyFilterData = new ProductFilterData();
+
+        $paginationResults = $listedProductViewRepository->getPaginatedForFilteredSearch('kitty', $emptyFilterData, ProductListOrderingConfig::ORDER_BY_NAME_ASC, 1, 10);
+        $listedProductViews = $paginationResults->getResults();
+
+        $this->assertArrayHasKey(1, $listedProductViews);
+
+        $this->assertInstanceOf(ListedProductViewInterface::class, $listedProductViews[1]);
+
+        $this->assertEquals('22" Sencor SLE 22F46DM4 HELLO KITTY', $listedProductViews[1]->getName());
```
simara-svatopluk: maybe compare just an ID?

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
```diff
@@ -0,0 +1,62 @@
...
+use Tests\ShopBundle\Test\FunctionalTestCase;
+
+class ImageExtensionTest extends FunctionalTestCase
+{
+    public function testGetImageHtmlWithMockedImageFacade(): void
+    {
+        $frameworkBundleImageExtension = $this->getContainer()->get(FrameworkImageExtension::class);
+        $domain = $this->getContainer()->get(Domain::class);
+
+        $productId = 2;
```
simara-svatopluk: I suggest to inline these variables as I don't see a benefit in their reuse. The `$extension` variable is even confusing as I expected `ImageExtension` in it, not a string

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
```diff
@@ -0,0 +1,69 @@
...
+            new AdditionalImageData('(min-width: 1200px)', sprintf('http://webserver:8080/%s/additional_0_%d.%s', $entityName, $productId, $extension)),
+            new AdditionalImageData('(max-width: 480px)', sprintf('http://webserver:8080/%s/additional_1_%d.%s', $entityName, $productId, $extension)),
+        ]);
+
+        $imageView = new ImageView($productId, $extension, $entityName, null);
+
+        $readModelBundleImageExtension = new ImageExtension($frameworkBundleImageExtension, $imageFacade, $domain);
+        $html = $readModelBundleImageExtension->getImageHtml($imageView);
+
+        $this->assertXmlStringEqualsXmlFile(__DIR__ . '/Resources/picture.twig', $html);
```
simara-svatopluk:
```suggestion
        $this->assertXmlStringEqualsXmlFile(__DIR__ . '/Resources/picture-mocked-facade.twig', $html);
```

... to match the test case name

project-base/tests/ReadModelBundle/Functional/Twig/Resources/picture.twig
```diff
@@ -0,0 +1,5 @@
+<picture>
+    <source media="(min-width: 1200px)" srcset="http://webserver:8080/product/additional_0_2.jpg" />
+    <source media="(max-width: 480px)" srcset="http://webserver:8080/product/additional_1_2.jpg" />
+    <img  alt="" title="" src="http://webserver:8080/product/2.jpg" class="image-product" itemprop="image" />
```
simara-svatopluk:
```suggestion
    <img alt="" title="" src="http://webserver:8080/product/2.jpg" class="image-product" itemprop="image" />
```

packages/read-model/LICENSE
grossmannmartin: After a talk with @LukasHeinz, we go for Shopsys license instead.

packages/read-model/easy-coding-standard.yml
grossmannmartin: Like this, it's consistent with other packages. I prefer the comment here to be clear on the first sight.

packages/read-model/src/Product/Listed/ListedProductView.php
```diff
@@ -0,0 +1,159 @@
...
+        int $id,
+        string $name,
+        ?string $shortDescription,
+        string $availability,
+        ProductPrice $sellingPrice,
+        array $flagIds,
+        ProductActionViewInterface $action,
+        ?ImageViewInterface $image
+    ) {
+        foreach ($flagIds as $flagId) {
```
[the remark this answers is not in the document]
grossmannmartin: extracted into validation method

packages/read-model/src/Product/Listed/ListedProductViewInterface.php
grossmannmartin: The main idea of these interfaces is not to model changeable behavior but provide type information about objects instead. It's known as a Marker Interface pattern. 

> According to the interface I cannot do anything with the object

To me, it's not a problem here because objects are directly used in twig templates only (and tests, but I'm ok with it there).


And the added value to me is honesty about type while extendind. Consider `createFromProduct` method from `ListedProductViewFactory` class. This method should return View object and is now defined like:
```php
protected function createFromProduct(Product $product, ?ImageViewInterface $imageView, ProductActionViewInterface $productActionView): ListedProductViewInterface
```

In a way how it's implemented, this method is first to be overridden if I need another field in the view object.

Let's presume, I declare this method to return the `ListedProductView` object directly. Like this:
```php
protected function createFromProduct(...): ReadModelBundle\ListedProductView
```
In my project, I cannot override this method and declare a different return type
```php
namespace Shopsys\ShopBundle\Model\Product\ViewFactory;

// Invalid: Return type declaration must be compatible with Shopsys\ReadModelBundle\Product\ListedProductViewFactory
protected function createFromProduct(...): ShopBundle\ListedProductView 
```

So I need to return parent class and then I lie about the return type – if I have the brand in my extended `ListedProductView`, returned object here looks like there is no brand at all, so in my point of view, this solves nothing.

And I'm not a fan of a different type in annotation and in typehint.

So Interface still seems to me like the cleanest possible solution.

simara-svatopluk: > So I need to return parent class and then I lie about the return type

This is not true. `B extends A` means `B IS A`, so if you declare `return A` and you return `B`, you don't lie.

> And the added value to me is honesty about type while extendind.

I don't see the added value. With both - interface and parent class declaration - you are honest and you don't lie.

> The main idea of these interfaces is not to model changeable behavior but provide type information about objects instead.

Type information is also provided by class declaration.


If you have an interface
```php
protected function createFromProduct(...): ListedProductViewInterface
```
You know almost nothing about returned object, only that it is somehow marked. What is this information good for?

If you have an interface with class
```php
protected function createFromProduct(...): ListedProductView
```
You know at least something about returned object. Definitely more than with the interface

packages/read-model/src/Image/ImageViewRepository.php
grossmannmartin: I had to rewrite it a little bit more because I realized `getImagesOrNullsByEntitiesIndexedByEntityId` does not do what claims to do. It's not returning null values for not existing ids. 
Please check it out.

packages/read-model/src/Product/Listed/ListedProductViewFacade.php
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Shopsys\ReadModelBundle\Product\Listed;
+
+use Shopsys\FrameworkBundle\Component\Paginator\PaginationResult;
+use Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData;
+
+/**
+ * @experimental
+ */
+class ListedProductViewFacade implements ListedProductViewFacadeInterface
```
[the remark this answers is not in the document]
grossmannmartin: Yes, this is related to the discussion about `Repository` suffix. I renamed all the repositories to `*Facade` and it should be less confusing and one class less it is.

packages/read-model/src/Product/Listed/ListedProductViewRepository.php
```diff
@@ -0,0 +1,177 @@
...
+        $this->topProductFacade = $topProductFacade;
+        $this->productOnCurrentDomainFacade = $productOnCurrentDomainFacade;
+        $this->listedProductViewFactory = $listedProductViewFactory;
+    }
+
+    /**
+     * @param int|null $limit
+     * @return \Shopsys\ReadModelBundle\Product\Listed\ListedProductViewInterface[]
+     */
+    public function getForTop(?int $limit = null): array
```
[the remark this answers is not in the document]
grossmannmartin: It's now only in Facade with names you suggest.

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
```diff
@@ -0,0 +1,51 @@
...
+use Shopsys\FrameworkBundle\Component\Paginator\PaginationResult;
+use Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData;
+
+interface ListedProductViewFacadeInterface
+{
+    /**
+     * @param int|null $limit Returns all products when "null" is provided
+     * @return \Shopsys\ReadModelBundle\Product\Listed\ListedProductViewInterface[]
+     */
+    public function getTop(?int $limit = null): array;
```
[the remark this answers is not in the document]
grossmannmartin: If I remember correctly, this was discussed with @PetrHeinz on PIA and the decision was to have the limit here even if it's not used to stay consistent with accessories and provide a way to render a different number of top products if needed. 
I personally would be glad if limit stays.
What do you think @vitek-rostislav?

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @param int $productId
+     * @param int|null $limit Returns all products when "null" is provided
+     * @return \Shopsys\ReadModelBundle\Product\Listed\ListedProductView[]
+     */
+    public function getAccessories(int $productId, ?int $limit = null): array;
```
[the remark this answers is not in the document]
grossmannmartin: I agree with you. 
@vitek-rostislav, do you know why this was decided like this?

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @param string|null $searchText
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData $filterData
+     * @param string $orderingModeId {@see Shopsys\FrameworkBundle\Model\Product\Listing\ProductListOrderingConfig}
+     * @param int $page Page number (starting with 1)
+     * @param int $limit Number of products per page (must be greater than 0)
+     * @return \Shopsys\FrameworkBundle\Component\Paginator\PaginationResult
+     */
+    public function getFilteredPaginatedForSearch(?string $searchText, ProductFilterData $filterData, string $orderingModeId, int $page, int $limit): PaginationResult;
```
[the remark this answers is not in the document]
grossmannmartin: I believe the main reason is that in controller we get searchtext by `$request->query->get(self::SEARCH_TEXT_PARAMETER);` and this method returns null if there is no such parameter in query string.

Currently is used `ProductOnCurrentDomainFacade::getPaginatedProductsForSearch()` method which defines `$searchText` in a same way.

I would happily remove `?` but I fear the potential bugs in projects. Need to check for null value from search parameter in controller can be easily forgotten and `/search` route (without `q` parameter) then raise 500 error.

packages/read-model/src/Product/Listed/ListedProductViewRepository.php
```diff
@@ -0,0 +1,177 @@
...
+    {
+        $topProducts = $this->topProductFacade->getAllOfferedProducts(
```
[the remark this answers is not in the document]
grossmannmartin: Considering the whole `TopProduct` layer in framework, not much. 
To start limiting results by default would need other changes in `TopProductFacade` and `TopProductRepository`. 
This is the road I don't want to go right now as we don't have any feedback of this been a problem.

packages/read-model/src/Product/Listed/ListedProductViewFactory.php
```diff
@@ -0,0 +1,118 @@
...
+use Shopsys\FrameworkBundle\Model\Product\ProductCachedAttributesFacade;
+use Shopsys\ReadModelBundle\Image\ImageViewInterface;
+use Shopsys\ReadModelBundle\Image\ImageViewRepository;
+use Shopsys\ReadModelBundle\Product\Action\ProductActionViewInterface;
+use Shopsys\ReadModelBundle\Product\Action\ProductActionViewRepository;
+
+/**
+ * @experimental
+ */
+class ListedProductViewFactory
```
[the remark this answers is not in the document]
grossmannmartin: Awesome!!! Thanks, it's much nicer!

packages/framework/src/Twig/ImageExtension.php
grossmannmartin: :+1: I will do the change on rebase to not mess with history right now

packages/read-model/src/Twig/ImageExtension.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $this->baseImageExtension->preventDefault($attributes);
+
+        try {
+            if ($imageView instanceof ImageViewInterface) {
+                $attributes['src'] = $this->imageFacade->getImageUrlNotUsingImageOrEntity(
+                    $this->domain->getCurrentDomainConfig(),
+                    $imageView->getId(),
+                    $imageView->getExtension(),
+                    $entityName = $imageView->getEntityName(),
```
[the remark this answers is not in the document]
grossmannmartin: I completely rewrote this extension. Please check it out.

packages/read-model/src/Twig/ImageExtension.php
```diff
@@ -0,0 +1,93 @@
...
+                    $attributes['size']
+                );
+                $additionalImagesData = $this->imageFacade->getAdditionalImagesDataNotUsingImageOrEntity(
+                    $this->domain->getCurrentDomainConfig(),
+                    $imageView->getId(),
+                    $imageView->getExtension(),
+                    $entityName = $imageView->getEntityName(),
+                    $imageView->getType(),
+                    $attributes['size']
+                );
```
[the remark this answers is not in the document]
grossmannmartin: I completely rewrote this extension. Please check it out.

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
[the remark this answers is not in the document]
vitek-rostislav: I would say that the method is more flexible with the `$limit` parameter so I am voting for keeping it this way :slightly_smiling_face:

project-base/src/Shopsys/ShopBundle/Controller/Front/CartController.php
```diff
@@ -171,6 +164,23 @@ public function addProductFormAction(Product $product, $type = 'normal')
         ]);
     }
 
+    /**
+     * @param \Shopsys\ReadModelBundle\Product\Action\ProductActionView $productActionView
```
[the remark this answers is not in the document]
grossmannmartin: Okay, specifically in this controller I agree with you, the interface is not ideal, because of line 173.

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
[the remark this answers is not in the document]
vitek-rostislav: ok, after I read the next comment, I like the solution with two methods :wink:

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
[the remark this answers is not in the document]
vitek-rostislav: TBH, I have no idea about the reasons behind this decision. I am absolutely ok with rewriting this into two methods

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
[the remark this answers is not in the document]
vitek-rostislav: I think we could simplify the method if we add a default value for the query parameter in `ProductController::searchAction`
```diff
- $searchText = $request->query->get(self::SEARCH_TEXT_PARAMETER);
+ $searchText = $request->query->get(self::SEARCH_TEXT_PARAMETER, '');
```

After that, even `ProductOnCurrentDomainFacade` could be refactored eventually (probably not in this PR)

packages/read-model/src/Image/ImageViewRepository.php
vitek-rostislav: @grossmannmartin oh, my apologies, I have rewritten the function but not its name :man_facepalming:

packages/framework/src/Component/Image/ImageFacade.php
vitek-rostislav: oh, I forget to rename it in `ImageLocator`...ok, renamed back to `...FromAttributes`

packages/read-model/.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+
+before_install:
+    - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+
+install:
+    - composer install
+
+script:
+    - php vendor/bin/parallel-lint ./src ./tests
+    - php vendor/bin/ecs check --verbose ./src ./tests
```
boris-brtan: hello we have created [this issue #880](https://github.com/shopsys/shopsys/issues/880) for removing --verbose from travis. I am not sure whether it make sense to create new scripts with that parameter set.

packages/read-model/.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+
+cache:
+    directories:
+        - ~/.composer/cache
+
+before_install:
+    - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+
+install:
+    - composer install
```
boris-brtan: in packages.json i see that it is dependent on shopsys/framework, please do not forget to set COMPOSER_MEMORY_LIMIT=-1 for the travis job.

packages/read-model/src/Product/Listed/ListedProductView.php
```diff
@@ -0,0 +1,167 @@
...
+        int $id,
+        string $name,
+        ?string $shortDescription,
+        string $availability,
+        ProductPrice $sellingPrice,
+        array $flagIds,
+        ProductActionViewInterface $action,
+        ?ImageViewInterface $image
+    ) {
+        $this->validateFlagIds($flagIds);
```
simara-svatopluk: I just saw even nicer solution!

```suggestion
        \Webmozart\Assert\Assert::allInteger($flagIds);
```

packages/read-model/src/Twig/FlagsExtension.php
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * @return array
+     */
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction(
+                'renderFlagsByIds',
+                function (array $flagIds, string $classAddition = '') {
+                    return $this->templating->render('@ShopsysShop/Front/Inline/Product/productFlags.html.twig', [
```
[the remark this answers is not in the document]
grossmannmartin: I moved this extension to project-base

packages/read-model/src/Twig/FlagsExtension.php
```diff
@@ -0,0 +1,79 @@
...
+                ['is_safe' => ['html']]
+            ),
+        ];
+    }
+
+    /**
+     * @param int[] $flagIds
+     * @return \Shopsys\FrameworkBundle\Model\Product\Flag\Flag[]
+     */
+    protected function getFlagsByIds(array $flagIds): array
```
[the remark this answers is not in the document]
grossmannmartin: Done 👍

packages/read-model/tests/Unit/Image/ImageViewRepositoryTest.php
grossmannmartin: Nicer, thanks :)

packages/read-model/tests/Unit/Image/ImageViewRepositoryTest.php
grossmannmartin: I hope it's better now

packages/read-model/tests/Unit/Product/Action/ProductActionViewRepositoryTest.php
grossmannmartin: I extracted the domain mock creation into the separate method as it is not needed to be class property.

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
grossmannmartin: I fixed this and I hope its better now :)

project-base/tests/ReadModelBundle/Functional/Image/ImageViewRepositoryTest.php
vitek-rostislav: Hi, I think it is not necessary here - we do not need the whole entity, just _an_ ID. Maybe constant would be more suitable, however, I am ok with the current state

grossmannmartin: It's in the constant now

project-base/tests/ReadModelBundle/Functional/Image/ImageViewRepositoryTest.php
grossmannmartin: it's in a constant now

project-base/tests/ReadModelBundle/Functional/Image/ImageViewRepositoryTest.php
grossmannmartin: done in a similar way

project-base/tests/ReadModelBundle/Functional/Product/Listed/ListedProductViewRepositoryTest.php
grossmannmartin: I made this hardcoded values into variables for better understanding.

project-base/tests/ReadModelBundle/Functional/Product/Listed/ListedProductViewRepositoryTest.php
grossmannmartin: To me, the name is more convincing :)

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
grossmannmartin: Test is now completely different, please check it out.

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
grossmannmartin: Test is now completely different, please check it out.

packages/read-model/src/Product/Listed/ListedProductView.php
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * @return \Shopsys\ReadModelBundle\Product\Action\ProductActionView
+     */
+    public function getAction(): ProductActionView
+    {
+        return $this->action;
+    }
+
+    /**
+     * @param array $flagIds
```
simara-svatopluk:
```suggestion
     * @param int[] $flagIds
```

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * @dataProvider getProductsData
+     * @param int $id
+     * @param string $productName
+     * @param string $shortDescription
+     * @param string $availabilityName
+     * @param int $priceAmount
+     * @param \Shopsys\ReadModelBundle\Image\ImageView $imageView
+     * @param \Shopsys\ReadModelBundle\Product\Action\ProductActionView $productActionView
+     * @param array|\Doctrine\Common\Collections\ArrayCollection $flags
```
simara-svatopluk:
```suggestion
     * @param Flag[] $flags
```

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
```diff
@@ -0,0 +1,156 @@
...
+
+        $this->assertEquals($expected, $listedProductView);
+    }
+
+    /**
+     * @param int $id
+     * @param string $name
+     * @param string $shortDescription
+     * @param string $availabilityName
+     * @param array|\Doctrine\Common\Collections\ArrayCollection $flags
```
simara-svatopluk:
```suggestion
     * @param Flag[] $flags
```

packages/read-model/src/Image/ImageViewRepository.php
simara-svatopluk: seems great :+1:

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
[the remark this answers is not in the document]
grossmannmartin: Done.

packages/read-model/src/Product/Listed/ListedProductViewFacadeInterface.php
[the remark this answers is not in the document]
grossmannmartin: Done.

packages/read-model/src/Product/Listed/ListedProductViewInterface.php
grossmannmartin: After a long discussion with @vitek-rostislav, with a heavy heart, I removed the Interfaces. 

I still believe it's a good concept but I understand your arguments.

simara-svatopluk: Thanks!
The interface topic deserves deeper explanation... I'd like to meet with you to not just explain but also to show when your approach makes sense... if you agree :-)

simara-svatopluk: I just checked all references of `ViewInterface` are gone :heavy_check_mark:

packages/read-model/src/Twig/FlagsExtension.php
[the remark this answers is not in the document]
simara-svatopluk: The new `FlagsExtension` is awesome. It is just clean, polished, pure... Great work!

For a while I had conversation in my head
- Why should be the `FlagsProvider` in the Read model? It should be in FW
- It just provides data for frontend!
- That makes totally sense!

packages/read-model/src/Flag/FlagsProvider.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param int[] $flagIds
+     * @return \Shopsys\FrameworkBundle\Model\Product\Flag\Flag[]
+     */
+    public function getFlagsByIds(array $flagIds): array
+    {
+        if ($this->allFlags === null) {
+            $this->allFlags = $this->flagFacade->getAll();
+        }
+
+        $flags = [];
```
simara-svatopluk: The provider itself is awesome, great job!

I was staring on the `foreach-if` for a while, and then I realized we just filter flags by something. So I've rewritten this statement to
```php
return array_filter($this->allFlags, function (Flag $flag) use ($flagIds) {
    return in_array($flag->getId(), $flagIds, true);
});
```

And then I realized how horrible filter/mapping/reduce operations looks in the PHP so... It's up to you - keep the original implementation and if you like the `filter`, change it :-)

simara-svatopluk: btw I didn't test it

packages/read-model/src/Twig/ImageExtension.php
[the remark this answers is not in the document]
simara-svatopluk: wow... so short! :heart:

packages/read-model/src/Twig/ImageExtension.php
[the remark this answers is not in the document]
simara-svatopluk: You used the inheritance exactly as it should be used - specialization - handling a new use-case (new behavior) without changing interface. :tophat:

packages/framework/src/Component/Image/ImageFacade.php
simara-svatopluk: thanks and sorry for changing opinions during two days

packages/read-model/tests/Unit/Image/ImageViewRepositoryTest.php
simara-svatopluk: nice

packages/read-model/src/Image/ImageView.php
```diff
@@ -0,0 +1,79 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\ReadModelBundle\Image;
+
+/**
+ * @experimental
+ *
+ * Class representing images in FE templates (to avoid usage of Doctrine entities a hence achieve performance gain)
+ */
+class ImageView implements ImageViewInterface
```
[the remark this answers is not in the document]
simara-svatopluk: Now I understand the `ImageView` concept. The `Image` concept in the SSFW means image metadata that can be used to construct URL or to work with image files. The `ImageView` reflects the `Image` concept - contains metadata for URL.

If the URL was in the `ImageView` concept, it won't match how the SSFW looks on images. We want to eliminate different mental models at least for images, so wee keep the `ImageView` as it is

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
simara-svatopluk: The data provider is great. Please adjust just the formatting of long method prototype and long object construction to...
```php
public function testCreateFromProduct(
    int $id,
    string $productName,
    ...
)

$expected = new ListedProductView($id,
    $productName,
    $shortDescription,
   ...
);
```

project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Product/productFlags.html.twig
grossmannmartin: 👍 I will do the change on rebase to not mess with history right now

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
simara-svatopluk: ok, I'll check it

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+        $imageFacade = $this->getContainer()->get(ImageFacade::class);
+
+        $productId = 1;
+        $entityName = 'product';
+        $fileExtension = 'jpg';
+
+        $imageView = new ImageView($productId, $fileExtension, $entityName, null);
+
+        $readModelBundleImageExtension = new ImageExtension('', $domain, $imageLocator, $imageFacade, $templating);
```
simara-svatopluk: At this point you can even get the `ImageExtension` from container, right? (maybe not because of the prefix, but I hope so). Same applies for following tests.

If you cannot receive the extension from container, please create method that will create the the object (and remove duplicits)

packages/read-model/.travis.yml
simara-svatopluk: @vitek-rostislav please fix it. By the thumbsup I assumed it is fixed

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
grossmannmartin: I changed it to `\Shopsys\FrameworkBundle\Model\Product\Flag\Flag[]|\Doctrine\Common\Collections\ArrayCollection` because it's important to me to test ArrayCollection behavior and like this, it explains why there is no typehint for flags in method argument.

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
grossmannmartin: I changed it to \Shopsys\FrameworkBundle\Model\Product\Flag\Flag[]|\Doctrine\Common\Collections\ArrayCollection because it's important to me to test ArrayCollection behavior and like this, it explains why there is no typehint for flags in method argument.

packages/read-model/src/Flag/FlagsProvider.php
grossmannmartin: I will keep the original implementation as I believe it is easier to read.

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
grossmannmartin: I created a method and simplified test cases with it.

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,129 @@
+# Extending Product List
```
simara-svatopluk: the cookbook is not linked anywhere... Please add it to cookbooks and maybe to article about extending entity

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,129 @@
...
+
+Frontend product lists leverage [the read model concept](/docs/model/introduction-to-read-model.md), i.e. special view objects are used instead of common Doctrine entities in the Twig templates.
```
simara-svatopluk: I'm missing sort of motivation... why should I read this? Something like "You need to add more information to product lists but adding them into entities doesn't work? That is because..."

The formulation is strange, but I would like to see a motivation, what problem this cookbook solves.

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,129 @@
...
+}
+```
+
+You need to register your new class as an alias for the one from the bundle in `services.yml`:
+
+```yml
+Shopsys\ReadModelBundle\Product\Listed\ListedProductViewFactory: '@Shopsys\ShopBundle\Model\Product\View\ListedProductViewFactory'
+```
+
+*Note: If you want to avoid ["the n+1 problem"](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping) here,
```
simara-svatopluk: We are suggesting to take a look at `Facade` but we are extending `Factory` in this article.

It sort of makes sense, but if you want to keep here this note, you have to instruct the reader what to do next. Or alternatively you can ass a new section **4** that deals with getting data in an optimized way.

packages/read-model/src/Product/Listed/ListedProductViewFactory.php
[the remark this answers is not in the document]
simara-svatopluk: Good implementation :+1:

packages/read-model/src/Product/Listed/ListedProductView.php
simara-svatopluk: @grossmannmartin what do you think about this solution?

packages/read-model/tests/Unit/Product/Listed/ListedProductViewFactoryTest.php
simara-svatopluk: :+1:

docs/upgrade/read-model-for-product-lists.md
```diff
@@ -0,0 +1,99 @@
...
+
+            return $bundles;
+        }
+
+        // ...
+    }
+    ```
+    - Note: the bundle needs to be registered after `ShopsysFrameworkBundle` as it overwrites its `image` Twig function
+- in all frontend controllers where you are listing products, use the new `Shopsys\ReadModelBundle\Product\Listed\ListedProductViewFacadeInterface`:
+    - in `ProductController::detailAction`:
```
simara-svatopluk: please add instruction that "adds `listedProductsFacade` property and modify constructor"... to all Controllers

docs/upgrade/read-model-for-product-lists.md
```diff
@@ -0,0 +1,99 @@
...
+        ```
+    - in `HomepageController::indexAction`:
+        ```diff
+        - $topProducts = $this->topProductFacade->getAllOfferedProducts(
+        -     $this->domain->getId(),
+        -     $this->currentCustomer->getPricingGroup()
+        - );
+        + $topProducts = $this->listedProductsFacade->getAllTop();
+        ```
+- edit your `productListMacro.html.twig` so it now works with instances of `ListedProductView` instead of `Product` entities
```
simara-svatopluk: I don't understand... what am I supposed to do? It seems I have to edit `productListMacro.html.twig`, please write down how.... the diff format is great!

docs/upgrade/read-model-for-product-lists.md
```diff
@@ -0,0 +1,99 @@
...
+    - to render product flags by their ids, you need to implement new [`FlagsExtension`](/project-base/src/Shopsys/ShopBundle/Twig/FlagsExtension.php) with `renderFlagsByIds` function and add a new [`productFlags.html.twig`](/project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Product/productFlags.html.twig) template
```
simara-svatopluk: I don't understand the instruction.. Does it mean something like "Download this class here" and "download this twig template here"?

docs/upgrade/read-model-for-product-lists.md
```diff
@@ -0,0 +1,99 @@
...
+    - to render "add to cart" form, add `CartController::productActionAction` and use it instead of `addProductFormAction`
```
simara-svatopluk: To render... add... what? And use it where?

If I have to implement something in the `CartController`, please provide a diff, it is super helpful.
And please change the "use it" instruction to "change this there, and change this there"

packages/read-model/src/Product/Listed/ListedProductView.php
grossmannmartin: I like the idea of "centralized assertion" (no need to invent it everywhere), but I wasn't sure if we really need this. Discussion with @PetrHeinz, you and some deep internet digging leave me with a very good feeling about it.
So I added the package to composer (as we didn't had it directly) and apply this suggestion.

grossmannmartin: I will reorder commits then so it will make sense :) I didn't want to do "used webmozart" commit

simara-svatopluk: :+1:

docs/model/introduction-to-read-model.md
```diff
@@ -0,0 +1,23 @@
+# Introduction to Read Model
+
+Read model is an extra layer that separates templates and [the application model](/docs/model/introduction-to-model-architecture.md).
+The read model is taken from [CQRS pattern](https://martinfowler.com/bliki/CQRS.html) where the main idea is usage of different objects for reading and writing data.
+
+<!--- TODO Replace with "nicer" image once the concept is approved --->
+![model architecture schema](./img/model-architecture-with-read-model.png 'Read model in Shopsys Framework architecture')
```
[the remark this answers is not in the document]
simara-svatopluk:
```
[Controller, CLI, REST]
 |                 |
 v                 v
ViewFacade - ---> View
```

docs/cookbook/extending-product-list.md
vitek-rostislav: in such a case, I would rather remove the note completely, as there are too many scenarios and solutions...it seems out of the scope to me

docs/upgrade/read-model-for-product-lists.md
vitek-rostislav: yes, rewritten :wink:

packages/read-model/.travis.yml
simara-svatopluk: @vitek-rostislav ping

packages/read-model/README.md
```diff
@@ -0,0 +1,69 @@
+# Shopsys Read Model
+
+[![Build Status](https://travis-ci.org/shopsys/read-model.svg?branch=master)](https://travis-ci.org/shopsys/read-model)
+[![Downloads](https://img.shields.io/packagist/dt/shopsys/read-model.svg)](https://packagist.org/packages/shopsys/read-model)
+
+This bundle for [Shopsys Framework](https://www.shopsys-framework.com) separates templates from model using [read model concept](https://github.com/shopsys/shopsys/blob/master/docs/model/introduction-to-read-model.md).
```
simara-svatopluk:
```suggestion
This bundle for [Shopsys Framework](https://www.shopsys-framework.com) separates templates from model using [Read Model](/docs/model/introduction-to-read-model.md) concept.
```

simara-svatopluk: @vitek-rostislav and the link https://www.shopsys-framework.com still ends up with error. Please change it to http://www.shopsys-framework.com OR https://www.shopsys.com

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,127 @@
...
+```php
+declare(strict_types=1);
+
+namespace Shopsys\ShopBundle\Model\Product\View;
+
+use Shopsys\FrameworkBundle\Model\Product\Product;
+use Shopsys\ReadModelBundle\Image\ImageView;
+use Shopsys\ReadModelBundle\Product\Action\ProductActionView;
+use Shopsys\ReadModelBundle\Product\Listed\ListedProductViewFactory as BaseListedProductViewFactory;
+use Shopsys\ReadModelBundle\Product\Listed\ListedProductView;
```
TomasLudvik:
```suggestion
use Shopsys\ReadModelBundle\Product\Listed\ListedProductView as BaseListedProductView;
```

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,127 @@
...
+
+class ListedProductViewFactory extends BaseListedProductViewFactory
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param \Shopsys\ReadModelBundle\Image\ImageView|null $imageView
+     * @param \Shopsys\ReadModelBundle\Product\Action\ProductActionView $productActionView
+     * @return \Shopsys\ReadModelBundle\Product\Listed\ListedProductView
+     */
+    protected function createFromProduct(Product $product, ?ImageView $imageView, ProductActionView $productActionView): ListedProductView
```
TomasLudvik:
```suggestion
    protected function createFromProduct(Product $product, ?ImageView $imageView, ProductActionView $productActionView): BaseListedProductView
```

TomasLudvik:
```suggestion
    public function createFromProduct(Product $product, ?ImageView $imageView, ProductActionView $productActionView): ListedProductView
```

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,127 @@
+# Extending Product List
+
+If you want add more information to product lists, you need to do a little bit more than [extending an entity](/docs/cookbook/adding-new-attribute-to-an-entity.md).
+That is because the frontend product lists leverage [the read model concept](/docs/model/introduction-to-read-model.md), i.e. special view objects are used instead of common Doctrine entities in the Twig templates.
+
+This cookbook describes how to extend the frontend product lists with step by step instructions.
```
TomasLudvik:
```suggestion
This cookbook describes how to extend the frontend product list with step by step instructions.
```

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,127 @@
...
+     * @param \Shopsys\ReadModelBundle\Image\ImageView|null $imageView
+     * @param \Shopsys\ReadModelBundle\Product\Action\ProductActionView $productActionView
+     * @return \Shopsys\ReadModelBundle\Product\Listed\ListedProductView
+     */
+    protected function createFromProduct(Product $product, ?ImageView $imageView, ProductActionView $productActionView): ListedProductView
+    {
+        return new ListedProductView(
+            $product->getId(),
+            $product->getName(),
+            $product->getBrand() ? $product->getBrand()->getName() : null,
```
TomasLudvik: Newly added parameter is third even it is nullable, why?

docs/upgrade/read-model-for-product-lists.md
```diff
@@ -0,0 +1,140 @@
...
+            // ...
+
+            return $bundles;
+        }
+
+        // ...
+    }
+    ```
+    - Note: the bundle needs to be registered after `ShopsysFrameworkBundle` as it overwrites its `image` Twig function
+- in all frontend controllers where you are listing products, add new `$listedProductViewFacade` property and modify the constructor to inject instance of `Shopsys\ReadModelBundle\Product\Listed\ListedProductViewFacadeInterface` to the property:
```
simara-svatopluk:
```suggestion
- in all frontend controllers where you are listing products, add new `$listedProductViewFacade` property and modify the constructor to inject instance of `Shopsys\ReadModelBundle\Product\Listed\ListedProductViewFacadeInterface` to the property. Eg.
```diff
class ProductController extends FrontBaseController
{
+     /**
+      * @var \Shopsys\ReadModelBundle\Product\Listed\ListedProductViewFacadeInterface
+      */
+     private $listedProductsFacade;
```

Please just add here an example, don't apply this suggestion as it would be badly applied probably

simara-svatopluk: And the rest should be in another bullet

* And in involved controllers use the facade for getting views
  * in ProductController::detailAction: ...

docs/upgrade/read-model-for-product-lists.md
```diff
@@ -0,0 +1,140 @@
...
+        - );
+        + $topProducts = $this->listedProductViewFacade->getAllTop();
+        ```
+- add new `CartController::productActionAction`:
+    ```diff
+    // src/Shopsys/ShopBundle/Controller/Front/CartController.php
+
+    + /**
+    +  * @param \Shopsys\ReadModelBundle\Product\Action\ProductActionView $productActionView
+    + * @param string $type
```
simara-svatopluk:
```suggestion
    +  * @param string $type
```

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,127 @@
...
+### 2. Extend `ListedProductViewFactory` so it returns the new required data
+
+The class is responsible for creating the view object. We need to ensure that the objects is created with proper brand name. We are able to get the brand name from the product entity, so we just need to overwrite `createFromProduct()` method.  
+
+```php
+declare(strict_types=1);
+
+namespace Shopsys\ShopBundle\Model\Product\View;
+
+use Shopsys\FrameworkBundle\Model\Product\Product;
```
TomasLudvik:
```suggestion
use Shopsys\ShopBundle\Model\Product\Product;
```

packages/read-model/README.md
vitek-rostislav: hi, I am not able to do that as the link in the split repository would lead to 404

docs/cookbook/extending-product-list.md
vitek-rostislav: the plural is intentional :wink:

docs/upgrade/read-model-for-product-lists.md
simara-svatopluk: @vitek-rostislav ping

packages/read-model/.travis.yml
vitek-rostislav: done in Travis configuration - https://travis-ci.org/shopsys/read-model/settings

docs/cookbook/extending-product-list.md
vitek-rostislav: dont know :laughing: fixed

docs/cookbook/extending-product-list.md
vitek-rostislav: I am not able to do that - it is not possible to overwrite the parent method definition

docs/cookbook/extending-product-list.md
```diff
@@ -0,0 +1,127 @@
...
+use Shopsys\ReadModelBundle\Image\ImageView;
+use Shopsys\ReadModelBundle\Product\Action\ProductActionView;
+use Shopsys\ReadModelBundle\Product\Listed\ListedProductViewFactory as BaseListedProductViewFactory;
+use Shopsys\ReadModelBundle\Product\Listed\ListedProductView as BaseListedProductView;
+
+class ListedProductViewFactory extends BaseListedProductViewFactory
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
```
TomasLudvik:
```suggestion
     * @param \Shopsys\ShopBundle\Model\Product\Product $product
```

packages/framework/src/Twig/ImageExtension.php
vitek-rostislav: in the end, no change that would deserve an additional description

project-base/src/Shopsys/ShopBundle/Resources/views/Front/Inline/Product/productFlags.html.twig
vitek-rostislav: I described the FlagsExtension itself in the commit message as there is no new macro

LukasHeinz:
Awesome job guys @vitek-rostislav @grossmannmartin . From business validation perspective:
- I need more clarification to upgrade notes
- I do not understand what should be done while extending in https://github.com/shopsys/shopsys/blob/rv-read-model-bck/docs/cookbook/extending-product-list.md

Please contact me so we can close this one :-) Thanks

PR #933, opened 2019-04-08
[shopsys] phing target test-db-demo now exports products into elasticsearch with different index

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Tests do not export data to elasticsearch. That would be problem if i have some products changed. This PR supports export products to elasticsearch with his own index for tests.
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Yes 
|Fixes issues| closes #865, closes #866 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1031, opened 2019-05-15
[shopsys] Functional, Smoke and Acceptance tests for PRs

MattCzerner:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to have completely tested each PR by travis, because of that docker-compose was used to create infrastructure in which it is possible.
|New feature| Yes
|Fixes issues| 


.travis.yml
```diff
@@ -1,28 +1,61 @@
...
+    -   TRAVIS_GROUP_ID=$(id -g)
 
 before_install:
-  - phpenv config-add project-base/docker/php-fpm/php-ini-overrides.ini
-  - echo '' | pecl install redis-4.1.1
-  - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
-  - phpenv config-rm xdebug.ini || return 0
-  - composer global require hirak/prestissimo
+-   sudo rm /usr/local/bin/docker-compose
+-   sudo add-apt-repository ppa:rmescandon/yq -y
```
boris-brtan: i think that after adding new repository it is needed to run `apt update`

.travis.yml
```diff
@@ -1,28 +1,44 @@
...
```
boris-brtan: https://docs.travis-ci.com/user/installing-dependencies/#adding-apt-sources

henzigo:
It is a pity that you need to build the entire application and not just tests.

.travis.yml
```diff
@@ -1,28 +1,44 @@
...
+-   mkdir -p $CACHE_DIR
+-   cp -f ./docker/conf/docker-compose.yml.dist ./docker-compose.yml
+-   yq write --inplace docker-compose.yml services.php-fpm.build.args.www_data_uid $TRAVIS_USER_ID
+-   yq write --inplace docker-compose.yml services.php-fpm.build.args.www_data_uid $TRAVIS_GROUP_ID
 
 install:
-  - COMPOSER_MEMORY_LIMIT=-1 travis_retry composer install --optimize-autoloader --no-interaction --no-progress --dev
-  - php phing npm
+-   docker load < $CACHE_FILE_IMAGE || true
+-   docker-compose up -d postgres elasticsearch redis selenium-server php-fpm webserver
```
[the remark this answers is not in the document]
MattCzerner: Very nice :)

.travis.yml
```diff
@@ -1,28 +1,44 @@
...
+-   if [[ ! -f "$CACHE_FILE_IMAGE" ]]; then docker save $(docker-compose images php-fpm | tail -1 | awk '{print $2}') | gzip -1 -c > $CACHE_FILE_IMAGE; fi
+-   docker ps
```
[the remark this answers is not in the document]
MattCzerner: In my opinio, docker ps outputs much more information then docker-compose ps. If you have some reason to use docker-compose please write it and i will use it if it makes sense

.travis.yml
```diff
@@ -1,28 +1,44 @@
-language: php
+language: bash
+dist: xenial
 
-php:
-  - 7.1
+services:
+-   docker
```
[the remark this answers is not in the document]
MattCzerner: its 18.06, why do we need 09? I

.travis.yml
MattCzerner: If it could be used like addons I would love to use it, but YQ is not in the addons packages and I find adding them using sourcelines more dirty than this.

MattCzerner:
Why? In your solution you would need to build application too in order to pass acceptance tests. You already built php-fpm image, so that does not change, but instead of installing third party services through apt-get i just download them with docker, so I dont see the pitty here. I think that this is much cleaner solution..

henzigo:
> Why? In your solution you would need to build application too in order to pass acceptance tests. You already built php-fpm image, so that does not change, but instead of installing third party services through apt-get i just download them with docker, so I dont see the pitty here. I think that this is much cleaner solution..

I meant that you are building database with demodata for application not only for tests (shopsys-test) but I absolutely agree that your solution is cleaner

.travis.yml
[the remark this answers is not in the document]
boris-brtan: 18.09 got buildpack https://medium.com/@tonistiigi/build-secrets-and-ssh-forwarding-in-docker-18-09-ae8161d066

.travis.yml
boris-brtan: OK then

boris-brtan:
> I meant that you are building database with demodata for application not only for tests (shopsys-test) but I absolutely agree that your solution is cleaner

@henzigo are we able to test application without the no-test database ? https://github.com/shopsys/shopsys/issues/1039
solution could be to modify parameters.yml or parameters-test.yml

henzigo:
@boris-brtan In current state of framework you simply can't. You need no-test database for some kind of setting value (I didn't examine it in more detail), for generate error pages and in 7.2 for export products to elastic (fixed in 8.0 https://github.com/shopsys/shopsys/pull/933).

.travis.yml
```diff
@@ -1,28 +1,56 @@
-language: php
-
-php:
-  - 7.1
+language: bash
+dist: xenial
+stages:
+-  Tests
+-  Third party services dependent tests
```
boris-brtan: this stage is not used so i think whole stages: section could be removed
```yaml
jobs:
    include:
    -   stage: Tests
        script: echo 1
    -   script: echo 2
    -   stage: 3rd party
        script: echo 3
    -   stage: 3rd party
        script: echo 4
    -   stage: 3rd party
        script: echo 5
```

would create 2 stages without this section

boris-brtan:
so squash it and OK

PR #1036, opened 2019-05-17
[shopsys] added note about recreating structure during deployment in upgrade guide for v7.2.0

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| during the upgrade of `shopsys/demoshop` I recreated the structure locally, but I forgot to recreate it during the deployment (as it is not a standard part of the deployment process). This lead to a *500 Internal Server Error* on the product list page.
|New feature| No 
|Fixes issues| ... 


henzigo:
Hello, 
I have question about that. Why I have to do this manually? That should be automatically from start using elastic.

PetrHeinz:
@henzigo Because currently, it's not a part of the standard deployment process. Structure recreation is only needed when the structure changes so it's not really necessary to recreate during every deployment. Also, it might be quite a lengthy operation because all product data need to be exported afterwards.

These issues could be resolved by change detection and/ or spinning up a separate instance/namespace/prefix in Elasticsearch for the new build version (similarly to #886) - none of these have been implemented yet.

We felt that until these are resolved, it would be best to keep it updated manually. As the Phing target configuration and the deployment process is part of the project repository, you have full control over it and you are able to make it recreate the structure automatically.

PR #1047, opened 2019-05-17
[shopsys] encapsulation of collections inside of entities

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| https://ocramius.github.io/doctrine-best-practices/#/43
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Administrator/Administrator.php
+++ b/packages/framework/src/Model/Administrator/Administrator.php
@@ -61,7 +61,8 @@ class Administrator implements UserInterface, Serializable, UniqueLoginInterface
     protected $email;
 
     /**
-     * @var \Doctrine\Common\Collections\ArrayCollection|\Shopsys\FrameworkBundle\Model\Administrator\AdministratorGridLimit[]
+     * @var \Shopsys\FrameworkBundle\Model\Administrator\AdministratorGridLimit[]|\Doctrine\Common\Collections\Collection
+     *
      * @ORM\OneToMany(
      *     targetEntity="Shopsys\FrameworkBundle\Model\Administrator\AdministratorGridLimit",
      *     mappedBy="administrator",
--- a/project-base/src/Shopsys/ShopBundle/Form/Front/Order/TransportAndPaymentFormType.php
+++ b/project-base/src/Shopsys/ShopBundle/Form/Front/Order/TransportAndPaymentFormType.php
@@ -95,7 +95,7 @@ class TransportAndPaymentFormType extends AbstractType
 
         $relationExists = false;
         if ($payment instanceof Payment && $transport instanceof Transport) {
-            if ($payment->getTransports()->contains($transport)) {
+            if (in_array($transport, $payment->getTransports(), true)) {
                 $relationExists = true;
             }
         }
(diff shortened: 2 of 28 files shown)


vitek-rostislav:
Hi @pk16011990, thanks for the PR. My apologies for the very late response. Processing of your PR is in our buffer, however, there are some other priorities in our sprint so it might take a while.

boris-brtan:
what do you think of this PR @vitek-rostislav ???

boris-brtan:
@pk16011990 if there is some change in frameworkBundle that needs changes in shopBundle tests it is considered as BC-break, so this should be pointed into 8.0

boris-brtan:
all bc-breaks should be based on new MAJOR version branch. In this case `8.0`
rebased on 8.0 resolved conflicts, added more annotation fixes

pk16011990:
I would like to ask you for a favor. Can you please finish this PR without me? After a long pause there were conflicts and I have no guarantee that I will solve it and there will be new conflicts before you look again after XY days.

I understand that it can delay the merge, but I'm OK with it.

packages/framework/src/Model/Administrator/Administrator.php
```diff
@@ -61,7 +61,8 @@ class Administrator implements UserInterface, Serializable, UniqueLoginInterface
     protected $email;
 
     /**
-     * @var \Doctrine\Common\Collections\ArrayCollection|\Shopsys\FrameworkBundle\Model\Administrator\AdministratorGridLimit[]
+     * @var \Shopsys\FrameworkBundle\Model\Administrator\AdministratorGridLimit[]|\Doctrine\Common\Collections\ArrayCollection
```
simara-svatopluk:
```suggestion
     * @var \Shopsys\FrameworkBundle\Model\Administrator\AdministratorGridLimit[]|\Doctrine\Common\Collections\Collection
```

Initially the `ArrayCollection` is true, but once the Doctrine hydrates the entity, it hydrates the `$gridLimit` as a `PersistentCollection`, and these two classes are different. The only think that they have in common is the interface `Collection`

simara-svatopluk: Please change it in all occurences in this PR

project-base/src/Shopsys/ShopBundle/Form/Front/Order/TransportAndPaymentFormType.php
```diff
@@ -95,7 +95,7 @@ public function validateTransportPaymentRelation(OrderData $orderData, Execution
 
         $relationExists = false;
         if ($payment instanceof Payment && $transport instanceof Transport) {
-            if ($payment->getTransports()->contains($transport)) {
+            if (in_array($transport, $payment->getTransports(), true) === true) {
```
simara-svatopluk:
```suggestion
            if (in_array($transport, $payment->getTransports(), true)) {
```

boris-brtan:
the whole day is github freezed and buggy https://travis-ci.org/shopsys/shopsys/builds/551349551

TomasLudvik:
I think that we should mention, that return values of some framework methods has changed - what do you think guys? @simara-svatopluk @boris-brtan @pk16011990

simara-svatopluk:
Please change the commit message (and then I approve the PR). Tomas is right - there are changes in return values and upgrade notes should reflect it

boris-brtan:
commimt message changed and upgrade notes updated

PR #1049, opened 2019-05-18
prepare DI configuration for project-base like framework

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Developer wants extend/override/use more services that only Facade/Factory/Repository in model. Namespaces Component and Twig are used too...
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| no 
|Fixes issues|  

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -97,6 +97,21 @@ There you can find links to upgrade notes for other versions too.
         -   $this->assertIdWithFilter($filter, [19, 17]);
         +   $this->assertIdWithFilter($filter, [17, 19]);
         ```
+- extend DI configuration for your project by updating ([#1049](https://github.com/shopsys/shopsys/pull/1049))
+    - `src/Shopsys/ShopBundle/Resources/config/services.yml`
+        ```diff
+        -    Shopsys\ShopBundle\Model\:
+        -        resource: '../../Model/**/*{Facade,Factory,Repository}.php'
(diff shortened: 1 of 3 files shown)


boris-brtan:
rebased on current master, added UPGRADE notes and solved conflicts.
Now we will do the CR of this PR.

pk16011990:
Hi, 

1) my point of view and examples from real implementation on project base are:
I extend/owerride a lot of class from SSFW by 
```
# src/Shopsys/ShopBundle/Resources/config/services.yml
Shopsys\FrameworkBundle\Model\Payment\PaymentDataFactoryInterface: '@Shopsys\ShopBundle\Model\Payment\PaymentDataFactory'
```
It works, but when I want extend something else (or elsewhere) what are not prepared in project base, I'm suprised why this way extension/overriding does not work. Or components are not intended for extending? Is it clear?


2) yes, i think, that it is DRY violation, This PR only fixs inconsitency from past in this DRY violation. This DRY violation caused me confused, why one class can be extended and another does not work.

vitek-rostislav:
Hi @pk16011990, thanks for your response.
>components are not intended for extending? 

Of course, they are :blush: My point was only that you do not need the empty folder there (when extending a component, it is up to you where to put your new class, and you can create the folder once it is needed) - so if I get it correctly, you just want to have the file structure unified with `framework` in `project-base` as well, right? I have no problem with that :wink: 

I will create an issue for that DRY problem :slightly_smiling_face:

PR #1053, opened 2019-05-20
[framework] Elasticsearch debug toolbar

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Because developers want it and need it
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchCollector.php
+++ b/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchCollector.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Elasticsearch\Debug;
+
+use Exception;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+
+class ElasticsearchCollector extends DataCollector
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Elasticsearch\Debug\ElasticsearchDebugStack
+     */
+    protected $elasticsearchDebugStack;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\Debug\ElasticsearchDebugStack $elasticSearchDebugStack
+     */
+    public function __construct(ElasticsearchDebugStack $elasticSearchDebugStack)
+    {
+        $this->elasticsearchDebugStack = $elasticSearchDebugStack;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function collect(Request $request, Response $response, Exception $exception = null): void
+    {
+        $totalRequestsTime = 0;
+        $collectedData = $this->elasticsearchDebugStack->getCollectedData();
+        foreach ($collectedData as $requestData) {
+            $totalRequestsTime += $requestData['duration'];
+        }
+
+        $this->data = [
+            'requests' => $collectedData,
+            'requestsCount' => count($collectedData),
+            'totalRequestsTime' => $totalRequestsTime,
+        ];
+    }
+
+    public function reset(): void
+    {
+        $this->data = [];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getName(): string
+    {
+        return 'shopsys.elasticsearch_collector';
+    }
+
+    /**
+     * @return array
+     */
+    public function getData(): array
+    {
+        return $this->data;
+    }
+}
--- a/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchCollector.php
+++ b/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchCollector.php
@@ -12,16 +12,16 @@ use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 class ElasticsearchCollector extends DataCollector
 {
     /**
-     * @var \Shopsys\FrameworkBundle\Component\Elasticsearch\Debug\ElasticsearchDebugStack
+     * @var \Shopsys\FrameworkBundle\Component\Elasticsearch\Debug\ElasticsearchRequestCollection
      */
-    protected $elasticsearchDebugStack;
+    protected $elasticsearchRequestCollection;
 
     /**
-     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\Debug\ElasticsearchDebugStack $elasticSearchDebugStack
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\Debug\ElasticsearchRequestCollection $elasticsearchRequestCollection
      */
-    public function __construct(ElasticsearchDebugStack $elasticSearchDebugStack)
+    public function __construct(ElasticsearchRequestCollection $elasticsearchRequestCollection)
     {
-        $this->elasticsearchDebugStack = $elasticSearchDebugStack;
+        $this->elasticsearchRequestCollection = $elasticsearchRequestCollection;
     }
 
     /**
@@ -29,16 +29,10 @@ class ElasticsearchCollector extends DataCollector
      */
     public function collect(Request $request, Response $response, Exception $exception = null): void
     {
-        $totalRequestsTime = 0;
-        $collectedData = $this->elasticsearchDebugStack->getCollectedData();
-        foreach ($collectedData as $requestData) {
-            $totalRequestsTime += $requestData['duration'];
-        }
-
         $this->data = [
-            'requests' => $collectedData,
-            'requestsCount' => count($collectedData),
-            'totalRequestsTime' => $totalRequestsTime,
+            'requests' => $this->elasticsearchRequestCollection->getCollectedData(),
+            'requestsCount' => $this->elasticsearchRequestCollection->getCollectedDataCount(),
+            'totalRequestsTime' => $this->elasticsearchRequestCollection->getTotalTime() * 1000,
         ];
     }
 
--- a/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchDebugStack.php
+++ b/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchDebugStack.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Elasticsearch\Debug;
+
+class ElasticsearchDebugStack
+{
+    /**
+     * @var array
+     */
+    protected $collectedData = [];
+
+    /**
+     * @return array
+     */
+    public function getCollectedData(): array
+    {
+        return $this->collectedData;
+    }
+
+    /**
+     * @param string $requestCurl
+     * @param string|null $requestJson
+     * @param mixed $requestData
+     * @param string $method
+     * @param string $uri
+     * @param int|null $statusCode
+     * @param mixed $response
+     * @param float $duration
+     */
+    public function addRequest(
+        string $requestCurl,
+        ?string $requestJson,
+        $requestData,
+        string $method,
+        string $uri,
+        ?int $statusCode,
+        $response,
+        float $duration
+    ): void {
+        $this->collectedData[] = [
+            'requestCurl' => $requestCurl,
+            'requestJson' => $requestJson,
+            'requestData' => $requestData,
+            'method' => $method,
+            'uri' => $uri,
+            'statusCode' => $statusCode,
+            'response' => $response,
+            'duration' => $duration,
+        ];
+    }
+}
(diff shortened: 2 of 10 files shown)


vitek-rostislav:
Hi @pk16011990, thanks for the PR and sorry for the late response. This PR is planned in our current sprint so it will be processed shortly.

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
```diff
@@ -0,0 +1,139 @@
...
+            {% set statusClass = 'status--ok' %}
+        {% endif %}
+        <table>
+            <tr style="">
+                <th class="{{ statusClass }}" style="width: 28px;">{{ request.statusCode }}</th>
+                <th class="{{ statusClass }}" style="width: 70px;">{{ request.method }}</th>
+                <th class="{{ statusClass }}">{{ request.uri }}</th>
+                <td class="{{ statusClass }}" style="width: 200px;">
+                    {% if request.requestJson is not null %}
+                        <button type="button" style="float: right;" onclick="this.style.display = 'none'; document.getElementById('js-elasticsearch-response-data-{{ index }}').style.display = ''; return false;">
```
boris-brtan: what about toggle functionality?

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
```diff
@@ -0,0 +1,139 @@
...
+            </tr>
+            <tr>
+                <td colspan="4" style="padding: 5px; word-break: break-all;">
+                    {{ request.requestCurl  }}
+                </td>
+            </tr>
+
+            {% if request.requestJson is not null %}
+                <tr id="js-elasticsearch-request-json-{{ index }}" style="display: none;">
+                    <td colspan="4"><pre>{{ request.requestJson }}</pre></td>
```
boris-brtan: is it possible to dump also json data?

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $this->elasticsearchDebugStack = $elasticsearchDebugStack;
+    }
+
+    /**
+     * @param mixed $level
+     * @param string $message
+     * @param array $context
+     */
+    public function log($level, $message, array $context = []): void
```
boris-brtan:
```suggestion
    public function log($level, string $message, array $context = []): void
```

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
```diff
@@ -0,0 +1,91 @@
...
+
+        $exceptionMessage = sprintf('Not supported log level `%s`', $level);
+        throw new NotSupportedException($exceptionMessage);
+    }
+
+    /**
+     * @param string $message
+     * @param array $context
+     */
+    protected function logRequest($message, array $context = []): void
```
boris-brtan:
```suggestion
    protected function logRequest(string $message, array $context = []): void
```

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
pk16011990: Why? Using of function `dump()` writes mess around like string length etc... Tag <pre> is exact what I needed for readable and copy-paste into Kibana for example...
So, It is my point of view, If SSFW team preferes dump, you can.

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
pk16011990: Yes I wanted toggle, but plain Javascript is too long and I didn't want a lot of JS in template.
However, I am agree to any improvements.

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
boris-brtan: ok, if that is some input into tools like KIBANA that is OK for me

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
boris-brtan: so what about this?
```diff
-                        <button type="button" style="float: right;" onclick="this.style.display = 'none'; document.getElementById('js-elasticsearch-response-data-{{ index }}').style.display = ''; return false;">
+                        <button type="button" style="float: right;" onclick="document.getElementById('js-elasticsearch-response-data-{{ index }}').classList.toggle('hidden'); return false;">
-                        <button type="button" style="float: right;" onclick="this.style.display = 'none'; document.getElementById('js-elasticsearch-request-data-{{ index }}').style.display = ''; return false;">
+                        <button type="button" style="float: right;" onclick="document.getElementById('js-elasticsearch-request-data-{{ index }}').classList.toggle('hidden'); return false;">
-                        <button type="button" style="float: right;" onclick="this.style.display = 'none'; document.getElementById('js-elasticsearch-request-json-{{ index }}').style.display = ''; return false;">
+                        <button type="button" style="float: right;" onclick="document.getElementById('js-elasticsearch-request-json-{{ index }}').classList.toggle('hidden'); return false;">
-                <tr id="js-elasticsearch-request-json-{{ index }}" style="display: none;">
+                <tr id="js-elasticsearch-request-json-{{ index }}" class="hidden">
-                <tr id="js-elasticsearch-request-data-{{ index }}" style="display: none;">
+                <tr id="js-elasticsearch-request-data-{{ index }}" class="hidden">
-                <tr id="js-elasticsearch-response-data-{{ index }}" style="display: none;">
+                <tr id="js-elasticsearch-response-data-{{ index }}" class="hidden">
```

boris-brtan: https://caniuse.com/#feat=classlist

pk16011990: excellent, thanks for something new for me :)

boris-brtan: so please make fixup, or you want me to do it?

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
```diff
@@ -0,0 +1,91 @@
...
+        if ($message !== 'Response:') {
+            $exceptionMessage = sprintf('Not supported message `%s`, It supports only exactly `Response:`', $message);
+            throw new NotSupportedException($exceptionMessage);
+        }
+
+        $matches = null;
+        $requestJson = null;
+        $requestData = null;
+
+        if (preg_match('/^.* -d \'(?<json>.*)\'$/U', $this->lastRequestCurl, $matches) !== 0) {
```
simara-svatopluk: It took me a while to understand what is this IF doing.. please extract the IF (including temporary variables) into meaningful protected functions... like 
```php
$requestData = $this->extractData($this->lastReqestCurl);
$requestJson = $this->formatData($requestData);
```

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchDebugStack.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Elasticsearch\Debug;
+
+class ElasticsearchDebugStack
```
simara-svatopluk: `stack` is a data structure with different properties than this data structure.

Please rename this class to match the responsibility. The best name I can come with is `ElasticsearchRequestCollection`

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchCollector.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function collect(Request $request, Response $response, Exception $exception = null): void
+    {
+        $totalRequestsTime = 0;
+        $collectedData = $this->elasticsearchDebugStack->getCollectedData();
+        foreach ($collectedData as $requestData) {
```
simara-svatopluk: I suggest to move this calculation to the `ElasticsearchDebugStack`. So in this place you'll call only `$totalRequestsTime = $this->elasticsearchDebugStack->getTotalTime()`

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchCollector.php
```diff
@@ -0,0 +1,65 @@
...
+            $totalRequestsTime += $requestData['duration'];
+        }
+
+        $this->data = [
+            'requests' => $collectedData,
+            'requestsCount' => count($collectedData),
```
simara-svatopluk: I suggest to move the counting to the `ElasticsearchDebugStack` in similar fastion as I suggest in a previous comment... `'requestsCount' => $this->elasticsearchDebugStack->getCount()`

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchCollector.php
simara-svatopluk: And after this refactoring you don't even need the temporary variable `$totalRequestsTime` and you can have cople of lines later following code:
`'totalRequestsTime' => $this->elasticsearchDebugStack->getTotalTime()`

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
```diff
@@ -0,0 +1,139 @@
+{% extends '@WebProfiler/Profiler/layout.html.twig' %}
+
+{% block toolbar %}
+    {% if collector.data.requestsCount > 0 %}
+        {% if collector.data.totalRequestsTime * 1000 > 200 %}
```
simara-svatopluk: I'm just wondering why these numbers? (I would wonder for any number without explanation) Could you please comment in the code the reason?

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
```diff
@@ -0,0 +1,139 @@
...
+            {% set status = 'red' %}
+        {% elseif collector.data.totalRequestsTime * 1000 > 100 %}
```
simara-svatopluk: `collector.data.totalRequestsTime` is always multiplied by 1000 in the template. I suggest to move this multiplication to the source of these data (ElasticsearchCollector, eventually ElasticsearchDebugStack if you accept moving methods idea)

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
```diff
@@ -0,0 +1,139 @@
...
+                <span class="sf-toolbar-label">in</span>
+                <span class="sf-toolbar-value">{{ '%0.2f'|format(collector.data.totalRequestsTime * 1000) }}</span>
+                <span class="sf-toolbar-label">ms</span>
+            </span>
+        {% endset %}
+
+        {% set text %}
+            <div class="sf-toolbar-info-piece">
+                <b>Requests</b>
+                <span class="sf-toolbar-status {{ collector.data.requestsCount > 5 ? 'sf-toolbar-status-yellow' : '' }}">{{ collector.data.requestsCount }}</span>
```
simara-svatopluk: Again, my obsession for hardcoded constants hits me :-) Could you please add a comment with explanation why the "5" is normal and why over "5" should be highlighted?

pk16011990:
I understand your points of view, but I don't care about it any more. You can do anything what you want with this PR. In my version it solves my problem, but I understand that you need more from it, I don't.

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
boris-brtan: there is interface which does not allow it

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
boris-brtan: there is interface which does not allow it

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
boris-brtan: or maybe it could be done like `        {% if collector.data.totalRequestsTime > 0.2 %}`

packages/framework/src/Resources/views/Debug/Elasticsearch/template.html.twig
boris-brtan: this is just based on doctrine toolbar divided by 10 collector.querycount > 50 ? 'sf-toolbar-status-yellow'
i added twig comment there

PR #1040, opened 2019-05-17
[shopsys] upgrade phpstan to level 4

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| we want to keep the code the best it can be so using higher level of phpstan is the best choice
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| closes #820
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1058, opened 2019-05-21
[framework] CountryFormType uses NotInArray instead of Callback constraint now

### LINKED ISSUE #1057: Creating/editing country with no code results in 500
### What is happening
I am in Administration Settings >  Lists and nomenclatures >  Countries >  Editing country - Czech republic

When I edit country and delete code (so there is nothung in code input), and save changes, it results in 500.

Note: also javascript validation for this input doesn't work.

### Expected result
There should be no 500 but standard info, that code is mandatory.

PetrHeinz:
- `CountryFacade::findByCode()` is now called only with a string parameter  
- validation of a value should not overwrite the `CountryFormType::$country` property
- fixed annotations (`CountryFormType::$country` can be `null` and the validator method accepts any value)

| Q             | A
| ------------- | ---
|Description, reason for the PR| validation in `CountryFormType` was failing a type check when calling `CountryFacade::findByCode()`, the validation method was unnecessarily nested and it was overwriting a form type property, which was suspicious
|New feature| No 
|Fixes issues| fixes #1057  


packages/framework/src/Form/Admin/Country/CountryFormType.php
```diff
@@ -119,18 +119,21 @@ public function configureOptions(OptionsResolver $resolver): void
     }
 
     /**
-     * @param $countryCodeValue
+     * @param mixed $countryCodeValue
```
boris-brtan: this was changed to string during [implementation of phpstan](https://github.com/shopsys/shopsys/pull/1040/files#diff-ccb16d8308d46464c96de7ce7571ef9bL122). What types can this have ?

PetrHeinz: I would expect `string|null`, but really forms can pass anything to a validator when a data transformer is added... That's why I annotated it `mixed`

PetrHeinz: I've resolved the conflict using a merge commit (I plan to merge this PR using a squash and merge)

PetrHeinz:
@LukasHeinz I've fixed the JS validation (even for the check of code uniqueness). It required me to rework the whole validation.

PR #1063, opened 2019-05-22
[migrations] Configuration::shouldExecuteMigration is fixed

boris-brtan:
- wrong return type annotation from doctrine/migrations package was overriden with the correct one
- wrongly implemented call of getVersion method for string instance was fixed

| Q             | A
| ------------- | ---
|Description, reason for the PR| after merge of #1040 phing target db-migrations started to fail because of the wrong return type annotation of method from doctrine/migrations one method was wrongly implemented when we simplifying it.
|New feature| No 
|Fixes issues| 

### DIFF
--- a/packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
+++ b/packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
@@ -11,8 +11,8 @@ use Doctrine\DBAL\Migrations\Version;
 use Shopsys\MigrationBundle\Component\Doctrine\Migrations\Exception\MethodIsNotAllowedException;
 
 /**
- * @see https://github.com/doctrine/migrations/pull/824
  * @method string[] getMigratedVersions()
+ * @see https://github.com/doctrine/migrations/pull/824
  */
 class Configuration extends DoctrineConfiguration
 {
--- a/packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
+++ b/packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
@@ -138,13 +138,7 @@ class Configuration extends DoctrineConfiguration
      */
     private function shouldExecuteMigration(Version $version, array $migratedVersions)
     {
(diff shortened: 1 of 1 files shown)


packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
```diff
@@ -10,6 +10,10 @@
 use Doctrine\DBAL\Migrations\Version;
 use Shopsys\MigrationBundle\Component\Doctrine\Migrations\Exception\MethodIsNotAllowedException;
 
+/**
```
vitek-rostislav: TBH? I am a little bit confused with this comment - could you please remove it? The first line would make sense in the commit message, the second line seems useless to me

boris-brtan: this should help developers with the future development when they will not be mistaken by wrong autosuggestion in php-IDE

packages/migrations/src/Component/Doctrine/Migrations/Configuration.php
```diff
@@ -129,13 +133,13 @@ public function getMigrationsToExecute($direction, $to)
...
      * @param \Doctrine\DBAL\Migrations\Version $version
-     * @param \Doctrine\DBAL\Migrations\Version[] $migratedVersions
+     * @param string[] $migratedVersions
      * @return bool
      */
     private function shouldExecuteMigration(Version $version, array $migratedVersions)
     {
         foreach ($migratedVersions as $migratedVersion) {
-            if ($version->getVersion() === $migratedVersion->getVersion()) {
+            if ($version->getVersion() === $migratedVersion) {
```
PetrHeinz: I guess we could simplify the whole method using `!in_array($version->getVersion(), $migratedVersions, true)`, just like it was before #1040...

boris-brtan:
please someone from @PetrHeinz or @vitek-rostislav for requestChanges or approval

PR #1076, opened 2019-05-27
[shopsys] composer.json in all shopsys packages now use latest version in 8.0

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Using `dev-master` as a dependency means that if I were to install a project repository from `8.0` branch, `shopsys/framework` and other packages would be installed in `dev-master` version (and thus latest `7.x` version). This should have been done during the creation of the `8.0` branch imo.
|New feature| No 
|Fixes issues| ... 


boris-brtan:
added also read-model package

PR #1092, opened 2019-05-30
[framework] ImageConfigDefinition bugfix - added additionalSizes under types

### LINKED ISSUE #1088: ImageConfigLoader - not correct loading/config of additionalSizes
### What is happening
There is a difference between definition of "images.yml" in ``ImageConfigDefinition`` and execution of this config in ``prepareSizes()`` function in ``ImageConfigLoader``. 

I have tried to add "types" into sliderItem config in images.yml - based on example in this file.
```
name: sliderItem
    class: Shopsys\FrameworkBundle\Model\Slider\SliderItem
    types:
        -   name: default
            sizes:
                -   name: ~
                    width: 1232
                    height: 480
                    crop: true
                    occurrence: 'Front-end: Main page slider'
```
This is just a piece of code, that should work. 

But it ends with the error - undefined index additionalSizes:
```$additionalSizes = $this->prepareAdditionalSizes($sizeName ?: '~', $sizeConfig[ImageConfigDefinition::CONFIG_SIZE_ADDITIONAL_SIZES]);```
 If I put it there, it returns - additionalSizes cannot be there. Because the config for sizes under the types do not allow additionalSizes. So it is an endless circle of add -> remove -> add -> remove additional sizes :)
### Expected result

Config file or loading function should be modified. For now, "type" functionality in images cannot be used.

### LINKED ISSUE #1093: DRY violation in ImageConfigDefinition
### What is happening

[ImageConfigDefinition](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Component/Image/Config/ImageConfigDefinition.php) violates [the DRY principle](https://github.com/shopsys/shopsys/blob/v7.2.0/docs/contributing/code-quality-principles.md#dont-repeat-yourself). The `'sizes'` configuration definition is identical in the entity image root and in the individual image types.

The violation of DRY caused #1088 during #766 where only one place was updated.

### Expected result

`ImageConfigDefinition` is refactored so that DRY principle is followed.

HakimFractal:
| Q             | A
| ------------- | ---
|Description, reason for the PR| bug in configuration file
|New feature| No 
|Fixes issues| closes #1088 and resolves #1093


PetrHeinz:
#1093 created

HakimFractal:
I had the same thoughts about just copying that piece of code :) I can create there some private function to include that code if you want

PetrHeinz:
@HakimFractal That would be great. I'm just a bit worried about that fluent syntax which may get hard to read when trying to extract some code...

HakimFractal:
It will be more complicated to understand how the config should be structured, but it depends how often someone will look for it. There is an example in images.yml which is easy to understand. 
So, what is a conclusion? Do a private function or not?

PetrHeinz:
I think it would be better to have it extracted :slightly_smiling_face:

HakimFractal:
There is a new commit. Please review.

PetrHeinz:
You're right about the type annotation - the problem was that PHPStan wasn't sure about the type.

The method `prototype('array')` has the same issue. Fortunately there is a specifically typed alias `arrayPrototype`.

I've fixed it in a new commit alongside a method rename `addSizesNode` -> `createSizesNode` as the method only creates the node, which is then added via the `append(...)`. Is that OK with you?

It should pass the static analysis now, I think.

HakimFractal:
I fine with your changes :)

PR #1099, opened 2019-06-02
[framework] administration: quicksearch for products via dashboard

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| for the speed purposes and better UX we would like to have direct connection between products and dashboard (default page after administrator is logged in)
|New feature| No 
|Fixes issues|  

### DIFF
--- a/packages/framework/src/Controller/Admin/DefaultController.php
+++ b/packages/framework/src/Controller/Admin/DefaultController.php
@@ -4,6 +4,8 @@ namespace Shopsys\FrameworkBundle\Controller\Admin;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Shopsys\FrameworkBundle\Component\Setting\Setting;
+use Shopsys\FrameworkBundle\Form\Admin\QuickSearch\QuickSearchFormData;
+use Shopsys\FrameworkBundle\Form\Admin\QuickSearch\QuickSearchFormType;
 use Shopsys\FrameworkBundle\Model\Mail\MailTemplateFacade;
 use Shopsys\FrameworkBundle\Model\Product\Availability\AvailabilityFacade;
 use Shopsys\FrameworkBundle\Model\Product\Unit\UnitFacade;
@@ -68,6 +70,7 @@ class DefaultController extends AdminBaseController
 
     /**
      * @Route("/dashboard/")
+     * @return \Symfony\Component\HttpFoundation\Response
      */
     public function dashboardAction()
     {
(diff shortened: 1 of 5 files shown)


packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -68,8 +71,10 @@ public function __construct(
 
     /**
      * @Route("/dashboard/")
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function dashboardAction()
+    public function dashboardAction(Request $request)
```
PetrHeinz: This would be considered a BC break. Please, extract the redirect into a new controller action to avoid it.

packages/framework/src/Resources/views/Admin/Content/Default/quickSearchFormContent.html.twig
```diff
@@ -0,0 +1,11 @@
...
+    <div class="box-quick-search">
+        {{ form_widget(quickSearchForm.text, { attr:
+            {
+                class: 'box-quick-search__input',
+                placeholder: 'Search by product name, catalogue number or product part number. It is possible to use wildcards * and ?.'|trans
+            }
+        })}}
+        {{ form_widget(quickSearchForm.submit, { label: 'Search [verb]'|trans, attr: { class: 'box-quick-search__btn'}, full_name: '' }) }}
+    </div>
+{{ form_end(quickSearchForm) }}
```
PetrHeinz:
```suggestion
{{ form_end(quickSearchForm) }}

```

packages/framework/src/Controller/Admin/DefaultController.php
boris-brtan: used action within formCreate method

boris-brtan:
i want to contact @LukasHeinz about the placeholder and design

boris-brtan:
rebased onto 8.0 @LukasHeinz @tomshopsys

packages/framework/src/Resources/views/Admin/Content/Default/index.html.twig
```diff
@@ -4,20 +4,33 @@
...
+    <div class="margin-bottom-20">
+        <div class="width-50-percent float-left">
+            <h2>{{ 'Quick product search'|trans }}</h2>
+        </div>
+        <div class="width-50-percent float-left text-right padding-top-10">
+            <a href="{{ url('admin_product_list', { as: [{ subject: 'productName' }] }) }}">
+                {{ 'Advanced search'|trans }}
+            </a>
+        </div>
+        <div class="in-tab__content clear">
```
PetrHeinz:
```suggestion
        <div class="clear">
```

The `in-tab__content` class is probably a left over from the previous implementation. Now it doesn't make sense as it is not in an element in a `in-tab` component (BEM).

PR #1109, opened 2019-06-06
[framework] checking of order item type is fixed by using type getters

### LINKED ISSUE #1100: Quick preview to order
### What is happening
If customer finish order, I can see like administrator in orders Quick Review of that order.

The problem is that I see 3 times constant "Shipping" in quick view. -> http://prntscr.com/nwukz0

### Expected result
In quick review show constants like -> item/shipping/payment

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| info in issue #1100
|New feature| No 
|Fixes issues| fixes #1100 

### DIFF
--- a/packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
@@ -13,9 +13,9 @@
             {% for item in order.items %}
                 <tr>
                     <td>
-                        {% if item.isTypeTransport() %}
+                        {% if item.typeTransport %}
                             {{ 'Shipping'|trans }} -
-                        {% elseif item.isTypePayment() %}
+                        {% elseif item.typePayment %}
                             {{ 'Payment'|trans }} -
                         {% endif %}
                         {{ item.name }}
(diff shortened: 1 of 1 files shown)


packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
```diff
@@ -13,9 +13,9 @@
             {% for item in order.items %}
                 <tr>
                     <td>
-                        {% if item.transport is defined %}
+                        {% if item.isTypeTransport() %}
```
PetrHeinz:
```suggestion
                        {% if item.typeTransport %}
```

packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
```diff
@@ -13,9 +13,9 @@
...
                             {{ 'Shipping'|trans }} -
-                        {% elseif item.payment is defined %}
+                        {% elseif item.isTypePayment() %}
```
PetrHeinz:
```suggestion
                        {% elseif item.typePayment %}
```

PR #1113, opened 2019-06-07
[shopsys] ProductAvailabilityCalculation: fix calculation of not-yet-persisted products

### LINKED ISSUE #1112: Creating product variants when main product variant is set to use stock
### What is happening
I am creating product A set up as “Use stocks” > yes and “Action after sellout:” > hide product or exclude from sale. 

I am choosing in the products overview “Create variant” and set up product A as “main variant” and any other product is chosen as sub variant. 

When trying to save it like that the application shows error E500 Doctrine\ORM\ORMInvalidArgumentException "Binding entities to query parameters only allowed for entities that have an identifier."

### Note
The issue is there when availability_id is set to null in the product

PetrHeinz:
- added new functional test for product variant creation to avoid regression (the issue is related to the one mentioned in #581)
- mocks of `EntityManager` that are used in the calculation class have to have the `contains` method return `true` in tests

| Q             | A
| ------------- | ---
|Description, reason for the PR| An unhandled exception was thrown when a variant is created from a product which has no availability set (i.e. newly created product using stock).
|New feature| No 
|Fixes issues| fixes #1112 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -59,7 +59,7 @@ There you can find links to upgrade notes for other versions too.
             +     return $entityManagerMock;
             + }
             ```
-    - you can copy-paste a new functional test [`ProductVariantCreationTest.php`](/project-base/tests/ShopBundle/Functional/Model/Product/ProductVariantCreationTest.php) into `tests/ShopBundle/Functional/Model/Product/` to avoid regression of issues with creating product variants in the future
+    - you can copy-paste a new functional test [`ProductVariantCreationTest.php`](https://github.com/shopsys/project-base/blob/master/tests/ShopBundle/Functional/Model/Product/ProductVariantCreationTest.php) into `tests/ShopBundle/Functional/Model/Product/` to avoid regression of issues with creating product variants in the future
 
 ### Configuration
 - update `phpstan.neon` with following change to skip phpstan error ([#1086](https://github.com/shopsys/shopsys/pull/1086))
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -33,6 +33,33 @@ There you can find links to upgrade notes for other versions too.
 ### Application
 - follow instructions in [the separate article](upgrade-instructions-for-read-model-for-product-lists.md) to introduce read model for frontend product lists into your project ([#1018](https://github.com/shopsys/shopsys/pull/1018))
     - we recommend to read [Introduction to Read Model](/docs/model/introduction-to-read-model.md) article
+- fix up your functional tests because of availability calculation patch in the `shopsys/framework` ([#1113](https://github.com/shopsys/shopsys/pull/1113))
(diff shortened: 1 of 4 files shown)


docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -33,6 +33,33 @@ There you can find links to upgrade notes for other versions too.
...
+            + protected function createEntityManagerMock(): MockObject
+            + {
+            +     $entityManagerMock = $this->createMock(EntityManager::class);
+            +
+            +     $entityManagerMock->method('contains')->willReturn(true);
+            +
+            +     return $entityManagerMock;
+            + }
+            ```
+    - you can copy-paste a new functional test [`ProductVariantCreationTest.php`](/project-base/tests/ShopBundle/Functional/Model/Product/ProductVariantCreationTest.php) into `tests/ShopBundle/Functional/Model/Product/` to avoid regression of issues with creating product variants in the future
```
PetrHeinz: I'd like to change this to the URL in the split repository:

```suggestion
    - you can copy-paste a new functional test [`ProductVariantCreationTest.php`](https://github.com/shopsys/project-base/blob/master/tests/ShopBundle/Functional/Model/Product/ProductVariantCreationTest.php) into `tests/ShopBundle/Functional/Model/Product/` to avoid regression of issues with creating product variants in the future
```

The URL returns 404 now, but it will be ok when this is merged into `master` and split...

project-base/tests/ShopBundle/Functional/Model/Product/ProductVariantCreationTest.php
```diff
@@ -0,0 +1,135 @@
...
+
+        $this->assertTrue($mainVariant->isMainVariant());
+        $this->assertContainsAllVariants([$mainProduct, $secondProduct, $thirdProduct], $mainVariant);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product[] $expectedVariants
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $mainVariant
+     */
+    private function assertContainsAllVariants(array $expectedVariants, Product $mainVariant)
```
boris-brtan:
```suggestion
    private function assertContainsAllVariants(array $expectedVariants, Product $mainVariant): void
```

project-base/tests/ShopBundle/Functional/Model/Product/ProductVariantCreationTest.php
```diff
@@ -0,0 +1,135 @@
...
+        ];
+    }
+
+    /**
+     * @dataProvider variantsWithStockCanBeCreatedProvider
+     * @param int $quantity
+     * @param string $outOfStockAction
+     * @param string|null $outOfStockAvailabilityReference
+     */
+    public function testVariantsWithStockCanBeCreated(int $quantity, string $outOfStockAction, ?string $outOfStockAvailabilityReference): void
```
boris-brtan:
```suggestion
    public function testVariantsWithStockCanBeCreated(int $quantity, string $outOfStockAction, ?string $outOfStockAvailabilityReference = null): void
```

PetrHeinz: I didn't want to make it optional, just nullable.

PR #1104, opened 2019-06-03
[shopsys] phing targets are sorted alphabetically

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| According to our new [Guidelines for Phing Targets](https://github.com/shopsys/shopsys/blob/master/docs/contributing/guidelines-for-phing-targets.md#sorting) introduced in #1068, targets should be sorted alphabetically. As there were a lot of them, I've implemneted an automated sorting command and applied it. It is even a part of the coding standards in monorepo. The command is not robust enough to be included in project repository coding standards imo.
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1116, opened 2019-06-07
[framework] rewrite of shopsys:phing-targets:sort command

PetrHeinz:
- renamed to `shopsys:phing-config:fix`
- the class was renamed to `PhingConfigFixerCommand`
- `SymfonyStyle` is used for input / output
- return codes are used instead of throwing Exceptions (Exceptions are still thrown for actual errors)
- messages were tweaked for better understandability
- the command also normalizes the whitespace in XML

| Q             | A
| ------------- | ---
|Description, reason for the PR| Usage of shopsys:phing-targets:sort introduced in #1104 was clumsy and caused a lot of hard-to-find errors.
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No (the class was not released yet) 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Command/SortPhingTargetsCommand.php
+++ b/packages/framework/src/Command/SortPhingTargetsCommand.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Shopsys\FrameworkBundle\Command;
 
 use SimpleXMLElement;
@@ -22,7 +24,7 @@ class SortPhingTargetsCommand extends Command
      */
     protected static $defaultName = 'shopsys:phing-targets:sort';
 
-    protected function configure()
+    protected function configure(): void
     {
         $this
             ->setDescription('Sort Phing targets alphabetically')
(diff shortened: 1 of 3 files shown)


simara-svatopluk:
@PetrHeinz sorry, I don't have time until next standup (and if there is anything about API, I cannot guarantee my time on monday)

packages/framework/src/Command/SortPhingTargetsCommand.php
```diff
@@ -31,31 +33,52 @@ protected function configure()
     /**
      * @param \Symfony\Component\Console\Input\InputInterface $input
      * @param \Symfony\Component\Console\Output\OutputInterface $output
+     * @return int
      */
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output): int
```
boris-brtan: so in this case, dont we want to use also `declare(strict_types=1);` ?

PetrHeinz: You're right, thanks for noticing!

boris-brtan:
TESTS seems OK now

PR #1122, opened 2019-06-10
[framework] fixed memory leak on shopsys:product-search:export-products command

### LINKED ISSUE #1021: initial command for products export into elasticsearch doesn't release memory usage during run
### What is happening

if we do an initial export of products into elasticsearch the command consumes big amount of memory
`php -dmemory_limit=-1 ./project-base/bin/console shopsys:product-search:export-products`
`php phing product-search-export-products`
50 000 items -> 2GB
200 000 items -> 8GB

so the command fails on memory_limit which is set by default to 512MB.
### Expected result

maybe we can do some fix based on this article ??? https://www.conradk.com/codebase/2018/04/22/memory-management-in-long-lived-symfony-commands/

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Exporting products to Elasticsearch needed more memory, the more products was present. Now the memory usage is stable and low enough (< 100MB). Progressbar representing the number of exported products was added.
|New feature| No 
|Fixes issues| closes #1021 

Disabling SQL logging is in our documentation as [best practice for working with many records](https://github.com/shopsys/shopsys/blob/master/docs/cookbook/basic-data-import.md#best-practices) and it can be found as a known solution – for example https://janhapke.com/blog/optimizing-doctrine-in-long-running-jobs/

Also, the identity map of Doctrine EntityManager is cleared after each batch to prevent memory leaks.

### DIFF
--- a/packages/framework/src/Model/Product/Search/Export/ProductSearchExportRepository.php
+++ b/packages/framework/src/Model/Product/Search/Export/ProductSearchExportRepository.php
@@ -5,6 +5,7 @@ namespace Shopsys\FrameworkBundle\Model\Product\Search\Export;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
+use Shopsys\FrameworkBundle\Component\Paginator\QueryPaginator;
 use Shopsys\FrameworkBundle\Model\Product\Product;
 use Shopsys\FrameworkBundle\Model\Product\ProductVisibility;
 
@@ -70,5 +71,17 @@ class ProductSearchExportRepository
             ->setParameter('variantTypeVariant', Product::VARIANT_TYPE_VARIANT);
 
         return $queryBuilder;
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $locale
+     * @return int
+     */
+    public function getProductTotalCountForDomainAndLocale(int $domainId, string $locale): int
+    {
+        $result = new QueryPaginator($this->createQueryBuilder($domainId, $locale));
+
+        return $result->getTotalCount();
     }
 }
--- a/packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
+++ b/packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
@@ -1,9 +1,16 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Shopsys\FrameworkBundle\Model\Product\Search\Export;
 
+use BadMethodCallException;
+use Doctrine\ORM\EntityManagerInterface;
+use Shopsys\FrameworkBundle\Component\Console\ProgressBarFactory;
+use Shopsys\FrameworkBundle\Component\Doctrine\SqlLoggerFacade;
 use Shopsys\FrameworkBundle\Model\Product\Search\ProductElasticsearchConverter;
 use Shopsys\FrameworkBundle\Model\Product\Search\ProductElasticsearchRepository;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 class ProductSearchExporter
 {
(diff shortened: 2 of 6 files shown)


packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
```diff
@@ -53,7 +115,42 @@ public function export(int $domainId, string $locale): void
             $exportedIds = array_merge($exportedIds, $batchExportedIds);
             $startFrom += static::BATCH_SIZE;
         } while (!empty($batchExportedIds));
-        $this->removeNotUpdated((string)$domainId, $exportedIds);
+        $this->removeNotUpdated($domainId, $exportedIds);
```
simara-svatopluk: is this change intentional? It is in deprecated function...

packages/framework/src/Model/Product/Search/Export/ProductSearchExportRepository.php
```diff
@@ -71,4 +72,16 @@ protected function createQueryBuilder(int $domainId, string $locale): QueryBuild
...
+    /**
+     * @param int $domainId
+     * @param string $locale
+     * @return int
+     */
+    public function getProductTotalCountForDomainAndLocale(int $domainId, string $locale): int
+    {
+        $result = new QueryPaginator($this->createQueryBuilder($domainId, $locale));
+
+        return $result->getTotalCount();
```
simara-svatopluk: cool trick. I thought it is slow, but it seems to be fast (only total count is fetched from DB, not paginated results) :+1:

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
```diff
@@ -53,7 +115,42 @@ public function export(int $domainId, string $locale): void
...
+    /**
+     * @param int $domainId
+     * @param string $locale
+     * @param \Symfony\Component\Console\Style\SymfonyStyle $symfonyStyleIo
+     */
+    public function exportWithOutput(int $domainId, string $locale, SymfonyStyle $symfonyStyleIo): void
+    {
+        $this->validateInjectedDependencies();
+
+        $this->sqlLoggerFacade->temporarilyDisableLogging();
```
simara-svatopluk: why is the logging disabled? Maybe it would worth to explain it in commit message

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
grossmannmartin: Yes because I added strict types notation and this would be broken.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportRepository.php
grossmannmartin: Thanks :)

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
grossmannmartin: Force pushed with extended commit message :)

grossmannmartin:
Thanks :)

I added an instruction to upgrade notes.

simara-svatopluk:
Tested on 5000 products for 2 domains
* works correctly :heavy_check_mark:  - export as same amount of products as the previous version
* consumes significantly lower amount of memory - 32MB vs old 344MB
* is slower :question:  - 55s vs old 43s

I think it is slower because of the cleaning memory operation. But I wanted to point it out so you can discuss this fact during BV

grossmannmartin:
> I think it is slower because of the cleaning memory operation. But I wanted to point it out so you can discuss this fact during BV

That's right because for the second domain the entities are fetched again instead of reusing data from identity map. 
On the BV we agreed it's not a problem.

PR #1125, opened 2019-06-11
[framework] generating of multidomain sitemaps

boris-brtan:
- removing of old sitemap files was separated for each domain
- copying of newly generated files from tmp to mainfilesystem was limited for sitemap domain prefix

| Q             | A
| ------------- | ---
|Description, reason for the PR| during generating sitemaps for multidomain shop, previous domain sitemaps were removed from mainfilesystem. Now there are just removed only those that are processed in the current item from cycle of domains list and also uploaded based on their prefix that contains unique name with domain id.
|New feature| No 
|Fixes issues|  

### DIFF
--- a/packages/framework/src/Model/Sitemap/SitemapDumper.php
+++ b/packages/framework/src/Model/Sitemap/SitemapDumper.php
@@ -57,7 +57,12 @@ class SitemapDumper extends Dumper
         $this->deleteExistingSitemaps($targetDir);
 
         $finder = new Finder();
-        foreach ($finder->files()->in($this->tmpFolder)->getIterator() as $file) {
+        $sitemapFileFinder = $finder
+            ->files()
+            ->name(sprintf('%s*.xml', $this->sitemapFilePrefix))
+            ->in($this->tmpFolder);
+
+        foreach ($sitemapFileFinder->getIterator() as $file) {
             $this->mountManager->move(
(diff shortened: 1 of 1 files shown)


boris-brtan:
mainfilesystem thing so TESTS needed also for GoogleCloud

simara-svatopluk:
tested locally `php bin/console shopsys:cron --module="Shopsys\FrameworkBundle\Model\Sitemap\SitemapCronModule"`

current `master` contains only 2. domain, and that was wrong

and results in 4 files
```
domain_1_sitemap.1.xml
domain_1_sitemap.xml
domain_2_sitemap.2.xml
domain_2_sitemap.xml
```

So it works correctly.

But I have one question - is it correct that the content file for 2. domain is named `domain_2_sitemap.2.xml`? shouldn't it be `domain_2_sitemap.1.xml`? The number should be just counter, am I right?

boris-brtan:
> But I have one question - is it correct that the content file for 2. domain is named `domain_2_sitemap.2.xml`? shouldn't it be `domain_2_sitemap.1.xml`? The number should be just counter, am I right?

good point @simara-svatopluk , seems like it is implementation of the library we use https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php#L241

simara-svatopluk:
ok, I just wanted to point it out

PR #1128, opened 2019-06-13
updated UPGRADE.md to include information about need to rebuild docker images without cache

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There might be situation when Docker build cache prevents to build images well and leads to unexpected errors.
|New feature| No 
|Fixes issues| ... 


boris-brtan:
TESTED OK

grossmannmartin:
I suggest add a note into docker troubleshooting article too.
Because this error can occur not only on upgrade but I encounter it when I tried to build and push image to dockerhub from clean project-base.

PR #1129, opened 2019-06-14
[framework] changed VAT calculation - not using a rounded coefficient now

PetrHeinz:
The rounded VAT coefficient is no longer used for calculating prices which means there will be slight differences in price calculation.
- these changes follow changes in the Czech law and should be in accordance with VAT law in other countries
- assuming you haven't changed the pricing mechanism yourself, it will only affect the selling prices without VAT (selling prices with VAT will not be affected by this change)

All prices in previously created orders will be persistent and will not change after the upgrade, which is ensured by a DB migration. The migration will pass only if it correctly persists all order item prices and the prices add up to the correct totals. This ensures that previously created orders will not have the prices changed.

There is a new option in the orders administration to set all prices manually, which can be used to edit previously created orders (which shouldn't use the new VAT calculation). This option can be used programmatically as well, which might help you to set exact prices with VAT / without VAT and exact totals for order items - see `OrderItemData::$usePriceCalculation` (eg. when importing an order from 3rd party system).    

| Q             | A
| ------------- | ---
|Description, reason for the PR| The previous implementation of VAT calculation used a VAT coefficient rounded to 4 decimal places to calculate the VAT amount included in price with VAT.<br />This can be a bit confusing and can be an issue for price calculation in other countries and now, after an amendment that has changed the [§ 37 of Czech VAT law](https://business.center.cz/business/pravo/zakony/dph/f6436200/#p37), also in Czech Republic.
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No (even though the behavior changes) 
|Fixes issues| ... 

For backward-compatibility some issues had to be addressed during the creation of this PR:

* Total prices for all order items must be persisted as VAT calculation can influence the calculation of total prices
* The prices of existing orders cannot change without explicit action from the administrator
* Prices that are not consistent with current price calculation must be persistable

For this I implemented a DB migration that creates two new columns for total prices, calculates their values using the previous VAT calculation algorithm and checks that the sum of these total prices is equal to the total order prices (the migration will fail if a project has customized the VAT calculation with the instructions how to fix it).

The pricing model has been modified so it now supports saving prices inconsistent with current implementation of price calculation which is essential for the possibility of editing the orders that were created in the past. To achieve this while maintaining BC, `OrderItemData::$usePriceCalculation` has been introduced which allows to control this.

As I cannot modify the method signatures in `OrderItemFactory` and they do not use the `OrderItemData` class, the total price properties in `OrderItem` are nullable. When a `null` is provided, the totals are meant to be calculated (just as they were before).

Administration of orders was modified to allow changing all the prices manually, which required some changes to styling and a fix to JS validation:
image (10)

**TODO before possible merge:**
* [x] writing upgrade notes (possible failing custom tests, ERP integration and `bool $usePriceCalculation`, custom pricing and DB migration, ~change method usage `calculateTotalPrice` -> `getTotalPrice`~)
* [x] consider writing more automated tests covering this functionality
* [x] writing functional tests for testing order editing without using price calculation
* [x] testing of VAT calculation, order creation and editing
* [x] testing from a project repository with input prices with VAT (especially the DB migration)
* [x] testing from a project repository with input prices without VAT (especially the DB migration)
* [x] add a hint icon to the *"Use price calculation"* form header with an explanation of this functionality


packages/framework/src/Form/Admin/Order/OrderItemFormType.php
```diff
@@ -40,6 +45,37 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ],
                 'error_bubbling' => true,
             ])
+            ->add('priceWithoutVat', MoneyType::class, [
+                'scale' => 6,
+                'constraints' => [
+                    new Constraints\NotBlank([
+                        'message' => 'Please enter price',
```
simara-svatopluk: Please distinguish error messages by the field. Eg. "Please enter unit price excluding VAT"

packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
```diff
@@ -53,6 +69,18 @@ public function configureOptions(OptionsResolver $resolver)
             ->setDefaults([
                 'data_class' => OrderItemData::class,
                 'attr' => ['novalidate' => 'novalidate'],
+                'validation_groups' => function (FormInterface $form) {
```
simara-svatopluk: I hear often about violation the DRY principle from colleagues... So here, you are violating the DRY principle. I personally don't care in this case, but the team may ;-)

simara-svatopluk: The block of code is the same in 3 form types

packages/framework/src/Form/Transformers/CopyTotalPricesOfOrderItemTransformer.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData|null $value
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData|null
+     */
+    public function transform($value): ?OrderItemData
+    {
+        if ($value === null) {
+            return null;
+        }
+
+        if ($value instanceof OrderItemData) {
```
simara-svatopluk: This block could be simplified to more readable:

```php
if (!$value instanceof OrderItemData) {
    throw new TransformationFailedException(...);
}

if ($value->quantity !== 1) {
    throw new TransformationFailedException(sprintf('A single item must provided for total prices to be copiable, quantity of %d provided.', $value->quantity));
}

$value->totalPriceWithVat = $value->priceWithVat;
$value->totalPriceWithoutVat = $value->priceWithoutVat;

return $value;
```
the advantage is less nested IFs

packages/framework/src/Migrations/Version20190611114955.php
```diff
@@ -0,0 +1,86 @@
...
+        );
+        $incorrectOrderCount = $statement->rowCount();
+
+        if ($incorrectOrderCount > 0) {
+            $message = sprintf('There are %d orders in your DB where the order item total prices do not add up to the total price of the order:', $incorrectOrderCount);
+
+            for ($i = 0; $i < min($incorrectOrderCount, static::MAX_LISTED_ORDERS); $i++) {
+                $incorrectOrder = $statement->fetch(PDO::FETCH_NUM);
+
+                $message .= sprintf("\n  - ID %d: order total %s, sum %s (with VAT); order total %s, sum %s (without VAT)", ...$incorrectOrder);
```
simara-svatopluk: this is pretty wild usage, I would prefer naming fields, eg. `$incorrectOrder['id']`

packages/framework/src/Model/Order/Item/OrderItem.php
```diff
@@ -192,6 +212,42 @@ public function getPriceWithVat(): Money
...
+        $this->totalPriceWithVat = $totalPrice !== null ? $totalPrice->getPriceWithVat() : null;
+        $this->totalPriceWithoutVat = $totalPrice !== null ? $totalPrice->getPriceWithoutVat() : null;
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasForcedTotalPrice(): bool
+    {
+        return $this->totalPriceWithoutVat !== null && $this->totalPriceWithVat !== null;
```
simara-svatopluk: there is an unhanded invalid state: One price is not null and one is null. In such case the class will behave as the prices are not forced. But it should throw an exception in my opinion as this case is simply invalid

packages/framework/src/Model/Order/Item/OrderItem.php
```diff
@@ -192,6 +212,42 @@ public function getPriceWithVat(): Money
...
+
+    /**
+     * The total price property can be used when order item has prices that differ from current price calculation implementation.
+     * Otherwise it should be set to NULL (which means it will be calculated automatically).
+     *
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Price|null $totalPrice
+     */
+    public function setTotalPrice(?Price $totalPrice): void
+    {
+        $this->totalPriceWithVat = $totalPrice !== null ? $totalPrice->getPriceWithVat() : null;
```
simara-svatopluk: this method doesn't set prices that are in the argument... it's a mistake in my opinion. If the behavior is correct, then the function is wrongly named

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
```diff
@@ -4,6 +4,20 @@
 
 class OrderItemDataFactory implements OrderItemDataFactoryInterface
 {
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemPriceCalculation|null
+     */
+    protected $orderItemPriceCalculation;
+
+    /**
+     * @required
```
simara-svatopluk: please add note that this will be changed to constructor injection in next major version (already done in the SSFW)

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
```diff
@@ -52,4 +73,30 @@ protected function addFieldsByOrderItemType(OrderItemData $orderItemData, OrderI
...
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData $orderItemData
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItem $orderItem
+     * @return bool
+     */
+    protected function canUsePriceCalculation(OrderItemData $orderItemData, OrderItem $orderItem): bool
+    {
+        if ($this->orderItemPriceCalculation === null) {
+            return false;
```
simara-svatopluk: shouldn't this throw an exception?

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
```diff
@@ -52,4 +73,30 @@ protected function addFieldsByOrderItemType(OrderItemData $orderItemData, OrderI
...
+        }
+
+        $calculatedPriceWithoutVat = $this->orderItemPriceCalculation->calculatePriceWithoutVat($orderItemData);
```
simara-svatopluk: I don't understand why is the function so complicated. Shouldn't it just return `return !$orderItem->hasForcedTotalPrice()`?

packages/framework/src/Model/Order/Order.php
```diff
@@ -1028,13 +1028,34 @@ public function edit(
...
         $this->calculateTotalPrice($orderPriceCalculation);
 
         return new OrderEditResult($statusChanged);
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData $orderItemData
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemPriceCalculation $orderItemPriceCalculation
+     */
+    protected function editOrderItemDataPrices(OrderItemData $orderItemData, OrderItemPriceCalculation $orderItemPriceCalculation): void
```
simara-svatopluk: This method invoked me something completely different... I thought it is about editing OrderItemPrices, but it's about data.... Please rename it to something like `calculateOrderItemDataPrices`

packages/framework/src/Model/Pricing/PriceCalculation.php
```diff
@@ -9,6 +9,8 @@
 
 class PriceCalculation
 {
+    protected const PRICE_CALCULATION_MAX_SCALE = 12;
```
simara-svatopluk: I'm wondering... why 12? please explain in the commit message

packages/framework/src/Form/Admin/Order/OrderItemFormType.php
PetrHeinz: Originally, I didn't want to add many translation messages, but I guess it's worth it. Addressed in f85fd7c4d

packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
PetrHeinz: You're right about the DRY violation, but resolving validation groups is always done in an inlined function like this, so I was just trying to be consistent. I guess I could extract it into a public static function in `OrderItemFormType` - do you think it would be better?

packages/framework/src/Form/Transformers/CopyTotalPricesOfOrderItemTransformer.php
PetrHeinz: You're right about that, fixed in e3e298f20.

Also, I just found out that `instanceof` has a [higher operator precendence](https://www.php.net/manual/en/language.operators.precedence.php) than negation... I always wrote `if (!($object instanceof ClassName)) {` and I was annoyed by how ugly it looks. Learning something every day :smile: Thanks!

packages/framework/src/Migrations/Version20190611114955.php
PetrHeinz: I was relieved when I got this shorter this way, because otherwise I would have to explicitly list five arguments (with an array index) here and also alias the selected columns.

I guess I see this as more friendly and less error-prone this way as the actual selected data are used exclusively in this `sprintf` call. I actually had the fields named during development but I made a typo in one of the aliases (the `_with_vat` / `_without_vat` postfixes can get mixed up when you see them for four hours straight...), which led to a embarrassingly long debugging.

packages/framework/src/Model/Order/Item/OrderItem.php
PetrHeinz: You're right about the invalid state. I let it slide as the state is encapsulated in the entity - the properties can be set only both at once in the `setTotalPrice` method. But I guess as long as the properties are only `protected` and the properties are set using reflection by Doctrine, a bit of defensive programming may be appropriate.

Addressed in 28a65a5a4 using a custom `RuntimeException`. Please note, that we always allow for passing the `$previous` exception as an constructor argument, even if it's unused. I'm keeping this up for consistency.

packages/framework/src/Model/Order/Item/OrderItem.php
PetrHeinz: I don't get your point. This setter serves for the purpose of setting either both total prices at the same time, or unsetting them (saying the total price should be set to `null`). Seeing that `Price` is just a value object for holding two monetary values (a with VAT / without VAT price pair) I don't think it could be misleading...

Also, it's consistent with `Order::setTotalPrice()`.

packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
simara-svatopluk: As I said "I personally don't core in this case". So it's up to you. I only wanted to point it out. Maybe you can explain your decision in the commit message

packages/framework/src/Form/Transformers/CopyTotalPricesOfOrderItemTransformer.php
simara-svatopluk: the has higher operator precedence.... it's like that since... ever (2004 I guess)

packages/framework/src/Migrations/Version20190611114955.php
simara-svatopluk: Simply I don't like this clever code approach...

It tights together how the SQL is written (order of columns in projection) and form of the error message.

But if you are ok with it, I don't care much as this is just a once-runned tool. Resolve the conversation if you wish :-)

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
PetrHeinz: I don't think we ever decided that we want to change *all* setter injections to constructon injections in the next (or every) major release. But here it makes sense I guess, added in 15dc0a240.

packages/framework/src/Model/Order/Item/OrderItem.php
simara-svatopluk: My apologiess, I read it as `... ? $this->getPriceWithVat() : null`

no problem in this place

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
PetrHeinz: My original thinking was that it should just return `false` meaning it cannot decide whether it should use the price calculation. But this would definitely lead to unexpected behavior, especially because I use the property outside of this method as well. Addressed in 3a67de5d9

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
simara-svatopluk: Depends what do you imagine by "we ever decided". If you think we had a whole team meeting and we've written down conclusion, then no. If you think couple of developers (3: PIA, DEV, CR) agreed on this, then yes :bird:

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
PetrHeinz: I implemented the `OrderItemDataFactory::canUsePriceCalculation()` method so that it uses price calculation always if possible so that the administration is easier. If the total prices is "forced" to be the same value that would be actually calculated by price calculation, it will still suggest to use the price calculation.

Only when the prices are inconsistent with the price calculation implementation it will set the `OrderItemData::$usePriceCalculation` to `false`.

Considering that I have to specify the wnated behavior to you I guess I should write a test for it, so that these expectations are explicit, right?

packages/framework/src/Model/Pricing/PriceCalculation.php
PetrHeinz: Well, you're right to be wondering. I've changed it to 6 in bf73ba5e2, and I'll add the reason (consistency with the scale of persisted prices of products) to the commit message during rebase.

packages/framework/src/Model/Order/Item/OrderItemPriceCalculation.php
```diff
@@ -57,6 +57,19 @@ public function calculatePriceWithoutVat(OrderItemData $orderItemData): Money
         return $orderItemData->priceWithVat->subtract($vatAmount);
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItem $orderItem
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\Price
+     */
+    public function getTotalPrice(OrderItem $orderItem): Price
```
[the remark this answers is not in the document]
PetrHeinz: Here my reasoning was: either you want to explicitly *calculate* the price, so you use the `calculateTotalPrice` or you just want to *get* the total price consistently with the application logic (which means it can be forced, or is recalculated on-the-fly as it was previously). Method `calculateTotalPriceIfNotForced` doesn't really sound to me as a method that should be used in majority of use-cases, it sounds more like an edge case...

packages/framework/src/Form/Admin/Order/OrderPaymentFormType.php
PetrHeinz: Well, there was quite a lot of duplicated code because of this, I did the extraction in 6f8c26436

packages/framework/src/Model/Order/Order.php
PetrHeinz: Renamed in 2fcd799d0

packages/framework/src/Model/Order/Item/OrderItemPriceCalculation.php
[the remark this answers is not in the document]
simara-svatopluk: ok, I wanted to give a suggestion.
`getTotalPrice` doesn't express it might calculate the price and also might return already calculated price. When should the implementator use `calculateTotalPrice` and when `getTotalPrice`? This naming doesn't help

packages/framework/src/Model/Order/Item/OrderItemDataFactory.php
simara-svatopluk: Let's assume following case:
* I open an order in administration
* uncheck "use price calculation" on order Item XXX
* save
* 2 total prices will be saved together with XXX
* I reopen the same order
* the "use price calculation" is unchecked (from user-point of view WHY?!)
* save
* 2 total prices are removed from XXX

Am I correct?
I don't assume this to be either user-friendly, reasonable scenario

PetrHeinz: Yes, this would be the behavior. I don't really know whether it is acceptable or user-friendly.

simara-svatopluk: It's doing "something" when I didn't change anything. It isn't reasonable for me. Please discuss it with our Product Owner

LukasHeinz: After discussion with @simara-svatopluk  this behavior is looking strange to me. Please can you describe use-case that create need for such behavior?
In my point of view, after migrating to this version, I would set all existing order items to "do not calculate price" setting. But new ones will use calculation. It is clear, consistent and easily explainable.

PetrHeinz: I've changed the behavior in 3cf8acd96

packages/framework/src/Model/Order/Item/OrderItemPriceCalculation.php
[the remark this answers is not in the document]
TomasLudvik: I think we should keep only calculateTotalPrice and it should do exactly what getTotalPrice does now. As @simara-svatopluk mentioned - the naming right now does not help in any way and therefore I would prefer to keep it as it was.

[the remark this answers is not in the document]
PetrHeinz: Oh... maybe it makes more sense now that https://github.com/shopsys/shopsys/pull/1129/commits/3cf8acd966cc0e8ec794f7c3aab9b2aa8dbfc9a6 was merged.

It would definitely make the change much much simpler.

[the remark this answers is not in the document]
PetrHeinz: changed in 5865c939e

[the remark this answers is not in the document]
PetrHeinz: I would like to squash some commits together as there are a lot of changes that revert of modify something built in previous commits... But I'd like to do that only after the changes are code-reviewed.

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -1090,6 +1090,9 @@ msgstr "IP adresa"
 msgid "Icon"
 msgstr "Ikona"
 
+msgid "If checked, all prices have to be handled manually. Otherwise, unit price without VAT and the total prices for that item will be recalculated automatically."
+msgstr "Pokud je zaškrtnuto, všechny ceny je třeba nastavovat ručně. Jinak se jednotková cena bez DPH a celkové ceny pro tuto položku přepočítají automaticky."
```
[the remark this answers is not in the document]
PetrHeinz: Fixed in aa1b4b7

LukasHeinz:
WIP testing report: 
the new price calculation and possibility to manually edit prices in orders are working fine. Great job. 

Now I move to migrations....

LukasHeinz:
All test passed. Great Work. 
Please update description of PR and add information what really changes - only calculation of displayed prices without VAT. There is no change in product prices with VAT or in previous orders.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -55,6 +55,26 @@ There you can find links to upgrade notes for other versions too.
         ```
 - update your tests to use interfaces of factories fetched from dependency injection container
     -  update tests same way as in PR ([#970](https://github.com/shopsys/shopsys/pull/970/files))
+- check your VAT calculations after it was modified in `shopsys/framework` ([#1129](https://github.com/shopsys/shopsys/pull/1129))
+    - see [the description of the PR](https://github.com/shopsys/shopsys/pull/1129) to understand the scope of this change
+    - run DB migrations on your demo data and on a copy of production database (if possible) to see that it works with your data well
```
LukasHeinz: It sounds like we are not sure if it works. Please rephrase. 

I would recommend starting this paragraph with "if you modified the price calculation or you altered the prices in the database directly"

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -55,6 +55,26 @@ There you can find links to upgrade notes for other versions too.
         ```
 - update your tests to use interfaces of factories fetched from dependency injection container
     -  update tests same way as in PR ([#970](https://github.com/shopsys/shopsys/pull/970/files))
+- check your VAT calculations after it was modified in `shopsys/framework` ([#1129](https://github.com/shopsys/shopsys/pull/1129))
+    - see [the description of the PR](https://github.com/shopsys/shopsys/pull/1129) to understand the scope of this change
```
LukasHeinz:
```suggestion
    - we strongly recommend seeing [the description of the PR](https://github.com/shopsys/shopsys/pull/1129) to understand the scope of this change
```

LukasHeinz: We offer the link to PR in every upgrade instruction. I would like to highlight it here because they really should see it.

PR #1131, opened 2019-06-17
[shopsys] improve backward compatibility of ProductAvailabilityCalculation patch

PetrHeinz:
- the patch in https://github.com/shopsys/shopsys/pull/1113 requried different EntityManager mock creation
- strict comparison of the result of `EntityManager::contains()` with `false` instead of a simple negation removes this requirement (because the mocked method will return `null`)
- tests require no changes to pass successfully after upgrade to a new version of `shopsys/framework` package
- the changes in `ProductAvailabilityCalculationTest` were reverted

| Q             | A
| ------------- | ---
|Description, reason for the PR| All tests should pass in `shopsys/project-base` in `v7.0.0` when upgrading all used `shopsys/*` packages via Composer. This was not true after merging #1113 as the functional tests required minor changes to pass.
|New feature| No 
|Fixes issues| ... 


boris-brtan:
travis, no-bc-break-check and other jobs passed.

PR #1133, opened 2019-06-17
[shopsys] added Elasticsearch structure migration via reindexing for easier deployment

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| this PR lowers the downtime of eshop when deploying by reindexing to a new index elasticsearch that is created only if the structure has changed
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/docs/installation/installation-using-docker-on-production-server.md
+++ b/docs/installation/installation-using-docker-on-production-server.md
@@ -348,6 +348,9 @@ docker-compose -p production up -d --force-recreate webserver
 # remove container for building the application image
 docker rm -f build-php-fpm-container
 ```
+***Note:** During `build-deploy-part-2-db-dependent` phing target `product-search-migrate-structure` is called and can cause error when you change the type of field to another (eg. you change it from `bool` to `integer`). If you need to make this change, please add new field with the correct type and delete the old field instead*
+
+***Note:** If you need to have freshly exported products in Elasticsearch after deploy, you can call phing target `product-search-export-products` during `build-deploy-part-2-db-dependent`.*
 
 ## Logging
 If you need to inspect your application logs, use `docker-compose logs` command.
--- a/docs/introduction/console-commands-for-application-management-phing-targets.md
+++ b/docs/introduction/console-commands-for-application-management-phing-targets.md
@@ -140,6 +140,20 @@ Consists of two subtasks that can be run independently:
 * `product-search-delete-structure` - deletes existing indexes structure
 * `product-search-create-structure` - creates new indexes structure by json definitions stored in [the resources directory](/project-base/src/Shopsys/ShopBundle/Resources/definition).
 
+#### product-search-migrate-structure
+Migrates Elasticsearch indexes if there is change between currently used structure and the one in `*.json`.
+Especially useful when you need to change the structure and don't need to have fresh data in Elasticsearch
+* creates new index without alias
+* reindexes data from old index to the new one
+* deletes old index
+* creates alias for the new index
+
+*Note: If you add field/s to the structure and reindex they won't be available until `product-search-export-products` is called.*
+*You application must handle the properties not being filled correctly until all products are exported.*
+
+*Note: Using this phing target after changing the type of field to another in structure (eg. changing it from `bool` to `integer`) will cause an error.*
+*If you need to make this change, please add new field with the correct type and delete the old field instead.*
+
 #### product-search-export-products
 Exports all visible products to Elasticsearch.
 
--- a/docs/introduction/console-commands-for-application-management-phing-targets.md
+++ b/docs/introduction/console-commands-for-application-management-phing-targets.md
@@ -149,7 +149,7 @@ Especially useful when you need to change the structure and don't need to have f
 * creates alias for the new index
 
 *Note: If you add field/s to the structure and reindex they won't be available until `product-search-export-products` is called.*
-*You application must handle the properties not being filled correctly until all products are exported.*
+*Your application must handle the properties not being filled correctly until all products are exported.*
 
 *Note: Using this phing target after changing the type of field to another in structure (eg. changing it from `bool` to `integer`) will cause an error.*
 *If you need to make this change, please add new field with the correct type and delete the old field instead.*
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -108,6 +108,25 @@ There you can find links to upgrade notes for other versions too.
     - modify the functional test `\Tests\ShopBundle\Functional\Component\Redis\RedisFacadeTest` so it creates `RedisFacade` using the two arrays and add a new test case `testNotCleaningPersistentClient`
         - you can copy-paste the [`RedisFacadeTest`](https://github.com/shopsys/project-base/blob/master/tests/ShopBundle/Functional/Component/Redis/RedisFacadeTest.php) from `shopsys/project-base`
 - implement `createFromIdAndName(int $id, string $name): FriendlyUrlData` method in your implementations of `FriendlyUrlDataFactoryInterface` as the method will be added to the interface in `v8.0.0` version ([#948](https://github.com/shopsys/shopsys/pull/948))
+- use aliases of index and build version in index name in Elasticsearch for better usage when deploying ([#1133](https://github.com/shopsys/shopsys/pull/1133))
+    - use method `ElasticsearchStructureManager::getCurrentIndexName` or `ElasticsearchStructureManager::getAliasName` instead of `ElasticsearchStructureManager::getIndexName` when calling a query to elasticSearch in order to always target the right index
+        - use `getAliasName` if you need to access the index for read operations (eg. searching, filtering)
+        - use `getCurrentIndexName` if need to write to the index or manipulate it (eg. product export)
+    - run `php phing product-search-recreate-structure` to generate new indexes with aliases
+    - use method `ElasticsearchStructureManager::deleteCurrentIndex` instead of `ElasticsearchStructureManager::deleteIndex` as it was deprecated
+    - if you have extended `ElasticsearchStructureManager` in `services.yml` you'll need to send the `build-version` parameter to the 4th argument of the constructor or call `setBuildVersion` setter injector like this:
+        ```diff
+          Shopsys\FrameworkBundle\Component\Elasticsearch\ElasticsearchStructureManager:
+              arguments:
+                  - '%shopsys.elasticsearch.structure_dir%'
+                  - '%env(ELASTIC_SEARCH_INDEX_PREFIX)%'
+        +     calls:
+        +         - method: setBuildVersion
+        +           arguments:
+        +               - '%build-version%'
+        ```
+    - copy a new functional test [`ElasticsearchStructureUpdateCheckerTest`](https://github.com/shopsys/project-base/blob/master/tests/ShopBundle/Functional/Component/Elasticsearch/ElasticsearchStructureUpdateCheckerTest.php) into `tests/ShopBundle/Functional/Component/Elasticsearch/` in your project
+        - this test will ensure that the check whether to update Elasticsearch structure works as intended
 
 ### Configuration
 - update `phpstan.neon` with following change to skip phpstan error ([#1086](https://github.com/shopsys/shopsys/pull/1086))
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -84,11 +84,11 @@
 
     <target name="build-deploy-part-1-db-independent" depends="build-version-generate,clean,composer-prod,npm,dirs-create,assets" description="First part of application build for production preserving your DB (can be run without maintenance page)."/>
 
-    <target name="build-deploy-part-2-db-dependent" depends="redis-check,db-migrations,domains-data-create,friendly-urls-generate,domains-urls-replace,grunt,error-pages-generate,warmup,clean-redis-old" description="Second part of application build for production preserving your DB (must be run with maintenance page when containing DB migrations)."/>
+    <target name="build-deploy-part-2-db-dependent" depends="product-search-migrate-structure,redis-check,db-migrations,domains-data-create,friendly-urls-generate,domains-urls-replace,grunt,error-pages-generate,warmup,clean-redis-old" description="Second part of application build for production preserving your DB (must be run with maintenance page when containing DB migrations)."/>
 
     <target name="build-dev" depends="build-version-generate,clean,composer-dev,npm,timezones-check,dirs-create,test-dirs-create,assets,db-migrations,domains-data-create,friendly-urls-generate,domains-urls-replace,product-search-recreate-structure,product-search-export-products,grunt,error-pages-generate,tests-acceptance-build,clean-redis-old,checks-diff" description="Builds application for development preserving your DB and runs checks on changed files."/>
 
-    <target name="build-dev-quick" depends="build-version-generate,clean,composer-dev,npm,dirs-create,test-dirs-create,assets,db-migrations,domains-data-create,friendly-urls-generate,domains-urls-replace,grunt,clean-redis-old" description="Builds application for development preserving your DB while skipping nonessential steps."/>
+    <target name="build-dev-quick" depends="build-version-generate,clean,composer-dev,npm,dirs-create,test-dirs-create,assets,db-migrations,domains-data-create,friendly-urls-generate,domains-urls-replace,product-search-migrate-structure,grunt,clean-redis-old" description="Builds application for development preserving your DB while skipping nonessential steps."/>
 
     <target name="build-new" depends="wipe,build-version-generate,composer-prod,redis-check,npm,dirs-create,assets,db-rebuild,grunt,error-pages-generate,warmup,product-search-recreate-structure,clean-redis-old" description="Builds application for production with clean DB (with base data only)."/>
 
@@ -688,6 +688,13 @@
         <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
             <arg value="${path.bin-console}"/>
             <arg value="shopsys:product-search:export-products"/>
+        </exec>
+    </target>
+
+    <target name="product-search-migrate-structure" description="Creates new structure, reindexes it from old one, deletes old structure and add alias to new structure.">
+        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+            <arg value="${path.bin-console}"/>
+            <arg value="shopsys:product-search:migrate-if-necessary"/>
         </exec>
     </target>
 
--- a/packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
+++ b/packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
@@ -2,7 +2,10 @@
 
 namespace Shopsys\FrameworkBundle\Component\Elasticsearch;
 
+use BadMethodCallException;
 use Elasticsearch\Client;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\Exception\ElasticsearchMoreThanOneCurrentIndexException;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\Exception\ElasticsearchNoCurrentIndexException;
 use Shopsys\FrameworkBundle\Component\Elasticsearch\Exception\ElasticsearchStructureException;
 
 class ElasticsearchStructureManager
@@ -23,15 +26,50 @@ class ElasticsearchStructureManager
     protected $client;
 
     /**
+     * @var string|null
+     */
+    protected $buildVersion;
+
+    /**
      * @param string $definitionDirectory
      * @param string $indexPrefix
      * @param \Elasticsearch\Client $client
+     * @param string|null $buildVersion
      */
-    public function __construct(string $definitionDirectory, string $indexPrefix, Client $client)
+    public function __construct(string $definitionDirectory, string $indexPrefix, Client $client, ?string $buildVersion = null)
     {
         $this->definitionDirectory = $definitionDirectory;
         $this->indexPrefix = $indexPrefix;
         $this->client = $client;
+        $this->buildVersion = $buildVersion;
+    }
+
+    /**
+     * @param string $buildVersion
+     * @internal Will be replaced with constructor injection in the next major release
+     */
+    public function setBuildVersion(string $buildVersion): void
+    {
+        if ($this->buildVersion !== null) {
+            throw new BadMethodCallException(sprintf('Method "%s" has been already called and cannot be called multiple times.', __METHOD__));
+        }
+
+        $this->buildVersion = $buildVersion;
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return string
+     * @deprecated Getting index name without a version using this method is deprecated since SSFW 7.3, use getVersionedIndexName or getAliasName instead
+     */
+    public function getIndexName(int $domainId, string $index): string
+    {
+        @trigger_error(
+            sprintf('Getting index name without a version using method "%s" is deprecated since SSFW 7.3, use %s or %s instead', __METHOD__, 'getVersionedIndexName', 'getAliasName'),
+            E_USER_DEPRECATED
+        );
+        return $this->getAliasName($domainId, $index);
     }
 
     /**
@@ -39,7 +77,69 @@ class ElasticsearchStructureManager
      * @param string $index
      * @return string
      */
-    public function getIndexName(int $domainId, string $index): string
+    public function getVersionedIndexName(int $domainId, string $index): string
+    {
+        return $this->indexPrefix . ($this->buildVersion ?? 'null') . $index . $domainId;
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return string
+     */
+    public function getCurrentIndexName(int $domainId, string $index): string
+    {
+        $aliasName = $this->getAliasName($domainId, $index);
+
+        $indexes = $this->client->indices();
+        if (!$indexes->existsAlias(['name' => $aliasName]) && $indexes->exists(['index' => $aliasName])) {
+            @trigger_error(
+                sprintf('Passing index name instead of alias name to method %s is deprecated since SSFW 7.3', __METHOD__),
+                E_USER_DEPRECATED
+            );
+            return $aliasName;
+        }
+
+        $indexNames = $this->getExistingIndexNamesForAlias($aliasName);
+
+        if (count($indexNames) > 1) {
+            throw new ElasticsearchMoreThanOneCurrentIndexException($aliasName);
+        } elseif (count($indexNames) === 0) {
+            throw new ElasticsearchNoCurrentIndexException($aliasName);
+        }
+
+        return reset($indexNames);
+    }
+
+    /**
+     * @param string $aliasName
+     * @return array
+     */
+    protected function getExistingIndexNamesForAlias(string $aliasName): array
+    {
+        $existingIndexNames = [];
+        $indexes = $this->client->indices();
+
+        if ($indexes->existsAlias(['name' => $aliasName])) {
+            $aliases = $indexes->getAlias([
+                'name' => $aliasName,
+            ]);
+            foreach (array_keys($aliases) as $indexName) {
+                if ($indexes->exists(['index' => $indexName])) {
+                    $existingIndexNames[] = $indexName;
+                }
+            }
+        }
+
+        return $existingIndexNames;
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return string
+     */
+    public function getAliasName(int $domainId, string $index): string
     {
         return $this->indexPrefix . $index . $domainId;
     }
@@ -50,11 +150,17 @@ class ElasticsearchStructureManager
      */
     public function createIndex(int $domainId, string $index)
     {
-        $definition = $this->getDefinition($domainId, $index);
+        $definition = $this->getStructureDefinition($domainId, $index);
         $indexes = $this->client->indices();
-        $indexName = $this->getIndexName($domainId, $index);
+        $indexName = $this->getVersionedIndexName($domainId, $index);
         if ($indexes->exists(['index' => $indexName])) {
-            throw new ElasticsearchStructureException(sprintf('Index %s already exists', $indexName));
+            $message = sprintf('Index %s already exists', $indexName);
+            if ($this->buildVersion === '0000000000000000' || $this->buildVersion === null) {
+                $message .= ' Please start using build version or delete current index before creating a new one.';
+            } else {
+                $message .= ' Maybe you forgot to execute "php phing generate-build-version" before creating new index?';
+            }
+            throw new ElasticsearchStructureException($message);
         }
 
         $indexes->create([
@@ -67,8 +173,26 @@ class ElasticsearchStructureManager
      * @param int $domainId
      * @param string $index
      */
+    public function migrateIndex(int $domainId, string $index)
+    {
+        $this->createIndex($domainId, $index);
+        $this->reindexFromCurrentIndexToNewIndex($domainId, $index);
+        $this->deleteCurrentIndex($domainId, $index);
+        $this->createAliasForIndex($domainId, $index);
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @deprecated Deleting index using this method is deprecated since SSFW 7.3, use deleteIndexesOfCurrentAlias instead
+     */
     public function deleteIndex(int $domainId, string $index)
     {
+        @trigger_error(
+            sprintf('Deleting index using method "%s" is deprecated since SSFW 7.3, use %s instead', __METHOD__, 'deleteIndexesOfCurrentAlias'),
+            E_USER_DEPRECATED
+        );
+
         $indexes = $this->client->indices();
         $indexName = $this->getIndexName($domainId, $index);
         if ($indexes->exists(['index' => $indexName])) {
@@ -79,9 +203,78 @@ class ElasticsearchStructureManager
     /**
      * @param int $domainId
      * @param string $index
+     */
+    protected function reindexFromCurrentIndexToNewIndex(int $domainId, string $index)
+    {
+        $indexes = $this->client->indices();
+        $indexName = $this->getVersionedIndexName($domainId, $index);
+        $currentIndexName = $this->getCurrentIndexName($domainId, $index);
+        if ($indexes->exists(['index' => $currentIndexName]) && $indexes->exists(['index' => $indexName])) {
+            $body = [
+                'source' => [
+                    'index' => $currentIndexName,
+                ],
+                'dest' => [
+                    'index' => $indexName,
+                ],
+            ];
+
+            $this->client->reindex([
+                'body' => $body,
+            ]);
+        }
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     */
+    public function createAliasForIndex(int $domainId, string $index)
+    {
+        $indexes = $this->client->indices();
+        $aliasName = $this->getAliasName($domainId, $index);
+        $indexName = $this->getVersionedIndexName($domainId, $index);
+
+        if (!$indexes->existsAlias(['name' => $aliasName, 'index' => $indexName]) && $indexes->exists(['index' => $indexName])) {
+            $indexes->putAlias([
+                'index' => $indexName,
+                'name' => $aliasName,
+            ]);
+        }
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     */
+    public function deleteCurrentIndex(int $domainId, string $index): void
+    {
+        try {
+            $indexName = $this->getCurrentIndexName($domainId, $index);
+        } catch (ElasticsearchNoCurrentIndexException $e) {
+            return;
+        }
+
+        $this->client->indices()->delete(['index' => $indexName]);
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
      * @return array
+     * @deprecated using method getDefinition is deprecated since SSFW 7.3. Use public method getStructureDefinition instead
      */
     protected function getDefinition(int $domainId, string $index): array
+    {
+        $this->getStructureDefinition($domainId, $index);
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return array
+     */
+    public function getStructureDefinition(int $domainId, string $index): array
     {
         $file = sprintf('%s/%s/%s.json', $this->definitionDirectory, $index, $domainId);
         if (!is_file($file)) {
--- a/packages/framework/src/Component/Elasticsearch/ElasticsearchStructureUpdateChecker.php
+++ b/packages/framework/src/Component/Elasticsearch/ElasticsearchStructureUpdateChecker.php
@@ -1,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Elasticsearch;
+
+use Elasticsearch\Client;
+use Shopsys\FrameworkBundle\Component\ArrayUtils\RecursiveArraySorter;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\Exception\ElasticsearchNoCurrentIndexException;
+
+class ElasticsearchStructureUpdateChecker
+{
+    /**
+     * @var \Elasticsearch\Client
+     */
+    protected $client;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Elasticsearch\ElasticsearchStructureManager
+     */
+    protected $elasticsearchStructureManager;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\ArrayUtils\RecursiveArraySorter
+     */
+    protected $recursiveArraySorter;
+
+    /**
+     * @param \Elasticsearch\Client $client
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\ElasticsearchStructureManager $elasticsearchStructureManager
+     * @param \Shopsys\FrameworkBundle\Component\ArrayUtils\RecursiveArraySorter $recursiveArraySorter
+     */
+    public function __construct(
+        Client $client,
+        ElasticsearchStructureManager $elasticsearchStructureManager,
+        RecursiveArraySorter $recursiveArraySorter
+    ) {
+        $this->client = $client;
+        $this->elasticsearchStructureManager = $elasticsearchStructureManager;
+        $this->recursiveArraySorter = $recursiveArraySorter;
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return bool
+     */
+    public function isNecessaryToUpdateStructure(int $domainId, string $index): bool
+    {
+        $definition = $this->elasticsearchStructureManager->getStructureDefinition($domainId, $index);
+
+        try {
+            $indexName = $this->elasticsearchStructureManager->getCurrentIndexName($domainId, $index);
+        } catch (ElasticsearchNoCurrentIndexException $e) {
+            return true;
+        }
+
+        $currentIndex = $this->client->indices()->get(['index' => $indexName]);
+
+        $currentIndexDefinition = $this->prepareDefinitionForComparing(reset($currentIndex));
+        $definition = $this->prepareDefinitionForComparing($definition);
+
+        return $currentIndexDefinition !== $definition;
+    }
+
+    /**
+     * @param array $definition
+     * @return array
+     */
+    protected function prepareDefinitionForComparing(array $definition): array
+    {
+        if (array_key_exists('analysis', $definition['settings'])) {
+            $definition['settings']['index']['analysis'] = $definition['settings']['analysis'];
+            unset($definition['settings']['analysis']);
+        }
+
+        unset(
+            $definition['aliases'],
+            $definition['settings']['index']['creation_date'],
+            $definition['settings']['index']['provided_name'],
+            $definition['settings']['index']['uuid'],
+            $definition['settings']['index']['version'],
+        );
+
+        $this->recursiveArraySorter->recursiveArrayKsort($definition);
+
+        return $this->castAllIntegersToStrings($definition);
+    }
+
+    /**
+     * @param array $structureDefinition
+     * @return array
+     */
+    protected function castAllIntegersToStrings(array $structureDefinition): array
+    {
+        return array_map(
+            function ($item) {
+                if (is_array($item)) {
+                    return $this->castAllIntegersToStrings($item);
+                }
+                return is_int($item) ? (string)$item : $item;
+            },
+            $structureDefinition
+        );
+    }
+}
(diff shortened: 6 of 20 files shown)


packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
```diff
@@ -42,6 +42,45 @@ public function createIndexes(OutputInterface $output)
...
+        }
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     */
+    public function createAliases(OutputInterface $output)
+    {
+        foreach ($this->domain->getAllIds() as $domainId) {
+            $output->writeln(sprintf('Creating alias for id %s', $domainId));
```
TomasLudvik: Foreach body violates DRY principle, because same code is in foreach of method createIndexesWithoutAliases - please create private method to avoid this.

packages/framework/build.xml
```diff
@@ -677,6 +677,20 @@
...
     </target>
 
+    <target name="product-search-create-structure-without-aliases" description="Creates structure without index aliases for searching via elasticsearch.">
+        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+            <arg value="${path.bin-console}"/>
+            <arg value="shopsys:product-search:create-structure-without-aliases"/>
+        </exec>
+    </target>
+
+    <target name="product-search-create-aliases" description="Creates index aliases for current elasticsearch index.">
```
TomasLudvik: I think that `product-search-create-structure` target should call this target too

packages/framework/build.xml
```diff
@@ -691,6 +705,20 @@
         </exec>
     </target>
 
+    <target name="product-search-export-products-if-migrations" description="Exports all products for searching via elasticsearch if there are migrations to be executed.">
```
TomasLudvik:
```suggestion
    <target name="product-search-export-products-if-new-migrations-exist" description="Exports all products for searching via elasticsearch if there are migrations to be executed.">
```

packages/framework/build.xml
```diff
@@ -691,6 +705,20 @@
...
+        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+            <arg value="${path.bin-console}"/>
+            <arg value="shopsys:product-search:export-products-migrations"/>
+        </exec>
+    </target>
+
+    <target name="product-search-export-products-if-no-migrations" description="Exports all products for searching via elasticsearch if there are no migrations to be executed.">
```
TomasLudvik:
```suggestion
    <target name="product-search-export-products-if-no-new-migrations-exist" description="Exports all products for searching via elasticsearch if there are no migrations to be executed.">
```

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -67,12 +84,55 @@ public function createIndex(int $domainId, string $index)
...
+            'index' => $indexName,
+            'name' => $aliasName,
+        ]);
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     */
+    public function deleteCurrentIndexOrIndexesOfAlias(int $domainId, string $index)
```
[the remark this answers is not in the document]
TomasLudvik: After discussion - first part should be replaced by some command "wipe-previous-elasticsearch-structure"

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -67,12 +84,55 @@ public function createIndex(int $domainId, string $index)
...
+            ]);
+
+            foreach (array_keys($alias) as $indexName) {
+                if ($indexes->exists(['index' => $indexName])) {
+                    $indexes->deleteAlias([
+                        'index' => $indexName,
+                        'name' => $aliasName,
+                    ]);
+
+                    $indexes->delete(['index' => $indexName]);
```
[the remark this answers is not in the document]
DavidOstrozlik: Yes, there can be only one alias per index. There cannot be two indexes for one alias so it cant end with an error

packages/framework/build.xml
DavidOstrozlik: I don't really like it. Maybe if `product-search-create-structure` called create without alias and then create alias but I think that it would just be more confusing

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -55,6 +55,20 @@ There you can find links to upgrade notes for other versions too.
         ```
 - update your tests to use interfaces of factories fetched from dependency injection container
     -  update tests same way as in PR ([#970](https://github.com/shopsys/shopsys/pull/970/files))
+- use aliases of index and build version in index name for better usage when deploying
```
simara-svatopluk: seeing this instructions leads me to a question: Is it a BC break?

packages/framework/build.xml
```diff
@@ -693,6 +693,13 @@
 
     <target name="product-search-recreate-structure" depends="product-search-delete-structure,product-search-create-structure" description="Recreates structure for searching via elasticsearch (deletes existing structure and creates new one)"/>
 
+    <target name="product-search-reindex-structure-if-necessary" description="Creates new structure, reindexes it from old one, deletes old structure and add alias to new structure.">
```
simara-svatopluk: naming is inconsistent, the phing target says "if necessary", the command says "reindex!" (for me means alwasy), please make them consistent (I prefer the "if necessary" variation)

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -34,21 +46,67 @@ public function __construct(string $definitionDirectory, string $indexPrefix, Cl
...
+
     /**
      * @param int $domainId
      * @param string $index
      * @return string
      */
     public function getIndexName(int $domainId, string $index): string
     {
-        return $this->indexPrefix . $index . $domainId;
+        return $this->indexPrefix . $this->buildVersion . $index . $domainId;
```
simara-svatopluk: you are using a different name of an index, it's like a different database name. It's a BC break in my opinion. The project simply won't work after `composer update`, am I right?

packages/framework/src/Migrations/Version20190617070314.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    use MultidomainMigrationTrait;
+
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        foreach ($this->getAllDomainIds() as $domainId) {
+            $this->sql('INSERT INTO setting_values (name, domain_id, value, type) VALUES
```
simara-svatopluk: So there is a new setting for each domain, and the value is an empty string `''`. So I update, and run migrations. And now when I call `ElasticStructureManger::getCurrentProductIndexName()`, It will return this empty string, am I right? That doesn't seem like a correct behavior

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -34,21 +46,67 @@ public function __construct(string $definitionDirectory, string $indexPrefix, Cl
...
+    }
+
+    /**
+     * @param int $domainId
+     * @return string
+     */
+    public function getCurrentProductIndexName(int $domainId)
```
simara-svatopluk: There are couple of methods to get an index name:
* `getIndexName`
* `getAliasName`
* `getCurrentProductIndexName`

When should I use what method? The naming doesn't help me unfortunately

simara-svatopluk: This is related with the new terminology: index name, alias name, current product index name, alias for current index, current alias, current index.
The terminology must be simple and consistent to be understandable. I cannot give you suggestion because I don't understand what methods even do

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -102,4 +210,76 @@ protected function getDefinition(int $domainId, string $index): array
...
+            }
+        }
+        return ksort($array);
+    }
+
+    /**
+     * @param array $currentIndex
+     * @return array
+     */
+    protected function prepareCurrentIndexDefinition(array $currentIndex): array
```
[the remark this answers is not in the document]
simara-svatopluk: me either

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -102,4 +210,76 @@ protected function getDefinition(int $domainId, string $index): array
...
+
+        $aliasName = $this->getAliasName($domainId, $index);
+        if ($indexes->existsAlias(['name' => $aliasName])) {
+            $aliases = $indexes->getAlias(['name' => $aliasName]);
+            foreach (array_keys($aliases) as $indexName) {
+                if ($indexes->exists(['index' => $indexName])) {
+                    $currentIndex = $indexes->get(['index' => $indexName]);
+                    $currentIndexDefinition = $this->prepareCurrentIndexDefinition($currentIndex);
+
+                    $this->recursiveArraySort($currentIndexDefinition);
```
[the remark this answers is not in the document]
DavidOstrozlik: You can't because `array_diff_assoc` only checks the first layer of array and not the other, more nested, layers

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -102,4 +242,104 @@ protected function getDefinition(int $domainId, string $index): array
...
+            $this->createAliasForCurrentIndex($domainId, $index);
+        }
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return bool
+     */
+    protected function isNecessaryToUpdateStructure(int $domainId, string $index)
```
simara-svatopluk: After all I have a strong feeling that the `ElasticStructureManager` is doing too much now and I don't understand what is it's responsibility. Maybe that's because it contains functionality that should be separated somewhere else

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -61,21 +119,103 @@ public function createIndex(int $domainId, string $index)
             'index' => $indexName,
             'body' => $definition,
         ]);
+        if ($setToSettings) {
+            $this->setting->setForDomain(Setting::ELASTICSEARCH_INDEX, $indexName, $domainId);
```
simara-svatopluk: This means when I `createIndex`, then this new index name will be set as  "the index" for the domain? This is quite strange functionality because it combines 2 things - create index and set the index as "the index" for the domain.

I suggest to create a new method that will be responsible for this setting, and I have a feeling that this method shouldn't be part of `ElasticsearchStructureManager`

docs/upgrade/UPGRADE-unreleased.md
simara-svatopluk: no

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
simara-svatopluk: no, it isn't BC break. You have to run a phing target for correct elastic data, but code is BC

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
simara-svatopluk: Even after discussions I suggest to remove this `setToSettings` from the function. You can create a new function 

```php
public function createIndexAndForDomain(int $domainId, string $index)
{
 $this->createIndex($domainId, $index);
 $indexName = $this->getIndexName($domainId, $index);
 $this->setting->setForDomain(Setting::ELASTICSEARCH_INDEX, $indexName, $domainId);
}
```

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -102,4 +242,104 @@ protected function getDefinition(int $domainId, string $index): array
...
+        return ksort($array);
+    }
+
+    /**
+     * @param array $currentIndex
+     * @return array
+     */
+    protected function prepareCurrentIndexDefinition(array $currentIndex): array
+    {
+        $currentIndexDefinition = reset($currentIndex);
```
simara-svatopluk: please pass into the method `$currentIndexDefinition` because the `$currentIndex` is not used in the function

`prepareCurrentIndexDefinition(reset($currentIndex))`

simara-svatopluk: Then you can use just `$indexDefinition` variable name instead of `$currentIndexDefinition`

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -102,4 +242,104 @@ protected function getDefinition(int $domainId, string $index): array
...
+        $currentIndexDefinition['settings'] = $currentIndexDefinition['settings']['index'];
+        $currentIndexDefinition['settings']['index'] = [
+            'number_of_replicas' => (int)$currentIndexDefinition['settings']['number_of_replicas'],
```
simara-svatopluk: So this means elastic returns `string` and when string is provided, then it doesn't accept it? Strange behavior

simara-svatopluk: please come with a better name for this function as this is not a preparation.... it's preparation for comparing... or something like that

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
simara-svatopluk:
```suggestion
    protected function isNecessaryToUpdateStructure(int $domainId, string $index): bool
```

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -102,4 +242,104 @@ protected function getDefinition(int $domainId, string $index): array
...
+                    if ($currentIndexDefinition !== $definition) {
+                        return true;
+                    }
+                }
+            }
+        } else {
+            return true;
+        }
+
+        return false;
```
simara-svatopluk: this line will never executed as there is `else` 4 lines above. I recommend to remove the `else` statement completely and keep here only `return true`

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -102,4 +242,104 @@ protected function getDefinition(int $domainId, string $index): array
...
+     * @param string $index
+     * @return bool
+     */
+    protected function isNecessaryToUpdateStructure(int $domainId, string $index)
+    {
+        $definition = $this->getDefinition($domainId, $index);
+        $indexes = $this->client->indices();
+
+        $aliasName = $this->getAliasName($domainId, $index);
+        if ($indexes->existsAlias(['name' => $aliasName])) {
```
simara-svatopluk: It's difficult to get what is going on in this deep function... Could you extract the `if, foreach, if` into a function `getExistingIndexesForAlias` (name is up to you) that will be used here
```php
foreach ($this->getExistingIndexesForAlias($aliasName))
{
 $currentIndex = $indexes->get(['index' => $indexName]);....
```

packages/framework/src/Model/Product/Search/Export/ProductSearchExportWithFilterRepository.php
```diff
@@ -65,7 +65,7 @@ public function getProductsData(int $domainId, string $locale, int $startFrom, i
                 'ean' => $product->getEan(),
                 'name' => $product->getName($locale),
                 'description' => $product->getDescription($domainId),
-                'shortDescription' => $product->getShortDescription($domainId),
+                'short_description' => $product->getShortDescription($domainId),
```
simara-svatopluk: This is a bugfix? I would appriciate if the change is separated into a standalone commit that describes reason for the change (together with changes in tests, upgrade instructions)

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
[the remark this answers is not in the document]
henzigo: That method should check multidimensional array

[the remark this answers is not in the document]
simara-svatopluk: Just write a test (one of you) and try to change the implementation... If it passes, no disscussion

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
DavidOstrozlik: I changed the name. The thing is that we have the information in *.json as integer and Elastic saves it to itself as a string. So to compare it strictly it necessary to retype it to int or change the way we have it in *.json

packages/framework/src/Model/Product/Search/Export/ProductSearchExportWithFilterRepository.php
DavidOstrozlik: This is kind of an bug fix that I need to fix in order to compare existing structure with the one in *.json. Do you really think that its necessary to move it to separate commit?

simara-svatopluk: if it is a bugfix, then definitely yes. The commit will be atomic and fulfill our https://github.com/shopsys/shopsys/blob/master/docs/contributing/guidelines-for-creating-commits.md#what-makes-a-good-commit

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureUpdateChecker.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        return false;
+    }
+
+    /**
+     * @param array $array
+     * @return bool
+     */
+    protected function recursiveArraySort(array &$array)
```
henzigo: again this should not be here

simara-svatopluk:
Please add `declare(strict_types=1);` to all php files that involves this PR

packages/framework/src/Model/Product/Search/Export/ProductSearchExportWithFilterRepository.php
DavidOstrozlik: done

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureUpdateChecker.php
DavidOstrozlik: done :)

packages/framework/src/Resources/config/services.yml
```diff
@@ -1006,6 +1006,10 @@ services:
         arguments:
             - '%shopsys.elasticsearch.structure_dir%'
             - '%env(ELASTIC_SEARCH_INDEX_PREFIX)%'
+        calls:
```
boris-brtan: is this backward compatible?
see the similar incompatibility after merge of https://github.com/shopsys/shopsys/pull/1158/files

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -67,21 +140,124 @@ public function createIndex(int $domainId, string $index)
...
+        return $existingIndexNames;
+    }
+
     /**
      * @param int $domainId
      * @param string $index
      * @return array
      */
-    protected function getDefinition(int $domainId, string $index): array
+    public function getDefinition(int $domainId, string $index): array
```
boris-brtan: is this backward compatible?

simara-svatopluk: you are right. This is a BC break. We agreed on this solution together with @DavidOstrozlik this morning (we were rushing).
@DavidOstrozlik please create a new method that will be public and use this old protected deprecated method

boris-brtan:
- [ ] something like "product-search-recreate-structure" is launched during deploy phing target
- [ ] index contains prefix based on structure version
- [ ] index should not be rebuild if the structure is the same as the one in use
- [ ] new index will be used in new version of deployed app (the old one in old version of deployed app)
- [ ] solved process of old indices removal
- [ ] edge-case structure changes and DB migrations is solved in process or documented
- [ ] during PIA / Review is coop with some implementation team

docs/installation/installation-using-docker-on-production-server.md
```diff
@@ -348,6 +348,8 @@ docker-compose -p production up -d --force-recreate webserver
 # remove container for building the application image
 docker rm -f build-php-fpm-container
 ```
+***Note:** During `build-deploy-part-2-db-dependent` phing target `product-search-reindex-structure-if-necessary` is called and can cause error when you change the type of field to another (eg. you change it from `bool` to `integer`). If you need to make this change, please add new field with the correct type and delete the old field instead*
+
```
boris-brtan:
```suggestion
```

boris-brtan:
another thing is that maybe we should change the cheat sheet of deploy so there will be the deploy scenario with indexes described.

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -34,16 +46,77 @@ public function __construct(string $definitionDirectory, string $indexPrefix, Cl
...
+    public function getVersionedIndexName(int $domainId, string $index): string
+    {
+        return $this->indexPrefix . $this->buildVersion . $index . $domainId;
+    }
+
+    /**
+     * @param int $domainId
+     * @return string
+     */
+    public function getCurrentIndexName(int $domainId): string
```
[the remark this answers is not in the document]
DavidOstrozlik: The index is not needed so why would we add it there?

packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
```diff
@@ -52,10 +125,10 @@ public function createIndexes(OutputInterface $output)
     public function deleteIndexes(OutputInterface $output)
     {
         foreach ($this->domain->getAllIds() as $domainId) {
-            $output->writeln(sprintf('Deleting index for id %s', $domainId));
-            $this->elasticsearchStructureManager->deleteIndex(
-                $domainId,
-                ProductElasticsearchRepository::ELASTICSEARCH_INDEX
+            $output->writeln(sprintf('Deleting index for domain with ID %s', $domainId));
+            $this->elasticsearchStructureManager->deleteNotUsedIndexes(
+                $this->elasticsearchStructureManager->getAliasName($domainId, ProductElasticsearchRepository::ELASTICSEARCH_INDEX),
```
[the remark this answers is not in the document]
DavidOstrozlik: that is not a setting but a constant.. and the thing with setting is that if the DB is created without running the migrations a lot of values in settings woudn't be there

packages/framework/src/Command/ProductSearchReindexIfNecessaryCommand.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @param \Symfony\Component\Console\Input\InputInterface $input
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $symfonyStyleIo = new SymfonyStyle($input, $output);
+        $output->writeln('Reindexing indexes');
+        $this->productSearchExportStructureFacade->reindexIndexesIfNecessary($output);
+        $symfonyStyleIo->success('Indexes reindexed successfully!');
```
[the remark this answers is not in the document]
DavidOstrozlik: I think this can be solved by adding the message about not necessary reindexing in the reindex if necessary funcion

packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
```diff
@@ -29,21 +38,85 @@ public function __construct(Domain $domain, ElasticsearchStructureManager $elast
...
+        $this->elasticsearchStructureUpdateChecker = $elasticsearchStructureUpdateChecker;
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     */
+    public function reindexIndexesIfNecessary(OutputInterface $output)
+    {
+        foreach ($this->domain->getAllIds() as $domainId) {
+            $output->writeln(sprintf('Reindexing index for domain with ID %s', $domainId));
```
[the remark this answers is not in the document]
DavidOstrozlik: I think this is a good idea because I want to have message that the reindexing started for every domain

packages/framework/build.xml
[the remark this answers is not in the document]
DavidOstrozlik: I don't think that it should be part of build-demo* because it's only useful for production. Maybe we can add it to our special job on CI to check it

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -75,6 +75,20 @@ There you can find links to upgrade notes for other versions too.
     - stop using the deprecated method `\Shopsys\FrameworkBundle\Model\Pricing\PriceCalculation::getVatCoefficientByPercent()`, use `PriceCalculation::getVatAmountByPriceWithVat()` for VAT calculation instead
     - if you want to customize the VAT calculation (eg. revert it back to the previous implementation), extend the service `@Shopsys\FrameworkBundle\Model\Pricing\PriceCalculation` and override the method `getVatAmountByPriceWithVat()`
     - if you created new tests regarding the price calculation they might start failing after the upgrade - in such case, please see the new VAT calculation and change the tests expectations accordingly
+- use aliases of index and build version in index name for better usage when deploying
+    - use method `ElasticsearchStructureManager::getAliasName` instead of `ElasticsearchStructureManager::getIndexName` when calling a query to elasticSearch in order to always target the right index
+        - fix test `FilterQueryTest::createFilter`
```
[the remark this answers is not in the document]
DavidOstrozlik: `getIndexName` is still the same so tests should work like before

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -89,6 +89,18 @@ There you can find links to upgrade notes for other versions too.
         -   shortDescription
         +   short_description
         ```
+    - use method `ElasticsearchStructureManager::deleteIndex` instead of `ElasticsearchStructureManager::deleteNotUsedIndexes` as it was deprecated
+    - if you have extended `ElasticsearchStructureManager` in `services.yml` you'll need call `setBuildVersion` and to send build version to it
```
boris-brtan: what about something like 
```suggestion
    - if you have extended `ElasticsearchStructureManager` in `services.yml` you'll need to call `setBuildVersion` method with $buildVersion as a first parameter
```

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -57,20 +57,50 @@ public function setBuildVersion(int $buildVersion): void
...
+    /**
+     * @param string $aliasName
+     * @return string
+     */
+    public function getCurrentIndexName(string $aliasName): string
+    {
+        $indexNames = $this->getExistingIndexNamesForAlias($aliasName);
+
+        if (count($indexNames) > 1) {
+            throw new ElasticsearchStructureException('There is more than one index set as alias. Please delete all non current ones.');
```
PetrHeinz: Rather *"There is more than one aliased index."* or *"The alias points to more than one index."*. From what I understand, you create an alias for an index (or alias an index), not set index as alias.

PetrHeinz: Also, can this actually happen in some scenario?

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -57,20 +57,50 @@ public function setBuildVersion(int $buildVersion): void
...
+
     /**
      * @param int $domainId
      * @param string $index
      * @return string
      */
     public function getAliasName(int $domainId, string $index): string
     {
-        return $this->indexPrefix . $index . $domainId;
+        return $this->indexPrefix . $index . $domainId . 'alias';
```
PetrHeinz: I would rather put a separator there: `. '_alias'`

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -79,57 +109,141 @@ public function getAliasName(int $domainId, string $index): string
...
     public function createIndex(int $domainId, string $index)
     {
-        $definition = $this->getDefinition($domainId, $index);
+        $definition = $this->getStructureDefinition($domainId, $index);
         $indexes = $this->client->indices();
-        $aliasName = $this->getAliasName($domainId, $index);
-        $indexName = $this->getIndexName($domainId, $index);
+        $indexName = $this->getVersionedIndexName($domainId, $index);
         if ($indexes->exists(['index' => $indexName])) {
+            if ($this->buildVersion === 0) {
```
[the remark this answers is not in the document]
PetrHeinz: I would rather check for the `null` here as `$this->buildVersion` is nullable.

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -79,57 +109,141 @@ public function getAliasName(int $domainId, string $index): string
...
+                throw new ElasticsearchStructureException('Index %s already exists. Please start using build version or delete current index before creating a new one.');
```
PetrHeinz: The placeholder for `$indexName` doesn't get replaced here.

Maybe the code would get clearer if you rewrote it with only one `sprintf` and one `throw`:
```php
$message = sprintf('Index %s already exists', $indexName);
if ($this->buildVersion === 0) {
    $message .= ' Please start using build version or delete current index before creating a new one.';
}
throw new ElasticsearchStructureException($message);
```

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -79,57 +109,141 @@ public function getAliasName(int $domainId, string $index): string
...
+            }
             throw new ElasticsearchStructureException(sprintf('Index %s already exists', $indexName));
```
PetrHeinz: Maybe when the buildVersion is in use and the index already exists, we can tell the user to run `php phing generate-build-version` as that seems like a thing people can miss. I've forgot to do that while testing this earlier:
```php
} else {
    $message .= ' Maybe you forgot to execute "php phing generate-build-version" before creating new index?';
}
throw new ElasticsearchStructureException($message);
```

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -79,57 +109,141 @@ public function getAliasName(int $domainId, string $index): string
...
+                'index' => $indexName,
+                'name' => $aliasName,
+            ]);
+        }
+    }
+
+    /**
+     * @param string $aliasName
+     */
+    public function deleteCurrentIndex(string $aliasName)
```
PetrHeinz: If the only reasonable usage is `$this->deleteCurrentIndex($this->getAliasName($domainId, $index))`, maybe the call to `getAliasName` should be inside the method.

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -34,12 +40,35 @@ public function __construct(string $definitionDirectory, string $indexPrefix, Cl
...
+    {
+        return $this->indexPrefix . $this->buildVersion . $index . $domainId;
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return string
+     */
+    public function getAliasName(int $domainId, string $index): string
```
PetrHeinz: Later, you call it "current" in some cases. Maybe renaming it to `getCurrentAliasName` could clarify things...

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -79,57 +109,141 @@ public function getAliasName(int $domainId, string $index): string
...
 
-        $indexes->putAlias([
-            'index' => $indexName,
-            'name' => $aliasName,
-        ]);
+    /**
+     * @param int $domainId
+     * @param string $index
+     */
+    public function migrate(int $domainId, string $index)
```
PetrHeinz: Shouldn't it be `migrateIndex` for consistency?

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -79,57 +109,141 @@ public function getAliasName(int $domainId, string $index): string
...
+            $indexes->delete(['index' => $indexName]);
+        }
     }
 
     /**
      * @param int $domainId
      * @param string $index
      */
-    public function deleteIndexesOfCurrentAlias(int $domainId, string $index)
+    public function reindexFromCurrentIndexToNewIndex(int $domainId, string $index)
```
PetrHeinz: This should be protected as it's called only within the class.

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureUpdateChecker.php
```diff
@@ -0,0 +1,97 @@
...
+        $this->elasticsearchStructureManager = $elasticsearchStructureManager;
+        $this->recursiveArraySorter = $recursiveArraySorter;
+    }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return bool
+     */
+    public function isNecessaryToUpdateStructure(int $domainId, string $index): bool
```
PetrHeinz: I'm still a bit worried about the hard-coded keys in `prepareCurrentIndexDefinitionForComparing`. Can you perhaps create a functional test that recreates the structure (or migrates it, if that's better in your opinion) and then asserts that `isNecessaryToUpdateStructure` returns `false`?

packages/framework/src/Component/Setting/Setting.php
```diff
@@ -21,6 +21,7 @@ class Setting
     public const FEED_NAME_TO_CONTINUE = 'feedNameToContinue';
     public const FEED_DOMAIN_ID_TO_CONTINUE = 'feedDomainIdToContinue';
     public const FEED_ITEM_ID_TO_CONTINUE = 'feedItemIdToContinue';
+    public const ELASTICSEARCH_INDEX = 'elasticsearchIndex';
```
PetrHeinz: This wasn't really used in the end.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
```diff
@@ -29,20 +38,80 @@ public function __construct(Domain $domain, ElasticsearchStructureManager $elast
         $this->elasticsearchStructureManager = $elasticsearchStructureManager;
     }
 
+    /**
+     * @required
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\ElasticsearchStructureUpdateChecker $elasticsearchStructureUpdateChecker
+     * @internal Will be replaced with constructor injection in the next major release
+     */
+    public function setElasticsearchStructureUpdateChecker(ElasticsearchStructureUpdateChecker $elasticsearchStructureUpdateChecker): void
```
PetrHeinz: Please check #1164 in order to make this setter injector consistent with the rest (i.e. making the use of constuctor injection possible).

packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
```diff
@@ -29,20 +38,80 @@ public function __construct(Domain $domain, ElasticsearchStructureManager $elast
...
+    public function migrateIndexesIfNecessary(OutputInterface $output)
+    {
+        foreach ($this->domain->getAllIds() as $domainId) {
+            $output->writeln(sprintf('Migrating index for domain with ID %s', $domainId));
+
+            $index = ProductElasticsearchRepository::ELASTICSEARCH_INDEX;
+
+            if ($this->elasticsearchStructureUpdateChecker->isNecessaryToUpdateStructure($domainId, $index)) {
+                $this->elasticsearchStructureManager->migrate($domainId, $index);
+                $output->writeln(sprintf('Migrating done for domain with ID %s', $domainId));
```
PetrHeinz: *"Migration done..."*

packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
```diff
@@ -29,20 +38,80 @@ public function __construct(Domain $domain, ElasticsearchStructureManager $elast
...
+    protected function createAlias(OutputInterface $output, int $domainId)
+    {
+        $output->writeln(sprintf('Creating alias for domain with ID %s', $domainId));
+
+        $this->elasticsearchStructureManager->createAliasForIndex(
+            $domainId,
+            ProductElasticsearchRepository::ELASTICSEARCH_INDEX
+        );
+
+        $output->writeln('Alias created');
```
PetrHeinz: I'm not sure with the message here as it seems that the alias is created only if it doesn't already exist...

packages/framework/src/Resources/config/services.yml
```diff
@@ -995,8 +995,12 @@ services:
         calls:
             - method: setBuildVersion
               arguments:
-                  - ['%build-version%']
+                  - '%build-version%'
```
PetrHeinz: This looks like a change that should have been in previous commit.

packages/framework/tests/Unit/Component/Elasticsearch/ElasticsearchStructureManagerTest.php
```diff
@@ -72,7 +72,7 @@ public function testDeleteSuccess(): void
         $expectedDelete = ['index' => 'test-product1'];
         $this->indices->expects($this->once())->method('delete')->with($expectedDelete);
 
-        $this->structureManager->deleteIndexesOfCurrentAlias(1, static::ELASTICSEARCH_INDEX);
+        $this->structureManager->deleteIndex(1, static::ELASTICSEARCH_INDEX);
```
PetrHeinz: This looks like a change that should have been squashed to one of previous commits. Also, the `deleteIndex()` method is deprecated. As a rule of thumb, you really shouldn't call a deprecated method, even that you have deprecated yourself.

If the method is still useful, maybe you shouldn't have deprecated at all. Or maybe there is some mocked behavior missing (for the aliasing part).

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -101,6 +101,43 @@ There you can find links to upgrade notes for other versions too.
...
+              arguments:
+                  - '%shopsys.elasticsearch.structure_dir%'
+                  - '%env(ELASTIC_SEARCH_INDEX_PREFIX)%'
+        +     calls:
+        +         - method: setBuildVersion
+        +           arguments:
+        +               - '%build-version%'
+        ```
+    - use method `ElasticsearchStructureManager::getCurrentIndexName` or `ElasticsearchStructureManager::getAliasName` instead of `ProductElasticsearchRepository::getIndexName` as it was deprecated
+    - change every number in mapping in `*.json` to string in order to be able to check definition for changes
```
PetrHeinz: So even the `settings.index.number_of_shards` and `settings.index.number_of_replicas`? Why haven't you changed them yourself?

Maybe a bit better approach would be to recursively traverse the definition yourself during the need-to-update check and cast all integers you encounter to strings. That would be automatic, not as brittle and wouldn't require a change in the structure.

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -152,4 +266,14 @@ protected function getDefinition(int $domainId, string $index): array
...
     }
+
+    /**
+     * @param int $domainId
+     * @param string $index
+     * @return array
+     */
+    public function getStructureDefinition(int $domainId, string $index): array
+    {
+        return $this->getDefinition($domainId, $index);
```
PetrHeinz: Instead of calling the deprecated method, you should move the code here and call `$this->getStructureDefinition($domainId, $index);` from the deprecated method.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -101,6 +101,43 @@ There you can find links to upgrade notes for other versions too.
...
 - implement `createFromIdAndName(int $id, string $name): FriendlyUrlData` method in your implementations of `FriendlyUrlDataFactoryInterface` as the method will be added to the interface in `v8.0.0` version ([#948](https://github.com/shopsys/shopsys/pull/948))
+- use aliases of index and build version in index name for better usage when deploying
+    - use method `ElasticsearchStructureManager::getCurrentIndexName` or `ElasticsearchStructureManager::getAliasName` instead of `ElasticsearchStructureManager::getIndexName` when calling a query to elasticSearch in order to always target the right index
+        - fix test `FilterQueryTest::createFilter`
+        ```diff
+        -   $elasticSearchIndexName = $elasticSearchStructureManager->getIndexName(1, self::ELASTICSEARCH_INDEX);
+        +   $elasticSearchIndexName = $elasticSearchStructureManager->getAliasName(1, self::ELASTICSEARCH_INDEX);
+        ```
+    - run `php phing product-search-recreate-structure` to generate new indexes with aliases
```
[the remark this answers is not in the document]
DavidOstrozlik: Its not the same as reindex. You should call recreate to recreate structure

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
DavidOstrozlik: Yes, it can happen in some scenario

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
DavidOstrozlik: I removed the suffix

packages/framework/src/Resources/config/services.yml
DavidOstrozlik: I'll squash the two commits

packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
DavidOstrozlik: I removed it

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureUpdateChecker.php
PetrHeinz: resolved in 19c21e422, 09503e5bb and e158d95a7

packages/framework/src/Model/Product/Search/Export/ProductSearchExportStructureFacade.php
```diff
@@ -45,11 +50,15 @@ public function __construct(Domain $domain, ElasticsearchStructureManager $elast
...
     public function setElasticsearchStructureUpdateChecker(ElasticsearchStructureUpdateChecker $elasticsearchStructureUpdateChecker): void
     {
-        if ($this->elasticsearchStructureUpdateChecker !== null) {
+        if ($this->elasticsearchStructureUpdateChecker !== null && $this->elasticsearchStructureUpdateChecker !== $elasticsearchStructureUpdateChecker) {
             throw new BadMethodCallException(sprintf('Method "%s" has been already called and cannot be called multiple times.', __METHOD__));
         }
 
-        $this->elasticsearchStructureUpdateChecker = $elasticsearchStructureUpdateChecker;
+        if ($this->elasticsearchStructureUpdateChecker === null) {
+            @trigger_error(sprintf('Method "%s" has been already called and cannot be called multiple times.', __METHOD__), E_USER_DEPRECATED);
```
PetrHeinz: The message should be `sprintf('The %s() method is deprecated and will be removed in the next major. Use the constructor injection instead.', __METHOD__);`, you just copied the `BadMethodCallException` message.

packages/framework/src/Component/Elasticsearch/ElasticsearchStructureManager.php
```diff
@@ -87,7 +87,7 @@ public function getVersionedIndexName(int $domainId, string $index): string
     public function getCurrentIndexName(string $aliasName): string
     {
         $indexes = $this->client->indices();
-        if ($indexes->exists(['index' => $aliasName])) {
+        if (!$indexes->existsAlias(['name' => $aliasName])) {
```
PetrHeinz: I would probably add ` && $indexes->exists(['index' => $aliasName]` to the condition, otherwise `ElasticsearchStructureManager::getCurrentIndexName('blabla')` will just return `'blabla'`

docs/introduction/console-commands-for-application-management-phing-targets.md
```diff
@@ -140,6 +140,17 @@ Consists of two subtasks that can be run independently:
...
+#### product-search-migrate-structure
+Migrates Elasticsearch indexes if there is change between currently used structure and the one in `*.json`.
+Especially useful when you need to change the structure and don't need to have fresh data in Elasticsearch
+* creates new index without alias
+* reindexes data from old index to the new one
+* deletes old index
+* creates alias for the new index
+
+*Note: If you add field/s to the structure and reindex they won't be available until `product-search-export-products` is called*
+*Note: Using this phing target after changing the type of field to another in structure (eg. changing it from `bool` to `integer`) will cause an error. If you need to make this change, please add new field with the correct type and delete the old field instead.*
```
PetrHeinz:
```suggestion

*Note: Using this phing target after changing the type of field to another in structure (eg. changing it from `bool` to `integer`) will cause an error.*
*If you need to make this change, please add new field with the correct type and delete the old field instead.*
```

docs/introduction/console-commands-for-application-management-phing-targets.md
```diff
@@ -140,6 +140,17 @@ Consists of two subtasks that can be run independently:
...
 
+#### product-search-migrate-structure
+Migrates Elasticsearch indexes if there is change between currently used structure and the one in `*.json`.
+Especially useful when you need to change the structure and don't need to have fresh data in Elasticsearch
+* creates new index without alias
+* reindexes data from old index to the new one
+* deletes old index
+* creates alias for the new index
+
+*Note: If you add field/s to the structure and reindex they won't be available until `product-search-export-products` is called*
```
PetrHeinz:
```suggestion
*Note: If you add field/s to the structure and reindex they won't be available until `product-search-export-products` is called.*
*You application must handle the properties not being filled correctly until all products are exported.*
```

docs/introduction/console-commands-for-application-management-phing-targets.md
```diff
@@ -140,6 +140,20 @@ Consists of two subtasks that can be run independently:
...
+#### product-search-migrate-structure
+Migrates Elasticsearch indexes if there is change between currently used structure and the one in `*.json`.
+Especially useful when you need to change the structure and don't need to have fresh data in Elasticsearch
+* creates new index without alias
+* reindexes data from old index to the new one
+* deletes old index
+* creates alias for the new index
+
+*Note: If you add field/s to the structure and reindex they won't be available until `product-search-export-products` is called.*
+*You application must handle the properties not being filled correctly until all products are exported.*
```
boris-brtan:
```suggestion
*Your application must handle the properties not being filled correctly until all products are exported.*
```

PR #1143, opened 2019-06-20
[shopsys] removed BC break in tests

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #970 was created BC break in tests that causes errors in tests. This PR fixes that.
|New feature| No 
|Fixes issues| ... 


TomasLudvik:
> HI @TomasLudvik,
> THX for the fix, what about CustomerDataFactoryInterface ? how was that done before?

Hello @boris-brtan,
in some magic way it was working even without being defined in the services_test.yml. I have no clue how it was possible, but I added it anyway.

boris-brtan:
@TomasLudvik seems like magic is gone
image

PR #1153, opened 2019-06-23
[framework] added immediate export to elastic on product update in administration

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In administration when product is created, edited or deleted it is now immediately exported into Elasticsearch. It helps to develop locally, where Cron jobs are not usually running and it's more comfortable for administrators. Changes are exported only when product is directly changed. Changes for example in pricing groups are not immediately propagated. This will be addressed with asynchronous tasks later. More description follows.
|New feature| Yes 
|Fixes issues| ... 

**Why it's called on `kernel.response` event and not directly in `ProductController`, where the product is saved?**

When a product is created, it's committed into the database, but prices and availability are only marked for immediate recalculations. And these recalculations are done thanks to `runImmediateRecalculations` which listen to the `kernel.response` event.
So data for the export are ready almost at the end of the request. 

_note: same logic applies for editing because we can edit attributes that affect availability and/or prices and needs to be recalculated too._

Having all this in mind, it's mandatory to set the priority of this listener to lower value to ensure it will be processed after recalculations and after the transaction is committed.

**What actions are immediately propagated into Elasticsearch?**
Create product, create variant and edit/delete of product, variant or main variant.

**Other**
As variants are not exported into Elasticsearch by itself, the main variant is exported when variant is updated/deleted.

When Product is edited in a way it's no longer visible, then it's deleted from Elasticsearch (it's the same behavior as `product-search-export-products` do).

Listening to `kernel.response` event has to be removed while running tests because there is no isolation of Elasticsearch and tests can unpredictably interfere with each other. 
This is a hotfix for a problem reported in #1171

### DIFF
--- a/packages/framework/src/Controller/Admin/ProductController.php
+++ b/packages/framework/src/Controller/Admin/ProductController.php
@@ -2,7 +2,6 @@
 
 namespace Shopsys\FrameworkBundle\Controller\Admin;
 
-use BadMethodCallException;
 use Doctrine\ORM\QueryBuilder;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
@@ -25,7 +24,6 @@ use Shopsys\FrameworkBundle\Model\Product\Product;
 use Shopsys\FrameworkBundle\Model\Product\ProductDataFactoryInterface;
 use Shopsys\FrameworkBundle\Model\Product\ProductFacade;
 use Shopsys\FrameworkBundle\Model\Product\ProductVariantFacade;
-use Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportScheduler;
 use Shopsys\FrameworkBundle\Model\Product\Unit\UnitFacade;
 use Shopsys\FrameworkBundle\Twig\ProductExtension;
 use Symfony\Component\HttpFoundation\Request;
@@ -104,11 +102,6 @@ class ProductController extends AdminBaseController
     protected $availabilityFacade;
 
     /**
-     * @var \Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportScheduler
-     */
-    protected $productSearchExportScheduler;
-
-    /**
      * @param \Shopsys\FrameworkBundle\Model\Product\MassAction\ProductMassActionFacade $productMassActionFacade
      * @param \Shopsys\FrameworkBundle\Component\Grid\GridFactory $gridFactory
      * @param \Shopsys\FrameworkBundle\Model\Product\ProductFacade $productFacade
@@ -157,19 +150,6 @@ class ProductController extends AdminBaseController
     }
 
     /**
-     * @param \Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportScheduler $productSearchExportScheduler
-     * @internal Will be replaced with constructor injection in the next major release
-     */
-    public function setProductSearchExportScheduler(ProductSearchExportScheduler $productSearchExportScheduler): void
-    {
-        if ($this->productSearchExportScheduler !== null) {
-            throw new BadMethodCallException(sprintf('Method "%s" has been already called and cannot be called multiple times.', __METHOD__));
-        }
-
-        $this->productSearchExportScheduler = $productSearchExportScheduler;
-    }
-
-    /**
      * @Route("/product/edit/{id}", requirements={"id" = "\d+"})
      * @param \Symfony\Component\HttpFoundation\Request $request
      * @param mixed $id
@@ -184,8 +164,6 @@ class ProductController extends AdminBaseController
 
         if ($form->isSubmitted() && $form->isValid()) {
             $this->productFacade->edit($id, $form->getData());
-
-            $this->scheduleProductOrMainVariantForImmediateExport($id, $product);
 
             $this->getFlashMessageSender()->addSuccessFlashTwig(
                 t('Product <strong>{{ product|productDisplayName }}</strong> modified'),
@@ -229,8 +207,6 @@ class ProductController extends AdminBaseController
 
         if ($form->isSubmitted() && $form->isValid()) {
             $product = $this->productFacade->create($form->getData());
-
-            $this->scheduleProductOrMainVariantForImmediateExport($product->getId(), $product);
 
             $this->getFlashMessageSender()->addSuccessFlashTwig(
                 t('Product <strong>{{ product|productDisplayName }}</strong> created'),
@@ -322,8 +298,6 @@ class ProductController extends AdminBaseController
 
             $this->productFacade->delete($id);
 
-            $this->scheduleProductOrMainVariantForImmediateExport($id, $product);
-
             $this->getFlashMessageSender()->addSuccessFlashTwig(
                 t('Product <strong>{{ product|productDisplayName }}</strong> deleted'),
                 [
@@ -364,8 +338,6 @@ class ProductController extends AdminBaseController
             $mainVariant = $formData[VariantFormType::MAIN_VARIANT];
             try {
                 $newMainVariant = $this->productVariantFacade->createVariant($mainVariant, $formData[VariantFormType::VARIANTS]);
-
-                $this->scheduleProductOrMainVariantForImmediateExport($newMainVariant->getId(), $newMainVariant);
 
                 $this->getFlashMessageSender()->addSuccessFlashTwig(
                     t('Variant <strong>{{ productVariant|productDisplayName }}</strong> successfully created.'),
@@ -450,31 +422,5 @@ class ProductController extends AdminBaseController
         }
 
         return true;
-    }
-
-    /**
-     * @internal Will be removed in the next major release
-     */
-    protected function validateInjectedDependencies(): void
-    {
-        if (!$this->productSearchExportScheduler instanceof ProductSearchExportScheduler) {
-            throw new BadMethodCallException(sprintf('Method "%s::setProductSearchExportScheduler()" has to be called in "services.yml" definition.', __CLASS__));
-        }
-    }
-
-    /**
-     * @param int $productId Id of deleted product is null, so it needs to be passed separately
-     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
-     */
-    protected function scheduleProductOrMainVariantForImmediateExport(int $productId, Product $product): void
-    {
-        $productIdToExport = $productId;
-
-        if ($product->isVariant()) {
-            $productIdToExport = $product->getMainVariant()->getId();
-        }
-
-        $this->validateInjectedDependencies();
-        $this->productSearchExportScheduler->scheduleProductIdForImmediateExport($productIdToExport);
     }
 }
--- a/packages/framework/src/Controller/Admin/ProductController.php
+++ b/packages/framework/src/Controller/Admin/ProductController.php
@@ -2,6 +2,7 @@
 
 namespace Shopsys\FrameworkBundle\Controller\Admin;
 
+use BadMethodCallException;
 use Doctrine\ORM\QueryBuilder;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
@@ -24,6 +25,7 @@ use Shopsys\FrameworkBundle\Model\Product\Product;
 use Shopsys\FrameworkBundle\Model\Product\ProductDataFactoryInterface;
 use Shopsys\FrameworkBundle\Model\Product\ProductFacade;
 use Shopsys\FrameworkBundle\Model\Product\ProductVariantFacade;
+use Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportScheduler;
 use Shopsys\FrameworkBundle\Model\Product\Unit\UnitFacade;
 use Shopsys\FrameworkBundle\Twig\ProductExtension;
 use Symfony\Component\HttpFoundation\Request;
@@ -102,6 +104,11 @@ class ProductController extends AdminBaseController
     protected $availabilityFacade;
 
     /**
+     * @var \Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportScheduler
+     */
+    protected $productSearchExportScheduler;
+
+    /**
      * @param \Shopsys\FrameworkBundle\Model\Product\MassAction\ProductMassActionFacade $productMassActionFacade
      * @param \Shopsys\FrameworkBundle\Component\Grid\GridFactory $gridFactory
      * @param \Shopsys\FrameworkBundle\Model\Product\ProductFacade $productFacade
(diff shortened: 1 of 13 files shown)


grossmannmartin:
Hi, thanks for your insights. The scheduler makes complete sense, thanks!
But the usage in the facade is problematic and I added an explanation into PR description.

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
```diff
@@ -153,6 +153,41 @@ public function exportWithOutput(int $domainId, string $locale, SymfonyStyle $sy
...
     }
 
+    /**
+     * @param int $domainId
+     * @param string $locale
+     * @param int[] $productIds
+     */
+    public function exportIds(int $domainId, string $locale, array $productIds): void
+    {
+        $exportedIds = $this->exportBatchIds($domainId, $locale, $productIds);
```
TomasLudvik: It took me some time to figure out how this works and why are only some IDs exported. I would move logic with loading and deciding what should be deleted here. 

You load data by ids, send them to exportBatch, convert data to ids and decide what should be deleted.

packages/framework/src/Controller/Admin/ProductController.php
```diff
@@ -423,4 +451,30 @@ protected function productCanBeCreated()
 
         return true;
     }
+
+    /**
+     * @internal Will be removed in the next major release
+     */
+    protected function validateInjectedDependencies(): void
```
TomasLudvik: This should be part of the `ProductSearchExportScheduler`

packages/framework/src/Model/Product/Search/Export/ProductSearchExportRepository.php
```diff
@@ -73,6 +73,22 @@ protected function createQueryBuilder(int $domainId, string $locale): QueryBuild
         return $queryBuilder;
     }
 
+    /**
+     * @param int $domainId
+     * @param string $locale
+     * @param int[] $productIds
+     * @return array
+     */
+    public function getProductsDataForIds(int $domainId, string $locale, array $productIds): array
```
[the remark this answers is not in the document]
TomasLudvik: This method should return Products and then be mapped to array as needed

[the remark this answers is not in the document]
grossmannmartin: Maybe, but now it's consistent with already present method `getProductsData`. 
Another argument against is – in query builder are selected only some fields and if the Products were returned, I would have to move selecting fields into Exporter.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportScheduler.php
```diff
@@ -0,0 +1,39 @@
...
+class ProductSearchExportScheduler
+{
+    protected $productIds = [];
+
+    /**
+     * @param int $productId
+     */
+    public function scheduleProductIdForImmediateExport(int $productId): void
+    {
+        $this->productIds[$productId] = $productId;
```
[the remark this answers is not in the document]
grossmannmartin: It's simple deduplication of IDs.  for this reason I'm using `array_values` below.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportScheduler.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        return $this->productIds !== [];
+    }
+
+    /**
+     * @return int[]
+     */
+    public function getProductIdsForImmediateExport(): array
+    {
+        return array_values($this->productIds);
```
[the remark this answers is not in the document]
grossmannmartin: Have to be when I'm using an associative array (see above).

But I can change it into array_unique and change associative array to normal array.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportWithFilterRepository.php
```diff
@@ -53,28 +53,62 @@ public function getProductsData(int $domainId, string $locale, int $startFrom, i
...
+        return $result;
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param int $domainId
+     * @param string $locale
+     * @return array
+     */
+    protected function extractResult(Product $product, int $domainId, string $locale): array
```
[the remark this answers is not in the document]
grossmannmartin: Probably the whole repository should be split into two classes, but I don't want to introduce this change in this PR as it will be probably a BC break

packages/framework/src/Controller/Admin/ProductController.php
grossmannmartin: Actually, it shouldn't because I need to know if the dependency was properly injected into `ProductController`.

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
grossmannmartin: Yeah, the thing is, I get some IDs, and I don't know if they need to be deleted or exported. 
So I export the IDs, from database I get list of IDs that was found and are valid to export to Elasticsearch and rest of the IDs have to be deleted from Elastic to prevent having old data in Elastic.

So you suggest to merge these two methods (`exportIds` and `exportBatchIds`)?

grossmannmartin:
Also with the fixup where listener is deregistered from kernel.response it should be possible to move scheduler calls to the facade, but I have a little trouble with it.
When you import data, they will still not be exported into elasticsearch as `kernel.response` is not called in CLI, and it just store IDs in memory to be just dropped afterwards.
If you need export to the Elasticsearch, you have to do it by yourself and you probably want to do it in batch anyway. 

As this is more like a hotfix and proper solution will be present with asynchronous queue, I'm ok with keeping scheduler directly in controller.

What do you say?

packages/framework/src/Model/Product/Search/Export/ProductSearchExportScheduler.php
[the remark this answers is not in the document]
TomasLudvik: Well I would prefer that because when I see something like this, I expect that there is some functionality connected to those keys

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
TomasLudvik: I would move the first part (loading data and extracting IDs) to exportIds and rest keep there and rename the method to exportBatch

packages/framework/src/Model/Product/Search/Export/ProductSearchExportWithFilterRepository.php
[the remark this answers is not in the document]
TomasLudvik: We should prevent that in 8.0 I think... And now I am not sure if this should be part of `8.0 deprecations removement` or somewhere else, but I think we should do that.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportRepository.php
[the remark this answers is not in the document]
TomasLudvik: BC -_-

TomasLudvik:
> Also with the fixup where listener is deregistered from kernel.response while running tests, it should be possible to move scheduler calls to the facade, but I have little doubts about it.
> When you import data, they will still not be exported into elasticsearch as `kernel.response` is not called in CLI, and it just store IDs in memory to be just dropped afterwards.
> If you need export to the Elasticsearch, you have to do it by yourself and you probably want to do it in batch anyway.
> 
> As this is more like a hotfix and proper solution will be present with asynchronous queue, I'm ok with keeping scheduler directly in controller.
> 
> What do you say?

I think that you should move logic to Facade and in the Listener create new method `exportProducts` that will be public and called from `onKernelResponse` this way it will not harm the tests and will be possible to call from for example some custom ProductImport

packages/framework/src/Model/Product/ProductFacade.php
```diff
@@ -442,4 +480,15 @@ public function getOneByCatnumExcludeMainVariants($productCatnum)
     {
         return $this->productRepository->getOneByCatnumExcludeMainVariants($productCatnum);
     }
+
+    /**
+     * @internal Will be removed in the next major release
+     * @deprecated
+     */
+    protected function validateInjectedDependencies(): void
```
TomasLudvik: Just wondering... should this be protected or private? :)

packages/framework/src/Model/Product/ProductVariantFacade.php
```diff
@@ -44,6 +46,11 @@ class ProductVariantFacade
      */
     protected $productAvailabilityRecalculationScheduler;
 
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportScheduler
+     */
+    protected $productSearchExportScheduler;
```
TomasLudvik: You are not using it here

packages/framework/src/Model/Product/Search/Export/ProductSearchExportListener.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->productSearchExportFacade = $productSearchExportFacade;
+        $this->entityManager = $entityManager;
+    }
+
+    /**
+     * @param \Symfony\Component\HttpKernel\Event\FilterResponseEvent $filterResponseEvent
+     */
+    public function onKernelResponse(FilterResponseEvent $filterResponseEvent): void
+    {
+        if ($this->productSearchExportScheduler->hasAnyProductIdsForImmediateExport()) {
```
TomasLudvik: Extract body of this method to another public method, so it can be called for example from imports

packages/framework/src/Model/Product/ProductFacade.php
grossmannmartin: According to our PrivateVisibilityFixer, it should be protected as in `Model` namespace is `private` forbidden

packages/framework/src/Model/Product/ProductVariantFacade.php
grossmannmartin: Thanks! How could I forget? Fixed.

packages/framework/src/Resources/config/services.yml
```diff
@@ -507,14 +507,6 @@ services:
             # must be run before ProductVisibilityFacade
             - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse, priority: 20 }
 
-    Shopsys\FrameworkBundle\Model\Product\ProductFacade:
```
TomasLudvik: I think this is still necessary

packages/framework/src/Model/Product/ProductVariantFacade.php
```diff
@@ -67,7 +68,8 @@ public function __construct(
         ImageFacade $imageFacade,
         ProductFactoryInterface $productFactory,
         ProductPriceRecalculationScheduler $productPriceRecalculationScheduler,
-        ProductAvailabilityRecalculationScheduler $productAvailabilityRecalculationScheduler
+        ProductAvailabilityRecalculationScheduler $productAvailabilityRecalculationScheduler,
+        ProductSearchExportScheduler $productSearchExportScheduler = null
```
TomasLudvik: `?ProductSearchExportScheduler` right?

packages/framework/src/Model/Product/ProductFacade.php
```diff
@@ -184,7 +185,8 @@ public function __construct(
         ProductCategoryDomainFactoryInterface $productCategoryDomainFactory,
         ProductParameterValueFactoryInterface $productParameterValueFactory,
         ProductVisibilityFactoryInterface $productVisibilityFactory,
-        ProductPriceCalculation $productPriceCalculation
+        ProductPriceCalculation $productPriceCalculation,
+        ProductSearchExportScheduler $productSearchExportScheduler = null
```
TomasLudvik: `?ProductSearchExportScheduler` right?

packages/framework/src/Resources/config/services_test.yml
```diff
@@ -162,9 +162,7 @@ services:
     Shopsys\FrameworkBundle\Model\Product\ProductDataFactoryInterface:
         alias: Shopsys\FrameworkBundle\Model\Product\ProductDataFactory
 
-    Shopsys\FrameworkBundle\Model\Product\ProductFacade:
```
TomasLudvik: I think this is still necessary

packages/framework/src/Resources/config/services_test.yml
```diff
@@ -177,9 +175,7 @@ services:
 
     Shopsys\FrameworkBundle\Model\Product\ProductVisibilityRepository: ~
 
-    Shopsys\FrameworkBundle\Model\Product\ProductVariantFacade:
```
TomasLudvik: I think this is still necessary

packages/framework/src/Resources/config/services.yml
grossmannmartin: I think Facade suffix is auto registered

packages/framework/src/Resources/config/services_test.yml
grossmannmartin: Scheduler is properly autowired even when is nullable

packages/framework/src/Model/Product/ProductFacade.php
grossmannmartin: Right!

packages/framework/src/Model/Product/ProductVariantFacade.php
grossmannmartin: Right!

packages/framework/src/Model/Product/ProductFacade.php
```diff
@@ -349,6 +342,8 @@ public function delete($productId)
         $this->em->remove($product);
         $this->em->flush();
 
+        $this->productSearchExportScheduler->scheduleProductIdForImmediateExport($productId);
```
TomasLudvik: I do not understand what changed by this, but ok

packages/framework/src/Resources/config/services_test.yml
```diff
@@ -210,7 +210,8 @@ services:
     Shopsys\FrameworkBundle\Model\Pricing\Vat\VatFacade: ~
 
     # Prevent listening to kernel.response event while running tests
-    Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportListener: ~
+    Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportListener:
+        tags: []
```
TomasLudvik: Was this necessary? Why?

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
```diff
@@ -242,7 +242,7 @@ public function deletePresent(int $domainId, array $deleteIds): void
                     'bool' => [
                         'must' => [
                             'ids' => [
-                                'values' => $deleteIds,
+                                'values' => array_values($deleteIds),
```
TomasLudvik: Why is this necessary?

packages/framework/src/Model/Product/ProductFacade.php
grossmannmartin: It’s suggested change to improve readability of a code. Prepare data, do something with data and then affect deleted product.

packages/framework/src/Resources/config/services_test.yml
grossmannmartin: It’s not necessary, just wanted to be clear about intentions. That it’s necessary to remove tag.

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
grossmannmartin: Because when you sent array indexed somehow different than from zero, it’s translated as an object not an array and Elasticsearch fails. 
For example this code is not really expected to fail 
```php
$productIds = [5,6,7];
$doNotUpdateIds = [5];
$toUpdateIds = array_diff($productIds, $doNotUpdateIds);

$elasticRepository->delete($toUpdateIds);

PR #1155, opened 2019-06-25
[shopsys] separated backend API from default project-base

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Not all projects needs backend API, so the backend API is not part of the project-base by default and is easily installable
|New feature| Yes
|Fixes issues| ... 


packages/backend-api/integration/integrate.sh
```diff
@@ -0,0 +1,79 @@
...
+    local SOURCE_FILE_PATH=$1
+
+    if [ ! -z $2 ]
+    then
+        SOURCE_FILE_PATH=$2
+    fi
+
+    echo "Applying patch for ${FILE_PATH}..."
+    local PATCH_DRY_RUN=`patch -st --dry-run ${PROJECT_BASE_PATH}/${FILE_PATH} ${INTEGRATION_DIR}/${SOURCE_FILE_PATH}.patch`
+    local PATCH_REVERSED=`grep "Reversed" <<< ${PATCH_DRY_RUN}`
```
boris-brtan: maybe we can use just return values. or could be there more result types than this?
and maybe `-e` will cover the whole thing

packages/backend-api/integration/integrate.sh
```diff
@@ -0,0 +1,79 @@
...
+apply_patch "app/config/parameters_common.yml"
+apply_patch "app/config/routing.yml"
+apply_patch "app/config/packages/security.yml"
+apply_patch "src/Shopsys/ShopBundle/Resources/config/routing.yml"
+apply_patch "app/AppKernel.php"
+
+# if running the monorepo integration, apply monorepo patches
+if [ "$1" == "monorepo" ]
+then
+    apply_patch "composer.json" "monorepo.composer.json"
```
boris-brtan: there is project_base_path 
i am not sure whether this will modify composer.json in monorepo which is in parent folder of project-base

build.xml
```diff
@@ -14,23 +14,23 @@
 
     <target name="build" depends="build-deploy-part-1-db-independent,build-deploy-part-2-db-dependent" description="Builds application for production preserving your DB."/>
 
-    <target name="build-demo" depends="wipe,build-version-generate,composer-prod,backend-api-oauth-keys-generate,redis-check,npm,dirs-create,assets,db-demo,grunt,error-pages-generate,warmup,product-search-recreate-structure,product-search-export-products,clean-redis-old" description="Builds application for production with clean demo DB."/>
+    <target name="build-demo" depends="wipe,build-version-generate,monorepo-backend-api-integrate,composer-prod,backend-api-oauth-keys-generate,redis-check,npm,dirs-create,assets,db-demo,grunt,error-pages-generate,warmup,product-search-recreate-structure,product-search-export-products,clean-redis-old,monorepo-backend-api-disintegrate" description="Builds application for production with clean demo DB."/>
```
boris-brtan: please use `shopsys_framework.build_demo`

.ci/build_kubernetes.sh
```diff
@@ -105,4 +105,4 @@ kubectl rollout status --namespace=${JOB_NAME} deployment/webserver-php-fpm --wa
 PHP_FPM_POD=$(kubectl get pods --namespace=${JOB_NAME} -l app=webserver-php-fpm -o=jsonpath='{.items[0].metadata.name}')
 
 # Run phing build targets for build of the application
-kubectl exec ${PHP_FPM_POD} --namespace=${JOB_NAME} ./phing backend-api-oauth-keys-generate test-db-create test-dirs-create checks-ci
+kubectl exec ${PHP_FPM_POD} --namespace=${JOB_NAME} ./phing backend-api-integrate backend-api-oauth-keys-generate test-db-create test-dirs-create checks-ci
```
boris-brtan: this would need to be also in `restart_kubernetes.sh`

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,30 @@
+# Backend API Installation
+
+Require `shopsys/backend-api` package in the `composer.json` or run
+
+```bash
+composer require shopsys/backend-api ^8.0
```
boris-brtan: hmm this is new notation. Till now we always use exact version like `composer require shopsys/backend-api:7.2.2`. But i like you solution more. But for consistency reasons.....

packages/backend-api/integration/build.xml.patch
```diff
@@ -0,0 +1,33 @@
+@@ -5,10 +5,32 @@
+ 
+     <property name="path.root" value="${project.basedir}"/>
+     <property name="path.vendor" value="${path.root}/vendor"/>
++    <property name="path.backend-api" value="${path.vendor}/backend-api"/>
```
boris-brtan: i dont think so. please see path.framework it is inside `shopsys/framework` so this will be inside `shopsys/backend-api`

packages/backend-api/integration/disintegrate.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+INTEGRATION_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
```
boris-brtan: :+1:

packages/backend-api/integration/disintegrate.sh
```diff
@@ -0,0 +1,67 @@
...
+PROJECT_BASE_PATH=`[ -d "$PWD/project-base" ] && echo "$PWD/project-base" || echo "$PWD"`
```
boris-brtan: what about to make it consistent with `$(....)` instead of `` `.....` `` or vice versa

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+Backend API integration is fully automated, so you have to run only (you have to be in the project root directory)
+
+```bash
+bash vendor/shopsys/backend-api/integration/integrate.sh
```
boris-brtan: what about disintegrate?

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+If it was successful, you'll see `Backend API integration was successful!`
+
+*Now you can continue with creating [OAuth client](/docs/backend-api/api-authentication-oauth2.md)*
+
+## Troubleshooting
+
+The automated integration is based on copying file from `vendor/shopssy/backend-api/integration` directory on applying patches.
```
boris-brtan: copying file|files|folders ?

boris-brtan:
```suggestion
The automated integration is based on copying the code from `vendor/shopsys/backend-api/integration` directory on applying patches.
```

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,30 @@
...
+
+Copying files can cause problems only if you already had a file with the same name in the same directory.
+In such case please rename your files and run the integration again.
```
boris-brtan: hmmm. but OK

docs/backend-api/backend-api-installation.md
boris-brtan: hmm i see why in composer.patch is no shopsys/backend-api...but still it is reversed

docs/backend-api/introduction-to-backend-api.md
```diff
@@ -3,6 +3,11 @@
 Shopsys Framework backend API is an interface to the application that is dedicated for integration with your third party systems systems (like IS, storage system, ERP, ...).
 We use [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) (implemented using [FOSRestBundle](https://github.com/FriendsOfSymfony/FOSRestBundle)) with JSON format and [OAuth2](https://oauth.net/2/) authorization (implemented using [oauth2-bundle](https://github.com/trikoder/oauth2-bundle)).
 
+## Installation
+
+The backend API is not part of the framework by default and you have to install it by yourself.
```
boris-brtan:
```suggestion
The backend API is not a part of the Shopsys Framework by default and you have to install it by yourself.
```

build.xml
```diff
@@ -4,11 +4,33 @@
...
     <property name="path.packages" value="${project.basedir}/packages"/>
+    <property name="path.backend-api" value="${path.packages}/backend-api"/>
     <property name="path.framework" value="${path.packages}/framework"/>
     <property name="path.utils" value="${project.basedir}/utils"/>
     <property name="translations.dump.flags" value=""/>
 
     <import file="${path.framework}/build.xml"/>
+    <import file="${path.backend-api}/build.xml"/>
+
+    <target name="build" depends="build-deploy-part-1-db-independent,build-deploy-part-2-db-dependent" description="Builds application for production preserving your DB."/>
```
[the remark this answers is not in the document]
simara-svatopluk: I had patches for monorepo `build.xml` and `composer.json` originally. But then I asked myself... why? What is the benefit of not having messed build.xml for the monorepo developer?

packages/backend-api/integration/composer.json.patch
```diff
@@ -0,0 +1,8 @@
+@@ -85,6 +85,7 @@
+         "symfony-cmf/routing": "^2.0.3",
+         "symfony-cmf/routing-bundle": "^2.0.3",
+         "tracy/tracy": "^2.4.13",
++        "trikoder/oauth2-bundle": "^1.1",
```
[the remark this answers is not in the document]
simara-svatopluk: monorepo `composer.json` is without a change. This patch is applied to `project-base/composer.json`

packages/backend-api/integration/monorepo.composer.json.patch
```diff
@@ -0,0 +1,16 @@
+@@ -64,6 +64,7 @@
+         "sensio/distribution-bundle": "^5.0.21",
+         "sensio/framework-extra-bundle": "^3.0.29",
+         "sensio/generator-bundle": "^3.1.7",
++        "shopsys/backend-api": "8.0.x-dev",
```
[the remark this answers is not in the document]
simara-svatopluk: This patch is applied to `project-base/composer.json` when integrating from monorepo

packages/backend-api/build.xml
```diff
@@ -3,6 +3,18 @@
...
     <property name="path.defuse-key-generate.executable" value="${path.bin}/generate-defuse-key"/>
 
+    <target name="monorepo-backend-api-disintegrate" description="Disintegrates backend API from the project" hidden="true">
+        <exec executable="${path.backend-api}/integration/disintegrate.sh" checkreturn="true" passthru="true">
+            <arg value="monorepo"/>
+        </exec>
+    </target>
+
+    <target name="monorepo-backend-api-integrate" description="Integrates backend API to the project" hidden="true">
+        <exec executable="${path.backend-api}/integration/integrate.sh" checkreturn="true" passthru="true">
```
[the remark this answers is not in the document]
simara-svatopluk: No. But now I'm!

Do I `install` the backend API or `integrate`? I would say integrate, installation is done by `composer require`... We can ask in the team slack what do you think?

packages/backend-api/integration/app/AppKernel.php.patch
```diff
@@ -0,0 +1,38 @@
+@@ -19,8 +19,10 @@
+             new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle(),
+             new Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle(),
+             new FM\ElfinderBundle\FMElfinderBundle(),
++            new FOS\RestBundle\FOSRestBundle(),
```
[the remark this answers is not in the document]
simara-svatopluk: Could make problems... yes. Could you please test it? How much problems? Because the `patch` program is pretty clever :-)

We also sort bundles alphabetically, that's why it is here.

What is more important - keeping our standards or making the integration less pain?

packages/backend-api/integration/build.xml.patch
```diff
@@ -0,0 +1,33 @@
...
+     <property name="path.framework" value="${path.vendor}/shopsys/framework"/>
+ 
+     <import file="${path.framework}/build.xml"/>
++    <import file="${path.backend-api}/build.xml"/>
+ 
+     <property name="is-multidomain" value="true"/>
+ 
++    <target name="build" depends="build-deploy-part-1-db-independent,build-deploy-part-2-db-dependent" description="Builds application for production preserving your DB."/>
```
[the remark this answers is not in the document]
simara-svatopluk: Should I prefix also the depends section?

packages/backend-api/integration/integrate.sh
simara-svatopluk: There are 3 states after the `dry run`
* no error -> will be patched on line 58
* error -> the script stops
* the `patch` detects the patch is already applied. In such case we are skipping the patch and continue with the script

So I would say we cannot use return values. Also the "Reverse" returns code `0`

packages/backend-api/integration/integrate.sh
simara-svatopluk: this will modify files inside the `project-base` directory... Should I change anything? Or explain it somehow?

build.xml
simara-svatopluk: I just tried to do it, and when I call from command line `./phing build-demo`, it uses the phing target from the framework, and not this one... And that's not correct behavior in my opinion

docs/backend-api/backend-api-installation.md
simara-svatopluk: I found only one occurence of `composer require` in our documentation and it is

> you can do this by `composer require helios-ag/fm-elfinder-bundle:^9.2 --update-with-dependencies

with the caret ("roof") annotation...

So what should be there?

packages/backend-api/integration/disintegrate.sh
simara-svatopluk: To be honest, I don't understand the difference between these notations in bash. Please just tell me "do this" and I'll do it :-)

docs/backend-api/backend-api-installation.md
simara-svatopluk: I assume programmers will do only the integration. If they need revert their work, they have `git checkout -- ...`

build.xml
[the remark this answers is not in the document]
simara-svatopluk: Well I didn't find a way to mix most of the targets...
The order isn't strictly `monorepo-backend-api-integrate`, `backend-api-oauth-keys-generate`, `the original build`, it is more like `do something from the original build`, `monorepo-backend-api-integrate`, `composer`, `backend-api-oauth-keys-generate`, `do the rest of the original`

Notice that `backend-api-oauth-keys-generate` must be always after `composer` :-(

packages/backend-api/integration/build.xml.patch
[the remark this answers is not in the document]
simara-svatopluk: And now I removed prefixes as I realized it was nonsense... As I explain in https://github.com/shopsys/shopsys/pull/1155#discussion_r297185902 I thinks it is almost impossible :-(

docs/backend-api/backend-api-installation.md
boris-brtan: in README.md of backed api is just `composer require shopsys/backend-api`

.ci/build_kubernetes.sh
boris-brtan: hmm should not we call here also `composer-dev` after integrate ?

packages/backend-api/integration/integrate.sh
boris-brtan: hmm if we apply patch for composer.json so i think we should run `composer-dev` or `composer-prod`

docs/backend-api/backend-api-installation.md
simara-svatopluk: changed to `composer require shopsys/backend-api`

build.xml
boris-brtan: i think this tatget seems to be easy
```xml
<target name="build-demo"  description="Builds application for production with clean demo DB."
    <phingcall target="monorepo-backend-api-integrate"/>
    <phingcall target="shopsys_framework.build-demo"/>
    <phingcall target="monorepo-backend-api-disintegrate"/>
</target>
```

build.xml
```diff
@@ -4,11 +4,31 @@
...
     <property name="path.framework" value="${path.packages}/framework"/>
     <property name="path.utils" value="${project.basedir}/utils"/>
     <property name="translations.dump.flags" value=""/>
 
     <import file="${path.framework}/build.xml"/>
+    <import file="${path.backend-api}/build.xml"/>
+
+    <target name="build-demo" depends="monorepo-backend-api-install,composer-prod,backend-api-oauth-keys-generate,shopsys_framework.build-demo,monorepo-backend-api-uninstall" description="Builds application for production with clean demo DB."/>
+
+    <target name="build-demo-ci" depends="monorepo-backend-api-install,composer-prod,backend-api-oauth-keys-generate,shopsys_framework.build-demo-ci,monorepo-backend-api-uninstall" description="Builds application for development with clean demo DB and runs CI checks."/>
```
boris-brtan:
```suggestion
    <target name="build-demo-ci" depends="monorepo-backend-api-install,composer-dev,backend-api-oauth-keys-generate,shopsys_framework.build-demo-ci,monorepo-backend-api-uninstall" description="Builds application for development with clean demo DB and runs CI checks."/>
```
you need to test it

packages/framework/build.xml
```diff
@@ -69,56 +67,30 @@
...
+    <target name="build-demo" depends="wipe,build-version-generate,composer-prod,redis-check,npm,dirs-create,assets,db-demo,grunt,error-pages-generate,warmup,product-search-recreate-structure,product-search-export-products,clean-redis-old" description="Builds application for production with clean demo DB."/>
 
-    <target name="build-demo-ci" depends="build-version-generate,backend-api-oauth-keys-generate,wipe-excluding-logs,timezones-check,dirs-create,test-dirs-create,db-demo,product-search-recreate-structure,product-search-export-products,grunt,error-pages-generate,clean-redis-old,checks-ci" description="Builds application for development with clean demo DB and runs CI checks."/>
+    <target name="build-demo-ci" depends="build-version-generate,wipe-excluding-logs,composer-prod,npm,timezones-check,dirs-create,test-dirs-create,db-demo,product-search-recreate-structure,product-search-export-products,grunt,error-pages-generate,clean-redis-old,checks-ci" description="Builds application for development with clean demo DB and runs CI checks."/>
```
boris-brtan:
```suggestion
    <target name="build-demo-ci" depends="build-version-generate,wipe-excluding-logs,composer-dev,npm,timezones-check,dirs-create,test-dirs-create,assets,db-demo,product-search-recreate-structure,product-search-export-products,grunt,error-pages-generate,tests-acceptance-build,clean-redis-old,checks-ci" description="Builds application for development with clean demo DB and runs CI checks."/>
```

packages/backend-api/install/uninstall.sh
```diff
@@ -0,0 +1,67 @@
...
+apply_patch_reverse "app/config/routing.yml"
+apply_patch_reverse "app/config/packages/security.yml"
+apply_patch_reverse "src/Shopsys/ShopBundle/Resources/config/routing.yml"
+apply_patch_reverse "app/AppKernel.php"
+apply_patch_reverse "build.xml"
+
+# if running the monorepo installation, apply monorepo patches
+if [ "$1" == "monorepo" ]
+then
+    apply_patch_reverse "composer.json" "monorepo.composer.json"
```
boris-brtan:
```suggestion
    apply_patch_reverse "composer.json" "monorepo.project-base.composer.json"
```

boris-brtan: sorry, i am still confused if there is just mentioned monorepo

.ci/restart_kubernetes.sh
```diff
@@ -31,4 +31,4 @@ kubectl rollout status --namespace=${JOB_NAME} deployment/webserver-php-fpm --wa
 PHP_FPM_POD=$(kubectl get pods -n ${JOB_NAME} -l app=webserver-php-fpm -o=jsonpath='{.items[0].metadata.name}')
 
 # Run phing build in the pod
-kubectl exec --namespace=${JOB_NAME} ${PHP_FPM_POD} ./phing clean dirs-create product-search-recreate-structure product-search-export-products grunt error-pages-generate
+kubectl exec --namespace=${JOB_NAME} ${PHP_FPM_POD} ./phing clean dirs-create monorepo-backend-api-install product-search-recreate-structure product-search-export-products grunt error-pages-generate
```
boris-brtan: missing `backend-api-oauth-keys-generate`

.travis.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     include:
     -   stage: Tests
         name: "Smoke, Functional and Acceptance Tests"
-        script: docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev-quick error-pages-generate checks-ci
+        script: docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev-quick monorepo-backend-api-install backend-api-oauth-keys-generate error-pages-generate checks-ci
 
     -   name: "Standards, Unit Tests"
         script: docker-compose exec php-fpm ./phing standards tests-unit
```
boris-brtan: maybe also here .. `monorepo-backend-api-install backend-api-oauth-keys-generate`

packages/backend-api/integration/integrate.sh
simara-svatopluk: not when running from monorepo (there are already all depencencies)... And from project-base? Yes, we should. I have just feeling it shouldn't be part of the integration script and I'd like to add it to install instructions (one argument is that in the install script we have no idea wheather `dev` or `prod` should be run)

build.xml
simara-svatopluk: `build-demo` uses `composer-prod`, why do you suggest `-dev`?

.travis.yml
simara-svatopluk: sorry, what should I do? I don't understand :innocent:

build.xml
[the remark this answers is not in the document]
simara-svatopluk: ok... so this comment is resolved?

packages/backend-api/integration/integrate.sh
simara-svatopluk: ... added to installation instructions

.travis.yml
boris-brtan:
```yaml
        script: docker-compose exec php-fpm ./phing monorepo-backend-api-install backend-api-oauth-keys-generate standards tests-unit
```

build.xml
boris-brtan: ci is for CI so you need to have to run tests and those are in dev dependencies

packages/framework/build.xml
simara-svatopluk: added manually

packages/backend-api/install/composer.json.patch
```diff
@@ -0,0 +1,8 @@
+@@ -85,6 +85,7 @@
+         "symfony-cmf/routing": "^2.0.3",
+         "symfony-cmf/routing-bundle": "^2.0.3",
+         "tracy/tracy": "^2.4.13",
++        "trikoder/oauth2-bundle": "^1.1",
```
grossmannmartin: Why do you need to patch composer.json? Can this dependency be in API package?

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+If it was successful, you'll see `Backend API installation was successful!`
+
+The last step is to update your composer because the `install.sh` added a composer dependency
+
+```bash
+composer update
+```
+
```
grossmannmartin: Should be running in docker considered in this guide? Because to me, it was all clear as I run it natively, but when you running in docker, you need to perform these steps inside of a container. 
I'm not really sure how we should document this type of things.

grossmannmartin: And what about db migrations? I need to run them, but neither here, end neither in following article is no information about it

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,36 @@
...
+Require `shopsys/backend-api` package in the `composer.json` or run
+
+```bash
+composer require shopsys/backend-api
+```
+
+Backend API installation is fully automated, so you have to run only (you have to be in the project root directory)
+
+```bash
+bash vendor/shopsys/backend-api/install/install.sh
```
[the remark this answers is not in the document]
simara-svatopluk: It's not necessary, I used the `./path` version

packages/backend-api/install/composer.json.patch
simara-svatopluk: Well... complicated question.

The backend API itself doesn't need the `oauth2` bundle. The `oauth2` bundle is needed only for securing the API, and that is a responsibility of the `project-base`.

Additionally we have to mention packages we use directly from `project-base` in the project-base `composer.json`, and because we register the bundle in `AppKernel`, we have to mention it.

packages/backend-api/install/uninstall.sh
```diff
@@ -0,0 +1,67 @@
...
+    echo "Reverting ${FILE_PATH}..."
+    local PATCH_DRY_RUN=`patch -Rf --dry-run ${PROJECT_BASE_PATH}/${FILE_PATH} ${INSTALL_DIR}/${SOURCE_FILE_PATH}.patch`
+    local PATCH_FAIL=`grep "FAILED" <<< ${PATCH_DRY_RUN}`
+
+    if [ ! -z "${PATCH_FAIL}" ]
+    then
+        local PATCH_APPLY_DRY_RUN=`patch -st --dry-run ${PROJECT_BASE_PATH}/${FILE_PATH} ${INSTALL_DIR}/${SOURCE_FILE_PATH}.patch`
+        if [ -z "${PATCH_APPLY_DRY_RUN}" ]
+        then
+            echo "Applied patch detected: ${FILE_PATH} already reverted, Doing nothing"
```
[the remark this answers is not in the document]
simara-svatopluk: Do you have the output? Because this should work as you described - repeated uninstall should be without an error

packages/backend-api/install/uninstall.sh
```diff
@@ -0,0 +1,67 @@
...
+        echo "Done"
+    fi
+}
+
+apply_patch_reverse "app/config/parameters_common.yml"
+apply_patch_reverse "app/config/routing.yml"
+apply_patch_reverse "app/config/packages/security.yml"
+apply_patch_reverse "src/Shopsys/ShopBundle/Resources/config/routing.yml"
+apply_patch_reverse "app/AppKernel.php"
+apply_patch_reverse "build.xml"
```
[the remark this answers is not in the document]
simara-svatopluk: fixed

packages/backend-api/install/app/AppKernel.php.patch
```diff
@@ -0,0 +1,34 @@
...
+             new Symfony\Cmf\Bundle\RoutingBundle\CmfRoutingBundle(),
++            new Trikoder\Bundle\OAuth2Bundle\TrikoderOAuth2Bundle(),
+             new VasekPurchart\ConsoleErrorsBundle\ConsoleErrorsBundle(),
+             new FOS\CKEditorBundle\FOSCKEditorBundle(), // has to be loaded after FrameworkBundle and TwigBundle
+             new Joschi127\DoctrineEntityOverrideBundle\Joschi127DoctrineEntityOverrideBundle(),
+@@ -106,3 +110,7 @@
++        if (file_exists(__DIR__ . '/config/packages/oauth2/parameters_oauth.yml')) {
++            $configs[] = __DIR__ . '/config/packages/oauth2/parameters_oauth.yml';
++        }
++
```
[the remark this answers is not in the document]
simara-svatopluk: What is the difference? I tried to copy the pasted code and It's the same... :thinking:

packages/backend-api/install/install.sh
```diff
@@ -0,0 +1,81 @@
...
+    local PATCH_DRY_RUN=`patch -st --dry-run ${PROJECT_BASE_PATH}/${FILE_PATH} ${INSTALL_DIR}/${SOURCE_FILE_PATH}.patch`
+    local PATCH_REVERSED=`grep "Reversed" <<< ${PATCH_DRY_RUN}`
+    local PATCH_FAIL=`grep "FAILED" <<< ${PATCH_DRY_RUN}`
+
+    if [ ! -z "${PATCH_REVERSED}" ]
+    then
+        echo "Applied patch detected: patch for ${FILE_PATH} already applied. Doing nothing"
+    elif [ ! -z "${PATCH_FAIL}" ]
+    then
+        echo ${PATCH_FAIL}
```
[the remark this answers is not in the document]
simara-svatopluk: In this case... yes, but in other case this message is helpful
`Hunk #1 FAILED at 19. Hunk #2 FAILED at 35. Hunk #3 FAILED at 45. Hunk #4 FAILED at 110. 4 out of 4 hunks FAILED`

So I see only 2 possibilities - remove the `echo` or keep it. I prefer keeping slightly

packages/backend-api/install/install.sh
```diff
@@ -0,0 +1,81 @@
...
+        echo "Patch for ${FILE_PATH} cannot be applied!"
+        exit 1
```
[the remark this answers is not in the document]
simara-svatopluk: done, looking for your feedback on the change ;-)

packages/backend-api/install/app/AppKernel.php.patch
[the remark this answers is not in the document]
grossmannmartin: There are three extra lines at the beginning of the hunk.

grossmannmartin:
Hi @simara-svatopluk,
I noticed one more thing from monorepo point of view. In phing targets for building application is installation of backend api. 
From certain point of view it makes sense, but I start wondering if it's really that beneficient. 
Because `build-demo-dev` is something I use really often and then it's really easy to commit one of "installed" files by mistake.
I'm thinking if leave it to developers choice whether to install or not is a better idea.

packages/backend-api/install/install.sh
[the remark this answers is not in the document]
grossmannmartin: Okay.

packages/backend-api/install/uninstall.sh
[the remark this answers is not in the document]
grossmannmartin: Output of repeated uninstall 
```
Reverting app/config/parameters_common.yml...
missing header for unified diff at line 1 of patch patching file /Volumes/work/shopsys/framework/project-base/app/config/parameters_common.yml Hunk #1 FAILED at 30. 1 out of 1 hunk FAILED -- saving rejects to file /Volumes/work/shopsys/framework/project-base/app/config/parameters_common.yml.rej
app/config/parameters_common.yml cannot be reverted!
Reverting app/config/routing.yml...
missing header for unified diff at line 1 of patch patching file /Volumes/work/shopsys/framework/project-base/app/config/routing.yml Hunk #1 FAILED at 5. 1 out of 1 hunk FAILED -- saving rejects to file /Volumes/work/shopsys/framework/project-base/app/config/routing.yml.rej
app/config/routing.yml cannot be reverted!
Reverting app/config/packages/security.yml...
missing header for unified diff at line 1 of patch patching file /Volumes/work/shopsys/framework/project-base/app/config/packages/security.yml Hunk #1 FAILED at 36. 1 out of 1 hunk FAILED -- saving rejects to file /Volumes/work/shopsys/framework/project-base/app/config/packages/security.yml.rej
app/config/packages/security.yml cannot be reverted!
Reverting src/Shopsys/ShopBundle/Resources/config/routing.yml...
missing header for unified diff at line 1 of patch patching file /Volumes/work/shopsys/framework/project-base/src/Shopsys/ShopBundle/Resources/config/routing.yml Hunk #1 FAILED at 6. 1 out of 1 hunk FAILED -- saving rejects to file /Volumes/work/shopsys/framework/project-base/src/Shopsys/ShopBundle/Resources/config/routing.yml.rej
src/Shopsys/ShopBundle/Resources/config/routing.yml cannot be reverted!
Reverting app/AppKernel.php...
missing header for unified diff at line 1 of patch patching file /Volumes/work/shopsys/framework/project-base/app/AppKernel.php Hunk #1 FAILED at 19. Hunk #2 FAILED at 33. Hunk #3 FAILED at 42. Hunk #4 FAILED at 103. 4 out of 4 hunks FAILED -- saving rejects to file /Volumes/work/shopsys/framework/project-base/app/AppKernel.php.rej
app/AppKernel.php cannot be reverted!
Reverting build.xml...
missing header for unified diff at line 1 of patch patching file /Volumes/work/shopsys/framework/project-base/build.xml Hunk #1 FAILED at 5. 1 out of 1 hunk FAILED -- saving rejects to file /Volumes/work/shopsys/framework/project-base/build.xml.rej
build.xml cannot be reverted!
Running from monorepo
Reverting composer.json...
missing header for unified diff at line 1 of patch patching file /Volumes/work/shopsys/framework/project-base/composer.json Hunk #1 FAILED at 64. Hunk #2 FAILED at 85. 2 out of 2 hunks FAILED -- saving rejects to file /Volumes/work/shopsys/framework/project-base/composer.json.rej
composer.json cannot be reverted!
Backend API uninstallation failed!

BUILD FAILED
/Volumes/work/shopsys/framework/packages/backend-api/build.xml:7:103: Task exited with code 1

Total time: 0.2582 seconds
```

packages/backend-api/install/composer.json.patch
grossmannmartin: I understand responsibilities.
But look at it from a different perspective. Information that API should be secured is already present in the backend-api package (install patches). 
And if you need API package, you need to require it in composer.
For ease installation (less files to patch, no need to run extra composer update) I'm opened to have the dependency in backend-api package.

packages/backend-api/install/install.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash -e
+
+INSTALL_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
+PROJECT_BASE_PATH=`[ -d "$PWD/project-base" ] && echo "$PWD/project-base" || echo "$PWD"`
```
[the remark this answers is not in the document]
grossmannmartin: Thanks for the change. I'm just wondering, if you already have proper path to project-base, why you forbid the installation from a different path?
If you use 
```
PROJECT_BASE_PATH=`realpath -e ${INSTALL_DIR}/../../../project-base`
``` 
(notice `-e`) command fails when path does not exists and otherwise everything else working properly even when I run it from different folder

packages/backend-api/install/install.sh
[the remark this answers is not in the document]
grossmannmartin: Nice!

packages/backend-api/install/install.sh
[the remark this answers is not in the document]
simara-svatopluk: two reasons
* `-e` doesn't allow me show human error message, it fails on "it doesn't exists"
* for the monorepo this approach is valid, but for project-base, there is just `../../../..` and that would exist

packages/backend-api/install/app/AppKernel.php.patch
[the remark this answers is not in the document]
simara-svatopluk: will change
```
@@ -104,6 +107,10 @@
         }

+        if (file_exists(__DIR__ . '/config/packages/oauth2/parameters_oauth.yml')) {
+            $configs[] = __DIR__ . '/config/packages/oauth2/parameters_oauth.yml';
+        }
+
         return $configs;
     }
```

packages/backend-api/install/uninstall.sh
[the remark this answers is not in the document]
simara-svatopluk: Probably only on Mac

packages/backend-api/install/composer.json.patch
simara-svatopluk: Check this during BV
* the proposed way would result in shorter documentation and less patches
* the proposed way is not "correct" in terms of naming transitive dependencies when we depend on them directly

packages/backend-api/install/uninstall.sh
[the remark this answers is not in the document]
PetrHeinz: Repeated uninstall is not very important use-case. I'm OK with it unsupported on MacOS.

packages/backend-api/install/composer.json.patch
PetrHeinz: `trikoder/oauth2-bundle` is only important for the integration part in `shopsys/project-base` (after the installation). I'm fine with complicating the installation / uninstallation in project repository this way in exchange for the benefit of having requirements where they are supposed to be.

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,55 @@
+# Backend API Installation
+
+*Please follow these instructions with clean `git status`. If the installation fails, you can easily revert all changes and start again.*
```
grossmannmartin:
```suggestion
*Please follow these instructions with clean git working tree (you can verify with `git status`). If the installation fails, you can easily revert all changes and start again.*
```

build.xml
```diff
@@ -4,11 +4,19 @@
...
     <property name="path.framework" value="${path.packages}/framework"/>
     <property name="path.utils" value="${project.basedir}/utils"/>
     <property name="translations.dump.flags" value=""/>
 
     <import file="${path.framework}/build.xml"/>
+    <import file="${path.backend-api}/build.xml"/>
+
+    <target name="build-demo-ci" depends="backend-api-install,composer-dev,backend-api-oauth-keys-generate,shopsys_framework.build-demo-ci,backend-api-uninstall" description="Builds application for development with clean demo DB and runs CI checks."/>
+
+    <target name="build-demo-dev" depends="backend-api-install,composer-dev,backend-api-oauth-keys-generate,shopsys_framework.build-demo-dev,backend-api-uninstall" description="Builds application for development with clean demo DB and runs checks on changed files."/>
```
grossmannmartin: I'm not happy with composer-dev running here and again in `shopsys_framework.build-demo-dev`.
But I understand why it's needed...

packages/backend-api/install/install.sh
```diff
@@ -0,0 +1,100 @@
...
+apply_patch "app/config/parameters_common.yml"
+apply_patch "app/config/routing.yml"
+apply_patch "app/config/packages/security.yml"
+apply_patch "src/Shopsys/ShopBundle/Resources/config/routing.yml"
+apply_patch "app/AppKernel.php"
+apply_patch "build.xml"
+
+if [ "$1" == "monorepo" ]
+then
+    echo "Running from monorepo, not applying patch for project-base composer.json because it has no effect in monorepo. Also composer is changed during API installation in a different way (composer require)."
```
grossmannmartin: I don't understand the second sentence (Also composer is changed during API installation in a different way (composer require).)
Is it something really important or can it be omitted?

build.xml
simara-svatopluk: before we can call `backend-api-oauth-keys-generate` we must be sure composer is installed, that's why.
Phing is also clever - it runs dependencies only once, so `composer-dev` runs only once - before `backend-api-oauth-key` and never again

packages/backend-api/install/install.sh
simara-svatopluk: omited :+1:

build.xml
grossmannmartin: Oh really? That's nice!

PetrHeinz: @grossmannmartin It's true, but only among dependencies of a single target. If you call multiple targets at once (via `phing target1 target2`) or use `<phingcall ... />` for a subcall, Phing will call the same target multiple times.

packages/backend-api/install/app/AppKernel.php.patch
```diff
@@ -0,0 +1,36 @@
...
+             new Symfony\Bundle\AsseticBundle\AsseticBundle(),
+@@ -42,6 +45,7 @@
+             new Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle(),
+             new Symfony\Bundle\TwigBundle\TwigBundle(),
+             new Symfony\Cmf\Bundle\RoutingBundle\CmfRoutingBundle(),
++            new Trikoder\Bundle\OAuth2Bundle\TrikoderOAuth2Bundle(),
+             new VasekPurchart\ConsoleErrorsBundle\ConsoleErrorsBundle(),
+             new FOS\CKEditorBundle\FOSCKEditorBundle(), // has to be loaded after FrameworkBundle and TwigBundle
+             new Joschi127\DoctrineEntityOverrideBundle\Joschi127DoctrineEntityOverrideBundle(),
+@@ -104,5 +108,9 @@
```
grossmannmartin:
```suggestion
@@ -104,4 +108,8 @@
```

Before it left me untracked file `AppKernel.php.orig` on uninstallation. With the suggestion applied everything works without problem. 
(Maybe something with the empty line at the end)

simara-svatopluk: applied manually. Both versions are ok in docker

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Patches
+
+Patching files can fail. It can happen if you changed something in your project and these changes are incompatible with the patch.
+
+The installation script tries to apply patches for all involved files and applies as much as possible.
+In other words the installation script doesn't stop on the first patch error.
+
+In case of error you'll see error messages like `Patch for app/AppKernel.php cannot be applied!`.
+In such case please take a look into [install directory](https://github.com/shopsys/backend-api/install), and apply all patches manually.
```
PetrHeinz: The link is broken (and will be even after merge and split).

Either change it to `/packages/backend-api/install/` (which is preferred as it will point out to the directory in current version of the docs) or `https://github.com/shopsys/backend-api/tree/master/install` (which will not work until the `master` will start handing `v8.0.0-dev` after `v7.3.0` release)

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,55 @@
+# Backend API Installation
+
+*Please follow these instructions with clean git working tree (you can verify with `git status`). If the installation fails, you can easily revert all changes and start again.*
+
+*If you installed the Shopsys Framework with docker please run all commands in the `php-fpm` docker container*
+
+Require `shopsys/backend-api` package in the `composer.json` or run
+
+```bash
+composer require shopsys/backend-api
```
PetrHeinz: This will not work until we release `v8.0.0`. Composer will fail:
```
  [InvalidArgumentException]                                                                                           
  Could not find a version of package shopsys/backend-api matching your minimum-stability (stable).
  Require it with an explicit version constraint allowing its desired stability.   
```
I am OK with this behavior as the error message is pretty explanatory and issues like this can be handled easily. Compatibility with the user's current version will be handled by Composer as `shopsys/backend-api` is dependent on `shopsys/framework` in a particular version (or version range).

docs/backend-api/backend-api-installation.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Backend API installation is fully automated, so you have to run only (you have to be in the project root directory)
+
+```bash
+./vendor/shopsys/backend-api/install/install.sh
```
PetrHeinz: It would be nice to register the script as [a Composer vendor binary](https://getcomposer.org/doc/articles/vendor-binaries.md) for easier and more standard usage, but that's just a nice-to-have.

packages/backend-api/build.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project name="backend_api" default="list">
+    <property name="path.openssl.executable" value="openssl"/>
+    <property name="path.defuse-key-generate.executable" value="${path.bin}/generate-defuse-key"/>
+
+    <target name="backend-api-uninstall" description="Uninstall backend API from the project" hidden="true">
+        <exec executable="${path.backend-api}/install/uninstall.sh" checkreturn="true" passthru="true">
+            <arg value="monorepo"/>
```
PetrHeinz: I didn't know that the `backend-api-(un)install` (originally `monorepo-backend-api-(un)install`) targets were defined here. As this build.xml gets included even in project repositories, the argument should be defined via property overriden in monorepo's `build.xml`:

```xml
<project name="backend_api" default="list">
    <property name="path.openssl.executable" value="openssl"/>
    <property name="path.defuse-key-generate.executable" value="${path.bin}/generate-defuse-key"/>
    <property name="backend-api.install-sh.arg" value=""/>

    <target name="backend-api-uninstall" description="Uninstall backend API from the project" hidden="true">
        <exec executable="${path.backend-api}/install/uninstall.sh" checkreturn="true" passthru="true">
            <arg value="${backend-api.install-sh.arg}"/>
         </exec>
    </target>
```

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project name="Shopsys Framework" default="list">

    <property name="path.root" value="${project.basedir}/project-base"/>
    <property name="path.vendor" value="${project.basedir}/vendor"/>
    <property name="path.packages" value="${project.basedir}/packages"/>
    <property name="path.framework" value="${path.packages}/framework"/>
    <property name="path.utils" value="${project.basedir}/utils"/>
    <property name="translations.dump.flags" value=""/>
    <property name="backend-api.install-sh.arg" value="monorepo"/>

```

PetrHeinz: Simply moving them to monorepo would make the most sense.

.travis.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     include:
     -   stage: Tests
         name: "Smoke, Functional and Acceptance Tests"
-        script: docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev-quick error-pages-generate checks-ci
+        script: docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev-quick backend-api-install backend-api-oauth-keys-generate error-pages-generate checks-ci
```
PetrHeinz: I think that [the Travis build fails](https://travis-ci.org/shopsys/shopsys/jobs/553287934) because you first run `build-demo-dev-quick` which executes DB migrations without the API bundle, then it installs the API bundle and does not run the additional migrations. Maybe add `db-migrations` before the `checks-ci`? Maybe `clean-redis` will be needed as well as the Doctrine metadata might get cached...

Not sure why it failed now though.

packages/backend-api/build.xml
```diff
@@ -0,0 +1,38 @@
...
+        <copy file="${path.app}/config/packages/oauth2/parameters_oauth.yml.dist" tofile="${path.app}/config/packages/oauth2/parameters_oauth.yml" overwrite="true">
+            <filterchain>
+                <replacetokens begintoken="%%" endtoken="%%">
+                    <token key="encryption_key" value="${encryption_key}"/>
+                </replacetokens>
+            </filterchain>
+        </copy>
+    </target>
+
+    <target name="list" description="Hidden target to make Phing list all targets when called without an argument." hidden="true">
```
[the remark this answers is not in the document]
simara-svatopluk: The phing plugin in the phpStorm suggested to add a `default` option, so I copied the list and default from `framework`. But it's not necessary, so I'm removing it

Issue #1108, opened 2019-06-06
make redis cache handling consistent


boris-brtan:
### What is happening

Now there seems to be inconsistency from the view of the redis-check redis-clean and redis-clean-old-version

- CheckRedisCommand
    ```yaml
    Shopsys\FrameworkBundle\Command\CheckRedisCommand:
        arguments:
            $cacheClients:
            - '@snc_redis.bestselling_products'
            - '@snc_redis.doctrine_metadata'
            - '@snc_redis.doctrine_query'
            - '@snc_redis.session'
           # missing framework_annotations
    ```
    ```php
    foreach ($this->cacheClients as $cacheClient) {
            try {
                $cacheClient->ping();
    ```
- RedisCleanCacheCommand
   ```yaml
    Shopsys\FrameworkBundle\Component\Redis\RedisFacade:
        arguments:
            $cacheClients:
            - '@snc_redis.bestselling_products'
            - '@snc_redis.doctrine_metadata'
            - '@snc_redis.doctrine_query'
   ```
    ```php
    $this->redisFacade->cleanCache();
        foreach ($this->cacheClients as $redis) {
            $prefix = (string)$redis->getOption(Redis::OPT_PREFIX);
            $pattern = $prefix . '*';
            if (!$this->hasAnyKey($redis, $pattern)) {
                continue;
            }
            $redis->eval("return redis.call('del', unpack(redis.call('keys', ARGV[1])))", [$pattern]);
        }
    ```
- RedisCleanCacheOldCommand
   ```yaml
    Shopsys\FrameworkBundle\Component\Redis\RedisVersionsFacade:
        arguments:
            $globalClient: '@snc_redis.global'
            $currentVersion: '%build-version%'
   ```
   ```php
   $this->redisVersionsFacade->cleanOldCache();
        do {
            $keys = $this->globalClient->scan($iterator, $versionPattern);

            if ($keys !== false) {
                $toRemove = [];

                foreach ($keys as $key) {
                    if (strpos($key, $currentVersionPrefix) === false) {
                        $keyWithoutPrefix = substr($key, strlen($prefix)); 
                        $toRemove[] = $keyWithoutPrefix;
                    }
                }
                $this->globalClient->unlink($toRemove);
   ```
As we can see from the check and clean commands, redis cache client can be stored on different redis servers, but as for the redis-clean-old-version removing of old cache is done just via global client.
[...]

PR #1158, opened 2019-06-25
[framework] phing targets redis-check and clean-redis clean also items of framework_annotations client

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| clean-redis and redis-check phing targets should use also framework_annotations snc_redis client
|New feature| No 
|Fixes issues| just partially #1108 

### DIFF
--- a/packages/framework/src/Resources/config/services.yml
+++ b/packages/framework/src/Resources/config/services.yml
@@ -975,9 +975,9 @@ services:
         arguments:
             $cacheClients:
             - '@snc_redis.bestselling_products'
-            - '@snc_redis.framework_annotations'
             - '@snc_redis.doctrine_metadata'
             - '@snc_redis.doctrine_query'
+            - '@snc_redis.framework_annotations'
 
     Shopsys\FrameworkBundle\Component\Redis\RedisVersionsFacade:
         arguments:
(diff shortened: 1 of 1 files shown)


henzigo:
Please do not fix it by this way. On projects you will always add new client. unfortunately they will not know about this shitty part of implementation. As we have troble with that on our internal project. I have no idea who can throw sticks like that under programmers feet

henzigo:
Please consider the consequences of this implementation. You can also contact Petr Jasnik for more details.

boris-brtan:
@LukasHeinz this doesn't fix the issue, it just fixes invalid state where one snc_redis client items were not cleaned.
The solution for fixing, or implementing the snc_redis in other way need to be more complex

pk16011990:
This solves the consequence and not the cause... My point of view is that it is only hotfix.

Whatever If it is good enought for you...

pk16011990:
If you want to solve this issue by this way, It should contain regres test that checks array $cacheClients contains all cache providers...

PetrHeinz:
@boris-brtan @pk16011990 @henzigo I've tried to come up with a slightly diffrerent solution for this problem, but it introduces a BC break - we could either try to make it backward-compatible or retarget to 8.0

I'm just curious about your feedback on such a solution.

pk16011990:
THX @PetrHeinz,  #1161 looks that solves it for me.

PR #1161, opened 2019-06-26
[framework] RedisFacade: unification of used clients

PetrHeinz:
- all `Redis` clients are injected into the facade via service tag
- clients that should not be cleaned are injected via manual service configuration, minimizing errors
- `CheckRedisCommand` uses `RedisFacade` for the pinging
- mentioning a Redis client that was not defined in previous versions (`@snc_redis.global` or `@snc_redis.framework_annotations`) explicitly in arguments would cause a BC break - this issue has been resolved by `RedisFacadeClientFilterCompilerPass` that removes such undefined services from `RedisFacade`

| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR tries to resolve the same issue as in #1158, but tries to do so without expecting users to hard-code everything, making the solution less error-prone (hopefully).
|New feature| No 
|Fixes issues| just partially #1108 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -78,6 +78,17 @@ There you can find links to upgrade notes for other versions too.
 - use automatic wiring of Redis clients for easier checking and cleaning ([#1161](https://github.com/shopsys/shopsys/pull/1161))
     - if you have redefined the service `@Shopsys\FrameworkBundle\Component\Redis\RedisFacade` or `@Shopsys\FrameworkBundle\Command\CheckRedisCommand` in your project, or you instantiate the classes in your code:
         - instead of instantiating `RedisFacade` with an array of cache clients to be cleaned by `php phing redis-clean`, pass an array of all redis clients and another array of redis clients you don't want to clean (eg. `global` and `session`)
+            ```diff
+                Shopsys\FrameworkBundle\Component\Redis\RedisFacade:
+                    arguments:
+            -           - '@snc_redis.doctrine_metadata'
+            -           - '@snc_redis.doctrine_query'
+            -           - '@snc_redis.my_custom_cache'
+            +           $allClients: !tagged snc_redis.client
+            +           $persistentClients:
+            +               - '@snc_redis.global'
+            +               - '@snc_redis.session'
+            ```
             - this allows you to use `!tagged snc_redis.client` in your DIC config for the first argument, ensuring that newly created clients will be registered by the facade
         - instead of instantiating `CheckRedisCommand` with an array of redis clients, pass an instance of `RedisFacade` instead
     - modify the functional test `\Tests\ShopBundle\Functional\Component\Redis\RedisFacadeTest` so it creates `RedisFacade` using the two arrays and add a new test case `testNotCleaningPersistentClient`
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -75,6 +75,13 @@ There you can find links to upgrade notes for other versions too.
     - stop using the deprecated method `\Shopsys\FrameworkBundle\Model\Pricing\PriceCalculation::getVatCoefficientByPercent()`, use `PriceCalculation::getVatAmountByPriceWithVat()` for VAT calculation instead
     - if you want to customize the VAT calculation (eg. revert it back to the previous implementation), extend the service `@Shopsys\FrameworkBundle\Model\Pricing\PriceCalculation` and override the method `getVatAmountByPriceWithVat()`
     - if you created new tests regarding the price calculation they might start failing after the upgrade - in such case, please see the new VAT calculation and change the tests expectations accordingly
+- use automatic wiring of Redis clients for easier checking and cleaning ([#1161](https://github.com/shopsys/shopsys/pull/1161))
+    - if you have redefined the service `@Shopsys\FrameworkBundle\Component\Redis\RedisFacade` or `@Shopsys\FrameworkBundle\Command\CheckRedisCommand` in your project, or you instantiate the classes in your code:
+        - instead of instantiating `RedisFacade` with an array of cache clients to be cleaned by `php phing redis-clean`, pass an array of all redis clients and another array of redis clients you don't want to clean (eg. `global` and `session`)
+            - this allows you to use `!tagged snc_redis.client` in your DIC config for the first argument, ensuring that newly created clients will be registered by the facade
+        - instead of instantiating `CheckRedisCommand` with an array of redis clients, pass an instance of `RedisFacade` instead
+    - modify the functional test `\Tests\ShopBundle\Functional\Component\Redis\RedisFacadeTest` so it creates `RedisFacade` using the two arrays and add a new test case `testNotCleaningPersistentClient`
+        - you can copy-paste the [`RedisFacadeTest`](https://github.com/shopsys/project-base/blob/master/tests/ShopBundle/Functional/Component/Redis/RedisFacadeTest.php) from `shopsys/project-base`
 
 ### Configuration
 - update `phpstan.neon` with following change to skip phpstan error ([#1086](https://github.com/shopsys/shopsys/pull/1086))
--- a/packages/framework/src/Command/CheckRedisCommand.php
+++ b/packages/framework/src/Command/CheckRedisCommand.php
@@ -2,6 +2,7 @@
 
 namespace Shopsys\FrameworkBundle\Command;
 
+use Shopsys\FrameworkBundle\Component\Redis\RedisFacade;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -14,18 +15,18 @@ class CheckRedisCommand extends Command
     protected static $defaultName = 'shopsys:redis:check-availability';
 
     /**
-     * @var \Redis[]
+     * @var \Shopsys\FrameworkBundle\Component\Redis\RedisFacade
      */
-    protected $cacheClients;
+    protected $redisFacade;
 
     /**
-     * @param \Redis[] $cacheClients
+     * @param \Shopsys\FrameworkBundle\Component\Redis\RedisFacade $redisFacade
      */
-    public function __construct(array $cacheClients)
+    public function __construct(RedisFacade $redisFacade)
     {
-        $this->cacheClients = $cacheClients;
-
         parent::__construct();
+
+        $this->redisFacade = $redisFacade;
     }
 
     protected function configure()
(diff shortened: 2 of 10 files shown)


packages/framework/src/Resources/config/services.yml
```diff
@@ -973,10 +973,10 @@ services:
 
     Shopsys\FrameworkBundle\Component\Redis\RedisFacade:
         arguments:
-            $cacheClients:
-            - '@snc_redis.bestselling_products'
-            - '@snc_redis.doctrine_metadata'
-            - '@snc_redis.doctrine_query'
+            $allClients: !tagged snc_redis.client
+            $persistentClients:
+                - '@snc_redis.global'
```
boris-brtan: from the view of RedisFacade it is persistentClient but from the view of `RedisVersionsFacade` it is not.

PetrHeinz: From the point of view of `RedisVersionsFacade`, it's still `$globalClient` and the facade knows what should get removed and what should not. Maybe `$persistentClient` is not a good name - do you have any suggestions?

boris-brtan: i do not.

PetrHeinz:
I've removed the BC breaking part by introducing fallbacks. If this approach is OK, I will add upgrade notes explaining the new way of configuration.

The `@snc_redis.global` is not registered in `v7.0.0`, which seems like BC breaking, but the same problem is with unregistered `@snc_redis.framework_annotations` service, so at least I don't add a new BC issue. We can tackle this problem independently.

PetrHeinz:
I try to tackle the problem of missing services in #1167

packages/framework/src/Command/CheckRedisCommand.php
```diff
@@ -2,30 +2,45 @@
...
     /**
      * @var string
      */
     protected static $defaultName = 'shopsys:redis:check-availability';
 
     /**
-     * @var \Redis[]
+     * @var \Shopsys\FrameworkBundle\Component\Redis\RedisFacade|\Redis[]
      */
-    protected $cacheClients;
```
boris-brtan: removed protected property of non final class

packages/framework/src/Component/Redis/RedisFacade.php
```diff
@@ -11,19 +11,38 @@ class RedisFacade
     /**
      * @var \Redis[]
      */
-    protected $cacheClients;
```
boris-brtan: removed protected property of non final class

packages/framework/src/Resources/config/services.yml
```diff
@@ -973,11 +977,10 @@ services:
 
     Shopsys\FrameworkBundle\Component\Redis\RedisFacade:
         arguments:
-            $cacheClients:
-            - '@snc_redis.bestselling_products'
-            - '@snc_redis.doctrine_metadata'
-            - '@snc_redis.doctrine_query'
-            - '@snc_redis.framework_annotations'
+            $allClients: !tagged snc_redis.client
```
boris-brtan: will this change be part of upgrade notes ?

packages/framework/src/Resources/config/services.yml
```diff
@@ -1014,12 +1017,3 @@ services:
             - method: setEntityManager
 
     Shopsys\FrameworkBundle\Model\Product\Search\ProductElasticsearchConverter: ~
-
-    Shopsys\FrameworkBundle\Command\CheckRedisCommand:
```
boris-brtan: will this be also part of upgrade note, in the case it was overriden in shopBundle ???

packages/framework/src/Command/CheckRedisCommand.php
PetrHeinz: fixed in e7806e5f6

packages/framework/src/Component/Redis/RedisFacade.php
PetrHeinz: fixed in d77db04ae

packages/framework/src/Resources/config/services.yml
PetrHeinz: Yes, I want to describe the new way of configuring it and how to use it correctly.

packages/framework/src/Resources/config/services.yml
PetrHeinz: Both changes will be a part of the upgrade note. If it was overriden, there are instances of Redis passed directly, which triggers the deprecation error.

PetrHeinz:
I've removed the BC break and written the upgrade notes.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -75,6 +75,13 @@ There you can find links to upgrade notes for other versions too.
     - stop using the deprecated method `\Shopsys\FrameworkBundle\Model\Pricing\PriceCalculation::getVatCoefficientByPercent()`, use `PriceCalculation::getVatAmountByPriceWithVat()` for VAT calculation instead
     - if you want to customize the VAT calculation (eg. revert it back to the previous implementation), extend the service `@Shopsys\FrameworkBundle\Model\Pricing\PriceCalculation` and override the method `getVatAmountByPriceWithVat()`
     - if you created new tests regarding the price calculation they might start failing after the upgrade - in such case, please see the new VAT calculation and change the tests expectations accordingly
+- use automatic wiring of Redis clients for easier checking and cleaning ([#1161](https://github.com/shopsys/shopsys/pull/1161))
+    - if you have redefined the service `@Shopsys\FrameworkBundle\Component\Redis\RedisFacade` or `@Shopsys\FrameworkBundle\Command\CheckRedisCommand` in your project, or you instantiate the classes in your code:
```
boris-brtan: what about `checkRedisCommandTest`

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -75,6 +75,13 @@ There you can find links to upgrade notes for other versions too.
...
+        - instead of instantiating `RedisFacade` with an array of cache clients to be cleaned by `php phing redis-clean`, pass an array of all redis clients and another array of redis clients you don't want to clean (eg. `global` and `session`)
```
boris-brtan: what about changing the argument name ?
```yaml
Shopsys\FrameworkBundle\Component\Redis\RedisFacade:`
    arguments:
        $cacheClients -> $allClients: 
```
will this work if they override service in ShopBundle ?

packages/framework/src/Component/Redis/RedisFacade.php
```diff
@@ -11,19 +11,45 @@ class RedisFacade
...
+     */
     protected $cacheClients;
 
     /**
-     * @param \Redis[] $cacheClients
+     * @param \Redis[] $allClients
+     * @param \Redis[] $persistentClients
      */
-    public function __construct(array $cacheClients)
+    public function __construct(iterable $allClients, iterable $persistentClients = [])
```
boris-brtan: shouldnt we keed there cacheClients variable name?

packages/framework/src/DependencyInjection/Compiler/RedisFacadeClientFilterCompilerPass.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
```
boris-brtan: `	declare(strict_types=1);` ???

packages/framework/src/Command/CheckRedisCommand.php
```diff
@@ -2,30 +2,55 @@
...
+    protected $redisFacade;
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Redis\RedisFacade|\Redis[] $redisFacadeOrClients
+     */
+    public function __construct($redisFacadeOrClients)
+    {
         parent::__construct();
+
+        if (!$redisFacadeOrClients instanceof RedisFacade) {
```
TomasLudvik: It is weird to have negation in if when there is also else present

docs/upgrade/UPGRADE-unreleased.md
PetrHeinz: That's a unit test in `shopsys/framework`

docs/upgrade/UPGRADE-unreleased.md
PetrHeinz: Configuring it with `$cacheClients:` would fail. I'm not sure how to deal with this because renaming parameters is not covered in our BC promise (and as such we probably shouldn't recommend users to use argument names in their configuration...)

packages/framework/src/Component/Redis/RedisFacade.php
PetrHeinz: Why? The meaning of the parameter has changed, because now `$cacheClients = $allClients - $persistentClients`. This change of meaning is still backward compatible because when you don't provide `$persistentClients` it holds true that `$cacheClients = $allClients`.

There might be a problem with the rename if they are using the named argument configuration, but I can't think of any better way how to handle this...

packages/framework/src/DependencyInjection/Compiler/RedisFacadeClientFilterCompilerPass.php
PetrHeinz: fixed in 8a02173d2

packages/framework/src/Command/CheckRedisCommand.php
PetrHeinz: You're right, especially if the else part actually contains the expected (non-deprecated) execution path. Fixed in 32626a9.

docs/upgrade/UPGRADE-unreleased.md
boris-brtan: yep this is only problem, in other way jobs for no-bc-break-checks passed so :+1:

packages/framework/src/Resources/config/services.yml
```diff
@@ -154,6 +154,10 @@ services:
             # must be ran before ProductVisibilityFacade
             - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse, priority: 20 }
 
+    Shopsys\FrameworkBundle\Command\CheckRedisCommand:
```
PetrHeinz:
```suggestion
    # this explicit definition will not be needed when the @deprecated argument type is removed
    Shopsys\FrameworkBundle\Command\CheckRedisCommand:
```

PetrHeinz:
Sorry for the bad merge, I thought I have the *Squash and Merge* option active but I haven't :slightly_frowning_face: 

```
*   4b62b31d6 (master) RedisFacade: unification of used clients (#1161)
|\  
| * 781eb4ae5 added example in upgrade note
| * cb181908a fixup! CheckRedisCommand and RedisFacade changed for backward-compatibility
| * 32626a96f fixup! CheckRedisCommand and RedisFacade changed for backward-compatibility
| * 8a02173d2 fixup! added compiler pass to avoid BC breaks because of missing redis clients in RedisFacade
| * 84ad2e97e added upgrade notes
| * f7dc6d5e1 add a not about BC in RedisCleanCacheOldCommand
| * c63e29810 added compiler pass to avoid BC breaks because of missing redis clients in RedisFacade
| * bd4df1e6d fixup! RedisFacade: unification of used clients
| * d77db04ae fixup! CheckRedisCommand and RedisFacade changed for backward-compatibility
| * e7806e5f6 fixup! CheckRedisCommand and RedisFacade changed for backward-compatibility
| * 8ab9fa515 CheckRedisCommand and RedisFacade changed for backward-compatibility
| * c2060b25e RedisFacade: unification of used clients
* |   771f5b752 remove redundant service definitions (#1166)
|\ \  
| * | e1380f921 remove service definitions that point to missing classes
| * | 4155a2e72 remove service definitions that are already auto-discovered
|/ /  
* | 9eaef07ce [framework] added immediate export to elastic on product update in administration (#1153)
* | 508e34a86 all services injected in order to prevent BC breaks are now injected same way (#1164)
|/  
* 6241405df remove unused route "front_category_panel" (#1042)
```

PR #1163, opened 2019-06-26
[project-base] use the monolog config option "excluded_http_codes" instead of "excluded_404s"

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The `excluded_404s` option is deprecated since `symfony/monolog-bundle v3.4`.
|New feature| No 
|Fixes issues| ... 


PetrHeinz:
It seems that the used version of Monolog bridge does not support the `excluded_http_codes`. We should probably resolve this in `8.0` release...

henzigo:
@grossmannmartin @malyMiso This sounds me like really important for Symfony 4

PR #1164, opened 2019-06-26
[shopsys] all services injected in order to prevent BC breaks are now injected same way

TomasLudvik:
- all services has been added to constructors but might be nullable and set via setters

| Q             | A
| ------------- | ---
|Description, reason for the PR| Service injections are now all solved same way. There are understandable deprecation errors triggered when services are injected via setter injection and users now can use constructor injection and avoid this deprecation errors.  Note: the removed method validateInjectedDependencies() does not cause a violation of a BC promise as the method was not yet part of any released version
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Router/DomainRouterFactory.php
+++ b/packages/framework/src/Component/Router/DomainRouterFactory.php
@@ -53,27 +53,37 @@ class DomainRouterFactory
      * @param \Shopsys\FrameworkBundle\Component\Router\LocalizedRouterFactory $localizedRouterFactory
      * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRouterFactory $friendlyUrlRouterFactory
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @param \Symfony\Component\HttpFoundation\RequestStack|null $requestStack
      */
     public function __construct(
         $routerConfiguration,
         LoaderInterface $configLoader,
         LocalizedRouterFactory $localizedRouterFactory,
         FriendlyUrlRouterFactory $friendlyUrlRouterFactory,
-        Domain $domain
+        Domain $domain,
+        ?RequestStack $requestStack = null
     ) {
         $this->routerConfiguration = $routerConfiguration;
         $this->configLoader = $configLoader;
         $this->localizedRouterFactory = $localizedRouterFactory;
         $this->domain = $domain;
         $this->friendlyUrlRouterFactory = $friendlyUrlRouterFactory;
+        $this->requestStack = $requestStack;
     }
 
     /**
+     * @required
      * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+     * @internal Will be replaced with constructor injection in the next major release
+     * @deprecated
      */
     public function setRequestStack(RequestStack $requestStack)
     {
-        $this->requestStack = $requestStack;
+        if ($this->requestStack === null) {
+            @trigger_error(sprintf('The %s() method is deprecated and will be removed in the next major. Use the constructor injection instead.', __METHOD__), E_USER_DEPRECATED);
+
+            $this->requestStack = $requestStack;
+        }
     }
 
     /**
--- a/packages/framework/src/Component/Router/DomainRouterFactory.php
+++ b/packages/framework/src/Component/Router/DomainRouterFactory.php
@@ -2,6 +2,7 @@
 
 namespace Shopsys\FrameworkBundle\Component\Router;
 
+use BadMethodCallException;
 use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRouterFactory;
@@ -74,11 +75,14 @@ class DomainRouterFactory
     /**
      * @required
      * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
-     * @internal Will be replaced with constructor injection in the next major release
-     * @deprecated
+     * @deprecated Will be replaced with constructor injection in the next major release
      */
     public function setRequestStack(RequestStack $requestStack)
     {
+        if ($this->requestStack !== null && $this->requestStack !== $requestStack) {
+            throw new BadMethodCallException(sprintf('Method "%s" has been already called and cannot be called multiple times.', __METHOD__));
+        }
+
         if ($this->requestStack === null) {
             @trigger_error(sprintf('The %s() method is deprecated and will be removed in the next major. Use the constructor injection instead.', __METHOD__), E_USER_DEPRECATED);
 
--- a/packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
+++ b/packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Model\Product\Search\Export;
 
-use BadMethodCallException;
 use Doctrine\ORM\EntityManagerInterface;
 use Shopsys\FrameworkBundle\Component\Console\ProgressBarFactory;
 use Shopsys\FrameworkBundle\Component\Doctrine\SqlLoggerFacade;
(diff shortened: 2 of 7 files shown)


packages/framework/src/Component/Router/DomainRouterFactory.php
```diff
@@ -53,27 +53,37 @@ class DomainRouterFactory
...
         $this->localizedRouterFactory = $localizedRouterFactory;
         $this->domain = $domain;
         $this->friendlyUrlRouterFactory = $friendlyUrlRouterFactory;
+        $this->requestStack = $requestStack;
     }
 
     /**
+     * @required
      * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+     * @internal Will be replaced with constructor injection in the next major release
```
PetrHeinz: Now, that there is an option to use the constructor injection, I would annotate it just as `@deprecated`. Previously I was hesitant to deprecate the usage if we don't provide an alternative - hence the `@internal`.

```suggestion
     * @deprecated Will be replaced with constructor injection in the next major release
```

packages/framework/src/Component/Router/DomainRouterFactory.php
```diff
@@ -53,27 +53,37 @@ class DomainRouterFactory
...
+     * @deprecated
```
PetrHeinz:
```suggestion
```

see above

packages/framework/src/Component/Router/DomainRouterFactory.php
```diff
@@ -53,27 +53,37 @@ class DomainRouterFactory
...
      */
     public function setRequestStack(RequestStack $requestStack)
     {
-        $this->requestStack = $requestStack;
+        if ($this->requestStack === null) {
+            @trigger_error(sprintf('The %s() method is deprecated and will be removed in the next major. Use the constructor injection instead.', __METHOD__), E_USER_DEPRECATED);
+
+            $this->requestStack = $requestStack;
+        }
```
PetrHeinz: If you want to set it only if the property is unset, I would definitely go with throwing an exception if `$this->requestStack !== $requestStack` at the end of the method.

Otherwise, you could call the setter injector twice with different argument like this:
```php
$domainRouterFactory = new DomainRouterFactory(...);
$domainRouterFactory->setRequestStack($firstRequestStack);
$domainRouterFactory->setRequestStack($secondRequestStack);
```
The code would not throw an exception, but the `$firstRequestStack` would get used, which is, in my point of view, counter-intuitive.

PetrHeinz: Or at the beggining of the method but inside `if ($this->requestStack !== null && $this->requestStack !== $requestStack)`...

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
```diff
@@ -180,22 +192,4 @@ protected function removeNotUpdated(int $domainId, array $exportedIds): void
     {
         $this->productElasticsearchRepository->deleteNotPresent($domainId, $exportedIds);
     }
-
-    /**
-     * @internal Will be removed in the next major release
-     */
-    protected function validateInjectedDependencies(): void
```
PetrHeinz: Please clarify in the description, that removal of this method is not a BC break as it was introduced in https://github.com/shopsys/shopsys/commit/7919c35af5f8ae499c4dfe6d7ad2b46daebc7302 which hasn't been released yet.

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
```diff
@@ -47,27 +47,35 @@ class ProductElasticsearchRepository
...
+     * @internal Will be replaced with constructor injection in the next major release
+     * @deprecated
      */
     protected function createFilterQueryFactory(): FilterQueryFactory
     {
-        return new FilterQueryFactory();
+        if ($this->filterQueryFactory === null) {
+            @trigger_error(sprintf('The %s() method is deprecated and will be removed in the next major. Use the constructor injection instead.', __METHOD__), E_USER_DEPRECATED);
+
+            $this->filterQueryFactory = new FilterQueryFactory();
```
PetrHeinz: The instance should be returned instead. This isn't a setter injector, but a factory method. It's signature must not change as it was already released in https://github.com/shopsys/shopsys/commit/95b111b7a0bcc3787e3158627a202c83e881986f

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
```diff
@@ -47,27 +47,35 @@ class ProductElasticsearchRepository
...
-        ElasticsearchStructureManager $elasticsearchStructureManager
+        ElasticsearchStructureManager $elasticsearchStructureManager,
+        ?FilterQueryFactory $filterQueryFactory = null
     ) {
         $this->indexPrefix = $indexPrefix;
         $this->client = $client;
         $this->productElasticsearchConverter = $productElasticsearchConverter;
         $this->elasticsearchStructureManager = $elasticsearchStructureManager;
-        $this->filterQueryFactory = $this->createFilterQueryFactory();
+        $this->filterQueryFactory = $filterQueryFactory;
```
PetrHeinz:
```suggestion
        $this->filterQueryFactory = $filterQueryFactory ?? $this->createFilterQueryFactory();
```
It seems that you have misunderstood the usage - `createFilterQueryFactory` is not a setter injector.

packages/framework/src/Resources/config/services.yml
```diff
@@ -465,10 +465,6 @@ services:
         arguments:
             - '@logger'
 
-    Shopsys\FrameworkBundle\Model\Mail\MailTemplateFacade:
```
PetrHeinz: You shouldn't remove a service definition because of backward-compatibility, even if it's not currently specified in our BC promise...

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
PetrHeinz: Adding the `@required` annotation to this method is what cause the Travis to fail.

Personally, I would revert the changes to this method with the exception of the added deprecation warning - that's useful.

packages/framework/src/Resources/config/services.yml
PetrHeinz: I've created https://github.com/shopsys/shopsys/issues/1165 to address the incomplete BC promise.

PetrHeinz:
```suggestion
    Shopsys\FrameworkBundle\Model\Mail\MailTemplateFacade: ~ # @deprecated
```

grossmannmartin: Even when the facade is defined automatically with the suffix?

PetrHeinz: I haven't thought of that, good point :+1:

It got me thinking whether we have more such redundancies, so I opened https://github.com/shopsys/shopsys/pull/1166

packages/framework/src/Component/Router/DomainRouterFactory.php
TomasLudvik: Changed

packages/framework/src/Component/Router/DomainRouterFactory.php
TomasLudvik: Changed

packages/framework/src/Component/Router/DomainRouterFactory.php
TomasLudvik: Updated

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
TomasLudvik: I do not think this is necessary

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
TomasLudvik: Done

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
TomasLudvik: Done

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
grossmannmartin: Short note in PR description is not a bad idea. 
_Note: removed methods does not cause violation of a BC promise as these methods was not yet part of any released version_

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
```diff
@@ -47,26 +47,32 @@ class ProductElasticsearchRepository
...
+        $this->filterQueryFactory = $filterQueryFactory ?? $this->createFilterQueryFactory();
+
     }
 
     /**
+     * @required
      * @return \Shopsys\FrameworkBundle\Model\Product\Search\FilterQueryFactory
      * @deprecated Will be replaced with constructor injection in the next major release
      */
     protected function createFilterQueryFactory(): FilterQueryFactory
```
grossmannmartin: Method has to be public to use `@required` annotation, but according to https://symfony.com/doc/current/contributing/code/bc.html#changing-classes this type of change is a violation of BC promise. 

But if you call it directly in constructor, you don't need `@required` at all, right?

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
TomasLudvik: Added

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
TomasLudvik: You are absolutely right, I forgot it there

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
```diff
@@ -47,18 +47,21 @@ class ProductElasticsearchRepository
...
+        ElasticsearchStructureManager $elasticsearchStructureManager,
+        ?FilterQueryFactory $filterQueryFactory = null
     ) {
         $this->indexPrefix = $indexPrefix;
         $this->client = $client;
         $this->productElasticsearchConverter = $productElasticsearchConverter;
         $this->elasticsearchStructureManager = $elasticsearchStructureManager;
-        $this->filterQueryFactory = $this->createFilterQueryFactory();
+        $this->filterQueryFactory = $filterQueryFactory ?? $this->createFilterQueryFactory();
+
```
PetrHeinz:
```suggestion
```

PR #930, opened 2019-04-04
[shopsys] use redis for doctrine and framework cache

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Moving doctrine and framework cache to redis storage will increase performance.
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| ... 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1167, opened 2019-06-27
[framework] defining empty services of Redis clients for BC

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Some Redis clients were added via configration in `shopsys/project-base` in previous versions (in #886 and #930). Projects may fail after upgrading on undefined services when generating the DIC. This PR avoids this problem by defining an empty service. Some work might still be needed to make those empty services behave correctly when called without Redis client configuration.<br>The backward compatibility should be tested thoroughly and the consequences of the missing clients should be explored.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Resources/config/services.yml
+++ b/packages/framework/src/Resources/config/services.yml
@@ -115,6 +115,16 @@ services:
         arguments:
             - '@snc_redis.framework_annotations'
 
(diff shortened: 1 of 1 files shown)


PetrHeinz:
This isn't a right way - the empty Redis client fail on essentially any method call and it's not straight-forward to implement some kind of `NoopRedisClient`... We're gonna have to find another way.

PR #1168, opened 2019-06-27
[shopsys] product does not open in new window

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Product detail was only page opening in new window. This behavior was weird and not well accepted by some customers so it has been removed.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Controller/Admin/ProductController.php
+++ b/packages/framework/src/Controller/Admin/ProductController.php
@@ -165,13 +165,16 @@ class ProductController extends AdminBaseController
         if ($form->isSubmitted() && $form->isValid()) {
             $this->productFacade->edit($id, $form->getData());
 
-            $this->getFlashMessageSender()->addSuccessFlashTwig(
-                t('Product <strong>{{ product|productDisplayName }}</strong> modified'),
-                [
-                    'product' => $product,
-                ]
-            );
-            return $this->redirectToRoute('admin_product_edit', ['id' => $product->getId()]);
+            $this->getFlashMessageSender()
+                ->addSuccessFlashTwig(
+                    t('Product <strong><a href="{{ url }}">{{ product|productDisplayName }}</a></strong> modified'),
+                    [
+                        'product' => $product,
+                        'url' => $this->generateUrl('admin_product_edit', ['id' => $product->getId()]),
+                    ]
+                );
+
+            return $this->redirectToRoute('admin_product_list');
         }
 
         if ($form->isSubmitted() && !$form->isValid()) {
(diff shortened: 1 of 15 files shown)


packages/framework/src/Controller/Admin/ProductController.php
```diff
@@ -165,13 +165,16 @@ public function editAction(Request $request, $id)
...
             $this->productFacade->edit($id, $form->getData());
 
-            $this->getFlashMessageSender()->addSuccessFlashTwig(
-                t('Product <strong>{{ product|productDisplayName }}</strong> modified'),
-                [
-                    'product' => $product,
-                ]
-            );
-            return $this->redirectToRoute('admin_product_edit', ['id' => $product->getId()]);
+            $this->getFlashMessageSender()
```
simara-svatopluk: this behavior change is not explained in the US. I didn't find a criteria that would say "if the product is successfully edited, redirect to list". Am I right?

simara-svatopluk: The same applies for a new product

packages/framework/src/Resources/scripts/admin/product.js
```diff
@@ -19,11 +19,6 @@
         Shopsys.product.toggleIsUsingAlternateAvailability($outOfStockActionSelection.val() === Shopsys.constant('\\Shopsys\\FrameworkBundle\\Model\\Product\\Product::OUT_OF_STOCK_ACTION_SET_ALTERNATE_AVAILABILITY'));
 
         Shopsys.product.initializeSideNavigation();
-
-        $('#js-close-without-saving').on('click', function () {
```
simara-svatopluk: I'm not sure whether it is a good idea to remove such functionality, the implementator can rely on it. If you need to remove it, please create upgrade instructions

TomasLudvik: I have added upgrade note

packages/framework/src/Controller/Admin/ProductController.php
TomasLudvik: Well, reason for this was that product stuff in admin behaved different way than the others. I believe this is right and will discuss during BV, are you ok with that?

simara-svatopluk:
Tests:
* the link for "closing without change" in product detail is now "back to overview", please revert the naming
* I tried to dump translations, but target `dump-translation` fails. Did you try to dump translations? Becuase there could be a new ones
* headline of product edit should be "Editing product: ", but you have "Editing product -"

packages/framework/src/Controller/Admin/ProductController.php
simara-svatopluk: sure

packages/framework/src/Resources/scripts/admin/product.js
simara-svatopluk: content is correct, but wrong upgrade.md, please move it to `upgrade-v8.0.0.md`

TomasLudvik: Oh you are right!

TomasLudvik:
> Tests:
> 
> * the link for "closing without change" in product detail is now "back to overview", please revert the naming
> * I tried to dump translations, but target `dump-translation` fails. Did you try to dump translations? Becuase there could be a new ones
> * headline of product edit should be "Editing product: ", but you have "Editing product -"

* same as in case of redirecting -> this should be unified as I did
* I have dumped translations (will create PR for fixing fail about dumping translations)
* again headline is same as for other entities (order, article ...)

simara-svatopluk:
Please discuss the textation of "back to overview" and format of headline during BV, because there is no information about this in the acceptance criteria

packages/framework/src/Controller/Admin/ProductController.php
TomasLudvik: @PetrHeinz our current business validator agreed with me

Issue #1098, opened 2019-06-01
Override/extend twig extensions is not possible


henzigo:
### What is happening

I'm not able to override twig extensions due to private methods that are used in extensions. When I need to change part of an extension or add a new function/filter, then I need to copy all class.

Typical use-case is when I want to implement image lazy load and I need to change attribute `src` to `data-original`. I have to copy whole `ImageExtension` class to change that.

### Expected result

Extensions should be better extendable/overridable.

PR #1173, opened 2019-06-27
[shopsys] all shopsys packages now use protected as highest visibility of properties and methods and use static instead of self

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Framework should be as much extendable as possible.
|New feature| No 
|Fixes issues| ... 


PetrHeinz:
This really increases the scope of our BC promise. It will be very hard to keep backward compatibility of all the protected function / methods / constants. As such, I am personally against such broad changes. I have not heard any problems with extensibility in these packages (#1098 was the last such feedback). The model namespace is always public or protected (eg. in feeds).

Also, it surprises me that you don't propose to change the analyzed namespaces in `shopsys/framework`.

LukasHeinz:
@TomasLudvik please can you react to Peter's concerns?

PR #1177, opened 2019-07-01
[shopsys] build.xml: build-deploy-* are fixed

boris-brtan:
- redis-check was moved from part-1-db-independet into part-2-db-dependent
- part-1-db-independent was used in production stage build in Dockerfile for php-fpm image so installation guide now make sense and build-version-generate is contained

| Q             | A
| ------------- | ---
|Description, reason for the PR| installation guide for production was not accurate since build-version was introduced and other thins were bound to it. Also production stage build is now simplified, since it is dependent on the things that were removed from old version of installation guide for production. phing target `build-deploy-part-1-db-independent` was fixed since redis-check is db-dependent target.
|New feature| No 
|Fixes issues|  

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -8,6 +8,13 @@ There you can find links to upgrade notes for other versions too.
 ## [shopsys/framework]
 
 ### Infrastructure
+- update your `docker/php-fpm/Dockerfile` production stage build ([#1177](https://github.com/shopsys/shopsys/pull/1177))
+    ```diff
+    RUN composer install --optimize-autoloader --no-interaction --no-progress --no-dev
+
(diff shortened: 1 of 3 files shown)


henzigo:
Why you need to call `redis-check` command separately, when all redis-* commands depends on `redis-check`. See https://github.com/shopsys/shopsys/blob/master/packages/framework/build.xml#L141

PetrHeinz:
@henzigo The reason is that using this approach we can first check whether Redis is ready and in case it isn't, abort the operation before DB migrations are executed.

boris-brtan:
hi @henzigo , i do not understand your comment. redis-check was moved from db-independent into db-dependent since redis is kind of db and redis-check will be now in the correct phing target as all redis-* targets. or ?

PR #1180, opened 2019-07-02
[shopsys] [BC-BREAK] fixed wrong field in exporting products to Elasticsearch

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Changed `shortDescription` to `short_description` in the export of products. It was already called `short_description` in the structure definition and queries and all other fields use underscore.
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Yes - functional tests will fail after upgrade (the fix is very easy to do and doing this the backward-compatible way would be harder and more confusing for both us and the end users) 
|Fixes issues| Fixes searching in `Product::$shortDescription` (for this to be fixed, a new product export must be executed) 

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -39,6 +39,13 @@ There you can find links to upgrade notes for other versions too.
     - if you deploy to the google cloud, copy new [`.ci/deploy-to-google-cloud.sh`](https://github.com/shopsys/project-base/blob/master/.ci/deploy-to-google-cloud.sh) script from `shopsys/project-base` ([#1126](https://github.com/shopsys/shopsys/pull/1126))
 
 ### Application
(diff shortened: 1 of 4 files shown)


boris-brtan:
hi @PetrHeinz @DavidOstrozlik , it seems that for this BC-break you need to set 
```
    Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportRepository:
        alias: Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExportWithFilterRepository
```
is that BC-break or not? because implementation of elasticsearch via projectbase alias i think is non-bc-breaking

PetrHeinz:
@boris-brtan It's BC breaking for projects that use Elasticsearch filtering introduced in v7.2.0.

PR #1189, opened 2019-07-04
[framework] UploadedFileDeleteDoctrineListener not throw not found enitity 

stanoMilan:
| Q             | A
| ------------- | ---
|Description| if add entity to config for uploading file and then i have optional file for this entity .. remove from db throw not found exception
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/UploadedFile/UploadedFileDeleteDoctrineListener.php
+++ b/packages/framework/src/Component/UploadedFile/UploadedFileDeleteDoctrineListener.php
@@ -36,8 +36,10 @@ class UploadedFileDeleteDoctrineListener
     {
         $entity = $args->getEntity();
         if ($this->uploadedFileConfig->hasUploadedFileEntityConfig($entity)) {
-            $uploadedFile = $this->uploadedFileFacade->getUploadedFileByEntity($entity);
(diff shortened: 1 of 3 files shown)


stanoMilan:
Example 

1.  I create new entity
2.  This entity add to config for upladed file (uploaded_files.yml)
3. Add row to db with upload file and add row to db without uploaded file
4. If i remove row with uploaded file via doctrine this row success remove this row
5.  If i remove row without uploaded file via doctrine this row is not remove and aplication return unhandled exception 


**expected behavior:
 if i have entity, wich not required file then if remove this record then app is not return unhandled exception**

vitek-rostislav:
Hi @stanoMilan, thanks for the fix! :ok_hand: 
It seems fine to me at a first glance, I would just change the commit message to something like that:
```
UploadedFileDeleteDoctrineListener::preRemove() now tries to remove UploadedFile instance only if it exists
``` 
We need to send it to review to another colleague of mine and test it when there is some time for that (it is not planned into our current sprint so it might take a while until the PR is processed)

boris-brtan:
Hi @vitek-rostislav, can you update the PR table (description, bc-break, ....) and also title ?

PR #1193, opened 2019-07-09
[shopsys] remove deprecations from 7.x.x

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| since we are almost near the release of next major version, we need to simplify code that was implemented to apply backward compatibility promise. One deprecated thing we kept, it is CartController:addProductFormAction since it is still used in productDetail but we have a plan to reimplement it with the new CartController:productActionAction.
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Yes 
|Fixes issues|  
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1195, opened 2019-07-10
[framework] fixed memory leak in cron ProductSearchExportCronModule

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Exporting products to Elasticsearch needed more memory, the more products were present. Now the memory usage is stable and low enough (< 100MB). This solves the same problem as #1122 does but for export with cron module.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Product/Search/Export/ProductSearchExportCronModule.php
+++ b/packages/framework/src/Model/Product/Search/Export/ProductSearchExportCronModule.php
@@ -4,6 +4,9 @@ namespace Shopsys\FrameworkBundle\Model\Product\Search\Export;
 
 use Shopsys\Plugin\Cron\SimpleCronModuleInterface;
 use Symfony\Bridge\Monolog\Logger;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 class ProductSearchExportCronModule implements SimpleCronModuleInterface
 {
@@ -29,6 +32,7 @@ class ProductSearchExportCronModule implements SimpleCronModuleInterface
 
     public function run()
     {
-        $this->productSearchExportFacade->exportAll();
+        $symfonyStyle = new SymfonyStyle(new ArrayInput([]), new NullOutput());
+        $this->productSearchExportFacade->exportAllWithOutput($symfonyStyle);
     }
 }


packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
```diff
@@ -128,18 +128,25 @@ public function setEntityManager(EntityManagerInterface $entityManager): void
     /**
      * @param int $domainId
      * @param string $locale
-     * @deprecated Use `exportWithOutput` instead
      */
     public function export(int $domainId, string $locale): void
```
PetrHeinz: Now this looks like a DRY violation. Maybe it could just call the `exportWithOutput` with null IO like this:
```php
$this->exportWithOutput(new SymfonyStyle(new StringInput(''), new NullOutput()));
```

simara-svatopluk: Great idea, I just tried
```php
$symfonyStyle = new SymfonyStyle(new ArrayInput([]), new NullOutput());
$this->exportWithOutput($domainId, $locale, $symfonyStyle);
```
and works as expected... My choice is `ArrayInput` as `StringInput` is trying to parse the string and that is unnecessary operation. But in the end it sort of doesn't matter

PetrHeinz: Also we can leave that method `@deprecated` and use this approach directly in the `ProductSearchExportCronModule`, which would be more appropriate in a patch version (considering this will be a part of `v7.3.1`).

boris-brtan: so will that be deprecated or not? i need to work with that in #1193

grossmannmartin: Hi, sorry for the late reply.
I think it's better to keep the export method (= remove deprecation) so this `emptySymfonyStyle` will remain together and the public interface of a class will be more clear.

I updated the PR, please check it out ;)

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
```diff
@@ -10,6 +10,8 @@
 use Shopsys\FrameworkBundle\Component\Doctrine\SqlLoggerFacade;
 use Shopsys\FrameworkBundle\Model\Product\Search\ProductElasticsearchConverter;
 use Shopsys\FrameworkBundle\Model\Product\Search\ProductElasticsearchRepository;
+use Symfony\Component\Console\Input\ArrayInput;
```
simara-svatopluk: please remove these unused imports

packages/framework/src/Model/Product/Search/Export/ProductSearchExporter.php
grossmannmartin: Edit: after internal discussion updated according to #1193

PR #1198, opened 2019-07-10
[project-base] gke-cluster.tf: remove 'google_container_engine_versions' data source

boris-brtan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Avoids error during `terraform apply`: `google_container_cluster.primary: node_version and min_master_version must be set to equivalent values on create`, for details, see [the docs](https://www.terraform.io/docs/providers/google/r/container_cluster.html#node_version).<br>As we don't really need running on the latest version of Kubernetes, it's safer to omit the whole data source and let GKE resolve version automatically.<br>Also, it makes the config simpler, which is nice.
|New feature| No 
|Fixes issues|  

### DIFF
--- a/project-base/infrastructure/google-cloud/gke-cluster.tf
+++ b/project-base/infrastructure/google-cloud/gke-cluster.tf
@@ -5,7 +5,6 @@ data "google_container_engine_versions" "primary" {
 resource "google_container_cluster" "primary" {
   name               = "primary"
   zone               = "${data.google_container_engine_versions.primary.zone}"
-  min_master_version = "${data.google_container_engine_versions.primary.latest_master_version}"
   node_version       = "${data.google_container_engine_versions.primary.latest_node_version}"
   initial_node_count = 3
 
--- a/project-base/infrastructure/google-cloud/gke-cluster.tf
+++ b/project-base/infrastructure/google-cloud/gke-cluster.tf
@@ -1,11 +1,6 @@
-data "google_container_engine_versions" "primary" {
-  zone = "${var.GOOGLE_CLOUD_REGION}-a"
-}
-
 resource "google_container_cluster" "primary" {
   name               = "primary"
-  zone               = "${data.google_container_engine_versions.primary.zone}"
-  node_version       = "${data.google_container_engine_versions.primary.latest_node_version}"
+  zone               = "${var.GOOGLE_CLOUD_REGION}-a"
   initial_node_count = 3
 
   node_config {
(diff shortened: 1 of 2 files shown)


PetrHeinz:
The idea behind specifying the `min_master_version` and `node_version` is that they get updated to higher version than GCP would pick by default. It's true that they should not differ when both are provided.

I'm not sure whether we should omit the `min_master_version` or `node_version`. After a short search I haven't found any best practices that would describe a reason in preferring one before the other. I've noticed that in [the Terraform `google_container_engine_versions` data source example usage](https://www.terraform.io/docs/providers/google/d/google_container_engine_versions.html#example-usage) they provide `node_version` and omit the `min_master_version`.

Maybe it would also make sense to omit both of the versions (along with `google_container_engine_versions` data source) and let GKE to decide which versions to use. This would keep our config as small as possible.

boris-brtan:
since it is impossible for me to do any changes, i close this PR right now.

PetrHeinz:
Hi @boris-brtan, whould you mind if we keep this open? We can finish it ourselves as soon as we get to it as the issue might break our GCP deploy scripts.

PetrHeinz:
I've rebased and retargeted the PR to `7.3` and made the config consistent with [the example usage](https://www.terraform.io/docs/providers/google/d/google_container_engine_versions.html#example-usage) as described in https://github.com/shopsys/shopsys/pull/1198#issuecomment-511430675.

s3tezsky:
The deploy is still failing. For my purposes helped me to remove also `node_version` from `google_container_cluster`.

Error on heimdall says 
```
google_container_cluster.primary: googleapi: Error 400: Desired node version "1.14.3-gke.11" cannot be greater than current master version "1.13.7-gke.8"., badRequest
```
IMO that might be fixed with specifying higher version by setting up `version_prefix` in `google_container_engine_versions` as is in example usage pasted by @PetrHeinz right above

PetrHeinz:
@s3tezsky: So I've omitted both versions and removed the data source altogether, which removes the problem and makes the config simpler at the same time. We don't really need to run it on the highest versions of Kubernetes, let GKE handle it automatically...

I'm running tests on our Heimdall CI on basically the same commit https://github.com/shopsys/project-base/commit/55c5d21da80751411fe97a5c020f81481fe360d1...

PetrHeinz:
The deployment to GCP succeeded on Heimdall CI :heavy_check_mark:

PR #1203, opened 2019-07-15
[backend-api] create, update, delete products

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Backend API supports writing operation for products - create, update (even partially), delete
|New feature| Yes
|Fixes issues| ... 

### DIFF
--- a/docs/backend-api/api-methods.md
+++ b/docs/backend-api/api-methods.md
@@ -6,6 +6,9 @@ You need to provide token via `Authorization` header to access requested data li
 - [Retrive access token](#retrieve-access-token)
 - [Retrieve a product](#retrieve-a-product)
 - [Retrieve a list of products](#retrieve-a-list-of-products)
+- [Add product](#add-product)
+- [Partial product update](#partial-product-update)
+- [Delete product](#delete-product)
 
 **Retrieve access token**
 ----
@@ -14,22 +17,22 @@ You need to provide token via `Authorization` header to access requested data li
 
   `/api/token`
 
-* **Method:**
+* **Method**
 
   `POST`
 
 *  **Data Params**
 
-   **Required:**
+   **Required**
 
    `grant_type=client_credentials`
    `client_id=[string]`
    `client_secret=[string]`
 
-* **Success Response:**
+* **Success Response**
 
-  * **Code:** `200` <br />
-     **Content:**
+  * **Code** `200` <br />
+     **Content**
         ```
         {
             "token_type": "Bearer",
@@ -38,10 +41,10 @@ You need to provide token via `Authorization` header to access requested data li
         }
         ```
 
-* **Error Response:**
+* **Error Response**
 
-  * **Code:** `401 UNAUTHORIZED` <br />
-    **Content:**
+  * **Code** `401 UNAUTHORIZED` <br />
+    **Content**
     ```
     {
         "error": "invalid_client",
@@ -51,9 +54,9 @@ You need to provide token via `Authorization` header to access requested data li
 
   OR
 
-  * **Code:** `400 Bad Request` <br />
+  * **Code** `400 Bad Request` <br />
 
-* **Notes:**
+* **Notes**
 Token is valid for one hour.
 You will need to generate new token after the first one expires.
 
@@ -64,20 +67,20 @@ You will need to generate new token after the first one expires.
 
   `/api/v1/products/{uuid}`
 
-* **Method:**
+* **Method**
 
   `GET`
 
 *  **URL Params**
 
-   **Required:**
+   **Required**
 
    `uuid=[string]`
 
-* **Success Response:**
+* **Success Response**
 
-  * **Code:** `200` <br />
-    **Content:**
+  * **Code** `200` <br />
+    **Content**
     ```
     {
       "uuid": "2ad32b16-bfb4-4cfa-8bbe-331e630769c5",
@@ -102,25 +105,25 @@ You will need to generate new token after the first one expires.
     }
     ```
 
-* **Error Response:**
+* **Error Response**
 
-  * **Code:** `400 Bad Request` <br />
-    **Content:**
+  * **Code** `400 Bad Request` <br />
+    **Content**
     ```
     {
         "code": 400,
-        "message": "This UUID is not valid: f564da76-6d51-4ecd-b004-895c8019a23"
+        "message": "This UUID is not valid: 123xxx"
     }
     ```
 
   OR
 
-  * **Code:** `401 UNAUTHORIZED` <br />
+  * **Code** `401 UNAUTHORIZED` <br />
 
   OR
 
-  * **Code:** `404 Not Found` <br />
-    **Content:**
+  * **Code** `404 Not Found` <br />
+    **Content**
     ```
     {
         "code": 404,
@@ -135,21 +138,21 @@ You will need to generate new token after the first one expires.
 
   `/api/v1/products`
 
-* **Method:**
+* **Method**
 
   `GET`
 
 *  **URL Params**
 
-   **Optional:**
+   **Optional**
 
    `uuids=[array of strings]`
    `page=[integer]`
 
-* **Success Response:**
+* **Success Response**
 
-  * **Code:** `200` <br />
-    **Content:**
+  * **Code** `200` <br />
+    **Content**
     ```
     [
        {
@@ -176,29 +179,29 @@ You will need to generate new token after the first one expires.
        # next products here
     ]
     ```
-    **Headers:**
+    **Headers**
     In headers there is list of links to another pages with products, it can be found under Link header
     `<http://127.0.0.1:8000/api/v1/products?page=2>; rel="next", <http://127.0.0.1:8000/api/v1/products?page=2>; rel="last"`
 
-* **Error Response:**
+* **Error Response**
 
-  * **Code:** `400 Bad Request` <br />
-    **Content:**
+  * **Code** `400 Bad Request` <br />
+    **Content**
     ```
     {
         "code": 400,
-        "message": "This UUID is not valid: f564da76-6d51-4ecd-b004-895c8019a23"
+        "message": "This UUID is not valid: 123xx"
     }
     ```
 
   OR
 
-  * **Code:** `401 UNAUTHORIZED` <br />
+  * **Code** `401 UNAUTHORIZED` <br />
 
   OR
 
-  * **Code:** `422 Unprocessable Entity` <br />
-    **Content:**
+  * **Code** `422 Unprocessable Entity` <br />
+    **Content**
     ```
     {
         "code": 422,
@@ -206,6 +209,233 @@ You will need to generate new token after the first one expires.
     }
     ```
 
-* **Sample Call:**
+* **Sample Call**
 
     `http://127.0.0.1:8000/api/v1/products?uuids[]=877a07d5-a276-49f8-9b73-6ac0edef83be&uuids[]=f564da76-6d51-4ecd-b004-895c8019a235x   `
+
+**Add product**
+----
+
+* **URL**
+
+  `/api/v1/products`
+
+* **Method**
+
+  `POST`
+
+* **Headers**
+
+  `Content-Type: application/json`
+
+* **Request Body**
+
+  * **Content**
+    ```
+    {
+      "name": {
+          "en": "22\" Sencor SLE 22F46DM4 HELLO KITTY",
+          "cs": "22\" Sencor SLE 22F46DM4 HELLO KITTY"
+      },
+      "hidden": false,
+      "sellingDenied": false,
+      "sellingFrom": "2000-01-16T00:00:00+00:00",
+      "sellingTo": "2100-01-16T00:00:00+00:00",
+      "catnum": "9177759",
+      "ean": "8845781245930",
+      "partno": "SLE 22F46DM4",
+      "shortDescription": {
+          "1": "Television LED, 55 cm diagonal ...",
+          "2": "Sencor SLE 22F46DM4 Hello Kitty je ..."
+      },
+      "longDescription": {
+          "1": "Television LED, 55 cm diagonal ...",
+          "2": "<p><strong>Sencor SLE 22F46DM4 ..."
+      }
+    }
+    ```
+
+* **Request Body for specified UUID**
+
+  * **Content**
+    ```
+    {
+      "uuid": "7a34a8cf-8f3b-405d-9662-f20d983cfc8a",
+      "name": {
+          "en": "22\" Sencor SLE 22F46DM4 HELLO KITTY",
+          "cs": "22\" Sencor SLE 22F46DM4 HELLO KITTY"
+      },
+      "hidden": false,
+      "sellingDenied": false,
+      "sellingFrom": "2000-01-16T00:00:00+00:00",
+      "sellingTo": "2100-01-16T00:00:00+00:00",
+      "catnum": "9177759",
+      "ean": "8845781245930",
+      "partno": "SLE 22F46DM4",
+      "shortDescription": {
+          "1": "Television LED, 55 cm diagonal ...",
+          "2": "Sencor SLE 22F46DM4 Hello Kitty je ..."
+      },
+      "longDescription": {
+          "1": "Television LED, 55 cm diagonal ...",
+          "2": "<p><strong>Sencor SLE 22F46DM4 ..."
+      }
+    }
+    ```
+
+* **Success Response**
+
+  * **Code** `201`  
+    **Headers**
+    `Location: http://127.0.0.1:8000/api/v1/products/7a34a8cf-8f3b-405d-9662-f20d983cfc8a`
+
+* **Error Response**
+
+  * **Code** `401 UNAUTHORIZED`
+
+  OR
+
+  * **Code** `415 Unsupported Media Type`
+
+  OR
+
+  * **Code** `400 Bad Request`  
+    **Content**
+    ```
+    {
+        "message": "Provided data did not pass validation",
+        "errors": {
+            "sellingTo": "The value \"12.1.2019\" is not a valid DateTime::ATOM format.",
+            "longDescription.1": "The value 345 is not a valid string."
+        }
+    }
+    ```
+
+    OR
+
+  * **Code** `422 Unprocessable Entity`  
+    **Content**
+    ```
+    {
+        "message": "Product with 74da76db-61ab-43fb-926e-43ba326325fc UUID already exists"
+    }
+    ```
+
+**Partial product update**
+----
+
+* **URL**
+
+  `/api/v1/products/{uuid}`
+
+* **Method**
+
+  `PATCH`
+
+* **Headers**
+
+  `Content-Type: application/json`
+
+*  **URL Params**
+
+   **Required**
+
+   `uuid=[string]`
+
+* **Request Body**
+
+  * **Content**
+    ```
+    {
+      "name": {
+          "en": "22\" Sencor SLE 22F46DM4 HELLO KITTY",
+          "cs": "22\" Sencor SLE 22F46DM4 HELLO KITTY"
+      },
+      "hidden": false,
+      "shortDescription": {
+          "1": "Television LED, 55 cm diagonal ..."
+      }
+    }
+    ```
+
+* **Success Response**
+
+  * **Code** `204`
+
+* **Error Response**
+
+  * **Code** `401 UNAUTHORIZED`
+
+  OR
+
+  * **Code** `415 Unsupported Media Type`
+
+  OR
+
+  * **Code** `400 Bad Request`  
+    **Content**
+    ```
+    {
+        "message": "Provided data did not pass validation",
+        "errors": {
+            "sellingTo": "The value \"12.1.2019\" is not a valid DateTime::ATOM format.",
+            "longDescription.1": "The value 345 is not a valid string."
+        }
+    }
+    ```
+
+  OR
+
+  * **Code** `404 Not Found`
+
+*Note: Updating variant/main variant is not supported yet*
+
+**Delete product**
+----
+
+* **URL**
+
+  `/api/v1/products/{uuid}`
+
+* **Method**
+
+  `DELETE`
+
+* **Headers**
+
+  `Content-Type: application/json`
+
+*  **URL Params**
+
+   **Required**
+
+   `uuid=[string]`
+
+* **Success Response**
+
+  * **Code** `204`
+
+* **Error Response**
+
+  * **Code** `400 Bad Request`  
+    **Content**
+    ```
+    {
+        "code": 400,
+        "message": "This UUID is not valid: 123xx"
+    }
+    ```
+
+  OR
+
+  * **Code** `401 UNAUTHORIZED`
+
+  OR
+
+  * **Code** `415 Unsupported Media Type`
+
+  OR
+
+  * **Code** `404 Not Found`
+
+*Note: Deleting variant/main variant is not supported yet*
--- a/docs/cookbook/adding-new-attribute-to-an-entity.md
+++ b/docs/cookbook/adding-new-attribute-to-an-entity.md
@@ -17,7 +17,6 @@ You can find instructions in [Extending Product List](./extending-product-list.m
 
     use Doctrine\ORM\Mapping as ORM;
     use Shopsys\FrameworkBundle\Model\Product\Product as BaseProduct;
-    use Shopsys\FrameworkBundle\Model\Product\ProductCategoryDomainFactoryInterface;
     use Shopsys\FrameworkBundle\Model\Product\ProductData as BaseProductData;
 
     /**
@@ -35,12 +34,11 @@ You can find instructions in [Extending Product List](./extending-product-list.m
 
         /**
          * @param \Shopsys\ShopBundle\Model\Product\ProductData $productData
-         * @param \Shopsys\FrameworkBundle\Model\Product\ProductCategoryDomainFactoryInterface $productCategoryDomainFactory
          * @param \Shopsys\ShopBundle\Model\Product\Product[]|null $variants
          */
-        protected function __construct(BaseProductData $productData, ProductCategoryDomainFactoryInterface $productCategoryDomainFactory, array $variants = null)
+        protected function __construct(BaseProductData $productData, ?array $variants = null)
         {
-            parent::__construct($productData, $productCategoryDomainFactory, $variants);
+            parent::__construct($productData, $variants);
 
             $this->extId = $productData->extId ?? 0;
         }
@@ -192,23 +190,19 @@ The original `ProductFormType` is set as the extended type by implementation of 
     ```php
     namespace Shopsys\ShopBundle\Model\Product;
 
-    use Shopsys\FrameworkBundle\Model\Product\ProductCategoryDomainFactoryInterface;
     use Shopsys\FrameworkBundle\Model\Product\ProductData as BaseProductData;
-    use Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPriceRecalculationScheduler;
 
     // ...
 
     /**
-     * @param \Shopsys\FrameworkBundle\Model\Product\ProductCategoryDomainFactoryInterface $productCategoryDomainFactory
+     * @param \Shopsys\FrameworkBundle\Model\Product\ProductCategoryDomain[] $productCategoryDomains
      * @param \Shopsys\ShopBundle\Model\Product\ProductData $productData
-     * @param \Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPriceRecalculationScheduler $productPriceRecalculationScheduler
      */
     public function edit(
-        ProductCategoryDomainFactoryInterface $productCategoryDomainFactory,
-        BaseProductData $productData,
-        ProductPriceRecalculationScheduler $productPriceRecalculationScheduler
+        array $productCategoryDomains,
+        BaseProductData $productData
     ) {
-        parent::edit($productCategoryDomainFactory, $productData, $productPriceRecalculationScheduler);
+        parent::edit($productCategoryDomains, $productData);
 
         $this->extId = $productData->extId;
     }
--- a/docs/cookbook/backend-api/adding-an-attribute-to-product-import.md
+++ b/docs/cookbook/backend-api/adding-an-attribute-to-product-import.md
@@ -1,0 +1,74 @@
+# Adding an Attribute to Product Import
+
+This short cookbook describes the steps that you need to do when you want to add a new attribute to the product import methods of the [backend API](/docs/backend-api/introduction-to-backend-api.md).
+
+Let's say you have added `extId` attribute to `Product` entity following [Adding New Attribute to an Entity cookbook](/docs/cookbook/adding-new-attribute-to-an-entity.md) and you want to include the attribute in the import as well.
+
+## 1. Extend `ProductDataFactory` for the API
+[`ProductDataFactory`](/packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php) is responsible for converting request body to [`ProductData`](/project-base/src/Shopsys/ShopBundle/Model/Product/ProductData.php).
+You have to add attributes to `ProductData` object during the conversion.
+
+```php
+namespace Shopsys\ShopBundle\Controller\Api\V1\Product;
+
+use Shopsys\BackendApiBundle\Controller\V1\Product\ProductDataFactory as BaseProductDataFactory;
+use Shopsys\FrameworkBundle\Model\Product\ProductData;
+
+class ProductDataFactory extends BaseProductDataFactory
+{
+    protected function setProductDataByApi(ProductData $productData, array $productApiData): void
+    {
+        parent::setProductDataByApi($productData, $productApiData);
+        $this->apiDataSetter->setValueIfExists('extId', $productApiData, $productData);
+    }
+}
+```
+
+*Take a look into the [`ApiDataSetter`](/packages/backend-api/src/Component/DataSetter/ApiDataSetter.php), it contains useful method for conversion like `setDateTimeValueIfExists`, `setMultidomainValueIfExists` and `setMultilanguageValueIfExists`.*
+
+## 2. Extend `ProductApiDataValidator`
+
+[`ProductApiDataValidator`](/packages/backend-api/src/Controller/V1/Product/ProductApiDataValidator.php) is responsible for validating imported data.
+If you need to add a field to the import, you have to always extend this validator.
+
+*We use pure [Symfony Validations](https://symfony.com/doc/3.4/validation.html), se please read about them first.*
+
+```php
+namespace Shopsys\ShopBundle\Controller\Api\V1\Product;
+
+use Shopsys\BackendApiBundle\Controller\V1\Product\ProductApiDataValidator as BaseProductApiDataValidator;
+use Symfony\Component\Validator\Constraints;
+
+class ProductApiDataValidator extends BaseProductApiDataValidator
+{
+    /**
+     * @return array
+     */
+    protected function getConstraintDefinition(): array
+    {
+        $definition = parent::getConstraintDefinition();
+        $definition['fields']['extId'] = new Constraints\Optional([
+            new Constraints\Type([
+                'type' => 'int',
+                'message' => 'The value {{ value }} is not a valid {{ type }}.',
+            ]),
+            new Constraints\NotNull(),
+        ]);
+
+        return $definition;
+    }
+}
+```
+
+The validation definition is used for both creating products and updating products.
+If you need different validation for creating and updating, feel free to implement your own [`ProductApiDataValidatorInterface`](/packages/backend-api/src/Controller/V1/Product/ProductApiDataValidatorInterface.php).
+
+## 3. Add information about the class extensions into the container configuration in [`services.yml`](/project-base/src/Shopsys/ShopBundle/Resources/config/services.yml)
+```yaml
+Shopsys\BackendApiBundle\Controller\V1\Product\ProductDataFactoryInterface: '@Shopsys\ShopBundle\Controller\Api\V1\Product\ProductDataFactory'
+Shopsys\BackendApiBundle\Controller\V1\Product\ProductApiDataValidatorInterface: '@Shopsys\ShopBundle\Controller\Api\V1\Product\ProductApiDataValidator'
+```
+
+## Conclusion
+Now, the `extId` attribute is included in the product import.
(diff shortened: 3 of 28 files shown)


packages/backend-api/src/Model/Product/ProductApiDataValidator.php
```diff
@@ -0,0 +1,119 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\BackendApiBundle\Model\Product;
+
+use DateTime;
+use Shopsys\BackendApiBundle\Component\Validator\AbstractValidator;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Symfony\Component\Validator\Constraints;
+
+class ProductApiDataValidator extends AbstractValidator implements ProductApiDataValidatorInterface
```
simara-svatopluk: The `AbstractValidator` shouldn't be abstract and should be kind of `ValidatorRunner` with public method. And should be used here as private `$validatorRunner`

packages/backend-api/install/tests/ShopBundle/Smoke/BackendApiCreateProductTest.php
```diff
@@ -0,0 +1,105 @@
...
+                2 => '<b>popisek dlouhy',
+            ],
+        ];
+        $response = $this->runOauthRequest('POST', '/api/v1/products', $product);
+
+        $this->assertSame(400, $response->getStatusCode());
+
+        $expectedResponse = [
+            'code' => 400,
+            'message' => 'During validation of data has been found these errors: [name][xx] This field was not expected., [sellingTo] The value "12.1.2019" is not a valid DateTime::ATOM format.',
```
simara-svatopluk: structure of errors should be different, see https://github.com/shopsys/shopsys/blob/f4a86c067fcc77f3cc057a13122e9803b19299fb/docs/backend-api/api-methods.md#add-product "error response"

packages/backend-api/src/Model/Product/ProductDataFactory.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Shopsys\BackendApiBundle\Model\Product;
+
+use Ramsey\Uuid\Uuid;
+use Shopsys\BackendApiBundle\Component\ApiFactory\ApiFactory;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Model\Product\Availability\AvailabilityFacade;
+use Shopsys\FrameworkBundle\Model\Product\ProductData;
+use Shopsys\FrameworkBundle\Model\Product\ProductDataFactoryInterface as BaseProductDataFactoryInterface;
+
+class ProductDataFactory extends ApiFactory implements ProductDataFactoryInterface
```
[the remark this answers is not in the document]
simara-svatopluk: for extensibility reasons - when I need to add eg. prices, I have to extend validator, factory but not controller. Interfaces helps with this

packages/backend-api/install/tests/ShopBundle/Smoke/BackendApiCreateProductTest.php
simara-svatopluk: So we agreed that `ProductApiDataValidatorInterface::validate` will return an array of errors. And controller will always check error count and if it is `> 0` then it will return a different `View`

packages/backend-api/install/tests/ShopBundle/Smoke/BackendApiCreateProductTest.php
```diff
@@ -0,0 +1,187 @@
...
+        $response = $this->runOauthRequest('POST', '/api/v1/products', $product);
+
+        $this->assertSame(201, $response->getStatusCode());
+
+        $location = $response->headers->get('Location');
+        $this->assertProductLocation($location);
+
+        $uuid = $this->extractUuid($location);
+
+        $this->getProduct($uuid);
```
grossmannmartin: Missing assert?

packages/backend-api/src/Component/DataSetter/ApiDataSetter.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->domain = $domain;
+    }
+
+    /**
+     * @param string $arrayKey
+     * @param array $apiData
+     * @param object $entityData
+     * @param string|null $entityDataKey
+     */
+    public function setValueIfExists(string $arrayKey, array $apiData, object $entityData, ?string $entityDataKey = null): void
```
grossmannmartin: `object` is really wide. I can pass any object I want? 
Couldn't it be narrowed?

docs/backend-api/api-methods.md
```diff
@@ -193,19 +197,262 @@ You will need to generate new token after the first one expires.
...
+
+  OR
+
+  * **Code** `400 Bad Request`  
+    **Content**
+    ```
+    {
+        "message": "Provided data did not pass validation",
+        "errors": [
+            "[sellingTo]" => "The value \"12.1.2019\" is not a valid DateTime::ATOM format.",
```
grossmannmartin: Why square brackets? Seems not important to me.

docs/backend-api/api-methods.md
```diff
@@ -193,19 +197,262 @@ You will need to generate new token after the first one expires.
...
+    }
+    ```
+
+  OR
+
+  * **Code** `422 Unprocessable Entity`  
+    **Content**
+    ```
+    {
+        "message": "Product with 74da76db-61ab-43fb-926e-43ba326325fc UUID already exists"
```
grossmannmartin: According to https://restfulapi.net/http-methods/ PUT method should allow update of the resource.
The PUT method should (have to?) be idempotent, but now multiple identical requests have a different outcome.

docs/backend-api/api-methods.md
```diff
@@ -193,19 +197,262 @@ You will need to generate new token after the first one expires.
...
+    }
+    ```
+
+**Update Product**
```
grossmannmartin: I suggest to name it **Partial Product Update** as it's exactly what it does.

docs/backend-api/api-methods.md
```diff
@@ -193,19 +197,262 @@ You will need to generate new token after the first one expires.
...
+  * **Code** `204`
+
+* **Error Response**
+
+  * **Code** `400 Bad Request`  
+    **Content**
+    ```
+    {
+        "code": 400,
+        "message": "This UUID is not valid: f564da76-6d51-4ecd-b004-895c8019a23"
```
grossmannmartin: More obvious invalid ID would be more helpful. 
```suggestion
        "message": "This UUID is not valid: 123"
```

Right now I was thinking if it's malformed UUID or not existing one in a database

docs/cookbook/adding-new-attribute-to-an-entity.md
```diff
@@ -35,12 +34,11 @@ You can find instructions in [Extending Product List](./extending-product-list.m
 
         /**
          * @param \Shopsys\ShopBundle\Model\Product\ProductData $productData
-         * @param \Shopsys\FrameworkBundle\Model\Product\ProductCategoryDomainFactoryInterface $productCategoryDomainFactory
          * @param \Shopsys\ShopBundle\Model\Product\Product[]|null $variants
          */
-        protected function __construct(BaseProductData $productData, ProductCategoryDomainFactoryInterface $productCategoryDomainFactory, array $variants = null)
+        protected function __construct(BaseProductData $productData, array $variants = null)
```
grossmannmartin:
```suggestion
        protected function __construct(BaseProductData $productData, ?array $variants = null)
```

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidator.php
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * @param array $productApiData
+     * @return string[]
+     */
+    public function validateUpdate(array $productApiData): array
+    {
+        $constraintCollection = new Constraints\Collection($this->getConstraintDefinition());
+
+        return $this->validationRunner->runValidation($productApiData, $constraintCollection);
```
grossmannmartin: As this method should behave the same as `validateCreate` (according to documentation), I prefer to simplify it to just 
```php
return $this->validateCreate($productApiData);
```

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidator.php
```diff
@@ -0,0 +1,168 @@
...
+                ),
+            ],
+            'allowExtraFields' => false,
+        ];
+    }
+
+    /**
+     * @return array
+     */
+    protected function createNameConstrains(): array
```
grossmannmartin: typo: (https://www.quora.com/What-is-the-difference-between-constraint-and-constrains)
```diff
-    protected function createNameConstrains(): array
+    protected function createNameConstraints(): array
```

Also, method is named createConstraints, but internally working with fields and return value is passed to fields too. Maybe rename it completely could be nice, but I'm not forcing it.

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidator.php
```diff
@@ -0,0 +1,168 @@
...
+            ]);
+        }
+
+        return $nameFields;
+    }
+
+    /**
+     * @return array
+     */
+    protected function createDescriptionConstrains(): array
```
grossmannmartin: typo: 
```diff
-    protected function createDescriptionConstrains(): array
+    protected function createDescriptionConstraints(): array
```

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidatorInterface.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
```
grossmannmartin: strict types please

packages/backend-api/src/Controller/V1/Product/ProductController.php
```diff
@@ -0,0 +1,285 @@
...
+
+namespace Shopsys\BackendApiBundle\Controller\V1\Product;
+
+use FOS\RestBundle\Controller\AbstractFOSRestController;
+use FOS\RestBundle\Controller\Annotations\Delete;
+use FOS\RestBundle\Controller\Annotations\Get;
+use FOS\RestBundle\Controller\Annotations\Patch;
+use FOS\RestBundle\Controller\Annotations\Post;
+use FOS\RestBundle\Controller\Annotations\Put;
+use FOS\RestBundle\Controller\Annotations\QueryParam;
```
grossmannmartin: I would personally use `FOS\RestBundle\Controller\Annotations` and annotate with `@Annotations\Get("...` But feel free to leave it like it is.

packages/backend-api/src/Controller/V1/Product/ProductController.php
```diff
@@ -0,0 +1,285 @@
...
+
+        $links = $this->linksTransformer->fromPaginationResult($productsResult, $request->getUri());
+
+        $view = View::create($productsArray, Response::HTTP_OK, ['Link' => $links->format()]);
+
+        return $this->handleView($view);
+    }
+
+    /**
+     * @Post("/products")
```
grossmannmartin: Get actions have description. Can you fill it for modifying actions too?

packages/backend-api/src/Controller/V1/Product/ProductController.php
```diff
@@ -0,0 +1,285 @@
...
+            }
+        }
+
+        return $errors;
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     */
+    protected function disallowWorkingWithVariants(Product $product): void
```
grossmannmartin: Method name suggest you somehow change behavior of import. I suggest `assertProductIsNotVariantType` instead (or so).

packages/backend-api/src/Controller/V1/Product/ProductController.php
```diff
@@ -0,0 +1,285 @@
...
+        $code = Response::HTTP_BAD_REQUEST;
+        $message = 'Provided data did not pass validation';
+
+        return View::create(['code' => $code, 'message' => $message, 'errors' => $errors], $code);
+    }
+
+    /**
+     * @param array $uuids
+     */
+    protected function validateUuids(array $uuids): void
```
grossmannmartin: I would like to see this method elsewhere, maybe `Validation` namespace so it could be reused

packages/backend-api/src/Controller/V1/Product/ProductController.php
```diff
@@ -0,0 +1,285 @@
...
+     * @param string|null $uuid
+     * @return array
+     */
+    protected function validateCreateProductInputs(Request $request, ?string $uuid): array
+    {
+        $errors = $this->productApiDataValidator->validateCreate($request->request->all());
+
+        if ($uuid !== null) {
+            if (!Uuid::isValid($uuid)) {
+                $errors['[uuid]'] = 'This UUID is not valid: ' . $uuid;
```
grossmannmartin: Why you use different approach from delete method for checking valid UUID?
Seems little as violation of DRY

packages/backend-api/src/Controller/V1/Product/ProductController.php
```diff
@@ -0,0 +1,285 @@
...
+     */
+    public function createProductAction(Request $request, ?string $uuid = null): Response
+    {
+        $errors = $this->validateCreateProductInputs($request, $uuid);
+
+        if (count($errors) > 0) {
+            return $this->handleView($this->createValidationView($errors));
+        }
+
+        $productData = $this->productDataFactory->createFromApi($request->request->all(), $uuid);
```
grossmannmartin: I like how `$request->request->all()` is extracted into variable in `patchProductAction` method.
It's clear what this value represents. Can you do it here too, please?

packages/backend-api/src/Controller/V1/Product/ProductController.php
```diff
@@ -0,0 +1,285 @@
...
+     * @param string $uuid
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function patchProductAction(string $uuid, Request $request): Response
+    {
+        $this->validateUuids([$uuid]);
+        $productApiData = $request->request->all();
+        $errors = $this->productApiDataValidator->validateUpdate($productApiData);
+        if (count($errors)) {
```
grossmannmartin: please be explicit with conditions
```suggestion
        if (count($errors) > 0) {
```

packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * @param array $productApiData
+     * @param string|null $uuid
+     * @return \Shopsys\FrameworkBundle\Model\Product\ProductData
+     */
+    public function createFromApi(array $productApiData, ?string $uuid = null): ProductData
+    {
+        $productData = $this->productDataFactory->create();
+        $productData->uuid = $uuid ?: Uuid::uuid4()->toString();
+        $productData->availability = $this->availabilityFacade->getDefaultInStockAvailability();
```
grossmannmartin: Does this mean any created product will be by default "In stock" without any way how to change it? 

I don't like it. I would like to have a way to set availability (even if it means I need to find proper value elsewhere)

packages/backend-api/src/Controller/V1/Product/ProductDataFactoryInterface.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
```
grossmannmartin: strict types please

packages/backend-api/src/Resources/config/services.yml
```diff
@@ -1,8 +1,20 @@
...
-    Shopsys\BackendApiBundle\Controller\:
-        resource: '../../Controller/'
-        public: false
+    _defaults:
         autowire: true
+        autoconfigure: true
+        public: false
+
+    Shopsys\BackendApiBundle\:
+        resource: '../../**/*{DataSetter,Factory,Transformer,ValidationRunner,Validator}.php'
```
grossmannmartin: It's mandatory to have ValidationRunner as it's not expected to be more of them and Validator as it still needs to be configured manually (because of the interface)

packages/backend-api/install/tests/ShopBundle/Test/OauthTestCase.php
```diff
@@ -80,18 +82,42 @@ protected function createOauthToken(): string
...
     {
         $client = $this->getClient();
         $headers = [
             'HTTP_AUTHORIZATION' => sprintf('Bearer %s', $this->createOauthToken()),
             'HTTP_ACCEPT' => '*/*',
+            'CONTENT_TYPE' => 'application/json',
         ];
-        $client->request($method, $uri, [], [], $headers, $content);
+        $uri = $this->getDomainBaseUrl() . $path;
+        $client->request($method, $uri, [], [], $headers, $content ? json_encode($content) : null);
```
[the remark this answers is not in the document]
simara-svatopluk: Well technically there is a difference. `null` sends nothing in content `[]` sends `[]` in content. When our API doesn't expect content, then it ignores it, so it doesn't matter.

But I can come with two reasons for `null`
* it's default `Client::request()` value
* We define in the documentation there should be no content

packages/backend-api/install/tests/ShopBundle/Test/OauthTestCase.php
```diff
@@ -80,18 +82,42 @@ protected function createOauthToken(): string
...
+        $domain = $this->getContainer()->get(Domain::class);
+
+        return $domain->getUrl();
+    }
+
+    /**
+     * @param string $url
+     * @return string
+     */
+    protected function extractUuid(string $url): string
```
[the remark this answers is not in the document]
simara-svatopluk: TODO: I'll adjust commit message

packages/backend-api/easy-coding-standard.yml
```diff
@@ -23,3 +23,5 @@ parameters:
     skip:
         Shopsys\CodingStandards\Sniffs\ObjectIsCreatedByFactorySniff:
             - '*/tests/*'
+        ObjectCalisthenics\Sniffs\Files\FunctionLengthSniff:
+            - '*/src/Controller/**/*Validator.php'
```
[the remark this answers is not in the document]
simara-svatopluk: You are right. Just one correction - it suppresses sniff for all *Validator.php in monorepo. When the monorepo is split, travis checks packages one by one and finds errors

packages/backend-api/install/tests/ShopBundle/Smoke/BackendApiCreateProductTest.php
simara-svatopluk: added expecting empty product - product with correct structure, but mostly `null` content

packages/backend-api/src/Component/DataSetter/ApiDataSetter.php
simara-svatopluk: Well idea of this `ApiDataSetter` is to be able to set any data object in the future. So it is true that right now it sets only `ProductData`, but in the future it could be `OrderData`, ...

So I can't come with a better typehint

docs/backend-api/api-methods.md
simara-svatopluk: That's because the Symfony validator produces it in this form and we didn't have a reason for removing them.

Nested errors have keys for example `[shortDescription][1]`, so "just removing" would be eventually complicated.

Do you think I shall change the format in some way? If so, please suggest one and keep in mind nested errors

docs/backend-api/api-methods.md
simara-svatopluk: You are right!!! And what more, you are right according to the RFC!

We - me, Tom, David - agreed on this solution because we want to provide only one way of updating - PATCH, but still we want to provide a way to create product with specific UUID.
And that's why this situation is documented.

It would be best if we discuss this topic personally (by phone) and come with solution.

docs/cookbook/adding-new-attribute-to-an-entity.md
simara-svatopluk: applied manually

docs/cookbook/backend-api/adding-an-attribute-to-product-export.md
```diff
@@ -5,12 +5,12 @@ This short cookbook describes the steps that you need to do when you want to add
...
 ## 1. Implement your own product transformer by extending `ApiProductTransformer`
-[`ApiProductTransformer`](/packages/backend-api/src/Controller/V1/ApiProductTransformer.php) is responsible for providing array of product data to backend API controllers.
+[`ApiProductTransformer`](/packages/backend-api/src/Controller/V1/Product/ApiProductTransformer.php) is responsible for providing array of product data to backend API controllers.
 You need to add your attribute to the array that is returned in `transform()` method.
 ```php
-namespace Shopsys\ShopBundle\Controller\Api\V1;
+namespace Shopsys\ShopBundle\Controller\Api\V1\Product;
 
-use Shopsys\BackendApiBundle\Controller\V1\ApiProductTransformer as BaseApiProductTransformer;
+use Shopsys\BackendApiBundle\Controller\V1\Product\ApiProductTransformer as BaseApiProductTransformer;
```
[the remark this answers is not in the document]
simara-svatopluk: Nice point. What is the reason for creating a new version of API? It is a major change (not addition) in URL or data structure.

Imagine we'll do major changes in data structure eg. removing fields, moving them into subsections. Could we in such case use anything from V1? It's unclear, depends on the change.

So IF there is not major change in data structure in V2, we could use/extract V1 validator; but when there is a major change in data structure, we'll have to write it from scratch (or copy and rewrite)

docs/cookbook/backend-api/adding-an-attribute-to-product-import.md
```diff
@@ -0,0 +1,73 @@
+# Adding an Attribute to Product Import
+
+This short cookbook describes the steps that you need to do when you want to add a new attribute to the product import methods of the [backend API](/docs/backend-api/introduction-to-backend-api.md).
+
+Let's say you have added `extId` attribute to `Product` entity following [the cookbook](/docs/cookbook/adding-new-attribute-to-an-entity.md) and you want to include the attribute in the import as well.
+
+## 1. Extend `ProductDataFactory` for the API
+[`ProductDataFactory`](/packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php) is responsible for converting API data to [`ProductData`](/project-base/src/Shopsys/ShopBundle/Model/Product/ProductData.php).
```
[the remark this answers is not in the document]
simara-svatopluk: changed to `...is responsible for converting request body to ProductData...`

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -258,5 +258,11 @@ There you can find links to upgrade notes for other versions too.
 ## [shopsys/coding-standards]
 - run `php phing standards-fix` to fix code style as we now disallow Yoda style for comparison in the Shopsys Framework coding standards ([#1209](https://github.com/shopsys/shopsys/pull/1209))
 
+## [shopsys/backend-api]
```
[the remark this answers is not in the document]
simara-svatopluk: That change isn't trivial - it is involved in multiple files like documentation, maybe configs, maybe tests... It would take I guess 2 hours of devel time to "do it"... Please come with this topic on standup

packages/backend-api/install/app/config/packages/test/fos_rest.yml
```diff
@@ -0,0 +1,3 @@
+fos_rest:
+    exception:
+        debug: true
```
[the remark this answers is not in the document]
simara-svatopluk: commit message explains it as
> fos rest bundle is set to be in debug mode in test environment so it returns exceptions correctly during smoke tests

Tom is author of this fix, it took him like 2 hours to figure it out :-)

packages/backend-api/src/Component/Validation/ValidationRunner.php
```diff
@@ -0,0 +1,41 @@
...
+
+        /** @var \Symfony\Component\Validator\ConstraintViolationList $violations */
+        $violations = $validator->validate(
+            $apiData,
+            $constraintCollection
+        );
+
+        $errors = [];
+
+        if (count($violations) > 0) {
```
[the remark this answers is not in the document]
simara-svatopluk: I prefer even `if ($violations->count()) {`

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidator.php
simara-svatopluk: renamed to
* `createNameConstraintDefinition`
* `createDescriptionConstraintDefinition`

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidatorInterface.php
simara-svatopluk: Sir! Yes, sir!

packages/backend-api/src/Controller/V1/Product/ProductController.php
simara-svatopluk: ok

packages/backend-api/install/tests/ShopBundle/Test/OauthTestCase.php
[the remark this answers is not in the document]
grossmannmartin: I'm not sure we understand each other. 
I'm referring to`$content` in a condition – whether it's intentional to convert empty array in `$content` to null, because passing empty array to method `runOauthRequest` does that.

packages/backend-api/src/Controller/V1/Product/ProductController.php
simara-svatopluk: This validation method bothered me for a while, and even Tom wanted to keep it in this fashion I decided to rewrite it to validate only if the UUID is setted... so please check `validateCreatingProductWithDefinedUuid` and beginning of the `createProductAction` method changed to
```php
        if ($uuid) {
            $this->validateCreatingProductWithDefinedUuid($uuid);
        }

        $productApiData = $request->request->all();
        $errors = $this->productApiDataValidator->validateCreate($productApiData);
```

I hate passing optional parameters when the optionality can be solved in the caller, so this is the result I like

packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php
simara-svatopluk: @PetrHeinz told me this is "the default correct behavior". I cannot argue against anything.

I only know that without availability the system breaks, and importing availability isn't in scope of this US

PetrHeinz: This is actually the only possible behavior I can think of, without widening the scope of the US.

packages/backend-api/easy-coding-standard.yml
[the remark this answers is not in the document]
grossmannmartin: I created #1249  to address this issue in the future.

packages/backend-api/src/Controller/V1/Product/ProductController.php
simara-svatopluk: Really? I personally find the sorter version better, sort of idiom

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidator.php
simara-svatopluk: renamed to
* `createNameConstraintDefinition`
* `createDescriptionConstraintDefinition`

packages/backend-api/src/Component/DataSetter/ApiDataSetter.php
grossmannmartin: Well the interface came in my mind, but I have some feeling that you will not agree 😄

docs/backend-api/api-methods.md
grossmannmartin: I liked the format initially introduced in documentation:
```json
{
  "errors": {
    "sellingTo":  "The value \"12.1.2019\" is not a valid DateTime::ATOM format.",
    "shortDescription.1": "Bad word found."
  }
}
```

or nested structure 
```json
{
  "errors": {
    "sellingTo":  "The value \"12.1.2019\" is not a valid DateTime::ATOM format.",
    "shortDescription": {
      "1": "Bad word found."
    }
  }
}
```
and the main argument to handle this: **Returned response is not a valid JSON right now.**

docs/upgrade/UPGRADE-unreleased.md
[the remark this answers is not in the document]
grossmannmartin: We decided to solve this in separate PR. Resolving conversation here.

packages/backend-api/install/app/config/packages/test/fos_rest.yml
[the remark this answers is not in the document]
grossmannmartin: Ou, I missed that. Thanks :)

packages/backend-api/src/Component/Validation/ValidationRunner.php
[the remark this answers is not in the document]
grossmannmartin: Well, I like being explicit :) 
But as we don't have any agreement about this topic in our team it's ok.

packages/backend-api/install/tests/ShopBundle/Test/OauthTestCase.php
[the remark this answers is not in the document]
simara-svatopluk: you are right, it is incorrect. Fixed

packages/backend-api/src/Component/DataSetter/ApiDataSetter.php
simara-svatopluk: correct

packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php
grossmannmartin: I know it's more MVP of API but still, without availability, it's not usable to me.
And if I saw properly, handling this is not planed in any future US in the MVP Epic nor in the roadmap plans.

Adding an optional field for availability really widen the scope so much?

docs/backend-api/api-methods.md
simara-svatopluk: I can only agree. Changed to dot notation

packages/backend-api/src/Controller/V1/Product/ProductController.php
grossmannmartin: Well, I like being explicit :) 
But as we don't have any agreement about this topic in our team it's ok.

packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php
simara-svatopluk: Well... I would leave this discussion for BV, what do you think?

docs/cookbook/backend-api/adding-an-attribute-to-product-export.md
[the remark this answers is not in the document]
grossmannmartin: Ok to me :)

docs/backend-api/api-methods.md
grossmannmartin: I still struggle with the format of the response.
```
{
    "message": "Provided data did not pass validation",
    "errors": [
        "sellingTo" => "The value \"12.1.2019\" is not a valid DateTime::ATOM format.",
        "longDescription.1" => "The value 345 is not a valid string.",
    ]
}
```

is not valid JSON and I believe it should be

packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php
simara-svatopluk: or we can involve @LukasHeinz without waiting for BV!

Martin:
> I know it's more MVP of API but still, without availability, it's not usable to me.
And if I saw properly, handling this is not planed in any future US in the MVP Epic nor in the roadmap plans.

Petr:
> This is actually the only possible behavior I can think of, without widening the scope of the US.

Me...
* Handling availability is not part of this US. We cannot always solve everything
* If implementators needs to work with availability immediately, they can implement handling it by themselves
* Unfortunatelly it is more complicated how should be handled... string constrants? Is it enough? What about stocks? What if the implementators extend availabilities by new ones?
* So I vote for not solving availabilities in this US, and create a new US that will solve only them and do the US right after this one

docs/backend-api/api-methods.md
simara-svatopluk: how did you get this response?

I just tried to POST with errors, and I got
```json
{
    "code": 400,
    "message": "Provided data did not pass validation",
    "errors": {
        "name.en": "The value \"22\" ......\" cannot be longer then 255 characters",
        "extId": "This field was not expected."
    }
}
```

And this is a valid JSON

packages/backend-api/src/Component/DataSetter/ApiDataSetter.php
grossmannmartin: okay then :)

docs/backend-api/api-methods.md
grossmannmartin: Ou I didn't test it live... The invalid format is in documentation `api-methods.md` and I assumed it's the same as the return value from the code :)

docs/backend-api/api-methods.md
simara-svatopluk: so we agreed that UUID will be part of request body and we'll remove PUT method

packages/backend-api/install/tests/ShopBundle/Smoke/BackendApiCreateProductTest.php
```diff
@@ -0,0 +1,215 @@
...
+        $this->assertSame(201, $response->getStatusCode());
+
+        $location = $response->headers->get('Location');
+        $this->assertProductLocation($location);
+
+        $extractedUuid = $this->extractUuid($location);
+        $this->assertSame($uuid, $extractedUuid);
+
+        $foundProduct = $this->getProduct($extractedUuid);
+        unset($product['uuid']);
```
grossmannmartin: You unset uuid here and also in `getProduct` method. But you can compare products incl. UUIDs, right?

packages/backend-api/src/Controller/V1/Product/ProductApiDataValidator.php
```diff
@@ -0,0 +1,172 @@
...
+        return $nameFields;
+    }
+
+    /**
+     * @return array
+     */
+    protected function createDescriptionConstraintDefinition(): array
+    {
+        $descriptionFields = [];
+        foreach ($this->domain->getAllIds() as $locale) {
```
grossmannmartin:
```suggestion
        foreach ($this->domain->getAllIds() as $domainId) {
```
and accordingly in usage below

packages/backend-api/install/tests/ShopBundle/Smoke/BackendApiCreateProductTest.php
simara-svatopluk: Well, I removed `unset` and had to fix tests with POST without UUID. The outcome is -5 and +5 lines :-D I'll keep it after the change

grossmannmartin:
Hi @simara-svatopluk, I noticed the upgrade notes are still present in this PR.
Can you rebase and remove them as they're not needed anymore?

grossmannmartin:
Hi, I did user testing.

1. It's mandatory to send header `Content-Type: application/json`, and withou it it just always create empty product no matter what you send in a request body. 
It was completely confusing to me and took long time to figure out what is going on.

2. sellingFrom allows to send datetime, but returned time is always midnight.
```
{
  "name": {
      "en": "22\" Sencor SLE 22F46DM4 HELLO KITTY",
      "cs": "22\" Sencor SLE 22F46DM4 HELLO KITTY"
  },
   "sellingFrom": "2010-01-16T07:05:00+00:00"
} 
```
get response 
```
{
    "uuid": "e99a4960-e96d-47eb-bb53-56c6b8a56c46",
    "name": {
        "en": "22\" Sencor SLE 22F46DM4 HELLO KITTY",
        "cs": "22\" Sencor SLE 22F46DM4 HELLO KITTY"
    },
    "hidden": false,
    "sellingDenied": false,
    "sellingFrom": "2010-01-16T00:00:00+00:00",
    "sellingTo": null,
```


3. PATCH method returns 204 status code, but documentation states it should be 201.

4. in PATCH uuid field is completely ignored even for validation
```
{
	"uuid": "sdfsdf91805aa2bc8",
  "hidden": false
} 
```
But we agreed on this and need for valid uuid which will be ignored in the end would be even more weird. Just wanted to point this out :) no need to do anything

And I'm going to test cookbooks (extensibility)

grossmannmartin:
Testing the cookbook:

PATCH method with null value in `extId` field results in 500 exception with following content

```json
{
    "code": 500,
    "message": "An exception occurred while executing 'UPDATE products SET ext_id = ?, recalculate_availability = ?, recalculate_visibility = ? WHERE id = ?' with params [null, 1, 1, 192]:\n\nSQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column \"ext_id\" violates not-null constraint\nDETAIL:  Failing row contains (192, 4, 2, 2, null, 2, null, null, null, , null, , null, null, null, f, f, f, f, f, null, null, t, f, f, t, none, 0, 4c2053a8-acb0-4f21-93dd-357ac2f3d1c6, null)."
}
```

_same occurs when trying to set `"hidden": null`_

simara-svatopluk:
1. fixed by checking Content-Type header, added documentation
2. that's true. We just use ATOM format, and it have to contain time; But the field itself is a date, I don't know how to fix this (is it even error?)
3. fixed docs
4. ok, I do nothing
5. fixed by not allowing `null` for `hidden`, `sellingDenied` and `extId` in cookbook

docs/backend-api/api-methods.md
```diff
@@ -176,36 +179,255 @@ You will need to generate new token after the first one expires.
...
+
+* **URL**
+
+  `/api/v1/products/{uuid}`
+
+* **Method**
+
+  `DELETE`
+
+*  **URL Params**
```
grossmannmartin: DELETE method also checks for content-type and can return error-response

simara-svatopluk: added documentation (I don't want to hack FOSRest bundle to skip this method)

packages/backend-api/src/Controller/V1/Product/ProductDataFactory.php
simara-svatopluk: This PR is a first step of the import API, so anyone can use it. But there is more work to do like categories, prices and of course availability. So availability won't be addressed now

grossmannmartin: Availability is literally only required field when you create a product. I appreciate if we push availabilities on the top of things to do in API :)

PR #1207, opened 2019-07-15
[framework] DailyFeedCronModule: use getter for queue to eliminate side effects in constructor

hason:
| Q             | A
| ------------- | ---
|Description, reason for the PR| I want to run `bin/console` to list all commands even if the database schema is not yet created.
|New feature| No
|Fixes issues| -

### DIFF
--- a/docs/upgrade/UPGRADE-unreleased.md
+++ b/docs/upgrade/UPGRADE-unreleased.md
@@ -249,6 +249,8 @@ There you can find links to upgrade notes for other versions too.
     - use helper classes `AdminCheckbox` and `AdminRadiobutton` instead of directly manipulating `input[type="checkbox"]` and `input[type="radio"]` elements in your administration acceptance tests
         - when you need to work with a particular input, create an instance of the appropriate class via static method `createByCss()` and use its methods to manipulate the input or assert its values
         - run `php phing tests-acceptance` to see that your acceptance test pass
+- use getter method instead of property inside `DailyFeedCronModule` ([#1207](https://github.com/shopsys/shopsys/pull/1207))
+    - if you extend `DailyFeedCronModule` in your project use the protected method `getFeedExportCreationDataQueue()` instead of `$this->feedExportCreationDataQueue` (which now might be `null`)
 
 ### Configuration
 - simplify local configuration ([#1004](https://github.com/shopsys/shopsys/pull/1004))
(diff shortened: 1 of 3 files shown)


PetrHeinz:
Hello @hason and thanks for your first time contribution to Shopsys Framework! :+1: 

I've looked through the suggestion and I do agree with removing the side effect from the constructor of `DailyFeedCronModule` - that `FeedFacade::getFeedNames('daily')` and `Domain::getAll()` calls shouldn't be there.

However, the PR actually changed the type of `DailyFeedCronModule::$feedExportCreationDataQueue` to `FeedExportCreationDataQueue|null`, which is a violation of our [BC promise](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md) and cannot be done in a `PATCH` version (as you're targeting your PR to `v7.3.2-dev`).
Imagine someone would extend the `DailyFeedCronModule` in their project and overriden some method using the now nullable property in their code. After a upgrade to the `v7.3.2` their code could end up with a `Call to a member function isEmpty() on null` error...

---

I have thought of a solution that would be both backward-compatible and achieve your end-goal: just set [the cron module service as lazy](https://symfony.com/doc/3.4/service_container/lazy_services.html) in `packages/framework/src/Resources/config/services/cron.yml`:

```diff
 Shopsys\FrameworkBundle\Model\Feed\DailyFeedCronModule:
     tags:
         - { name: shopsys.cron, hours: '*/6', minutes: '0' }
+    lazy: true

 Shopsys\FrameworkBundle\Model\Feed\HourlyFeedCronModule:
```

What would you think of such a solution?

---

If you agree, you might implement that change instead and it will get merged to the `7.3` branch pretty shortly. I would do that myself, but I didn't want to take the authorship of your contribution :slightly_smiling_face: 

If you want, you can also create a PR for the proper solution targeted into `master` so the side-effect gets actually removed in `v8.0.0-dev`. I can help with writing a proper upgrade note so our users know they should check their extended classes and overwritten methods.
In that case, please run `php phing standards-fix` so that the annotation gets fix and update the annotation of the `DailyFeedCronModule::$feedExportCreationDataQueue` property to a `@var \Shopsys\FrameworkBundle\Model\Feed\FeedExportCreationDataQueue|null` so it's not misleading.

PetrHeinz:
Seeing that this is a BC break, I've rebased and retargeted the PR into `master` (`v8.0.0-dev`).

PR #1211, opened 2019-07-16
[framework] customers grid in admin has now better "Items per page" text

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Customers grid had "Items per page" texting in grid. That was changed to "Customers per page".
|New feature| No 
|Fixes issues| ... 


LukasHeinz:
Currently, it is consistent through the whole administration - everywhere it is "items per page". It eases the implementation of new lists because you do not need to think about it. 

If we introduce this thing we need to make administration consistent again - to add this to every grid in admin (products, orders, brands, ...). 

It is not a bug. This might be a nice feature, but it must be everywhere. 
P.S. I do not recommend to work on it during the admin facelift phase.

PR #1214, opened 2019-07-17
[framework] AbstractMigration: removed deprecation annotation

### LINKED ISSUE #1213: There is deprecation in prepared new major version (8.0.0)
### What is happening

During resolving deprecation in preparation of new major #1193 there was one deprecation left.

There is a lot of opinions about whether this is good.

UPDATE: there are more deprecations than just one.

### Expected result

It should be resolved or there should be a strong explanation of why to leave it here.

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The method is not in fact deprecated, it's rather forbidden as explained in https://github.com/shopsys/shopsys/issues/1213#issuecomment-512242620
|New feature| No 
|Fixes issues| ~resolves~ *helps with* #1213 

### DIFF
--- a/packages/migrations/src/Component/Doctrine/Migrations/AbstractMigration.php
+++ b/packages/migrations/src/Component/Doctrine/Migrations/AbstractMigration.php
@@ -9,7 +9,6 @@ abstract class AbstractMigration extends DoctrineAbstractMigration
 {
     /**
      * {@inheritDoc}
-     * @deprecated use "sql" method instead
      */
     protected function addSql($sql, array $params = [], array $types = [])
     {


PetrHeinz:
Hi @TomasLudvik,  the `CartController::addProductFormAction()` seems deprecated but is being used on product detail... Methods in `StrictWebDriver` look like it's very similar to this case, to be honest I don't want to change it now there seems to be more work (renaming `DeprecatedMethodException`, moving the `getDeprecatedMethodExceptionMessage` method body into the exception class and writing upgrade instruction as it a part of `project-base` repo).

I was working on the assumption that there is a single deprecation left (see #1213).

PetrHeinz:
@vitek-rostislav I would be OK with such resolution.

TomasLudvik:
@vitek-rostislav yes I am ok

PR #1215, opened 2019-07-18
[framework] admin: display real name of logged user instead of icon

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Displaying real name of logged administrator in the account menu (upper right corner) makes much more sense than just displaying an icon.
|New feature| No 
|Fixes issues| ... 

**Before:**
image

**After:**
image

### DIFF
--- a/packages/framework/src/Resources/styles/admin/components/box/dropdown.less
+++ b/packages/framework/src/Resources/styles/admin/components/box/dropdown.less
@@ -57,13 +57,15 @@
             }
         }
 
-        &--account {
-            color: @color-base;
-
+        &--with-icon {
             .svg {
                 margin-right: 5px;
                 vertical-align: bottom;
             }
+        }
+
+        &--grey {
+            color: @color-grey-middle;
         }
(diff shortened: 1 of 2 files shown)


henzigo:
What about to display name with icon ? O.o

PetrHeinz:
@henzigo Well, I thought about the exact same thing but there should be no icon according to the design of admin face-lift:

image

Maybe we can reconsider? @LukasHeinz

Also, the color should be more subtle - I guess that's something @tomshopsys will do after this PR gets merged.

PetrHeinz:
I've managed to make it look like this:
image

Can you please take a look on it @tomshopsys? Thx!

tomshopsys:
I just split modification for future use. Now we can use dropdown with icon and with grey text separately.

PetrHeinz:
@PatrikBajer I've fixed that in #1227 as it is a separate issue, thx for noticing. Is this PR OK otherwise?

PR #1229, opened 2019-07-19
[framework] admin: currency is displayed in the order table header instead of behind every input

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Having all inputs with currencies behind them was a 
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Yes (simplification of Twig macros) 
|Fixes issues| ... 

**Before:**
image

**After:**
image


PetrHeinz:
@PatrikBajer Well, the table header gets wrapped automatically, I don't see any viable systematic solution. And I think @tomshopsys plans to make the columns narrower.

PR #1232, opened 2019-07-19
[framework] LegalConditionsFacade::findArticle() returns null when the Article doesn't exist

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Repository and facade methods starting with `find` should return the entity or `null`. Methods starting with `get` should return the entity or throw an exception (see [our Coding Standards that are not checked automatically](https://github.com/shopsys/shopsys/blob/master/docs/contributing/coding-standards.md#standards-that-are-not-checked-automatically)). Here `get` was called from within a `find`, failing with error when the expected `Article` was not found.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/LegalConditions/LegalConditionsFacade.php
+++ b/packages/framework/src/Model/LegalConditions/LegalConditionsFacade.php
@@ -113,7 +113,7 @@ class LegalConditionsFacade
         $articleId = $this->setting->getForDomain($settingKey, $domainId);
 
         if ($articleId !== null) {
-            return $this->articleFacade->getById($articleId);
+            return $this->articleFacade->findById($articleId);
         }
 
         return null;


packages/framework/src/Model/LegalConditions/LegalConditionsFacade.php
```diff
@@ -113,7 +113,7 @@ protected function findArticle($settingKey, $domainId)
         $articleId = $this->setting->getForDomain($settingKey, $domainId);
 
         if ($articleId !== null) {
-            return $this->articleFacade->getById($articleId);
+            return $this->articleFacade->findById($articleId);
```
grossmannmartin: How about to skip if and returning null and return a result of `ArticleFacade::findById` directly?

grossmannmartin:
And please update PR description ;)

packages/framework/src/Model/LegalConditions/LegalConditionsFacade.php
PetrHeinz: But `ArticleFacade::findById()` does not accept `null` as a parameter, only `int`.

There are three cases:
1. The article is not set.
2. The article is set but does not exists. (this case is fixed here)
3. The article is set and exists.

grossmannmartin: Sorry I forgot that findById does not accept null as an argument. It's okay.

PR #1233, opened 2019-07-19
[framework] GridView: remove .form-line__side before rendering

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Forms that appear in inline edit are not displayed well. Problems arise from the fact that `form_row` is used to render the form and the `theme.html.twig` assumes it will be rendered in a standard full-page form. Fixing it properly would require an overhaul of the form rendering and of the current theme. I thought of adding a one more boolean parameter (`display_side`) that could toggle rendering of the `.form-line__side` but that would be probably also pretty brittle. Furthermore, the `.form-line > .form-line__side > .form-line__item` HTML structure that appears in our forms is a violation of BEM.<br>In the end, I propose a kind of a dirty solution - just filtering all occurrences of `form-line__side` from the rendered output in `GridView`. There will probably never be a case where this element would make sense in a grid and the string is unambiguous so I'm not afraid it will ever remove something it shouldn't.<br>I cannot honestly say I'm proud of this solution, but in the end, I think it will get the job done.
|New feature| No 
|Fixes issues| ... 

**Before:**
image

**After:**
image

### DIFF
--- a/packages/framework/src/Component/Grid/GridView.php
+++ b/packages/framework/src/Component/Grid/GridView.php
@@ -6,6 +6,7 @@ use Symfony\Component\Form\FormView;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
+use Twig\TemplateWrapper;
 use Twig_Environment;
 
 class GridView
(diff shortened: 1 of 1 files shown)


PetrHeinz:
@grossmannmartin OK, I agree.

I'm closing this PR, @tomshopsys will resolve this issue via CSS in his `tg-admin-facelift-issues` branch.

PR #1240, opened 2019-07-22
[shopsys] added Phing target that checks whether there is a project-base/vendor dir

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When someone installs Composer dependencies in project-base directory in monorepo, the autoloader will start failing on duplicate class definitions. I added a check that will fail in such a case and tell the dev what to do. Only added to monorepo build.xml as it is only relevant there.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -206,7 +206,7 @@
         <if>
             <available file="${path.root}/vendor" type="dir" property="dir.Exists"/>
             <then>
-                <echo level="error" message="Composer dependencies are installed in ${path.root} directory which would break the monorepo autoloader. Remove ${path.root}/vendor before re-running the command,"/>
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -200,6 +202,16 @@
         </exec>
     </target>
 
+    <target name="project-base-vendor-check" description="Check there exists no vendor directory in project-base/ as that would break the autoloader" hidden="true">
+        <if>
+            <available file="${path.root}/vendor" type="dir" property="dir.Exists"/>
+            <then>
+                <echo level="error" message="Composer dependencies are installed in ${path.root} directory which would break the monorepo autoloader. Remove ${path.root}/vendor before re-running the command,"/>
```
grossmannmartin:
```suggestion
                <echo level="error" message="Composer dependencies are installed in ${path.root} directory which would break the monorepo autoloader. Remove ${path.root}/vendor before re-running the command."/>
```

PR #906, opened 2019-03-25
[shopsys] admin design facelift

tomshopsys:
| Q             | A
| ------------- | ---
|Description, reason for the PR| administration wasn't well arranged and modern. This is just the first phase that doesn't include big changes in structure. There is preparation for the statistics on dashboard 
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| ... 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1241, opened 2019-07-22
[framework] Admin facelift issues and tweaks

tomshopsys:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Resolve issues after administration design facelift in #906 
|New feature| Yes 
|Fixes issues| ... 


PetrHeinz:
@tomshopsys The `selectize.js` library removes `option`s from the original `select` (other then the currently selected one) which breaks the acceptance tests (they work with the original hidden `select` element instead of the graphical one).

See the original input (I've removed the `display:none;` for demonstration):
image

**This must be addressed before this PR can be merged.**

Also, I've noticed that the `selectize.js` copies all classes of the original input to the newly created elements. This might be wanted from the point of view of style, but it may create problems when it copies the `js-*` classes - our scripts sometimes assume that the `.js-*` elements will be of certain type or something like that...

This can be amended by a simple change to filter only non-js-specific classes in the `packages/framework/src/Resources/scripts/admin/plugins/selectize.js` (line 1175):

```diff
	var classes_plugins;

	inputMode         = self.settings.mode;
-	classes           = $input.attr('class') || '';
+	classes           = $input.attr('class').replace(/\bjs-[a-zA-Z0-9_-]\b/g, '') || '';

	$wrapper          = $('<div>').addClass(settings.wrapperClass).addClass(classes).addClass(inputMode);
	$control          = $('<div>').addClass(settings.inputClass).addClass('items').appendTo($wrapper);
```

@LukasHeinz: Speaking of modification to `selectize.js`, can we make direct modifications to the script which is under *Apache License, Version 2.0*? If so, are there any restrictions to the changes? Do we need to make clear which parts we modified?

---

While testing the advanced filter I found our that some filter types (eg. Flags) don't really fit the bounding box:
image

LukasHeinz:
For `Apache License, Version 2.0` applies restriction that when we change the code we must provide information that we changed it. I would suggest adding NOTICE.md file to the folder with the library (or where its copyright or license file is)and write down that we did these changes + add our copyright...

packages/framework/src/Resources/views/Admin/Content/PersonalData/index.html.twig
```diff
@@ -5,71 +5,80 @@
...
             <div class="form-group margin-bottom-20">
                 <h2 class="form-group__title">
                     {{ 'Link to the personal information overview'|trans }}
                 </h2>
-                <div class="form-line">
-                    <div class="form-line__line">
-                        <div class="form-line__label">
-                            {{ getDomainName(domainId) }}
-                            {% if isMultidomain() %}
+                <div>
```
PetrHeinz: Why is this `div` element important? Can't we just get rid of it?

packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
```diff
@@ -1,36 +1,33 @@
...
-        </tbody>
-        <tfoot>
-            <tr>
-                <td><strong>{{ 'Total including VAT'|trans }}</strong></td>
-                <td></td>
-                <td class="text-right"><strong>{{ order.totalPriceWithVat|priceWithCurrency(order.currency) }}</strong></td>
-            </tr>
-        </tfoot>
-    </table>
+<div class="window-content__heading text-left pl30">
```
PetrHeinz: This element is not rendered in `.window-content`, so according to BEM it must not have `window-content__heading` class.

Also, the heading doesn't look quite well imo, maybe there's something broken...
image

When I've tried to add a wrapping `.window-content` the heading was rendered over the table header because of the `position: absolute;`.

packages/framework/src/Resources/views/Admin/Content/Image/overview.html.twig
```diff
@@ -19,27 +19,27 @@
...
+        <table class="table-grid">
             <thead>
-                <tr>
-                    <th class="table-col-20">{{ 'Size name'|trans }}</th>
-                    <th class="table-col-10">{{ 'Width [px]'|trans }}</th>
-                    <th class="table-col-10">{{ 'Height [px]'|trans }}</th>
-                    <th class="table-col-50">{{ 'Occurrence'|trans }}</th>
-                    <th class="table-col-10 text-right">{{ 'Cropping'|trans }}</th>
+                <tr class="table-grid__row">
+                    <th class="table-grid__cell table-col-20">{{ 'Size name'|trans }}</th>
```
PetrHeinz: I've always thought this structure (`.table-grid > .table-grid__row > .table-grid__cell`) is also a violation of BEM. Is there some reason for it?

packages/framework/src/Resources/views/Admin/Content/Currency/currencySettings.html.twig
```diff
@@ -1,12 +1,14 @@
...
-        </div>
-        {{ form_row(form.defaultCurrency, { label: 'Default administration currency'|trans, icon: { title: 'When changing the default currency there is no recalculation of prices'|trans } }) }}
-        {% for item in form.domainDefaultCurrencies %}
-            {{ form_row(item, { label: 'Basic frontend currency %domainName%'|trans({'%domainName%': domainNames[item.vars.name]})}) }}
-        {% endfor %}
+        <div class="form-group">
+            <div class="form-group__title">
+                {{ 'Settings'|trans }}
+            </div>
+            <div>
```
PetrHeinz: Why are you adding the `div`?

packages/framework/src/Resources/views/Admin/Content/Brand/list.html.twig
```diff
@@ -3,34 +3,38 @@
 {% block title %}- {{ 'Brand overview'|trans }}{% endblock %}
 {% block h1 %}{{ 'Brand overview'|trans }}{% endblock %}
 
+{% block btn %}
```
PetrHeinz: This will look pretty weird in online stores with many domains - the button will be too far from the grid imo.

image

packages/framework/src/Resources/views/Admin/Content/Country/list.html.twig
```diff
@@ -4,12 +4,13 @@
 
 {% block h1 %}{{ 'Countries overview'|trans }}{% endblock %}
 
+{% block btn %}
```
PetrHeinz: I guess you haven't done this on other pages (`Settings >  Lists and nomenclatures >  Flags`, `Settings >  Lists and nomenclatures >  Parameters`, `Settings >  Lists and nomenclatures >  Status of orders`, `Pricing >  Promo codes`) because the use inline editing, right?

image

packages/framework/src/Resources/styles/admin/core/form/custom-inputs.less
```diff
@@ -0,0 +1,114 @@
+@custom-inputs-margin-left: 30px;
```
PetrHeinz: This file is indented by `TAB`s instead of spaces and uses `\r\n` instead of `\n` for line ending. Did you copy it from a different project, or do you have a misconfigured IDE?

packages/framework/src/Twig/DomainExtension.php
```diff
@@ -108,7 +108,7 @@ class="in-image__in"
                 </span>';
         } else {
             return '
-                <span class="in-image in-image--' . $size . '">
+                <span class="in-image">
```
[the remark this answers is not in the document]
tomshopsys: In first step I want to unify all icon to same size. But finallly I have to used smaller one in order list. 

There are new domain icons with SSFW logo. It is possible situation that we dont have icon?

packages/framework/src/Resources/views/Admin/Content/PersonalData/index.html.twig
tomshopsys: It makes inside element form-line first and last child - and it means that gets border radius

packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
tomshopsys: I tried to unify display in hover and display in popup. So i used `window-content__heading` for both.

packages/framework/src/Resources/views/Admin/Content/Image/overview.html.twig
tomshopsys: Yes thats true is little bit of violation. But we are using this bcs the source code is smaller and more readable.

Specially in table where can be hundreds of elements, can be source code much bigger - 80% source code for elements, 20% for content. So in cases where it makese sence we can use it like this.

packages/framework/src/Resources/views/Admin/Content/PersonalData/index.html.twig
PetrHeinz: That makes sense, thanks!

packages/framework/src/Resources/views/Admin/Content/Currency/currencySettings.html.twig
tomshopsys: Same as higher. This parent makes child - first child and last child so it gets border radius

packages/framework/src/Twig/DomainExtension.php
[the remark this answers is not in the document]
PetrHeinz: Yes - for example when someone doesn't use demo data or adds a new domain for which we don't have an icon by default.

packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
PetrHeinz: I'm asking because of the big red warning about not doing this in our own CSS documentation:
image

Could you please explain why is it OK in this case?

packages/framework/src/Resources/views/Admin/Content/Country/list.html.twig
tomshopsys: If I would be honest - when i found add button outside of block btn a tried to move it. But on some pages I didn´t find add button. Is there any posibilty to review & unify it?

packages/framework/src/Resources/styles/admin/core/form/custom-inputs.less
tomshopsys: Yes, it is from another project so I fixed it.

packages/framework/src/Twig/DomainExtension.php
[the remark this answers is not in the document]
tomshopsys: After discussion, we made decision to revert removal of size class.

packages/framework/src/Resources/views/Admin/Content/Brand/list.html.twig
tomshopsys: Ok this can be reverted.

packages/framework/src/Resources/views/Admin/Content/Country/list.html.twig
tomshopsys: After discussion we will unify it later.

packages/framework/src/Twig/DomainExtension.php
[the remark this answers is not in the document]
PetrHeinz: Fixed in a23a533c5

packages/framework/src/Resources/views/Admin/Content/Image/overview.html.twig
PetrHeinz: OK, makes sense

packages/framework/src/Resources/views/Admin/Content/Currency/currencySettings.html.twig
PetrHeinz: :+1:

packages/framework/src/Resources/views/Admin/Content/Brand/list.html.twig
PetrHeinz: Fixed in 5d44d49ba

PetrHeinz:
@tomshopsys I've added the upgrade notes in 9d98289e3 + 2d49a51d3, please review it whether it's sufficient:

image

packages/framework/src/Resources/views/Admin/Content/Order/preview.html.twig
PetrHeinz: After a discussion it seems that it's no OK, waiting for reimplementation.

LukasHeinz:
Reporting bugs:
1) Admin>Product detail: the textarea is short
Snímek obrazovky 2019-07-25 v 12 07 09
2) Admin>Product detail: flags next to radiobuttons are in boxes that makes no sense. Please remove boxes
Snímek obrazovky 2019-07-25 v 12 08 40

PR #1245, opened 2019-07-23
[shopsys] acceptance test helper classes for graphical inputs

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| During admin design face-lift the checkboxes and radiobuttons in administration were redesigned using CSS (see #1241). These changes hide the original inputs which breaks acceptance tests (Selenium can manipulate only elements that are visible).<br>In this PR I've introduced helper classes that solve this problem. I've created classes for both types of inputs, even though they are not used in our current acceptance tests suite (in `shopsys/project-base`).<br>I've decided to keep the helper classes in `shopsys/framework` similarly as the helper classes for PHPUnit introduced in #821 (see also #871).
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No (the BC break was introduced in #1241) 
|Fixes issues| ... 


packages/framework/tests/Test/Codeception/ActorInterface.php
```diff
@@ -0,0 +1,963 @@
+<?php
```
vitek-rostislav: wow, it was really hard to review the whole class :smile: 
And it seems that it will be quite hard to maintain this class when Codeception is upgraded in the future - how can we assure the interface is up to date with possible changes in the generated `AcceptanceTesterActions` trait and with the other modules?

I do not like we need this cumbersome interface only because of the fact that `AcceptanceTester` class is in `project-base`. Could this be solved by https://github.com/shopsys/shopsys/issues/1246?

packages/framework/tests/Test/Codeception/AdminSelectbox.php
```diff
@@ -42,10 +44,66 @@ public static function createByCss(ActorInterface $tester, string $cssSelector):
...
+        $message = sprintf('Label for value "%s" not found in admin selectbox "%s".', $optionValue, $this->cssSelector);
+        Assert::assertNotEmpty($label, $message);
+
+        $this->tester->seeInCss($label, sprintf('.%s .selectric', $wrapperClass));
+    }
+
+    /**
+     * @return string
+     */
+    protected function getWrapperClass(): string
```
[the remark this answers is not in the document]
vitek-rostislav: > Could you maybe describe this in the commit message, please? Or in the method annotation?

The same request applies to similar methods in `AdminCheckbox` and `AdminRadiobutton`

packages/framework/src/Resources/styles/admin/core/form/custom-inputs.less
```diff
@@ -33,7 +35,7 @@ input[type=radio].css-radio + .css-radio__image, input[type=radio].css-radio + .
 		content: "";
 		position: absolute;
 		display: inline-block;
-		top: -9px;
+		top: 0;
```
vitek-rostislav: this change is because of the set dimensions?

packages/framework/tests/Test/Codeception/AdminRadiobutton.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @param string $radioValue
+     */
+    public function assertSelected(string $radioValue): void
+    {
+        $imageElementClass = $this->getImageElementClassByValue($radioValue);
+
+        $this->tester->canSeeElement(['css' => '.' . $imageElementClass]);
+
+        $script = sprintf('return $(".%s").is("input:checked + *")', $imageElementClass);
```
vitek-rostislav: oh, what does this syntax mean? `input:checked + *`

packages/framework/src/Resources/styles/admin/core/form/custom-inputs.less
tomshopsys: i think this is because Selenia can´t find element. These less changes are ok for me.

packages/framework/tests/Test/Codeception/AdminRadiobutton.php
pk16011990: It is combination of https://api.jquery.com/next-adjacent-selector/ and https://api.jquery.com/all-selector/

 `sprintf()` is hard to reading for me in this case...

packages/framework/tests/Test/Codeception/ActorInterface.php
PetrHeinz: I might be helped in #1246, but I don't really know for sure. This will at least give us some way to see whether some methods, that were previously supported, get removed. And of course, if we in future don't need a particular `Actor` method we don't have to add it to the interface. Adding methods to the interface would be considered a BC break as per our promise.

For me, the question was "Which methods should I add to the interface?" I could have chosen only the methods that I actually use in the helper classes but that seemed pretty shortsighted to me. The other obvious choice was all of them (which I've chosen). There could be some arbitrary subset of the supported methods but I couldn't find a reasonable dividing line...

packages/framework/src/Resources/styles/admin/core/form/custom-inputs.less
PetrHeinz: Yes, because I've set the height to 18px, I don't need to set negative `top`.

packages/framework/tests/Test/Codeception/AdminRadiobutton.php
PetrHeinz: Yes, it answers the question whether the element with `$imageElementClass` is directly adjacent to a checked `input`.

The `sprintf` might be a bit hard to read but it was even worse when I've used concatenation - it was pretty easy to miss a quote.

packages/framework/tests/Test/Codeception/AdminSelectbox.php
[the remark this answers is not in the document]
PetrHeinz: Done in 1a264bfd1 and 7ef16601c.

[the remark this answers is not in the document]
PetrHeinz: It's not really because of more selectboxes on a page (the `$this->cssSelector` could handle that), it's more because the particular element is not easily targeted by a CSS selector.

When I'm trying to click a particular item in a menu, I need to target list item with a particular `data-index` attribute which resides in `.selectric-scroll` of a common parent wrapper element with the selectbox (that is defined by the `$this->cssSelector`).
I saw that this targeting was pretty complex, so I've marked the element in question with a generated class via JS to avoid all that complexity.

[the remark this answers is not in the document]
vitek-rostislav: ok, thanks for the clarification!

packages/framework/tests/Test/Codeception/ActorInterface.php
vitek-rostislav: ok, thanks for the answer :+1:

packages/framework/tests/Test/Codeception/AdminRadiobutton.php
```diff
@@ -0,0 +1,89 @@
...
+     * @param string $cssSelector
+     */
+    protected function __construct(ActorInterface $tester, string $cssSelector)
+    {
+        $this->tester = $tester;
+        $this->cssSelector = $cssSelector;
+    }
+
+    /**
+     * @param \Tests\FrameworkBundle\Test\Codeception\ActorInterface $tester
```
vitek-rostislav:
```suggestion
	 * It is important to understand that AdminRadiobutton represents the whole group of radio inputs with the same name.
	 * I.e. you need to use the name or CSS class of the group (and not an ID of a particular radio input) as the $cssSelector
	 * E.g. '[name="product_form[displayAvailabilityGroup][hidden]"]'
     * @param \Tests\FrameworkBundle\Test\Codeception\ActorInterface $tester
```

PetrHeinz:
I've just rebased and force-pushed the branch (using `git rb --onto origin/tg-admin-facelift-issues f8e8a1f49`).

Now, after the switch from `selectric.js` to `select2.js` we don't really have to implement `AdminSelectbox` after all... So this PR should be updated to only help with graphical checkboxes and radiobuttons.

docs/upgrade/UPGRADE-unreleased.md
```diff
@@ -209,9 +209,27 @@ There you can find links to upgrade notes for other versions too.
...
-        - remove `bg/`, `flags/`, `logo.svg`, and `preloader.gif` from `web/assets/admin/images/`
-        - if you need to use these assets yourself, you can find them in `web/assets/bundles/shopsysframework/` (they are copied using `php phing assets` which is a part of standard build targets)
+- clean your project repository after administration image assets were moved from `shopsys/project-base` into `shopsys/framework` repository ([#1243](https://github.com/shopsys/shopsys/pull/1243))
+    - remove `bg/`, `flags/`, `logo.svg`, and `preloader.gif` from `web/assets/admin/images/`
+    - if you need to use these assets yourself, you can find them in `web/assets/bundles/shopsysframework/` (they are copied using `php phing assets` which is a part of standard build targets)
+- update your administration acceptance tests after design face-lift ([#1245](https://github.com/shopsys/shopsys/pull/1245))
+    - you don't have to do anything if you haven't updated the admin acceptance tests or written your own scenarios
```
vitek-rostislav:
```suggestion
    - you don't have to do anything if you haven't modified the admin acceptance tests or written your own scenarios
```

PetrHeinz:
I've rebased and retargeted the PR

PR #1250, opened 2019-07-25
[backend-api] moved backend API Product Controller classes to subdirectory

simara-svatopluk:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This change will be done anyway in #1203 and if we do it before 8.0 is release, then we make life for the implementation teams brighter
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/backend-api/install/install.sh
+++ b/packages/backend-api/install/install.sh
@@ -33,8 +33,8 @@ cp ${INSTALL_DIR}/app/config/packages/oauth2/parameters_oauth.yml.dist ${PROJECT
 echo "Done"
 
 echo "Creating directory src/Shopsys/ShopBundle/Controller/Api/V1..."
-mkdir -p ${PROJECT_BASE_PATH}/src/Shopsys/ShopBundle/Controller/Api/V1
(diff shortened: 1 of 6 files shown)


packages/backend-api/install/install.sh
```diff
@@ -33,7 +33,7 @@ cp ${INSTALL_DIR}/app/config/packages/oauth2/parameters_oauth.yml.dist ${PROJECT
 echo "Done"
 
 echo "Creating directory src/Shopsys/ShopBundle/Controller/Api/V1..."
-mkdir -p ${PROJECT_BASE_PATH}/src/Shopsys/ShopBundle/Controller/Api/V1
+mkdir -p ${PROJECT_BASE_PATH}/src/Shopsys/ShopBundle/Controller/Api/V1/Product
 touch ${PROJECT_BASE_PATH}/src/Shopsys/ShopBundle/Controller/Api/V1/.gitkeep
```
grossmannmartin: What about `.gitkeep` in Product folder? Otherwise, it will not be committed.

PR #1254, opened 2019-07-26
[coding-standards] explicit visibility is part of the general coding-standards

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Having explicit visibility for `const`, `method`, and `property` is now required even in project repositories. It leads to clearer code which will help our users and there is an automatic fixer for it, so it will not require extra work from them.
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Not sure, actually 
|Fixes issues| ... 


PetrHeinz:
@vitek-rostislav Exactly - in #904 :slightly_smiling_face:

PR #1256, opened 2019-07-26
[project-base] enforce strict typing in project repository

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Using [strict typing](https://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.strict) is a general best-practice in PHP. It leads to safer code which will help our users and there is an automatic fixer for it, so it will not require extra work from them, except for fixing wrongly typed variables. <br>Even though there is a lot to do for complete switch to strict typing across the whole codebase (there are few problems in our typing and the change could have serious backward-compatibility implications), we can recommend our users to enable the `declare(strict_types=1)` in their project.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/easy-coding-standard.yml
+++ b/easy-coding-standard.yml
@@ -5,6 +5,10 @@ imports:
 
 parameters:
     skip:
+        PhpCsFixer\Fixer\Strict\DeclareStrictTypesFixer:
+            - 'packages/*'
+            - 'utils/*'
+
         Shopsys\CodingStandards\Sniffs\ConstantVisibilityRequiredSniff:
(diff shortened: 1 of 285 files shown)


easy-coding-standard.yml
```diff
@@ -3,8 +3,15 @@ imports:
     - { resource: 'packages/*/easy-coding-standard.yml' }
     - { resource: 'project-base/easy-coding-standard.yml' }
 
+services:
```
vitek-rostislav: is this change necessary? it is already declared in the imported config from `project-base` :thinking:

PR #1262, opened 2019-07-29
[migrations] add missing strict types and return types into template

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Add missing strict types and return types for migration template.
|New feature| No
|Fixes issues| ...

I know it's a little stupid thing because it doesn't bring anything new, but I like the order in the code 😄 If you don't want this PR, I will accept it.

### DIFF
--- a/packages/migrations/src/Resources/views/Migration/migration.php.twig
+++ b/packages/migrations/src/Resources/views/Migration/migration.php.twig
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace {{ namespace }};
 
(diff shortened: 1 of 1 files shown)


sspooky13:
P.S.: I didn't know why this code https://github.com/shopsys/shopsys/blob/master/packages/migrations/src/Resources/views/Migration/migration.php.twig#L15-L17 has weird indents so I left it that way

PetrHeinz:
@sspooky13 I'd say the indentation is there because we don't want the lines to appear in the generated file.

PR #1244, opened 2019-07-23
[shopsys] prepare releaser for work in multiple branches

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Originally, the new version of Shopsys Framework was always released from the current master. But since we started supporting multiple versions (`7.3`, master (`8.0`), ...), we need to tweak the tool to be prepared for these use cases
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

**Important note:** Beside this PR itself, there are other things that need review:

- [x] new Heimdall tool (http://heimdall:8080/view/Tools/job/tool-monorepo-split-branch/) should be reviewed (it was copied from `tool-monorepo-split` and `SPLIT_BRANCH` parameter has been added instead of hardcoded `master` - `CreateAndPushGitTagReleaseWorker` suggests to use this tool when release is performed on other than `master` branch)
- [x] https://github.com/Symplify/ChangelogLinker/pull/3
- [x] https://github.com/Symplify/MonorepoBuilder/pull/4

**TODO:** I asked @PetrHeinz for help as I am missing an answer to the following question:

- [x] How come that Jenkins build is failing after I introduced the dependency on forked `Sympify\ChangelogLinker` in https://github.com/shopsys/shopsys/commit/d8b271c10c75c22bdb342dd9b10b0dcb3a87a07b? (local composer installation works well) - a suggestion of solution [here](https://github.com/shopsys/shopsys/pull/1244#discussion_r309154778) 
- [x] Change the bullet point in the commit message `upgrade files do not use "Unreleased" phrase anymore` - from ~UPGRADE-unreleased.md renamed to UPGRADE-v8.0.0-dev.md~ to `UPGRADE-unreleased.md renamed to UPGRADE-v7.3.2-dev.md`

- [x] Change commit message "CheckChangelogForTodaysDateReleaseWorker now checkouts to "rc-" branch and pushes the changes automatically" (checkout to rc branch has been extracted into a separate commit by fixup)

- [x] Create a proper PR to Symplify current monorepo
[...]

PR #1264, opened 2019-07-29
[shopsys] releaser now commits package and composer locks

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to ensure that there are no problems with versions of dependencies so we lock them when releasing
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/utils/releaser/src/ReleaseWorker/AfterRelease/RemoveLockFilesReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/AfterRelease/RemoveLockFilesReleaseWorker.php
@@ -37,13 +37,8 @@ final class RemoveLockFilesReleaseWorker extends AbstractShopsysReleaseWorker
         $this->processRunner->run('git rm project-base/package-lock.json');
         $this->commit('removed locked versions of dependencies for unreleased version');
 
-        if ($this->initialBranchName === 'master') {
-            $this->symfonyStyle->note('You need to push the master branch manually, however, you have to wait until the previous (tagged) master build is finished on Heimdall. Otherwise, master-project-base would have never been built from the source codes where there are dependencies on the tagged versions of shopsys packages.');
-            $this->confirm('Confirm you have waited long enough and then pushed the master branch.');
-        } else {
-            $this->symfonyStyle->note(sprintf('You need to push the "%s" branch manually', $this->initialBranchName));
-            $this->confirm(sprintf('Confirm you have pushed the "%s "branch.', $this->initialBranchName));
-        }
+        $this->symfonyStyle->note('You need to push the master branch manually, however, you have to wait until the previous (tagged) master build is finished on Heimdall. Otherwise, master-project-base would have never been built from the source codes where there are dependencies on the tagged versions of shopsys packages.');
+        $this->confirm('Confirm you have waited long enough and then pushed the master branch.');
     }
 
     /**
--- a/utils/releaser/src/ReleaseWorker/AfterRelease/RemoveLockFilesReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/AfterRelease/RemoveLockFilesReleaseWorker.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\Releaser\ReleaseWorker\AfterRelease;
+
+use PharIo\Version\Version;
+use Shopsys\Releaser\ReleaseWorker\AbstractShopsysReleaseWorker;
+use Shopsys\Releaser\Stage;
+
+final class RemoveLockFilesReleaseWorker extends AbstractShopsysReleaseWorker
+{
+    /**
+     * @param \PharIo\Version\Version $version
+     * @return string
+     */
+    public function getDescription(Version $version): string
+    {
+        return 'Remove lock files from the repository, commit the change, and [Manually] push';
+    }
+
+    /**
+     * Higher first
+     * @return int
+     */
+    public function getPriority(): int
+    {
+        return 180;
+    }
+
+    /**
+     * @param \PharIo\Version\Version $version
+     */
+    public function work(Version $version): void
+    {
+        $this->processRunner->run('git rm project-base/composer.lock');
+        $this->processRunner->run('git rm project-base/package-lock.json');
+        $this->commit('removed locked versions of dependencies for unreleased version');
+
+        if ($this->initialBranchName === 'master') {
+            $this->symfonyStyle->note('You need to push the master branch manually, however, you have to wait until the previous (tagged) master build is finished on Heimdall. Otherwise, master-project-base would have never been built from the source codes where there are dependencies on the tagged versions of shopsys packages.');
+            $this->confirm('Confirm you have waited long enough and then pushed the master branch.');
+        } else {
+            $this->symfonyStyle->note(sprintf('You need to push the "%s" branch manually', $this->initialBranchName));
+            $this->confirm(sprintf('Confirm you have pushed the "%s "branch.', $this->initialBranchName));
+        }
+    }
+
+    /**
+     * @return string
+     */
+    public function getStage(): string
+    {
+        return Stage::AFTER_RELEASE;
+    }
+}
--- a/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
@@ -49,7 +49,6 @@ final class CreateAndCommitComposerLockAndPackageLockReleaseWorker extends Abstr
         $this->symfonyStyle->confirm(sprintf('confirm that composer.lock and package-lock.json are pushed to master branch'));
 
         $this->processRunner->run('rm -rf project-base/vendor');
-        $this->processRunner->run('rm -f project-base/composer.lock');
         $this->processRunner->run('composer install');
     }
 
--- a/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
@@ -33,7 +33,7 @@ final class CreateAndCommitComposerLockAndPackageLockReleaseWorker extends Abstr
      */
     public function work(Version $version): void
     {
-        $this->symfonyStyle->note('Removing vendor/, node_modules/, composer.lock and package-lock.json');
+        $this->symfonyStyle->note('Removing vendor/, node_modules/, composer.lock, and package-lock.json');
         $this->processRunner->run('rm -rf project-base/vendor');
         $this->processRunner->run('rm -rf project-base/node_modules');
         $this->processRunner->run('rm -f project-base/composer.lock');
--- a/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
@@ -33,7 +33,7 @@ final class CreateAndCommitComposerLockAndPackageLockReleaseWorker extends Abstr
      */
     public function work(Version $version): void
     {
-        $this->symfonyStyle->note('Removing vendor and node_modules folders and composer.lock and package-lock.json');
+        $this->symfonyStyle->note('Removing vendor/, node_modules/, composer.lock and package-lock.json');
         $this->processRunner->run('rm -rf project-base/vendor');
         $this->processRunner->run('rm -rf project-base/node_modules');
         $this->processRunner->run('rm -f project-base/composer.lock');
--- a/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\Releaser\ReleaseWorker\Release;
+
+use PharIo\Version\Version;
+use Shopsys\Releaser\ReleaseWorker\AbstractShopsysReleaseWorker;
+use Shopsys\Releaser\Stage;
+
+final class CreateAndCommitComposerLockAndPackageLockReleaseWorker extends AbstractShopsysReleaseWorker
+{
+    /**
+     * @param \PharIo\Version\Version $version
+     * @return string
+     */
+    public function getDescription(Version $version): string
+    {
+        return 'Create and commit composer.lock and package-lock.json and [Manually] push it';
+    }
+
+    /**
+     * Higher first
+     * @return int
+     */
+    public function getPriority(): int
+    {
+        return 630;
+    }
+
+    /**
+     * @param \PharIo\Version\Version $version
+     */
+    public function work(Version $version): void
+    {
+        $this->processRunner->run('rm -rf project-base/vendor');
+        $this->processRunner->run('rm -rf project-base/node_modules');
+        $this->processRunner->run('rm -f project-base/composer.lock');
+        $this->processRunner->run('rm -f project-base/package-lock.json');
+        $this->processRunner->run('cd project-base && composer install && npm install');
+
+        $this->processRunner->run('git add -f project-base/composer.lock');
+        $this->processRunner->run('git add -f project-base/package-lock.json');
+        $message = sprintf('locked versions of dependencies for %s release', $version->getVersionString());
+        $this->commit($message);
+
+        $this->symfonyStyle->note('push last commit with composer and package locks');
+        $this->symfonyStyle->confirm(sprintf('confirm that composer.lock and package-lock.json are pushed to "%s" branch', $this->initialBranchName));
+
+        $this->processRunner->run('rm -rf project-base/vendor');
+        $this->processRunner->run('rm -f project-base/composer.lock');
+        $this->processRunner->run('composer install');
+    }
+
+    /**
+     * @return string
+     */
+    public function getStage(): string
+    {
+        return Stage::RELEASE;
+    }
+}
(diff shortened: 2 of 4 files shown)


utils/releaser/src/ReleaseWorker/AfterRelease/RemoveLockFilesReleaseWorker.php
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * @param \PharIo\Version\Version $version
+     */
+    public function work(Version $version): void
+    {
+        $this->processRunner->run('git rm project-base/composer.lock');
+        $this->processRunner->run('git rm project-base/package-lock.json');
+        $this->commit('removed locked versions of dependencies for unreleased version');
+
+        if ($this->initialBranchName === 'master') {
```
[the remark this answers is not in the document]
DavidOstrozlik: Removed it from `SetFrameworkBundleVersionToDevReleaseWorker`

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,73 @@
...
+        $versionString = $version->getVersionString();
+
+        $this->processRunner->run('mkdir temp');
+        foreach (self::PACKAGES_WITHOUT_MONOREPO_AND_PROJECT_BASE as $packageName) {
+            $this->processRunner->run('cd temp && git clone https://github.com/shopsys/' . $packageName . '.git');
+            $this->processRunner->run('cd temp/' . $packageName . ' && git checkout ' . $this->initialBranchName . ' && git tag ' . $versionString);
+            $this->processRunner->run('cd temp/' . $packageName . ' && git push origin ' . $versionString);
+        }
+
+        $this->symfonyStyle->note('Wait for packagist to get new versions of all packages excluding monorepo and project-base');
```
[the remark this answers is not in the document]
DavidOstrozlik: I don't know  about this one.. do we have this logic somewhere in releaser?

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Shopsys\Releaser\ReleaseWorker\Release;
+
+use PharIo\Version\Version;
+use Shopsys\Releaser\ReleaseWorker\AbstractShopsysReleaseWorker;
+use Shopsys\Releaser\Stage;
+
+final class CreateAndPushGitTagsExceptProjectBaseReleaseWorker extends AbstractShopsysReleaseWorker
+{
+    private const PACKAGES_WITHOUT_MONOREPO_AND_PROJECT_BASE = [
```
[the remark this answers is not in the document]
DavidOstrozlik: Fixed it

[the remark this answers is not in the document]
PetrHeinz: It's already missing `shopsys/backend-api`. We should definitely aim to have only one list to maintain.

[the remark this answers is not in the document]
DavidOstrozlik: @PetrHeinz 1 minute too late. It's changed already. :smile:

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public function work(Version $version): void
+    {
+        $versionString = $version->getVersionString();
+
+        $this->processRunner->run('mkdir temp');
+        foreach (self::PACKAGES_WITHOUT_MONOREPO_AND_PROJECT_BASE as $packageName) {
+            $this->processRunner->run('cd temp && git clone https://github.com/shopsys/' . $packageName . '.git');
+            $this->processRunner->run('cd temp/' . $packageName . ' && git checkout ' . $this->initialBranchName . ' && git tag ' . $versionString);
+            $this->processRunner->run('cd temp/' . $packageName . ' && git push origin ' . $versionString);
```
PetrHeinz: Can there be some way of ensuring that it will tag the right commits? Remember that tagging a package is something **we cannot revert**. Maybe instead of pushing the tags right away, I would create an array of package names and commit msgs (package name is `sprintf('shopsys/%s', $packageName)` and you can get the commit msg by executing `'git show ' . $versionString . ' --format="%s"'`).

After the developer confirms that correct commits are tagged, the releaser could loop through all repositories in `temp/`, push the tags and remove `temp/`. This would seem a lot safer to me.

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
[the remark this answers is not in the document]
vitek-rostislav: well, depends on what you mean by "this logic" :grin: check of the packagist is implemented in the mentioned `CheckPackagesOnPackagistReleaseWorker`, the while loops are on our wishlist, I think it would be really helpful to implement it here, or not? :slightly_smiling_face:

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -5,26 +5,53 @@
...
+     * @var string[]
+     */
+    private const EXCLUDED_PACKAGES = [
+        // excluded from the initial tagging as there needs to be another commit with composer.lock and package-lock.json
+        // @see https://github.com/shopsys/shopsys/pull/1264
+        'shopsys/shopsys',
+        'shopsys/project-base',
+        'shopsys/product-feed-interface',
+        'shopsys/phpstorm-inspect',
+        // forks
```
vitek-rostislav: from here down, it is DRY violation again, however, it might be justified as the list may differ from the one used in `TravisStatusReporter`. 

But maybe it would make sense to create a list of by default excluded packages in `PackageProvider` and then use the default list here and in `TravisStatusReporter` a let developers to extend the array somehow (like here we need to add `shopsys/shopsys` and `shopsys/project-base` to the array :thinking:

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
PetrHeinz: Not all packages will actually have the same commit. There is nothing changed in coding-standards and monorepo-tools, as they have no dependencies on `shopsys/*` packages...

PetrHeinz:
I think this should be retargeted into `7.3` so it can help us with releasing `v7.3.2`, similarly to #1244

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
DavidOstrozlik: I personally think that it should not be in one basic array as (as you said) the arrays may differ in the future

vitek-rostislav: ok :)

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -76,22 +71,37 @@ public function getPriority(): int
...
+                $this->processRunner->run('cd temp/' . $packageName . ' && git push origin ' . $versionString);
+            }
 
-        $this->processRunner->run('rm -r temp');
+            $this->processRunner->run('rm -r temp');
+            $this->symfonyStyle->note('Wait for packagist to get new versions of all packages excluding monorepo and project-base');
+            $this->confirm('Confirm that there are new versions of all packages excluding monorepo and project-base');
+        } else {
+            $packageNamesWithProblemsMessage = sprintf('package%s %s', count($packageNamesWithProblems) === 1 ? '' : 's', implode(', ', $packageNamesWithProblems));
+            $this->confirm(sprintf('Please fix the problem in %s and split the monorepo again This step will be repeated.', $packageNamesWithProblemsMessage));
```
vitek-rostislav: missing dot before "This step..."

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $packages = $this->packageProvider->getPackagesByOrganization('shopsys', self::EXCLUDED_PACKAGES);
+        $packages = str_replace('shopsys/', '', $packages);
+
+        $versionString = $version->getVersionString();
+
+        $this->processRunner->run('mkdir temp');
+        $packageNamesWithProblems = [];
+        foreach ($packages as $packageName) {
+            $this->processRunner->run('cd temp && git clone https://github.com/shopsys/' . $packageName . '.git');
```
simara-svatopluk: This line takes a lot of time sometimes, it would be nice to have a progressbar of the `git clone` here. If not progress bar then at least a note like "Clonning shopsys/xxx, this can take couple of minutes" and annimation that would tell me the process is working. As an animation it would be nice to have dots `.`->`..`->`...`->`.`->`..`->and again, and again....

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,114 @@
...
+            $this->processRunner->run('cd temp/' . $packageName . ' && git checkout master && git tag ' . $versionString);
+            $this->processRunner->run('cd temp/' . $packageName . ' && git log --graph --oneline --decorate=short --color | head', true);
+            $pushTag = $this->symfonyStyle->ask('Is the tag on right commit and should be pushed?', 'yes');
```
simara-svatopluk: It keeps asking me "Is the tag on right commit?" But I have no idea about what package is it talking about. Please ass there this information

utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        return 630;
+    }
+
+    /**
+     * @param \PharIo\Version\Version $version
+     */
+    public function work(Version $version): void
+    {
+        $this->processRunner->run('rm -rf project-base/vendor');
```
simara-svatopluk: When it started running, I thought it was an error and I have to do something manually. After 10 seconds, the PC fan started to run and I realized it probably does something.

Please add here an information that it is running, optionally animation so I won't be tempted to do something

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,114 @@
...
+                $packageNamesWithProblems[] = $packageName;
+            }
+        }
+
+        if (count($packageNamesWithProblems) === 0) {
+            foreach ($packages as $packageName) {
+                $this->processRunner->run('cd temp/' . $packageName . ' && git push origin ' . $versionString);
+            }
+
+            $this->processRunner->run('rm -r temp');
```
simara-svatopluk: I didn't remove the `temp` directory (could happen on some error), and the following worker `CreateAndCommitComposerLockAndPackageLockReleaseWorker` commited them. I really recommend different temp directory

utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
```diff
@@ -0,0 +1,62 @@
...
+        $this->processRunner->run('git add -f project-base/composer.lock');
+        $this->processRunner->run('git add -f project-base/package-lock.json');
+        $message = sprintf('locked versions of dependencies for %s release', $version->getVersionString());
+        $this->commit($message);
+
+        $this->symfonyStyle->note('push last commit with composer and package locks');
+        $this->symfonyStyle->confirm(sprintf('confirm that composer.lock and package-lock.json are pushed to master branch'));
+
+        $this->processRunner->run('rm -rf project-base/vendor');
+        $this->processRunner->run('rm -f project-base/composer.lock');
```
simara-svatopluk: I guess this file shouldn't be removed because it will be removed by `RemoveLockFilesReleaseWorker`

(if it should be deleted, then git shows for `git status` deleted file that is uncommited, and that is wrong)

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,114 @@
...
+use Shopsys\Releaser\ReleaseWorker\AbstractShopsysReleaseWorker;
+use Shopsys\Releaser\Stage;
+
+final class CreateAndPushGitTagsExceptProjectBaseReleaseWorker extends AbstractShopsysReleaseWorker
+{
+    /**
+     * Packages that are not released - old packages or forks
+     * @var string[]
+     */
+    private const EXCLUDED_PACKAGES = [
```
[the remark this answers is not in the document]
DavidOstrozlik: I added the note about what package are you tagging, so I think that is safe because you'll know what package are you tagging and can add it to the excluded ones, if you don't want to tag it.

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
DavidOstrozlik: I added a note

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
DavidOstrozlik: I added name of the package to the note

utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
DavidOstrozlik: I added a note as well

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
```diff
@@ -0,0 +1,115 @@
...
+        $packageNames = str_replace('shopsys/', '', $packages);
+
+        $versionString = $version->getVersionString();
+
+        $tempDirectory = trim($this->processRunner->run('mktemp -d -t shopsys-release-XXXX'));
+        $packageNamesWithProblems = [];
+        foreach ($packageNames as $packageName) {
+            $this->symfonyStyle->note(sprintf('Cloning shopsys/%s. This can take a while.', $packageName));
+            $this->processRunner->run(sprintf('cd %s && git clone https://github.com/shopsys/%s.git', $tempDirectory, $packageName));
+            $this->processRunner->run(sprintf('cd %s/%s && git checkout master && git tag %s', $tempDirectory, $packageName, $versionString));
```
PetrHeinz: It would rather have the path as a single parameter:
```suggestion
            $this->processRunner->run(sprintf('cd %s && git checkout master && git tag %s', $tempDirectory . '/' . $packageName, $versionString));
```

DavidOstrozlik: For me `%s/%s` is more readable because I know just from reading that I want to go into one particular folder in some other folder so I don't want to do this.

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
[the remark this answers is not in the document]
PetrHeinz: @DavidOstrozlik I do agree that it's safe to use and it's a consistent approach in the releaser. But the list of packages is currently in more than one place (eg. `monorepo.md`, `.ci/monorepo_functions.sh`, or `TravisStatusReporter`). This will create problems for us in the future imo and is a clear violation of DRY.

Please, create an issue explaining the situation so we'll fix it eventually.

utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
PetrHeinz: The note is a bit ill-structured, but fine.

I'd use a list: *Removing vendor/, node_modules/, composer.lock, and package-lock.json* instead of *Removing vendor and node_modules folders and composer.lock and package-lock.json*...

utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php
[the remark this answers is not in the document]
DavidOstrozlik: I added #1269  for this.

utils/releaser/src/ReleaseWorker/Release/CreateAndCommitComposerLockAndPackageLockReleaseWorker.php
DavidOstrozlik: I changed it to what you wrote.

PetrHeinz: You forgot the [oxford comma](https://en.wikipedia.org/w/index.php?title=Oxford_comma).

PR #1273, opened 2019-08-01
[project-base] enable entity extension from components

malyMiso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| setup mappings for entities in project components.  It saves a lot of time searching, why e.g extension of Image doesn't work :slightly_smiling_face: I followed instructions from [Entity extension](https://github.com/shopsys/shopsys/blob/master/docs/extensibility/entity-extension.md) but I was not able to make it work. 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/docs/upgrade/UPGRADE-v7.3.2-dev.md
+++ b/docs/upgrade/UPGRADE-v7.3.2-dev.md
@@ -4,3 +4,24 @@ This guide contains instructions to upgrade from version v7.3.1 to v7.3.2-dev.
 
 **Before you start, don't forget to take a look at [general instructions](/UPGRADE.md) about upgrading.**
 There you can find links to upgrade notes for other versions too.
+
+## [shopsys/framework]
+
+### Configuration
+- update your `app/config/packages/doctrine.yml` ([#1273](https://github.com/shopsys/shopsys/pull/1273))
+    ```diff
+       ShopsysShopBundle:
+           type: annotation
+           dir: '%shopsys.root_dir%/src/Shopsys/ShopBundle/Model'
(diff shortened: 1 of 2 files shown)


PetrHeinz:
Good idea, haven't thought of that 👍 Thanks @sspooky13! Seeing that the PR changes something in the `project-base` repository, we should add an upgrade note before merging.

vitek-rostislav:
Hi @LukasHeinz, well we were discussing this with @TomasLudvik for a long time. The identification of BC breaks is sometimes really hard, but in the end, we agreed that a change of configuration should be always a BC break.

Maybe we need some third opinion here, e.g. @PetrHeinz, could you please add your thoughts here?

PetrHeinz:
@vitek-rostislav There is [a whole section](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md#project-base-repository) of our BC promise dedicated to changes in `project-base` repository.

> The project-base repository is not meant to be extended or depended upon. For this reason, the **changes in it are not subject to the rules of this BC promise**.

According to our BC promise, this is not a BC break.

Also, targeting PRs such as these will really clutter the upgrade notes for major version with non-critical changes, which will lead to a messy upgrading imo. Another consequence will be postponing the fixes for our users. I see this as a fix of a bug in `shopsys/project-base` - it's our responsibility to notify the users as soon as possible and explain how to fix the issue (seeing that we cannot do that for them). A patch release is a great opportunity for that imo.

I would definitely retarget this to `7.3`, to help users that have their projects built on the LTS version.

PR #1276, opened 2019-08-01
[shopsys] docker-compose-win.yml.dist: remove php.ini volume mount

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The mounting of `php.ini` is not necessary as it is copied into the Docker image during build (the mounting was removed in #497 but reappeared 3 weeks later in #499 - possibly during a rebase?).
|New feature| No 
|Fixes issues| ... 


PetrHeinz:
@vitek-rostislav I'm not sure there is anything to test, actually. It works exactly like that on the other platforms... There is nothing Windows specific here.

PR #1289, opened 2019-08-05
[shopsys] composer.json enhancements (alphabetical sorting of extensions, 4 spaces as indentation)

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| extensions were not sorted alphabetically and monorepo composer.json was not indented by 4 spaces.
|New feature| No 
|Fixes issues| ... 


PetrHeinz:
If we want to keep packages alphabetically ordered when adding them via `composer require`, we should use the `sort-packages` config option in Composer. See [official docs](https://getcomposer.org/doc/06-config.md#sort-packages).

vitek-rostislav:
Hi @PetrHeinz, thanks for the tip, however, we can do this in a separate PR, am I right? :slightly_smiling_face:

PetrHeinz:
@vitek-rostislav Sure thing :+1: It seemed related to eaa4f96 at first, but as I see now it only sorts the extensions, not packages.

TomasLudvik:
Hello @vitek-rostislav,

- yes you are right and I have updated CM
- it is not BC break so I have targeted it so it gets in 8.0.1

vitek-rostislav:
Hi @TomasLudvik 
> it is not BC break so I have targeted it so it gets in 8.0.1

Then it should be targeted to `7.3`, or not? :thinking:

PetrHeinz:
@vitek-rostislav To summarize:
If it fixes a bug and it's not a BC break, it should be targeted to a supported patch where it occurs first (i.e. `7.3` or `8.0` if it was introduced in version 8).
Otherwise, if it's not a BC break, it should be targeted to `master`.
Otherwise, it should be targeted to next major (i.e. `9.0`).

sspooky13:
Hello guys,
as said @PetrHeinz, it looks like only extensions are alphabetically sorted so I sorted only them

vitek-rostislav:
Hi, after discussion with @DavidOstrozlik, we have this consensus: I will separate the first commit (adding the missing dom extension) into a new PR that will be targeted to `7.3` branch (as this is really a bugfix and it will get to the higher versions in the near future once the `7.3` branch is merged into `8.0`, etc.). The other two commits (sorting of extensions in composer.jsons and composer.json indentations) will remain in this branch and this PR will be targeted to master :wink:

PR #1291, opened 2019-08-05
[framework] Fix error for less>=3.0.1

hason:
https://github.com/less/less.js/issues/3188

| Q             | A
| ------------- | ---
|Description, reason for the PR| less.js < 3.0 automaticaly detects if imported file is less or css. But from 3.0.1 it should be defined by `(css)` before url to avoid errors during build, see https://github.com/less/less.js/issues/3188
|New feature| No
|Fixes issues| -

### DIFF
--- a/packages/framework/src/Resources/styles/admin/core/variables.less
+++ b/packages/framework/src/Resources/styles/admin/core/variables.less
@@ -13,7 +13,7 @@
 /*******************************************************************
 // FONTS
 *******************************************************************/
-@import url('https://fonts.googleapis.com/css?family=Roboto:400,500&subset=latin-ext');
+@import (css) url('https://fonts.googleapis.com/css?family=Roboto:400,500&subset=latin-ext');
 
 @font: "Roboto", Arial, Helvetica, sans-serif;
 


vitek-rostislav:
Hi @hason, thanks for the PR :+1: could anyone please provide a better description for a person like me? I am no frontend developer, however, I am trying to understand this. Does it mean something (import of fonts) is broken at the moment? In such a case the PR should be targetted to `8.0` branch as the problem is there as well, am I right @tomshopsys?

tomshopsys:
@vitek-rostislav 
Description of problem: less.js < 3.0 automaticaly detects if imported file is less or css. But from 3.0.1 it should be defined by (css) before url to avoid errors during build.

And is it true to add this PR to 8.0 too.

vitek-rostislav:
Hi @hason, we wanted to test your modification, however, we were not able to reproduce the original problem so even though I believe your solution helps (based on the linked issue and on the fact that everything works in your branch), I am not able to confirm that.

Could you please provide us exact steps to reproduce the original problem?

PR #1302, opened 2019-08-08
[project-base] Dockerfile: error demonstration resulting in unbuildable image

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This is a demonstration of a problem I've noticed in https://github.com/shopsys/demoshop/pull/59#issuecomment-519141095 - the Docker image for `php-fpm` doesn't get rebuilt as it is loaded from cache (cache from `master` build is used for new PRs).
|New feature| No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Yes (it's actually just a general break of the application) 
|Fixes issues| ... 

### DIFF
--- a/project-base/docker/php-fpm/Dockerfile
+++ b/project-base/docker/php-fpm/Dockerfile
@@ -25,24 +25,8 @@ COPY ${project_root}/docker/php-fpm/docker-install-composer /usr/local/bin/docke
 RUN chmod +x /usr/local/bin/docker-install-composer && \
     docker-install-composer
 
-# libpng-dev needed by "gd" extension
-# libzip-dev needed by "zip" extension
-# libicu-dev for intl extension
-# libpg-dev for connection to postgres database
-# autoconf needed by "redis" extension
-RUN apt-get install -y \
-    libpng-dev \
-    libjpeg-dev \
-    libfreetype6-dev \
-    libzip-dev \
-    libicu-dev \
-    libpq-dev \
-    vim \
-    nano \
-    mc \
-    htop \
-    autoconf && \
-    apt-get clean
+# fail instead of installing required PHP extensions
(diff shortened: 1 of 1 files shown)


PetrHeinz:
The [Travis CI build passes](https://travis-ci.org/shopsys/shopsys/builds/569249851) even with [the error in the `Dockerfile`](https://github.com/shopsys/shopsys/pull/1302/files#diff-50a0e02c146dc64c2a172b42022589faR28-R29). This is a major flaw in our Travis build configuration (since #1031).

Fortunately, we have our own CI build that check branches in our repository. However, these checks do not run on PR from forks (i.e. PRs from external contributors). We should be super-careful with such PRs that modify `Dockerfile` of `php-fpm`, otherwise we would notice the issue only after it's merged to `master`.

Further explanation and a suggestion on different ways to handle this can be found in https://github.com/shopsys/demoshop/pull/59#issuecomment-519141095.

grossmannmartin:
Hi, thanks for reporting. 
To me, it makes the most sense to omit the whole "cache image" thing.
But if the change prolongs the time of the build, we can cache images with some hash suffix (for example crc of Dockerfile)

PetrHeinz:
@grossmannmartin Let's see how long it takes... On the demoshop it seemed that it's even a bit faster as the saving and loading of the cache is a quite long operation. But I was saving both `php-fpm` and `elasticsearch` images, so might be bit slower now...

PetrHeinz:
@grossmannmartin Sadly, creating a unique hash is not that easy. Checksum of Dockerfile is not enough. In principle, you would have to compute a checksum of the whole Docker context along with the Dockerfile. And that would change with almost every PR...

PR #941, opened 2019-04-10
improve Phing property is-multidomain to detect domains automatically

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| You dont need to set is-multidomain property in buld.xml anymore.
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No 
|Fixes issues| ... 
|Standards and tests pass| Yes
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1309, opened 2019-08-09
[shopsys] domain configuration is autoloaded in build.xml

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The `is-multidomain` and `translations.dump.locales` properties are now loaded automatically using a new Symfony command `shopsys:domains:info` instead of having developers set them up manually. Simplifies configuration of domains and locales. Also provides a way to show info about domains via CLI.
|New feature| Yes 
|Fixes issues| closes #941 

### DIFF
--- a/packages/framework/src/Command/DomainInfoCommand.php
+++ b/packages/framework/src/Command/DomainInfoCommand.php
@@ -21,6 +21,7 @@ class DomainInfoCommand extends Command
     protected const ARG_ID = 'domainId';
 
     protected const OPTION_DEDUPLICATE = 'deduplicate';
+    protected const OPTION_ONELINE = 'oneline';
 
     protected const RETURN_CODE_OK = 0;
     protected const RETURN_CODE_ERROR = 1;
@@ -54,7 +55,8 @@ class DomainInfoCommand extends Command
             ->setDescription('Loads and displays domain info.')
             ->addArgument(static::ARG_PROPERTY_NAME, InputArgument::OPTIONAL, 'Property that should be loaded', 'id')
             ->addArgument(static::ARG_ID, InputArgument::OPTIONAL, 'Domain ID (if omitted, the command will output all values)')
-            ->addOption(static::OPTION_DEDUPLICATE, 'd', InputOption::VALUE_NONE, 'Return only unique property values (sorted alphabetically)');
+            ->addOption(static::OPTION_DEDUPLICATE, 'd', InputOption::VALUE_NONE, 'Return only unique property values (sorted alphabetically)')
+            ->addOption(static::OPTION_ONELINE, 'o', InputOption::VALUE_NONE, 'Return property values on one line separated by tabs');
     }
 
     /**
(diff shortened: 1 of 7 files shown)


packages/framework/src/Command/DomainInfoCommand.php
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            $propertyValues[] = $propertyAccessor->getValue($domainConfig, $propertyName);
+        }
+
+        if (!$input->getOption(static::OPTION_ALL)) {
+            sort($propertyValues);
+            $propertyValues = array_unique($propertyValues);
+        }
+        $io->writeln(implode("\t", $this->formatPropertyValues($propertyValues)));
```
grossmannmartin: Wouldn't be better to print each value on a separate line? It's then easier to use this command in shell scripts like (without need to use regular expressions / awk, ... )
```
for x in `./bin/console shopsys:domains:info`; do ..... 
```

and to me personally, it's more understandable separated with new lines, then tabs

packages/framework/src/Command/DomainInfoCommand.php
```diff
@@ -0,0 +1,149 @@
...
+     * @return \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig[]
+     */
+    protected function getDomainConfigs(InputInterface $input): array
+    {
+        $domainConfigs = $this->domain->getAllIncludingDomainConfigsWithoutDataCreated();
+
+        $domainId = $input->getArgument(static::ARG_ID);
+        if ($domainId !== null) {
+            foreach ($domainConfigs as $domainConfig) {
+                if ($domainId === (string)$domainConfig->getId()) {
```
grossmannmartin:
```suggestion
                if ($domainConfig->getId() === (int)$domainId) {
```
(if you accept this suggestion, don't forget to change sprintf on line 111)

PetrHeinz: I would rather make it fail when domain ID `"1not-really-a-number"` is provided

packages/framework/src/Command/DomainInfoCommand.php
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure(): void
+    {
+        $this
+            ->setDescription('Loads and displays domain info.')
+            ->addArgument(static::ARG_PROPERTY_NAME, InputArgument::OPTIONAL, 'Property that should be loaded', 'id')
+            ->addArgument(static::ARG_ID, InputArgument::OPTIONAL, 'Domain ID (if omitted, the command will output all distinct values)')
+            ->addOption(static::OPTION_ALL, 'a', InputOption::VALUE_NONE, 'Display all property values (without sorting and deduplication)');
```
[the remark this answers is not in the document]
PetrHeinz: you're right about that, fixed in b417c8c19

packages/framework/src/Command/DomainInfoCommand.php
PetrHeinz: :+1: Good idea. However, because of how Phing works with the output I left the possibility of tab-seaprated output using `--oneline` option. See 27578e7bf

packages/framework/src/Command/DomainInfoCommand.php
grossmannmartin: that's fine by me :)

grossmannmartin:
@PetrHeinz, Merge branch `master` into `ph-auto-domain-config` is intentional or by mistake?
If intentional why it's not just rebased?

PetrHeinz:
@grossmannmartin I've merged it so the conflict gets resolved and Travis CI starts checking the PR. I didn't want to rebase & force push it to not complicate your review.

PR #1313, opened 2019-08-09
[project-base] force locale for number formatting

### LINKED ISSUE #1184: MoneyType should work with different locale
### What is happening

If we have set locale that contains decimal_separator comma instead of dot, Money functionality starts to fail.
### Expected result

Money will work also with different locale set

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When server has "wrong" locale, `Money` or sitemap does not work properly. For more info see #1184
|New feature| No
|Fixes issues| closes #1184

### DIFF
--- a/docs/upgrade/UPGRADE-v8.1.0-dev.md
+++ b/docs/upgrade/UPGRADE-v8.1.0-dev.md
@@ -146,4 +146,7 @@ There you can find links to upgrade notes for other versions too.
         +       arguments:
         +           $useInlineEditation: false
         ```
+- add `setlocale(LC_NUMERIC, 'en_US.utf8');` in your `Bootstrap.php` right behind `setlocale(LC_CTYPE, 'en_US.utf8');` ([#1313](https://github.com/shopsys/shopsys/pull/1313/))
+    - add [`\Tests\ShopBundle\Unit\NumberFormattingTest`](https://github.com/shopsys/shopsys/blob/master/project-base/tests/ShopBundle/Unit/NumberFormattingTest.php)
+
 [shopsys/framework]: https://github.com/shopsys/framework
--- a/project-base/app/Bootstrap.php
+++ b/project-base/app/Bootstrap.php
@@ -15,6 +15,7 @@ use Symfony\Component\HttpFoundation\Request;
 require_once __DIR__ . '/autoload.php';
 
 setlocale(LC_CTYPE, 'en_US.utf8');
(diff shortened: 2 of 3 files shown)


pk16011990:
I totaly agree with `LC_ALL`, but I dont know why was setted only `LC_CTYPE`...

One of our server has this behaviour:
```bash
php -a
Interactive mode enabled

php > var_dump(sprintf("%01.2f", 123456.789));
string(9) "123456,79"
php > var_dump(setlocale(LC_ALL, 0));
string(11) "cs_CZ.UTF-8"
php > var_dump(setlocale(LC_NUMERIC, 0));
string(11) "cs_CZ.UTF-8"
php > setlocale(LC_NUMERIC, 'en_US.utf8');
php > var_dump(sprintf("%01.2f", 123456.789));
string(9) "123456.79"
```

With `PHP 7.1.16-1+0~20180405085419.16+stretch~1.gbp4937aa` and `Debian GNU/Linux 9 (stretch)`

maybe it depends on kind of locales instaled in system (same name but different behaviour?)... However the problem is very difficult to notice.

grossmannmartin:
Hi @pk16011990, Thank you very much for providing detailed information.

For some reason, it's not reflecting the server settings to me... 

I was able to reproduce the problem. But then more questions occurred to me.

setLocale works as expected only if the locale is actually installed on the system. For example I don't have any en_US locale installed:

```sh
$ locale -a
C
cs_CZ.utf8
C.UTF-8
POSIX
```

and the testing scenario you provided:

```
php > var_dump(setlocale(LC_NUMERIC, 0));
string(11) "cs_CZ.UTF-8"

php > var_dump(sprintf("%01.2f", 123456.789));
string(9) "123456,79"

php > var_dump(setlocale(LC_NUMERIC, 'en_US.UTF-8'));
bool(false)

php > var_dump(sprintf("%01.2f", 123456.789));
string(9) "123456,79"

php > var_dump(setlocale(LC_NUMERIC, 0));
string(11) "cs_CZ.UTF-8"
```

I'm really worried about the false confidence here... 

But as I was saying, I'm ok with merging this, but you still need be careful about server setting, though.

TomasLudvik:
Hello @grossmannmartin and @pk16011990,

I do not want to force LC_ALL as I am not sure with outcome. I have added super easy unit test to check if formatting is OK. What do you think guys?

PR #1317, opened 2019-08-09
[shopsys] demo data: 1st domain has EUR, 2nd domain has CZK, other domains have EUR

### LINKED ISSUE #305: Demo data - currencies on domains
### What is happening

The prices on the first (eng) domain are in CZK currency and prices on the second (czech) domain are in EUR currency.

### Expected result

There should be CZk currency on the second domain and EUR currency on the first domain

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| 1st domain is EN and currency EUR is more common then CZK... And 2nd CZ domain has CZK...
|New feature| No
|Fixes issues| closes #305


vitek-rostislav:
Thank you for this PR @pk16011990! I'll try to push it to the upcoming sprint so it is processed as soon as possible!

LukasHeinz:
@pk16011990 , thank you for this PR.

To validate this PR I would need a functional project on top of this branch. @vitek-rostislav please, can you prepare it for me?

packages/framework/src/Migrations/Version20180603135342.php
```diff
@@ -14,7 +14,7 @@ public function up(Schema $schema)
     {
         $currenciesCount = $this->sql('SELECT count(*) FROM currencies')->fetchColumn(0);
         if ($currenciesCount <= 0) {
-            $this->sql('INSERT INTO currencies (id, name, code, exchange_rate) VALUES (1, \'Česká koruna\', \'CZK\', 1)');
+            $this->sql('INSERT INTO currencies (id, name, code, exchange_rate) VALUES (1, \'Euro\', \'EUR\', 1)');
```
PetrHeinz: Changing migrations isn't really safe, a new migration should be generated if necessary.

pk16011990: I know, but i this case... New migration would cause really big problem for existing projects. And new projects install this migration with right currency.

vitek-rostislav:
Hi @LukasHeinz, I have prepared the running instance and sent the link to you via email :wink:

packages/framework/src/Migrations/Version20180603135342.php
PetrHeinz: But this introduces a major BC break. If this PR would be just about changing data fixture, it would modify only the `project-base` repository and thus would not complicate upgrading.

pk16011990: This changes (whole PR) will be complicated for projects... It's my point of view... And I want to do final solution, not half-solution because BC break. Will it be updated in next major releases without confusing code?

PetrHeinz: In my opinion, it will be similarly difficult both ways. When the modifications are only in the project repository, users will at least have the option to omit this on their project if they choose to (as the project will still work without the changes).

To be honest, I don't really see having CZK with ID=1 and EUR with ID=2 as a half-solution. This change should be about domains and that can be resolved using demo-data only.

s3tezsky:
As @PetrHeinz mentioned in [comment](https://github.com/shopsys/shopsys/pull/1317/files#r312872069), changing old migration script is not OK for us.

I have also tried to do it just by changing datafixtures which was much cleaner but many issues still remained.

Due to hardoced curency to domain it will affect many tests which we cannot solve in pretty way (e.g. divide prices by currency rate) and as a result it will produce tons on upgrade notes which do not seem fine for us neither for clients.

As a better solution I would like to see independent currencies from domain. Then there would be uncomfortable update notes more reasonable.

For now this PR is too much pain for almost no result.

LukasHeinz:
In my point of view the result of this PR is significant - the SSFW starts making sense in terms of domains and currencies.

s3tezsky:
We could reach this goal almost painless if we had separate _demo data_ (for presentation) and _testing data_. Then we could change _demo data_ as we want and without breaking tests.

We will delay this PR before we solve the issue about separating datasets.

PetrHeinz:
@s3tezsky There is no reason for the currency mismatch in the testing data set, so in my point of view the separation of the data sets would just postpone and hide the actual issue...

henzigo:
Hello, do you plan to merge this PR?

TomasLudvik:
Hello, after changing demodata there is much easier way to do this, as @vitek-rostislav introduced in #1542. I am closing this PR.

PR #1318, opened 2019-08-12
[shopsys] UpdateUpgradeReleaseWorker: added note about checking upgrade instructions from lower versions

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We need to ensure that upgrade instructions for patch versions are included in the all relevant `UPGRADE` files
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/utils/releaser/src/ReleaseWorker/ReleaseCandidate/UpdateUpgradeReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/ReleaseCandidate/UpdateUpgradeReleaseWorker.php
@@ -97,7 +97,8 @@ final class UpdateUpgradeReleaseWorker extends AbstractShopsysReleaseWorker
             'Typically, you need to:
             - check the correctness of the order of Shopsys packages and sections, 
             - check whether there are no duplicated instructions for modifying docker related files, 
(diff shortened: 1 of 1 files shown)


vitek-rostislav:
We are closing this after discussion with @LukasHeinz. For this (check of upgrade notes) to happen is ensured in the user story for the closest release, and we are currently discussing the way how we work with upgrade notes generally (see https://github.com/shopsys/shopsys/issues/1327)

PetrHeinz:
@LukasHeinz I don't understand this - why not merge this? It's just an added note about merging that will help the dev performing the release. I see that we're thinking about doing things differently - but this note will have to modified anyhow... Let's just merge it until then. Why copying the note from this PR to the US for the release?

vitek-rostislav:
We do not merge this because it is in conflict with the current consensus - when I release a version, I should merge the branch into the higher branch and resolve such things. 

The note implies I should do it backward and we do not want to do it this way at all.

I did not write it precisely in the previous comment, however, there will be no need of checking previous (lower) upgrading instruction during `7.3.2` release which is the closest one now. During this release, we will merge the branch higher so the proper upgrading notes for `8.0.1` version will be ensured by this.

PR #1319, opened 2019-08-12
[shopsys] make promo codes editable on separate pages

### LINKED ISSUE #1267: Inline editation doesn't match ours needs
### What is happening

In current state promo codes are edited by inline editation. Many projects, we have been working on, require advanced promo codes. Usually we meet with requirements like:
- different types of promo codes (promo code by percentage or nominal value)
- validity of promo code
- usage limit for promo code (unlimited, one time usage, ...)
- other limitations (maximum of usage, minimal nominal value...)
- etc.
It is not very user friendly to edit all this settings by inline editation.

### Expected result

I think it would be great to change promo code inline editation to classic form editation (edit promo code on detail page) in SSFW itself. It will save a lot of time to programmers to switch from inline edition to classic form editation

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| promo codes are extended on a lot of project so inline edit might not be perfect when working with a lot of attributes
|New feature| Yes 
|Fixes issues| fixes #1267  


project-base/tests/ShopBundle/Acceptance/acceptance/VatInlineEditCest.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $inlineEditPage->assertSeeInColumn(1, 'name', 'test edited');
+    }
+
+    /**
+     * @param \Tests\ShopBundle\Test\Codeception\AcceptanceTester $me
+     * @param \Tests\ShopBundle\Acceptance\acceptance\PageObject\Admin\LoginPage $loginPage
+     * @param \Tests\ShopBundle\Acceptance\acceptance\PageObject\Admin\InlineEditPage $inlineEditPage
+     */
+    public function testVatDeleteAndCreate(AcceptanceTester $me, LoginPage $loginPage, InlineEditPage $inlineEditPage)
```
TomasLudvik: I have only question - why have you joined those two tests?

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -66,5 +66,7 @@ There you can find links to upgrade notes for other versions too.
               $debugNote = 'Order page should redirect by 302 as the cart is empty by default.';
               $config->changeDefaultRequestDataSet($debugNote)
         ```
-
+- make promo codes editable on separate page ([#1319](https://github.com/shopsys/shopsys/pull/1319))
+    - replace `PromoCodeInlineEditCest` with [`VatInlineEditCest`](https://github.com/shopsys/project-base/blob/v8.1.0/tests/ShopBundle/Acceptance/acceptance/VatInlineEditCest.php) in order to have inline edit still acceptance tested
```
TomasLudvik: Should not link be changed during release of 8.1.0? Until then it should be linked to master right?

packages/framework/src/Controller/Admin/PromoCodeController.php
```diff
@@ -59,6 +113,27 @@ public function listAction()
         ]);
     }
 
+    /**
+     * @Route("/promo-codes/list")
```
TomasLudvik: Yesterday we were discussing it and we agreed that it would be better to keep current route and decide which site to display by some parameter. Why did you chose otherwise?

packages/framework/src/Model/AdminNavigation/SideMenuBuilder.php
```diff
@@ -148,7 +148,9 @@ protected function createPricingMenu(): ItemInterface
         $menu->addChild('vat', ['route' => 'admin_vat_list', 'label' => t('VAT and rounding')]);
         $menu->addChild('free_transport_and_payment', ['route' => 'admin_transportandpayment_freetransportandpaymentlimit', 'label' => t('Free shipping and payment')]);
         $menu->addChild('currencies', ['route' => 'admin_currency_list', 'label' => t('Currencies')]);
-        $menu->addChild('promo_codes', ['route' => 'admin_promocode_list', 'label' => t('Promo codes')]);
+        $promoCodesMenu = $menu->addChild('promo_codes', ['route' => 'admin_promocode_pageeditlist', 'label' => t('Promo codes')]);
```
TomasLudvik: I think that this might be considered BC break. You kept old page with promo codes, but changed link to new implementation. In this case it is highly possible, that people will be forced to new version.

If you would have use parameter you could even decide what version of menu would be displayed.

packages/framework/src/Controller/Admin/PromoCodeController.php
DavidOstrozlik: Because if I would want to do it with GET parameter I would need to add new `Request` as a argument to the `listAction` and that would be a BC break

docs/upgrade/UPGRADE-v8.1.0-dev.md
DavidOstrozlik: I actually don't know. I can change it if you want.

project-base/tests/ShopBundle/Acceptance/acceptance/VatInlineEditCest.php
DavidOstrozlik: because if you want to delete Vat that is used you need select a new one to replace it which takes some time so the VAT is still in the table which means I need to create a new one that isn't used => can be deleted

packages/framework/src/Model/AdminNavigation/SideMenuBuilder.php
DavidOstrozlik: I couldn't use it as I described above. The old page is deprecated so it shouldn't be used anymore.

docs/upgrade/UPGRADE-v8.1.0-dev.md
TomasLudvik: it would be nice

packages/framework/src/Controller/Admin/PromoCodeController.php
TomasLudvik: Ok, I have discussed this with @PetrHeinz and we think that there should be some parameter in framework saying: useOldPromoCodes = true, this parameter would be now overwritten in project-base parameters saying useOldPromoCodes=false. So our current customers would have same behavior as before and new customers would have the new one. 

You then could use this parameter in menu to decide if to show old / new menu and in other parts (even inlineEdit in forms).

packages/framework/src/Model/AdminNavigation/SideMenuBuilder.php
TomasLudvik: I tried to describe some concept in previous comment.

henzigo:
Hello, I'm glad that you created this PR, but you missed dump new translations. @LukasHeinz

LukasHeinz:
Hi @henzigo , do not worry, final check of translations is part of release process. If the translations were missed here, it will be fixed when this feature will be released.

henzigo:
But this is bad practice for me. Imagine that you don't know this rule and you will create new feature with some translations like me and command translations-dump will dump you sume random texts and beyond you should have one PR for one feature not just half of feature in PR and then another PR  with translations

LukasHeinz:
Good point, thank you. I will discuss it with the team and we might change out DoD.

PetrHeinz:
@henzigo @LukasHeinz I've drafted up #1420 to see how it could be checked automatically...

PR #1320, opened 2019-08-12
[shopsys] simplification of Travis CI configuration

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| As demonstrated in #1302, our Travis CI always reused older Docker images, which might result in false tests pass. This PR resolves the issue by significantly simplifying the build and removing the faulty caching mechanism altogether. This comes with the price of not being able to run different jobs in parallel (meaning that when standards fail, we won't know whether tests pass).<br><br>Sadly, it makes the build run for 20 minutes instead of just 15...
|New feature| No 
|Fixes issues| closes #1302 


PetrHeinz:
The parallelization in 7521cc91 doesn't help much. It saved only 43 seconds due to `docker-compose pull` having `--parallel` as the default option and the build of `elasticsearch` image not really taking too long.

There is also a disadvantage of the log being rather confusing as the outputs of both build processes are being printed at the same time:

image

PetrHeinz:
We can simplify the Travis config even further as shown in https://github.com/shopsys/demoshop/pull/66...

PR #1323, opened 2019-08-13
[shopsys] commit migrations-lock.yml with other lock files when releasing

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to have migrations-lock.yml present in every version so you don't need to commit it every time you start a new project
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitLockFilesReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/Release/CreateAndCommitLockFilesReleaseWorker.php
@@ -8,7 +8,7 @@ use PharIo\Version\Version;
 use Shopsys\Releaser\ReleaseWorker\AbstractShopsysReleaseWorker;
 use Shopsys\Releaser\Stage;
 
-final class CreateAndCommitComposerLockAndPackageLockReleaseWorker extends AbstractShopsysReleaseWorker
+final class CreateAndCommitLockFilesReleaseWorker extends AbstractShopsysReleaseWorker
 {
     /**
      * @param \PharIo\Version\Version $version
@@ -16,7 +16,7 @@ final class CreateAndCommitComposerLockAndPackageLockReleaseWorker extends Abstr
      */
     public function getDescription(Version $version): string
     {
-        return 'Create and commit composer.lock and package-lock.json and [Manually] push it';
+        return 'Create and commit composer.lock, package-lock.json, and migrations-lock.yml and [Manually] push it';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


utils/releaser/src/ReleaseWorker/Release/CreateAndCommitLockFilesReleaseWorker.php
```diff
@@ -33,15 +33,19 @@ public function getPriority(): int
...
     {
-        $this->symfonyStyle->note('Removing vendor/, node_modules/, composer.lock, and package-lock.json');
+        $this->symfonyStyle->note('Removing vendor/, node_modules/, composer.lock, migrations-lock.yml, and package-lock.json');
         $this->processRunner->run('rm -rf project-base/vendor');
         $this->processRunner->run('rm -rf project-base/node_modules');
         $this->processRunner->run('rm -f project-base/composer.lock');
         $this->processRunner->run('rm -f project-base/package-lock.json');
+        $this->processRunner->run('rm -f project-base/migrations-lock.yml');
         $this->processRunner->run('cd project-base && composer install && npm install');
+        $this->processRunner->run('cd project-base && php phing db-rebuild');
```
LukasHeinz: Why to rebuild?

DavidOstrozlik: Because we want to drop the old database and create new, fresh one with only the migrations that you need so you can generate right `migrations-lock.yml`

PetrHeinz:
Shouldn't this PR be retargeted to `7.3` so it helps with releasing `v7.3.x` and `v8.0.x` as well? To me, it's very similar to #1244 and #1264 in this respect.

utils/releaser/src/ReleaseWorker/Release/CreateAndCommitLockFilesReleaseWorker.php
```diff
@@ -33,23 +33,29 @@ public function getPriority(): int
...
         $this->processRunner->run('rm -f project-base/composer.lock');
         $this->processRunner->run('rm -f project-base/package-lock.json');
+        $this->processRunner->run('rm -f project-base/migrations-lock.yml');
         $this->processRunner->run('cd project-base && composer install && npm install');
 
         $this->processRunner->run('git add -f project-base/composer.lock');
         $this->processRunner->run('git add -f project-base/package-lock.json');
+
+        $this->processRunner->run('rm -rf project-base/vendor');
+        $this->processRunner->run('composer install');
```
TomasLudvik: This composer install is not necessary

DavidOstrozlik: It was added to ensure that application is still working after this step (that calling composer install in project base didn't brake anything)

PR #1324, opened 2019-08-13
[coding-standards] easy-coding-standard.yml: declare compliance with PSR-12

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR bring compliance of our coding-standards package with [PSR-12](https://www.php-fig.org/psr/psr-12/). It uses the [PSR-12 set](https://github.com/Symplify/EasyCodingStandard/blob/master/config/set/psr12.yaml) in `Symplify/EasyCodingStandard` package for the automated checking.<br>Some parts of the standard are not checked by ECS yet (for example [multiline conditionals](https://www.php-fig.org/psr/psr-12/#51-if-elseif-else)) - this will probably be fixed in new versions of ECS in the future (we can even create a PR with new such fixers and help the PHP community).<br>There were no changes required in the code for compliance with the new standards.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/coding-standards/easy-coding-standard.yml
+++ b/packages/coding-standards/easy-coding-standard.yml
@@ -73,12 +73,7 @@ services:
     PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace\LogicalOperatorSpacingSniff: ~
     PhpCsFixer\Fixer\ArrayNotation\ArraySyntaxFixer:
         syntax: short
-    PhpCsFixer\Fixer\Operator\BinaryOperatorSpacesFixer:
-        default: single_space
-    PhpCsFixer\Fixer\PhpTag\BlankLineAfterOpeningTagFixer: ~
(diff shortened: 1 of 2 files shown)


henzigo:
Hello, isn't this BC-break because you need to do some steps to upgrade your application. Otherwise, your CI may fail.

PetrHeinz:
@henzigo Well, I do get your point... But I don't really see this as a true BC break - it doesn't actually complicate things for the end users. They can fix every new violation to the CS using `php phing standards-fix` (and please note there wasn't actually anything to fix in `shopsys/project-base` in this PR).

Saying this is a BC break would probably also mean that every bug-fix in the coding-standards package (assuming it would fix something differently to the previous version) would have to be released in the next `MAJOR` version instead of a patch. This point of view would simply be impractical in the long-term. For example, see #1328, which has been merged into the planned patch version `v8.0.1-dev`.

henzigo:
Hello, what with this PR?

PR #1335, opened 2019-08-16
[shopsys] fix typos in translation messages and demo data

### LINKED ISSUE #884: check and fix translation and validation messages
### What is happening

i found out that there are some translation bugs based on english language where instead of word `than` there is used `then` and also there is some strange form of mail word like capitalized it is without dash but not capitalized with dash like `Email cannot be longer then` and `Please enter e-mail`
```
First name cannot be longer then {{ limit }} characters
Email cannot be longer then {{ limit }} characters
Password cannot be longer then {{ limit }} characters

```

### Expected result

have it fixed and maybe some checker for *.po files

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The usage of "then" vs. "than" was corrected and the usage of "e-mail" was replaced by "email" in English translations. Changes were done according to our [BC promise](https://github.com/shopsys/shopsys/blob/v7.3.0/docs/contributing/backward-compatibility-promise.md#translation-messages).
|New feature| No 
|Fixes issues| resolves #884


PetrHeinz:
I just realized that some of these messages are in `shopsys/project-base`. We should just search and replace there (and create an upgrade note for this change), as modifying just the English translations have no added value there.

Also, I've noticed that 1 test scenario in `CustomerRegistrationCest` is dependent on the message `"This e-mail is already registered"` and fails now.

I'm marking this as a BC break as I don't have the time to finish it right now.

PetrHeinz:
Maybe we should split this PR to the changes to `shopsys/framework` and `shopsys/project-base`. The changes in `shopsys/framework` fix the majority of the typos (38/61) and don't suffer from the issues explained above - they'd be much easier to merge.

LukasHeinz:
This is connected to issue #884 where all this was reported.

LukasHeinz:
I recommend doing two PR - one to 7.3 and one to 9.0

PetrHeinz:
@LukasHeinz There is actually no BC break as the changes in `shopsys/project-base` are [excluded from our BC promise](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md#project-base-repository). Old version of `project-base` will work with the newer versions of `shopsys/framework` just fine - both the acceptance test and the translation message in question are in the project repository.

henzigo:
Will you rename those constants in next major release?

LukasHeinz:
@henzigo we will change the message ids in next major release. It is part of our release process.

PR #1338, opened 2019-08-16
[framework] restrict access to currency settings to superadmin

### LINKED ISSUE #791: Unexpected behavior of changing basic front-end currency on a domain
During prototyping of #745 we have found an unexpected behavior which may lead to an unpleasant surprise when encountered in production environment.

### What is happening
When changing a basic frontend currency on a domain in administration, the product prices are not recalculated or changed in any way. They remain the same, but in a different currency. This may happen without superadmin permission and without any warning or prompt.

**1. Front-end with CZK prices** (without any changes to demo data):
image

**2. Adminsitration -> Pricing -> Currencies** (without any changes to demo data):
image

**3. As an admin, I change the basic frontend currency on the first domain** (from CZK to EUR):
image

**4. As a customer, now I can see the same prices as before, but in a different currency** (without any recalculation, which means that all products are ~25&times; more expensive):
image

This issue might have gone unnoticed because automatic price type was a default value for products  before #595 was merged. Automatic product pricing did not suffer from this problem by being recalculated by the currency exchange rate.

### Expected result

The scenario above is either impossible or it's much more obvious that this is what will happen. I can think of some possible solutions:

- only a superadmin can make such a change
- this setting is moved to `domains.yml` instead of being editable in administration
- there is a *Are you really sure about this?* type of prompt (similar to removing a currency)
- all product prices for pricing groups on the edited domain are immediately recalculated using the currencies' exchange rates

As a side note, implementing the Money type as described in #745 might be a robust way to avoid similar situations in the future.

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When changing a basic frontend currency on a domain in administration, the product prices are not recalculated or changed in any way. They remain the same, but in a different currency. This may happen without any warning or prompt. Because of this, I suggest restricting the access to currency setting to superadmin only.
|New feature| Yes 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No <br> ~~(for this reason, I've used `@Security` annotation instead of a record in `security.yml`, which is used for restricting access to other pages)~~<br>Record in security.yml is necessaary as it is project-base thing same as upgrading smoke tests. The app will not break when upgrade (except missing link to affected page). 
|Fixes issues| resolves #791 

### DIFF
--- a/packages/framework/src/Controller/Admin/CurrencyController.php
+++ b/packages/framework/src/Controller/Admin/CurrencyController.php
@@ -2,6 +2,7 @@
 
 namespace Shopsys\FrameworkBundle\Controller\Admin;
 
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Shopsys\FrameworkBundle\Component\ConfirmDelete\ConfirmDeleteResponseFactory;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Router\Security\Annotation\CsrfProtection;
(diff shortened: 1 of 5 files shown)


pk16011990:
hi, what is missing for accepting?

PR #1340, opened 2019-08-20
[framework] replace constant from other classes for value

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| What I know it's not good to use constants which can be changed in time and migrations must not adjust over time, right? So I changed these constants to their value.
|New feature| No 
|Fixes issues| ... 

P.S.: I didn't change classes with which I was not sure:

- [here](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Migrations/Version20190121094400.php#L107) is used `Doctrine\DBAL\Connection\Connection::PARAM_INT_ARRAY` constant and I don't know that is ok
- [I am not sure about using this constant](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Migrations/MultidomainMigrationTrait.php#L20)

P.P.S.: after consultation with @pk16011990 I heard PHP class constant is OK because this constant/classes should be changed only in major version (e.c. from PHP 7 to PHP 8). If it's true then I am ok with PDO constant and other PHP constants.

### DIFF
--- a/packages/framework/src/Migrations/MultidomainMigrationTrait.php
+++ b/packages/framework/src/Migrations/MultidomainMigrationTrait.php
@@ -3,7 +3,6 @@
 namespace Shopsys\FrameworkBundle\Migrations;
 
 use PDO;
-use Shopsys\FrameworkBundle\Component\Setting\Setting;
 
 /**
  * This trait can be used in classes
@@ -17,7 +16,7 @@ trait MultidomainMigrationTrait
     protected function getAllDomainIds()
     {
         return $this
-            ->sql('SELECT domain_id FROM setting_values WHERE name = :baseUrl', ['baseUrl' => Setting::BASE_URL])
+            ->sql('SELECT domain_id FROM setting_values WHERE name = :baseUrl', ['baseUrl' => 'baseUrl'])
             ->fetchAll(PDO::FETCH_COLUMN, 'domain_id');
     }
 
--- a/packages/framework/src/Migrations/Version20160129122637.php
+++ b/packages/framework/src/Migrations/Version20160129122637.php
@@ -3,7 +3,6 @@
 namespace Shopsys\FrameworkBundle\Migrations;
 
 use Doctrine\DBAL\Schema\Schema;
-use Shopsys\FrameworkBundle\Component\Setting\Setting;
 use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
 
 class Version20160129122637 extends AbstractMigration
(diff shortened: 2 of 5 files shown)


vitek-rostislav:
Hi @sspooky13, thanks for the PR, I believe we are able to process it in the following sprint (ie. within 3 weeks :wink:)

s3tezsky:
Hey @sspooky13 could you please replase constant in the trait as well?

https://github.com/shopsys/shopsys/blob/05768ca61f2ee447ac954548885064d923a916f2/packages/framework/src/Migrations/MultidomainMigrationTrait.php#L20

sspooky13:
> 
> 
> Hey @sspooky13 could you please replase constant in the trait as well?
> 
> https://github.com/shopsys/shopsys/blob/05768ca61f2ee447ac954548885064d923a916f2/packages/framework/src/Migrations/MultidomainMigrationTrait.php#L20

Hello @s3tezsky,
Fixed. Next time you can fix these little things by yourself. It's an ineffective wait to fix these little issues from contributors 🙂 but if you don't hurry then it's alright 😁

s3tezsky:
I got that ineffective point, but these are standard process. Even this is really small change with low severity I did not want to change those things and then review it by myself.

Could you just please rebase it to actual master? It sould fix the build. Then it would be OK from me 👍

sspooky13:
Rebased.
Hmm, it's new to me. I created several PR and no one wanted to change these things (rebase, change 1 row, etc.) from me. With amend commits were no problem too. Were rules have been changed? 🤔

s3tezsky:
Thank you for that!

[Rules](https://github.com/shopsys/shopsys/blob/master/docs/contributing/guidelines-for-pull-request.md#1-create-pull-request) are the same for a long time and looks like nothing important has not changed.

I am new to SSFW and these are one of firsts community PRs which may resulted in different interpretation of the guideline.

I may explain you my points:
- fix one line - As me as a reviewer do not want to make any changes because it will results in approving my own commits (even that small ones)
- avoid ammending - Ammending is fine for squashing fixup commits but I seem it a bit unappropriete doing it when new changes appears after code review. There is commit history lost and than the reviewer need to review whole PR again instead last changes only.
- rebasing - As you should start working on checkouted actual master you should keep the branch as actual as possible to avoid any conflicts. If it is not done I seem a bit risky approving those changes because of breaking some code with unwanted wrong resolved conficts which might appear after the approve.

Sorry for these inconvinience I will discuss about proper processing with colleagues (and try to get better next time). Anyway thanks again for contrubuting an hope I have not distracted you from future contribution.

Cheers!

Issue #152, opened 2018-05-28
Dates on the frontend are not localized


jakubhavlak:
Problem is in DatePickerType, which uses a constant format (Czech localization: `dd.MM.yyyy`), both in admin and front-end:

![image](https://user-images.githubusercontent.com/10008612/54756227-121e9200-4be8-11e9-9a6a-e014bd9330e5.png)

A localized date format should be used.

boris-brtan:
is this already implemented?

henzigo:
Stalebot open it!

PetrHeinz:
@henzigo Can you please provide some more info, ideally some steps to reproduce?

I've taken a look at list of orders on front-end on both English and Czech domain and the date localization seems fine ("2019-03-15" for English / "21. 3. 2019" for Czech):

![image](https://user-images.githubusercontent.com/10008612/54748606-661e7c00-4bd2-11e9-8434-2f2ade3cd028.png)

![image](https://user-images.githubusercontent.com/10008612/54748616-6cacf380-4bd2-11e9-92c6-fcc5639b9425.png)

Tested on current `master` installed via Docker. I don't suppose anything has changed in this behavior lately.

henzigo:
Ye sorry. if you work only with output its fine. But framework is not prepared for working with localized dates in forms and javascript. Problem is when you want to use input with date or use datepicker for example.

PetrHeinz:
Thanks for clarification :+1:

VitMickal:
This issue is going to solved within User Story which has been raised - US-5162

henzigo:
open it

henzigo:
Any new information?

sslt:
This issue has been resolved in #1343 if I am not mistaken. Correct me if I am wrong @henzigo

pk16011990:
nope, #1343 is about timezones, not about localization in form fields.

This issue has few problems (in version SSFW 6.x):
* We want 4-digits in CS years. Library INTL needs variant MEDIUM for 4-digits year. However MEDIUM variant in Hungarian locale means month formated as 3-letters word (eg. `oct`). Hungarian needs SHORT variant (month as number in hungarian).
* Format definition is different for JS datepicker and for parsing in `DatePickerType` in PHP.
* If we want same formatting for all places, problem is: Few languages (eg. czech) put space after dot ("1. 1. 2000"), It causes problem in JS Datepicker, because it requires space too in manual user input.
[...]

PR #1343, opened 2019-08-22
[shopsys] Added support to display dates in different timezone

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| All date-time values are internally in UTC and this PR allows to present date-time values in different timezone.
|New feature| Yes 
|Fixes issues| ... 

new class `DisplayTimeZoneProvider` implementing `DisplayTimeZoneProviderInterface` is used to provide desired timezone.
All dates should be presented to the user via Twig template and should use one of filters (`formatDate`, `formatTime`, `formatDateTime`)

Current implementation `DisplayTimeZoneProvider` use parameter `shopsys.display_timezone` to select proper time zone.
For new installations is used `Europe/Prague` as default timezone

- after running database migrations, products will use a DateTime type for columns for "Selling start date" (selling_from) and "Selling end date" (selling_to)
    - not considered a BC break
    - the tests will not fail because on the testing database are run DB migrations before the testing starts (when it's run as `php phing tests`)
    - even the `php phing checks` target will fail as the database schema is not satisfying ORM (Phing target `db-check-schema`) the application still behaves the same
    - DateTime is used to properly handle timezones and should not be used to set exact time to start/end selling products as the recalculations run at midnight by default
    - columns will have set the 00:00:00 as time
    - please check [`Version20190823110846`](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Migrations/Version20190823110846.php)


packages/framework/src/Migrations/Version20190821060201.php
```diff
@@ -0,0 +1,30 @@
...
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20190821060201 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE products ADD selling_from_date_time TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL');
```
TomasLudvik: Would not that be easier to update type of current columns? I was looking at the entity and there are mixed names "sellingFrom" and "sellingDateTimeFrom" or something like that and I think you are also little confused how to use that right way.

docs/introduction/using-form-types.md
```diff
@@ -108,6 +109,16 @@ DatePickerType has 2 constants that can be used:
...
 
+### [DateTimeType](../../packages/framework/src/Form/DateTimeType.php)
+Displays a text field that allows to set date and time.  
+Value is internally converted from [display time zone](./working-with-date-time-values.md) to UTC.
+
+#### format
+Defaults to `DateTimeType::FORMAT_PHP`.  
+Defines in what format should be the date shown.
+DatePickerType has constant that can be used:
+- FORMAT_PHP = 'dd.MM.yyyy HH:mm:ss'
```
TomasLudvik: Oh here it is... should not this be stored in parameters so implementators can change it easily?

packages/framework/src/Resources/config/parameters_common.yml
```diff
@@ -8,5 +8,8 @@ parameters:
     # Administration locale. Must be registered in domains.yml
     shopsys.admin_locale: en
 
+    # All time values in administration are displayed in this timezone (default is PHP timezone)
+    shopsys.display_timezone: ~
```
TomasLudvik: I would prefer if it was set here to prague

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -79,4 +85,20 @@ There you can find links to upgrade notes for other versions too.
         +       arguments:
         +           $useInlineEditation: false
         ```
+- add support to display date-time values in different timezone ([#1343](https://github.com/shopsys/shopsys/pull/1343))
+    - you can read more about how to [work with display timezone in documentation](/docs/introduction/working-with-date-time-values.md)
+    - you can delete test `tests/ShopBundle/Functional/Twig/DateTimeFormatterExtensionTest.php` as it was moved to the FrameworkBundle, if you don't test any your specific use-case
+    - class `CustomDateTimeFormatterFactory` is deprecated and should not be used anymore
+        - if you have extended this factory to alter configuration of `DateTimeFormatPatternRepository`, extend new `DateTimeFormatPatternRepositoryFactory` instead (as `DateTimeFormatPatternRepository` configuration was moved to this class from the factory)
+        - custom `DateTimeFormatter` have to be created with the DIC directly if needed
```
TomasLudvik: has to be created

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -79,4 +85,20 @@ There you can find links to upgrade notes for other versions too.
         +       arguments:
         +           $useInlineEditation: false
         ```
+- add support to display date-time values in different timezone ([#1343](https://github.com/shopsys/shopsys/pull/1343))
+    - you can read more about how to [work with display timezone in documentation](/docs/introduction/working-with-date-time-values.md)
+    - you can delete test `tests/ShopBundle/Functional/Twig/DateTimeFormatterExtensionTest.php` as it was moved to the FrameworkBundle, if you don't test any your specific use-case
```
TomasLudvik: ~don't~ do not

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -79,4 +85,20 @@ There you can find links to upgrade notes for other versions too.
...
+    - class `CustomDateTimeFormatterFactory` is deprecated and should not be used anymore
+        - if you have extended this factory to alter configuration of `DateTimeFormatPatternRepository`, extend new `DateTimeFormatPatternRepositoryFactory` instead (as `DateTimeFormatPatternRepository` configuration was moved to this class from the factory)
+        - custom `DateTimeFormatter` have to be created with the DIC directly if needed
+            ```yaml
+            Shopsys\FrameworkBundle\Component\Localization\DateTimeFormatterInterface:
+                class: 'Shopsys\ShopBundle\Component\Localization\DateTimeFormatter'
+            ```
+    - class `DateTimeFormatter` no longer supports extending via `shopsys.entity_extension.map` as it's not entity
```
TomasLudvik: ~it's~ it is

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -79,4 +85,20 @@ There you can find links to upgrade notes for other versions too.
...
+        -if you have extend this class with `shopsys.entity_extension.map`, register your class for `DateTimeFormatterInterface` in `services.yml` instead
```
TomasLudvik: extended

packages/framework/src/Migrations/Version20190821060201.php
TomasLudvik: Why are you updating this? I know that selling denied is better in exact time, but it is not part of this US, right?

packages/framework/src/Resources/config/parameters_common.yml
```diff
@@ -8,5 +8,8 @@ parameters:
     # Administration locale. Must be registered in domains.yml
     shopsys.admin_locale: en
 
+    # All time values in administration are displayed in this timezone (default is PHP timezone)
```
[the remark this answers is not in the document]
grossmannmartin: No, it's not :) Fixed in a128df8ecf5943a29dc46f41bb0d50b779d437c7

packages/framework/tests/Unit/Component/Localization/DateTimeFormatterTest.php
```diff
@@ -0,0 +1,55 @@
...
+            // Mountain Standard Time (UTC -7)
+            ['inputDateTime' => new DateTime('2019-08-21T06:52:47+00:00'), 'displayTimeZone' => 'America/Phoenix', 'result' => 'Aug 20, 2019, 11:52:47 PM'],
+        ];
+    }
+
+    /**
+     * @dataProvider formatDateTimeDataProvider
+     * @param \DateTime $inputDateTime
+     * @param string $dateTimeZone
+     * @param mixed $result
```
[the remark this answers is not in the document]
grossmannmartin: String! Fixed in 109b2419fb7710793d6034545e67aad5c34cb1df

packages/framework/src/Form/DateTimeType.php
```diff
@@ -0,0 +1,48 @@
...
+namespace Shopsys\FrameworkBundle\Form;
+
+use Shopsys\FrameworkBundle\Component\Localization\DisplayTimeZoneProviderInterface;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\DateTimeType as SymfonyDateTimeType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class DateTimeType extends AbstractType
+{
+    protected const FORMAT_PHP = 'dd.MM.yyyy HH:mm:ss';
```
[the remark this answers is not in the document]
grossmannmartin: I followed the trend set in `DatePickerType` to easier usage as it's similar to something people already know.

packages/framework/src/Migrations/Version20190821060201.php
grossmannmartin: Change the type of column is considered a BC break – https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md#database-migrations 

You're right the naming is confusing, but the idea is: just the name of the protected property is changed and the rest of the application still works as before (still use sellingFrom)

And why I'm updating this. Consider someone in Phoenix on 21.02.2019 20:00 (UTC-7) sets SellingFrom to another day (22.02.2019).

What date should be saved in the database that is in UTC? 

Let's say I save exactly what user enters – 22.02.2019. 
Server time is at the moment of change 22.02.2019 03:00 and running recalculation make the product visible immediately. Not good.

Let's say I convert Date(Time) to UTC before saving.
Convert the DateTime returned from Symfony Form `2019-02-22 00:00:00.0 America/Phoenix (-07:00)` to UTC. 
That is `2019-02-22 07:00:00.0 UTC (+00:00)` but into the database i still save only 22-02-2019. And we're at the exactly same spot as before.

Having the time saved solves this problem as I can now save and compare with exact time moment.

But the time currently should not be used to set product visible at the exact time of the day (ie. at 13:25) because the recalculations are not executed so often.

docs/introduction/using-form-types.md
grossmannmartin: I consider it outside the scope because then I would need to change DatePickerType too (would I?).

packages/framework/src/Resources/config/parameters_common.yml
grossmannmartin: I cannot set it here to Prague, because that could change the times for someone who upgrade to the newest version (My server is set to 'America/Phoenix' and I count with that in application). Setting it here means their time shifts on the frontend.

I set it in `ShopBundle` `parameters_common.yml` file only on purpose so new installations are in Europe/Prague (acceptance criteria), but upgrade does not change a thing because default is server timezone.

docs/upgrade/UPGRADE-v8.1.0-dev.md
grossmannmartin: fixed in ad2dcfae1808de200ae563f458e1aa9fc60420a2

docs/upgrade/UPGRADE-v8.1.0-dev.md
grossmannmartin: fixed in ad2dcfae1808de200ae563f458e1aa9fc60420a2

docs/upgrade/UPGRADE-v8.1.0-dev.md
grossmannmartin: fixed in ad2dcfae1808de200ae563f458e1aa9fc60420a2

docs/upgrade/UPGRADE-v8.1.0-dev.md
grossmannmartin: fixed in ad2dcfae1808de200ae563f458e1aa9fc60420a2

packages/framework/src/Migrations/Version20190821060201.php
TomasLudvik: Ok good point, but I consider changing date to date_time as exception from our BC break same as we allow changing array to iterable. It will still behave same way.

docs/introduction/using-form-types.md
TomasLudvik: Hmmm maybe yes. Can you at least create issue for that?

packages/framework/src/Resources/config/parameters_common.yml
TomasLudvik: Ok good point.

TomasLudvik: You forgot to update parameters_common in project-base

grossmannmartin: it's in different commit – 9cadeb4442c9a351daa297cd99a192ed19aa0e62

docs/introduction/using-form-types.md
grossmannmartin: It actually exists - #152 (Dates on the frontend are not localized)
And we already have the User Story for that too (internal number 6364)

packages/framework/src/Migrations/Version20190821060201.php
grossmannmartin: Hmm... it makes sense... 
the tests will not fail because on the testing database are run DB migrations before the testing starts (when it's run with Phing).

The `./phing checks` target will fail as the database schema is not satisfying ORM (Phing target `db-check-schema`)

But apart from that everything seems to be working.

grossmannmartin: fixed in 610c3c3df026edf41e47854a89938bb9ecacc0fb and 60713fae9f68e33d67c5fd8867d71f1b8adf680a

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -40,11 +40,13 @@ There you can find links to upgrade notes for other versions too.
         ```
 
 ### Database migrations
-- after running database migrations, products will have new columns for "Selling start date" (selling_from_date_time) and "Selling end date" (selling_to_date_time) as DateTime fields ([#1343](https://github.com/shopsys/shopsys/pull/1343))
+- after running database migrations, products will use a DateTime type for columns for "Selling start date" (selling_from) and "Selling end date" (selling_to) ([#1343](https://github.com/shopsys/shopsys/pull/1343))
```
TomasLudvik: UPGRADE notes should only include instructions what users are supposed to do

grossmannmartin: fixed in b5bff4734bd5938a3e57b9ccc7f8111778a21fa5

grossmannmartin:
Rebased to current master before tests/bv

PR #1345, opened 2019-08-22
[framework] creating OrderItem type product from OrderItemData is now done via OrderItemFactory

### LINKED ISSUE #1266: Add new function to be able to create OrderItem from OrderItemData
### What is happening

If I want to add a new field to `OrderItem`, e.g. product EAN, I override method `OrderItemFactoryInterface::createProduct`, and set EAN from the product, if exists.  
But If I want to edit that field from the administration, I have to override `edit`method.

### Expected result

I would like to have a separate method, called `OrderItemFactoryInterface::createProductFromOrderItemData`, which accepts parameters `Order`, `OrderItemData`. 
If I want to edit EAN from the administration I just add new field to `OrderItemData` and override `createProductFromOrderItemData`, I don't need to override whole `edit` function.

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Extending OrderItem cannot be extended in easy way. There are many uncomfortable changes required. We can partially bypass those with adding a new factory method.
|New feature| Yes
|Fixes issues| closes #1266 

With this PR comes up also branch https://github.com/shopsys/shopsys/tree/ds-create-order-item-from-data-test-code which can provide basic code for testing it or to demostrate extending OrderItem class

### DIFF
--- a/docs/upgrade/UPGRADE-v8.1.0-dev.md
+++ b/docs/upgrade/UPGRADE-v8.1.0-dev.md
@@ -79,4 +79,8 @@ There you can find links to upgrade notes for other versions too.
         +       arguments:
         +           $useInlineEditation: false
         ```
+- introduced `\Shopsys\FrameworkBundle\Model\Order\Item\OrderItemFactory::createProductFromOrderItemData()` which provides more flexibility when `\Shopsys\FrameworkBundle\Model\Order\Item\OrderItem` is extended
+	- simplifies SSFW usage in `\Shopsys\FrameworkBundle\Model\Order\OrderFacade::refreshOrderItemsWithoutTransportAndPayment()` where the logic was moved into new method which can be well overriden without needs to duplicate `OrderFacade`'s code
+	- `\Shopsys\FrameworkBundle\Model\Order\Item\OrderItemFactoryInterface` is deprecated and method `createProductFromOrderItemData()` will be added in next major version
+
 [shopsys/framework]: https://github.com/shopsys/framework
--- a/packages/framework/src/Model/Order/Item/OrderItemFactory.php
+++ b/packages/framework/src/Model/Order/Item/OrderItemFactory.php
@@ -63,6 +63,31 @@ class OrderItemFactory implements OrderItemFactoryInterface
         return $orderProduct;
     }
 
+    public function createProductFromOrderItemData(Order $order, OrderItemData $orderItemData): OrderItem
+    {
+        $classData = $this->entityNameResolver->resolve(OrderItem::class);
+
+        $orderItem = new $classData(
+            $order,
+            $orderItemData->name,
+            new Price(
+                $orderItemData->priceWithoutVat,
+                $orderItemData->priceWithVat
+            ),
+            $orderItemData->vatPercent,
+            $orderItemData->quantity,
+            OrderItem::TYPE_PRODUCT,
+            $orderItemData->unitName,
+            $orderItemData->catnum
+        );
+        if (!$orderItemData->usePriceCalculation) {
+            $orderItem->setTotalPrice(
+                new Price($orderItemData->totalPriceWithoutVat, $orderItemData->totalPriceWithVat)
+            );
+        }
+        return $orderItem;
+    }
+
     /**
      * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
      * @param string $name
--- a/packages/framework/src/Model/Order/Item/OrderItemFactory.php
+++ b/packages/framework/src/Model/Order/Item/OrderItemFactory.php
@@ -63,6 +63,11 @@ class OrderItemFactory implements OrderItemFactoryInterface
         return $orderProduct;
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData $orderItemData
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\OrderItem
+     */
     public function createProductFromOrderItemData(Order $order, OrderItemData $orderItemData): OrderItem
     {
         $classData = $this->entityNameResolver->resolve(OrderItem::class);
--- a/packages/framework/src/Model/Order/OrderFacade.php
+++ b/packages/framework/src/Model/Order/OrderFacade.php
@@ -661,20 +661,32 @@ class OrderFacade
         foreach ($orderData->getNewItemsWithoutTransportAndPayment() as $newOrderItemData) {
             $this->calculateOrderItemDataPrices($newOrderItemData);
 
-            $newOrderItem = $this->orderItemFactory->createProduct(
-                $order,
-                $newOrderItemData->name,
-                new Price(
-                    $newOrderItemData->priceWithoutVat,
-                    $newOrderItemData->priceWithVat
-                ),
-                $newOrderItemData->vatPercent,
-                $newOrderItemData->quantity,
-                $newOrderItemData->unitName,
-                $newOrderItemData->catnum
-            );
-            if (!$newOrderItemData->usePriceCalculation) {
-                $newOrderItem->setTotalPrice(new Price($newOrderItemData->totalPriceWithoutVat, $newOrderItemData->totalPriceWithVat));
+            if (method_exists($this->orderItemFactory, 'createProductFromOrderItemData')) {
+                $this->orderItemFactory->createProductFromOrderItemData($order, $newOrderItemData);
+            } else {
+                $message = sprintf(
+                    'Adding new OrderItem via "%s" is deprecated since SSFW 8.1, implement "%s" instead.'
+                    . 'Interface "%s" will be changed in SSFW 9.',
+                    get_class($this->orderItemFactory) . '::createProduct()',
+                    get_class($this->orderItemFactory) . '::createProductFromOrderItemData()',
+                    OrderItemFactoryInterface::class
+                );
+                @trigger_error($message, E_USER_DEPRECATED);
+                $newOrderItem = $this->orderItemFactory->createProduct(
+                    $order,
+                    $newOrderItemData->name,
+                    new Price(
+                        $newOrderItemData->priceWithoutVat,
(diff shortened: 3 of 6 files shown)


packages/framework/src/Model/Order/Item/OrderItemFactory.php
```diff
@@ -63,6 +63,36 @@ public function createProduct(
         return $orderProduct;
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData $orderItemData
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\OrderItem
+     */
+    public function createProductFromOrderItemData(Order $order, OrderItemData $orderItemData): OrderItem
```
TomasLudvik: Just detail - when method name includes "from order item data" I would expect it to be first argument

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -661,20 +661,32 @@ protected function refreshOrderItemsWithoutTransportAndPayment(Order $order, Ord
...
-                $newOrderItemData->unitName,
-                $newOrderItemData->catnum
-            );
-            if (!$newOrderItemData->usePriceCalculation) {
-                $newOrderItem->setTotalPrice(new Price($newOrderItemData->totalPriceWithoutVat, $newOrderItemData->totalPriceWithVat));
+            if (method_exists($this->orderItemFactory, 'createProductFromOrderItemData')) {
+                $this->orderItemFactory->createProductFromOrderItemData($order, $newOrderItemData);
+            } else {
+                $message = sprintf(
+                    'Adding new OrderItem via "%s" is deprecated since SSFW 8.1, implement "%s" instead.'
```
TomasLudvik: We do not use SSFW as official shortcut, you can replace "SSWF" with "version". And you can update that Interface will be updated in next major release.

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -661,20 +661,32 @@ protected function refreshOrderItemsWithoutTransportAndPayment(Order $order, Ord
...
+                    . 'Interface "%s" will be changed in SSFW 9.',
+                    get_class($this->orderItemFactory) . '::createProduct()',
```
TomasLudvik: get_class returns FQCN, I would use only 
```suggestion
                    'OrderItemFactory::createProduct()',
```

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -661,20 +661,32 @@ protected function refreshOrderItemsWithoutTransportAndPayment(Order $order, Ord
...
+                    get_class($this->orderItemFactory) . '::createProductFromOrderItemData()',
```
TomasLudvik:
```suggestion
                    'OrderItemFactory::createProductFromOrderItemData()',
```

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -661,20 +661,32 @@ protected function refreshOrderItemsWithoutTransportAndPayment(Order $order, Ord
...
+                    OrderItemFactoryInterface::class
```
TomasLudvik:
```suggestion
                    'OrderItemFactoryInterface'
```

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -661,20 +661,32 @@ protected function refreshOrderItemsWithoutTransportAndPayment(Order $order, Ord
...
+                );
+                @trigger_error($message, E_USER_DEPRECATED);
+                $newOrderItem = $this->orderItemFactory->createProduct(
```
TomasLudvik: new line before this one please

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -661,20 +661,32 @@ protected function refreshOrderItemsWithoutTransportAndPayment(Order $order, Ord
...
+                    new Price(
+                        $newOrderItemData->priceWithoutVat,
+                        $newOrderItemData->priceWithVat
+                    ),
+                    $newOrderItemData->vatPercent,
+                    $newOrderItemData->quantity,
+                    $newOrderItemData->unitName,
+                    $newOrderItemData->catnum
+                );
+                if (!$newOrderItemData->usePriceCalculation) {
```
TomasLudvik: New line before this one please

packages/framework/src/Model/Order/Item/OrderItemFactory.php
TomasLudvik:
```suggestion
    public function createProductOrderItemFromOrderItemData(Order $order, OrderItemData $orderItemData): OrderItem
```

packages/framework/src/Model/Order/Item/OrderItemFactory.php
```diff
@@ -63,6 +63,36 @@ public function createProduct(
...
+    {
+        $classData = $this->entityNameResolver->resolve(OrderItem::class);
```
TomasLudvik: I would probably check if OrderItemData has transport and payment set to null

packages/framework/src/Resources/config/services.yml
```diff
@@ -720,7 +720,7 @@ services:
         alias: Shopsys\FrameworkBundle\Model\Order\OrderNumberSequenceFactory
 
     Shopsys\FrameworkBundle\Model\Order\Item\OrderItemFactoryInterface:
-        alias: Shopsys\FrameworkBundle\Model\Order\Item\OrderItemFactory
+        alias: Shopsys\ShopBundle\Model\Order\Item\OrderItemFactory
```
TomasLudvik: We do not use ShopBundle classes in framework, because framework package does not depend on project-base

project-base/src/Shopsys/ShopBundle/Model/Order/Item/OrderItemFactory.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
TomasLudvik: Why do you think that it is necessary to create this class?

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -79,4 +79,8 @@ There you can find links to upgrade notes for other versions too.
         +       arguments:
         +           $useInlineEditation: false
         ```
+- introduced `\Shopsys\FrameworkBundle\Model\Order\Item\OrderItemFactory::createProductFromOrderItemData()` which provides more flexibility when `\Shopsys\FrameworkBundle\Model\Order\Item\OrderItem` is extended
```
TomasLudvik: Upgrade notes should be directives of what should user do when upgrading.

packages/framework/src/Model/Order/OrderFacade.php
TomasLudvik: I would update it like this:
```suggestion
                    'Adding new OrderItem via "%s" is deprecated since version 8.1, use "%s" method instead.'
```

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -661,20 +661,32 @@ protected function refreshOrderItemsWithoutTransportAndPayment(Order $order, Ord
...
-                $newOrderItemData->catnum
-            );
-            if (!$newOrderItemData->usePriceCalculation) {
-                $newOrderItem->setTotalPrice(new Price($newOrderItemData->totalPriceWithoutVat, $newOrderItemData->totalPriceWithVat));
+            if (method_exists($this->orderItemFactory, 'createProductFromOrderItemData')) {
+                $this->orderItemFactory->createProductFromOrderItemData($order, $newOrderItemData);
+            } else {
+                $message = sprintf(
+                    'Adding new OrderItem via "%s" is deprecated since SSFW 8.1, implement "%s" instead.'
+                    . 'Interface "%s" will be changed in SSFW 9.',
```
TomasLudvik:
```suggestion
                    . 'Interface "%s" will be changed in next major release.',
```

docs/upgrade/UPGRADE-v8.1.0-dev.md
TomasLudvik: I would say something like:
- if you are extending `OrderFacade::refreshOrderItemsWithoutTransportAndPayment()` method in your project, you need to update your usage according to ([#1345](https://github.com/shopsys/shopsys/pull/1345/)) where was logic creating `OrderItem` moved to `OrderItemFactory:: createProductOrderItemFromOrderItemData()` method

docs/upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -79,4 +79,8 @@ There you can find links to upgrade notes for other versions too.
...
+	- simplifies SSFW usage in `\Shopsys\FrameworkBundle\Model\Order\OrderFacade::refreshOrderItemsWithoutTransportAndPayment()` where the logic was moved into new method which can be well overriden without needs to duplicate `OrderFacade`'s code
+	- `\Shopsys\FrameworkBundle\Model\Order\Item\OrderItemFactoryInterface` is deprecated and method `createProductFromOrderItemData()` will be added in next major version
```
TomasLudvik: That interface is not deprecated

packages/framework/src/Model/Order/Item/OrderItemFactory.php
```diff
@@ -63,6 +63,36 @@ public function createProduct(
...
+
+        $orderItem = new $classData(
```
TomasLudvik: This should be done via createProduct method

LukasHeinz:
OrderItem was previously (and maybe still is) a problem for some project implementation. Please look for feedback from implementation teams before proceeding with this one.

DavidOstrozlik:
This problem should be resolved more globally by refactoring the way that OrderItem and OrderItemData are created which will be obviously a BC break. I created a user story for this so we can solve this problem. I'm closing this PR as it doesn't make sense anymore.

Issue #1337, opened 2019-08-16
docker logs are not limited in size


PetrHeinz:
### What is happening
During finding more information for #1326 we found out that Docker logs can get pretty large and are not limited in size by default.

For that reason we could introduce some logging options to the `docker-compose.yml` to configure log rotation:

```diff
  webserver:
      image: nginx:1.13-alpine
      container_name: shopsys-framework-webserver
      depends_on:
          - php-fpm
      volumes:
          - .:/var/www/html
          - ./docker/nginx/nginx.conf:/etc/nginx/conf.d/default.conf
      ports:
          - "8000:8080"
+     logging:
+          options:
+              max-size: "10m"
+              max-file: "10"
```

This example config will keep the logs of the `webserver` container at the maximum size of 100 MiB.

See [Compose file reference](https://docs.docker.com/compose/compose-file/#logging) and [Docker logging drivers configuration docs](https://docs.docker.com/config/containers/logging/configure/) for more details.

### Expected result

Logs are rotated by default, at least in our `docker-compose.prod.yml.dist`, or there is a different mechanism to keep the log files from bloating. This behavior is documented (why it's done and how to change the limits).

PR #1362, opened 2019-08-30
[shopsys] nginx.conf: remove access_log directive and use default settings

### LINKED ISSUE #1326: wrong handling of nginx access logs in webserver container
### What is happening

There is a directive `access_log /var/log/nginx/shopsys-framework.access.log` in our [nginx.conf](https://github.com/shopsys/shopsys/blob/v7.3.1/project-base/docker/nginx/nginx.conf) configuration. This config is used for running nginx in Docker and in containerized applications we rather use `stdout` for logs (see [12-factor app](https://12factor.net/logs)).

This means there will be a log file in the `webserver` with duplicate contents with logs maintained by Docker itself (as it reads and stores `stdout` from all containers). This logfile can get pretty big pretty soon...

### Expected result

The nginx in container should be configured to not save logs into a file. Let Docker (or Kubernetes) handle the logging.

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Configuring the access log filepath is not neccessary, `/var/log/nginx/access.log` is used by default.<br>In `nginx` Docker image, the `/var/log/nginx/access.log` is a symlink to `/dev/stdout` which means that using the path is in accordance with [12-factor app recommendation](https://12factor.net/logs).<br><br>There is still an open related issue #1337
|New feature| Yes (access logs are now on standard output which can be seen as a feature, but if we agree it's a bugfix we should retarget to `7.3` instead) 
|Fixes issues| resolves #1326 

### DIFF
--- a/docker/nginx/nginx.conf
+++ b/docker/nginx/nginx.conf
@@ -4,7 +4,6 @@ upstream php-upstream {
 
 server {
     listen 8080;
-    access_log /var/log/nginx/shopsys-framework.access.log;
     root /var/www/html/project-base/web;
     server_tokens off;
     client_max_body_size 32M;
--- a/docs/upgrade/UPGRADE-v7.3.2-dev.md
+++ b/docs/upgrade/UPGRADE-v7.3.2-dev.md
@@ -48,5 +48,15 @@ There you can find links to upgrade notes for other versions too.
     +      prefix: Shopsys\ShopBundle\Component
     +      is_bundle: false
     ```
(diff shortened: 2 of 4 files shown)


TomasLudvik:
Tested and is ok

DavidOstrozlik:
> I am OK with PR itself but I do not understand the info regarding "New Feature" in PR desctription.

It's just that it's a new feature because we didn't have it yet but it could be seen as a bug fix. The only question about this is if we want to have this in the next minor or as a bug fix in 7.3

PR #1364, opened 2019-09-03
ProductCodeController now autowires PromoCodeDataFactoryInterface instead of PromoCodeDataFactory

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| To extend promo codes you should not need to extend PromoCodeController just because of wrong argument type hint.
|New feature| No 
|Fixes issues| ... 


LukasHeinz:
@s3tezsky if it is BC break, it should be targeted to 9.0. Please change targeted branch or PR desc.

s3tezsky:
I've changed BC breake to `NO`. Actually in v8.0 it won't break anything since it is depending on https://github.com/shopsys/shopsys/pull/1319 which hasn't been released yet.

PR #1368, opened 2019-09-03
[framework] all phing targets that should not be run in production now ask for permission if in production environment

### LINKED ISSUE #1236: Do not allow destructive phing actions in production without warning
### What is happening
When developer has open more consoles and one of them is in production environemnt, there is chance for running a destructive phing command in production. I do not wish anyone that feeling :smile: 

### Expected result
All destructive phing targets cannot be run without a obstacle. 
We used somethinkg like:
```xml
    <target name='check-dev-enabled'>
        <if>
            <and>
                <isset property="enable-dev-commands" />
                <equals arg1="${enable-dev-commands}" arg2="false" />
            </and>
            <then>
                <fail message = "You can't use this command. If you want to use it you have to edit build.local.properties file."/>
            </then>
        </if>
    </target>
```
and set it like dependency for actions: `build-new`, `build-demo`, `db-create` etc...

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Many phing targets should not be run in production environment unless really intended.
|New feature| Yes 
|Fixes issues| closes #1236 

### DIFF
--- a/docs/upgrade/UPGRADE-v9.0.0-dev.md
+++ b/docs/upgrade/UPGRADE-v9.0.0-dev.md
@@ -7,6 +7,15 @@ There you can find links to upgrade notes for other versions too.
 
 ## [shopsys/framework]
 
+### Infrastructure
+- update your `kubernetes/deployments/webserver-php-fpm.yml` file: ([#1368](https://github.com/shopsys/shopsys/pull/1368))
+    ```diff
+    -   command: ["sh", "-c", "cd /var/www/html && ./phing db-create dirs-create db-demo product-search-recreate-structure product-search-export-products grunt error-pages-generate warmup"]
+    +   command: ["sh", "-c", "cd /var/www/html && ./phing -D production.confirm.action=y db-create dirs-create db-demo product-search-recreate-structure product-search-export-products grunt error-pages-generate warmup"]
+    ```
+- check all the phing targets that depend on the new `production-protection` target
+    - if you use any of the targets in your automated build scripts in production environment, you need to pass the confirmation to the phing using `-D production.confirm.action=y`
+
 ### Application
 
 - update your twig files ([#1284](https://github.com/shopsys/shopsys/pull/1284/)):
@@ -23,6 +32,7 @@ There you can find links to upgrade notes for other versions too.
         - `src/Resources/views/Admin/Content/Brand/detail.html.twig`
         - `src/Resources/views/Admin/Content/Category/detail.html.twig`
         - `src/Resources/views/Admin/Content/Product/detail.html.twig`
+- add [`app/getEnvironment.php`](https://github.com/shopsys/shopsys/blob/9.0/project-base/app/getEnvironment.php) file to your project ([#1368](https://github.com/shopsys/shopsys/pull/1368))
 
 ### Tools
 
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -703,9 +703,14 @@
         <if>
             <equals arg1="${environment}" arg2="prod"/>
             <then>
-                <input propertyName="production.confirm.action" validArgs="y,n" defaultValue="n" message="Warning! You are now in production environment and performing action that is not supposed to be ran in this environment! Are you sure you want to continue?"/>
                 <if>
-                    <equals arg1="${production.confirm.action}" arg2="n"/>
+                    <not><isset property="production.confirm.action"/></not>
+                    <then>
+                        <input propertyName="production.confirm.action" validArgs="y,n" defaultValue="n" message="Warning! You are now in production environment and performing action that is not supposed to be ran in this environment! Are you sure you want to continue?"/>
+                    </then>
+                </if>
+                <if>
+                    <not><equals arg1="${production.confirm.action}" arg2="y"/></not>
                     <then>
                         <fail message="Terminated by user" status="0"/>
                     </then>
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -72,13 +72,13 @@
 
     <target name="build" depends="build-deploy-part-1-db-independent,build-deploy-part-2-db-dependent" description="Builds application for production preserving your DB."/>
 
-    <target name="build-demo" depends="wipe,build-version-generate,composer-prod,redis-check,npm,dirs-create,assets,db-demo,grunt,error-pages-generate,warmup,product-search-recreate-structure,product-search-export-products,clean-redis-old" description="Builds application for production with clean demo DB."/>
+    <target name="build-demo" depends="production-protection,wipe,build-version-generate,composer-prod,redis-check,npm,dirs-create,assets,db-demo,grunt,error-pages-generate,warmup,product-search-recreate-structure,product-search-export-products,clean-redis-old" description="Builds application for production with clean demo DB."/>
 
-    <target name="build-demo-ci" depends="build-version-generate,wipe-excluding-logs,composer-dev,npm,timezones-check,dirs-create,test-dirs-create,assets,db-demo,product-search-recreate-structure,product-search-export-products,grunt,error-pages-generate,tests-acceptance-build,clean-redis-old,checks-ci" description="Builds application for development with clean demo DB and runs CI checks."/>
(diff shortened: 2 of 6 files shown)


packages/framework/build.xml
```diff
@@ -488,6 +629,17 @@
         </exec>
     </target>
 
+    <target name="environment-load" description="Load Symfony environment into Phing properties." hidden="true">
+        <if>
+            <not><isset property="environment"/></not>
+            <then>
+                <exec command="php bin/console about | grep Environment | sed 's/Environment//;s/ //g'" outputProperty="environment"/>
```
vitek-rostislav: I do not understand `sed` enough to be able to validate this, however, it works, so ok :grimacing:

packages/framework/build.xml
```diff
@@ -679,6 +831,23 @@
 
     <target name="product-search-recreate-structure" depends="product-search-delete-structure,product-search-create-structure" description="Recreates structure for searching via elasticsearch (deletes existing structure and creates new one)"/>
 
+    <target name="production-protection" depends="environment-load" description="Asks user if wants to run desctructive actions in production environemnt" hidden="true">
+        <if>
+            <and>
+                <not><isset property="productionConfirmDestructiveAction"/></not>
```
vitek-rostislav: we do not use camelCase for phing properties, however, I am not sure, what should be the proper name here, maybe just `production.confirm.destructive.action`?

packages/framework/build.xml
```diff
@@ -72,13 +72,94 @@
...
 
-    <target name="build-demo" depends="wipe,build-version-generate,composer-prod,redis-check,npm,dirs-create,assets,db-demo,grunt,error-pages-generate,warmup,product-search-recreate-structure,product-search-export-products,clean-redis-old" description="Builds application for production with clean demo DB."/>
+    <target name="build-demo" depends="production-protection" description="Builds application for production with clean demo DB.">
+        <if>
+            <equals arg1="${productionConfirmDestructiveAction}" arg2="y"/>
+            <then>
+                <phingcall target="wipe"/>
+                <phingcall target="build-version-generate"/>
+                <phingcall target="composer-prod"/>
+                <phingcall target="redis-check"/>
```
vitek-rostislav: you are missing `npm` here

packages/framework/build.xml
```diff
@@ -679,6 +831,23 @@
 
     <target name="product-search-recreate-structure" depends="product-search-delete-structure,product-search-create-structure" description="Recreates structure for searching via elasticsearch (deletes existing structure and creates new one)"/>
 
+    <target name="production-protection" depends="environment-load" description="Asks user if wants to run desctructive actions in production environemnt" hidden="true">
+        <if>
+            <and>
```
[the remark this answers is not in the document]
TomasLudvik: Yes, you can have only one argument in if... so you need to nest it :)

packages/framework/build.xml
TomasLudvik: fixed

packages/framework/build.xml
TomasLudvik: fixed

TomasLudvik:
> Hi, I have accidentally submitted the review before writing the comment 😄 I have only few things for you to fix 😉
> 
> And one question - have you consulted the list of targets that should be protected with anyone from implementation teams? To me, it seems more or less ok (see below), but it would be nice to have it validated by them.
> 
> I am only thinking about why `build-dev(-quick)` and `product-search-delete-structure` are not protected? I know you need to call the second one on production, however, it can cause troubles when you call it by accident, or not? 🤔

@vitek-rostislav those build-dev does not run any destructive action, the product delete structure is necessary on production, so I do not want to add that check there.

packages/framework/build.xml
vitek-rostislav: originally, the `npm` was called after `redis-check`, I think you should not change that unless you have some reason for that :innocent:

vitek-rostislav:
> @vitek-rostislav those build-dev does not run any destructive action, the product delete structure is necessary on production, so I do not want to add that check there.

Ok, and could you please confirm that the list of targets was confirmed by someone from implementation? It seems crucial to me

packages/framework/build.xml
henzigo: I'm not sure if phingcall will not call duplicate targets more then once.

TomasLudvik: @henzigo I do not understand what you mean

henzigo: Oh, I will show you an example:

Look about those commands:
```
    <target name="db-check-mapping" depends="clean-redis">
    ...
    </target>

    <target name="db-check-schema" depends="clean-redis,db-check-mapping">
       ...
    </target>
```

If you will have command `db-check` that has to depend on those two commands then `clean-redis` will be called only once. 

But if you will call `php phing db-check-mapping db-check-schema` then `clean-redis` will be called twice and I'm not sure which of those rules will be applied if you will use phingcall

TomasLudvik: @henzigo I understand now. I am afraid that security of production is more important then little longer build, do you agree with me?

TomasLudvik:
@vitek-rostislav you can do new CR

packages/framework/build.xml
henzigo: But you can easily check that in production-protection command (and throw error if you will not confirm that) and then add only depends on that target without change any of those commands, so you will use depends and you will have protection.

packages/framework/build.xml
```diff
@@ -634,9 +640,10 @@
         <if>
             <not><isset property="environment"/></not>
             <then>
-                <exec command="php bin/console about | grep Environment | sed 's/Environment//;s/ //g'" outputProperty="environment"/>
+                <exec command="php ${path.app}/getEnvironment.php" outputProperty="environment"/>
 
                 <property name="environment" value="${environment}"/>
+                <echo>${environment}</echo>
```
vitek-rostislav: not sure whether this is necessary, it means that the environment info is now printed every time any target with protection is executed, but on the other hand, why not, I do not mind it :grin:

project-base/app/getEnvironment.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Shopsys;
+
+require_once __DIR__ . '/autoload.php';
+
+file_put_contents("php://output", Environment::getEnvironment(false));
```
vitek-rostislav: why the `false` value for `$console` parameter? You are calling this from the console only so I am confused now

docs/upgrade/UPGRADE-v9.0.0-dev.md
```diff
@@ -7,6 +7,13 @@ There you can find links to upgrade notes for other versions too.
 
 ## [shopsys/framework]
 
+### Infrastructure
+- update your `kubernetes/deployments/webserver-php-fpm.yml` file: ([#1368](https://github.com/shopsys/shopsys/pull/1368))
+    ```diff
+    -   command: ["sh", "-c", "cd /var/www/html && ./phing db-create dirs-create db-demo product-search-recreate-structure product-search-export-products grunt error-pages-generate warmup"]
+    +   command: ["sh", "-c", "cd /var/www/html && ./phing -D productionConfirmDestructiveAction=y db-create dirs-create db-demo product-search-recreate-structure product-search-export-products grunt error-pages-generate warmup"]
```
vitek-rostislav: there should be `production.confirm.destructive.action` (and `production.confirm.action` after 6845dcb204b04115e0fdc94be8ad4c4c8f4c3940)

packages/framework/build.xml
```diff
@@ -708,291 +885,476 @@
...
+        </if>
     </target>
 
-    <target name="test-product-search-delete-structure" description="Deletes structure for searching via elasticsearch for test environment." hidden="true">
-        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
-            <arg value="${path.bin-console}"/>
-            <arg value="--env=test"/>
-            <arg value="shopsys:product-search:delete-structure"/>
-        </exec>
+    <target name="test-product-search-create-structure" depends="production-protection" description="Creates structure for searching via elasticsearch for test environment." hidden="true">
```
vitek-rostislav: I have to ask once more - why this command is protected, and `product-search-delete-structure` is not? I know you need to delete structure on production, however, such an action should be protected IMHO.

packages/framework/build.xml
vitek-rostislav: Hi @TomasLudvik, I do agree with @henzigo here, it would be easier to move the logic to the protection target, and then let the other targets be protected by adding `production-protection` target as their dependency, what do you say? The thing is, it is not just about longer build - sometimes running one target multiple times might be harmful. And moreover, repeating this condition
```
<if>
   <equals arg1="${production.confirm.action}" arg2="y"/>
   <then>
...
```
in multiple targets is DRY violaion :slightly_frowning_face:

Issue #1059, opened 2019-05-21
Use phpstan-symfony


DavidOstrozlik:
Use https://github.com/phpstan/phpstan-symfony to check symfony-related problems.

There is hopefully no trouble right now, but it is a step towards safer code.

PetrHeinz:
I've tried doing that, but there are problems with dependencies. Our forks of Symplify repositories and some other dependencies use `nette/utils` in `2.5` and `phpstan-symfony` requires it in `3.0`.

The best approach would probably be fixing the CS in https://github.com/Symplify/Symplify/pull/1589 so it can be (hopefully) merged, abandoning our forks, and upgrading all packages that require the older `nette/utils`...

@vitek-rostislav can you fix the PR? Running `ecs` with `--fix` and commiting the result should suffice.

vitek-rostislav:
Hi @PetrHeinz, I am sorry for the late response. I have fixed the PR (I actually realized that a part of it is completely unnecessary so I created a new PR in our repository as well :slightly_smiling_face:) https://github.com/shopsys/shopsys/pull/1370

ondrejmirtes:
PHPStan 0.12 doesn't declare any dependencies so I recommend upgrading to get rid of the conflict and start using phpstan-symfony,

sslt:
Open it stalebot!

PR #1370, opened 2019-09-04
[shopsys] releaser simplification

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to allow releasing of a lower tag than the newest one as we we are supporting multiple versions. This is now allowed by the usage of our own fork of MonorepoBuilder (see https://github.com/Symplify/MonorepoBuilder/pull/4), however, I found out that it is completely unnecessary. The desired result can be achieved by configuration :man_facepalming: Thanks to that, I could also simplify my PR https://github.com/Symplify/Symplify/pull/1589
|New feature| No 
|Fixes issues| ... 

After this is merged, [the fork](https://github.com/shopsys/MonorepoBuilder) should be removed from github and from packagist (despite the fact it is a bad practice, I believe we can do that and cause no harm to anybody)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -150,8 +150,8 @@
     "nette/utils": "^2.5",
     "phpstan/phpstan": "^0.11",
     "symplify/easy-coding-standard-tester": "^5.2.17",
+    "symplify/monorepo-builder": "^5.2.17",
     "shopsys/changelog-linker": "^5.5.0",
-    "shopsys/monorepo-builder": "^5.5.0",
     "phpstan/phpstan-doctrine": "^0.11.2",
     "phpstan/phpstan-phpunit": "^0.11.2"
   },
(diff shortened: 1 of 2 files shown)


PetrHeinz:
Good catch! Imo we should just mark the fork as abandoned, otherwise earlier vesions get uninstallable...

Why did you choose 5.2.17 instead of 5.5.0?

vitek-rostislav:
Thanks :slightly_smiling_face: Well, I returned the original version restriction that we used before we started using the fork. Actually, there is no `5.5.0` version of the Symplify package - we released our own tag - I was surprised by that as well :grin: 

It seems like https://github.com/Symplify/Symplify/pull/1589 will be merged soon :tada: I believe after the new version of Symplify pakcages is released, it will be a good time update our dependencies. Not only because of https://github.com/shopsys/shopsys/issues/1059, I believe we had some more reasons for that, I thought we already had some issue or user-story for that, however, I am not able to find it now.

Btw should not you be on a vacation? :laughing:

PetrHeinz:
@vitek-rostislav Rest assured I'm vacating properly on sunny Croatian beaches 😎 That doesn't prevent me from peeking on GitHub from my mobile under a fig tree 😉

PR #1372, opened 2019-09-05
[framework] only set new url as main when entity doesn't have one

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| if you have your custom url and delete the one that was generated from name of the entity, application will create a new url from name and set it as main
|New feature| No 
|Fixes issues| #668  

### DIFF
--- a/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
+++ b/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
@@ -75,7 +75,8 @@ class FriendlyUrlFacade
         $friendlyUrls = $this->friendlyUrlFactory->createForAllDomains($routeName, $entityId, $namesByLocale);
         foreach ($friendlyUrls as $friendlyUrl) {
             $locale = $this->domain->getDomainConfigById($friendlyUrl->getDomainId())->getLocale();
-            $this->resolveUniquenessOfFriendlyUrlAndFlush($friendlyUrl, $namesByLocale[$locale]);
+            $mainFriendlyUrl = $this->findMainFriendlyUrl($friendlyUrl->getDomainId(), $routeName, $entityId);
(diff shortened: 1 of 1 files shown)


packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
```diff
@@ -96,8 +97,9 @@ public function createFriendlyUrlForDomain($routeName, $entityId, $entityName, $
     /**
      * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrl $friendlyUrl
      * @param string $entityName
+     * @param bool $setAsMain
      */
-    protected function resolveUniquenessOfFriendlyUrlAndFlush(FriendlyUrl $friendlyUrl, $entityName)
+    protected function resolveUniquenessOfFriendlyUrlAndFlush(FriendlyUrl $friendlyUrl, $entityName, bool $setAsMain = true)
```
vitek-rostislav: sadly, this is a BC break :disappointed: If someone extended the facade and overwrote the method, his application would start failing as the declaration would not be compatible with the parent anymore...could you please rewrite this to be backward compatible? Or maybe retargeting to `9.0` would be ok, however, please, consult it with @LukasHeinz - we wanted to have this fixed in `7.3.2`

DavidOstrozlik:
Ok we have this problem. We don't know how should the URLs behave in certain use-cases.
Lets say that I have category "Electonics" with main URL "/electronics/" that was generated from the name of the category.
1. If I change the name of the category to "Books" the application creates new URL "/books/". Should it set "/books/" as the main URL of the category?
2. I add my custom URL to the category "/elec-123/" and set it as main URL. Then I rename the category to "Books". The application creates new URL "/books/". Should it set "/books/" as the main URL of the category or keep the custom one?
Can you please make this clear @LukasHeinz ?

LukasHeinz:
Just for reference - the quite similar topic was rised here - https://github.com/shopsys/shopsys/pull/669#issuecomment-446506677

PR #1380, opened 2019-09-06
added password environment variable from hosts

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Special jobs started failing smoke tests because of admin could not be logged in.
|New feature| No 
|Fixes issues| ... 

---
I've added environmet variable do `docker-compose.yml.dist` but I'm a bit unsure if we want to add it into rest of docker-composes dists.


vitek-rostislav:
We discussed this with @grossmannmartin and we would rather document the usage of ENV variable better (I will create a new PR for that) and fix the special jobs directly on our CI by manipulating with `docker-compose` there.

PR #1384, opened 2019-09-06
[shopsys] simplified filling variables in Kubernetes manifests

### LINKED ISSUE #1271: Build application in kubernetes should be simplified
### What is happening
In deploy/build scripts is often used yq to replace placeholders or some values with the proper one right for the environment.

This replacement often looks like 
```sh
# Set environment variables to container and initContainer for Google Cloud Storage connection - set it to null
yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.containers[0].env[0].value 'bucket-name'
yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.containers[0].env[1].value 'project-id'
yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.initContainers[1].env[0].value 'bucket-name'
yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.initContainers[1].env[1].value 'project-id'
```

Which transforms (shortened example)

```yaml
spec:
    template:
            initContainers:
                -   name: copy-source-codes-to-volume
                -   name: initialize-database
                    env:
                        -   name: GOOGLE_CLOUD_STORAGE_BUCKET_NAME
                            value: ~
                        -   name: GOOGLE_CLOUD_PROJECT_ID
                            value: ~
            containers:
                -   name: php-fpm
                    env:
                        -   name: GOOGLE_CLOUD_STORAGE_BUCKET_NAME
                            value: ~
                        -   name: GOOGLE_CLOUD_PROJECT_ID
                            value: ~
```

into

```yaml
spec:
    template:
            initContainers:
                -   name: copy-source-codes-to-volume
                -   name: initialize-database
                    env:
                        -   name: GOOGLE_CLOUD_STORAGE_BUCKET_NAME
                            value: bucket-name
                        -   name: GOOGLE_CLOUD_PROJECT_ID
                            value: project-id
            containers:
                -   name: php-fpm
                    env:
                        -   name: GOOGLE_CLOUD_STORAGE_BUCKET_NAME
                            value: bucket-name
                        -   name: GOOGLE_CLOUD_PROJECT_ID
                            value: project-id
```

This is violation of DRY and most of all very fragile solution.
It's really easy to prepend new env variable as the @PetrHeinz correctly pointed out.

Maybe we could use Kustomize variables (see [vars documentation](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/fields.md#vars) or [an example using vars](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/wordpress/README.md)) to not repeat ourselves like this.

### Expected result

Scripts should be less error-prone and simplified.

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Kubernetes parameters in configuration files are now replaced via `yq` which is affecting specified configuration nodes - that makes the `build_script.sh` dependent on nodes order.<br>For better stability and maintainability placeholders were introduced in kubernetes config files. Build script now iterate all config files in kubernetes folder and replace them from environment variables.
|New feature| No 
|Fixes issues| closes #1271  


.ci/build_kubernetes.sh
```diff
@@ -9,18 +9,25 @@ echo ${DOCKER_PASSWORD} | docker login --username ${DOCKER_USERNAME} --password-
...
+)
+VARS=(
+    FIRST_DOMAIN_HOSTNAME
+    SECOND_DOMAIN_HOSTNAME
+)
 
-# Set domain into webserver hostnames
-yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.hostAliases[0].hostnames[+] ${FIRST_DOMAIN_HOSTNAME}
-yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.hostAliases[0].hostnames[+] ${SECOND_DOMAIN_HOSTNAME}
+for FILE in "${FILES[@]}"
```
PetrHeinz: It would make more sense to me to first copy and modify the files in `project-base/app/config/`, build and push the images and then replace all the variables in Kubernetes manifests and Kustomize config in one loop.

project-base/kubernetes/deployments/webserver-php-fpm.yml
```diff
@@ -42,40 +44,40 @@ spec:
...
+                    image: "{{DOCKER_PHP_FPM_IMAGE}}"
                     securityContext:
                         runAsUser: 33
                     command: ["sh", "-c", "cp -r /var/www/html/. /tmp/source-codes"]
                     volumeMounts:
                         -   name: source-codes
                             mountPath: /tmp/source-codes
                         -   name: domains-urls
-                            mountPath: /var/www/html/app/config/domains_urls.yml
+                            mountPath: "{{PATH_CONFIG_DOMAIN_URL}}"
```
PetrHeinz: Why did you parameterize the paths?

.ci/build_kubernetes.sh
```diff
@@ -9,18 +9,25 @@ echo ${DOCKER_PASSWORD} | docker login --username ${DOCKER_USERNAME} --password-
...
-
-# Set domain into dev-stack (redis-admin, elasticsearch, adminer) hostnames
-yq write --inplace project-base/kubernetes/kustomize/overlays/ci/ingress-patch.yaml [0].value.host adminer.${FIRST_DOMAIN_HOSTNAME}
-yq write --inplace project-base/kubernetes/kustomize/overlays/ci/ingress-patch.yaml [1].value.host elasticsearch.${FIRST_DOMAIN_HOSTNAME}
-yq write --inplace project-base/kubernetes/kustomize/overlays/ci/ingress-patch.yaml [2].value.host redis-admin.${FIRST_DOMAIN_HOSTNAME}
+FILES=(
+    project-base/kubernetes/ingress.yml
+    project-base/kubernetes/kustomize/overlays/ci/ingress-patch.yaml
+    project-base/kubernetes/deployments/webserver-php-fpm.yml
+)
```
henzigo: What about to get all files recursively from path `kubernetes`? This can easily make a mistake, because you have to think that you need to add file path here.

.ci/build_kubernetes.sh
henzigo: So you want to push sensitive information like database password to external docker repository? Probably you do not need to set all parameters before push, but you can do that after push to docker repository.

project-base/kubernetes/deployments/webserver-php-fpm.yml
s3tezsky: It was also replaced by `yq` but from your comment I guess it was not necessary.

s3tezsky: I have noticed that replacement was different (`/var/www/html/project-base/app/config/domains_urls.yml`). After discussion we decided to parametrize only directory path.

.ci/build_kubernetes.sh
PetrHeinz: @henzigo I understand your concern, but it's already done like that, I'm just speaking about making it simpler without changing the resulting behavior. The parameters don't actually contain sensitive information as the Kubernetes infrastructure doesn't allow access from outside network.

But generally speaking, it would be best to have info in `parameters.yml` not pushed to the Docker repository (ideally not saved in a file at all as per https://12factor.net/config).

project-base/kubernetes/deployments/webserver-php-fpm.yml
PetrHeinz: Makes sense. In that case, I would parameterize just the path to `app/config`.

.ci/build_kubernetes.sh
s3tezsky: That would be really nice to see in it. I will train my bash skills for next level ;)

.ci/build_kubernetes.sh
henzigo: @PetrHeinz We are pushing only default parameters from `parameters.yml.dist` and then we are overriding them with our custom parameters.

You are right that Kubernetes infrastructure doesn't allow access from outside network, but we are using services like `adminer` where you can use those parameters.

.ci/build_kubernetes.sh
```diff
@@ -58,27 +45,32 @@ docker image pull ${DOCKER_USERNAME}/elasticsearch:${DOCKER_ELASTIC_IMAGE_TAG} |
...
-yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.containers[0].env[1].value ''
-yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.initContainers[1].env[0].value ''
-yq write --inplace project-base/kubernetes/deployments/webserver-php-fpm.yml spec.template.spec.initContainers[1].env[1].value ''
+DOCKER_PHP_FPM_IMAGE=${DOCKER_USERNAME}/php-fpm:${DOCKER_IMAGE_TAG}
+DOCKER_ELASTIC_IMAGE=${DOCKER_USERNAME}/elasticsearch:${DOCKER_ELASTIC_IMAGE_TAG}
+PATH_CONFIG_DIRECTORY='/var/www/html/project-base/app/config'
+GOOGLE_CLOUD_STORAGE_BUCKET_NAME=''
+GOOGLE_CLOUD_PROJECT_ID=''
+
+FILES=(find . -type f -regex 'project-base\/kubernetes\/.*\.ya?ml$')
```
PetrHeinz: Run the command in a sub-shell via `$(...)`:

```suggestion
FILES=$(find . -type f -regex 'project-base\/kubernetes\/.*\.ya?ml$')
```
---
*BTW do you know about https://devhints.io/bash? It's a great reference for Bash, I use it quite often :slightly_smiling_face:*

s3tezsky:
As it is affecting only project-base I guess it is not BC break. Current clients will still have an old build script untill they upgrade.

henzigo:
Hello, please look at my two points: 

1. What about parameters.yml? There are many parameters on project and you need currently to set them with yq. (On our project currently 47 lines).
2. Please do not forget that there exists another build script here https://github.com/shopsys/shopsys/blob/3208be11d559cc882bd69cbe4fbe27ef7f3e7403/project-base/.ci/deploy-to-google-cloud.sh

s3tezsky:
> What about parameters.yml? There are many parameters on project and you need currently to set them with yq. (On our project currently 47 lines).
I do not think that parametrize `parameters.yml` is a good idea. It will require a different solution (maybe force override parameters with environment variables?). Because of that I would rather extract it into new PR.

.ci/build_kubernetes.sh
```diff
@@ -58,27 +45,32 @@ docker image pull ${DOCKER_USERNAME}/elasticsearch:${DOCKER_ELASTIC_IMAGE_TAG} |
...
+    FIRST_DOMAIN_HOSTNAME
+    SECOND_DOMAIN_HOSTNAME
+    DOCKER_PHP_FPM_IMAGE
+    DOCKER_ELASTIC_IMAGE
+    PATH_CONFIG_DIRECTORY
+    GOOGLE_CLOUD_STORAGE_BUCKET_NAME
+    GOOGLE_CLOUD_PROJECT_ID
+)
+
+for FILE in $FILES
```
PetrHeinz: I would use single line syntax for initiation the cycle just like in other our shell/bash scripts:

```suggestion
for FILE in $FILES; do
```

.ci/build_kubernetes.sh
```diff
@@ -58,27 +45,32 @@ docker image pull ${DOCKER_USERNAME}/elasticsearch:${DOCKER_ELASTIC_IMAGE_TAG} |
...
+do
```
PetrHeinz:
```suggestion
```

.ci/build_kubernetes.sh
```diff
@@ -58,27 +45,32 @@ docker image pull ${DOCKER_USERNAME}/elasticsearch:${DOCKER_ELASTIC_IMAGE_TAG} |
...
+    for VAR in ${VARS[@]}
```
PetrHeinz:
```suggestion
    for VAR in ${VARS[@]}; do
```

.ci/build_kubernetes.sh
```diff
@@ -58,27 +45,32 @@ docker image pull ${DOCKER_USERNAME}/elasticsearch:${DOCKER_ELASTIC_IMAGE_TAG} |
...
+	do
```
PetrHeinz:
```suggestion
```

PetrHeinz: And the same is true for `project-base/.ci/deploy-to-google-cloud.sh`

PetrHeinz:
@s3tezsky The `v7.3.2` with #1198 bugfix was already released and merged, so we can finally finish this :slightly_smiling_face:

s3tezsky:
As it is affecting ` .ci/build_kubernetes.sh` which relays on k8s's configuration located in project base I'm marking it as a BC break again.

For same reaason I'm not sure about placing upgrade notes in `UPGRADE-v8.1.0-dev`. Shouldn't I create `UPGRADE-v9.0.0-dev`? (I'm a bit confused we have none yet)

PetrHeinz:
@s3tezsky I would still argue it's not a BC break. Changes in `project-base` repo are excluded from our BC promise and there is no compatibility issue between monorepo's `.ci/build_kubernetes.sh` and `project-base` - monorepo is not a package and will *always* contain the same version of `project-base` (by definition, as the repositories are developed together in the monorepo). If the script would be in any of the packages (`framework`, `http-smoke-testing`, ...), it would be a problem; but not in the monorepo itself.

Furthermore, `UPGRADE-v9.0.0-dev.md` exists, but only in `9.0` branch where it's being developed.

upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -102,6 +102,45 @@ There you can find links to upgrade notes for other versions too.
     - ENV ELASTIC_SEARCH_INDEX_PREFIX=''
 
     ```
+- parametrize variables in kubernetes configuration [[#1384]](https://github.com/shopsys/shopsys/pull/1384)
+    - walk through your [`.ci/deploy-to-google-cloud.sh`](https://github.com/shopsys/shopsys/blob/v8.0.0/project-base/.ci/deploy-to-google-cloud.sh) and notice every occurrences of using `yq` command which is affecting `yml` or `yaml` files in [`project-base/kubernetes`](https://github.com/shopsys/shopsys/tree/v8.0.0/project-base/kubernetes) (k8s' configuration files)
+        - in k8s' configuration files replace these occurrences with placeholder like this `{{FIRST_DOMAIN}}` (placeholder will be replaced by ENV variable with the same name)
```
PetrHeinz: I would add an example using a diff:
```suggestion
        - in k8s' configuration files replace these occurrences with placeholder like this `{{FIRST_DOMAIN_HOSTNAME}}` (the placeholder will be replaced by ENV variable with the same name)
            ```diff
             spec:
                 rules:
            -        -   host: ~
            +        -   host: "{{FIRST_DOMAIN_HOSTNAME}}"
                         http:
                             paths:
            ```
```

upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -102,6 +102,45 @@ There you can find links to upgrade notes for other versions too.
...
+        - in  `deploy-to-google-cloud.sh` replace `yq` commands by new code bellow
+            - change shebang of `.ci/deploy-to-google-cloud.sh`
+                ```diff
+                - #!/bin/sh -ex
+                + #!/bin/bash -ex
+                ```
+            - add code bellow to find all k8s' configuration files in `kubernetes` folder
```
PetrHeinz: We try not use the shorthand K8s for Kubernetes ([as said in the docs](https://docs.shopsys.com/en/latest/kubernetes/introduction-to-kubernetes/#terminology-of-kubernetes)), could you replace it by Kubernetes?

s3tezsky:
@PetrHeinz Thanks a lot for clarifying. Totally aggree with you. Monorepo != framework which make sense!

upgrade/UPGRADE-v8.1.0-dev.md
s3tezsky: Replaced

upgrade/UPGRADE-v8.1.0-dev.md
```diff
@@ -102,6 +102,53 @@ There you can find links to upgrade notes for other versions too.
     - ENV ELASTIC_SEARCH_INDEX_PREFIX=''
 
     ```
+- parametrize variables in kubernetes configuration [[#1384]](https://github.com/shopsys/shopsys/pull/1384)
```
LukasHeinz:
```suggestion
- parametrize variables in kubernetes configuration ([#1384])(https://github.com/shopsys/shopsys/pull/1384)
```

LukasHeinz: Just wrong brackets...

grossmannmartin:
Google Cloud has not domains configured properly. In `domain_urls.yml` is still default domain setting (`http://127.0.0.x:8000`) instead of required `https://gcX.shopsys-test.com`
And trusted proxies setting is not propagated too.

Can it be because in `build_kubernetes.sh` you replace parameters before image build and in `deploy-to-google-cloud.sh` after the image is built?

PR #1360, opened 2019-08-29
disallow administrator login with default credentials in production e…

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Disallow administrator login with default credentials in production environment. Login will fail and flash message is thrown.
|New feature| Yes
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| No
|Fixes issues| closes #489 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1389, opened 2019-09-10
[project-base] ErrorController uses environment from DIC parameter

DavidOstrozlik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Shopsys\Environment should be used only in Shopsys\Bootstrap
|New feature| No 
|Fixes issues| ... 

- [x] add upgrade notes

### DIFF
--- a/project-base/src/Shopsys/ShopBundle/Controller/Front/ErrorController.php
+++ b/project-base/src/Shopsys/ShopBundle/Controller/Front/ErrorController.php
@@ -40,21 +40,29 @@ class ErrorController extends FrontBaseController
     private $domain;
 
     /**
+     * @var string
+     */
+    private $environment;
+
+    /**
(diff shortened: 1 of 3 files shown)


PetrHeinz:
@vitek-rostislav the resons for this change are described in [a review comment](https://github.com/shopsys/shopsys/pull/1360/files#r321710747) that lead to #1386.

s3tezsky:
@PetrHeinz this cannot be merged into `7.3`. You may consider it as a bugfix but it changes the interface of `ErrorController::__construct()` according to our BC promisse it is not considered as a BC break but in case anyone is using extended `ErrorController` and calling this `__construct()` the app will throw `500`

PetrHeinz:
@s3tezsky But the change is done in project-base repository and as such it's excluded from our BC promise.

PR #1395, opened 2019-09-10
[shopsys] added sniff fobidden default values in Doctrine

### LINKED ISSUE #1242: Default values of entity properties should not be set using Doctrine
### What is happening
Currently (as of `v7.3.1`) we use Doctrine default values for 5 properties (see below).

Configuring default values in ORM is not really a best practice. Doctrine should shield the model from the database and having default values configured at the DB level achieves the exact opposite.

Doctrine docs mention that it's not supported to set the default value (see [the official docs](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/faq.html#how-can-i-add-default-values-to-a-column)) - that's why it has to be done via the `options={"default" = <value>}` parameter.

Default values should be defined in the entity constructor (or directly as a default value of the parameter itself) or the data object. Remember that in the best-case scenario, entities should work without the ORM or the DB itself as said by Marco Pivetta (Ocramius) in [Doctrine best practices](http://ocramius.github.io/doctrine-best-practices/#/26).

Imagine you create a new entity and try getting the attribute:
- If it has it's default value only defined at the DB level, you're gonna end up with a `TypeError` (or worse with `NULL`).
- If it has a default value only in constructor you will be fine - you'll get the default as expected.
- And if it has the default set both in constructor and on the DB level you'll be fine - but you'll be in danger of those two default values being changed to different values as it is a DRY violation.

Furthermore, it poses a problem for [entity extension](https://github.com/shopsys/shopsys/blob/v7.3.1/docs/extensibility/entity-extension.md) as annotation cannot be easily overriden (even though we managed to [get that working on our demoshop by overriding `LoadORMMetadataSubscriber` class](https://github.com/shopsys/demoshop/blob/master/src/Shopsys/ShopBundle/Component/EntityExtension/LoadORMMetadataSubscriber.php)).

### Expected result
There are no default values defined using the ORM, or their correct usage is documented (if there is a valid reason for that).

Default values are only defined on the DB level in migrations and the the end of the migrations, they are removed.

---

https://github.com/shopsys/shopsys/blob/aef43bab59ad65d8f0cfecd80b563ab3a91c35a4/packages/framework/src/Model/Product/Product.php#L187-L192
https://github.com/shopsys/shopsys/blob/aef43bab59ad65d8f0cfecd80b563ab3a91c35a4/packages/framework/src/Model/Product/Product.php#L221-L233
https://github.com/shopsys/shopsys/blob/aef43bab59ad65d8f0cfecd80b563ab3a91c35a4/packages/framework/src/Model/Newsletter/NewsletterSubscriber.php#L39-L43
https://github.com/shopsys/shopsys/blob/aef43bab59ad65d8f0cfecd80b563ab3a91c35a4/packages/framework/src/Component/Cron/CronModule.php#L28-L33

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR prevents using a DB default value in doctrine properties (with a sniff) and also removes usages through the codebase.
|New feature| Yes 
|Fixes issues| closes #1242 

### DIFF
--- a/packages/coding-standards/src/Sniffs/ForbiddenDoctrineDefaultValueSniff.php
+++ b/packages/coding-standards/src/Sniffs/ForbiddenDoctrineDefaultValueSniff.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\Sniffs;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\TokenHelper;
+
+class ForbiddenDoctrineDefaultValueSniff implements Sniff
+{
+    /**
+     * @return array
+     */
+    public function register(): array
+    {
+        return [\T_CLASS];
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $classPosition
+     */
+    public function process(File $file, $classPosition): void
+    {
+        $tokens = $file->getTokens();
+        $docBlockOpeningTagPositions = $this->getAllDocBlockOpeningTagPositions($file, $classPosition);
+
+        foreach ($docBlockOpeningTagPositions as $docBlockOpenTagPosition) {
+            $docBlockToken = $tokens[$docBlockOpenTagPosition];
+
+            $content = TokenHelper::getContent($file, $docBlockOpenTagPosition, $docBlockToken['comment_closer']);
+
+            if ($this->annotationContainsDefaultValue($content)) {
+                $file->addError(
+                    'Default value of entity properties cannot be used.',
+                    $docBlockOpenTagPosition,
+                    self::class
+                );
+            }
+        }
+    }
+
+    /**
+     * @param string $annotationString
+     * @return bool
+     */
+    protected function annotationContainsDefaultValue(string $annotationString): bool
+    {
+        return (bool)preg_match('~options\s*=\s*\{\s*.*"default"~', $annotationString);
+    }
+
+    /**
+     * @param \PHP_CodeSniffer\Files\File $file
+     * @param int $startPosition
+     * @return int[]
+     */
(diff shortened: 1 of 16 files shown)


pk16011990:
Hi, this feature looks great, what is missing for finishing?

PetrHeinz:
@pk16011990 It seems that a usage in the actual coding standards is missing - if it actually works as it should, it would not be hard to finish at all.

packages/coding-standards/src/Sniffs/ForbiddenDoctrineDefaultValueSniff.php
```diff
@@ -0,0 +1,71 @@
...
+                );
+            }
+        }
+    }
+
+    /**
+     * @param string $annotationString
+     * @return bool
+     */
+    protected function annotationContainDefaultValue(string $annotationString): bool
```
TomasLudvik, line 49:
```suggestion
    protected function annotationContainsDefaultValue(string $annotationString): bool
```

grossmannmartin: thanks, fixed manually (due to usage)

packages/coding-standards/src/Sniffs/ForbiddenDoctrineDefaultValueSniff.php
```diff
@@ -0,0 +1,71 @@
...
+        $docBlockOpeningTagPositions = $this->getAllDocBlockOpeningTagPositions($file, $classPosition);
+
+        foreach ($docBlockOpeningTagPositions as $docBlockOpenTagPosition) {
+            $docBlockToken = $tokens[$docBlockOpenTagPosition];
+
+            $content = TokenHelper::getContent($file, $docBlockOpenTagPosition, $docBlockToken['comment_closer']);
+
+            if ($this->annotationContainsDefaultValue($content)) {
+                $file->addError(
+                    'Default value of entity properties should not be used',
```
TomasLudvik, line 37:
```suggestion
                    'Default value of entity properties cannot be used.',
```

PR #1405, opened 2019-09-13
releaser: excluded forked shopsys/changelog-linker from travis check

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #1244 we introduced new forked package, but Travis status of all shopsys packages is checked while releasing the new version. Forked packages should be excluded from these checks and this PR exclude newly added shopsys/changelog-linker fork.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/utils/releaser/src/Travis/TravisStatusReporter.php
+++ b/utils/releaser/src/Travis/TravisStatusReporter.php
@@ -24,6 +24,7 @@ final class TravisStatusReporter
         'shopsys/postgres-search-bundle',
         'shopsys/doctrine-orm',
         'shopsys/jparser',
+        'shopsys/changelog-linker',
         // old packages
         'shopsys/syscart',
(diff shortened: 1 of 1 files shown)


PetrHeinz:
Why should we stop checking forked packages if we use the forks in SSFW?

vitek-rostislav:
@PetrHeinz I think we have never checked them in the first place, or have we?

grossmannmartin:
True, none of our forked packages is checked on Travis. Another question is "do we want to check them?".
shopsys/MonorepoBuilder (package excluded at the same time as this one) is archived and original package should be used instead (as changes we need was merged into the original). So that one I want to skip.
About this one, the changes we need was merged to original package and we plan to use Symplify/ChangelogLinker again. Until then this package could be checked, but I don't see much added value as it's already checked on travis for the original package.

PR #1414, opened 2019-09-17
[project-base] fix set category description for first domain

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixed set first domain description for categories
|New feature| Yes/No 
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Yes/No 
|Fixes issues| ... 

As you can see https://3v4l.org/r29p0 function `array_merge()` not replace value for 1. domain id http://prntscr.com/p73onr but `array_replace` works http://prntscr.com/p73pge 😊


PetrHeinz:
Oh, you're exactly right, category descriptions on the first domain were broken since #568 because of incorrect usage of `array_merge` (since it behaves differently with string and integer keys).

PetrHeinz:
I've just rebased your PR and retargeted it to `7.3` so we can get it fixed in upcoming `v7.3.2` :slightly_smiling_face:

PR #1428, opened 2019-09-25
files outside the docs now link the docs on docs.shopsys.com

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We have a new website (https://docs.shopsys.com/) where the documentation should be viewed primarly so it is sutiable to link the documentation on the site, not in github
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/read-model/README.md
+++ b/packages/read-model/README.md
@@ -3,7 +3,7 @@
 [![Build Status](https://travis-ci.org/shopsys/read-model.svg?branch=master)](https://travis-ci.org/shopsys/read-model)
 [![Downloads](https://img.shields.io/packagist/dt/shopsys/read-model.svg)](https://packagist.org/packages/shopsys/read-model)
 
-This bundle for [Shopsys Framework](https://www.shopsys.com) separates templates from model using [read model concept](https://github.com/shopsys/shopsys/blob/master/docs/model/introduction-to-read-model.md).
+This bundle for [Shopsys Framework](https://www.shopsys.com) separates templates from model using [read model concept](https://docs.shopsys.com/en/7.3/model/introduction-to-read-model/).
 The bundle is dedicated for projects based on Shopsys Framework (i.e. created from [`shopsys/project-base`](https://github.com/shopsys/project-base)) exclusively.
 
 This repository is maintained by [shopsys/shopsys] monorepo, information about changes are in [monorepo CHANGELOG.md](https://github.com/shopsys/shopsys/blob/master/CHANGELOG.md).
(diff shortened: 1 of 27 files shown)


packages/read-model/README.md
```diff
@@ -3,7 +3,7 @@
 [![Build Status](https://travis-ci.org/shopsys/read-model.svg?branch=master)](https://travis-ci.org/shopsys/read-model)
 [![Downloads](https://img.shields.io/packagist/dt/shopsys/read-model.svg)](https://packagist.org/packages/shopsys/read-model)
 
-This bundle for [Shopsys Framework](https://www.shopsys.com) separates templates from model using [read model concept](https://github.com/shopsys/shopsys/blob/master/docs/model/introduction-to-read-model.md).
+This bundle for [Shopsys Framework](https://www.shopsys.com) separates templates from model using [read model concept](https://docs.shopsys.com/en/7.3/model/introduction-to-read-model).
```
grossmannmartin:
```suggestion
This bundle for [Shopsys Framework](https://www.shopsys.com) separates templates from model using [read model concept](https://docs.shopsys.com/en/7.3/model/introduction-to-read-model/).
```

packages/read-model/src/Image/ImageView.php
```diff
@@ -9,7 +9,7 @@
  *
  * Class representing images in frontend
  *
- * @see https://github.com/shopsys/shopsys/blob/master/docs/model/introduction-to-read-model.md
+ * @see https://docs.shopsys.com/en/7.3/model/introduction-to-read-model
```
grossmannmartin:
```suggestion
 * @see https://docs.shopsys.com/en/7.3/model/introduction-to-read-model/
```

packages/read-model/src/Product/Listed/ListedProductView.php
```diff
@@ -14,7 +14,7 @@
  *
  * Class representing products in lists in frontend
  *
- * @see https://github.com/shopsys/shopsys/blob/master/docs/model/introduction-to-read-model.md
+ * @see https://docs.shopsys.com/en/7.3/model/introduction-to-read-model
```
grossmannmartin:
```suggestion
 * @see https://docs.shopsys.com/en/7.3/model/introduction-to-read-model/
```

vitek-rostislav:
Hi, @grossmannmartin, thanks for the review :ok_hand: 

> What about Changelog file? Lines 863, 955, 994 and 1030 are without changed links on purpose, or it's a mistake?

This was on purpose - I was not sure in which version I should link the articles, the latest? :thinking: As those are old entries (for alpha versions) in the changelog, I do not mind keeping it as is, but I can change it if you want :slightly_smiling_face: 

> And I was a little confused with the whole contributing section. Do me it almost looks like articles that should not be versioned (and for example, BC Promise is really linked to latest version).

Actually, this is a good idea, I was not sure about that section too, always linking the files in contributing section to the latest versions makes sense to me, I will change it this way :wink:

Issue #1322, opened 2019-08-12
Cursor in input does not look the same when empty and filled


grossmannmartin:
### What is happening

When the input field is empty and focused, cursor is to the full height of the input. But when you start to type something, the cursor shrinks.

Tested on Safari 12.1.2 on MacOS 10.14.6 (Mojave)

![cursor](https://user-images.githubusercontent.com/1177414/62891590-5bfa4700-bd46-11e9-99a4-9b1109ed856b.gif)

### Expected result
The cursor should look the same when empty and filled.

PR #1448, opened 2019-10-04
[framework] fix cursor size in form inputs for Apple devices

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fix cursor size when input is focused at Apple devices.
|New feature| No 
|Fixes issues| #1322  

### DIFF
--- a/packages/framework/src/Resources/styles/admin/core/form/input.less
+++ b/packages/framework/src/Resources/styles/admin/core/form/input.less
@@ -4,6 +4,7 @@
 @input-border-size: 2px;
 @input-line-height: @input-height - 2*@input-border-size;
 @input-padding-horizontal: 14px;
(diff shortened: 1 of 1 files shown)


packages/framework/src/Resources/styles/admin/core/form/input.less
```diff
@@ -18,8 +19,8 @@
     width: @input-width;
     max-width: 100%;
     height: @input-height;
-    line-height: @input-line-height;
```
s3tezsky: Can you delete the variable definition? Are we using it somewhere else?

TomasGottvald: Yes, we can. Deleted, thank you for your good point

Issue #1278, opened 2019-08-02
Column header in administration moves when I sort content by this column


LukasHeinz:
# What is happening
Administration > Settings > List and Nomenclatures > Flags: when I sort the list of flags by color, the headers of column jump to another location. This appears on almost every in line editation table.

![Snímek obrazovky 2019-08-01 v 14 58 23](https://user-images.githubusercontent.com/33071107/62295228-e13f4b00-b46c-11e9-818f-65b6cd4456c6.png)

### Expected result
Columns do not move when sorting is changed.

PR #1452, opened 2019-10-07
[framework] design - admin - sorting arrows not changing column width

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixes sort arrow in table header so, now it don´t change columns width.
|New feature| No 
|Fixes issues| #1278  

### DIFF
--- a/packages/framework/src/Resources/styles/admin/components/in/sort-arrows.less
+++ b/packages/framework/src/Resources/styles/admin/components/in/sort-arrows.less
@@ -19,9 +19,14 @@
 
         &--up {
             .draw-arrow-top(5px, @color-base, 5px, -8px);
+            position: absolute;
+            top: 14px;
         }
+
         &--down {
             .draw-arrow-bottom(5px, @color-base, 5px, -6px);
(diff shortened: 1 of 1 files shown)


packages/framework/src/Resources/styles/admin/components/in/sort-arrows.less
```diff
@@ -19,9 +19,14 @@
 
         &--up {
             .draw-arrow-top(5px, @color-base, 5px, -8px);
+            position: absolute;
+            top: 14px;
```
topolanekmartin: Should not be here `top: 0`? Or why is top position same as for down arrow?

TomasGottvald: It is arrow position in column at table header. Arrow down is on same position but show another arrow.

topolanekmartin: In this case it is okay. Thanks for explanation.

PR #1463, opened 2019-10-14
[project-base] - Basic styleguide

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Add basic default styleguide to project base.
|New feature| Yes 
|Fixes issues| ... 


TomasGottvald:
I have minified codemirrors JS and CSS.

project-base/styleguide/src/Shopsys/ShopBundle/Resources/scripts/frontend/responsiveToggle.js
```diff
@@ -0,0 +1,79 @@
+(function ($) {
```
DavidOstrozlik: why don't we use the JS files that are already in project-base? This adds another layer of things that we'll need to maintain, because after every change in these files in project-base this needs to be updated as well

project-base/styleguide/README.md
```diff
@@ -0,0 +1,70 @@
...
+### Note
+Colors in styleguide content are NOT loaded from LESS files of SSFW project )-; And need to be added manually.
+
+### Inspired by
+<a href="https://hugeinc.github.io/styleguide/">https://hugeinc.github.io/styleguide/</a>
+
+## ChangeLog
+
+07-10-2019
+- content according to Shopsys Framework Stable version 8.0
```
DavidOstrozlik: This shouldn't be here

docs/frontend/styleguide.md
```diff
@@ -0,0 +1,70 @@
+# Shopsys Framework StyleGuide
```
DavidOstrozlik: Why did you add the same file twice?

project-base/styleguide/README.md
```diff
@@ -0,0 +1,70 @@
+# Shopsys Framework StyleGuide
```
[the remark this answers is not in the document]
TomasGottvald: unified to lowrcase

project-base/styleguide/README.md
TomasGottvald: Deleted

project-base/styleguide/src/Shopsys/ShopBundle/Resources/scripts/frontend/responsiveToggle.js
TomasGottvald: This is minimum js setup we need for styleguide as standalone folder. If we can add some functionality to copy these files from project-base by cmd command it will be better - same as default css files.

docs/frontend/styleguide.md
TomasGottvald: Because of standalone styleguide without project. Should we add only link to github docs md file?

DavidOstrozlik: Yes, we don't want to have duplicated files so it would be great to just add a link

project-base/styleguide/src/Shopsys/ShopBundle/Resources/scripts/frontend/responsiveToggle.js
DavidOstrozlik: We can remove them, link the JS files to the ones in project-base and add a note in the standalone install guide that you need to copy these files in the folder and change the links to them so users will have a way to have it working as a standalone

docs/frontend/styleguide.md
TomasGottvald: Removed.

project-base/styleguide/src/Shopsys/ShopBundle/Resources/scripts/frontend/responsiveToggle.js
TomasGottvald: We need to create phing target which will copy needed files to right folder. Some files are from vendor folder.

pesektomas: I apologize for the interference, but couldn't we copy the files using grunt? Or ... wait for it ... do not you want to try webpack? :smile:

TomasGottvald: Actually is on @DavidOstrozlik to create phing target which will copy all files from current sources to our styleguide folder.

pesektomas: Ok, I wanted show second view on copy files problem. Btw you wanted try webpack. I think so this is great opportunity to try it :smile:

henzigo: Why do you want to create a style guide as a standalone app?

pesektomas: I don't understand question. your comment is the answer to my comment about webpack? If yes - my point was about replace gruntfile as webpack, thats all. Goťák wanted explore webpack and I think so he can try replace gruntfile in styleguide. It is real project and I think so it is the best thing for exploring.

henzigo: No, my comment was directed to @DavidOstrozlik and his comment:

> add a note in the standalone install guide that you need to copy these files in the folder and change the links to them so users will have a way to have it working as a standalone

Anyway I do not think that you should implement webpack in this PR. I like your mindset, but there is https://github.com/symfony/webpack-encore and I think that there can be some threats with it. Second point could be that you have planned to upgrade SSFW to Symfony 4 and you will have to use webpack encore instead of assetic, so you will have to explore Webpack anyway

TomasGottvald: @pesektomas I will think about it in next version of styleguide.
@henzigo We have more than one usecase of styleguide - Try to imagine, that you want to sent styleguide to your graphic design or to somebody - is it really neccessary to install whole project on his computer ? I don´t think so. So I want to let styleguide as easy HTML page which can be wrapped to 0,5MB ZIP and used everywhere you want. And it can be used on other projects not only on SSFW.

pesektomas: @henzigo ok, thanks for explanation. I wanted tell Goťák point to he can exploit webpack on this styleguide. No in this PR, no webpack encore (we can make a phone call about webpack and webpack encore :smile: )

TomasGottvald: So you mean, that styleguide could be a nice playground for exploring webpack, isn´t it?

DavidOstrozlik:
Styleguide is moved to PR #1485 .

PR #1475, opened 2019-10-20
[framework] Clean redis cache via scan and unlink

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Finding keys by pattern must be done via SCAN, not KEYS. KEYS is blocking command and It causes performance problems for traffic in production.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Redis/RedisFacade.php
+++ b/packages/framework/src/Component/Redis/RedisFacade.php
@@ -52,10 +52,25 @@ class RedisFacade
         foreach ($this->getCacheClients() as $redis) {
             $prefix = (string)$redis->getOption(Redis::OPT_PREFIX);
             $pattern = $prefix . '*';
-            if (!$this->hasAnyKey($redis, $pattern)) {
-                continue;
+            $this->cleanCacheByScan($redis, $pattern);
+        }
+    }
+
+    /**
+     * @param \Redis $redisClient
+     * @param string $pattern
+     */
+    protected function cleanCacheByScan(Redis $redisClient, string $pattern): void
+    {
+        $suggestedScanBatchSize = 1000;
+
+        $iterator = null;
+        $keys = null;
+        while ($keys !== false) {
+            $keys = $redisClient->scan($iterator, $pattern, $suggestedScanBatchSize);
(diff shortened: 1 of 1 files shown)


malyMiso:
Hi @henzigo, you rock :slightly_smiling_face: , thank you for your pull request.

Can you please fix failing test ? Then I'll do my best to process this PR ASAP.
Or do you need help with fixing it ?

henzigo:
Hello, @malyMiso I fixed my code. :)

grossmannmartin:
Hi guys, I retarget this PR to 7.3 as it makes sense to me to deliver this in patch version and deprecation note in changed files states it's deprecated since 7.3.3.

packages/framework/src/Component/Redis/RedisFacade.php
```diff
@@ -52,10 +52,25 @@ public function cleanCache(): void
...
+    protected function cleanCacheByScan(Redis $redisClient, string $pattern): void
+    {
+        $suggestedScanBatchSize = 1000;
+
+        $iterator = null;
+        $keys = null;
+        while ($keys !== false) {
+            $keys = $redisClient->scan($iterator, $pattern, $suggestedScanBatchSize);
+            if (is_array($keys) && count($keys) > 0) {
+                $redisClient->eval("return redis.call('unlink', ARGV[1])", $keys);
```
grossmannmartin: why `eval` and not `unlink` method directly? I would like to know if it's for some specific reason

henzigo: Hello, that's because Redis clients are weird. `$redisClient->scan()` will return all keys with some pattern, but when I want to use `$redisClient->unlink()` I need to remove prefix from all keys to be able to unlink them (Prefix is defined in `snc_redis.yml` file). Instead of removing prefix I prefer to use `$redisClient->eval()` to circumvent this behavior.

henzigo: And I had that `eval` function wrong. It was removing only first key. Now it's working as expected

grossmannmartin: Thank you 👍

PR #1478, opened 2019-10-25
[shopsys] bad constraint message for minimum password length

### LINKED ISSUE #1518: Wrong msgid of translation
There is translation `msgid` `Password cannot be longer then {{ limit }} characters`. In fact it is used in Length Constraints to validate minimal (!) length of password. So `msgid` should be e.g. `Password must be {{ limit }} characters long or longer`. Czech translation is correct and is according to reality.

len

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| "Password must be longer than {{ limit }} characters" was not good for min length constraint. This PR fixes that.
|New feature| No 
|Fixes issues| closes #1518 

PS: Again. I want dump translation so I want to use `php phing dump-translations` as I am used to but this command not exists 😐. Only exists info `php phing dump-translations` [is deprecated](https://github.com/shopsys/shopsys/blame/master/upgrade/UPGRADE-v7.3.0.md#L130). Maybe I am just dumb but I don't see exactly message `php phing dump-translations` is renamed/deleted. Maybe I understand wrong changelogs? 🤔 

PPS: It's BC break so I wanted to change all "then" to "than" in constraint but `php phing translations-dump` stopped working (I don't know why 😄) so I give it up 😄

### DIFF
--- a/packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
+++ b/packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
@@ -99,7 +99,7 @@ class AdministratorFormType extends AbstractType
     private function getFirstPasswordConstraints($scenario)
     {
         $constraints = [
-            new Constraints\Length(['min' => 6, 'minMessage' => 'Password cannot be longer then {{ limit }} characters']),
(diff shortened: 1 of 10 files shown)


sslt:
[bad constraint message for min password length](https://trello.com/c/EntH3hZK/15-bad-constraint-message-for-min-password-length)

PR #1483, opened 2019-10-30
[project-base] added support for image and iframe lazy loading

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Add Lazyload functionality for images and iframe as default feature.
|New feature| Yes 
|Fixes issues| ... 

Screen before:
obrazek

Screen after:
obrazek


packages/framework/src/Twig/ImageExtension.php
```diff
@@ -210,6 +210,19 @@ protected function getImageHtmlByEntityName(array $attributes, $entityName, $add
...
+        $isAttributeClassExistsAndNotEmpty = array_key_exists('class', $attributes) && $attributes['class'] !== '';
+        $htmlAttributes['class'] = sprintf(
+            '%s%s',
+            $useLazyLoading ? 'lazy' : '',
+            $isAttributeClassExistsAndNotEmpty ? ' ' . $attributes['class'] : ''
+        );
+
+        if ($useLazyLoading) {
+            $htmlAttributes['data-src'] = $htmlAttributes['src'];
+            $htmlAttributes['src'] = '';
```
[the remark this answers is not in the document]
pesektomas: Ok, I see that it is resolved by minylazyload library. But. I have problem with native lazyload. I think so you need attribute loading in img (or iframe) element, isn't it? I watched on this (https://web.dev/native-lazy-loading or in Czech https://www.vzhurudolu.cz/prirucka/lazy-loading-obrazku#nativni) articles and in your PR I miss loading attribute.

project-base/src/Shopsys/ShopBundle/Resources/scripts/frontend/lazyLoadInit.js
```diff
@@ -0,0 +1,18 @@
...
+    placeholder: 'placeholder.png'
+}, '.lazy'));
+
+(function ($) {
+
+    Shopsys = window.Shopsys || {};
+    Shopsys.lazyLoadCall = Shopsys.lazyLoadCall || {};
+
+    Shopsys.lazyLoadCall.inContainer = function (container) {
+        $(container).find('.lazyx').each(function () {
```
s3tezsky: Why is there a `x` in `find('.lazyx')`? Is that a typo or it has some reason?

pesektomas: Oh yes, i forgot write comment. Btw I think so we should use selector like this `[loading=lazy]`. I waiting for TG reaction ...

packages/framework/src/Twig/ImageExtension.php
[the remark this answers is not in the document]
TomasGottvald: Yes thats good point. Even in new minilazyload version is it recommended to use [loading=lazy] selector. Refactored. Please check it out.

project-base/src/Shopsys/ShopBundle/Resources/scripts/frontend/lazyLoadInit.js
TomasGottvald: "x" was my mistake. I forgot test value. Removed 

Yes thats good point. Even in new minilazyload version is it recommended to use [loading=lazy] selector. Refactored. Please check it out.

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -210,6 +210,14 @@ protected function getImageHtmlByEntityName(array $attributes, $entityName, $add
         $htmlAttributes = $attributes;
         unset($htmlAttributes['type'], $htmlAttributes['size']);
 
+        $useLazyLoading = array_key_exists('lazy', $attributes) ? (bool)$attributes['lazy'] : true;
+        $isAttributeClassExistsAndNotEmpty = array_key_exists('class', $attributes) && $attributes['class'] !== '';
+        $htmlAttributes['class'] = sprintf(
+            '%s%s',
+            $useLazyLoading ? 'lazy' : '',
+            $isAttributeClassExistsAndNotEmpty ? ' ' . $attributes['class'] : ''
```
malyMiso: I would prefer to set up space between classes in format pattern, `%s %s` and then use `trim()` to remove unnecessary whitespaces. This will also solve stripping whitespaces from the beginning and end of a string `$attributes['class']`

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -210,6 +210,14 @@ protected function getImageHtmlByEntityName(array $attributes, $entityName, $add
         $htmlAttributes = $attributes;
         unset($htmlAttributes['type'], $htmlAttributes['size']);
 
+        $useLazyLoading = array_key_exists('lazy', $attributes) ? (bool)$attributes['lazy'] : true;
```
malyMiso: Do we want every image to use lazy loading by default ? So every image in administration (e.g. icons) will be lazy loaded as well. Is it desired behavior ?

docs/frontend/lazyload.md
```diff
@@ -0,0 +1,21 @@
+# Lazyload
+Lazyload for images and iframes are implemented with [Minilazyload library](https://www.npmjs.com/package/minilazyload).
+
+All images are lazy loaded as default. Default selector for lazyload is [loading=lazy].
```
malyMiso:
```suggestion
All images are lazy loaded by default. Default selector for lazyload is [loading=lazy].
```

docs/frontend/lazyload.md
```diff
@@ -0,0 +1,21 @@
...
+Native lazy load is overrided - we are waiting for bigger browser support.
+
+## Disable lazyload
+At homepage slider is used Slick to rotate slider images. We dont want to use lazyload on these images. Slick has own lazyload function. We can disable minilazyload function by `lazy: false` attribute in img tag.
```
malyMiso:
```suggestion
Slick component is used as homepage slider for cycling through images. We don't want to use lazy loading for these images. Slick has own lazyload function. We can disable minilazyload function by setting attribute `lazy: false` in img tag.
```

docs/frontend/lazyload.md
```diff
@@ -0,0 +1,21 @@
...
+
+```twig
+{{ image(item, { lazy: false }) }}
+```
+
+### Manual lazyload call
+If you need in some situations (like popup menu with images) call lazyload function manually, or if you want load images on ajax content, you can use function:
```
malyMiso:
```suggestion
If you need lazy loading in some situations (like popup menu with images) call lazyload function manually, or if you want load images on ajax content, you can use function:
```

docs/frontend/lazyload.md
```diff
@@ -0,0 +1,21 @@
...
+
+```javascript
+Shopsys.lazyLoadCall.inContainer(container);
+```
+
+As container varibale use css selector for container with images. All images in this container will be loaded after function call.
```
malyMiso:
```suggestion
As container variable use CSS selector for container with images. All images in this container will be loaded after function call.
```

packages/framework/src/Twig/ImageExtension.php
pesektomas: Yes, you are right and I'll think about it in the future. However, this part of code was  remove.

packages/framework/src/Twig/ImageExtension.php
TomasGottvald: We have only two posibilities. One default disabled, other one enabled. For better performance I choosed default enabled. Icons are actually made by svg font. And I hope even this will be changed. Go go to better performance.

malyMiso: Fine by me

PR #1489, opened 2019-11-04
[frontend-api] Product variants and main variant can now be loaded via frontend API

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Added support for loading product variants and main variant via frontend API
|New feature| Yes
|Fixes issues| ...

### DIFF
--- a/packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
+++ b/packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
@@ -278,9 +278,9 @@ class ProductOnCurrentDomainFacade implements ProductOnCurrentDomainFacadeInterf
     /**
      * @return array
      */
-    public function getAllListableProducts(): array
+    public function getAllOfferedProducts(): array
     {
-        return $this->productRepository->getAllListableProducts(
+        return $this->productRepository->getAllOfferedProducts(
             $this->domain->getId(),
             $this->currentCustomer->getPricingGroup()
         );
--- a/packages/framework/src/Model/Product/ProductRepository.php
+++ b/packages/framework/src/Model/Product/ProductRepository.php
@@ -792,8 +792,8 @@ class ProductRepository
      * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup $pricingGroup
      * @return array
      */
-    public function getAllListableProducts(int $domainId, PricingGroup $pricingGroup): array
+    public function getAllOfferedProducts(int $domainId, PricingGroup $pricingGroup): array
     {
-        return $this->getAllListableQueryBuilder($domainId, $pricingGroup)->getQuery()->execute();
+        return $this->getAllOfferedQueryBuilder($domainId, $pricingGroup)->getQuery()->execute();
     }
 }
--- a/packages/frontend-api/src/Model/Resolver/Products/ProductResolver.php
+++ b/packages/frontend-api/src/Model/Resolver/Products/ProductResolver.php
@@ -47,16 +47,10 @@ class ProductResolver implements ResolverInterface, AliasedInterface
         }
 
         try {
-            $product = $this->productFacade->getByUuid($uuid);
+            return $this->productFacade->getByUuid($uuid);
         } catch (\Shopsys\FrameworkBundle\Model\Product\Exception\ProductNotFoundException $productNotFoundException) {
             throw new UserError($productNotFoundException->getMessage());
         }
-
-        if ($product->isVariant() === true) {
-            throw new UserError('This product is variant and there is no support for variants yet.');
-        }
-
-        return $product;
     }
 
     /**
--- a/packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
+++ b/packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
@@ -38,16 +38,37 @@ class ProductResolverMap extends ResolverMap
     {
         return [
             'Product' => [
-                'shortDescription' => function (Product $product) {
-                    return $product->getShortDescription($this->domain->getId());
-                },
-                'link' => function (Product $product) {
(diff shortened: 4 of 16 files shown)


packages/frontend-api/src/Model/Resolver/Products/ProductResolver.php
```diff
@@ -52,10 +52,6 @@ public function resolver(string $uuid): Product
             throw new UserError($productNotFoundException->getMessage());
         }
 
-        if ($product->isVariant() === true) {
-            throw new UserError('This product is variant and there is no support for variants yet.');
-        }
-
         return $product;
```
s3tezsky: Variable `$product` might be removed and replaced by `return $this->productFacade->getByUuid($uuid);` on line 50.

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
```diff
@@ -38,16 +38,38 @@ protected function map(): array
...
-                    return $product->getCategoriesIndexedByDomainId()[$this->domain->getId()];
+                self::RESOLVE_TYPE => function (Product $product) {
+                    if ($product->isMainVariant()) {
+                        return 'MainVariant';
+                    }
+                    if ($product->isVariant()) {
+                        return 'Variant';
+                    }
+
+                    return 'RegularProduct';
```
s3tezsky: Do not you like if/elseif/else?

upgrade/UPGRADE-v9.0.0-dev.md
```diff
@@ -90,6 +90,17 @@ There you can find links to upgrade notes for other versions too.
...
+- add optional frontend API - products variant to your project ([#1489](https://github.com/shopsys/shopsys/pull/1489)):
+    - copy necessary type definitions:
+        [MainVariant.types.yml from Github](https://github.com/shopsys/shopsys/blob/9.0/project-base/src/Shopsys/ShopBundle/Resources/graphql-types/MainVariant.types.yml) to `src/Shopsys/ShopBundle/Resources/graphql-types/MainVariant.types.yml`
+        [RegularProduct.types.yml from Github](https://github.com/shopsys/shopsys/blob/9.0/project-base/src/Shopsys/ShopBundle/Resources/graphql-types/RegularProduct.types.yml) to `src/Shopsys/ShopBundle/Resources/graphql-types/RegularProduct.types.yml`
+        [Variant.types.yml from Github](https://github.com/shopsys/shopsys/blob/9.0/project-base/src/Shopsys/ShopBundle/Resources/graphql-types/Variant.types.yml) to `src/Shopsys/ShopBundle/Resources/graphql-types/Variant.types.yml`
+    - update your [`Products.types.yml`](https://github.com/shopsys/shopsys/blob/9.0/project-base/src/Shopsys/ShopBundle/Resources/graphql-types/Products.types.yml)
+      ```diff
+          -    type: object
+          +    type: interface
+      ```
```
s3tezsky: This bullet is not necessary as `Products.types.yml` has not been released yet.

packages/frontend-api/src/Model/Resolver/Products/ProductResolver.php
pesektomas: :+1:

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
pesektomas: Hmm, you are right :smile:

upgrade/UPGRADE-v9.0.0-dev.md
pesektomas: :+1:

packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
```diff
@@ -278,9 +278,9 @@ public function getProductFilterCountDataForSearch(
     /**
      * @return array
      */
-    public function getAllListableProducts(): array
+    public function getAllOfferedProducts(): array
```
s3tezsky: What is the reason for renaming this method? Is that so important for us to cahnge the interface? I guess is is beceause `Offered` should return all products and `Listable` only return some chunk of product dependent on pagination?

It is a bit confusing to me but I belive we will not want to return all products in one request. Also this is not the scope of this PR.

packages/framework/src/Model/Product/ProductRepository.php
```diff
@@ -792,8 +792,8 @@ public function findByProductQueryParams(ProductQueryParams $query): PaginationR
      * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup $pricingGroup
      * @return array
      */
-    public function getAllListableProducts(int $domainId, PricingGroup $pricingGroup): array
+    public function getAllOfferedProducts(int $domainId, PricingGroup $pricingGroup): array
```
s3tezsky: Same thougths as in `ProductOnCurrentDomainFacade`.

packages/frontend-api/src/Model/Resolver/Products/ProductsResolver.php
```diff
@@ -38,7 +38,7 @@ public function __construct(
      */
     public function resolve(): array
     {
-        return $this->productOnCurrentDomainFacade->getAllListableProducts();
+        return $this->productOnCurrentDomainFacade->getAllOfferedProducts();
```
s3tezsky: According to previous changes I'm just curious if we want to use `ProductOnCurrentDomainFacade` only. It will make sense to me stick to the interface `ProductOnCurrentDomainFacadeInterface` and have the same functionality implementent even in `ProductOnCurrentDomainElasticFacade`. Then anyone could just switch the implementation in `services.yml`. But this is not the scope of this PR maybe we should create an issue or new PR for this. What do you think?

PR #1501, opened 2019-11-11
[shopsys] fixed copying codeception logs from pod

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When there were no codeception logs `.ci/export_logs.sh` failed on non-existing directory.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/.ci/export_logs.sh
+++ b/.ci/export_logs.sh
@@ -1,4 +1,4 @@
-#!/bin/sh -ex
+#!/bin/sh -e
 
 # used to copy logs of codeception which are not streamed
 WEBSERVER_PHP_FPM_CONTAINER_NAME="webserver-php-fpm"
--- a/.ci/export_logs.sh
+++ b/.ci/export_logs.sh
@@ -1,4 +1,4 @@
-#!/bin/sh -e
+#!/bin/sh -ex
 
 # used to copy logs of codeception which are not streamed
 WEBSERVER_PHP_FPM_CONTAINER_NAME="webserver-php-fpm"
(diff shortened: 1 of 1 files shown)


.ci/export_logs.sh
```diff
@@ -15,6 +15,8 @@ for POD_NAME in $(kubectl get pods -n ${JOB_NAME} | grep -v ^NAME | cut -f 1 -d
     if [[ ${POD_NAME} == *${WEBSERVER_PHP_FPM_CONTAINER_NAME}* ]]; then
         # copy codeception logs from php-fpm pod to local
         # we do not need to specify the php-fpm container because it is picked by default
-        kubectl cp ${JOB_NAME}/${POD_NAME}:/var/www/html/project-base/var/logs/codeception/ ${WORKSPACE}/logs/codeception
+        if [[ -n $(kubectl -n ${JOB_NAME} exec ${POD_NAME} -- ls -1 var/logs |grep '^codeception$') ]]; then
```
grossmannmartin: while running from monorepo context, I believe it should be 
```suggestion
        if [[ -n $(kubectl -n ${JOB_NAME} exec ${POD_NAME} -- ls -1 project-base/var/logs | grep '^codeception$') ]]; then
```

but if the point of the test is to check presence of the directory, I suggest to use 
```sh
if kubectl -n ${JOB_NAME} exec ${POD_NAME} -- test -d project-base/var/logs/codeception; then
```

s3tezsky: Yes, you are right with that project-base folder. Also I like the test command 👍

PR #1505, opened 2019-11-13
[framework] replaced egeloen/ordered-form by a fork

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR replaces package `egeloen/ordered-form` with fork `shopsys/ordered-form` as original seems to be unmaintained any more and is not compatible with symfony 4 (at least its `composer.json` says so).
|New feature| No 
|Fixes issues| ... 


composer.json
```diff
@@ -54,6 +54,12 @@
             "Shopsys\\Releaser\\Tests\\": "utils/releaser/tests"
         }
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/shopsys/ivory-ordered-form.git"
+        }
+    ],
```
grossmannmartin: If we want to maintain the fork, we should publish it to packagist.

composer.json
```diff
@@ -92,7 +98,7 @@
         "doctrine/doctrine-fixtures-bundle": "^3.0.2",
         "doctrine/doctrine-migrations-bundle": "^1.3.0",
         "doctrine/migrations": "^1.6.0",
-        "egeloen/ordered-form": "^3.0",
+        "egeloen/ordered-form": "^4.0",
```
grossmannmartin: this should be updated in `packages/framework/composer.json` too

s3tezsky: Fixed

PR #1506, opened 2019-11-13
[shopsys] fixed version of symfony/monolog-bundle

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Hotfix. After releasing `symfony/monolog-bundle v3.5.0` the composer installs `monolog/monolog v2` which results in `Symfony 5 is required for Monolog 2 support. Please downgrade Monolog to version 1.`
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -167,6 +167,7 @@
     },
     "conflict": {
         "symfony/dependency-injection": "3.4.15|3.4.16",
+        "symfony/monolog-bundle": "3.5.0",
         "symplify/better-phpdoc-parser": "5.4.14|5.4.15",
         "twig/twig": "2.6.1"
     },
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -167,6 +167,7 @@
     },
     "conflict": {
         "symfony/dependency-injection": "3.4.15|3.4.16",
+        "symfony/monolog-bundle": "3.5.0",
```
TomasLudvik: Well every newer version of monolog-bundle will be in conflict, we should lock version in last 3.4

TomasLudvik: + this branch should be build upon 7.3 as this problem will occur even there

s3tezsky: Maybe they just did something wrong with their dependecies. It does not make any sence to me that monolog v2 requires symfony 5 which is in beta since yeasterday.

s3tezsky: In 7.3 it is not a problem, I guess there are some older dependencies which blocks upgrading.

Anyway I'm ok to fix `symfony/monolog-bundle` to 3.4

PR #1512, opened 2019-11-18
[shopsys] disabled CSRF protection in logout actions

### LINKED ISSUE #769: 403 when trying to logout after my session expires
Reported by @stanoMilan 
### What is happening

Notes: 
* probably connected to #315 

### Expected result

s3tezsky:
| Q             | A
| ------------- | ---
|New feature| No 
|Fixes issues| fixes #769 

### DIFF
--- a/packages/framework/src/Model/Security/AdministratorLoginFacade.php
+++ b/packages/framework/src/Model/Security/AdministratorLoginFacade.php
@@ -102,11 +102,11 @@ class AdministratorLoginFacade
     {
         $token = $this->tokenStorage->getToken();
         if ($token !== null) {
-            /** @var \Shopsys\FrameworkBundle\Model\Administrator\Administrator $currentAdministrator */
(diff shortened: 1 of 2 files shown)


pk16011990:
CSRF protection on logout is for security... Logout action without CSRF can be triggered by attacker. User is forced login again and write user's password.

Please, do not remove this security feature!

s3tezsky:
Actually it has no added value. It may be concerned only as a security through obscurity.

This protection will not prevent any phising (which may be done via poor designed email), man in the middle and sniffing comunitaction in case it will not be running through secured conenction and even any XSS.

henzigo:
@s3tezsky Maybe this page can help you to understand how you can create security issue with this PR ... https://security.stackexchange.com/questions/62769/should-login-and-logout-action-have-csrf-protection

s3tezsky:
I'm closing it as it will be totally user unfriendly and annoying without it.

PR #1529, opened 2019-11-28
[project-base] prevent throwing error 500 after order is sent if SMTP is down

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When application cannot connect to SMTP server, then application will throw error 500 for user. This PR prevents to throw error. Instead of it, user will be notified with flash message that is saying `Unable to send some emails, please contact us for order verification`
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/project-base/src/Shopsys/ShopBundle/Controller/Front/OrderController.php
+++ b/project-base/src/Shopsys/ShopBundle/Controller/Front/OrderController.php
@@ -225,7 +225,7 @@ class OrderController extends FrontBaseController
 
                 try {
                     $this->sendMail($order);
-                } catch (\Shopsys\FrameworkBundle\Model\Mail\Exception\MailException $e) {
+                } catch (\Exception $e) {
                     $this->getFlashMessageSender()->addErrorFlash(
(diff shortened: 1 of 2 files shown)


grossmannmartin:
Hi @henzigo,
thanks for the PR, unfortunately, I was not able to reproduce the behavior this PR should fix.
I tried to set `mailer_host` parameter to some nonsense value and even set `mailer_master_email_address: null` and `mailer_delivery_whitelist: null` to be sure it's not something dev mode related.

Can you please provide some reproducible case, this PR fixes? 

Thanks in advance :)

vitek-rostislav:
Hi @grossmannmartin, when I tested this on our project, I just shut down the `smtp` container using `docker-compose stop smtp-server` :wink:

henzigo:
@grossmannmartin Yes that's my fail. On our project, we have disabled spool so this issue is not relevant for current configuration on SSFW. You can close it.

grossmannmartin:
@henzigo I see... This is an exemplary pull request (description, upgrade notes), I am almost sad it's not merged.

But... closing

PR #1538, opened 2019-12-03
[framework] Full-width layout in administration

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| With new design, there is in administration so much unused space. This can be problem when you want to add some elements for example to orders table. This PR creates full-width administration.
|New feature| Yes 
|Fixes issues| ... 

Some info: 
- The administration is now resizing from 1200px to undefined ☣️ 
- We can add additional columns to orders table without any deformation
- We have space for two-column grid 💯 
- I like it 💪 
- You will like it too

Before:
Snímek obrazovky 2019-12-03 v 17 41 20

After: 
Snímek obrazovky 2019-12-03 v 17 41 39


grossmannmartin:
Thanks for the change!
@TomasGottvald can you please validate and review this PR?

vitek-rostislav:
nice! :ok_hand:

packages/framework/src/Resources/styles/admin/components/box/quick-search.less
```diff
@@ -7,10 +7,14 @@
     font-size: 0;
 
     &__input {
-        vertical-align: middle;
-        width: 89%;
-        height: @input-height;
-        line-height: 30px;
+        display: inline;
+        position: relative;
+        input {
```
TomasGottvald:
```suggestion
        
        input {
```

PR #1541, opened 2019-12-04
[framework] restrict date picker settings per instance 

milanStano:
| Q             | A
| ------------- | ---
|Description| If there are more datepicker on the page, the settings are copied among themselves. For example, I have two datepicker for the first set max date to today and the second not. At this point, the secound datepicker has max date set and should not have.
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| 
No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Resources/scripts/common/components/datePicker.js
+++ b/packages/framework/src/Resources/scripts/common/components/datePicker.js
@@ -5,7 +5,7 @@
     var datePicker = function ($container) {
         $container.filterAllNodes('.js-date-picker').each(function () {
             // Loads regional settings for current locale
-            var options = $.datepicker.regional[global.locale] || $.datepicker.regional[''];
+            var options = $.extend({}, $.datepicker.regional[global.locale] || $.datepicker.regional['']);
 
             // Date format is fixed so that it is understood by back-end
             options.dateFormat = Shopsys.constant('\\Shopsys\\FrameworkBundle\\Form\\DatePickerType::FORMAT_JS');


packages/framework/src/Resources/scripts/common/components/datePicker.js
```diff
@@ -5,7 +5,7 @@
     var datePicker = function ($container) {
         $container.filterAllNodes('.js-date-picker').each(function () {
             // Loads regional settings for current locale
-            var options = $.datepicker.regional[global.locale] || $.datepicker.regional[''];
+            var options = Object.assign({}, $.datepicker.regional[global.locale] || $.datepicker.regional['']);
```
pesektomas: Hello @milanStano, thank you for your PR. I have one little thing. Object.assign is not suporrted in IE. You can look at [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign). You can use [extend](https://api.jquery.com/jquery.extend/) method from jquery. When we will have babel in framework (and I hope that it will be early) we will replace jquery extend as assign.

Could you change it, please?

henzigo: IE is not supported aanymore if im not wrong

pesektomas: IE 11 too?

milanStano: @pesektomas Ok i fixed it :) it is okay ?

pesektomas: A lots of thanks, @stanoMilan. I think that is ok. I have to test it in real use case :smile:

PR #1559, opened 2019-12-20
[project-base] Do not throw exception if CSRF token is expired on logout

### LINKED ISSUE #769: 403 when trying to logout after my session expires

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR fixes problem with CSRF token on logout. Instead of error there will be flash message shown for user.
|New feature| No 
|Fixes issues| closes #769 


sslt:
[[project-base] Do not throw exception if CSRF token is expired on logout](https://trello.com/c/n0VgzJmA/14-project-base-do-not-throw-exception-if-csrf-token-is-expired-on-logout)

TomasLudvik:
Hi, this has been solved in #1909, closing.

PR #1563, opened 2019-12-22
Order item has new type `discount`

### LINKED ISSUE #1263: Promo code is custom instance of OrderItem
This issue was discussed once, in https://github.com/shopsys/shopsys/issues/355, it was closed and today we can create new instance of `OrderItem` with new order item type for promo code. 
### What is happening

Promo codes are still used in SSFW and have type `OrderItem::TYPE_PRODUCT`, it is confusing. 

It is worth to mention one more point:
There is no relation between promo code `OrderItem` and product `OrderItem` to which it belongs to. From my experience with working with promo codes, there should exists such a relation.

Why ?
Many projects, which we implemented, work with discount per product. We should send them product and its discount. Now we have to override working with promo codes from stabil. 
1. add new order item type
2. allow have a relation between order items

one more example from project, we are currently implementing: gifts. 
Some products can have associated gift with special price. So we created new order type, but also we had to define relation between order items (if you remove main order items, its gift shoul be also removed).  

### Expected result
1. there exists new order type item `OrderType::TYPE_PROMO_CODE`
2. there exists relation between order items (promo code discount is related to certain product)

If you are interested, maybe we can provide PR for this issue, from project, where we solved it

### LINKED ISSUE #1390: PromoCodes and OrderRounding are OrderItems with type product
### What is happening

Promo codes and order rounding are OrderItems with type "product" which doesn't really makes sense as they're not products but behave like it.

### Expected result

Maybe it would be better to have new types of order items for promo codes and order rounding so we don't use type product for everything.
I would be glad to hear more ideas how to solve this or if it would make sense to do this at all.

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Until now, all discounts had typed `product`. That was unwanted for a lot of projects because they need to know what is product and what is discount for product. This PR adds new type `discount` to recognition product discount.
|New feature| Yes 
|Fixes issues| closes #1390, closes #1263 

### DIFF
--- a/packages/framework/src/Model/Order/Item/OrderItem.php
+++ b/packages/framework/src/Model/Order/Item/OrderItem.php
@@ -22,7 +22,8 @@ class OrderItem
     public const
         TYPE_PAYMENT = 'payment',
         TYPE_PRODUCT = 'product',
-        TYPE_TRANSPORT = 'transport';
+        TYPE_TRANSPORT = 'transport',
+        TYPE_DISCOUNT = 'discount';
 
     /**
      * @var int|null
@@ -355,7 +356,7 @@ class OrderItem
      */
     public function getProduct(): ?Product
     {
-        $this->checkTypeProduct();
+        $this->checkTypeProductOrDiscount();
         return $this->product;
     }
 
@@ -364,7 +365,7 @@ class OrderItem
      */
     public function hasProduct()
     {
-        $this->checkTypeProduct();
+        $this->checkTypeProductOrDiscount();
         return $this->product !== null;
     }
 
@@ -373,7 +374,7 @@ class OrderItem
      */
     public function setProduct(?Product $product): void
     {
-        $this->checkTypeProduct();
+        $this->checkTypeProductOrDiscount();
 
         if ($product !== null && $product->isMainVariant()) {
             throw new MainVariantCannotBeOrderedException();
@@ -406,6 +407,21 @@ class OrderItem
         return $this->type === self::TYPE_TRANSPORT;
     }
 
+    /**
+     * @return bool
+     */
+    public function isTypeDiscount(): bool
+    {
+        return $this->type === self::TYPE_DISCOUNT;
+    }
+
+    protected function checkTypeDiscount(): void
+    {
+        if (!$this->isTypeDiscount()) {
+            throw new WrongItemTypeException(self::TYPE_DISCOUNT, $this->type);
+        }
+    }
+
     protected function checkTypeTransport(): void
     {
         if (!$this->isTypeTransport()) {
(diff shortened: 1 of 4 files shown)


packages/framework/src/Model/Order/Item/OrderItem.php
```diff
@@ -406,6 +407,21 @@ public function isTypeTransport(): bool
...
+        return $this->type === self::TYPE_DISCOUNT;
+    }
+
+    protected function checkTypeDiscount(): void
+    {
+        if (!$this->isTypeDiscount()) {
+            throw new WrongItemTypeException(self::TYPE_DISCOUNT, $this->type);
+        }
+    }
+
```
PetrHeinz: Unused method
```suggestion
```

packages/framework/src/Model/Order/Item/OrderItem.php
```diff
@@ -420,10 +436,20 @@ protected function checkTypePayment(): void
         }
     }
 
+    /**
+     * @deprecated This method is deprecated since 8.1.0 and will be removed in next major release, use OrderItem::checkTypeProductOrDiscount() instead
```
PetrHeinz: Seems like triggering a user deprecation warning is missing.

PetrHeinz:
Cool @henzigo! Finally! 🎉

paolo42:
The "discount" OrderItem subtype has been missing for as long as I can remember. My childhood will be ended by this pull request :cry:

pesektomas:
Hello @paolo42, so enjoy the last moments of your childhood.

henzigo:
Hello guys, I'm glad that you are interested in this PR, but I'm afraid that this PR will not be merged soon.

pesektomas:
Be patient @henzigo, The Rome was not build in a day.

paolo42:
Let's just be glad that foundation of Rome wasn't managed by Shopsys :hourglass:

pesektomas:
Maybe the Colosseum would have been built before the city :smiling_imp:

CarabineSK:
`getProduct()` contains `$this->checkTypeProductOrDiscount()`
Is this some kind of polymorphism attempt? It sounds wrong.

It all seems like this PR is adding product and discount confusion.

paolo42:
The confusion has been there from the start. This `OrderItem` subtype should've been created as two separate subtypes in the beginning. The purpose of this PR is to get rid of most of the confusion, this is the simplest way to do this.

This `checkTypeProductOrDiscount` method is just backwards compatibility for completed orders - a completed order has items for products that maybe don't exist anymore (deleted, hidden, whatever). This is a place to say that the product was not found and the only thing left is a row in an old order represented by an "empty" `OrderItem` instance. There is no point in being more exact than this, no more value will be gained from a different implementation.

paolo42:
Happy birthday, PR 1563! :birthday:

PR #1572, opened 2020-01-09
[project-base] - lazyload call after cart recalculation, price with VAT table position

### LINKED ISSUE #1570: Images of products disappear after change of product count in cart
### What is happening
When I am in cart and I increase count of products the product images disappear.
This seems to be connected with lazyload of images somehow.

### Expected result
Images should stay

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| After adding lazyload, we need to call lazyload function manually after cart recalculation. Price with VAT position in table was too on right. Now it is correct.
|New feature| No 
|Fixes issues| closes #1570 

Screen before:
https://www.screencast.com/t/M4lCCmm4
obrazek

Screen after:
obrazek

### DIFF
--- a/project-base/src/Resources/scripts/frontend/cart/cartRecalculator.js
+++ b/project-base/src/Resources/scripts/frontend/cart/cartRecalculator.js
@@ -63,7 +63,6 @@
 
                 Shopsys.register.registerNewContent($mainContent);
                 Shopsys.register.registerNewContent($cartBox);
-                Shopsys.lazyLoadCall.inContainer('.js-cart');
             }
         });
     };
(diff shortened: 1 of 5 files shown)


project-base/src/Resources/scripts/frontend/cart/cartRecalculator.js
```diff
@@ -63,6 +63,7 @@
 
                 Shopsys.register.registerNewContent($mainContent);
                 Shopsys.register.registerNewContent($cartBox);
+                Shopsys.lazyLoadCall.inContainer('.js-cart');
```
pesektomas: I think that we can start thinging about use Register.callback in lazyLoadCall component. What do you think about it?

malyMiso: I like the idea @pesektomas, nice point :+1:

TomasGottvald: Discussed with @pesektomas and refactored. All these call were removed, and added register.callback to lazyload function

PR #1574, opened 2020-01-09
[project-base] - fix brand list without panel

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| After removing left panel from frontend, there was bad layout of product list by brand.
|New feature| No 
|Fixes issues| ... 

Before: 
obrazek

After:
obrazek

### DIFF
--- a/project-base/src/Resources/styles/front/common/components/box/list.less
+++ b/project-base/src/Resources/styles/front/common/components/box/list.less
@@ -46,18 +46,28 @@
     }
 
     &__producer {
-        .clearfix();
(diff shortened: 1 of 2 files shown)


TomasGottvald:
> I thing that space between title and products list is so big and it is not pretty to look. Can't you do something with it?

It wasn't our scope, but after consultation I refactored brand list description.

Screen after:
image

Image is on left side, vertically centered. Text on right side is vertically centered too. If image is not available, text wiill appear on 100% width.

PR #1575, opened 2020-01-10
[shopsys] Added npm link to npm phing target

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #1545 we created script for coping common javascripts from packages/framework/assets to project-base/assets/framework. This pull request introduces better approach. Now we can use npm package @shopsys/framework for management common dependencies and  npm link to management common dependecies for development
|New feature| Yes
|Fixes issues| -
|ID|US-7418|


build.xml
```diff
@@ -180,6 +182,36 @@
...
+    <target name="npm" description="Installs node modules.">
+        <exec executable="${path.npm.executable}" dir="${path.root}" logoutput="true" passthru="true" checkreturn="true">
+            <arg value="prune"/>
+            <arg value="--no-audit"/>
+        </exec>
+
+        <exec executable="${path.npm.executable}" dir="${path.root}" logoutput="true" passthru="true" checkreturn="true">
+            <arg value="install"/>
+        </exec>
+
```
grossmannmartin: You can call the parent NPM target, in that case, you don't need to change anything in a project if we make a change in this target in framework's build.xml file. (for example, remove prune :D)

packages/framework/.gitignore
```diff
@@ -2,5 +2,6 @@
 /components
 /vendor
 /composer.lock
-/node_modules
-/assets/js/translations.json
+/assets/node_modules
+/assets/package-lock.json
```
grossmannmartin: What do you think about putting these ignores into the separate .gitignore file into assets folder? the same way as eslintignore file is

packages/framework/assets/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@shopsys/framework",
+  "version": "9.0.0-alpha.1",
```
grossmannmartin: We talked about 9.0.0-dev version so the package will reflect the version of SSFW. Do I assume properly it's not possible and so you decide to use alpha.1 version instead? 
Or can you explain it to me? I realized I don't know much about releasing node modules.

project-base/.eslintignore
```diff
@@ -1,7 +1,3 @@
-/src/Resources/scripts/frontend/plugins
-
-/src/Resources/scripts/styleguide
```
grossmannmartin: Styleguide related files are now also part of a node module?

project-base/assets/js/frontend/components/hoverIntent.js
```diff
@@ -1,7 +1,8 @@
 import Responsive from '../utils/responsive';
 import HoverIntentSetting from './HoverIntentSetting';
 import Register from 'framework/common/utils/register';
-import 'jquery-hoverintent';
+import $ from 'jquery';
```
grossmannmartin: I noticed you used different way to get jQuery in previous files.
`const $ = window.jQuery || global.jQuery || jQuery;` 

Is any difference between these two? Can it be unified?

project-base/package.json
```diff
@@ -35,29 +35,19 @@
...
-        "webpack-notifier": "^1.6.0"
+        "webpack-notifier": "^1.6.0",
+        "pofile": "^1.1.0"
     },
     "dependencies": {
         "@babel/parser": "^7.7.7",
         "@babel/traverse": "^7.7.4",
-        "bazinga-translator": "^2.6.6",
-        "chart.js": "^2.9.3",
+        "@shopsys/framework": "9.0.0-alpha.1",
```
grossmannmartin: 😍

build.xml
pesektomas: Thx,I changed it

packages/framework/.gitignore
pesektomas: Yes,we can:)Fied

packages/framework/assets/package.json
pesektomas: Sorry, my mistake. We can use dev. I fixed it. I think that version system in npm is same as we use yet. https://docs.npmjs.com/about-semantic-versioning

project-base/.eslintignore
pesektomas: Yes. And I think that we dont have to ignore coding standards of styleguide. 

```js
// package.json
"codemirror": "^5.49.2",
```

```js
// webpack.config.js
.addEntry('styleguide', './assets/js/styleguide/styleguide.js')
```

project-base/assets/js/frontend/components/hoverIntent.js
pesektomas: It is my nightmare. When we use hack for using jsformvalidator out of webpack, we have to set jquery as global variable and some jquery plugins have problems with it. And we have our own plugins (To make things worse). I belive that we can unite this when we use form validator in webpack.

But in admin/index.js I removed global jquery. It was my forgot parth of code for testing.

packages/framework/build.xml
```diff
@@ -653,29 +653,28 @@
...
         </exec>
 
         <exec executable="${path.npm.executable}" dir="${path.root}" logoutput="true" passthru="true" checkreturn="true">
             <arg value="run"/>
-            <arg value="copyFw"/>
+            <arg value="build"/>
         </exec>
+    </target>
 
+    <target name="npm-install-dependencies" description="Installs node modules dependencies.">
```
grossmannmartin: _nitpick_
Can be this target used by itself? (eg. without running `trans` and `build`)

If it's unlikely, I suggest hiding this target (argument ` hidden="true"` as it is in phplint for example)

grossmannmartin:
Tests ok, just wondering. When I build project (not from monorepo, but from project-base) I see untracked folder `assets/css/` is it some my mistake, or this happens?

pesektomas:
Hello @grossmannmartin, thx for testing. Assets/css folder is created with install webpack-encore. I did not solve this because I hope we will move all styles to this folder in the next sprint.

packages/framework/build.xml
pesektomas: Ok, fixed

PR #1579, opened 2020-01-10
[framework] Allow to run tests with test database only

### LINKED ISSUE #1039: I want to be able to run all tests/standards without the use of no-test database
### What is happening

I have a problem to run test phing targets with empty db-create database, i needed to modified parameters.yml database_name to shopsys-test to be able to generate error pages and other things like grunt and so on
### Expected result

I want to launch tests without the need to launch db-demo phing target but just test-db-demo phing target. So build test checks will be faster processed.

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The tests are very time consuming because we need to build whole application on CI server. With this upgrade, you will be able to run tests only with test database. That will decrease time-consuming for test stage.
|New feature| No 
|Fixes issues| closes #1039 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,10 @@ install:
     - docker-compose up -d postgres elasticsearch redis selenium-server php-fpm webserver
     - docker-compose exec php-fpm composer install --no-interaction
 
-script:
-    - docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev tests-acceptance
+jobs:
+    include:
+        -   stage: Tests
+            name: Build all
+            script: docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev tests-acceptance
+        -   name: Build tests only
+            script: docker-compose exec php-fpm ./phing test-db-create tests tests-acceptance
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -1010,7 +1010,7 @@
 
     <target name="test-elasticsearch-index-recreate" depends="production-protection,test-elasticsearch-index-delete,test-elasticsearch-index-create" description="Recreates structure for searching via elasticsearch in test environment (deletes existing structure and creates new one)"/>
 
-    <target name="test-error-pages-generate" depends="redis-check" description="Generates error pages displayed in production environment.">
+    <target name="test-error-pages-generate" depends="redis-check" description="Generates error pages for testing environment.">
         <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
             <arg value="${path.bin-console}"/>
             <arg value="shopsys:error-page:generate-all"/>
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -1010,6 +1010,14 @@
 
     <target name="test-elasticsearch-index-recreate" depends="production-protection,test-elasticsearch-index-delete,test-elasticsearch-index-create" description="Recreates structure for searching via elasticsearch in test environment (deletes existing structure and creates new one)"/>
 
+    <target name="test-error-pages-generate" depends="redis-check" description="Generates error pages displayed in production environment.">
+        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+            <arg value="${path.bin-console}"/>
+            <arg value="shopsys:error-page:generate-all"/>
+            <arg value="--env=test"/>
+        </exec>
+    </target>
+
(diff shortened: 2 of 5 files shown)


packages/framework/build.xml
```diff
@@ -1056,7 +1064,7 @@
         </phingcall>
     </target>
 
-    <target name="tests" depends="npm-install-dependencies,error-pages-generate,test-db-demo,test-elasticsearch-index-recreate,test-elasticsearch-export,tests-unit,tests-functional,tests-smoke" description="Runs unit, functional and smoke tests. Builds new test database in the process."/>
+    <target name="tests" depends="npm-install-dependencies,test-db-demo,test-error-pages-generate,test-elasticsearch-index-recreate,test-elasticsearch-export,tests-unit,tests-functional,tests-smoke" description="Runs unit, functional and smoke tests. Builds new test database in the process."/>
```
grossmannmartin, line 1067: When you now run the whole build (`build-demo-dev`), you will end with generated error pages from test DB, or?

TomasLudvik:
Hello @henzigo, do you intend to finish this draft or you no longer need it and it can be closed?

henzigo:
Hello @TomasLudvik, I'm glad that you are asking. I will try to finish this PR by the end of this week. Anyway, this is not what I need to be done, but as a Shopsys Framework developer, you should be interested to do this thing and not only close this.

henzigo:
@TomasLudvik Tests are failing because of some bug in master.

packages/framework/build.xml
henzigo: Each environment has its own folder for storing generated HTML files

TomasLudvik:
Hello @henzigo,

there is so much work to be done that not finished drafts are not in my top priorities and it depends on its authors if they want to finish it or not. In this case I thank you for finishing your work so we can decide in near future in what patch we will include this. 

There is no bug in master - PHPStan released new version that adds new control for its definition in PHPStan, there is already PR #1942 to solve it.

.travis.yml
```diff
@@ -15,5 +15,10 @@ install:
...
 
-script:
-    - docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev tests-acceptance
+jobs:
+    include:
+        -   stage: Tests
+            name: Build all
+            script: docker-compose exec php-fpm ./phing db-create test-db-create build-demo-dev tests-acceptance
+        -   name: Build tests only
+            script: docker-compose exec php-fpm ./phing test-db-create tests tests-acceptance
```
grossmannmartin, line 24: Why do you run tests twice? (part of build-demo-dev build and separately)

packages/framework/build.xml
```diff
@@ -1010,6 +1010,14 @@
 
     <target name="test-elasticsearch-index-recreate" depends="production-protection,test-elasticsearch-index-delete,test-elasticsearch-index-create" description="Recreates structure for searching via elasticsearch in test environment (deletes existing structure and creates new one)"/>
 
+    <target name="test-error-pages-generate" depends="redis-check" description="Generates error pages displayed in production environment.">
```
grossmannmartin, line 1013:
```suggestion
    <target name="test-error-pages-generate" depends="redis-check" description="Generates error pages for testing environment.">
```

project-base/config/paths.yaml
```diff
@@ -1,6 +1,5 @@
 parameters:
     shopsys.content_dir_name: 'content'
-    shopsys.css_version_filepath: '%shopsys.web_dir%/assets/cssVersion'
```
grossmannmartin, line 3: this is not related to a scope of this PR and I suggest to remove the change to avoid confusion

.travis.yml
henzigo: Because I want to test if tests are working without "production" database. But we could build application in first stage and tests in second stage. But this is on you guys.

project-base/config/paths.yaml
henzigo: That's fair point

.travis.yml
grossmannmartin: I understand and I agree with you – this new feature should be tested to ensure this will work in the future.

Do you need to somehow prepare the application before `test-db-create tests tests-acceptance` is run? 
If not, What about splitting the jobs and run standards in the first stage and tests in the second, as you did on the internal GitLab server?
That way it could run parallel and faster, or am I mistaken?

henzigo:
It's already implemented.

PR #1585, opened 2020-01-14
[framework] remove route prefix (app_) for project-base

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| routes do not have prefix `shopsys_shop_` yet from using flex. It caused duplication of routes after override controller without overriden action (one with prefix `app_admin_` and another one only with prefix `admin_`). It could cause problems with breadcumb etc... Another problem is all new routes in administration have annoying and inconsistent prefix `app_`
|New feature| No
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| No
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1586, opened 2020-01-14
[framework] admin environment is now correctly detected for project-base admin routes

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Administration/AdministrationFacade.php
+++ b/packages/framework/src/Model/Administration/AdministrationFacade.php
@@ -26,6 +26,6 @@ class AdministrationFacade
      */
     public function isInAdmin(): bool
     {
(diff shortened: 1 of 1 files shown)


henzigo:
Hello, this is not the best solution to this problem. After introducing flex, all overridden routes from admin now get prefix `app_`. This PR only fixes that SSFW will correctly detect if route is from admin, but there is a more problematic situation. You need to override all links (templates, controllers) to your overridden routes because your new route will be prefixed with `app_`. 

There is a solution in PR #1585, but after it will be merged then your solution will be useless.

I suggest to close this PR and devote energy to PR from @pk16011990.

PR #1588, opened 2020-01-15
[project-base] added missing symfony/stopwatch dependency

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #1492 `symfony/symfony` dependency was removed and replaced with individual packages. Unfortunatelly `symfony/stopwatch` was forgotten in project composer file causing installation of packages without dev dependencies fails on non-existing classes. This PR fixes it.
|New feature| No 
|Fixes issues| ... 

Upgrade notes not written as previous change is not part of any released version.

### DIFF
--- a/packages/framework/composer.json
+++ b/packages/framework/composer.json
@@ -93,6 +93,7 @@
         "symfony/property-info": "^3.4",
         "symfony/proxy-manager-bridge": "^3.4",
         "symfony/security": "^3.4",
+        "symfony/stopwatch": "^3.4",
         "symfony/swiftmailer-bundle": "^3.2.2",
         "symfony/var-dumper": "^3.4",
         "symfony/web-server-bundle": "^3.4",


s3tezsky:
I don't know about any usage of `Stopwatch` so we should not require it by ourselves. Is not it a problem in some package?

grossmannmartin:
> I don't know about any usage of `Stopwatch` so we should not require it by ourselves. Is not it a problem in some package?

Stopwatch is used in `TraceableEventDispatcher`. Component (`symfony/event-dispatcher`)  have the dependency in their dev requirements. 

In framework's `services.yml` file we define service TraceableEventDispatcher, so we need to have this dependency installed too. 

But thanks to this question I now wonder if `symfony/stopwatch` dependency should be in framework's `composer.json` instead of project-base. What do you think @pesektomas, @s3tezsky ?

pesektomas:
Hello @grossmannmartin,

At first: @s3tezsky has right and I think that is is mistake of EventDispatcher component that it not have StopWatch dependency in require (only in require-dev).

And propably yes. We can have stopwatch dependency in framework's composer.json.

grossmannmartin:
From the [documentation](https://symfony.com/doc/current/components/event_dispatcher/traceable_dispatcher.html) I think the `TraceableEventDispatcher` is meant for debugging purposes only (namespaces suggest this too). In that case, having it in dev is completely ok. 

We use this dispatcher directly in `Authenticator` class

```yaml
 Shopsys\FrameworkBundle\Model\Security\Authenticator:
    arguments:
        - '@security.token_storage'
        - '@Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher'

Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher:
    arguments:
        - '@event_dispatcher'
```

so to create an instance of `TraceableEventDispatcher` we need `Stopwatch`. We may consider rewriting the Authenticator class (and tests), though.


I moved the dependency into framework's `composer.yml` file as it makes more sense there and no upgrade notes are needed.

s3tezsky:
I aggree it is  framework part and it should be there. But I still do not like the whole concept.

> From the [documentation](https://symfony.com/doc/current/components/event_dispatcher/traceable_dispatcher.html) I think the `TraceableEventDispatcher` is meant for debugging purposes only (namespaces suggest this too). In that case, having it in dev is completely ok.

In that case I thing we should not use it as a dependency of `Shopsys\FrameworkBundle\Model\Security\Authenticator`. IMHO `EventDispatcher` should do just fine in production environment.

If there is any benefit of using `TraceableEventDispatcher` we will use it in `services_test` and/or `services_dev`.

---

I know it is not about adding a missing dependency any more but it seems as the best solution here. As you wrote `stopwatch` is a dev dependency of `event-dispatcher` and in that way it should be used only for development. Not as a core of application.

s3tezsky:
I will be OK with merging it as is. Just wanted to point at the `Authenticator` dependency as it is not needed for me (at least not in production) and ask if we wanted to have it solved.

grossmannmartin:
I created issue #1597 so the Authenticator/TraceableEventDispatcher is not forgotten.

henzigo:
😔

PR #1603, opened 2020-01-20
Change access modifier in AdvancedSearchOperatorTranslation

LuKillman:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Inability to add a new operator among existing ones.
|New feature| No
|Fixes issues| 

### DIFF
--- a/packages/framework/src/Form/Admin/AdvancedSearch/AdvancedSearchOperatorTranslation.php
+++ b/packages/framework/src/Form/Admin/AdvancedSearch/AdvancedSearchOperatorTranslation.php
@@ -9,7 +9,7 @@ class AdvancedSearchOperatorTranslation
     /**
      * @var string[]
      */
-    private $operatorsTranslations;
+    protected $operatorsTranslations;
 
     public function __construct()
     {


s3tezsky:
Hi @LuKillman, thanks for PR! The change seems legit since you cannot add/get/remove any item.

Do not you thing that implementing serving methods will do just fine without direct access to property?

LuKillman:
@s3tezsky, I know it's possible to do it that way but it feels different than usual approach when extending framework classes.

s3tezsky:
> @s3tezsky, I know it's possible to do it that way but it feels different than usual approach when extending framework classes.

In case you want to add a new operator it seems to me as a better option to implement some method for addinng which you can use e.g. in child's constructor.
When we change visibility of the property even as protected it could tie our hands in future development because of out BC promisse.

pk16011990:
I disagree with you, @s3tezsky. It is very tricky, because major part of SSFW use protected for extendability, and there developers try use private attribute without declaration in child class -> It creates dynamicly public attribut with same name and it does not work. 

True reason is wrong configuration of `ForbiddenPrivateVisibilityFixer`. See #1650 to complex solution

s3tezsky:
Hi @pk16011990  thanks for clarification. I'm closing this one (sorry @LuKillman) and we will continue in your PR.

PR #1619, opened 2020-01-26
[framework] production protection phing target now returns 1 on error

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Production protection phing target now not return 0 as the return code when user chooses to not proceed.
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| maybe (when relying on the return code 0 on fail) 
|Fixes issues| ... 

The default value is `n` (do not proceed) and when some "dangerous" target is run on the CI server (ie. Jenkins), the console output says "FAIL", but job result is a success (as the job/phing target exited with 0). This is very confusing and can lead to false assumptions that my tests pass while they weren't actually even executed.

### DIFF
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -810,7 +810,7 @@
                 <if>
                     <not><equals arg1="${production.confirm.action}" arg2="y"/></not>
                     <then>
-                        <fail message="Terminated by user" status="0"/>
+                        <fail message="Terminated by user"/>
                     </then>
                 </if>
             </then>


TomasLudvik:
Hi @grossmannmartin,

I think that build should end as successful when the user decided that it should not be run.

grossmannmartin:
> I think that build should end as successful when the user decided that it should not be run.

Right now it's really confusing. 
Locally – choosing not to continue, end up with big red message "BUILD FAILED", but return code says everything is OK.
On CI – when I run build in production environment, job on CI ends green (successful) and I'm happy. But after close inspection of output, I can see the build was not executed at all. That way I can easily deploy defect version of my application.

henzigo:
Hello, right now our deploy to production just passed even when all of our phing targets didn't run because of `production-protection` command. 

That can result in large issue when not completed application is deployed

PR #1602, opened 2020-01-18
Upgrade to Elasticsearch 7

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| As SSFW developer I need Elasticsearch updated to current version. Also developers need Kibana for local development.
|New feature| Yes
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

I didn't update UPGRADE.md

Issue #1287, opened 2019-08-04
How to create and maintain custom Elasticsearch indexes?


pk16011990:
### What is happening
I need to create new index in Elasticsearch for my project. SSFW has prepared some methods for creating/migrating/deleting indexes, however there is only command `product-search-migrate-structure`. 

1. Name of this command is wrong. The command doesn't apply to searching only

2. When I want another index rather than products in Elasticsearch, I can not use command `product-search-migrate-structure`. It does not support neither easy extension nor automatic loading of all definitions in `project-base/src/Shopsys/ShopBundle/Resources/definition/*/*.json`

Automatic loading would have a problem with non-multidomain indexes... :-/

### Expected result
Existing target for all Elasticsearch structure migrations (like DB migrations)

pk16011990:
.

grossmannmartin:
resolved with #1622 
Also, you can check new cookbook (https://docs.shopsys.com/en/latest/cookbook/adding-a-new-elasticsearch-index/) how to create a new elasticsearch index

Issue #1286, opened 2019-08-04
Unfriendly filepath for Elasticsearch configuration


pk16011990:
### What is happening
Elasticsearch indexes configuration is in json files with filepath `project-base/src/Shopsys/ShopBundle/Resources/definition/product/{domainId}.json`.
It is not intuitive, I dont know what is number (domain id) and for what is this definition (Elasticsearch). 

Next problem use-case is finding of this file in IDE... I search file with name `1.json`... or `definition/product/1.json`. That is not good DX.

### Expected result
Filepaths of Elasticsearch definition files say `hey, I'm for Elasticsearch` and magic numbers have prefix like `domain` or somethink like that.

vitek-rostislav:
Hi @pk16011990, thanks for the feedback, it definitely makes sense to me. Do you have any particular suggestion for the filepath? What about simply renaming the `definition` folder to `elasticsearch-index-definition`? I do agree with renaming the files using `domain_` prefix :+1:

pk16011990:
It is not stale, filepaths are still meaningless...

PR #1622, opened 2020-01-27
[shopsys] easier elasticsearch extensibility

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Current elastic operations are too tight to product search. The goal of this PR is to make commands (create, export, migrate, delete) and cron jobs (export) more abstract to be easily extended for other usage without duplicating the logic.
|New feature| No 
|Fixes issues| ... 


packages/framework/src/Component/Elasticsearch/AbstractIndex.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Elasticsearch;
+
+abstract class AbstractIndex
+{
+    protected const BATCH_SIZE = 100;
```
grossmannmartin: Constant BATCH_SIZE is here and in DataProviderInterface. It's confusing to me. Can you explain the reason, please?

packages/framework/src/Component/Elasticsearch/Exception/ElasticsearchIndexException.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * @param string $definitionFilename
+     * @return self
+     */
+    public static function cantReadDefinitionFile(string $definitionFilename): self
+    {
+        return new self(sprintf(
+            'Can\'t read definition file at path %s. Please check that file exists and has permissions for reading.',
```
grossmannmartin: _nitpick_ 🤓 
Variable is named `definitionFilename`, but here it's referred as the path. It's confusing to me

packages/framework/src/Component/Elasticsearch/Exception/ElasticsearchIndexException.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * @param string $indexName
+     * @param array $error
+     * @return self
+     */
+    public static function createIndexError(string $indexName, array $error): self
+    {
+        return new self(sprintf(
+            'Error when creating index %s:' . PHP_EOL . '%s',
```
grossmannmartin: I suggest putting index name into quotation marks in the whole file, so it's easier to spot spaces or any other unwanted characters.

(like `'Error when creating index "%s":' . PHP_EOL . '%s',`)

packages/framework/src/Component/Elasticsearch/Exception/ElasticsearchIndexException.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * @param string $alias
+     * @return static
+     */
+    public static function aliasDoesntExists(string $alias): self
+    {
+        return new self(sprintf(
+            'Can\' found any index with alias %s.',
```
grossmannmartin:
```suggestion
            'Can\'t found any index with alias "%s".',
```

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
```diff
@@ -0,0 +1,132 @@
...
+    protected $indexPrefix;
+
+    /**
+     * @var int
+     */
+    protected $domainId;
+
+    /**
+     * DocumentDefinition constructor.
+     *
```
grossmannmartin: This looks like something forgotten

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
```diff
@@ -0,0 +1,132 @@
+<?php
+
```
grossmannmartin:
```suggestion

declare(strict_types=1);

```

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
```diff
@@ -0,0 +1,132 @@
...
+     */
+    public function getIndex(): AbstractIndex
+    {
+        return $this->index;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDefinitionsDirectory(): string
```
grossmannmartin: Is this getter even necessary as it's only used in getDefinitionFilename?

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
```diff
@@ -0,0 +1,132 @@
...
+            throw ElasticsearchIndexException::cantReadDefinitionFile($definitionFilename);
+        }
+
+        return file_get_contents($definitionFilename);
+    }
+
+    /**
+     * @return string
+     */
+    public function getDocumentDefinitionVersion(): string
```
grossmannmartin: Should this method be public? Is it desirable to use it outside the class?

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
```diff
@@ -0,0 +1,132 @@
...
+            );
+        }
+
+        return $decodedDefinition;
+    }
+
+    /**
+     * @return string
+     */
+    protected function getDefinitionFilename(): string
```
grossmannmartin: _nitpick_ 🤓 
Seems more like Filepath to me

packages/framework/src/Component/Elasticsearch/IndexManager.php
```diff
@@ -0,0 +1,260 @@
+<?php
+
```
grossmannmartin:
```suggestion

declare(strict_types=1);

```

packages/framework/src/Component/Elasticsearch/IndexRegistry.php
```diff
@@ -0,0 +1,72 @@
...
+     * @var \Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractIndex[]
+     */
+    protected $registeredIndexes;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractIndex[] $indexes
+     */
+    public function __construct(array $indexes)
+    {
+        foreach ($indexes as $index) {
```
grossmannmartin: You can use here a check of passed objects in array, like `Assert::allIsInstanceOf($indexes, AbstractIndex::class);`
to be more defensive.

What do you think?

packages/framework/src/Model/Product/Search/Export/ProductSearchExportWithFilterRepository.php
```diff
@@ -83,50 +83,28 @@ public function __construct(
...
 
         $results = [];
         /** @var \Shopsys\FrameworkBundle\Model\Product\Product $product */
         foreach ($query->getResult() as $product) {
-            $results[] = $this->extractResult($product, $domainId, $locale);
+            $results[$product->getId()] = $this->extractResult($product, $domainId, $locale);
         }
 
         return $results;
     }
```
grossmannmartin: maybe it's time to rename `ProductSearchExportWithFilter` to something better, what do you think?

packages/framework/src/Model/Product/Search/Export/ProductSearchExportScheduler.php
```diff
@@ -4,36 +4,8 @@
...
-    public function getProductIdsForImmediateExport(): array
-    {
-        return array_unique($this->productIds);
-    }
-
-    public function cleanScheduleForImmediateExport(): void
-    {
-        $this->productIds = [];
-    }
+class ProductSearchExportScheduler extends AbstractExportScheduler
```
grossmannmartin: What about to change the name (everywhere) to something without the 'Search' as ES is used for many more things now. What do you think?

packages/framework/src/Model/Product/Elasticsearch/ProductIndex.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Elasticsearch;
+
+use Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractIndex;
+
+class ProductIndex extends AbstractIndex
+{
+    public const INDEX_NAME = 'product';
```
grossmannmartin: as the getter exists, is it necessary to have the constant public?

packages/framework/src/Command/Elasticsearch/AbstractElasticsearchIndexCommand.php
```diff
@@ -0,0 +1,131 @@
...
+        IndexRegistry $indexRegistry,
+        IndexFacade $indexFacade,
+        IndexDefinitionLoader $indexDefinitionLoader,
+        Domain $domain
+    ) {
+        $this->indexRegistry = $indexRegistry;
+        $this->indexFacade = $indexFacade;
+        $this->indexDefinitionLoader = $indexDefinitionLoader;
+        $this->domain = $domain;
+        parent::__construct();
```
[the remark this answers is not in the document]
s3tezsky: Ther is no reason. It should be called as first.

packages/framework/src/Command/Elasticsearch/AbstractElasticsearchIndexCommand.php
```diff
@@ -0,0 +1,131 @@
...
+        foreach ($this->getAffectedIndexes($indexName) as $index) {
+            foreach ($this->domain->getAll() as $domainConfig) {
+                $this->executeCommand(
+                    $this->indexDefinitionLoader->getIndexDefinition($index, $domainConfig->getId()),
+                    $output
+                );
+            }
+        }
+
+        $symfonyStyleIo->success($this->getActionFinishedMessage());
```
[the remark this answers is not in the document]
s3tezsky: Agree. We should not write deprecated code.

packages/framework/src/Command/Elasticsearch/ElasticsearchIndexesCreateCommand.php
```diff
@@ -0,0 +1,60 @@
...
+        return 'Creates structure in Elasticsearch';
+    }
+
+    /**
+     * @return string
+     */
+    protected function getArgumentNameDescription(): string
+    {
+        return sprintf(
+            'Which index will be created? Available indexes: "%s"',
```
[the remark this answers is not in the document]
s3tezsky: This is a good point. It seems a bit confusing.

I am not really sure which term to use - maybe keep `index` as it is corrent name in elastic search terminology. Even I do not really like it as it is too common word.

packages/framework/src/Command/Elasticsearch/ElasticsearchIndexesDeleteCommand.php
```diff
@@ -0,0 +1,60 @@
...
+        return 'Creates structure in Elasticsearch';
+    }
+
+    /**
+     * @return string
+     */
+    protected function getArgumentNameDescription(): string
+    {
+        return sprintf(
+            'Which index will be created? Available indexes: "%s"',
```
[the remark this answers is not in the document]
s3tezsky: Thanks!

packages/framework/src/Command/Elasticsearch/ElasticsearchIndexesMigrateCommand.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->indexFacade->migrateByIndexDefinition($indexDefinition, $output);
+    }
+
+    /**
+     * @return string
+     */
+    protected function getCommandDescription(): string
+    {
+        return 'Migrate indexes in Elasticsearch';
```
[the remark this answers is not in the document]
s3tezsky: The behavior is kept right in `IndexFacade` https://github.com/shopsys/shopsys/blob/4b92eb3fdc5ae15cdfbed7e2967d758b93fad29e/packages/framework/src/Component/Elasticsearch/IndexFacade.php#L81

packages/framework/src/Component/Elasticsearch/AbstractIndex.php
s3tezsky: This one stayed forgotten here. It is not used anywhere. The right place I want to keep it is `DataProviderInterface`.

I am removing it from here.

packages/framework/src/Component/Elasticsearch/Exception/ElasticsearchIndexException.php
s3tezsky: renamed to `$definitionFilePath` also with other related wrong namings.

packages/framework/src/Component/Elasticsearch/Exception/ElasticsearchIndexException.php
s3tezsky: Nice catch. Totally agree.

packages/framework/src/Component/Elasticsearch/Exception/ElasticsearchIndexException.php
s3tezsky: Fixed with others.

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
s3tezsky: Fixed here and in my IDE!

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
s3tezsky: I personally do not like accessing properties directly. Even it seems unnecessary I would like keep it here. It may help us one day when we will want to change where documents are stored (for any reason) without affecting the property.

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
s3tezsky: Fixed. It should not be used elsewhere.

packages/framework/src/Component/Elasticsearch/IndexRegistry.php
s3tezsky: Interesting... Can find out if it is more paranoia or is the check really reasonable.

As this is by desing just kind of user input we could check it to stop it ASAP. If we are planning to autowire all children of `AbstractIndex` I would not check it here again.

s3tezsky: After discussion we decided to autowire Index classes automatically. 

This check is not necessary.

packages/framework/src/Command/Elasticsearch/AbstractElasticsearchIndexCommand.php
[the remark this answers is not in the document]
s3tezsky: There is reason. We need to assign services before `::configure()` is called - which is called via parent constructor.

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
s3tezsky: After discussion we decided to access property directly inside its class and remove the getter.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportWithFilterRepository.php
s3tezsky: It is not necessary to solve this now.

packages/framework/src/Model/Product/Elasticsearch/ProductIndex.php
s3tezsky: Make sense.

packages/framework/src/Component/Elasticsearch/AbstractExportListener.php
```diff
@@ -0,0 +1,91 @@
...
+
+namespace Shopsys\FrameworkBundle\Component\Elasticsearch;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Model\Product\Elasticsearch\ProductIndex;
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+
+abstract class AbstractExportListener
```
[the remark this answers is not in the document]
s3tezsky: Good point, now it is switched to using subscriber.

packages/framework/src/Model/Product/Search/Export/ProductSearchExportScheduler.php
s3tezsky: We can change it to `ExportScheduler` if you think it is reasonable. Also with others like:

- `ProductSearchExportCronModule` -> `ExportCronModule`
- `ProductSearchExportSubscriber` -> `ExportSubscriber`
- `ProductSearchExportWithFilterRepository` -> `ExportWithFilterRepository`

---
I do not like the names too but I dont think it is wise to change it just for a better looking and no added value.

s3tezsky: If we decide to rename it anyway I would suggest to change the namespace too. From `Shopsys\FrameworkBundle\Model\Product\Search\Export` to `Shopsys\FrameworkBundle\Model\Product\Elasticsearch`

packages/framework/build.xml
```diff
@@ -735,28 +735,32 @@
     <target name="product-search-create-structure" description="Creates structure for searching via elasticsearch." hidden="true">
         <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
             <arg value="${path.bin-console}"/>
-            <arg value="shopsys:product-search:create-structure"/>
+            <arg value="shopsys:elasticsearch:indexes-create"/>
+            <arg value="product"/>
```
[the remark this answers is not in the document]
s3tezsky: I have deprecated the old phing targets and created a new ones. Which can handle all indexes or just one specific via property.

packages/framework/src/Component/Elasticsearch/IndexRepository.php
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\IndexDefinition $indexDefinition
+     * @param array $restrictToIds
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     */
+    public function export(IndexDefinition $indexDefinition, array $restrictToIds, OutputInterface $output): void
+    {
+        $indexAlias = $indexDefinition->getIndexAlias();
```
grossmannmartin: I noticed you removed disabled SQL Logging (`SqlLoggerFacade`).
This results in high memory usage on a larger data set
Snímek obrazovky 2020-02-07 v 16 26 45
.

packages/framework/src/Component/Elasticsearch/IndexRepository.php
```diff
@@ -0,0 +1,262 @@
...
+        } else {
+            $totalCount = $index->getTotalCount($indexDefinition->getDomainId());
+        }
+
+        $progressBar = $this->progressBarFactory->create($output, $totalCount);
+
+        $lastProcessedId = 0;
+        $updatedIds = [];
+        do {
+            $currentBatchData = $index->getExportData($domainId, $lastProcessedId, $restrictToIds);
```
grossmannmartin: I was a bit lost, on this method. it exports data, but the last two arguments are mutually exclusive (= it doesn't make sense to set them both).
I'm thinking about split the whole export method into two separate (batch and ids). What do you think? 
I believe it could help with readability.

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * @var string
+     */
+    protected $indexPrefix;
+
+    /**
+     * @var int
+     */
+    protected $domainId;
```
pk16011990: in my opinion Index cannot be depended on domain. ... I'm just saying...

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
```diff
@@ -0,0 +1,124 @@
...
+
+        return $decodedDefinition;
+    }
+
+    /**
+     * @return string
+     */
+    protected function getDefinitionFilepath(): string
+    {
+        return $this->definitionsDirectory . $this->getIndex()->getName() . '/' . $this->getDomainId() . '.json';
```
pk16011990: filename 1.json etc is very bad to searching.... Imagine all entities called `Entity` and only namespace says what is it...

It haas very bad DX... see #1286

packages/framework/src/Model/Product/Elasticsearch/ExportCronModule.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Shopsys\FrameworkBundle\Model\Product\Elasticsearch;
+
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractExportCronModule;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexDefinitionLoader;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexFacade;
+
+class ExportCronModule extends AbstractExportCronModule
```
pk16011990: class name `ExportCronModule` is too general in my point of view. It is like rename `ProductFacade` to `Facade`... Namespace for specification is too small...

packages/framework/src/Model/Product/Elasticsearch/ExportSubscriber.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Shopsys\FrameworkBundle\Model\Product\Elasticsearch;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractExportSubscriber;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexDefinitionLoader;
+use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexRepository;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+class ExportSubscriber extends AbstractExportSubscriber
```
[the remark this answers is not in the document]
pk16011990: I totally agree. It is problem in more classes in this PR

pk16011990:
I suggest to finish #1602 first. I caused better testing of new version of Elasticsearch.

henzigo:
Hello, will be #1287 fixed by this PR?

packages/framework/src/Model/Product/Elasticsearch/ExportSubscriber.php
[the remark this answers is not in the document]
s3tezsky: You are right. I tried to be minimalistic too much here.

packages/framework/build.xml
```diff
@@ -541,6 +541,44 @@
...
+    <target name="elasticsearch-index-export" description="Exports indexes data into elasticsearch.">
+        <property name="elasticsearch.index" value=""/>
+        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+            <arg value="${path.bin-console}"/>
+            <arg value="shopsys:elasticsearch:data-export"/>
+            <arg value="${elasticsearch.index}"/>
+        </exec>
+    </target>
+
+    <target name="elasticsearch-index-migrate" description="Creates new indexes, reindex them from old one, deletes old indexes and add alias to the new one." hidden="true">
```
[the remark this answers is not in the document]
s3tezsky: Thanks for noticing this one 👍 ! I have missed that. Maybe we should not use too long lines to prevent these.

packages/framework/build.xml
```diff
@@ -541,6 +541,44 @@
...
+    <target name="elasticsearch-index-delete" depends="production-protection" description="Deletes indexes into elasticsearch." hidden="true">
+        <property name="elasticsearch.index" value=""/>
+        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+            <arg value="${path.bin-console}"/>
+            <arg value="shopsys:elasticsearch:indexes-delete"/>
+            <arg value="${elasticsearch.index}"/>
+        </exec>
+    </target>
+
+    <target name="elasticsearch-index-export" description="Exports indexes data into elasticsearch.">
```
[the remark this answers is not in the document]
s3tezsky: I agree. I will use `elasticsearch-export`.

packages/framework/build.xml
```diff
@@ -936,47 +979,83 @@
...
         <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
             <arg value="${path.bin-console}"/>
             <arg value="--env=test"/>
-            <arg value="shopsys:product-search:export-products"/>
+            <arg value="shopsys:plugin-data-fixtures:load"/>
         </exec>
     </target>
 
-    <target name="test-product-search-recreate-structure" depends="production-protection,test-product-search-delete-structure,test-product-search-create-structure" description="Recreates structure for searching via elasticsearch in test environment (deletes existing structure and creates new one)"/>
+    <target name="elasticsearch-index-recreate" depends="production-protection, test-elasticsearch-index-delete, test-elasticsearch-index-create" description="Recreates indexes into elasticsearch (deletes existing indexes and creates new one)"/>
```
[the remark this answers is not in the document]
s3tezsky: What an invader! Thanks for noticing 👍

packages/framework/src/Resources/config/services.yml
```diff
@@ -1010,11 +1009,20 @@ services:
             $definitionDirectory: '%shopsys.elasticsearch.structure_dir%'
             $indexPrefix: '%env(ELASTIC_SEARCH_INDEX_PREFIX)%'
 
-    Shopsys\FrameworkBundle\Model\Product\Search\Export\ProductSearchExporter: ~
+    Shopsys\FrameworkBundle\Model\Product\Elasticsearch\ProductIndex: ~
```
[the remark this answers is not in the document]
s3tezsky: Sadly it is due to used dependencies in other classes. I may refactor them to remove this class as the only thing what is used is `getAliasName()` which is now reduntant to new implementation.

packages/framework/src/Resources/config/services_test.yml
```diff
@@ -225,3 +219,7 @@ services:
             $indexPrefix: '%env(ELASTIC_SEARCH_INDEX_PREFIX)%'
```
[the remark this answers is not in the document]
s3tezsky: Actually it exists. But I do not want to see it existing any longer. (same as a comment above)

packages/framework/tests/Unit/Model/Elasticsearch/__fixtures/CategoryIndex.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Tests\FrameworkBundle\Unit\Model\Elasticsearch\__fixtures;
+
+use Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractIndex;
+use Symplify\BetterPhpDocParser\Exception\NotImplementedYetException;
```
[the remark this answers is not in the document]
s3tezsky: Not really. Replaced via `Symfony\Component\Intl\Exception\MethodNotImplementedException`

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
s3tezsky: Actually it is. Each index is created per domain by design. That is because you want to have different configuration according to the language. I know domain != language, but it is the way we work with it and we are not changing it in this PR. If you have any improvement tips please let us know we want to hear it.

packages/framework/src/Component/Elasticsearch/IndexDefinition.php
s3tezsky: I udestand your trouble and agree with you. We are not planning to change it in this PR as it is not a part of this change. I have looked into the linked PR and I am missing any suggestion from you how it could be done better for you.

When we come up with a better solution it is not hard to change later.

s3tezsky:
> I suggest to finish #1602 first. I caused better testing of new version of Elasticsearch.

@pk16011990  This will not be possible due to our priority. I am not expecting any problem if we get merged this PR first.

> Hello, will be #1287 fixed by this PR?

@henzigo  Yes, it will. There will be new phing targets `elasticsearch-index-(create|migrate|delete)` and `elasticsearch-data-export` which will do the action for all indexes. Also it will be possible to use it for a single index with optional parameter.

packages/framework/build.xml
```diff
@@ -748,34 +786,39 @@
...
+            <property name="elasticsearch.index" value="product"/>
+        </phingcall>
     </target>
 
-    <target name="product-search-migrate-structure" description="Creates new structure, reindexes it from old one, deletes old structure and add alias to new structure.">
-        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
-            <arg value="${path.bin-console}"/>
-            <arg value="shopsys:product-search:migrate-if-necessary"/>
-        </exec>
+    <target name="product-search-migrate-structure" description="Creates new structure, reindexes it from old one, deletes old structure and add alias to new structure." hidden="true">
```
grossmannmartin: As all `product-search-` targets are now deprecated, it would be nice to change dependencies in other targets. Currently running tests, build, or so results in warnings in the console about the deprecated targets and the user have no way to fix it

project-base/config/parameters_test.yml.dist
```diff
@@ -14,3 +14,5 @@ parameters:
     test_mailer_password: ~
 
     shopsys.content_dir_name: 'content-test'
+
+    env(ELASTIC_SEARCH_INDEX_PREFIX): 'test'
```
grossmannmartin: Unfortunately, this does not work as a prefix... 
when I set environment variable ELASTIC_SEARCH_INDEX_PREFIX (in docker-compose file for example), then running tests will create index with the same name as production/development environment does

packages/framework/src/Model/Product/Search/ProductElasticsearchRepository.php
```diff
@@ -6,16 +6,15 @@
...
 class ProductElasticsearchRepository
 {
-    public const ELASTICSEARCH_INDEX = 'product';
-
     /**
-     * @var string
+     * @deprecated will be removed in the next major release
      */
-    protected $indexPrefix;
+    public const ELASTICSEARCH_INDEX = 'product';
```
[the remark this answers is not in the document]
s3tezsky: When anyone is using this constant somewehere else (as it is a public) we may leave it still available for a while. As this is only one deprecation I created it is not worthy anyway.

Maybe we should think about it in feature.

project-base/config/parameters_test.yml.dist
s3tezsky: This is really weird. This line is specifying the default value which might be overriden by value from ENV (docker-compose.yml).

I do not know about aby other solution except use for testing purposes second ENV (e.g. `TEST_ELASTIC_SEARCH_INDEX_PREFIX`). What do you think?

packages/framework/src/Component/Elasticsearch/IndexRepository.php
```diff
@@ -0,0 +1,329 @@
...
+
+        if (isset($result['error'])) {
+            throw ElasticsearchIndexException::deleteIndexError($indexName, $result['error']);
+        }
+    }
+
+    /**
+     * @param string $indexAlias
+     * @param int $domainId
+     * @param array $ids
```
grossmannmartin:
```suggestion
     * @param int[] $ids
```

packages/framework/src/Component/Elasticsearch/IndexRepository.php
```diff
@@ -0,0 +1,329 @@
...
+
+        $this->sqlLoggerFacade->reenableLogging();
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractIndex $index
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\IndexDefinition $indexDefinition
+     * @param int[] $restrictToIds
+     */
+    public function exportIds(AbstractIndex $index, IndexDefinition $indexDefinition, array $restrictToIds): void
```
grossmannmartin: I tested the following scenario
- all products are exported in ES
- I go to product detail and remove all category assignments on the first domain and save product

Expected result: The product is no longer visible in product lists on the first domain.
Actual result: Product is visible in list and product detail returns 404

In the previous implementation was some condition to remove ids from elastic if they're not found with detExportData method.

project-base/config/parameters_test.yml.dist
grossmannmartin: it's moved from parameters into services

packages/framework/src/Model/Product/Elasticsearch/ExportCronModule.php
grossmannmartin: Renamed to `ProductExportCronModule` (and other similar classes were renamed too)

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
++
+    public function getName(): string
+    {
+-       // TODO: Implement getName() method.
++       return static::INDEX_NAME;
+    }
+```
+
+So far it is the most minimalistic implementation.
+Now we are able to create an index in elasticsearch by running `./phing elasticsearch-index-create -D elasticsearc.index=category`.
```
grossmannmartin: In this point, I get the following error:
```
There is no index "category" registered (uncaught exception) at
packages/framework/src/Component/Elasticsearch/Exception/ElasticsearchIndexException.php line 97 
while running console command `shopsys:elasticsearch:indexes-create`
```

and I had to manually tag the service with elasticsearch.index....
```
   App\Model\Category\Elasticsearch\CategoryIndex:
        tags: ['elasticsearch.index']
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
++           'description' => $category->getDescription($domainId),
++           'parentId' => $category->getParent()->getId(),
++           'level' => $category->getLevel(),
++           'uuid' => $category->getUuid(),
++       ];
++   }
+```
+
+!!! note
+The `getExportDataForBatch()` must return serialized array of rows indexed by its ID
```
grossmannmartin: admonitions have to be indented

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+
+Now we can export categories data (name, description, parentId, level, and uuid) into elasticsearch index via `./phing elasticsearch-export -D elaticsearch.index=category` (we need to have created index first, see the step above).
```
grossmannmartin:
```suggestion
Now we can export categories data (name, description, parentId, level, and uuid) into elasticsearch with `./phing elasticsearch-export -D elasticsearch.index=category` (index have to be created first, see the step above).
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Exporting via cron
+
+We may automate the export process via CronModule which is super easy.
```
grossmannmartin:
```suggestion
We may automate the export process thanks to the `CronModule` which is super easy.
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+All we need to achieve this goal is to create a new class `CategoryExportCronModule` which extends `AbstractExportCronModule`.
```
grossmannmartin:
```suggestion
All we need to do to achieve this goal is to create a new class `CategoryExportCronModule` which extends `AbstractExportCronModule`.
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+
+The most important task here is to override parent constructor with changing a type hint of the first argument to our created index (`CategoryIndex`).
```
grossmannmartin:
```suggestion
The most important task here is to override parent constructor and change the type-hint of the first argument to our created index (`CategoryIndex`).
```

docs/cookbook/adding-a-new-elasticsearch-index.md
grossmannmartin: it would be nice to say the best place where to create the new `CategoryExportCronModule` (it's later guessable from namespace, but at first it made me thinking)

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+        IndexDefinitionLoader $indexDefinitionLoader,
+        Domain $domain
+    ) {
+        parent::__construct($index, $indexFacade, $indexDefinitionLoader, $domain);
+    }
+}
+```
+
+Now if we have [crons](../introduction/cron.md) properly configured the new cron will be started automatically with others.
+Or you may want to [configure](../cookbook/working-with-multiple-cron-instances.md) this cron module with different timing
```
grossmannmartin:
```suggestion
Or you may want to [configure](../cookbook/working-with-multiple-cron-instances.md) this cron module with different timing.
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+### Implement partial update
+
+Sometimes we want to export data immediately after the original data are changed (hiding category, rename, ...).
+For this purpose we need to implement `CategoryIndex::getExportDataForIds()`, scheduler for queuing, and [subscriber](https://symfony.com/doc/current/event_dispatcher.html#creating-an-event-subscriber) for processing queue.
+
+Anywhere you want to add row to immediate export you may call `CategoryExportScheduler::scheduleRowIdForImmediateExport()` for adding a row into the queue.
+
+#### Scheduler
+
+Scheduler is used as a queue of IDs which we want to export. When we make any changes we may add an affected category ID into this queue and Subscriber will pick it up after request is done.
```
grossmannmartin:
```suggestion
Scheduler is used as a queue of IDs we want to export. When we make any changes we may add an affected category ID into this queue and subscriber will pick it up after request is done.
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+
+class CategoryExportScheduler extends AbstractExportScheduler
+{
+}
+```
+
+#### Subscriber
+
+Create class `CategoryExportSubscriber` which extends `AbstractExportSubsriber`.
+Override its `__construct()`, and `getSubscribedEvents()`.
```
grossmannmartin:
```suggestion
Override its `__construct()` and `getSubscribedEvents()` methods.
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+
+Here is important to override constructors arguments type hint.
+Instead of abstract classes from `\Shopsys\FrameworkBundle\Component\Elasticsearch` you need to replace it with our new implementation (`CategoryExportScheduler`, `CategoryIndex`).
+
+Implementation of `getSubscribedEvents()` is desired to use `exportScheduledRows()` from abstract class but also we can implement in by ourselves.
```
grossmannmartin: The method  `getSubscribedEvents()` is abstract in the parent class. So it has to be implemented

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+                ['exportScheduledRows', 0],
+            ],
+        ];
+    }
+}
+```
+
+#### CategoryIndex::getExportDataForIds()
+
+To finish partial exports we need to implement the last unimplemented method in `CategoryIndex` to return all categories by their identifiers.
```
grossmannmartin:
```suggestion
To finish partial exports we need to implement the last unimplemented method in `CategoryIndex` – to return all categories by their identifiers.
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
++       foreach ($categories as $category) {
++           $result[$category->getId()] = $this->convertToElastic($category, $domainId, $locale);
++       }
++       return $result;
+    }
+```
+
+## Conclusion
+
+We have created a new index category into elasticsearch. We are able to fill it with data (by a cron or immediately after row is changed).
```
grossmannmartin:
```suggestion
We have created a new index category into elasticsearch. We were able to fill it with data (by a cron or immediately after a row is changed).
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+From now you are able to use elasticsearch as a data source (data providing functionality is needed to be implemented) instead PostgreSQL which will save your performance.
```
grossmannmartin:
```suggestion
From now you are able to use elasticsearch as a data source (data providing functionality is needed to be implemented) instead of PostgreSQL which will improve the performance of your application.
```

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+    Shopsys\FrameworkBundle\Model\Category\Elasticsearch\CategoryIndex: ~
+```
+
+## Create new index into elasticsearch
+
+To create an index into elasticsearch we need to implement `getName()` method in `CategoryIndex`.
+The best practice is to define index name into constant due to later usage for obtaining data.
+
+```diff
++   public const INDEX_NAME = 'category';
```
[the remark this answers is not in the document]
s3tezsky: The purpose of public constant is to be available when you implement read logic. 
e.g. https://github.com/shopsys/shopsys/blob/e6076b845e80d3ba80b8d6cddff364fa4356a9bb/packages/framework/src/Model/Product/ProductOnCurrentDomainElasticFacade.php#L343

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
++
+    public function getName(): string
+    {
+-       // TODO: Implement getName() method.
++       return static::INDEX_NAME;
```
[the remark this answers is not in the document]
s3tezsky: IMHO it may be overridden when you need implement export methods by yourself.

docs/cookbook/adding-a-new-elasticsearch-index.md
s3tezsky: This is weird. Did not help you cleaning cache? I guess it should help - also it might be mentioned here.

docs/cookbook/adding-a-new-elasticsearch-index.md
s3tezsky: What about an Oxfords comma? 😄

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+```php
+declare(strict_types=1);
+
+namespace App\Model\Category\Elasticsearch;
+
+use Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractIndex;
+
+class CategoryIndex extends AbstractIndex
+{
+    public function getName(): string
```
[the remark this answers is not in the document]
s3tezsky: sure

packages/framework/src/Component/Elasticsearch/IndexRepository.php
s3tezsky: Fixed in https://github.com/shopsys/shopsys/pull/1622/commits/d9ad0cdeb3f3c663cdc93205baa312ecf6dbb3b9. Really thanks for noticing 👍

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Export data into elasticsearch
+
+Creating and deleting index is nice but it is not really useful.
+As a next step we will implement method `getTotalCount()` and `getExportDataForBatch()` to be able export data.
+
+We can use already existing method in `\Shopsys\FrameworkBundle\Model\Category\CategoryRepository::getTranslatedVisibleSubcategoriesByDomain()`.
+The method `getTranslatedVisibleSubcategoriesByDomain()` needs as a second argument an instance of `DomainConfig` so we need to inject an instance of `Domain` class along with instance of `CategortRepository` into `CategoryIndex`.  
+
+```diff
```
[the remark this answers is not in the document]
s3tezsky: I have changed it. The original purpose was to be more clear but it is a bit annoying when everything is highlighted by green background.

docs/cookbook/adding-a-new-elasticsearch-index.md
```diff
@@ -0,0 +1,294 @@
+# Adding a New Elasticsearch Index
+
+In this cookbook, we will add a new elasticsearch index for categories, implement basic functions for data export, implement cron module, and support for partial export.
+
+## New elasticsearch mapping
+
+As a first step we need to define [elasticsearch mapping](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html) in `src/Resources/definition/category/` for all domains (e.g. for 2 domains with ID 1 and 2: `1.json`, `2.json`).
+
```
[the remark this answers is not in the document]
grossmannmartin: What do you think about some minimalist configuration like
```
{
    "settings": {
        "index": {
            "number_of_shards": 1,
            "number_of_replicas": 0
        }
    },
    "mappings": {
        "_doc": {
            "properties": {
                "name": {
                    "type": "text"
                },
                "description": {
                    "type": "text"
                },
                "parentId": {
                    "type": "integer"
                },
                "level": {
                    "type": "integer"
                },
                "uuid": {
                    "type": "text"
                }
            }
        }
    }
}
```

docs/cookbook/adding-a-new-elasticsearch-index.md
grossmannmartin: I'm glad you're thinking about it 👍 
But Oxford comma takes place in a list of three or more items https://www.lexico.com/definition/oxford_comma

docs/cookbook/adding-a-new-elasticsearch-index.md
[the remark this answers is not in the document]
grossmannmartin: I see... Maybe making the getName() method static could be a little more straightforward to use.

docs/cookbook/adding-a-new-elasticsearch-index.md
grossmannmartin: Cleaning cache did not help.
I have `CategoryIndex extends AbstractIndex` and in `project-base/config/services.yaml` I registered this class as a service.
When I remove the tag from CategoryIndex from services.yaml, it fails on the error above.

upgrade/UPGRADE-v9.0.0-dev.md
```diff
@@ -446,9 +446,9 @@ There you can find links to upgrade notes for other versions too.
         +   }
         ```
         Be aware that to make this change in production environment you'll need to delete old structure and then create
-        a new one because of the change of `type` in `main_variant` field. If you want to know more you can see [this article](../docs/introduction/console-commands-for-application-management-phing-targets.md#product-search-migrate-structure)
+        a new one because of the change of `type` in `main_variant` field. If you want to know more you can see [this article](../docs/introduction/console-commands-for-application-management-phing-targets.md#elasticsearch-index-migrate)
 
-    - change and include new fields in ProductSearchExportWithFilterRepositoryTest
+    - change and include new fields in ProductExportRepositoryTest
```
grossmannmartin: That means the #1622 instruction should be done before this one, right?

docs/cookbook/adding-a-new-elasticsearch-index.md
s3tezsky: Fixed via https://github.com/shopsys/shopsys/pull/1622/commits/403af0cffe379157eecc3a018b482043eab43c46

upgrade/UPGRADE-v9.0.0-dev.md
s3tezsky: Fixed

upgrade/UPGRADE-v9.0.0-dev.md
```diff
@@ -657,6 +657,75 @@ There you can find links to upgrade notes for other versions too.
...
+		
+        - update `createFilter()` method
+
+            ```diff
+                protected function createFilter(): FilterQuery
+                {
+            -       $elasticSearchIndexName = $this->elasticSearchStructureManager->getAliasName(Domain::FIRST_DOMAIN_ID, self::ELASTICSEARCH_INDEX);
+            -	
+            -       $filter = $this->filterQueryFactory->create($elasticSearchIndexName);
+            +       $indexDefinition = $this->indexDefinitionLoader->getIndexDefinition(ProductIndex::getName(), Domain::FIRST_DOMAIN_ID);
```
grossmannmartin: use of ProductIndex is missing

vitek-rostislav:
Hi guys, I am experiencing a weird bug(?) - when I have `ELASTIC_SEARCH_INDEX_PREFIX` env variable set, my functional tests fail on `index_not_found_exception`. I solved my problem by discarding the "test" prefix in my `services_test.yml`, however, I have a strong feeling that this is not a proper solution as this way, I am using the same index for both test and prod/dev environment, am I right? :thinking:

vitek-rostislav:
ad the previous comment - it was a problem on our project where CI build target was overridden and was not dependent on `test-elasticsearch-index-recreate` - hence the problem with not existing index. Everything fixed now, THX @grossmannmartin and @s3tezsky for consultations :+1:

PR #1644, opened 2020-02-02
[project-base] Order can now be completed when successful flash message exists

### LINKED ISSUE #1235: Order cannot be sent with success flash message
### What is happening
Imagine use case, when project needs simple success flash message in all order steps. Adding success flash message solves this requirement, but order cannot be sent. It is weird in the first moment.

Order cannot be sent with existing any frontend flash message. It is OK for error messages and info messages but success messages should not stop the order from being sent, I think.

### Expected result
Order cannot be sent only with error or info flash message. Or order can be sent with success flash message.

pk16011990:
| Q             | A
| ------------- | ---
|New feature| No 
|Fixes issues| closes #1235 

### DIFF
--- a/project-base/src/Controller/Front/OrderController.php
+++ b/project-base/src/Controller/Front/OrderController.php
@@ -213,7 +213,7 @@ class OrderController extends FrontBaseController
         if ($isValid) {
             if ($orderFlow->nextStep()) {
                 $form = $orderFlow->createForm();
-            } elseif ($this->isFlashMessageBagEmpty()) {
+            } elseif (count($this->getErrorMessages()) === 0 && count($this->getInfoMessages()) === 0) {
                 $deliveryAddress = $orderData->deliveryAddressSameAsBillingAddress === false ? $frontOrderFormData->deliveryAddress : null;
                 $order = $this->orderFacade->createOrderFromFront($orderData, $deliveryAddress);
                 $this->orderFacade->sendHeurekaOrderInfo($order, $frontOrderFormData->disallowHeurekaVerifiedByCustomers);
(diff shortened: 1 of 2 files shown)


project-base/src/Controller/Front/OrderController.php
```diff
@@ -213,7 +213,7 @@ public function indexAction()
         if ($isValid) {
             if ($orderFlow->nextStep()) {
                 $form = $orderFlow->createForm();
-            } elseif ($this->isFlashMessageBagEmpty()) {
+            } elseif (count($this->getErrorMessages()) === 0 && count($this->getInfoMessages()) === 0) {
```
s3tezsky, line 216: It seems really strange the order flow is dependent on FlashMessages.

Could not `$form->isValid()` or `count($form->getErrors) === 0` be used here instead?

---

Otherwise it still should be stopped only when errors appears.

```suggestion
            } elseif (count($this->getErrorMessages()) === 0) {
```

TomasLudvik: You might want to display some success flash message and that would prevent Order from completion. Info messages might be used for situations like: product in your cart has new price, do you still want to finish the order?

s3tezsky: I sticked to original expectation:

> Expected result
>
> Order cannot be sent only with error flash message. Or order can be sent with info flash message.

I aggree you may do not want to continue in ordering when product price has changed. But should not even here be some kind of form validation instead of throwing flash message?

sslt: This conversation feels unresolved. Going by guts feelings, I have to agree with @s3tezsky .... flash messages should not govern the logic. I do get the example you raised @TomasLudvik (price change) but message should not be the only indicator affecting the flow of order process. 
Lets get third prg opinion on this one please.

henzigo: I absolutely agree with @sslt and @s3tezsky . There should be some validation instead of checking flash messages.

pk16011990: Order process must be postponed by anything with error/information for user. Validation in order form can not check everything in app. 

Or do you want create validator for empty flash message bag? How would you imagine such a validator?

TomasLudvik: @henzigo this conversation remained unresolved even we updated issue and agreed with @s3tezsky and @pk16011990 that current solution solves the problem, we are closing this now

PR #1650, opened 2020-02-04
[framework] use protected properties in Transformers and AdvancedSearch

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| simpler extensibility, Eg. creating new operator for Advanced search was pain (array with translations was private)<br> We forbid using private properties in framework classes due to extensibility. This PR changes its visibility to protected.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/easy-coding-standard.yml
+++ b/packages/framework/easy-coding-standard.yml
@@ -12,6 +12,8 @@ services:
                             - Shopsys\FrameworkBundle\Model
                             - Shopsys\FrameworkBundle\Component
                             - Shopsys\FrameworkBundle\Controller
+                            - Shopsys\FrameworkBundle\Form\Admin\AdvancedSearch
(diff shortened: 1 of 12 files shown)


packages/framework/easy-coding-standard.yml
```diff
@@ -12,6 +12,7 @@ services:
                             - Shopsys\FrameworkBundle\Model
                             - Shopsys\FrameworkBundle\Component
                             - Shopsys\FrameworkBundle\Controller
+                            - Shopsys\FrameworkBundle\Form\Transformer
```
s3tezsky: According to the other changes do not you want to extend the analysis for `Shopsys\FrameworkBundle\Form\Admin\AdvancedSearch` too?

pk16011990: good point, fixed

PR #1656, opened 2020-02-06
upgrade readme shops

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Upgraded list of shops built on shopsys framework. B2C demos is built on same app as B2B so I don't know we want his on list. If we don't want him on list, delete him.
|New feature| No 
|Fixes issues| ... 


sslt:
Thanks for this PR sspooky13, but list of projects using SSFW will be removed from readme file with SSFW 9 release.

PR #1658, opened 2020-02-06
[shopsys] coding standards for using strict comparison operators

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Disallows using loose == and != comparison operators. Use === and !== instead, they are much more secure, predictable and consistent.
|New feature| No 
|Fixes issues| ... 

Fixer needs manual check, because some code could rely on less strict comparisons -> BC break in my opinion.

### DIFF
--- a/project-base/easy-coding-standard.yml
+++ b/project-base/easy-coding-standard.yml
@@ -3,6 +3,8 @@ imports:
 
 services:
     PhpCsFixer\Fixer\Strict\DeclareStrictTypesFixer: ~
+    # @deprecated This will be moved from project-base to coding-standards package in next major version
+    SlevomatCodingStandard\Sniffs\Operators\DisallowEqualOperatorsSniff: ~
 
 parameters:
     exclude_files:
--- a/upgrade/UPGRADE-v9.1.0-dev.md
+++ b/upgrade/UPGRADE-v9.1.0-dev.md
@@ -39,3 +39,7 @@ There you can find links to upgrade notes for other versions too.
         - `Shopsys\FrameworkBundle\Model\Product\Product::$recalculateAvailability`
         - `Shopsys\FrameworkBundle\Model\Product\Product::$recalculatePrice`
         - `Shopsys\FrameworkBundle\Model\Product\Product::$recalculateVisibility`
+
+- introduced sniff for strict comparasion ([#1658](https://github.com/shopsys/shopsys/pull/1658))
+    - this change is temporary to help you prepare your project for next major version, where this will be required directly in coding-standards package
+    - see #project-base-diff to update your project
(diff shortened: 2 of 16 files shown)


s3tezsky:
Thanks for the PR!

I guess it is not necessary to waait until v10. This PR might be splitted into 2 separate PR.

1. Apply this standard rule in shopsys/framework, shopsys/project-base (as an optional upgrade) and maybe others packages in v9.1.0
2. Apply this standard rule in shopsys/coding-standards in v10 and drop previous rules from specific packages.

What do you think @TomasLudvik, @sslt?

upgrade/UPGRADE-v9.1.0-dev.md
```diff
@@ -31,3 +31,6 @@ There you can find links to upgrade notes for other versions too.
 - add phpstan-symfony extension ([#1961](https://github.com/shopsys/shopsys/pull/1961))
     - see #project-base-diff to update your project
 
+- introduced sniff for strict comparasion ([#1658](https://github.com/shopsys/shopsys/pull/1658))
+    - this change is optional but will be required in next major version
```
TomasLudvik, line 35: Please remove this line as it is not optional, cause it will be required ;)

s3tezsky: It will be required in next major release, not before ;)

TomasLudvik: My point still stands. It will be required. Upgrade notes are required when upgrading to next major, not to patch or minor version.

project-base/easy-coding-standard.yml
```diff
@@ -3,6 +3,7 @@ imports:
 
 services:
     PhpCsFixer\Fixer\Strict\DeclareStrictTypesFixer: ~
+    SlevomatCodingStandard\Sniffs\Operators\DisallowEqualOperatorsSniff: ~
```
TomasLudvik, line 6:
```suggestion
    # @deprecated This will be moved from project-base to coding-standards package in next major version
    SlevomatCodingStandard\Sniffs\Operators\DisallowEqualOperatorsSniff: ~
```

TomasLudvik: Add this same to framework package

upgrade/UPGRADE-v9.1.0-dev.md
```diff
@@ -31,3 +31,6 @@ There you can find links to upgrade notes for other versions too.
...
+    - see #project-base-diff to update your project
```
TomasLudvik, line 36:
```suggestion
- introduced sniff for strict comparasion ([#1658](https://github.com/shopsys/shopsys/pull/1658))
    - this change is temporary to help you prepare your project for next major version, where this will be required directly in coding-standards package
    - see #project-base-diff to update your project
```

s3tezsky: Nice!

PR #1659, opened 2020-02-06
[shopsys] do not change product availability to default when availability cannot be calculated immediately

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| It is better to keep current availability, then changing it to default, until calculation is run.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Product/Availability/ProductAvailabilityCalculation.php
+++ b/packages/framework/src/Model/Product/Availability/ProductAvailabilityCalculation.php
@@ -63,11 +63,11 @@ class ProductAvailabilityCalculation
     public function calculateAvailability(Product $product)
     {
         // If the product is not managed by EntityManager yet, it's not possible to calculate its availability consistently
-        // Let's return a default availability for the moment and mark the product for recalculation
+        // Let's return a same availability for the moment, do not change it now and mark the product for recalculation
         if ($this->em->contains($product) === false) {
             $product->markForAvailabilityRecalculation();
 
-            return $this->availabilityFacade->getDefaultInStockAvailability();
+            return $product->getCalculatedAvailability();
         }
 
         if ($product->isMainVariant()) {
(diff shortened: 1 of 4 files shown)


pk16011990:
+ next problem: some products after `db-demo` have null availability -> It can cause fatal like  
```php
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 1 passed to Shopsys\FrameworkBundle\Model\Product\Product::setCalculatedAvailability() must be an instance of Shopsys\FrameworkBundle\Model\Product\Availability\Availability, null given, called in /var/www/html/vendor/shopsys/framework/src/Model/Product/Availability/ProductAvailabilityRecalculator.php on line 94 in /var/www/html/vendor/shopsys/framework/src/Model/Product/Product.php:619
```

s3tezsky:
IMO the real problem here is having column `availability_id` as a nullable.
Maybe we should set availabilities to product which has `null` to `out_of_stock` reference and set the column NOT NULL.
What do you think @pk16011990?

s3tezsky:
> next problem: some products after db-demo have null availability -> It can cause fatal like

This is not true, I walked through ProductDataFixtures and each product has `availability_id` set to not null value.

Also I tried to throw the fatal error you have mentioned with no success. I just fixed the return type of `Product::getAvailability()` as the current one is invalid.

The only case the error was thrown is when you have invalid data in database (`availability_id` is `null` and `using_stock` is false).

TomasLudvik:
Hello @s3tezsky, @pk16011990 just informed me, that such fatal appears on master, when you set some product for availability recalculation, when such product has availability set to null.

pk16011990:
@s3tezsky, yes, problem with:
> * next problem: some products after `db-demo` have null availability -> It can cause fatal like
> 
> ```
> PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 1 passed to Shopsys\FrameworkBundle\Model\Product\Product::setCalculatedAvailability() must be an instance of Shopsys\FrameworkBundle\Model\Product\Availability\Availability, null given, called in 
/var/www/html/vendor/shopsys/framework/src/Model/Product/Availability/ProductAvailabilityRecalculator.php on line 94 in /var/www/html/vendor/shopsys/framework/src/Model/Product/Product.php:619
> ```

is not problem in SSFW, it is problem in my project :) sorry for mystification

Issue #907, opened 2019-03-25
Services needed in integration tests should not be manually registered as public


grossmannmartin:
### What is happening

It would be best if we wouldn't have to make every service that is taken from DIC in tests public. 

These copy-pasted service definitions can cause a lot of trouble. Either stop using the container directly (can we autowire services in integration tests?) or have all services public in test env by default (maybe via a compiler pass?)...

### Expected result
We don't need to manually add services into services_test.yml.

grossmannmartin:
Seems like we can fix this issue by upgrading to Symfony 4.1 – https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing

simara-svatopluk:
I don't even understand why do we use the container in tests. We're just hacking tests and using service locator instead of standard dependency injection... Isn't it possible with PHPUnit to just inject the service?

```php
class ArticleFacadeTest {

    private $facade;

    public function __construct(ArticleFacade $facade) {
        $this->facade = $facade;
    }

    // ...
}
```

grossmannmartin:
I don't know if it's by PHPUnit, Symfony or app (I didn't do enough digging), but a parameter in constructor is not working. 

First of all, Functional tests inherit from `PHPUnit\Framework\TestCase` and there is defined constructor `public function __construct($name = null, array $data = [], $dataName = '')`

And as a second, adding something as a parameter of the constructor throws exception `ArgumentCountError: Too few arguments to function ...::__construct(), 0 passed in ... and exactly 1 expected`

simara-svatopluk:
I'm bit annoyed by this issue, so I tried to search a bit... And there maybe is a hack solution that "looks like dependency injection", and should work with private services... I'll try it tomorrow if there is a time

https://github.com/jakzal/phpunit-injector#symfony-container-symfony-34--40

simara-svatopluk:
It works :smiling_imp: https://github.com/shopsys/shopsys/commit/a3f9e1361da5ab299ea232c839c0e19b5c9b9e1a

vitek-rostislav:
Come on stalebot, do not close this! :angry: 

We already have a solution prototype thanks to @simara-svatopluk :heart: 

It should be quite easy to create a proper PR now and get rid of `services_test.yml` once for all. The only question is, does it pay off to work on it or shall we wait for the upgrade on Symfony 4?

simara-svatopluk:
[...]

PR #1661, opened 2020-02-07
[shopsys] stop of services_test.yml maitenance 

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| functional tests sometimes fails with weird error after overriding of part of SSFW. Solition was annoying maintaining of services_test.yml
|New feature| No 
|Fixes issues| #907 

### DIFF
--- a/packages/framework/src/DependencyInjection/Compiler/PublicServicesForFunctionalTestsCompilerPass.php
+++ b/packages/framework/src/DependencyInjection/Compiler/PublicServicesForFunctionalTestsCompilerPass.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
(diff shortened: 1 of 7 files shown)


TomasLudvik:
Hello, @pk16011990 is this still issue? After updating to Symfony 4 this should be already solved, right?

TomasLudvik:
Ok I have discussed this with @pk16011990 and have some more information about this:
- first of all @pk16011990 does not think that everything has been solved, because there is many services left in `services_test.yaml` file, that has been removed in this PR. I think that needs some thorough investigation to find reasons why they are still there and if some needs to be kept add some comments describing why, so it is not subject of discussion in future.
- excluded services are present in this PR only because when they are registered via autodiscovery, `PublicServicesForFunctionalTestsCompilerPass` marks them as public an in such case they are not removed while Symfony optimizes container by removing unused services.

I would suggest as written at beginning, to investigate services_test.yaml and decide:
- which services can be removed
- which services has to be kept and add comment why it is that way

pk16011990:
> * which services can be removed

All what I removed in this PR... All tests pass, App works... what more is it needed?

TomasLudvik:
Hello @pk16011990, the milestone is set for this PR, so it will be merged before release of 9.1.0.

TomasLudvik:
After discussion with @pk16011990 I am closing this PR as #1957 solves this using Symfony test container

Issue #1387, opened 2019-09-09
Recalculations are called in deploy target even if there is not new domain


henzigo:
### What is happening

In `build-deploy-part-2-db-dependent` there is `domains-data-create` target that creates domain data for newly created domain and runs `shopsys:recalculations` command even if there is no new domain. 

This can increase deployment time which is undesirable.

### Expected result

`shopsys:recalculations` command should be run in `domains-data-create` target only if there are some new domains.

s3tezsky:
In my point of view there are two solutions:

1. Creating new phing target called `shopsys:domains-data:update`
    - which will respect current data and process only differences
2. Change `shopsys:domains-data:create` to behave like the `...:update` above
    - what will result in drastical change of behavior of this target (not good for me)
    - target name will be confisung according to it's purpose

Also it will need to require to extend `shopsys:recalculations` to accept parameter with domain ID (?) or hide this login into `shopsys:domains-data:update` target.

I would decide for creating new phing target for domains update and extend recalculations with an argument.

Before implementation I will need a PIA with some one else.

vitek-rostislav:
Hi guys, we would really appreciate if you solved this :slightly_smiling_face:

s3tezsky:
Hi @vitek-rostislav, I have to dissappoint you. We are still keeping it in our buffer with lower priority. Maybe we could get it done in a month but cannot guarantee anything.

s3tezsky:
closed by #1669

PR #1669, opened 2020-02-08
[framework] Run all recalculators during deploy only if domain is created

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Current implementation is unacceptable for projects with more products. Recalculations are now run only for newly created domains. Any products marked for recalculation are calculated with cron only and not during the deploy.
|New feature| No 
|Fixes issues| #1387 


grossmannmartin:
rebased on actual master (to be able to test the fix in context of actual master)

Issue #1449, opened 2019-10-04
Admin - Slider - domain editation dont have any effect


TomasGottvald:
### What is happening

1) When you edit domain for Slider item, it has no effect after save changes. 
2) When you click on second domain it shows only info "No page found"

### Expected result

I can edit domain for sliders and I can see filtered slider pages according to domains.

pk16011990:
not stale

grossmannmartin:
closed with #1673

PR #1673, opened 2020-02-08
[framework] fixed editing of SliderItem's domain id

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR fixed editing of SliderItem's domain id
|New feature| No 
|Fixes issues| #1449 

### DIFF
--- a/packages/framework/src/Model/Slider/SliderItem.php
+++ b/packages/framework/src/Model/Slider/SliderItem.php
@@ -72,6 +72,7 @@ class SliderItem implements OrderableEntityInterface
      */
     public function edit(SliderItemData $sliderItemData)
     {
+        $this->domainId = $sliderItemData->domainId;
         $this->name = $sliderItemData->name;
         $this->link = $sliderItemData->link;
         $this->hidden = $sliderItemData->hidden;


packages/framework/src/Model/Slider/SliderItem.php
```diff
@@ -72,6 +72,7 @@ public function __construct(SliderItemData $sliderItemData)
      */
     public function edit(SliderItemData $sliderItemData)
     {
+        $this->domainId = $sliderItemData->domainId;
         $this->name = $sliderItemData->name;
         $this->link = $sliderItemData->link;
         $this->hidden = $sliderItemData->hidden;
```
grossmannmartin: I noticed (SonarCloud noticed actually) that the implementation of `::edit` method is the same as `__construct` is. 
Does it make sense to call the `edit` method from the constructor?

pk16011990: It is "problem" a lot of entities and it is different "problem" in my opinion.

I don't know what is better, because in some case editation has little differences :-/

grossmannmartin: Okay, resolving this conversation as wont-fix. Further refactoring should take place instead.

PR #1676, opened 2020-02-10
[shopsys] added change and remove methods to js register class

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| If we want override  some js class (with constructor or init method) we can't use monkey patching, but we have to use standard classes overriding (like this class B extend A).  For this use case we have to change registered callback and this PR added method to change callback.
|New feature| No 
|Fixes issues| ... 


docs/frontend/npm-and-webpack.md
```diff
@@ -143,3 +143,32 @@ CustomizeBundle.showFormErrorsWindow = myOverridedShowFormErrorsWindow;
 ```
 
 This principle is called [Monkey Patching](https://www.sitepoint.com/pragmatic-monkey-patching/).
+
+### I want to override class from @shopsys/framework common package
+
+You can use ES6 syntax for override class. You certainly know key word `extend`. You can use it in es world now.
```
grossmannmartin:
```suggestion
You can use ES6 syntax to override class.
You certainly know the key word `extend`.
You can use it in the ES world now.
```

packages/framework/assets/js/admin/validation/form/freeTransportAndPayment.js
```diff
@@ -18,4 +18,4 @@ export default function validationFreTransportAndPayment () {
     });
 }
 
-(new Register()).registerCallback(validationFreTransportAndPayment);
+(new Register()).registerCallback(validationFreTransportAndPayment, 'validationFreTransportAndPayment');
```
grossmannmartin: here it looks like a typo (fre X free). I'm not sure if the class should be renamed in this PR, but the key can be correct even now, or?

packages/framework/assets/js/admin/validation/form/mailTemplate.js
```diff
@@ -20,4 +20,4 @@ export default function validationMailTemplate () {
     });
 }
 
-(new Register()).registerCallback(validationMailTemplate);
+(new Register()).registerCallback(validationMailTemplate, 'mailTemplate.validationMailTemplate');
```
grossmannmartin: is `mailTemplate.validationMailTemplate` correct? For example `validationAdvert` uses just `validationAdvert` without any prefix

packages/framework/assets/js/admin/validation/form/freeTransportAndPayment.js
pesektomas: 745c39a79812facb4e2922516390a1c9e137086a

packages/framework/assets/js/admin/validation/form/mailTemplate.js
pesektomas: 1c9bd3f272d46c82caa4dc1e1722134ef74d13f8

docs/frontend/npm-and-webpack.md
pesektomas: 7872665e959a169f7ab08abe7e64147a91d51fac

docs/frontend/npm-and-webpack.md
```diff
@@ -143,3 +143,40 @@ CustomizeBundle.showFormErrorsWindow = myOverridedShowFormErrorsWindow;
...
+        console.log('patched constructor');
+    }
+
+    init () {
+        new MyCategoryTreeSorting();
+    }
+}
+```
+
+If you override framework's common javascript class, you will have to change registered callback with original class init.
```
grossmannmartin:
```suggestion
If you override framework's javascript class, you will have to change registered callback of the original class init to your implementation.
```

PR #1680, opened 2020-02-12
[framework] responsive administration

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Makes admninstration responsive for mobile devices
|New feature| No 
|Fixes issues| ... 


packages/framework/assets/js/admin/components/sideMenu.js
```diff
@@ -4,12 +4,23 @@ import Register from '../../common/utils/register';
...
         this.$sideMenu = $sideMenu;
         this.$items = this.$sideMenu.filterAllNodes('.js-side-menu-item');
-
-        this.$items.click(function () {
-            $(this).filterAllNodes('.js-side-menu-submenu').show();
-            $(this).addClass('open');
+        const _this = this;
+
+        this.$items.click( event => {
+            if($(event.currentTarget).hasClass('open')){
```
pesektomas: I would recommend create from `open`, `hidden`, `.js-side-menu-submenu` and `.js-side-menu-item` strings constants

TomasGottvald: open and hidden are state class ... like .active we usually use them - and they are important for using css attributes.

pesektomas: A understand, but my point was about it that you use `open` string by eight times, `hidden` and `.js-side-menu-submenu`  by four times and `.js-side-menu-item` by two times. And it is weird for me :smile: But when you vouch for it will never be changed ...

PR #1684, opened 2020-02-14
[project-base] suppress Phpstan in App tests for getReference()

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When getReferrence() is used in test for project-base class, php stan fails. It is caused by changing of namespace from ShopBundle to App.
|New feature| No 
|Fixes issues| ... 


pk16011990:
It is unnecessary now, because https://github.com/shopsys/shopsys/commit/2fffcbc34f99c6496986e4066e338cf4fbae448a solve it in another way (scratch this check for all) -> Close

PR #1690, opened 2020-02-18
[framework] fix wrongly placed annotation in BasePriceCalculation

malyMiso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Function `calculateBasePriceRoundedByCurrency` for price calculation with rounding by currency has deprecated annotation, which is wrong. This annotation belongs to `applyCoefficients`. This PR fixes wrongly placed anntoation. 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Pricing/BasePriceCalculation.php
+++ b/packages/framework/src/Model/Pricing/BasePriceCalculation.php
@@ -79,6 +79,8 @@ class BasePriceCalculation
      */
     public function applyCoefficients(Price $price, Vat $vat, array $coefficients): Price
     {
+        @trigger_error(sprintf('The %s() method is deprecated and will be removed in the next major. This method is not used, only in tests.', __METHOD__), E_USER_DEPRECATED);
+
         $priceWithVatBeforeRounding = $price->getPriceWithVat();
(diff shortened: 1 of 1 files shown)


packages/framework/src/Model/Pricing/BasePriceCalculation.php
```diff
@@ -72,6 +70,8 @@ public function calculateBasePriceRoundedByCurrency(
     }
 
     /**
+     * @deprecated method is deprecated and will be removed in the next major. This method is not used, only in tests
```
pesektomas: It would be great to added a method `@trigger_error` into method's body. Or not?

malyMiso: Why not, fixed.

pesektomas: Thanks to fix. Why do I think it's good to have it there? Because phpunit bridge logs occurrence of trigger_error function :smile:

grossmannmartin:
rebased to current 8.1 to fix tests due to date error fixed in #1677

PR #1696, opened 2020-02-24
[framework] added fallback to migrate previous elasticsearch structure to current one

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| After merging #1622 the alias names has changed which caused throwing exception whne running migration command. Also after implementing product variants in #1557 there was incompatible change in Elasticsearch mapping. <br>This PR adds fallback to resolve previous index name and fixes incompatible change in mapping to be able to migrate data without need of recreating and exporting whole structure.
|New feature| No 
|Fixes issues| ... 


packages/framework/src/Component/Elasticsearch/IndexFacade.php
```diff
@@ -167,7 +168,15 @@ public function migrate(IndexDefinition $indexDefinition, OutputInterface $outpu
     {
         $indexName = $indexDefinition->getIndexName();
         $domainId = $indexDefinition->getDomainId();
-        $existingIndexName = $this->indexRepository->findCurrentIndexNameForAlias($indexDefinition->getIndexAlias());
+
+        try {
+            $existingIndexName = $this->indexRepository->findCurrentIndexNameForAlias($indexDefinition->getIndexAlias());
+        } catch (ElasticsearchIndexException $exception) {
+            if (strpos($exception->getMessage(), 'Can\'t found any index with alias') === false) {
```
vitek-rostislav: This is quite fragile, especially when I see the exception message that will be probably proposed for fixing sooner or later - "Can't found" is wrong, should be rather "Not able to find" :grin: 

What about throwing and catching more specific Exception? :slightly_smiling_face:

PR #1707, opened 2020-02-27
[framework] add js-symbol-after-input class to symbolAfterInput span

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| this might be very handy - eg. I want to change the currency symbol after my `Money` input when domain is changed in another form field
|New feature| it is a little tweak
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Resources/views/Admin/Form/theme.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Form/theme.html.twig
@@ -1015,7 +1015,7 @@
 
 {% block appendix_block %}
     {% if symbolAfterInput is defined %}
-        <span class="form-line__item form-line__item--text form-line__item--fixed-width-left">{{ symbolAfterInput }}</span>
+        <span class="form-line__item form-line__item--text form-line__item--fixed-width-left js-symbol-after-input">{{ symbolAfterInput }}</span>
     {% endif %}
     {% if attr.icon is defined %}
         <i class="svg svg-{{ attr.iconType|default('info') }} in-icon cursor-help form-line__item form-line__item--info js-tooltip"


vitek-rostislav:
Hi @pesektomas, thanks and sorry for the late response :slightly_smiling_face: 

TBH, I do not think there is such a use case in the stable version. My use case was a creation of price level setting for order gifts - there was a select for domain and then input for price level setting :wink:

TomasLudvik:
Hello @vitek-rostislav,

I am sorry, but I do not feel it should be in Shopsys Framework if it has no usage there as it violates the YAGNI principle. Would you be ok, if we closed this PR?

vitek-rostislav:
How come it violates the YAGNI? It is not used in project-base, however, I needed it on my project so I believe I (or anyone else) am gonna need it in the future again :slightly_smiling_face:

vitek-rostislav:
See the description of the PR, I found that class handy on project (and I feel that another project might need it too - and it is not easy to find out what needs to be overridden), however, you are probably right, in the stable version of the framework it is an unused piece of code :+1: 

So if there is more of you in the dev team that thinks this should not be in the framework, I am absolutely ok with closing the PR :wink:

vitek-rostislav:
:laughing: I did not want to close it right away :laughing:

henzigo:
I didn't know about such functionality. This could save me a lot of work if I knew it before or it was used by default

grossmannmartin:
Hi, thanks for the work you did!
I understand it can help with your particular use-case, but I'm afraid of adding "this may be handy" features. It's easy to end with code full of one-time use-case things. 

Instead, I would rather see the theme separated into multiple files and make it easily overridable.
I'm sorry, but I'm voting for closing this PR.

TomasLudvik:
After voting, this feature is in violence with YAGNI principle and we voted to close this, sorry.

PR #1719, opened 2020-03-05
[framework] ProductExportRepository: prevent Undefined offset notice

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| `$product->getCategoriesIndexedByDomainId` may not contain categories for provided domain ID
|New feature| Yes 
|Fixes issues| ... 


TomasLudvik:
Hello @vitek-rostislav,

could you please add some cases when this may occur? Thank you!

vitek-rostislav:
Hi, I am not sure if I remember it correctly - I believe when you transfer categories from IS, it may happen that you transfer only categories for a certain domain...

Or, when you create a new product another way than using the form in admin, than `productData::$categoriesByDomainId` is an empty array -> `Product::$categories` is an empty ArrayCollection

PR #1722, opened 2020-03-09
[framework] deliveryAddressChoiceFields.html.twig: fix risk of getter call on null

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| `DeliveryAddress::$country` is nullable so it might happen that getter on null is called in the template
|New feature| No 
|Fixes issues| ... 


packages/framework/src/Resources/views/Front/Form/deliveryAddressChoiceFields.html.twig
```diff
@@ -16,7 +16,9 @@
                                         {{ deliveryAddress.companyName }}<br>
                                     {% endif %}
                                     {{ deliveryAddress.street }}, {{ deliveryAddress.city }} {{ deliveryAddress.postcode }}<br>
-                                    {{ deliveryAddress.country.name }}<br>
+                                    {% if deliveryAddress.country %}
```
TomasLudvik, line 19:
```suggestion
                                    {% if deliveryAddress.country is not null %}
```

PR #1726, opened 2020-03-11
[shopsys] added transport and payment to frontend API

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to add orders to frontend API, in order to do that we need other connected entities to be on API too. This PR adds first of them - transport and payment
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/src/Model/Resolver/Payment/PaymentResolver.php
+++ b/packages/frontend-api/src/Model/Resolver/Payment/PaymentResolver.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Resolver\Payment;
+
+use Overblog\GraphQLBundle\Definition\Resolver\AliasedInterface;
+use Overblog\GraphQLBundle\Definition\Resolver\ResolverInterface;
+use Overblog\GraphQLBundle\Error\UserError;
+use Ramsey\Uuid\Uuid;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Model\Payment\Payment;
+use Shopsys\FrameworkBundle\Model\Payment\PaymentFacade;
+
+class PaymentResolver implements ResolverInterface, AliasedInterface
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Payment\PaymentFacade
+     */
+    protected $paymentFacade;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Domain\Domain
+     */
+    protected $domain;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Payment\PaymentFacade $paymentFacade
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     */
+    public function __construct(PaymentFacade $paymentFacade, Domain $domain)
+    {
+        $this->paymentFacade = $paymentFacade;
+        $this->domain = $domain;
+    }
+
+    /**
+     * @param string $uuid
+     * @return \Shopsys\FrameworkBundle\Model\Payment\Payment
+     */
+    public function resolver(string $uuid): Payment
+    {
+        if (Uuid::isValid($uuid) === false) {
+            throw new UserError('Provided argument is not valid UUID.');
+        }
+
+        try {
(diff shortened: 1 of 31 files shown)


packages/frontend-api/src/Model/Resolver/Payment/PaymentResolver.php
```diff
@@ -0,0 +1,63 @@
...
+
+use Overblog\GraphQLBundle\Definition\Resolver\AliasedInterface;
+use Overblog\GraphQLBundle\Definition\Resolver\ResolverInterface;
+use Overblog\GraphQLBundle\Error\UserError;
+use Ramsey\Uuid\Uuid;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Model\Payment\Payment;
+use Shopsys\FrameworkBundle\Model\Payment\PaymentFacade;
+
+class PaymentResolver implements ResolverInterface, AliasedInterface
```
pesektomas: I am thinking about use case for this resolver on frontend (same as TransportResolver) :thinking:

project-base/tests/FrontendApiBundle/Functional/Payment/PaymentTest.php
```diff
@@ -0,0 +1,58 @@
...
+    protected $payment;
+
+    protected function setUp(): void
+    {
+        $this->payment = $this->paymentFacade->getById(2);
+
+        parent::setUp();
+    }
+
+    public function testCategoryNameByUuid(): void
```
pesektomas: I don't see a context in the method name :thinking:

project-base/tests/FrontendApiBundle/Functional/Payment/PaymentsTest.php
```diff
@@ -0,0 +1,108 @@
...
+            ],
+        ];
+
+        $this->assertQueryWithExpectedArray($query, $arrayExpected);
+    }
+
+    /**
+     * @return string
+     */
+    protected function getLocaleForFirstDomain(): string
```
pesektomas: I am thinking about move this method into `GraphQlTestCase`, what do you think? Same method is created in ~three~ five test classes at now.

packages/frontend-api/src/Model/Resolver/Transport/TransportsResolver.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->transportFacade = $transportFacade;
+        $this->paymentFacade = $paymentFacade;
+    }
+
+    /**
+     * @return array
+     */
+    public function resolve(): array
+    {
+        return $transports = $this->transportFacade->getVisibleOnCurrentDomain($this->paymentFacade->getVisibleOnCurrentDomain());
```
pesektomas: I'm a little confused. Why visible payments are resolved in this method, but visible transports are not resolved in PaymentFacade::getVisibleOnCurrentDomain?

project-base/tests/FrontendApiBundle/Functional/Payment/PaymentTest.php
TomasLudvik: forgotten copy paste, fixed

project-base/tests/FrontendApiBundle/Functional/Payment/PaymentsTest.php
TomasLudvik: moved

packages/frontend-api/src/Model/Resolver/Payment/PaymentResolver.php
TomasLudvik: Well I think it might be useful in some cases

packages/frontend-api/src/Model/Resolver/Transport/TransportsResolver.php
TomasLudvik: I think that reason there is that you will not display transport that has no visible payment

pesektomas: I am just surprised that this is not same as the opposite :thinking:

packages/frontend-api/src/Model/Resolver/Payment/PaymentResolver.php
pesektomas: Yes, but I can't do the case use to find :smile:

PR #1731, opened 2020-03-18
[shopsys] added user login to frontend API

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fully functional frontend API requires ability to log in users. This PR adds such functionality.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/src/Model/Mutation/Login/LoginMutation.php
+++ b/packages/frontend-api/src/Model/Mutation/Login/LoginMutation.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Mutation\Login;
+
+use GraphQL\Error\UserError;
+use Overblog\GraphQLBundle\Definition\Argument;
+use Overblog\GraphQLBundle\Definition\Resolver\AliasedInterface;
+use Overblog\GraphQLBundle\Definition\Resolver\MutationInterface;
+use Shopsys\FrameworkBundle\Model\Customer\User\FrontendCustomerUserProvider;
+use Shopsys\FrontendApiBundle\Model\Token\TokenFacade;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
+
+class LoginMutation implements MutationInterface, AliasedInterface
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Customer\User\FrontendCustomerUserProvider
+     */
+    protected $frontendUserProvider;
+
+    /**
+     * @var \Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface
+     */
+    protected $userPasswordEncoder;
+
+    /**
+     * @var \Shopsys\FrontendApiBundle\Model\Token\TokenFacade
+     */
+    protected $tokenFacade;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Customer\User\FrontendCustomerUserProvider $frontendCustomerUserProvider
+     * @param \Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface $userPasswordEncoder
+     * @param \Shopsys\FrontendApiBundle\Model\Token\TokenFacade $tokenFacade
+     */
+    public function __construct(
+        FrontendCustomerUserProvider $frontendCustomerUserProvider,
+        UserPasswordEncoderInterface $userPasswordEncoder,
+        TokenFacade $tokenFacade
+    ) {
+        $this->frontendUserProvider = $frontendCustomerUserProvider;
+        $this->userPasswordEncoder = $userPasswordEncoder;
+        $this->tokenFacade = $tokenFacade;
+    }
+
+    /**
+     * @param \Overblog\GraphQLBundle\Definition\Argument $argument
+     * @return string[]
+     */
+    public function login(Argument $argument): array
+    {
+        $input = $argument['input'];
+
+        try {
+            $user = $this->frontendUserProvider->loadUserByUsername($input['email']);
+        } catch (UsernameNotFoundException $e) {
+            throw new UserError('Log in failed.');
+        }
+
+        if (!$this->userPasswordEncoder->isPasswordValid($user, $input['password'])) {
+            throw new UserError('Log in failed.');
+        }
+
+        return [
+            'accessToken' => $this->tokenFacade->generateAccessTokenByCustomerUser($user),
+        ];
+    }
+
+    /**
+     * @return string[]
+     */
+    public static function getAliases(): array
+    {
+        return [
+            'login' => 'user_login',
+        ];
+    }
+}
--- a/packages/frontend-api/src/Model/Token/TokenAuthenticator.php
+++ b/packages/frontend-api/src/Model/Token/TokenAuthenticator.php
@@ -1,0 +1,148 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Token;
+
+use GraphQL\Error\FormattedError;
+use InvalidArgumentException;
+use Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser;
+use Shopsys\FrameworkBundle\Model\Customer\User\FrontendCustomerUserProvider;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+use Symfony\Component\Security\Guard\AbstractGuardAuthenticator;
+
+class TokenAuthenticator extends AbstractGuardAuthenticator
+{
+    protected const HEADER_AUTHORIZATION = 'Authorization';
+
+    protected const BEARER = 'Bearer ';
+
+    /**
+     * @var \Shopsys\FrontendApiBundle\Model\Token\TokenFacade
+     */
+    protected $tokenFacade;
+
+    /**
+     * @param \Shopsys\FrontendApiBundle\Model\Token\TokenFacade $tokenFacade
+     */
+    public function __construct(TokenFacade $tokenFacade)
+    {
+        $this->tokenFacade = $tokenFacade;
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return bool
+     */
+    public function supports(Request $request): bool
+    {
+        return $request->headers->has(static::HEADER_AUTHORIZATION) && strpos($request->headers->get('Authorization'), static::BEARER) === 0;
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return string|null
+     */
+    public function getCredentials(Request $request): ?string
+    {
+        $authorizationHeader = $request->headers->get(static::HEADER_AUTHORIZATION);
+
+        return substr($authorizationHeader, strlen(static::BEARER));
+    }
+
+    /**
+     * @param string|null $credentials
+     * @param \Symfony\Component\Security\Core\User\UserProviderInterface $userProvider
+     * @return \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser|null
+     */
+    public function getUser($credentials, UserProviderInterface $userProvider): ?CustomerUser
+    {
+        if ($credentials === null) {
+            return null;
+        }
+
+        if (!$userProvider instanceof FrontendCustomerUserProvider) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'The user provider must be an instance of FrontendUserProvider ("%s" was given).',
+                    get_class($userProvider)
+                )
+            );
+        }
+
+        $token = $this->tokenFacade->getTokenByString($credentials);
+
+        return $userProvider->loadUserByUsername($token->getClaim('email'));
+    }
+
+    /**
+     * @param string|null $credentials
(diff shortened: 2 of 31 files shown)


packages/frontend-api/src/Resources/config/services.yml
```diff
@@ -5,7 +5,11 @@ services:
         public: false
 
     Shopsys\FrontendApiBundle\:
-        resource: '../../**/*{Resolver,Map}.php'
+        resource: '../../**/*{Facade,Map,Resolver}.php'
+
+    Shopsys\FrontendApiBundle\Model\Mutation\:
+        resource: '../../Model/Mutation/'
+        tags: ['overblog_graphql.mutation']
```
grossmannmartin: Why is it necessary to tag mutation services, when Resolvers are not and without the tag, it works.
(I assume you used this https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/resolver.md#the-service-way)

packages/frontend-api/src/Resources/config/services.yml
```diff
@@ -5,7 +5,11 @@ services:
         public: false
 
     Shopsys\FrontendApiBundle\:
-        resource: '../../**/*{Resolver,Map}.php'
+        resource: '../../**/*{Facade,Map,Resolver}.php'
```
grossmannmartin: 👍

project-base/config/parameters_common.yml
```diff
@@ -33,3 +33,8 @@ parameters:
     build-version: '0000000000000000'
     shopsys.display_timezone: Europe/Prague
     shopsys.image.enable_lazy_load: true
+
+# To enable Frontend API on your first two domains uncomment next three lines
+#    shopsys.frontend_api.domains:
+#        - 1
+#        - 2
```
grossmannmartin: 😍

project-base/config/services.yaml
```diff
@@ -11,6 +11,8 @@ services:
         autowire: true
         autoconfigure: true
         public: false
+        bind:
+            $secret: '%secret%'
```
grossmannmartin: Why is this needed? I thought the secret is obtained from ParameterBag

project-base/tests/FrontendApiBundle/Functional/Login/LoginTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+declare(strict_types=1);
+
+namespace Tests\FrontendApiBundle\Functional\Login;
+
+use Tests\FrontendApiBundle\Test\GraphQlTestCase;
+
+class LoginTest extends GraphQlTestCase
+{
+    public function testPaymentNameByUuid(): void
```
grossmannmartin: I believe the name of the test should be different ;)

pesektomas: 96da1cf35ecd5d514da9ad2e7f5455b8fc013a8e

project-base/config/services.yaml
pesektomas: c1c034f8fb112803cd13dbc0dd0f9fd5a767394d

packages/frontend-api/src/Resources/config/services.yml
pesektomas: 159be85a16709a2efae20f1e598baaeb2614b0e3

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
```diff
@@ -0,0 +1,136 @@
...
+        $this->tokenFacade = $tokenFacade;
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return bool
+     */
+    public function supports(Request $request): bool
+    {
+        return $request->headers->has('X-AUTH-TOKEN');
```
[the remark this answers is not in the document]
pesektomas: cdec432490db60b3683d69fb968cceddd9799d9f

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return string|null
+     */
+    public function getCredentials(Request $request): ?string
+    {
+        return $request->headers->get('X-AUTH-TOKEN');
```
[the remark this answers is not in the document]
pesektomas: cdec432490db60b3683d69fb968cceddd9799d9f

packages/frontend-api/src/Model/Token/TokenCustomerUserTransformer.php
```diff
@@ -0,0 +1,24 @@
...
+class TokenCustomerUserTransformer
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser $user
+     * @return array
+     */
+    public static function transform(CustomerUser $user): array
+    {
+        return [
+            'id' => $user->getId(),
```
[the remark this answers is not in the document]
pesektomas: d818c93eb7f16d5173d4992b0a778fcb09a03c2e

packages/frontend-api/src/Model/Mutation/Login/LoginMutation.php
```diff
@@ -0,0 +1,80 @@
...
+     * @return string[]
+     */
+    public function login(Argument $argument): array
+    {
+        $input = $argument['input'];
+
+        try {
+            $user = $this->frontendUserProvider->loadUserByUsername($input['email']);
+        } catch (UsernameNotFoundException $e) {
+            throw new UserError(t('Log in failed.'));
```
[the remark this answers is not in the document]
pesektomas: I think that the API should not return translated errors and the application should take care of the translation.

Why do I think so? Because we have documentation in English and in the first phase the developer should agree with the API and if api returns the translated errors, the developer has no chance to find some information in the documentation.

Maybe we can start thinking about error codes and list of these errors in documentation.

2dd241f12ee7f74798ed079df8b046141c87aeff

packages/frontend-api/src/Model/Token/TokenFacade.php
```diff
@@ -0,0 +1,121 @@
...
+        return new Signer\Hmac\Sha256();
+    }
+
+    /**
+     * @param string $tokenString
+     * @return \Lcobucci\JWT\Token
+     */
+    public function getTokenByString(string $tokenString): Token
+    {
+        return (new Parser())->parse($tokenString);
```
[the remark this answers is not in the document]
pesektomas: Hmm, I think you get 500 when you send an expired token as well.

But I do not know if exceptions should be caught here or at the place where this method will be called :thinking:

packages/frontend-api/src/Model/Token/TokenCustomerUserTransformer.php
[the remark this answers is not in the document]
grossmannmartin: thanks

packages/frontend-api/src/Resources/config/services.yml
grossmannmartin: What about adding Mutation into the general definition on line 8?
like 
```diff
    Shopsys\FrontendApiBundle\:
-       resource: '../../**/*{Facade,Map,Resolver}.php'
+       resource: '../../**/*{Facade,Map,Mutation,Resolver,}.php'
```

instead of a separate definition?

packages/frontend-api/src/Model/Mutation/Login/LoginMutation.php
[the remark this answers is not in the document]
grossmannmartin: I agree, thank you :)

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @param \Symfony\Component\Security\Core\Exception\AuthenticationException|null $authException
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function start(Request $request, ?AuthenticationException $authException = null): Response
+    {
+        $responseData = [
+            'message' => t('Authentication Required'),
```
grossmannmartin: I notice another translated message here :)

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+        return $request->headers->has(static::HEADER_AUTHORIZATION);
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return string|null
+     */
+    public function getCredentials(Request $request): ?string
+    {
+        return $request->headers->get(static::HEADER_AUTHORIZATION);
```
grossmannmartin: I noticed you changed the header. What I read, the authorization header can be used for basic and token auth, so it's usually in the following form `Authorization: Bearer <api-key>`.

Is it right to let it without the Bearer claim?
https://learning.postman.com/docs/postman/sending-api-requests/authorization/#bearer-token

packages/frontend-api/src/Model/Token/TokenFacade.php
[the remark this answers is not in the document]
grossmannmartin: The method is used several times in TokenAuthenticator. I would handle the exception here and throw UserError with some meaningful message ("token is not token at all" :D) instead of doing it every time I need to use the method.
But I don't have really strong preference about it.

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
pesektomas: You are right, sorry - 30d7cdd6bb1753470e5fb2940fe7ca655e4a30d9

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
pesektomas: 5ed1d0911f09c94b7166838f55b0bf9d21f4b7f7

packages/frontend-api/src/Resources/config/services.yml
pesektomas: 41dfa325e9f286aeef4b6f70ca940962ac311db7

packages/frontend-api/src/Model/Token/TokenFacade.php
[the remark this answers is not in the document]
pesektomas: Hmm, I wrote about method `checkCredentials`. I don't know why. 8b5c43ac34426ebc0cd316016f355933b821573e

packages/frontend-api/src/Resources/config/services.yml
```diff
@@ -5,7 +5,7 @@ services:
         public: false
 
     Shopsys\FrontendApiBundle\:
-        resource: '../../**/*{Resolver,Map}.php'
+        resource: '../../**/*{Facade,Map,Resolver,Mutation}.php'
```
grossmannmartin:
```suggestion
        resource: '../../**/*{Facade,Map,Mutation,Resolver}.php'
```

I think it's nice to keep things alphabetically ordered, what do you think?

packages/frontend-api/src/Model/Token/TokenFacade.php
```diff
@@ -0,0 +1,127 @@
...
+
+        return (string)$tokenBuilder->getToken($this->getSigner(), $this->getKey());
+    }
+
+    /**
+     * @return \Lcobucci\JWT\Signer\Key
+     */
+    public function getKey(): Key
+    {
+        return new Key($this->parameterBag->get('secret'));
```
grossmannmartin: I still struggle with this, as we spoke earlier. 
Right now it's really to enable frontend-api without changing the default secret. That would be fatal for any production shop (considering it the same as leaving default admin with the default password in production, but much easier to do by mistake as it's not in the DB, but in parameters.yml file).
I think it's really important to keep the key in a different location than DB as it's now.

documentation about signing https://lcobucci-jwt.readthedocs.io/en/latest/configuration/

What do you say?

packages/frontend-api/src/Resources/config/services.yml
pesektomas: 5be21b9f98706bd45ef8b5bbeda2e988435bca90

packages/frontend-api/src/Model/Token/TokenFacade.php
pesektomas: We add US to sign tokens with a certificate. :+1:

pesektomas:
> Hi, I did some testing with following results:

Hello @grossmannmartin, thank you for testing. I fixed points 1-3 in 50d10748cebe0a30a1c776acf4dddcc94447a755. Point 4 will be solved in US for sign tokens by certificate and point 5 is same as our login form - we could solved it in closely future.

packages/frontend-api/src/Model/Token/TokenFacade.php
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * @param string $tokenString
+     * @return \Lcobucci\JWT\Token
+     */
+    public function getTokenByString(string $tokenString): Token
+    {
+        try {
+            return (new Parser())->parse($tokenString);
+        } catch (\Throwable $throwable) {
```
grossmannmartin: _nitpick_ 🤓 
Shouldn't it be just `Throwable` with `use Throwable` at the beginning? (same with BadMethodCallException on line 127)

packages/frontend-api/src/Resources/config/graphql-types/LoginDecorator.types.yml
```diff
@@ -0,0 +1,11 @@
+LoginDecorator:
```
grossmannmartin: _nitpick_ 🤓 

What about naming input objects like `LoginInputDecorator` and `LoginInput`? I think it helps knowing purpose of each type (when creating and consuming API)

pesektomas: Hmm, I tried it and I don!t think so. Why? Whith naming as LoginInput query would look like `mutation {LoginInput(input: {email: \"no-reply@shopsys.com\", password: \"user123\"}) {accessToken}}` - this is too many input in my opinion :smile:

PR #1735, opened 2020-03-19
[framework] categories with empty name are now considered not visible

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When importing data from IS, it might happen easily that you persist empty string for your category name. Thanks to the original condition (`ct.name IS NOT NULL`), such a category is considered visible then, because it is not null indeed, however, that is just wrong. This PR improves the condition for category visibility calculation - category with the empty name should not be visible as well. 
|New feature| No 
|Fixes issues| ... 

It might be a good idea to inspect the codebase for similar conditions :thinking:

### DIFF
--- a/packages/framework/src/Model/Category/CategoryVisibilityRepository.php
+++ b/packages/framework/src/Model/Category/CategoryVisibilityRepository.php
@@ -109,7 +109,7 @@ class CategoryVisibilityRepository
                 SET visible = (
                     cd.enabled = TRUE
                     AND
-                    ct.name IS NOT NULL
+                    (ct.name IS NOT NULL OR ct.name != '')
                     AND
                     parent_cd.visible = TRUE
                 )


vitek-rostislav:
Hi @pk16011990, why the downvote? Maybe a better solution could be that the application itself would ensure that an empty string is transformed into `NULL`?

TomasLudvik:
Hi @vitek-rostislav, 

yes, it would be probably better to check for empty string and then return null. For example, your solution would still be bad if somebody added one space there. So some trim and comparing empty string when saving would be better.

vitek-rostislav:
True, I will fix it one day :grin:

TomasLudvik:
Hello @vitek-rostislav,
we discussed it internally and we decided, that better way would be to sanitize input value to trim string and if it is empty save null instead. Also migration to sanitize all previous data would be nice.

s3tezsky:
Fixed in #2206

PR #1739, opened 2020-03-25
[shopsys] cron definitions moved to project-base

### LINKED ISSUE #350: Plugins should not add cron modules automatically into application
### What is happening
*Reported by @jDolba*.
It is very weird that a plugin can register new cron module simply by tagging a service with `shopsys.cron` tag (and implementing the cron module interface). This might be bad for at least two reasons:
- any bundle can add cron module into application and a developer might not even notice it
- the plugin's cron modules are run with higher priority that those that are defined in `ShopsysFrameworkBundle` and `ShopsysShopBundle` (as these bundles are loaded last) and a developer is not able to influence the order of cron modules.

Note: You can see order of cron modules along with launch times by adding following code into `\Shopsys\ShopBundle\Component\Cron\Config\CronConfig::getCronModuleConfigsByTime()`
```
d(array_map(function(CronModuleConfig $config) {return $config->getServiceId() . $config->getTimeHours() . $config->getTimeMinutes();}, $this->cronModuleConfigs));
exit;
```

### Expected result
A developer has the definite power over cron modules in his application (i.e. is able to enable/disable a cron module, change their order or time of execution easily) - inlcuding cron modules installed by external bundles.

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| It was not obvious how to disable / change order of crons. Moving them to project-base enables developers to easily control them.
|New feature| No 
|Fixes issues| closes #350 


vitek-rostislav:
Please do not merge it this way - I discovered one "bug" - it does not make sense run export to Elastic before availability and visibility of products are recalculated, please check the order of the modules here. 

Even price recalculation might be considered, I am not sure

vitek-rostislav:
ok, the proper recalculations after order creation should be probably covered by https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Order/Item/OrderProductFacade.php#L104 so it is probably ok in such a case :man_shrugging: 

But still, the order of modules does not make sense - every five minutes are products exported to Elastic and right after that, some attributes (that were already exported) are recalculated - this just does not make sense :weary:

upgrade/UPGRADE-v9.0.0-dev.md
```diff
@@ -1345,6 +1345,9 @@ There you can find links to upgrade notes for other versions too.
 - method name in `HeurekaCategoryDownloader` has been changed ([#1740](https://github.com/shopsys/shopsys/pull/1740))
     - `HeurekaCategoryDownloader::convertToShopEntities()` has been renamed to `HeurekaCategoryDownloader::convertToHeurekaCategoriesData()` update your project appropriately
 
+- move cron definitions in your project so it is easier to control them ([#])
```
grossmannmartin:
```suggestion
- move cron definitions in your project so it is easier to control them ([#1739](https://github.com/shopsys/shopsys/pull/1739))
```

PR #1750, opened 2020-03-29
[framework] Product::setProductCategoryDomains() fix

vitek-rostislav:
- the method now adds the new elements only when they are not present in the current data
- this solution prevents the application from possible duplicate key errors
- original problem: when flush without parameter is called on entity manager, elements
that are removed from collection are scheduled for DELETE, however,
INSERT queries are executed before DELETE, so the application fails here
- the behavior is connected with `orphanremoval=true` and `cascade={"persist"}` settings of `Product::$productCategoryDomains`

| Q             | A
| ------------- | ---
|Description, reason for the PR| the application might fail on duplicate key error when flush without parameter is called during the product editing
|New feature| No 
|Fixes issues| ... 

Original authors of this fix are @pk16011990 and @machicek so kudos to them! :+1: 

:warning: Btw, someone should really examine the other collections in the application - there might be a similar problem. :warning:

### DIFF
--- a/packages/framework/src/Model/Product/Product.php
+++ b/packages/framework/src/Model/Product/Product.php
@@ -634,17 +634,38 @@ class Product extends AbstractTranslatableEntity
      */
     public function setProductCategoryDomains(array $productCategoryDomains)
     {
-        $this->productCategoryDomains->clear();
-
-        foreach ($productCategoryDomains as $productCategoryDomain) {
-            $this->productCategoryDomains->add($productCategoryDomain);
+        foreach ($this->productCategoryDomains as $productCategoryDomain) {
+            if ($this->isProductCategoryDomainInArray($productCategoryDomain, $productCategoryDomains) === false) {
+                $this->productCategoryDomains->removeElement($productCategoryDomain);
+            }
         }
-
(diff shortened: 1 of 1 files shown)


vitek-rostislav:
Oh, and maybe this should be targetted against `7.3` branch (provided it is still supported)

grossmannmartin:
T.Hanks for the PR here. I think it would be nice to have a new test for this behavior ;)

And as it's not a BC break and really it is a bugfix, I would like to see it merged into 7.3 branch.

vitek-rostislav:
You're welcome :wink: 

I wanted to re-target the branch and then I realized that it is not a problem in `7.3` :grin: There is the old approach with using services in entities and the problem should be covered there, see https://github.com/shopsys/shopsys/blob/7.3/packages/framework/src/Model/Product/Product.php#L623

> I think it would be nice to have a new test for this behavior

You are right, however, I am not sure, if I will have time for that. I will try to, but if you want to, you can definitely write the test yourselves :innocent:  :grin:

TomasLudvik:
I did some investigation and it seems that this is only place that needs to be solved.

PR #1771, opened 2020-04-09
[frontend-api] added frontend api methods docs for apiary

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Frontend API methods were not documented, this has been fixed by this PR.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/apiary.apib
+++ b/packages/frontend-api/apiary.apib
@@ -1526,3 +1526,44 @@ Creates complete order with products and addresses
             }
         }
     }
+
+## Create order without supplying required fields
+
+### POST [POST]
+
+Returns validation errors
+
+- Request (application/json)
+    {
+        "errors": [
+            {
+                "message": "validation",
+                "extensions": {
+                    "category": "arguments_validation_error",
+                    "validation": {
+                        "input.companyName": [
+                            {
+                                "message": "Please enter company name",
+                                "code": "c1051bb4-d103-4f74-8988-acbcafc7fdc3"
+                            }
+                        ],
+                        "input.companyNumber": [
+                            {
+                                "message": "Please enter identification number",
+                                "code": "c1051bb4-d103-4f74-8988-acbcafc7fdc3"
+                            }
+                        ]
+                    }
+                },
+                "locations": [
+                    {
+                        "line": 2,
+                        "column": 9
+                    }
+                ],
+                "path": [
+                    "CreateOrder"
+                ]
+            }
+        ]
+    }
--- a/packages/frontend-api/apiary.apib
+++ b/packages/frontend-api/apiary.apib
@@ -1322,3 +1322,207 @@ Refreshes access and refresh tokens
             }
         }
     }
+
+## Create order [/graphql]
+
+### POST [POST]
+
+Creates complete order with products and addresses
+
+- Request (application/json)
+
+    mutation {
+        CreateOrder(
+            input: {
+                firstName: "firstName"
+                lastName: "lastName"
+                email: "user@example.com"
+                telephone: "+53 123456789"
+                onCompanyBehalf: true
+                companyName: "Airlocks s.r.o."
+                companyNumber: "1234"
+                companyTaxNumber: "EU4321"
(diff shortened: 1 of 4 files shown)


packages/frontend-api/apiary.apib
```diff
@@ -0,0 +1,1132 @@
...
+### POST [POST]
+
+Run a GraphQL query
+
+- Request (application/json)
+
+    mutation {
+        Login(input: {
+            email: "no-reply@shopsys.com"
+            password: "user123"
```
pesektomas, line 1116: Hmm, I think so we don't need password here

packages/frontend-api/apiary.apib
```diff
@@ -0,0 +1,1132 @@
...
+
+## Login [/graphql]
+
+### POST [POST]
+
+Run a GraphQL query
+
+- Request (application/json)
+
+    mutation {
```
pesektomas, line 1113: I think so I might like a description of each method. In the beginning, the methods have only `Runs a graphQL query.` description. 

In this mutation (login) I would also like information about the lifetime of tokens, because I think so this pair of tokens will always be expired - the mocking probably will not affect, but we should probably write info. What do you think?

packages/frontend-api/apiary.apib
```diff
@@ -0,0 +1,1132 @@
...
+                    "priceWithVat": "139.96",
+                    "priceWithoutVat": "115.67",
+                    "vatAmount": "24.29"
+                },
+                "images": [
+                    {
+                        "type": null,
+                        "position": null,
+                        "size": "default",
+                        "url": "http://127.0.0.1:8000/content/images/product/default/1.jpg",
```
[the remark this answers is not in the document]
TomasLudvik: Ok we can change it to https://demo.shopsys.com/content/images/product/default/1.jpg

packages/frontend-api/apiary.apib
TomasLudvik: I would leave it

packages/frontend-api/apiary.apib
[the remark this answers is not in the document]
ShopsysBot: Well I leave it as it is because we have no good solution for URLs right now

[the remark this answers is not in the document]
TomasLudvik: Well I leave it as it is because we have no good solution for URLs right now

packages/frontend-api/apiary.apib
TomasLudvik: updated

packages/frontend-api/apiary.apib
```diff
@@ -7,7 +7,7 @@ HOST: https://polls.apiblueprint.org/
 
 ### POST [POST]
 
-Runs a GraphQL query.
+Returns list of product that can be paginated using `first`, `last`, `before` and `after` keywords
```
pesektomas, line 10:
```suggestion
Returns list of product that can be paginated using `first`, `last`, `before` and `after` keywords. You can read more about Connection specification in [this](https://relay.dev/graphql/connections.htm) article.
```

packages/frontend-api/apiary.apib
```diff
@@ -1102,11 +1183,93 @@ Run a GraphQL query
...
+            }
+        }
+    }
+
 ## Login [/graphql]
 
 ### POST [POST]
 
-Run a GraphQL query
+Login user and return access and refresh tokens
```
pesektomas, line 1272:
```suggestion
Login user and return access and refresh tokens (Tokens in mock are expired)
```

packages/frontend-api/apiary.apib
```diff
@@ -1130,3 +1293,31 @@ Run a GraphQL query
             }
         }
     }
+
+## Refresh tokens [/graphql]
+
+### POST [POST]
+
+Refreshes access and refresh tokens
```
pesektomas, line 1301:
```suggestion
Refreshes access and refresh tokens (Tokens in mock are expired)
```

PR #1704, opened 2020-02-26
[shopsys] update symfony to version 4.4

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR updates dependencies for last LTS version of symfony (version ^4.4.0)
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes 
|Fixes issues| closes #1388, closes #907 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1776, opened 2020-04-13
[framework] fix ErrorExtractor::getAllErrorsAsArray()

vitek-rostislav:
- according to the usage of the method, the second parameter should be an array
- the bug was introduced in https://github.com/shopsys/shopsys/pull/1704
(3761f8e930931aa7a661fe3c0202ac77fa7cbe71)

| Q             | A
| ------------- | ---
|Description, reason for the PR| there is a conflict in method interface and it's usage. I had two options - change the method's interface to fit the usages, or fix the usages. I chose the first one as it seemed easier to use it this way. If you want to fix the problem another way, I am ok with it :slightly_smiling_face: Maybe `FlashMessageTrait` could provide a getter for `\Symfony\Component\HttpFoundation\Session\Flash\FlashBag` instance :thinking: 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/FlashMessage/ErrorExtractor.php
+++ b/packages/framework/src/Component/FlashMessage/ErrorExtractor.php
@@ -1,17 +1,19 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Shopsys\FrameworkBundle\Component\FlashMessage;
 
-use Symfony\Component\Form\Form;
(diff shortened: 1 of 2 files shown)


grossmannmartin:
Hi, thanks for the PR. I rebased to the current master to fix failing Travis (see #1779)

Issue #1693, opened 2020-02-20
ImageConfig doesn't accept extended entities


RostislavKreisinger:
### What is happening
When I use the following function with extended entity (entity has different class name). This will throw the exception. The class name of extended entity isn't among the registered entities.

`$imageViews = $this->imageViewFacade->getForEntityIds(Product::class, $productIds);`

`
array:10 [▼
  "Shopsys\FrameworkBundle\Model\Payment\Payment" => ImageEntityConfig {#205 ▶}
  "Shopsys\FrameworkBundle\Model\Category\Category" => ImageEntityConfig {#238 ▶}
  "Shopsys\FrameworkBundle\Model\Product\Product" => ImageEntityConfig {#153 ▶}
  "App\Model\Product\Series\ProductSeries" => ImageEntityConfig {#135 ▶}
  "Shopsys\FrameworkBundle\Model\Slider\SliderItem" => ImageEntityConfig {#124 ▶}
  "Shopsys\FrameworkBundle\Model\Transport\Transport" => ImageEntityConfig {#121 ▶}
  "Shopsys\FrameworkBundle\Model\Advert\Advert" => ImageEntityConfig {#104 ▶}
  "Shopsys\FrameworkBundle\Model\Product\Brand\Brand" => ImageEntityConfig {#107 ▶}
  "App\Model\Blog\Category\BlogCategory" => ImageEntityConfig {#102 ▶}
  "App\Model\Blog\Article\BlogArticle" => ImageEntityConfig {#95 ▶}
]
`

vitek-rostislav:
I believe this could be fixed by using `EntityNameResolver` in `ImageConfig` service.

As a workaround, is it possible to pass the original entity, @RostislavKreisinger?

RostislavKreisinger:
Yes, the solution to the problem is obvious. 

but using:
`$productClassName = 'Shopsys\FrameworkBundle\Model\Product\Product';
        $imageViews = $this->imageViewFacade->getForEntityIds($productClassName, $this->getIdsForProducts($products));`
is realy stupid.

grossmannmartin:
Thanks for reporting the issue 👍

grossmannmartin:
fixed in #1777

PR #1777, opened 2020-04-13
[shopsys] ImageConfig accepts extended entities

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| See #1693... This issue is unfixable on project, because class `ImageClass` contains extra semicolon in annotation (first commit). The extra semicolon fails in phing target annotation fix. Please accept this PR ASAP.
|New feature| No 
|Fixes issues| #1693 

### DIFF
--- a/packages/framework/src/Component/Image/Config/ImageConfig.php
+++ b/packages/framework/src/Component/Image/Config/ImageConfig.php
@@ -2,6 +2,7 @@
 
 namespace Shopsys\FrameworkBundle\Component\Image\Config;
 
+use Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver;
 use Shopsys\FrameworkBundle\Component\Image\Image;
 
 class ImageConfig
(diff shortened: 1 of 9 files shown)


TomasLudvik:
Hello, I wanted to push this PR to version 9.0.1, but I found, that changing interface of ImageConfig would cause BC break, as it is not created via Factory. I have no idea how to resolve this without BC break right now, but I am open to suggestions. 

I am setting milestone to version 10. If we do not find any other way, how to push it without BC break, it would be great to add ImageConfigFactory, so in the future such problem could be solved without BC break.

pk16011990:
I don't see changes in interface, but only in constructor (new not nullable argument -> BC Break) - is that it?

I suggest making it nullable and if it is null, this fix is not applied. Or set it via setter injection (It does not solve injection in tests).

Can you imagine it this way without a BC break?

TomasLudvik:
Yes you are right, that are the reasons. Feel free to update it to non-BC break version

PR #1789, opened 2020-04-17
[framework] fix Authenticator definition

### LINKED ISSUE #1597: Authenticator should not be dependent on `TraceableEventDispatcher`
### What is happening

`Shopsys\FrameworkBundle\Model\Security\Authenticator` class is dependent on `Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher`, which is according to [documentation](https://symfony.com/doc/current/components/event_dispatcher/traceable_dispatcher.html) more a development thing (TraceableEventDispatcher needs a `symfony/stopwatch` which is in dev-requirements only) 

We use this dispatcher directly in Authenticator class

```yaml
Shopsys\FrameworkBundle\Model\Security\Authenticator:
    arguments:
        - '@security.token_storage'
        - '@Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher'

Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher:
    arguments:
        - '@event_dispatcher'
```

EventDispatcher class should be sufficient for production use. 

### Expected result

The authenticator class is rewritten, or reasons are documented.

vitek-rostislav:
- explicit definition of "EventDispatcherInterface $eventDispatcher" argument causes production application build to fail because TraceableEventDispatcher requires Stopwatch in its constructor and the dependency is not installed in production

| Q             | A
| ------------- | ---
|Description, reason for the PR| I am not able to install application in production env (`composer install --no-dev` fails on `Cannot autowire service "Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher": argument "$stopwatch" of method Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher::__construct()" has type "Symfony\Component\Stopwatch\Stopwatch" but this class was not found.`
|New feature| No 
|Fixes issues| fixes #1597  

### DIFF
--- a/packages/framework/src/Resources/config/services.yml
+++ b/packages/framework/src/Resources/config/services.yml
@@ -498,11 +498,6 @@ services:
     Shopsys\FrameworkBundle\Model\Security\Authenticator:
         arguments:
             - '@security.token_storage'
-            - '@Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher'
-
-    Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher:
-        arguments:
-            - '@event_dispatcher'
 
     Shopsys\FrameworkBundle\Component\Filesystem\FilepathComparator: ~
 


vitek-rostislav:
also, it would be probably a good idea to automatically test `project-base` installation without dev dependencies, as we discussed with @grossmannmartin :wink:

grossmannmartin:
> also, it would be probably a good idea to automatically test `project-base` installation without dev dependencies, as we discussed with @grossmannmartin 😉

I thought about it, but this may be a tricky task because without dev dependencies there are no tools to run standards, tests, etc.

PR #1792, opened 2020-04-20
[framework] fixed phpstan error due to not existent class from different package

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In the framework we are working with FrontendApi class only in case, that bundle is installed. PHPStan ignores class_exists condition and throws an error about not the existing class. This has been resolved in this PR.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -165,7 +165,7 @@
         "codeception/phpunit-wrapper": "^8.0",
         "jakub-onderka/php-parallel-lint": "^0.9",
         "nette/utils": "^3.0",
-        "phpstan/phpstan": "^0.12",
+        "phpstan/phpstan": "^0.12.19",
         "phpstan/phpstan-doctrine": "^0.12",
         "phpstan/phpstan-phpunit": "^0.12",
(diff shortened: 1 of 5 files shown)


grossmannmartin:
class_exists in phpstan should be resolved in [v0.12.19 release](https://github.com/phpstan/phpstan/releases/tag/0.12.19)

grossmannmartin:
Hi, thanks for the PR.

A few points from testing follows.

- since v0.12.19 release of PHPStan the first ignored pattern is no longer reported and build fails on that (see https://travis-ci.org/github/shopsys/framework/builds/677717938). I suggest requiring phpstan in ^0.12.19 or `--prefer-lowest` installation fails.
-  `$user instanceof FrontendApiUser` does not emit any error or warning, when class does not exists (https://www.php.net/manual/en/internals2.opcodes.instanceof.php#108164, https://3v4l.org/dkKFN), so checking if the class exists seems unnecessary (and I'm really surprised by that)

PR #1796, opened 2020-04-20
[project-base] Free memory after each functional test

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Functional tests consume about 500MB memory. It causes fail after add few new tests on project. Problem is probably combination of #1704 (injecting services by TestContainer - this PR caused double memory useage!) and PhpUnit (https://github.com/sebastianbergmann/phpunit/issues/3915). I found small part of leaked memory - injected services into all test's instance. It is only 15% in my case, but it is better then nothing. Next precaution I suggest increase memory limit in tests.
|New feature| No 
|Fixes issues| ... 


pk16011990:
I didn't run acceptance tests on local and probably I will not try fix it in following days.

TomasLudvik:
Hi @pk16011990, I have reviewed reasons for problems with memory leaks in sources you have included and most of them talks about memory leaks in PHP 7.4. It seems, that PHP team needs to fix that.

Second of all, I have reviewed your work and it seems fine, but problem are those failing acceptance tests. I am putting this PR on hold, until that problem is solved.

Third - we have already increased memory limit for PHPUnit in phpunit.xml so until first part is resolved in PHP, I hope, that 768MB will be sufficient.

TomasLudvik:
Oh and PHP leak is reported here already: https://bugs.php.net/bug.php?id=79519

pk16011990:
Thanks for your exploration. I understand and agree with that. I will fix acceptence tests later...maybe :D

PR #1800, opened 2020-04-23
[shopsys] added double submit protection on all forms

### LINKED ISSUE #584: Admin form allows to submit form several times
### What is happening
In admin category detail I click Save and then click it several times again. Then the page reload and I see category overview and I can see that category was created multiple times - I have four same categories.

Maybe this happens to other admin forms.

### Expected result
I should be able to submit form only once even if I click button sever times.

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When someone is quicker and double-clicks the submit button, our forms are submitted twice. This caused among other things problems when sending the order, where the second form submit redirected the customer to an empty cart. It shouldn't happen now.
|New feature| Yes 
|Fixes issues| closes #584 


grossmannmartin:
also, this PR closes #584

packages/framework/assets/js/common/validation/customizeFpValidator.js
```diff
@@ -33,39 +45,50 @@ FpJsFormValidator.customizeMethods._submitForm = FpJsFormValidator.customizeMeth
 //   there can be loaded callbacks from last form submit which can cause duplicated form error windows
 // (the rest is copy&pasted from original method; eg. ajax validation)
 FpJsFormValidator.customizeMethods.submitForm = function (event) {
+
+    if ($(':focus').hasClass('js-no-validate-button')) {
+        return;
+    }
+
```
grossmannmartin, line 52: is the comment above (starting with `// Custom behavior:`) still up to date? 
Shouldn't be something from the new behavior mentioned there?

pesektomas: You are right, I added new comment.

PR #1814, opened 2020-04-28
[shopsys] unified config files extension to yaml

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #1492 we changed configuration files which lead to inconsistency with naming. Some files were named *.yml while some *.yaml. This PR unifies project config files to *.yaml (Symfony's default)
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes (backend-api installation) 
|Fixes issues| ... 


packages/frontend-api/src/Model/Order/OrderDataFactory.php
```diff
@@ -98,7 +98,7 @@ public function createOrderDataFromArgument(Argument $argument): OrderData
         }
 
         $orderData->domainId = $this->domain->getId();
-        $orderData->origin = self::ORDER_ORIGIN_GRAPHQL;
+        $orderData->origin = static::ORDER_ORIGIN_GRAPHQL;
```
TomasLudvik, line 101: Interesting change, why was this made?

grossmannmartin: While testing individual packages, I notice standards are failing in this package. And for me, it was more like a typo (as we have a rule for this).
Do you want a separate PR?

grossmannmartin:
> you have still yml extension in the descriptions in ImageType class.

Can you please paste the full path? I was unable to find what you mean. 

> There are some files, that I think are intentional

yes, those files were left intentionally. 
Apart from you mentioned, also docker related files, readthedocs/mkdocs configurations.

> think if it might be good to remove yml from Kernel::CONFIG_EXTS.

I'm not really sure. It could bring a lot of pain if someone installs a package and copies `yml` files into the config directory (or reuse some previous implementation of a feature in a new project).
It would be really hard to find, why it's not working. 
Besides, Symfony uses the same https://github.com/symfony/demo/blob/master/src/Kernel.php

PR #1817, opened 2020-04-28
[shopsys] removed webserver-bundle in favor of Symfony local server

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| `symfony/web-server-bundle` is deprecated since Symfony 4.4. This PR removes this bundle in favor of "Symfony Local Web Server" (https://symfony.com/doc/current/setup/symfony_server.html) which has more features. 
|New feature| Yes 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes (removed commands) 
|Fixes issues| ... 


docs/installation/native-installation.md
```diff
@@ -45,15 +45,28 @@ php phing build-demo-dev
 
 ## Run integrated HTTP server
 
-```sh
-php bin/console server:run
+First follow [installation instructions](https://symfony.com/doc/current/setup/symfony_server.html#installation) for the `symfony` binary which provides its own web server.
+
+Then you can run integrated webserver for the first domain:
+```bash
+symfony server:start
```
TomasLudvik, line 52: You have updated phing target server-run for that would not be better to suggest that?

TomasLudvik, line 52: Also, I think the best solution (might be the next interaction) to create a command that will provide a user with all available domains.

TomasLudvik, line 52: If you do not want to paste phing target here.... would not be better to use php bin/console server:start as we use it in other docs?

grossmannmartin: This is a good point, but I'm afraid of the confusion if for the first domain it would be phing target and for the other symfony command (as it's not possible to pass an argument into the phing exec command, afaik).

And in order to run properly second domain, `domain_urls.yml` has to be changed, so the second phing target is up to no good.

grossmannmartin: I agree, but sadly I don't know a quick&nice™ way how to achieve this behavior now.

grossmannmartin: `php bin/console server:start` was provided by symfony/webserver-bundle and this command no longer exists. Instead, the Symfony command is used. 
Although it's possible to create a command that calls `symfony server:run` internally I'm not a big fan of that solution because it counts on executable available on the machine and in the end it's just a different name and I think symfony executable provides many great features and it can be counterproductive to hide it behind our command.

_(yes, the same can be said of the phing target and I seriously consider removing that target completely)_

I searched for usages of `php bin/console server:start` in the documentation, but didn't find any. Do you have an example, please?

TomasLudvik: We have some commands with options, but I do understand it might not be part of this US. Can you please create another one?

TomasLudvik: I meant, that we do not use commands starting with symfony. I just expectted it is alias for php bin/console, which we use...

grossmannmartin: Yup, created in our backlog.

PR #1606, opened 2020-01-21
[shopsys] Redis upgrade

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There is newer version of redis client available. We want to keep our platform up to date.
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes (upgrading dependency to new major version) 
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

Issue #1002, opened 2019-05-07
remove command that build symfony kernel from composer.json


henzigo:
Command `composer install` is usually called before we have an application ready (database, redis, ... are not ready yet). But now, `post-install-cmd` and `post-update-cmd` commands calls command `Shopsys\FrameworkBundle\Command\ComposerScriptHandler` that build symfony kernel. 

Because of this, we have in our project some hacks (#988, FallbackCacheFactory).

This command only checks if `domains_urls.yml` file exists and if doesn't, it will create that.  
Probably we can call that command as phing target after all containers are started. It will allow us to remove that hacks and we will have fewer problems with that in the future.

henzigo:
open it

pk16011990:
remove stale label

grossmannmartin:
Hi @henzigo,
thank you for the issue. 

On the current master branch, the `Shopsys\FrameworkBundle\Command\ComposerScriptHandler` no longer exists, but I assume this class is not the cause of a problem. 

Do I understand properly, the main issue here are commands executed after composer install (update) that needs working connections (Redis, for example)?

I think, removal of `shopsys:domains-urls:configure` from composer will not be enough, because the main problem here is running any Symfony command.

currently, we have 
```
"shopsys:domains-urls:configure": "symfony-cmd",
"ckeditor:install --clear=skip --release=full --tag=4.5.11": "symfony-cmd",
"cache:clear --no-warmup": "symfony-cmd",
```

and any of those commands will fail without `fallbackCacheFactory` with Redis disabled.

pk16011990:
I see still problem... All this scripts (`shopsys:domains-urls:configure` etc...) needs builded container and it can cause fatal, because cache is cleared as last.

FInally the biggest paradox is: symfony command `cache:clear` needs container, but container can be broken and it needs clear cache.

In my opinion clearing of cache should be cleared without symfony command - eg by phing target with cmd command...

grossmannmartin:
may be resolved with #1606 (updated version of snc/redis-bundle)

pk16011990:
Nope, please read my previous message...

It is necessary clear cache (not by Symfony command) before any symfony command. `composer.json`'s auto-scripts in project-base does not do it...

grossmannmartin:
I see. I created #1820 to clear cache first and with combination with #1606 I believe this issue is solved.
What do you think?

pk16011990:
yes, #1820 solves it :) good job :+1:

[...]

PR #1820, opened 2020-04-30
[project-base] improved clearing cache behavior

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Cache is now cleared before any other commands in composer scripts and does not rely on working application. Built Docker image does not contain cache generated on build. 
|New feature| No 
|Fixes issues| #1002 

_Note: why it's necessary to clear the cache first? 
For example, after #1606 it's necessary to clear cache after composer update. But if any symfony command is run before clearing cache, application try to use old cache with incompatible method definition and fails on `Declaration of Redis_ca5fc0f::multi() should be compatible with Redis::multi($mode = NULL) in /var/www/html/project-base/var/cache/prod/ContainerWnR8ZiQ/Redis_ca5fc0f.php`_ 

_Note2: In Linux Kernel 3.10 with XFS filesystem and Overlay storage, this change results in an un-buildable image due to error in Linux kernel. (see https://github.com/moby/moby/issues/10294#issuecomment-223909234)_

### DIFF
--- a/project-base/docker/php-fpm/Dockerfile
+++ b/project-base/docker/php-fpm/Dockerfile
@@ -127,7 +127,7 @@ COPY --chown=www-data:www-data / /var/www/html
 
 RUN composer install --optimize-autoloader --no-interaction --no-progress --no-dev
 
-RUN php phing build-deploy-part-1-db-independent
+RUN php phing build-deploy-part-1-db-independent clean
 
 ########################################################################################################################
 
@@ -140,3 +140,5 @@ RUN composer install --optimize-autoloader --no-interaction --no-progress --dev
 RUN php phing composer-dev dirs-create test-dirs-create assets npm standards tests-unit tests-acceptance-build
 
 RUN ./bin/console shopsys:environment:change prod
+
+RUN php phing clean
--- a/composer.json
+++ b/composer.json
@@ -194,9 +194,9 @@
             "@auto-scripts"
         ],
         "auto-scripts": {
+            "php phing clean": "script",
             "shopsys:domains-urls:configure": "symfony-cmd",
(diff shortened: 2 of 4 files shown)


pk16011990:
Why do not we clear cachce by `rm -rf` (or multiplatform equivalent in phing)? It can cause still problems during development after switch between branches, because it is chicken egg problem (see https://github.com/symfony/symfony/issues/23685)

grossmannmartin:
From what I found and tested it's pretty hard to get into a state in which `clear:cache` command fails (see https://github.com/symfony/symfony/pull/23792 – rebootable kernel)

Anyway even with `clear:cache --no-warmup` or `php phing clean` as a first script I'm still getting errors on our CI while building docker image (travis is ok, local builds too) [stage CI, step php phing composer-dev ...]

```
!!
#25 6.816 !!  In AbstractTypeGenerator.php line 381:
#25 6.816 !!
#25 6.816 !!    Warning: mkdir(): Invalid argument
```

project-base/docker/php-fpm/Dockerfile
```diff
@@ -129,6 +129,8 @@ RUN composer install --optimize-autoloader --no-interaction --no-progress --no-d
 
 RUN php phing build-deploy-part-1-db-independent
 
+RUN php phing clean
```
henzigo, line 132: Why no merge `RUN php phing build-deploy-part-1-db-independent` with `clean` command together? :)

grossmannmartin: good point

grossmannmartin:
> Why do not we clear cache by `rm -rf` (or multiplatform equivalent in phing)?

I end up using `php phing clean`.

PR #1823, opened 2020-05-04
[framework] remove final from relevant classes

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Shopsys Framework uses class extendability like main way, how to override it. However few classes break is with their `final` definition. And it is inconsistent (3 DB migrations are final and others not). I didn't solve classes in tests, coding standards package and utils. Maybe it can be watched by automatic sniff like protected property, because protected property in final class is funny and sad in one time... 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Collector/ShopsysFrameworkDataCollector.php
+++ b/packages/framework/src/Component/Collector/ShopsysFrameworkDataCollector.php
@@ -12,7 +12,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 
-final class ShopsysFrameworkDataCollector extends DataCollector
+class ShopsysFrameworkDataCollector extends DataCollector
 {
     /**
      * @var \Shopsys\FrameworkBundle\Component\Domain\Domain
--- a/packages/framework/src/Component/Doctrine/MoneyType.php
+++ b/packages/framework/src/Component/Doctrine/MoneyType.php
@@ -7,7 +7,7 @@ use Doctrine\DBAL\Types\ConversionException;
 use Doctrine\DBAL\Types\Type;
 use Shopsys\FrameworkBundle\Component\Money\Money;
 
-final class MoneyType extends Type
+class MoneyType extends Type
 {
     /**
      * {@inheritDoc}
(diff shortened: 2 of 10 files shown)


TomasLudvik:
Hello @pk16011990, final is intended for Money classes, as it is directly instantiated in code and has no factory. It cannot be changed from project, so it is set as final. Removing final from migrations is similar to me as migrations are not intended to be extended, but mainly I am neutral in that case. I am ok with removing final for collectors and transformers. Can you please update your PR? Thank you

pk16011990:
No, because I don't agree. 

Money - yes, but I don't see any problem for framework, when it is not final class. When I want to create new Money with something special (eg. price from), I must create something like decorator. And for using in existing codebase use inner decorated object. I is more complicated in project base and I don't see anything what will be worst in framework without this final class. (major part of framework is protected - no encapsulation). If you want apply this strict logic (it is not intended for extending), make more class final (eg. bundle class). Now it is not consitent and this PR makes it more consitent.

Migrations - consistency. Why are few migrations final? Developer must think about it - unnecessary.

I guess, the `final` in framework was used because "hype" in the past, because few programers read good article about it and it was good reason for use it...

But I'm not maitainer of framework, If you are ok with your suggestions, feel free to do it. It is only my opinion.

TomasLudvik:
Hello @pk16011990, 

the example you had described about Money, is that you would be using it wrong. Money is always consistent and has one absolute value, that is why it is also Immutable object. It cannot have price from or anything else. 

Money represents amount of money like amount of money in your purse. It is absolute number as you can empty your purse and count it. Then you can do some things with your money (e.g. split it with your wife) represented by mathematical functions presented in Money and also some comparison functions to work with it. 

I believe that final for this class is important in order to prevent people from using it wrong. I also believe that all operations you might need to do with Money are already present there, but if I am mistaken, please let me know and we can include it in Money class.

With the rest of the class you made your point and I agree with you, that final should be removed there.

pk16011990:
> Money is always consistent and has one absolute value, that is why it is also Immutable object.
:-1: It is true, but `final` is not relevant for compliance it. It can be still true without final class. Immutable !== final class

And Shopsys Framework does not allow make Money with Currency in project (by easy way)... thanks...

TomasLudvik:
I have not said, that immutable === final, that was some misunderstanding.

I am still standing by what I said but I also am hearing what you are saying, so I have decided to accept your request.

PR #1830, opened 2020-05-08
[framework] RegisterExtendedEntitiesCompilerPass: catch MappingException

vitek-rostislav:
- see https://prnt.sc/scsxyl

| Q             | A
| ------------- | ---
|Description, reason for the PR| if there is a translatable entity in project-base (i.e. it extends AbstractTranslatableEntity), the compiler pass fails on the exception as it is not possible to load metadata for the non-entity. This PR adds a check if parent is a real entity class.
|New feature| No 
|Fixes issues| ... 


grossmannmartin:
related to #1663

pk16011990:
Extended solution can be about testing `$parentClass` if it is abstract class.

grossmannmartin:
> Extended solution can be about testing of `$parentClass` if it is abstract class.

Unfortunately, it's not possible. The same error occurred when I tried to extend any non-entity class from Shopsys namespace. 

I created a new entity that extends `\Shopsys\FrameworkBundle\Component\Csv\CsvReader` and I got the error.

PR #1842, opened 2020-05-12
[project-base] Automatically delete sessions after 7 days of user inactivity

stanoMilan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Session were never deleted from Redis and this caused some high disk space usage. This is fixed by this PR.
|New feature| No 
|Fixes issues| ... 

Delete a session after 24 hours of user inactivity, because the session stays on the server forever and takes up memory space.

### DIFF
--- a/project-base/config/packages/snc_redis.yaml
+++ b/project-base/config/packages/snc_redis.yaml
@@ -36,4 +36,5 @@ snc_redis:
             dsn: 'redis://%redis_host%'
     session:
         client: 'session'
+        ttl: 604800
         prefix: '%env(REDIS_PREFIX)%session:'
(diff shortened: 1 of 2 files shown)


henzigo:
That's not true. Sessions are deleted after 30 days. See https://github.com/shopsys/shopsys/blob/6b54709e620b80b27ad309c70c5457fcd737b817/project-base/config/packages/security.yaml#L66

pk16011990:
@henzigo, `remember_me` is performed by cookie and it is not depend on session, or how do you mean it?

grossmannmartin:
This should work (including remember me feature)
Admin will be forced to log again into administration after 24hours of inactivity even when he does not close the browser window.

pk16011990:
@s3tezsky why :-1: ? It is realy problem for production eshops, because it creates records in redis to forever (and it requires RAM). Please explaint it to me.

henzigo:
Hello, could you merge this PR? Sessions in Redis may increase more than 10GB of RAM after some time in running production because of this

stanoMilan:
@TomasLudvik  for the old session does not set the  TTL to 604800  but this sessions remains set to -1(for all time ).

s3tezsky:
> @s3tezsky why 👎 ? It is realy problem for production eshops, because it creates records in redis to forever (and it requires RAM). Please explaint it to me.

Just wanted to bring more attention here.
At first I thought we may reduce size of session first. After some digging I find out they are pretty small.

PR #1848, opened 2020-05-14
[shopsys] save generated error pages via flysystem adapter

### LINKED ISSUE #1628: error pages violates 12FactorApp principle
### What is happening
 See https://12factor.net/processes

`\Shopsys\FrameworkBundle\Component\Error\ErrorPagesFacade` saves generated error pages with Symfony's filesystem (always local).
In the case of scaled application, error pages are not shared between replicas, which can lead to unexpected behavior.

Error pages can contain some dynamic data known in the time of generation (for example articles in the header), so there is a Cron Module that generates error pages once a day.

_may be related to #950_

### Expected result
A possible solution is to use `MainFilesystem` to store/retrieve error pages and if error related to the main filesystem occurs (eg. connection to S3 fails), some pre-generated page could be rendered instead.

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Until now, generated error pages were saved into local var dir. It contradicts the principles of [12-factor app](https://12factor.net/processes). That is problem when someone has production in cluster because he couldn't upgrade error pages via cron. This PR is changing this behaviour and uses flysystem adapter for saving error pages. 
|New feature| No 
|Fixes issues| closes #1628 

### DIFF
--- a/project-base/config/paths.yaml
+++ b/project-base/config/paths.yaml
@@ -7,7 +7,7 @@ parameters:
     shopsys.domain_images_dir: '/web%shopsys.domain_images_url_prefix%'
     shopsys.domain_images_url_prefix: '/%shopsys.content_dir_name%/admin/images/domain'
     shopsys.domain_urls_config_filepath: '%shopsys.root_dir%/config/domains_urls.yaml'
-    shopsys.error_pages_dir: '%shopsys.root_dir%/var/errorPages/'
+    shopsys.error_pages_dir: '/var/errorPages/'
     shopsys.feed_dir: '/web%shopsys.feed_url_prefix%'
     shopsys.feed_url_prefix: '/%shopsys.content_dir_name%/feeds/'
     shopsys.filemanager_upload_dir: '%shopsys.web_dir%/%shopsys.filemanager_upload_web_dir%'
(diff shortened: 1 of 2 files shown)


project-base/config/paths.yaml
```diff
@@ -7,7 +7,7 @@ parameters:
     shopsys.domain_images_dir: '/web%shopsys.domain_images_url_prefix%'
     shopsys.domain_images_url_prefix: '/%shopsys.content_dir_name%/admin/images/domain'
     shopsys.domain_urls_config_filepath: '%shopsys.root_dir%/config/domains_urls.yaml'
-    shopsys.error_pages_dir: '%shopsys.root_dir%/var/errorPages/'
+    shopsys.error_pages_dir: '/var/errorPages/'
```
grossmannmartin, line 10: Is necessary to update this path? Isn't it broken in monorepo?

henzigo: Flysystem is configured to project root by default. So I need to change that because I don't want to have errorPages in path `/var/www/html/var/errorPages/` on S3 file storage. I do not know if it will break monorepo.

pk16011990: It may not work without Flysystem on localhost, but I didn't test it.

henzigo: It works :)

henzigo: @pk16011990 It's working like uploading images etc. So if there are working upload images on localhost this should also work.

grossmannmartin:
closed by https://github.com/shopsys/shopsys/pull/2609

PR #1866, opened 2020-05-22
[framework] change password in facade now do explicit flush operation

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Method in facade can be easily used separately in controller. For example moving change password into separate customer section. Mostly facades do a flush after changing data. The necessity to call an explicit flush after calling the facade method can be easily overlooked and cause troubles. Not mentioning the flush should not be called in controller (at least in my mind 😄)
|New feature| No 
|Fixes issues| ... 


TomasLudvik:
Hi, this was already added with frontend API changePassword method, closing.

PR #1874, opened 2020-05-28
[framework] Redirect URLs with a trailing slash (code 301)

RostislavKreisinger:
| Q             | A
| ------------- | ---
|Description | PR handles automatic redirection of URLs with and without a slash. This feature has been part of symfony since symfony 4, but our multilingual adjustments did not allow this redirection. https://symfony.com/doc/4.1/routing/redirect_trailing_slash.html
|New feature| Yes 
|Fixes issues| ... 


packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Routing\Exception\ResourceNotFoundException;
+
+class NormalizeUrlTrailingSlashSubscriber implements EventSubscriberInterface
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Router\CurrentDomainRouter
+     */
+    protected $router;
```
s3tezsky, line 19: _Nitpick warning_

As this property is not for any router interface I would not name it so general.

```suggestion
    protected $currentDomainRouter;
```

It is longer but more describing in further context.

packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+        if ($event->getThrowable() instanceof NotFoundHttpException) {
+            $pathInfo = $event->getRequest()->getPathInfo();
+
+            if (substr($pathInfo, -1) === '/') {
+                $pathInfo = rtrim($pathInfo, '/');
+            } else {
+                $pathInfo .= '/';
+            }
+
+            $this->redirectIfPathExists($pathInfo, $event);
```
s3tezsky, line 53: `redirectIf*` does not sound like a good method name to me.

Why just do not name it `redirectToPath` or simillar. The method do the rest (throw an exception when path does not exist).

packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * @param \Symfony\Component\HttpKernel\Event\ExceptionEvent $event
+     */
+    public function onKernelException(ExceptionEvent $event)
+    {
+        if ($event->getThrowable() instanceof NotFoundHttpException) {
+            $pathInfo = $event->getRequest()->getPathInfo();
+
+            if (substr($pathInfo, -1) === '/') {
```
s3tezsky, line 47: _nitpick warning_

I am not sure if we are using it anywhere else. I can imagine it (and folowing 2 more lines) as a part of some `UrlHelper` with methods:
- `hasTrailingSlash`
- `removeTrailingSlash`
- `addTrailingSlash`

packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+     * @param \Symfony\Component\HttpKernel\Event\ExceptionEvent $event
+     */
+    protected function redirectIfPathExists(string $newPath, ExceptionEvent $event): void
+    {
+        try {
+            $this->router->match($newPath);
+
+            $pathInfo = $event->getRequest()->getPathInfo();
+            $fullPath = $event->getRequest()->getRequestUri();
+            $pathToRedirect = str_replace($pathInfo, $newPath, $fullPath);
```
pk16011990, line 68: This is magic and It does not work with path like `www` or `http`...

pk16011990: or with same string in GET parameter

TomasLudvik: I have been testing this on full url with filter parameters, can you be more specific please?

pk16011990: I did not test it, however I guess this model situation:
```php
$fullPath = 'https://example.com/com/?foo=example.com/bar';
$pathInfo = 'com/';
$newPath = 'com';
$pathToRedirect = 'https://example.com/com?foo=example.combar'; # !!!
```
It can by solved by replace only first occur, but:
```php
$fullPath = 'https://example.com/example';
$pathInfo = 'example';
$newPath = 'example/';
$pathToRedirect = 'https://example/com/example/'; # !!!
```
Maybe it can be fixed by ignoring protocol and domain, but It can be obstacle for individual routing, when pathinfo contains some prefix like `https://example.com/en/order`. 

It is reason, why I prefer for generation url by router

packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
TomasLudvik: Done

packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
TomasLudvik: I have used redirectToExistingPath

packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
TomasLudvik: I think this is pretty straightforward so I would keep it

packages/framework/src/Component/Router/NormalizeUrlTrailingSlashSubscriber.php
TomasLudvik: I have updated it to be more accurate

PR #1829, opened 2020-05-07
[shopsys] some little seo optimalizations

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR|When someone removed category or brand with some url, someone don't have possibility to create new page with category's (or brand's) url and this url will return 404 for every time. We remove urls after delete category or brand. Next adjustment: when someone delete product, we return 410 now.  This is because we are building a platform for large ecommerce where products are spinning. When the eshop deletes a lot of products in this way and we return 410, we save the crowler budget for indexing the eshop.
|New feature| Yes 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes 
|Fixes issues| closes #1011 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1886, opened 2020-06-04
[framework] Admin can create new Category URL and remove old automaticly created URL

Swejzi:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Admin can't create own Category URL without Category generated from Category name
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Category/CategoryFacade.php
+++ b/packages/framework/src/Model/Category/CategoryFacade.php
@@ -152,7 +152,6 @@ class CategoryFacade
         }
         $this->em->flush();
         $this->friendlyUrlFacade->saveUrlListFormData('front_product_list', $category->getId(), $categoryData->urls);
-        $this->friendlyUrlFacade->createFriendlyUrls('front_product_list', $category->getId(), $category->getNames());
         $this->imageFacade->manageImages($category, $categoryData->image);
 
         $this->pluginCrudExtensionFacade->saveAllData('category', $category->getId(), $categoryData->pluginData);


paolo42:
Hello, I think this could be unwanted behavior in some cases for some people.

The best way to avoid BC break here might be to add a checkbox with label `Create also friendly URLS for the product` to the editation page. If admin unchecks the checkbox, the line removed in this PR gets skipped.

The default state of the checkbox should be `checked` to make sure this doesn't change URL behavior for people who update SSFW.

I also think Jakub Dolba tried something similar a few years back and eventually decided he didn't want to do it. You can try to ask him if he remembers anything useful.

Swejzi:
Hi, 
I don't think it is BC break, because main function is not modified and after update SSFW main behavior of app doesn't change.
This PR only allow release URL address created from category name for product for example.
For example I have same name for product "AA bateries" and category "AA bateries" and I want URL for product "/aa-bateries" and for category "/category-aa-bateries". 
I can't do this.

TomasLudvik:
Hello @Swejzi,
we actually implemented more thorough solution for this in #1829. Thank you for pointing out, that this PR has not been closed yet.

Swejzi:
Hi @TomasLudvik, 
no, https://github.com/shopsys/shopsys/pull/1829 doesn't solve my problem. 
I still can't remove URL adress /electrinics/ from category Electronics.

TomasLudvik:
Hello @Swejzi,
we have discussed it internally and we came to this:
- we agree it is not ideal situation, that you cannot delete URL for current category name, also this problem is same for products and probably other entities
- also we do not think that not creating url by current name is right, it could cause some problem
- we want to address this issue in near future, but I cannot tell you when, as we have a lot of important work in our backlog

TomasLudvik:
> @TomasLudvik
> What problem could cause for example?

For example you are not checking whether there is at least one URL.

TomasLudvik:
Ok, we have decided how we wanna solve this. We want to implement that automatically created URL for entity is when entity is renamed (so not every time it is saved). Also we wanna add checker, that during saving is at least one URL created. This should apply for all entities with friendly URLs - product, category, article and brand.

s3tezsky:
I am closing this PR as it is not solving the problem properly. The fix for this issue is in #1972.

PR #1909, opened 2020-06-30
[framework] fixed error 500 during logout when the user is already logged out

### LINKED ISSUE #769: 403 when trying to logout after my session expires

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Application did end with 500 error page when user was trying to logout, when had not valid CSRF token. This has been fixed.
|New feature| No 
|Fixes issues| closes #769 closes #1559 

### DIFF
--- a/packages/framework/src/Component/Error/LogoutExceptionSubscriber.php
+++ b/packages/framework/src/Component/Error/LogoutExceptionSubscriber.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Error;
+
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\FlashMessage\FlashMessage;
+use Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
+use Symfony\Component\HttpKernel\Event\ExceptionEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\Exception\LogoutException;
+
+class LogoutExceptionSubscriber implements EventSubscriberInterface
+{
+    /**
+     * @var \Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface
+     */
+    protected $flashBag;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser
+     */
+    protected $currentCustomerUser;
+
+    /**
+     * @var \Symfony\Component\Routing\RouterInterface
+     */
+    protected $router;
+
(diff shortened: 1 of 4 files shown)


packages/framework/src/Component/Error/LogoutExceptionSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+     * @param \Symfony\Component\HttpKernel\Event\ExceptionEvent $event
+     */
+    public function onKernelException(ExceptionEvent $event): void
+    {
+        if ($event->getThrowable() instanceof LogoutException || $event->getThrowable()->getPrevious() instanceof LogoutException) {
+            if ($this->currentCustomerUser->findCurrentCustomerUser() !== null) {
+                $domainId = $this->currentCustomerUser->findCurrentCustomerUser()->getDomainId();
+                $locale = $this->domain->getDomainConfigById($domainId)->getLocale();
+
+                $this->flashBag->add(FlashMessage::KEY_ERROR, t('There was an error during logout attempt. If you really want to sign out, please try it again.', [], 'messages', $locale));
```
henzigo, line 78: I do not like this message in Framework bundle. As you know this translation will not be dumped in project-base, but it can be shown on Frontend. So you are not able to translate this message when you will do not know about that.

TomasLudvik: Ok, we have added upgrade note to warn developers about this

henzigo: But what about new projects? There will be same issue as I described

TomasLudvik: This is pros vs cons. 

Pro is that user do not end with 500 error during logging out, which might be really confusing and lead to bad feeling about such e-commerce solution. 

Con is that in really really weird state when:
- developers did not follow UPGRADE notes, because they do not need to when upgrading to patch / minor version
- user is logging out and somehow has wrong CSRF token
- on domain that is not in both English and Czech
happens:
- message is displayed in English saying, that there was some error and he should try it again

What do you think that is more possible to happen and what do you think is worse for both user and e-commerce solution owner?

TomasLudvik: Ok I was thinking what is the problem here and I now understand - framework translations displayed on frontend must be manually added to translations in project-base so they can be translated. 

Well I never thinked about this and found several such translations in framework. This is not easy thing to solve. I will create issue for that.

TomasLudvik: #1911

henzigo: I'm ok with issue that you created :) Thank you

TomasLudvik: Great, thank YOU for bringing this up!

henzigo:
Hello, this PR didn't close issue and my PR. Can you look at it, please? :)

TomasLudvik:
Hello @henzigo, it is because this PR was merged into 9.0 branch and not master. It should close after we merge 9.0 into master. This will happen in next days.

henzigo:
Ok, thanks 👍

PR #1916, opened 2020-07-02
[framework] fixed throwing OrderItemNotFoundException

### LINKED ISSUE #1898: Wrong argument sent to OrderItemNotFoundException
### What is happening

Class `Shopsys\FrameworkBundle\Model\Order\Order` attempts to throw `OrderItemNotFoundException` on line 725 but it supplies `array` instead of `string|null` as the first argument so `TypeError` is thrown instead. 

### Expected result

Either `OrderItemNotFoundException` should override its parent construct to accept `array` as the first argument or `string` should be supplied as the first argument.

s3tezsky:
| Q             | A
| ------------- | ---
|New feature| No 
|Fixes issues| closes #1898  

### DIFF
--- a/packages/framework/src/Model/Order/Order.php
+++ b/packages/framework/src/Model/Order/Order.php
@@ -722,7 +722,10 @@ class Order
                 return $orderItem;
             }
         }
-        throw new \Shopsys\FrameworkBundle\Model\Order\Item\Exception\OrderItemNotFoundException(['id' => $orderItemId]);
+        throw new \Shopsys\FrameworkBundle\Model\Order\Item\Exception\OrderItemNotFoundException(sprintf(
+            'Order item `%s` not found.',
+            $orderItemId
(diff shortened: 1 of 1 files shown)


packages/framework/src/Model/Order/Order.php
```diff
@@ -722,7 +722,10 @@ public function getItemById($orderItemId)
                 return $orderItem;
             }
         }
-        throw new \Shopsys\FrameworkBundle\Model\Order\Item\Exception\OrderItemNotFoundException(['id' => $orderItemId]);
+        throw new \Shopsys\FrameworkBundle\Model\Order\Item\Exception\OrderItemNotFoundException(sprintf(
+            'Order item `%s` not found.',
```
TomasLudvik, line 726:
```suggestion
            'Order item id `%d` not found.',
```

PR #1922, opened 2020-07-13
[shopsys] raised phpstan analyse to level 5

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want our code to be shiny. Raising the phpstan level will help us to write better code in future.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Command/DomainInfoCommand.php
+++ b/packages/framework/src/Command/DomainInfoCommand.php
@@ -125,7 +125,7 @@ class DomainInfoCommand extends Command
 
         $propertyExtractor = new ReflectionExtractor();
         $io->writeln('You can access these properties:');
-        $io->listing($propertyExtractor->getProperties($domainConfig));
+        $io->listing($propertyExtractor->getProperties(get_class($domainConfig)));
     }
 
     /**
--- a/packages/framework/src/Component/Translation/PhpFileExtractor.php
+++ b/packages/framework/src/Component/Translation/PhpFileExtractor.php
@@ -87,8 +87,10 @@ class PhpFileExtractor implements FileVisitorInterface, NodeVisitor
     {
         if ($this->isTransMethodOrFuncCall($node)) {
             if (!$this->isIgnored($node)) {
-                $messageId = $this->getMessageId($node);
-                $domain = $this->getDomain($node);
+                /** @var \PhpParser\Node\Expr\FuncCall|\PhpParser\Node\Expr\MethodCall $transNode */
+                $transNode = $node;
+                $messageId = $this->getMessageId($transNode);
+                $domain = $this->getDomain($transNode);
 
                 $message = new Message($messageId, $domain);
                 $message->addSource(new FileSource((string)$this->file->getFilename(), $node->getLine()));
--- a/packages/framework/src/Model/Customer/User/CustomerUserRepository.php
+++ b/packages/framework/src/Model/Customer/User/CustomerUserRepository.php
@@ -76,7 +76,7 @@ class CustomerUserRepository
     {
         $customerUser = $this->findById($id);
         if ($customerUser === null) {
-            throw new \Shopsys\FrameworkBundle\Model\Customer\Exception\CustomerUserNotFoundException($id);
+            throw new \Shopsys\FrameworkBundle\Model\Customer\Exception\CustomerUserNotFoundException((string)$id);
         }
         return $customerUser;
     }
--- a/packages/framework/src/Model/Feed/FeedExportFactory.php
+++ b/packages/framework/src/Model/Feed/FeedExportFactory.php
@@ -67,11 +67,22 @@ class FeedExportFactory
     /**
      * @param \Shopsys\FrameworkBundle\Model\Feed\FeedInterface $feed
      * @param \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig $domainConfig
-     * @param string|null $lastSeekId
+     * @param int|null $lastSeekId
      * @return \Shopsys\FrameworkBundle\Model\Feed\FeedExport
      */
-    public function create(FeedInterface $feed, DomainConfig $domainConfig, ?string $lastSeekId = null): FeedExport
+    public function create(FeedInterface $feed, DomainConfig $domainConfig, $lastSeekId = null): FeedExport
     {
+        if ($lastSeekId !== null && !is_int($lastSeekId)) {
+            @trigger_error(
+                sprintf(
+                    'The argument "$lastSeekId" passed to method "%s()" should be type of int or null.'
+                    . ' Argument will be strict typed in the next major.',
+                    __METHOD__
+                ),
+                E_USER_DEPRECATED
+            );
+        }
+
         $feedRenderer = $this->feedRendererFactory->create($feed);
         $feedFilepath = $this->feedPathProvider->getFeedFilepath($feed->getInfo(), $domainConfig);
         $feedLocalFilepath = $this->feedPathProvider->getFeedLocalFilepath($feed->getInfo(), $domainConfig);
@@ -86,7 +97,7 @@ class FeedExportFactory
             $this->em,
             $feedFilepath,
             $feedLocalFilepath,
-            $lastSeekId
+            (int)$lastSeekId
         );
     }
 }
--- a/packages/framework/src/Model/Order/OrderFacade.php
+++ b/packages/framework/src/Model/Order/OrderFacade.php
@@ -261,7 +261,7 @@ class OrderFacade
      */
     public function createOrder(OrderData $orderData, OrderPreview $orderPreview, ?CustomerUser $customerUser = null)
     {
-        $orderNumber = $this->orderNumberSequenceRepository->getNextNumber();
+        $orderNumber = (string)$this->orderNumberSequenceRepository->getNextNumber();
         $orderUrlHash = $this->orderHashGeneratorRepository->getUniqueHash();
         $toFlush = [];
 
@@ -590,7 +590,12 @@ class OrderFacade
             );
 
             if ($quantifiedItemDiscount !== null) {
-                $this->addOrderItemDiscount($orderItem, $quantifiedItemDiscount, $locale, $orderPreview->getPromoCodeDiscountPercent());
+                $this->addOrderItemDiscount(
+                    $orderItem,
+                    $quantifiedItemDiscount,
+                    $locale,
+                    (float)$orderPreview->getPromoCodeDiscountPercent()
+                );
             }
         }
     }
@@ -657,7 +662,7 @@ class OrderFacade
                 $order,
                 t('Rounding', [], 'messages', $locale),
                 $orderPreview->getRoundingPrice(),
-                0,
+                '0',
                 1,
                 null,
                 null,
--- a/packages/framework/tests/Unit/Command/CheckRedisCommandTest.php
+++ b/packages/framework/tests/Unit/Command/CheckRedisCommandTest.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Tests\FrameworkBundle\Unit\Command;
 
-use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Redis;
 use Shopsys\FrameworkBundle\Command\CheckRedisCommand;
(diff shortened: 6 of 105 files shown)


packages/framework/src/Model/Customer/User/CustomerUserRepository.php
```diff
@@ -76,7 +76,7 @@ public function getCustomerUserById($id)
     {
         $customerUser = $this->findById($id);
         if ($customerUser === null) {
-            throw new \Shopsys\FrameworkBundle\Model\Customer\Exception\CustomerUserNotFoundException($id);
+            throw new \Shopsys\FrameworkBundle\Model\Customer\Exception\CustomerUserNotFoundException((string)$id);
```
TomasLudvik, line 79: Well you should probably send nothing here right?

packages/framework/src/Model/Feed/FeedExportFactory.php
```diff
@@ -67,10 +67,10 @@ public function __construct(
     /**
      * @param \Shopsys\FrameworkBundle\Model\Feed\FeedInterface $feed
      * @param \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig $domainConfig
-     * @param string|null $lastSeekId
+     * @param int|null $lastSeekId
      * @return \Shopsys\FrameworkBundle\Model\Feed\FeedExport
      */
-    public function create(FeedInterface $feed, DomainConfig $domainConfig, ?string $lastSeekId = null): FeedExport
+    public function create(FeedInterface $feed, DomainConfig $domainConfig, ?int $lastSeekId = null): FeedExport
```
TomasLudvik, line 73: This is BC break

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -269,7 +269,7 @@ public function createOrder(OrderData $orderData, OrderPreview $orderPreview, ?C
 
         $order = $this->orderFactory->create(
             $orderData,
-            $orderNumber,
+            (string)$orderNumber,
```
TomasLudvik, line 272: this would be better on line 264

packages/framework/tests/Unit/Command/CheckRedisCommandTest.php
```diff
@@ -43,21 +42,23 @@ public function testPingAllRedisClients(bool $expectSuccess, RedisFacade $redisF
     }
 
     /**
-     * @return \PHPUnit\Framework\MockObject\MockObject
+     * @return \Redis
      */
-    private function createRedisMockExpectingPing(): MockObject
+    private function createRedisMockExpectingPing(): \Redis
```
TomasLudvik, line 47:
```suggestion
    private function createRedisMockExpectingPing(): Redis
```

import it

packages/framework/tests/Unit/Model/AdvancedSearch/AdvancedSearchQueryBuilderExtenderTest.php
```diff
@@ -19,6 +19,7 @@ public function testExtendByAdvancedSearchData()
         $ruleData->operator = 'testOperator';
         $ruleData->value = 'testValue';
 
+        /** @var \Shopsys\FrameworkBundle\Model\AdvancedSearch\AdvancedSearchRuleData[]|null[] $advancedSearchData */
```
TomasLudvik, line 22: This cannot be array of nulls, right?

packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
```diff
@@ -57,7 +60,9 @@ public function testPaymentCannotBeSetForWrongType(): void
         $orderItem = $this->createOrderProduct();
 
         $this->expectException(WrongItemTypeException::class);
-        $orderItem->setPayment($this->createPaymentMock());
+        /** @var \App\Model\Payment\Payment $paymentMock */
```
TomasLudvik, line 63: What about set that createMock methods return mock or instance of this?

packages/product-feed-google/src/Model/FeedItem/GoogleFeedItemFacade.php
```diff
@@ -56,6 +56,7 @@ public function __construct(
     public function getItems(DomainConfig $domainConfig, ?int $lastSeekId, int $maxResults): iterable
     {
         $pricingGroup = $this->pricingGroupSettingFacade->getDefaultPricingGroupByDomainId($domainConfig->getId());
+        /** @var \Shopsys\FrameworkBundle\Model\Product\Product[] $products */
```
TomasLudvik, line 59: It is already set in return annotation, why is it necessary here?

project-base/build.xml
```diff
@@ -10,7 +10,7 @@
     <property name="path.framework.assets" value="${path.root}/node_modules/@shopsys/framework"/>
     <property name="path.frontend-api" value="${path.vendor}/shopsys/frontend-api"/>
 
-    <property name="phpstan.level" value="4"/>
+    <property name="phpstan.level" value="5"/>
```
TomasLudvik, line 13: I do not think it is good idea to enforce phpstan level 5 in projects

project-base/tests/App/Functional/Model/Cart/CartFacadeDeleteOldCartsTest.php
```diff
@@ -190,7 +190,7 @@ private function getCartFacadeForCustomerUser(CustomerUserIdentifier $customerUs
     /**
      * @param \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserIdentifier $customerUserIdentifier
      *
-     * @return \PHPUnit\Framework\MockObject\MockObject
+     * @return \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserIdentifierFactory
```
TomasLudvik, line 193: You should have both in return, right?

project-base/tests/App/Functional/Model/Transport/IndependentTransportVisibilityCalculationTest.php
```diff
@@ -26,7 +26,7 @@ class IndependentTransportVisibilityCalculationTest extends TransactionFunctiona
     private $independentTransportVisibilityCalculation;
 
     /**
-     * @var \Shopsys\FrameworkBundle\Model\Transport\TransportDataFactoryInterface
```
TomasLudvik, line 29: Why is it necessary to replace interfaces with class?

packages/framework/src/Command/DomainInfoCommand.php
```diff
@@ -125,7 +125,7 @@ protected function outputPropertyNotAccessible(SymfonyStyle $io, DomainConfig $d
 
         $propertyExtractor = new ReflectionExtractor();
         $io->writeln('You can access these properties:');
-        $io->listing($propertyExtractor->getProperties($domainConfig));
+        $io->listing($propertyExtractor->getProperties(get_class($domainConfig)));
```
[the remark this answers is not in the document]
s3tezsky: `getProperties()` expects string for desired class name, not an instance

packages/framework/src/Component/Translation/PhpFileExtractor.php
```diff
@@ -87,8 +87,10 @@ public function enterNode(Node $node)
     {
         if ($this->isTransMethodOrFuncCall($node)) {
             if (!$this->isIgnored($node)) {
-                $messageId = $this->getMessageId($node);
-                $domain = $this->getDomain($node);
+                /** @var \PhpParser\Node\Expr\FuncCall|\PhpParser\Node\Expr\MethodCall $transNode */
```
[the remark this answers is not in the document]
s3tezsky: We cannot do it as the method `enterNode()` have to consume interface `Node`. Then it is check if it is `FuncCall` or `MethodCall`. So this is the right place for such an anotation.

packages/framework/src/Model/Customer/User/CustomerUserRepository.php
s3tezsky: Dissagree. Why we should psas nothing here? I do not want to change the behavior so I just cast `$id` to string.

packages/framework/src/Model/Feed/FeedExportFactory.php
s3tezsky: But actually it is a fix of type mismatch. See the only usage we have and the variable previous type hint. https://github.com/shopsys/shopsys/blob/762a90cc3543b33bce1daca2feb8c48adcd3e79a/packages/framework/src/Model/Feed/FeedFacade.php#L87

packages/framework/tests/Unit/Model/AdvancedSearch/AdvancedSearchQueryBuilderExtenderTest.php
s3tezsky: Not array of nulls but there might be a null value.

https://github.com/shopsys/shopsys/blob/762a90cc3543b33bce1daca2feb8c48adcd3e79a/packages/framework/tests/Unit/Model/AdvancedSearch/AdvancedSearchQueryBuilderExtenderTest.php#L23

project-base/build.xml
s3tezsky: It is not enforcing as it is an optional update. Only new projects will start with this level so it is OK to me.

project-base/tests/App/Functional/Model/Transport/IndependentTransportVisibilityCalculationTest.php
s3tezsky: Because it is creating `TransportData` which are passed into project-base's `Transport`.

I chosed to change annotation of that factory then annotating all related variables. As it is test in project-base where the `TransportDataFactory` is overiden it is better way to me.

packages/framework/src/Model/Feed/FeedExportFactory.php
TomasLudvik: So remove the typehint and use deprecation annotation to describe what should be done in next major

packages/framework/src/Model/Order/OrderFacade.php
TomasLudvik: Thx for thumbs up, can you incorporate it?

packages/framework/tests/Unit/Command/CheckRedisCommandTest.php
TomasLudvik: You did not use this suggestion

packages/framework/tests/Unit/Model/AdvancedSearch/AdvancedSearchQueryBuilderExtenderTest.php
TomasLudvik:
```suggestion
        /** @var array $advancedSearchData */
```

packages/product-feed-google/src/Model/FeedItem/GoogleFeedItemFacade.php
TomasLudvik: Remove this and remove ArrayCollection from return of that method

project-base/build.xml
TomasLudvik: you should then say in Upgrade Notes that this is up to developer to decide if he wants to increase phpstan level

packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
```diff
@@ -26,7 +26,9 @@ public function testTransportCannotBeSetForWrongType(): void
         $orderItem = $this->createOrderPayment();
 
         $this->expectException(WrongItemTypeException::class);
-        $orderItem->setTransport($this->createTransportMock());
+
+        $transportMock = $this->createTransportMock();
+        $orderItem->setTransport($transportMock);
```
TomasLudvik, line 31: These kind of changes are no longer necessary

s3tezsky: That is right, fixed.

s3tezsky:
> but I am not sure, if we still should not think about that PHPStan extension of ours.

Everything before `but` is ***.

I really do not think we need to have custom extension as the changes you do not like (seems as unnecessary) are just really pointing to our poor code (e.g. not using factories for creating entity).

TomasLudvik (approved):

sslt (approved):

PR #1932, opened 2020-07-16
[framework] updated AdministratorFrontSecurityFacade::isAdministratorLogged() to work well if used on frontend

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When method `AdministratorFrontSecurityFacade::isAdministratorLogged()` is called on front-end (eg. in GTM layer), it causes redirect to `front_login` route. And if this calling is in login page, It causes `Too many redirections`. All exceptions catched can be thrown in `AdministratorUserProvider::refreshUser`.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Administrator/Security/AdministratorFrontSecurityFacade.php
+++ b/packages/framework/src/Model/Administrator/Security/AdministratorFrontSecurityFacade.php
@@ -60,7 +60,10 @@ class AdministratorFrontSecurityFacade
     {
         try {
             $token = $this->getAdministratorToken();
-        } catch (\Shopsys\FrameworkBundle\Model\Administrator\Security\Exception\InvalidTokenException $e) {
+        } catch (
+            \Shopsys\FrameworkBundle\Model\Administrator\Security\Exception\InvalidTokenException |
+            \Symfony\Component\Security\Core\Exception\AuthenticationException $e
+        ) {
             return false;
         }
 


packages/framework/src/Model/Administrator/Security/AdministratorFrontSecurityFacade.php
```diff
@@ -60,7 +60,12 @@ public function isAdministratorLogged()
     {
         try {
             $token = $this->getAdministratorToken();
-        } catch (\Shopsys\FrameworkBundle\Model\Administrator\Security\Exception\InvalidTokenException $e) {
+        } catch (
+            \Shopsys\FrameworkBundle\Model\Administrator\Security\Exception\InvalidTokenException |
+            \Symfony\Component\Security\Core\Exception\UnsupportedUserException |
+            \Symfony\Component\Security\Core\Exception\AuthenticationExpiredException |
+            \Symfony\Component\Security\Core\Exception\UsernameNotFoundException $e
```
s3tezsky, line 67: These exception are all extended from `Symfony\Component\Security\Core\Exception\AuthenticationException`. Even in this context it should be rater catch by the parrent one to prevent any other similar unhadled states.

s3tezsky, line 67: These exception are all extended from `Symfony\Component\Security\Core\Exception\AuthenticationException`. Even in this context it should be rater catch by the parrent one to prevent any other similar unhadled states.

TomasLudvik: Good idea, updated.

PR #1947, opened 2020-07-22
[releaser] fixed checking builds in after-release stage

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In `after-release` stage of releaser there is check for passing builds on travis. The check should validated the released version instead initial branch. This PR fixes that.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/utils/releaser/src/ReleaseWorker/AbstractCheckPackagesTravisBuildsReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/AbstractCheckPackagesTravisBuildsReleaseWorker.php
@@ -41,6 +41,11 @@ abstract class AbstractCheckPackagesTravisBuildsReleaseWorker extends AbstractSh
     }
 
     /**
+     * @return string
+     */
+    abstract protected function getCheckingBranchName(): string;
+
+    /**
      * @param \PharIo\Version\Version $version
      */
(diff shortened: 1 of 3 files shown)


utils/releaser/src/ReleaseWorker/AbstractCheckPackagesTravisBuildsReleaseWorker.php
```diff
@@ -40,12 +40,20 @@ public function getDescription(Version $version): string
         return 'Check Travis build status for all packages';
     }
 
+    /**
+     * @return string
+     */
+    abstract protected function getCheckingBranchName(): string;
```
TomasLudvik, line 46: What does the 'checkingBranch' stands for?

s3tezsky: It is name for branch you want to have checked. Maybe form the method name it is not really clear but I did not came up with better one.

TomasLudvik: Maybe just `getBranchName` or `getCurrentlyCheckedBranchName`

PR #1952, opened 2020-07-24
[framework] JS validation: fix creating of prototype of compound item in collection

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When compound item has property with reserved name (eg. `id`), it causes broken FpJsElement.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/assets/js/common/validation/customizeFpValidator.js
+++ b/packages/framework/assets/js/common/validation/customizeFpValidator.js
@@ -17,7 +17,7 @@ FpJsFormValidator.preparePrototype = function (prototype, name) {
 
     if (typeof prototype.children === 'object') {
         for (let childName in prototype.children) {
-            prototype[childName] = this.preparePrototype(prototype.children[childName], name);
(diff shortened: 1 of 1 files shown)


packages/framework/assets/js/common/validation/customizeFpValidator.js
```diff
@@ -17,7 +17,7 @@ FpJsFormValidator.preparePrototype = function (prototype, name) {
 
     if (typeof prototype.children === 'object') {
         for (let childName in prototype.children) {
-            prototype[childName] = this.preparePrototype(prototype.children[childName], name);
+            prototype.children[childName] = this.preparePrototype(prototype.children[childName], name);
```
s3tezsky, line 20:
```suggestion
            prototype.children[childName] = this.preparePrototype(childName, name);
```

I am really not a JS expert, but would not be this one more obvious?

TomasLudvik: childName is the key, so sadly no

PR #1954, opened 2020-07-24
[shopsys] Remove unused dependencies and attributes

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| removed unused dependencies and attributes to make the code cleaner
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGeneratorFacade.php
+++ b/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGeneratorFacade.php
@@ -53,7 +53,7 @@ class FriendlyUrlGeneratorFacade
     public function generateUrlsForSupportedEntities(OutputInterface $output)
     {
         foreach ($this->domain->getAll() as $domainConfig) {
-            $output->writeln(' Start of generating friendly urls for domain ' . $domainConfig->getUrl() . '');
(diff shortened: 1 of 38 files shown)


s3tezsky:
Thanks @pk16011990  for PR!

I am really suprised how many dependencies are dead.
It would be great to add [phpstan rule for detecting dead code](https://github.com/slevomat/coding-standard#cleaning---detecting-dead-code) with this cahange too.

TomasLudvik (changes requested):
As @s3tezsky suggested, it would be nice to introduce checks for this so it will be checked automatically in the future. I found this could be implemented https://psalm.dev/docs/running_psalm/configuration/#findunusedvariablesandparams

grossmannmartin:
> As @s3tezsky suggested, it would be nice to introduce checks for this so it will be checked automatically in the future. I found this could be implemented https://psalm.dev/docs/running_psalm/configuration/#findunusedvariablesandparams

I tried Psalm, but it seems like it needs some work to configure it to match our needs – I get almost 7000 errors just after installation.

Also, I'm not sure how to perform such checks automatically... It's not always possible to delete it (due to bc promise for example).

I would like to merge this cleanup as it is and maybe introduce some check in the future.

TomasLudvik (approved):
If some US would be created for this - it would make me happy

PR #1956, opened 2020-07-27
[framework] IndexExportedEvent is called after elasticsearch export cron module finished

### LINKED ISSUE #1884: Export (changed) products to Elasticsearch is infinite
Version 9.0.0

Crons `ProductExportCronModule` and `ProductExportChangedCronModule` do not mark exported products as exported. It means that once marked products are still exported by next run of  `ProductExportChangedCronModule`.

My personal opinion from debuging this: solution by events it is crazy and for me it has not good DX.

Expected result: Product is marked as exported ASAP after export (not after whole export - it can take tens minutes)

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| If I export data to the ES, the data is never set as exported...
|New feature| No 
|Fixes issues|

### DIFF
--- a/packages/framework/src/Component/Elasticsearch/AbstractExportChangedCronModule.php
+++ b/packages/framework/src/Component/Elasticsearch/AbstractExportChangedCronModule.php
@@ -4,10 +4,12 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Component\Elasticsearch;
 
+use BadMethodCallException;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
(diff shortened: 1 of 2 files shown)


TomasLudvik:
Hi @pesektomas, setter needs to meet some requirements for example @require is missing - here is example how it should be done: https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Product/ProductDataFactory.php#L131

pk16011990:
I think It does not solve all in #1884. Because export to elasticsearch can take tens minutes but marking products as exported is called at the end. This might cause some product changes not to be exported.

PR #1961, opened 2020-07-29
[symfony] added phpstan-symfony extension

### LINKED ISSUE #1059: Use phpstan-symfony

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| #1059 
|New feature| Yes 
|Fixes issues| closes #1059 

### DIFF
--- a/project-base/tests/FrontendApiBundle/Functional/Product/ProductVariantTest.php
+++ b/project-base/tests/FrontendApiBundle/Functional/Product/ProductVariantTest.php
@@ -11,12 +11,12 @@ use Tests\FrontendApiBundle\Test\GraphQlTestCase;
 class ProductVariantTest extends GraphQlTestCase
 {
     /**
-     * @var \Shopsys\FrameworkBundle\Model\Product\Product
+     * @var \App\Model\Product\Product
      */
     private $productAsMainVariant;
 
     /**
-     * @var \Shopsys\FrameworkBundle\Model\Product\Product
+     * @var \App\Model\Product\Product
      */
     private $productAsVariant;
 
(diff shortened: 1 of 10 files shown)


project-base/tests/FrontendApiBundle/Functional/Product/ProductVariantTest.php
```diff
@@ -11,12 +11,12 @@
 class ProductVariantTest extends GraphQlTestCase
 {
     /**
-     * @var \App\Model\Product\Product
+     * @var \Shopsys\FrameworkBundle\Model\Product\Product
```
s3tezsky, line 14: This seems weird to me.

I undestand the purpose of this change but IMHO having annotaded assigning value to this property will do better. This applies to other same changes.

TomasLudvik: Ok, you made me think about this and also test it on one of our projects. The problem here is that ProductFacade on line 27 and 28 here in the test returns \Shopsys\FrameworkBundle\Model\Product\Product which is not accepted by \App\Model\Product\Product, but it works in opposite case.

There are no other problems like this in running projects, so this two cases are only ones.

TomasLudvik:
I have updated build.xml files

project-base/tests/FrontendApiBundle/Functional/Product/ProductVariantTest.php
s3tezsky: Does any project use FEAPI?

In project-base you alreaady know that Product entity has been extended so it will be resolved everytime. IMHO the setup should look like this.

```
protected function setUp(): void
    {
        $productFacade = $this->getContainer()->get(ProductFacade::class);

        /** @var \App\Model\Product\Product $product */
        $productAsMainVariant = $productFacade->getById(150);
        $this->productAsMainVariant = $productAsMainVariant;

        /** @var \App\Model\Product\Product $productAsVariant */
        $productAsVariant = $productFacade->getById(75);
        $this->productAsVariant = $productAsVariant;

        parent::setUp();
    }
```

TomasLudvik: changed

PR #1960, opened 2020-07-29
[frontend-api] Added resolvers for orders

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR adds posibility to retreive order list or single order for given customer user or combination of uuid and urls hash.
|New feature| Yes 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| No 
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #1964, opened 2020-07-30
[shopsys] fixed filtering of products in frontend API

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Keyword last was working wrong way. This was caused by wrong implementation of Paginator. This has been solved in this PR.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
+++ b/packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
@@ -200,6 +200,24 @@ class ProductOnCurrentDomainFacade implements ProductOnCurrentDomainFacadeInterf
     }
 
     /**
+     * @return int
+     */
+    public function getProductsCountOnCurrentDomain(): int
+    {
+        $queryBuilder = $this->productRepository->getAllListableQueryBuilder(
+            $this->domain->getId(),
+            $this->currentCustomerUser->getPricingGroup()
+        );
+
+        $this->productRepository->addTranslation($queryBuilder, $this->domain->getLocale());
+
+        return $queryBuilder
+            ->select('count(p.id)')
+            ->getQuery()
+            ->getSingleScalarResult();
+    }
+
+    /**
      * @param string $orderingModeId
      * @param int $page
      * @param int $limit
--- a/packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
+++ b/packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
@@ -209,8 +209,6 @@ class ProductOnCurrentDomainFacade implements ProductOnCurrentDomainFacadeInterf
             $this->currentCustomerUser->getPricingGroup()
         );
 
-        $this->productRepository->addTranslation($queryBuilder, $this->domain->getLocale());
-
         return $queryBuilder
             ->select('count(p.id)')
             ->getQuery()
--- a/packages/framework/src/Model/Product/ProductOnCurrentDomainFacadeInterface.php
+++ b/packages/framework/src/Model/Product/ProductOnCurrentDomainFacadeInterface.php
@@ -123,4 +123,9 @@ interface ProductOnCurrentDomainFacadeInterface
      * @return array
      */
     public function getProductsOnCurrentDomain(int $limit, int $offset, string $orderingModeId): array;
+
+    /**
+     * @return int
+     */
+    public function getProductsCountOnCurrentDomain(): int;
 }
--- a/project-base/tests/FrontendApiBundle/Functional/Product/ProductsTest.php
+++ b/project-base/tests/FrontendApiBundle/Functional/Product/ProductsTest.php
@@ -121,4 +121,43 @@ class ProductsTest extends GraphQlTestCase
 
         return $arrayExpected;
     }
+
+    public function testLastProduct(): void
+    {
+        $query = '
+            query {
+                products (last: 1) {
+                    edges {
+                        node {
+                            name
+                        }
+                    }
+                }
+            }
+        ';
+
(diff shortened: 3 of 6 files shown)


packages/framework/src/Model/Product/ProductOnCurrentDomainFacadeInterface.php
```diff
@@ -123,4 +123,9 @@ public function getProductsByCategory(Category $category, int $limit, int $offse
      * @return array
      */
     public function getProductsOnCurrentDomain(int $limit, int $offset, string $orderingModeId): array;
+
+    /**
+     * @return int
+     */
+    public function getProductsCountOnCurrentDomain(): int;
```
s3tezsky, line 130: Should it be called `getAllListableProductCount...`?

packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
```diff
@@ -199,6 +199,24 @@ public function getProductsOnCurrentDomain(int $limit, int $offset, string $orde
...
+     * @return int
+     */
+    public function getProductsCountOnCurrentDomain(): int
+    {
+        $queryBuilder = $this->productRepository->getAllListableQueryBuilder(
+            $this->domain->getId(),
+            $this->currentCustomerUser->getPricingGroup()
+        );
+
+        $this->productRepository->addTranslation($queryBuilder, $this->domain->getLocale());
```
s3tezsky, line 212: Why do we need translation here?

packages/framework/src/Model/Product/ProductOnCurrentDomainFacadeInterface.php
TomasLudvik: I do not think so, as no method in this facade has it in name and most of them are getting data from listable methods in repository

packages/framework/src/Model/Product/ProductOnCurrentDomainFacade.php
TomasLudvik: 798a42377d9d1ac2c134986715dad926fb6c35e6

TomasLudvik:
I have also added another fixup (c3287e4f5d657ef15ee50f246245d74b6fabcd86) for solving defult limit + resolving for categories.

TomasLudvik:
> Nice fix. Could you please write some test to prevent it will not happen again?

Sorry I do not known how could I write test preventing wrong usage of Paginator in resolvers. I am open to ideas.

TomasLudvik:
> Nice fix. Could you please write some test to prevent it will not happen again?

added

packages/framework/src/Model/Product/ProductOnCurrentDomainFacadeInterface.php
s3tezsky: > no method in this facade has it in name and most of them are getting data from listable methods in repository

Yeah and it is really confusing. The interface/facade name is `ProductOnCurrentDomain...` not `ListableProductOnCurrentDomain...`.

And also you wrote the second confusing thing - `most of them are getting data from listable methods`. It seems wrong to me as I cannot be sure what products will I get unless I walk through the code.

project-base/tests/FrontendApiBundle/Functional/Product/ProductsTest.php
```diff
@@ -121,4 +121,43 @@ private function getExpectedDataForNineteenthProduct(): array
...
+                            name
+                        }
+                    }
+                }
+            }
+        ';
+
+        $jsonExpected = $this->getExpectedDataForLastProduct();
+
+        $this->assertQueryWithExpectedJson($query, $jsonExpected);
```
s3tezsky, line 141: Maybe we should discuss the way we are writing tests for the API.

I like this one as it is easy to write but I am worried about bulletproof in future. Against this one I wrote more prozaic test which takes a bit more time to write/read but it is testing more than just product. https://github.com/shopsys/shopsys/pull/1960/commits/7b57ab628568b1e138308aeb88f9529862d32daf

packages/framework/src/Model/Product/ProductOnCurrentDomainFacadeInterface.php
TomasLudvik: Well you should think of it from other side - OnCurrentDomainFacades are working with products, that can be displayed (are not hidden, have price name...)

project-base/tests/FrontendApiBundle/Functional/Product/ProductsTest.php
TomasLudvik: I wanted to use that, but that has been done in 9.1 and is not available in 9.0

packages/framework/src/Model/Product/ProductOnCurrentDomainFacadeInterface.php
s3tezsky: I know that somehow (deep in myself) but on other hand the name is really confusing here. But fixing it would take more than just renaming one method 🤔 .

Think of it from other side - you do not know what is behind this interface, you will use it, because it gives you products for given domain (fine - I know there are 50 products on any domain). Now comes a surprise  - the method will give me only 40 of them, because it is not obvious to mi it is limitting hidden products and some others as well. 😞 

I guess we cannot do anything to improve it here and now. But we should really thing about the naming and maybe consider renaming methods or the whole interface.

project-base/tests/FrontendApiBundle/Functional/Product/ProductsTest.php
s3tezsky: Ok, that is fair enough 👍

PR #1966, opened 2020-07-31
[framework] fixed clearing form options in product picker

### LINKED ISSUE #1187: Allowed choice of variant product in product picker window after search with not allowed variants
### What is happening
I use `Shopsys\FrameworkBundle\Form\ProductsType` in form with option `allow_variants` set to `false`.

So when I open products picker window, I can't choose variant - it is correct behavior.

But when I use search in that same product picker window, I can choose variant from searched results - it is not correct behavior, because I set `allow_variants` to `false`.

### Expected result
I can't choose variant product after use search in product picker window (with using `ProductsType` with option `allow_variants` set to `false`).

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixes linked issue
|New feature| No 
|Fixes issues| closes #1187 

### DIFF
--- a/packages/framework/src/Controller/Admin/ProductPickerController.php
+++ b/packages/framework/src/Controller/Admin/ProductPickerController.php
@@ -63,11 +63,13 @@ class ProductPickerController extends AdminBaseController
     }
 
     /**
-     * @Route("/product-picker/pick-multiple/{jsInstanceId}/")
+     * @Route("/product-picker/pick-multiple/{jsInstanceId}/{allowMainVariants}/{allowVariants}")
      * @param \Symfony\Component\HttpFoundation\Request $request
      * @param string $jsInstanceId
+     * @param bool $allowMainVariants
+     * @param bool $allowVariants
      */
-    public function pickMultipleAction(Request $request, $jsInstanceId)
+    public function pickMultipleAction(Request $request, $jsInstanceId, bool $allowMainVariants=true, bool $allowVariants=true)
     {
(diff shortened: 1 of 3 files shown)


packages/framework/src/Controller/Admin/ProductPickerController.php
```diff
@@ -63,12 +63,18 @@ public function __construct(
...
-     * @Route("/product-picker/pick-multiple/{jsInstanceId}/")
+     * @Route("/product-picker/pick-multiple/{jsInstanceId}/{allowMainVariants}/{allowVariants}")
      * @param \Symfony\Component\HttpFoundation\Request $request
      * @param string $jsInstanceId
+     * @param bool $allowMainVariants
+     * @param bool $allowVariants
      */
-    public function pickMultipleAction(Request $request, $jsInstanceId)
-    {
+    public function pickMultipleAction(
```
TomasLudvik, line 72: This will not work with strict_types = 1, you are right now changing type, please update it

s3tezsky: eh? I do not understaand this. What type am I changing?

On other hand here is not used `strict_types=1`.

s3tezsky: Actually even with strict_types it workl like a charm. I am leaving it as is.

TomasLudvik: ok

PR #1971, opened 2020-08-03
[framework] fixed joining the same table with different alias

### LINKED ISSUE #1417: QueryBuilderExtender ignores different alias for joined table
### What is happening
When I call `QueryBuilderExtender::addOrExtendJoin()` with join which exists in query builder but with another alias, QueryBuilderExtender extends join instead of adding new join with different alias. Condition for new join contains new alias but It does not exist...

### Expected result
Passing test:
```php
public function testDifferentAlias()
{
    /** @var \Doctrine\ORM\EntityManager $entityManager */
    $entityManager = $this->getMockBuilder(EntityManagerInterface::class)
        ->disableOriginalConstructor()
        ->getMockForAbstractClass();

    $queryBuilder = new QueryBuilder($entityManager);
    $queryBuilder
        ->select('c')
        ->from(Category::class, 'c')
        ->join(Product::class, 'p', Join::WITH, 'p.id = 1');

    $queryBuilderExtender = new QueryBuilderExtender();
    $queryBuilderExtender->addOrExtendJoin($queryBuilder, Product::class, 'p2', 'p2.id = 2');

    $dql = $queryBuilder->getDQL();
    $this->assertSame('SELECT c FROM ' . Category::class . ' c INNER JOIN ' . Product::class . ' p WITH p.id = 1 INNER JOIN ' . Product::class . ' p2 WITH p2.id = 2', $dql);
}
```
not: 
```
1) Tests\FrameworkBundle\Unit\Component\Doctrine\QueryBuilderExtenderTest::testDifferentAlias
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'SELECT c FROM Shopsys\FrameworkBundle\Model\Category\Category c INNER JOIN Shopsys\ShopBundle\Model\Product\Product p WITH p.id = 1 INNER JOIN Shopsys\ShopBundle\Model\Product\Product p2 WITH p2.id = 2'
+'SELECT c FROM Shopsys\FrameworkBundle\Model\Category\Category c INNER JOIN Shopsys\ShopBundle\Model\Product\Product p WITH p.id = 1 WHERE p2.id = 2'

/var/www/html/packages/framework/tests/Unit/Component/Doctrine/QueryBuilderExtenderTest.php:36
```

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| see #1417
|New feature| No 
|Fixes issues| closes #1417 

### DIFF
--- a/packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
+++ b/packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
@@ -11,13 +11,21 @@ class DuplicatedAliasException extends LogicException
 {
     /**
      * @param string $alias
+     * @param string $aliasingClass
+     * @param string $aliasedClass
      * @param \Throwable|null $previous
      */
-    public function __construct(string $alias, ?Throwable $previous = null)
-    {
+    public function __construct(
+        string $alias,
+        string $aliasingClass,
+        string $aliasedClass,
+        ?Throwable $previous = null
+    ) {
         $message = sprintf(
-            'Alias "%s" is already assigned to different entity.',
-            $alias
+            'You cannot use "%s" as "%s" because this alias is already assigned to "%s" in the same QueryBuilder instance.',
+            $aliasingClass,
+            $alias,
+            $aliasedClass
         );
         parent::__construct(
             $message,
--- a/packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
+++ b/packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
@@ -11,10 +11,9 @@ class DuplicatedAliasException extends LogicException
 {
     /**
      * @param string $alias
-     * @param int|null $code
      * @param \Throwable|null $previous
      */
-    public function __construct(string $alias, ?int $code = 0, ?Throwable $previous = null)
+    public function __construct(string $alias, ?Throwable $previous = null)
     {
         $message = sprintf(
             'Alias "%s" is already assigned to different entity.',
@@ -22,7 +21,7 @@ class DuplicatedAliasException extends LogicException
         );
         parent::__construct(
             $message,
-            $code,
+            0,
             $previous
         );
     }
--- a/packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
+++ b/packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Doctrine\Exception;
+
+use LogicException;
+use Throwable;
+
+class DuplicatedAliasException extends LogicException
+{
+    /**
+     * @param string $alias
+     * @param int|null $code
+     * @param \Throwable|null $previous
+     */
+    public function __construct(string $alias, ?int $code = 0, ?Throwable $previous = null)
+    {
+        $message = sprintf(
+            'Alias "%s" is already assigned to different entity.',
+            $alias
+        );
+        parent::__construct(
+            $message,
+            $code,
+            $previous
+        );
+    }
+}
--- a/packages/framework/src/Component/Doctrine/QueryBuilderExtender.php
+++ b/packages/framework/src/Component/Doctrine/QueryBuilderExtender.php
@@ -37,8 +37,7 @@ class QueryBuilderExtender
         $joinAlreadyUsed = false;
         $resolvedClass = $this->entityNameResolver->resolve($class);
         foreach ($joins as $join) {
-            $resolvedJoinClass = $this->entityNameResolver->resolve($join->getJoin());
-            if ($resolvedJoinClass === $resolvedClass) {
+            if ($join->getAlias() === $alias) {
                 $joinAlreadyUsed = true;
                 break;
             }
--- a/packages/framework/src/Component/Doctrine/QueryBuilderExtender.php
+++ b/packages/framework/src/Component/Doctrine/QueryBuilderExtender.php
@@ -4,6 +4,7 @@ namespace Shopsys\FrameworkBundle\Component\Doctrine;
 
 use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
+use Shopsys\FrameworkBundle\Component\Doctrine\Exception\DuplicatedAliasException;
 use Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver;
 
 class QueryBuilderExtender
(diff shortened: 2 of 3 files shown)


packages/framework/tests/Unit/Component/Doctrine/QueryBuilderExtenderTest.php
```diff
@@ -91,4 +92,35 @@ public function extendJoinWithExtendedEntityProvider(): array
...
+        );
+
+        $dql = $queryBuilder->getDQL();
+        $this->assertSame(
+            'SELECT c FROM ' . Category::class . ' c'
+            . ' INNER JOIN ' . Product::class . ' p WITH p.id = 1'
+            . ' INNER JOIN ' . Product::class . ' p2 WITH p2.id = 2',
+            $dql
+        );
+    }
```
grossmannmartin, line 125: 😍

s3tezsky: Yeah! Thanks to @pk16011990 for excelent issue 👏

packages/framework/src/Component/Doctrine/QueryBuilderExtender.php
```diff
@@ -37,8 +37,7 @@ public function addOrExtendJoin(QueryBuilder $queryBuilder, $class, $alias, $con
         $joinAlreadyUsed = false;
         $resolvedClass = $this->entityNameResolver->resolve($class);
         foreach ($joins as $join) {
-            $resolvedJoinClass = $this->entityNameResolver->resolve($join->getJoin());
-            if ($resolvedJoinClass === $resolvedClass) {
+            if ($join->getAlias() === $alias) {
```
pk16011990, line 40: I think there should be compared alias and joined class together.

s3tezsky: Why do you think so? I think comparing alias only is enough as you cannot have same alias even for different classes.

grossmannmartin: Hi,
I have similar thoughts after testing.

Imagine the following code:

```php
$qb->select('oi.id')
    ->from(OrderItem::class, 'oi');

$queryBuilderExtender->addOrExtendJoin($qb, Product::class, 'p', 'oi.product = p.id');
$queryBuilderExtender->addOrExtendJoin($qb, Payment::class, 'p', 'oi.payment = p.id');
```
 resulting SQL is not what I expect (no join on payment)
```sql
SELECT o0_.id AS id_0 
FROM order_items o0_ 
INNER JOIN products p1_ ON (o0_.product_id = p1_.id) 
WHERE o0_.payment_id = p1_.id
```

now it's a pretty obvious mistake, but `addOrExtendJoin` can be in a different class, and then it could be really confusing because no error/warning is raised and code just doesn't do what you expect.

s3tezsky: Thanks for testing it this way! I have added one more test and fix this issue https://github.com/shopsys/shopsys/pull/1971/commits/3d1b7b943783c10afbf155ab43761d3d1f35eb88

packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
```diff
@@ -0,0 +1,29 @@
...
+use Throwable;
+
+class DuplicatedAliasException extends LogicException
+{
+    /**
+     * @param string $alias
+     * @param int|null $code
+     * @param \Throwable|null $previous
+     */
+    public function __construct(string $alias, ?int $code = 0, ?Throwable $previous = null)
```
grossmannmartin, line 17: is the `$code` argument necessary? Usually, it's not used in extended exceptions (see other exceptions in the same namespace)

packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
```diff
@@ -0,0 +1,29 @@
...
+    {
+        $message = sprintf(
+            'Alias "%s" is already assigned to different entity.',
+            $alias
+        );
+        parent::__construct(
+            $message,
```
grossmannmartin, line 24: Personally, I would prefer to have the sprintf on one line here, but as it's only a matter of personal preference I leave it up to you if you change it or not.

packages/framework/src/Component/Doctrine/QueryBuilderExtender.php
grossmannmartin: nice!

packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * @param string $alias
+     * @param int|null $code
+     * @param \Throwable|null $previous
+     */
+    public function __construct(string $alias, ?int $code = 0, ?Throwable $previous = null)
+    {
+        $message = sprintf(
+            'Alias "%s" is already assigned to different entity.',
```
grossmannmartin, line 20: I'm thinking about a more verbose message, so it's really clear what's going on. 
What do you think of 

```php
sprintf('You cannot use "%s" as "%s" because this alias is already assigned to "%s" in the same QueryBuilder instance.',
    $resolvedClass,
    $alias,
    $resolvedJoinClass
);
```

`You cannot use "App\Model\Payment\Payment" as "p" because this alias is already assigned to "App\Model\Product\Product" in the same QueryBuilder instance.`

packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
s3tezsky: Yeah you are right.

packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
s3tezsky: Usually I would do it the way you are describing. I did not because the sprintf method is kind a large (and I personally do not like multiline function as an argument) and maybe will be even larger as you are suggesting change the message.

packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
s3tezsky: Nice point. I gues it will be the most descriptive exception in FW 😄

packages/framework/src/Component/Doctrine/Exception/DuplicatedAliasException.php
grossmannmartin: 👌

PR #1972, opened 2020-08-03
[framework] fixed removing friendly urls

### LINKED ISSUE #668: [framework] Custom URL for Category

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There was not any way how to remove default friendly URL for entities (product, category, brand or article). This PR makes it possible unless the entity is renamed.
|New feature| No 
|Fixes issues| closes #668  , also solves already closed PR #669

### DIFF
--- a/packages/framework/src/Component/Domain/Domain.php
+++ b/packages/framework/src/Component/Domain/Domain.php
@@ -155,9 +155,9 @@ class Domain implements DomainIdsProviderInterface
     }
 
     /**
-     * @return int[][]
+     * @return array
      */
-    public function getDomainIdsByLocale(): array
+    public function getDomainIdsIndexedByLocale(): array
     {
         $domainIdsByLocale = [];
         foreach ($this->domainConfigs as $domainConfig) {
--- a/packages/framework/src/Component/Domain/Domain.php
+++ b/packages/framework/src/Component/Domain/Domain.php
@@ -155,18 +155,6 @@ class Domain implements DomainIdsProviderInterface
     }
 
     /**
-     * @return array
-     */
-    public function getDomainIdsIndexedByLocale(): array
-    {
-        $domainIdsByLocale = [];
-        foreach ($this->domainConfigs as $domainConfig) {
-            $domainIdsByLocale[$domainConfig->getLocale()][] = $domainConfig->getId();
-        }
-        return $domainIdsByLocale;
-    }
-
-    /**
      * @param int $domainId
      */
     public function switchDomainById($domainId)
--- a/packages/framework/src/Component/Domain/Domain.php
+++ b/packages/framework/src/Component/Domain/Domain.php
@@ -155,6 +155,18 @@ class Domain implements DomainIdsProviderInterface
     }
 
     /**
+     * @return int[][]
+     */
+    public function getDomainIdsByLocale(): array
+    {
+        $domainIdsByLocale = [];
+        foreach ($this->domainConfigs as $domainConfig) {
+            $domainIdsByLocale[$domainConfig->getLocale()][] = $domainConfig->getId();
+        }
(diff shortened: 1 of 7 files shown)


packages/framework/src/Component/Domain/Domain.php
```diff
@@ -154,6 +154,18 @@ public function getDomainConfigById($domainId)
         throw new \Shopsys\FrameworkBundle\Component\Domain\Exception\InvalidDomainIdException();
     }
 
+    /**
+     * @return int[][]
```
TomasLudvik, line 158:
```suggestion
     * @return int[]
```

We are trying to avoid this

packages/framework/src/Model/Category/CategoryFacade.php
```diff
@@ -435,4 +438,27 @@ public function getListableProductCountsIndexedByCategoryId(
             $domainId
         );
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Category\Category $category
+     * @param array $originalNames
+     */
+    protected function createFriendlyUrlWhenRenamed(Category $category, array $originalNames): void
```
TomasLudvik, line 446: This code is too complex, better would be to create only array of changed names and createUrls same as it was done in previous version on line 155.

packages/framework/src/Component/Domain/Domain.php
s3tezsky: Ok I will chaange that to `array` as your suggestion is not valid.

packages/framework/src/Model/Category/CategoryFacade.php
s3tezsky: In previous version frienly urls were creaated for all domains every time.

Anyway, I reduced complexity in https://github.com/shopsys/shopsys/pull/1972/commits/cada8e61ed46f9ba8764d5d84d674a005564175b

packages/framework/src/Model/Category/CategoryFacade.php
```diff
@@ -435,4 +438,33 @@ public function getListableProductCountsIndexedByCategoryId(
...
+    {
+        $this->friendlyUrlFacade->createFriendlyUrls(
```
TomasLudvik, line 448: You should call this only when some names are available

packages/framework/src/Model/Category/CategoryFacade.php
```diff
@@ -435,4 +438,33 @@ public function getListableProductCountsIndexedByCategoryId(
             $domainId
         );
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Category\Category $category
+     * @param array $originalNames
+     */
+    protected function createFriendlyUrlWhenRenamed(Category $category, array $originalNames): void
```
TomasLudvik, line 446:
```suggestion
    protected function createFriendlyUrlsWhenRenamed(Category $category, array $originalNames): void
```

s3tezsky: fixed

packages/framework/src/Model/Category/CategoryFacade.php
s3tezsky: fixed

PR #1985, opened 2020-08-11
[shopsys] redis build-version now includes application environment

### LINKED ISSUE #860: Run tests with different Redis database
### What is happening
Tests are running on the same database as the shop is running.
Tests first cleans the DB (the shop cache), and then runs themselves - so they are not affected by cache state but can affect the shop after they finishes. And just cleaning cache after tests finishes is not optimal solution in my opinion.

### Expected result
Tests should run in different DB than shop.

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Application was using same redis tables for each environment and this could result in unexpected problems. This has been solved by adding environment into build-version and using it for all redis tables.
|New feature| No 
|Fixes issues| closes #860 

### DIFF
--- a/project-base/config/packages/snc_redis.yaml
+++ b/project-base/config/packages/snc_redis.yaml
@@ -37,4 +37,4 @@ snc_redis:
     session:
         client: 'session'
         ttl: 604800
-        prefix: '%env(REDIS_PREFIX)%%build-version%session:'
+        prefix: '%env(REDIS_PREFIX)%session:'
--- a/project-base/config/packages/snc_redis.yaml
+++ b/project-base/config/packages/snc_redis.yaml
@@ -29,7 +29,7 @@ snc_redis:
             alias: 'global'
             dsn: 'redis://%redis_host%'
             options:
-                prefix: '%env(REDIS_PREFIX)%'
+                prefix: '%env(REDIS_PREFIX)%%build-version%'
         session:
             type: 'phpredis'
             alias: 'session'
@@ -37,4 +37,4 @@ snc_redis:
     session:
         client: 'session'
         ttl: 604800
-        prefix: '%env(REDIS_PREFIX)%session:'
+        prefix: '%env(REDIS_PREFIX)%%build-version%session:'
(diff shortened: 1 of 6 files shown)


project-base/config/packages/snc_redis.yaml
```diff
@@ -29,12 +29,12 @@ snc_redis:
...
+                prefix: '%env(REDIS_PREFIX)%%build-version%'
         session:
             type: 'phpredis'
             alias: 'session'
             dsn: 'redis://%redis_host%'
     session:
         client: 'session'
         ttl: 604800
-        prefix: '%env(REDIS_PREFIX)%session:'
+        prefix: '%env(REDIS_PREFIX)%%build-version%session:'
```
grossmannmartin, line 40: I'm not sure, but this log out all users after deploy, or?

project-base/config/packages/snc_redis.yaml
```diff
@@ -29,12 +29,12 @@ snc_redis:
             alias: 'global'
             dsn: 'redis://%redis_host%'
             options:
-                prefix: '%env(REDIS_PREFIX)%'
+                prefix: '%env(REDIS_PREFIX)%%build-version%'
```
grossmannmartin, line 32: Hmm, I'm not sure if this solves the problem mentioned in the issue. 

Checks on CI (see `.ci/build_kubernetes.sh`) are run in the production environment anyway, so it will be still problematic.
Same we recommend in our documentation, https://docs.shopsys.com/en/9.0/cookbook/jenkins-configuration/ and 
indirectly in https://docs.shopsys.com/en/9.0/kubernetes/continuous-integration-using-kubernetes/

Internal projects run tests in the development environment, so for our teams it probably helps.
Maybe we should change the recommendations for running tests and how we run them ourselves.

project-base/config/packages/snc_redis.yaml
TomasLudvik: You are right

project-base/config/packages/snc_redis.yaml
TomasLudvik: But checks do change the environment to test so it should be ok.

s3tezsky (approved):

sslt (approved):

PR #1990, opened 2020-08-17
[project-base] added condition for get accessories if module is enabled only

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The accessories is load with every addProductAjaxAction request now. I added condition into controller for get accessories if module is enabled only. Condition in view is to late in my opinion. 
|New feature| No
|Fixes issues| ... 

### DIFF
--- a/project-base/src/Controller/Front/CartController.php
+++ b/project-base/src/Controller/Front/CartController.php
@@ -10,6 +10,7 @@ use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\FlashMessage\ErrorExtractor;
 use Shopsys\FrameworkBundle\Model\Cart\AddProductResult;
 use Shopsys\FrameworkBundle\Model\Cart\CartFacade;
+use Shopsys\FrameworkBundle\Model\Module\ModuleFacade;
 use Shopsys\FrameworkBundle\Model\Module\ModuleList;
 use Shopsys\FrameworkBundle\Model\Order\Preview\OrderPreviewFactory;
 use Shopsys\FrameworkBundle\Model\Product\Product;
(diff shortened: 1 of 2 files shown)


project-base/src/Controller/Front/CartController.php
```diff
@@ -406,4 +417,20 @@ private function isCartHoverEnable(): bool
 
         return !in_array($masterRequest->get('_route'), self::PAGES_WITH_DISABLED_CART_HOVER, true);
     }
+
+    /**
+     * @required
+     * @param \Shopsys\FrameworkBundle\Model\Module\ModuleFacade $moduleFacade
+     * @internal This function will be replaced by constructor injection in next major
```
s3tezsky, line 424:
```suggestion
     * @deprecated This function will be replaced by constructor injection in next major
```

project-base/src/Controller/Front/CartController.php
```diff
@@ -275,14 +284,16 @@ public function addProductAjaxAction(Request $request)
...
-                    $addProductResult->getCartItem()->getProduct()->getId(),
-                    self::AFTER_ADD_WINDOW_ACCESSORIES_LIMIT
-                );
+                $accessories = [];
+                if ($this->moduleFacade->isEnabled(ModuleList::ACCESSORIES_ON_BUY)) {
+                    $accessories = $this->listedProductViewFacade->getAccessories(
+                        $addProductResult->getCartItem()->getProduct()->getId(),
+                        self::AFTER_ADD_WINDOW_ACCESSORIES_LIMIT
+                    );
+                }
```
s3tezsky, line 293: It would be really lovely if accessories is provided via separated method. Do not you think?

s3tezsky (changes requested):

project-base/src/Controller/Front/CartController.php
TomasLudvik: It should be internal as it is called by DIC.

project-base/src/Controller/Front/CartController.php
TomasLudvik: Well I do not see this as important. This PR is bugfix, no need to do refactoring here

s3tezsky (approved):

project-base/src/Controller/Front/CartController.php
```diff
@@ -79,7 +86,8 @@ public function __construct(
         OrderPreviewFactory $orderPreviewFactory,
         ErrorExtractor $errorExtractor,
         ListedProductViewFacadeInterface $listedProductViewFacade,
-        RequestStack $requestStack
+        RequestStack $requestStack,
+        ?ModuleFacade $moduleFacade = null
```
sslt, line 90: I am sorry, I have urgent need to confirm something...
the question mark at the beginning of line is on keyboard very close to uppercase M.
Perhaps it is a typo?

sslt (approved):

PR #1994, opened 2020-08-19
[framework] added link into breadcrumb item in admin if uri is not null

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We have breadcrumb navigation in the administration, but if there is a link to the route, this navigation will not display the link. This PR adds a link to the navigation.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Resources/views/Admin/Inline/Breadcrumb/breadcrumb.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Inline/Breadcrumb/breadcrumb.html.twig
@@ -7,7 +7,13 @@
         {% elseif loop.last %}
             <span class="in-breadcrumb__item in-breadcrumb__item--actual">{{ breadcrumbOverrider.lastItemOverridden ? breadcrumbOverrider.lastItemLabel : item.label }}</span>
         {% else %}
+            {% if item.uri is not null %}
+                <a href="{{ item.uri }}">
+            {% endif %}
             <span class="in-breadcrumb__item">{{ item.label }}</span>
+            {% if item.uri is not null %}
+                </a>
+            {% endif %}
         {% endif %}
 
         {% if not loop.last %}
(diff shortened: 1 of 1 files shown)


packages/framework/src/Resources/views/Admin/Inline/Breadcrumb/breadcrumb.html.twig
```diff
@@ -7,7 +7,13 @@
         {% elseif loop.last %}
             <span class="in-breadcrumb__item in-breadcrumb__item--actual">{{ breadcrumbOverrider.lastItemOverridden ? breadcrumbOverrider.lastItemLabel : item.label }}</span>
         {% else %}
+            {% if item.uri is not null %}
+                <a href="{{ item.uri }}">
+            {% endif %}
             <span class="in-breadcrumb__item">{{ item.label }}</span>
+            {% if item.uri is not null %}
+                </a>
+            {% endif %}
```
s3tezsky, line 16:
```suggestion
            {% if item.uri is not null %}
                <a href="{{ item.uri }}"><span class="in-breadcrumb__item">{{ item.label }}</span></a>
            {% else %}
                <span class="in-breadcrumb__item">{{ item.label }}</span>
            {% endif %}
```

s3tezsky (changes requested):
Hi @pesektomas,

thanks for PR. I would just make it a different way. Your solution causes a whitespace to be underscored.

packages/framework/src/Resources/views/Admin/Inline/Breadcrumb/breadcrumb.html.twig
pesektomas: Hello @s3tezsky, wht not. I love latte `n:tag-if` and I'm very sad that twig doesn't have something like that.

s3tezsky (approved):

sslt (approved):

PR #2000, opened 2020-08-24
[shopsys] timezone for running crons

### LINKED ISSUE #1924: wrong calculation of visibility by domain and product visibility date
### What is happening

If I set the end of the sale date to today’s in the administration and save it, then the visibility of the product is recalculated on individual domains. But the calculation uses new Datetime() and the date from the form, since I only set the date, it returns with the time 00:00:00, and since we have a timezone set on the front end to Prague, this date is changed to UTC, ie. - 2 hours. So this date will become yesterday for me and so the product will be hidden, which is not right and should remain visible. This can also be a problem with other dates where the time is not set.

Why not set timezone for domain in domain.yml ? 
### Expected result

Calculate visibility must be by domain time zone. (https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Product/ProductVisibilityRepository.php#L183)
 ```
/**
     * @return \DateTime
     */
    public function createDateWithTimezoneCorrection(): \DateTime
    {
        $dateByTimeZoneProvider = new \DateTime('midnight', $this->getDisplayTimeZone());
        $dateByTimeZoneProvider->setTimezone(new \DateTimeZone('UTC'));

        return $dateByTimeZoneProvider;
    }



```

### PR DESCRIPTION
**Pull request by TomasLudvik**
| Q             | A
| ------------- | ---
|Description, reason for the PR| With introduction of setting all dates and times in UTC timezone, problems with running crons in desired times has occurred.
|New feature| No 
|Fixes issues| closes #1924 

### DIFF
--- a/packages/framework/src/Command/CronCommand.php
+++ b/packages/framework/src/Command/CronCommand.php
@@ -2,8 +2,10 @@
 
 namespace Shopsys\FrameworkBundle\Command;
 
+use BadMethodCallException;
 use DateTime;
 use DateTimeImmutable;
+use DateTimeZone;
 use Shopsys\FrameworkBundle\Component\Cron\Config\CronModuleConfig;
 use Shopsys\FrameworkBundle\Component\Cron\CronFacade;
 use Shopsys\FrameworkBundle\Component\Cron\MutexFactory;
@@ -12,6 +14,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 
 class CronCommand extends Command
 {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/framework/src/Command/CronCommand.php [hunk 1]
```diff
@@ -162,7 +189,10 @@ private function runCron(InputInterface $input, CronFacade $cronFacade, MutexFac
      */
     private function getCurrentRoundedTime()
     {
-        $time = new DateTime(null);
+        $cronTimezone = $this->parameterBag->get('shopsys.cron_timezone');
+        $cronTimezone = $cronTimezone ?? date_default_timezone_get();
+
+        $time = new DateTime(null, new DateTimeZone($cronTimezone));
```
s3tezsky, line 195:
```suggestion
        $time = new DateTime('now', new DateTimeZone($cronTimezone));
```

https://www.php.net/manual/en/datetime.construct.php

packages/framework/src/Command/CronCommand.php
```diff
@@ -162,7 +189,10 @@ private function runCron(InputInterface $input, CronFacade $cronFacade, MutexFac
      */
     private function getCurrentRoundedTime()
     {
-        $time = new DateTime(null);
+        $cronTimezone = $this->parameterBag->get('shopsys.cron_timezone');
+        $cronTimezone = $cronTimezone ?? date_default_timezone_get();
```
s3tezsky, line 193: I would move it to something like `getDateTimeZone(): DateTimeZone`

s3tezsky (changes requested):

packages/framework/src/Command/CronCommand.php
TomasLudvik: updated

s3tezsky (approved):

sslt (approved):

PR #2001, opened 2020-08-25
[framework] AbstractFileUploadType: remove fileConstraints class property

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When there are multiple instances of `AbstractFileUploadType` in a given form (both `FileUploadType` and `ImageUploadType`), it might happen that the constraints from the first one are used for the second type in `AbstractFileUploadType::validateUploadedFiles` method. I am not sure why this even happens, however, the problem is solved by removing the class property and using a local variable instead.
|New feature| No 
|Fixes issues| ... 


grossmannmartin (approved):
Thanks!

sslt (approved):

PR #2025, opened 2020-09-07
[shopsys] optimized monorepo docker files

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Docker-sync is needed mainly for PHP files, so project runs smoothly. Other files can be synced via internal docker volumes, as its speed does not impact project performance much. This optimizations are made in hope to decrease number of files maintained by docker-sync. .git folder added to enable optimized phing targets on Win and Mac in the future. Only monorepo files are included in order to test impact of these changes.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/docker/conf/docker-compose-mac.yml.dist
+++ b/docker/conf/docker-compose-mac.yml.dist
@@ -28,7 +28,7 @@ services:
         depends_on:
             - php-fpm
         volumes:
-            - shopsys-framework-web-sync:/var/www/html/project-base/web
+            - ./project-base/web:/var/www/html/project-base/web:cached
             - ./docker/nginx/nginx.conf:/etc/nginx/conf.d/default.conf:delegated
         ports:
             - "8000:8080"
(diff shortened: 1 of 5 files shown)


docker/conf/docker-compose-mac.yml.dist
```diff
@@ -28,7 +28,7 @@ services:
         depends_on:
             - php-fpm
         volumes:
-            - shopsys-framework-web-sync:/var/www/html/project-base/web
+            - ./project-base/web:/var/www/html/project-base/web:cached
```
s3tezsky, line 31: If synced `./project-base/web` will do fine it could be replaced in `docker-compose.yml.dist` for webserver too.

TomasLudvik: On Linux it does not matter what folder do you wanna create volume from as actually there are no volumes created on linux

s3tezsky: Yes there are.

s3tezsky (approved):

sslt (approved):

PR #2032, opened 2020-09-09
[shopsys] remove FE API only dependencies from framework

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When working on FE API we have added some methods to framework classes that should not be there. This PR fixes that.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/src/Model/Order/OrderRepository.php
+++ b/packages/frontend-api/src/Model/Order/OrderRepository.php
@@ -25,6 +25,7 @@ class OrderRepository
 
     /**
      * @return \Doctrine\ORM\QueryBuilder
+     * @internal This will be replaced by \Shopsys\FrameworkBundle\Model\Order::getOrderRepository() with visibility set to public
      */
     protected function createOrderQueryBuilder()
     {
--- a/packages/frontend-api/src/Model/Order/OrderRepository.php
+++ b/packages/frontend-api/src/Model/Order/OrderRepository.php
@@ -1,0 +1,117 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Order;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser;
+use Shopsys\FrameworkBundle\Model\Order\Order;
+
+class OrderRepository
+{
+    /**
+     * @var \Doctrine\ORM\EntityManagerInterface
+     */
+    protected $em;
+
+    /**
+     * @param \Doctrine\ORM\EntityManagerInterface $entityManager
+     */
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->em = $entityManager;
+    }
+
+    /**
+     * @return \Doctrine\ORM\QueryBuilder
+     */
+    protected function createOrderQueryBuilder()
+    {
+        return $this->em->createQueryBuilder()
+            ->select('o')
+            ->from(Order::class, 'o')
+            ->where('o.deleted = FALSE');
+    }
+
+    /**
+     * @param string $uuid
+     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser $customerUser
+     * @return \Shopsys\FrameworkBundle\Model\Order\Order|null
+     */
+    protected function findByUuidAndCustomerUser(string $uuid, CustomerUser $customerUser)
+    {
+        return $this->createOrderQueryBuilder()
+            ->andWhere('o.uuid = :uuid')->setParameter(':uuid', $uuid)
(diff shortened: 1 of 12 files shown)


packages/frontend-api/src/Model/Order/OrderRepository.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->em = $entityManager;
+    }
+
+    /**
+     * @return \Doctrine\ORM\QueryBuilder
+     */
+    protected function createOrderQueryBuilder()
```
s3tezsky, line 29: Should not it be also deprecated and reusing `OrderRepository::createOrderQueryBuilder()`?

packages/frontend-api/src/Model/Resolver/Order/OrderResolver.php
```diff
@@ -32,19 +33,27 @@ class OrderResolver implements ResolverInterface, AliasedInterface
...
+     * @param \Shopsys\FrontendApiBundle\Model\Order\OrderFacade $frontendApiOrderFacade
      */
     public function __construct(
         CurrentCustomerUser $currentCustomerUser,
         OrderFacade $orderFacade,
-        Domain $domain
+        Domain $domain,
+        FrontendApiOrderFacade $frontendApiOrderFacade
     ) {
         $this->orderFacade = $orderFacade;
```
s3tezsky, line 53: I am not sure if original `OrderFacade` should be used here.

It is a bit confusing having 2 different order facades. This one is only used on line 63 `$this->orderFacade->getByUrlHashAndDomain`. Maybe this method could be exposed into FEAPI Order Facade too?

packages/frontend-api/src/Model/Resolver/Order/OrdersResolver.php
```diff
@@ -32,15 +33,25 @@ class OrdersResolver implements ResolverInterface, AliasedInterface
...
      */
-    public function __construct(CurrentCustomerUser $currentCustomerUser, OrderFacade $orderFacade)
-    {
+    public function __construct(
+        CurrentCustomerUser $currentCustomerUser,
+        OrderFacade $orderFacade,
+        FrontendApiOrderFacade $frontendApiOrderFacade
+    ) {
         $this->currentCustomerUser = $currentCustomerUser;
         $this->orderFacade = $orderFacade;
```
s3tezsky, line 52: This property is unused so far.

s3tezsky (changes requested):

packages/frontend-api/src/Model/Resolver/Order/OrderResolver.php
TomasLudvik: I think this is absolutely ok. Those are two different facades and we do not want to duplicate code in FE API, we just wanna prevent methods from FE API to appear in framework.

packages/frontend-api/src/Model/Resolver/Order/OrdersResolver.php
TomasLudvik: Nice catch

packages/frontend-api/src/Model/Order/OrderRepository.php
TomasLudvik: Fixed

packages/frontend-api/src/Model/Resolver/Order/OrderResolver.php
s3tezsky: I do not want to duplicate the logic I was more thinking about bridging this method through FEAPI Order Facade...

It will be hard to think about what facade should I use. It is sad we cannot use the advantages of inheritance here...

grossmannmartin: I don't think it's a problem. 
We cannot use inheritance, because then it would be hard to extend the proper facade. 

I would like to not composite facades (like `$frontendApiOrderFacade->getOrderFacade()->...`).

s3tezsky: As `FrontendApiOrderFacade` is already dependent on `OrderFacade` it is useless and confusing as a delevoler to have both dependencies in `OrderResolver` - when one would do just fine (easier developmnet, testing, more coupled).

I was not suggesting making composition from the resolver. I was more thinking about exposing the method in `FrontendApiOrderFacade` like:

```php
// OrderResolver

$this-frontendApiOrderFacade->getMeAny();
```

```php
// FrontendApiOrderFacade

public function getMeAny(): Any
{
    return $this->orderFacade->getMeAny();
}
```

s3tezsky (approved):

sslt (approved):

PR #2057, opened 2020-10-01
[shopsys] added missing data to Product frontend API type

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Some data of Product has been missing in frontend API, these has been added by this PR.
|New feature| Yes
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/apiary.apib
+++ b/packages/frontend-api/apiary.apib
@@ -56,6 +56,12 @@ Authorization header is not required.
                         brand {
                             name
                         }
+                        accessories {
+                            name
+                        }
+                        isSellingDenied
+                        description
+                        orderingPriority
                     }
                 }
             }
@@ -97,7 +103,11 @@ Authorization header is not required.
                                 "vatAmount": "0.84"
                             },
                             "images": [],
-                            "brand": null
+                            "brand": null,
+                            "accessories": [],
+                            "isSellingDenied": false,
+                            "description": "Coupon valued to 100 Czech crowns. You can cash it at any exchange office",
+                            "orderingPriority": null
                         }
                     },
                     {
@@ -220,7 +230,18 @@ Authorization header is not required.
                             ],
                             "brand": {
                                 "name": "Sencor"
-                            }
+                            },
+                            "accessories": [
+                                {
+                                    "name": "Kabel HDMI A - HDMI A M/M 2m gold-plated connectors High Speed HD"
+                                },
+                                {
+                                    "name": "Defender 2.0 SPK-480"
+                                }
+                            ],
+                            "isSellingDenied": false,
+                            "description": "Television LED, 55 cm diagonal, 1920x1080 Full HD, DVB-T MPEG4 tuner with USB recording and playback (DivX, XviD, MP3, WMA, JPEG), HDMI, SCART, VGA, pink execution, energ. Class B",
+                            "orderingPriority": null
                         }
                     }
                 ]
@@ -272,6 +293,12 @@ Returns product filtered using UUID
             brand {
                 name
             }
+            accessories {
+                name
+            }
+            isSellingDenied
+            description
+            orderingPriority
         }
     }
 
@@ -397,7 +424,18 @@ Returns product filtered using UUID
                 ],
                 "brand": {
                     "name": "Sencor"
-                }
+                },
+                "accessories": [
+                    {
+                        "name": "Kabel HDMI A - HDMI A M/M 2m gold-plated connectors High Speed HD"
+                    },
+                    {
+                        "name": "Defender 2.0 SPK-480"
+                    }
+                ],
+                "isSellingDenied": false,
+                "description": "Television LED, 55 cm diagonal, 1920x1080 Full HD, DVB-T MPEG4 tuner with USB recording and playback (DivX, XviD, MP3, WMA, JPEG), HDMI, SCART, VGA, pink execution, energ. Class B",
+                "orderingPriority": null
             }
         }
     }
--- a/packages/frontend-api/apiary.apib
+++ b/packages/frontend-api/apiary.apib
@@ -62,6 +62,14 @@ Authorization header is not required.
                         isSellingDenied
                         description
                         orderingPriority
+                        parameters {
+                            uuid
+                            name
+                            values {
+                                uuid
+                                text
+                            }
+                        }
                     }
                 }
             }
@@ -107,7 +115,8 @@ Authorization header is not required.
                             "accessories": [],
                             "isSellingDenied": false,
                             "description": "Coupon valued to 100 Czech crowns. You can cash it at any exchange office",
-                            "orderingPriority": null
+                            "orderingPriority": null,
+                            "parameters": []
                         }
                     },
                     {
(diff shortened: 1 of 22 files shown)


packages/framework/src/Model/Product/Accessory/ProductAccessoryFacade.php
```diff
@@ -31,4 +33,19 @@ public function getTopOfferedAccessories(Product $product, $domainId, PricingGro
...
         return $this->productAccessoryRepository->getTopOfferedAccessories($product, $domainId, $pricingGroup, $limit);
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param int $domainId
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup $pricingGroup
+     * @return array
+     */
+    public function getAllAccessories(Product $product, int $domainId, PricingGroup $pricingGroup): array
```
s3tezsky, line 43: This method exist because of FEAPI only - should not it be in its package?

project-base/tests/FrontendApiBundle/Functional/Product/ProductTest.php
```diff
@@ -67,18 +67,33 @@ public function testProductDetailWithAllAttributesByUuid(): void
...
+                    isSellingDenied,
+                    description,
+                    orderingPriority
+                    parameters {
+                        parameter {
+                            name
+                        }
+                        value {
+                            text
+                        }
```
s3tezsky, line 96: Are you sure this structure is the one we want?

What do you think about having it like this:
```
parameters {
    name
    values {
        text
    }
}
```
Whould not it be better especially when we start using it for displaying filter?

s3tezsky (changes requested):

project-base/tests/FrontendApiBundle/Functional/Product/ProductTest.php
TomasLudvik: There can be only one parameter of same type per product, so this makes no sense

packages/framework/src/Model/Product/Accessory/ProductAccessoryFacade.php
TomasLudvik: Nice catch, just sorry, I fixed it in commit not in fixup, I hope it is not a problem

project-base/tests/FrontendApiBundle/Functional/Product/ProductTest.php
dostrozlik: I agree with @s3tezsky here. I think that the usage with values inside parameter will be a lot better on FE (especially in filter) and the structure won't be as confusing and will follow the "GraphQL way" of relations

s3tezsky: > There can be only one parameter of same type per product, so I do not see need for that

This is sad, as you may want to have more values to one parameter. I guess we should not tighten us by introducing this structure in FEAPI too.

E.g.: When you have a wifi router you will have parameter called frequency where you can have both of 2,4GHz and 5GHz and so on.

packages/framework/src/Model/Product/Accessory/ProductAccessoryFacade.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
```
s3tezsky, line 4: I really do not mind it added. But maybe it is unwanted here?

project-base/tests/FrontendApiBundle/Functional/Product/ProductTest.php
TomasLudvik: Guys, I understand, what you are saying and basically I agree. But I see there possible confusion about returning different type, than actual used in application. 

For example, we do know, that many of e-commerce solutions uses multiple warehouses and wants to return stock per warehouse, but we return only integer of stock quantity, because it is current implementation. 

So I see this as subject of discussion and would like to see at least @grossmannmartin opinion and @sslt opinion would be also appreciated.

TomasLudvik: I have updated structure as requrested

packages/framework/src/Model/Product/Accessory/ProductAccessoryFacade.php
TomasLudvik: fixed

packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $productParameterValues = $this->productCachedAttributesFacade->getProductParameterValues($product);
+
+        $valuesByParameterId = [];
+        $parameters = [];
+        $parametersWithValues = [];
+
+        foreach ($productParameterValues as $productParameterValue) {
+            $valuesByParameterId[$productParameterValue->getParameter()->getId()][] = $productParameterValue->getValue();
+            $parameters[$productParameterValue->getParameter()->getId()] = $productParameterValue->getParameter();
```
s3tezsky, line 51: Just a nitpick to make it more readable.

```suggestion
            $parameterId = $productParameterValue->getParameter()->getId();
            $valuesByParameterId[parameterId][] = $productParameterValue->getValue();
            $parameters[parameterId] = $productParameterValue->getParameter();
```

packages/frontend-api/src/Model/Resolver/Parameter/ParameterResolverMap.php
```diff
@@ -0,0 +1,31 @@
...
+                },
+                'name' => function (ParameterWithValues $parameterWithValues) {
+                    return $parameterWithValues->getParameter()->getName();
+                },
+                'visible' => function (ParameterWithValues $parameterWithValues) {
+                    return $parameterWithValues->getParameter()->isVisible();
+                },
+            ],
+        ];
+    }
```
s3tezsky, line 30: Maybe this could be automated with some serializer inside `ParameterWithValues` which will make it more easier to extend.

But I really do not mind it now. It just an idea for future.

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
```diff
@@ -84,7 +91,8 @@ public function __construct(
         ?ProductAccessoryFacade $productAccessoryFacade = null,
         ?CurrentCustomerUser $currentCustomerUser = null,
         ?ProductFacade $productFacade = null,
-        ?ProductCachedAttributesFacade $productCachedAttributesFacade = null
+        ?ProductCachedAttributesFacade $productCachedAttributesFacade = null,
```
s3tezsky, line 94: This dependency should not be necessary if I am not missing anything.

s3tezsky:
Great work!

I left last few details, please check them out and fix travis build. Then it will be perfect 👌

packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
TomasLudvik: done

packages/frontend-api/src/Model/Resolver/Parameter/ParameterResolverMap.php
TomasLudvik: done

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
TomasLudvik: fixed

packages/frontend-api/apiary.apib
```diff
@@ -272,6 +354,22 @@ Returns product filtered using UUID
...
+            orderingPriority
+            parameters {
+                parameter {
+                    uuid
+                    name
+                }
+                value {
+                    uuid
+                    text
+                }
```
s3tezsky, line 371: This needs to be updated

TomasLudvik: fixed

s3tezsky (approved):

sslt (approved):

PR #2062, opened 2020-10-06
[shopsys] fixed fraction digits of prices in FEAPI

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Frontend API does not respects fraction digits per currency.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
+++ b/packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Component\Price;
+
+use Shopsys\FrameworkBundle\Component\Money\Money;
+
+class MoneyFormatterHelper
+{
+    protected const MAX_FRACTION_DIGITS = 6;
+    protected const DECIMAL_POINT = '.';
+    protected const THOUSANDS_SEPARATOR = '';
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money $money
+     * @return string
+     */
+    public static function formatWithMaxFractionDigits(Money $money): string
+    {
+        return number_format(
+            $money->getAmount(),
+            static::MAX_FRACTION_DIGITS,
+            static::DECIMAL_POINT,
+            static::THOUSANDS_SEPARATOR
+        );
+    }
+}
--- a/project-base/tests/FrontendApiBundle/Functional/Order/GetOrderAsAuthenticatedCustomerUserTest.php
+++ b/project-base/tests/FrontendApiBundle/Functional/Order/GetOrderAsAuthenticatedCustomerUserTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Tests\FrontendApiBundle\Functional\Order;
 
 use Shopsys\FrameworkBundle\Model\Order\Order;
+use Shopsys\FrontendApiBundle\Component\Price\MoneyFormatterHelper;
 use Tests\FrontendApiBundle\Test\GraphQlWithLoginTestCase;
 
 class GetOrderAsAuthenticatedCustomerUserTest extends GraphQlWithLoginTestCase
@@ -14,6 +15,12 @@ class GetOrderAsAuthenticatedCustomerUserTest extends GraphQlWithLoginTestCase
      * @inject
      */
     private $orderFacade;
+
+    /**
+     * @var \Shopsys\FrontendApiBundle\Component\Price\MoneyFormatterHelper
+     * @inject
+     */
+    private $moneyFormatterHelper;
 
     public function testGetOrder(): void
     {
@@ -61,7 +68,9 @@ class GetOrderAsAuthenticatedCustomerUserTest extends GraphQlWithLoginTestCase
                 $order->getUuid(),
                 [
                     'status' => $order->getStatus()->getName(),
-                    'totalPriceWithVat' => $order->getTotalPriceWithVat()->getAmount(),
+                    'totalPriceWithVat' => MoneyFormatterHelper::formatWithMaxFractionDigits(
+                        $order->getTotalPriceWithVat()
+                    ),
                 ],
             ];
         }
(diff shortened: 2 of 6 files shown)


packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
```diff
@@ -0,0 +1,73 @@
...
+    protected $currencyFormatterFactory;
+
+    /**
+     * @var \CommerceGuys\Intl\Currency\CurrencyRepositoryInterface
+     */
+    protected $intlCurrencyRepository;
+
+    /**
+     * MoneyFormatterHelper constructor.
+     *
```
TomasLudvik, line 31: Remove this

packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Money\Money $money
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Component\Money\Money
+     */
+    public function getMoneyFormattedAsDecimalWithFractionDigitsByCurrencySetting(Money $money, int $domainId): Money
+    {
+        $locale = 'en';
```
TomasLudvik, line 53: You should get locale for first domain here

packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
```diff
@@ -0,0 +1,73 @@
...
+        return Money::create(
+            $currencyFormatter->format(
+                $money->getAmount(),
+                $intlCurrency->getCurrencyCode(),
+                [
+                    'currency_display' => 'none',
+                    'use_grouping' => false,
+                ]
+            )
+        );
```
TomasLudvik, line 71: I do not understand this at all... Money should not include any formatting, it is supposed to be value only. My head is spinning from thinking, if and how this could work.

TomasLudvik: If you need formatting on output, I think better place to do it is `MoneyType::serialize` method, as it responsible for outputing all money values in API.

project-base/tests/FrontendApiBundle/Functional/Order/GetOrderAsAuthenticatedCustomerUserTest.php
```diff
@@ -61,7 +67,10 @@ private function getOrderDataForCurrentlyLoggedCustomerUserProvider(): array
                 $order->getUuid(),
                 [
                     'status' => $order->getStatus()->getName(),
-                    'totalPriceWithVat' => $order->getTotalPriceWithVat()->getAmount(),
+                    'totalPriceWithVat' => $this->moneyFormatterHelper->getMoneyFormattedAsDecimalWithFractionDigitsByCurrencySetting(
```
TomasLudvik, line 70: In tests we are using `FunctionalTestCase::getPriceWithVatConvertedToDomainDefaultCurrency()` for getting right price for output. If you need same function but add formatting I would expect to create new method in same class, something like`getFormattedPriceWithVatConvertedToDomainDefaultCurrency` which could be alias of some helper, but I think that it should be here.

TomasLudvik (changes requested):
Hello,
I was confused during checking this PR. I hope, we can find some solution to be satisfied.
I have one question: US says: "If it is possible, same method used for outputing price on frontend is used if possible. If not we will consider creating one." Did you consider it? Why was it not created?

project-base/tests/FrontendApiBundle/Functional/Order/GetOrderAsAuthenticatedCustomerUserTest.php
s3tezsky: I do not see a good reason to (not) doing so.

It may comes handy when writing a new test and having this method right in `FunctionalTestCase`. On other hand I am a bit worried if it is the right place to be. In some time the `FunctionalTestCase` will became something like `ProductFacade`.

packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
s3tezsky: > I do not understand this at all... Money should not include any formatting, it is supposed to be value only. My head is spinning from thinking, if and how this could work.

This is the only code responsible for proper price formatting. I have already discussed it with @grossmannmartin about not ideal flow of doing that. Sadly it is what we have got.

In future I would like to make the formatting more readable/extandable/clean.

> If you need formatting on output, I think better place to do it is MoneyType::serialize method, as it responsible for outputing all money values in API.

This is really good point! Thanks!

packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
s3tezsky: No it should not.

That is why we need this method. We want to have formatted price with the decimal dot (not comma).

s3tezsky:
> I have one question: US says: "If it is possible, same method used for outputing price on frontend is used if possible. If not we will consider creating one." Did you consider it? Why was it not created?

That is exactly what I have done here. I found out in framework, there is always used formatting by `PriceExtension` which cannot be used here. The way how the extension work is extracted into helper (`MoneyFormatterHelper`) and used widely for all occurences.

packages/frontend-api/src/Component/Price/MoneyFormatterHelper.php
s3tezsky: Sadly it cannot be done as the MoneType uses static methods. Because of that we are not able to format it there.

TomasLudvik: In such case some documentation describing, how people should format Money type in FE API would be appreciated.

s3tezsky:
I have accidentally overwrite the PR's history, sorry about that.

Anyway from internal discussions we agreed on a diferent way how to fix it which was tightly connected with planned fixes to support FEAPI tests for any currency.

That is why I have created a new PR #2075 and closed this one.

PR #1228, opened 2019-07-19
[shopsys] Recalculation of Product visibilities via async queue

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to include async queues in the framework so we can do some stuff asynchronously. Good example are recalculations of all kinds in Product and Category, that are done at the end of application life cycle in the onKernelResponse event. This will be all moved to async queue so they do not need to wait for end of Request.
|New feature| Yes
|[BC breaks](https://github.com/shopsys/shopsys/blob/master/docs/contributing/backward-compatibility-promise.md)| Yes
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #2076, opened 2020-10-14
[shopsys] poc of rabbitMQ

pesektomas:
| Q             | A
| ------------- | ---
|New feature| Yes 
|Fixes issues| ... 

Hello,

I've been thinking for a while about how to synchronize data with other systems as quickly as possible, and instead of shortening cron time, I'd like to use an asynchronous queue.

In this PR, I tried to outline the implementation and start a debate about whether it makes sense or not. I tried to export products to elastic using RbbitMq using the BunnyPhp library implemented by RabbitMqBundle.

I came across two points that I can't deal with in this solution, and at least I would like someone to try to explain to me how this could be solved.

The first point is how to put consumers into operation correctly?

And the second point is that when I send data to the queue with the help of a producer in ProductFacade, it happens that the asynchronous consumer starts downloading the data before the doctrine stores the data in the database in the original request. I fixed it here with the x-delay header, but of course I don't like it.

I'd like to talk to someone about it and compare my thoughts. 

Thank you!


grossmannmartin:
Hi @pesektomas,

thank you for this, you rock 👍 🤘 🆒 

We experimented with rabbit earlier (see #1202, #1228 and #1321) and I would like to have this prepared in the framework

As for the points, you mentioned.

- I'm nut sure about BunnyPhp, but running a consumer should be just a matter of running  command (at least in previous prototypes it was). So you will need another container running this command as entry-point. I believe in #1321 was also somehow solved changing the source codes in running app (consumer is running indefinitely, and by default changes in source code is not propagated into running php app).
In production I see it as a new pod (autoscaled 😈 ) doing the hard work.

- Yup, problem is a master transaction wrapping each request. So you send a message, but db transaction is committed later. 
I have some ideas (remove master transaction, create delayed messenger that sends data after request), but I'm not sure which way is the right way.

pesektomas:
Hello @grossmannmartin,

yes yes, bunnyPhp is also started by the command (I prepared a phing target). I prepared my own container for it and it looks ok, I was just wondering if it's not an overkill. In addition, I was not very good at synchronizing containers, but that will be my poor knowledge of docker.

And yes, I also know about the main transaction, I solved it by delay, but I'm also not sure if it's ok. I think it would take something better, but I don't have much idea how to do it :)

And now I see that I have no idea why I built the build :(

TomasLudvik:
Hello @pesektomas, thank you for this PR. It has surged some discussion around queues again. Sadly I see here same problems (master transaction, product recalculations, ...) that stopped implementing queues in PRs that @grossmannmartin mentioned. Also I think that those PRs had a bit better solution of consumers in containerized environment.

That leads me to suggestion to replace this PR with issue(s) addressing these problems and including queues in Shopsys Framework.

I hope you understand.

sslt:
Hello @pesektomas, thank you PR, we have discussed already and right now we decided to close this PR. Topic will be addressed next year.

PR #2080, opened 2020-10-17
[framework] optimization: flush the whole identity map when an order is created

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| For a long time, I thought that for improving performance, you should always flush just the relevenat entities, not the whole identity map. However, during optimizations on my project, I found out, that the previous is probably not 100 % true. When you have the new entities in the identity map only (e.g. when creating a new order), then flush without parameters is much faster. This single change really helped me on the project - it **decreased the flush time from 2.75s to 955ms**! It would be great if you had time for exploring this behavior further and maybe change other pieces of code that call flush with an array of new entities :thinking: 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Order/OrderFacade.php
+++ b/packages/framework/src/Model/Order/OrderFacade.php
@@ -263,7 +263,6 @@ class OrderFacade
     {
         $orderNumber = (string)$this->orderNumberSequenceRepository->getNextNumber();
         $orderUrlHash = $this->orderHashGeneratorRepository->getUniqueHash();
-        $toFlush = [];
 
         $this->setOrderDataAdministrator($orderData);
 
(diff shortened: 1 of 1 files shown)


TomasLudvik:
Hello @vitek-rostislav, thank you for bringing this to our attention. This is pretty interesting and I think we should remove all flushes with parameters and run performance tests to find if it has any influence there.

vitek-rostislav:
Hi @TomasLudvik, well I suppose it will not be as easy - in many cases, it really helped me to call flush with a parameter so removing them all will probably worsen the performance, call without flush seems to help in the described scenario only :shrug: 

But if we want to upgrade to Doctrine3 eventually, there will be no other option, so it might be a good idea to explore that now :slightly_smiling_face: :+1:

s3tezsky (approved):
Thanks for PR!

Sadly I have no clue why it is better and also I had not measured such a performance gain, but it is still better (~400ms), good job!

sslt (approved):

PR #2083, opened 2020-10-19
[framework] fixed checking of empty settings values

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #2035 empty function was removed, but checking for settings values was not properly changed. Based on this value, some fields (ga tracking code or extra checkbox field in order) are rendered. Those values are considered empty when null. This PR fixes usage. 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Heureka/HeurekaSetting.php
+++ b/packages/framework/src/Model/Heureka/HeurekaSetting.php
@@ -64,7 +64,7 @@ class HeurekaSetting
      */
     public function isHeurekaShopCertificationActivated($domainId)
     {
-        return $this->setting->getForDomain(static::HEUREKA_API_KEY, $domainId) !== '';
+        return $this->setting->getForDomain(static::HEUREKA_API_KEY, $domainId) !== null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


pk16011990:
Isn't the problem that the value should not be an empty string? I see only one place where it is edited and it is form. Form fills null into empty values. (There is wrong annotation/parameter's type)

Eg. only null is tested for testing empty product name.

s3tezsky (approved):
>Isn't the problem that the value should not be an empty string? 

It is, we should better handle the form input. Maybe the explicit checkbox for enable/disable would be fine.
Checks for empty strings are necessary because on some project it may be already saved as empty string in database.

I am ok with this PR but maybe we could consider to not returning empty strings from `getForDomain` and return `null` instead.
Or rather sanitize those empty stings when tey are saving into db. AFAIK We know the desired type of value and we are not allowing to change it.
What do you think?

pk16011990:
@s3tezsky, yes, now I understand and agree with you. May it should be deprecated when it is called with empty string + next major version of framework removes the deprecation...?

grossmannmartin:
Thanks for the points 👍 

I checked the underlying code and I think it's not possible to create the empty string value with type `StringValue::TYPE_NULL` (see `SettingValue::setValue` and `SettingValue::getValueType`).

Converting an empty string to null on saving seems like a good idea.
I'm afraid of inevitable BC break – changed behavior may easily break the current code:
- an empty string is now saved as string type and this "may" be a valid value
- implicit convert to null on save/load can break a code
So I would like to not introduce this change in this PR. 

I changed this PR, so it's now checked for a null value to fix an imminent problem (unnecessary check for the empty string was removed)

s3tezsky (approved):
nice! reviewed and tested

sslt (approved):

PR #2084, opened 2020-10-19
[framework] do not run checks for cart modifications on cart delete

### LINKED ISSUE #2082: Unnecessary checkCartModifications after order created when deleting cart
When creating order via createOrderFromFront, order is created, there are products subtracted from stock and then cart of current customer should be deleted. In this deleting function, there is find for the cart that checks for modifications onece more.
If you extend checkCartModifications with for example checkModifiedQuantities, you will get error if the product is no longer in stock.

### What is happening

1. the sequence is
```
$order = $this->createOrder($orderData, $orderPreview, $customerUser);
$this->orderProductFacade->subtractOrderProductsFromStock($order->getProductItems());
$this->cartFacade->deleteCartOfCurrentCustomerUser();
```
- that means order is created (and checked for modifications multiple times before)
- products are subtracted from stock
- delete cart of current customer used calls `findCartOfCurrentCustomerUser` -> `findCartByCustomerUserIdentifier`
->  `checkCartModifications`
2. we have function `checkModifiedQuantities` in `checkCartModifications` function
3. order is valid with valid items, but we also get error message due checkCartModifications beying called while deleting cart:  
image

### Expected result

`findCartByCustomerUserIdentifier` does the magic of checking for cart modifications and if the cart is empty then deleting cart. That would be ok. I don't think that we should call this find function in `$this->cartFacade->deleteCartOfCurrentCustomerUser();` because we have already created order, subtracted order products from stock, we should just throw the cart away as it is.

malyMiso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Cart is deleted after creating of an order. On cart delete, checks for cart modifications are run. At this point some checks can fail, because ordered products are sellout already. Customer then see warning flash message on sent page. More info can be found in reported issue #2082
|New feature| No 
|Fixes issues| closes #2082  

### DIFF
--- a/packages/framework/src/Model/Cart/CartFacade.php
+++ b/packages/framework/src/Model/Cart/CartFacade.php
@@ -195,7 +195,9 @@ class CartFacade
 
     public function deleteCartOfCurrentCustomerUser()
     {
-        $cart = $this->findCartOfCurrentCustomerUser();
+        $customerUserIdentifier = $this->customerUserIdentifierFactory->get();
+
+        $cart = $this->cartRepository->findByCustomerUserIdentifier($customerUserIdentifier);
 
         if ($cart !== null) {
             $this->deleteCart($cart);


malyMiso:
I was asked to provide steps to reproduce a scenario, which can be tested. I was not able to reproduce such a scenario in SSFW itself. 
Expected steps to reproduce it:
1. as admin on product detail page of product `Hello Kitty` set option "use stocks" to true and "available in stock" set to 100.
2. save and switch to FE.
3. add 101 units of product `Hello Kitty` to cart.
4. proceed order process until you reach 3rd step "3. Delivery data".
5. finish the order.
6. on sent page you see that order has been created OK, without any flash message.

If everything worked correctly as it is implemented (see [createOrderFromFront](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Mode
place obstacles in someone's wayl/Order/OrderFacade.php#L300)) you would see error flash message "Product Hello Kitty you had in cart is no longer available. Please check your order." 

But it doesn't work this way, because of some magic mentioned in #2108

s3tezsky (approved):

s3tezsky:
Hi @malyMiso thanks a lot for your PR!

grossmannmartin (approved):
Looks good, thank you! Rebased to current master.

sslt (approved):

PR #2086, opened 2020-10-20
[shopsys] allow placing scripts in administration after content

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Scripts (various javascript) created in administration can be now placed after content to avoid negative impact on page performance. Rendering scripts was also changer from rendering controller to calling function to improve page performance.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Script/ScriptFacade.php
+++ b/packages/framework/src/Model/Script/ScriptFacade.php
@@ -127,17 +127,19 @@ class ScriptFacade
 
     /**
      * @return string[]
+     * @deprecated use getAllPagesBeforeContentScriptCodes() instead
      */
     public function getAllPagesScriptCodes()
     {
-        $allPagesScripts = $this->scriptRepository->getScriptsByPlacement(Script::PLACEMENT_ALL_PAGES);
-        $scriptCodes = [];
+        @trigger_error(
+            sprintf(
+                'The "%s()" method is deprecated and will be removed in the next major. Use "getAllPagesBeforeContentScriptCodes()" instead.',
+                __METHOD__
+            ),
+            E_USER_DEPRECATED
+        );
 
-        foreach ($allPagesScripts as $script) {
-            $scriptCodes[] = $script->getCode();
-        }
-
-        return $scriptCodes;
+        return $this->getAllPagesBeforeContentScriptCodes();
     }
 
     /**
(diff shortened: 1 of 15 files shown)


packages/framework/src/Model/Script/ScriptFacade.php
```diff
@@ -146,14 +148,14 @@ public function getAllPagesScriptCodes()
...
-        $scriptCodes = [];
-
-        foreach ($scripts as $script) {
-            $scriptCodes[] = $this->replaceVariables($script->getCode(), $order);
-        }
-
-        return $scriptCodes;
+        $scriptCodes = $this->getScriptCodesByPlacement(Script::PLACEMENT_ORDER_SENT_PAGE);
+
+        return array_map(
```
s3tezsky, line 153: I am not realy fan of array_map function. Could we rather keep foreach here?

I know it is not a big deal, but foreach is more obvious to me (and probably to others too). Array_map is harder to read and do not have to much inpact on microoptimization.

E.g. it could be also used here but it is not https://github.com/shopsys/shopsys/pull/2086/files#diff-417f6bc61f2a1d726925f3c1e70085bfd50f3ff6a292c1c8a947be31bbecf201R228.

packages/framework/src/Twig/ScriptsExtension.php
```diff
@@ -0,0 +1,64 @@
...
+        return implode("\n", $this->scriptFacade->getAllPagesAfterContentScriptCodes());
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @return string
+     */
+    public function getOrderSentPageScripts(Order $order): string
+    {
+        return implode("\n", $this->scriptFacade->getOrderSentPageScriptCodesWithReplacedVariables($order));
```
s3tezsky, line 62: Maybe the imploding by `\n` deserves its own method. What do oyu think?

packages/framework/src/Model/Script/ScriptFacade.php
grossmannmartin: okay, changed in b6f5a32bc

_btw. I see a difference between those two methods. In first the input is an array of strings and I need to pass every value through a function, but the result is just still an array of (modified) strings. 
In the second I have an array of objects, and I need just one property from the objects._

packages/framework/src/Twig/ScriptsExtension.php
grossmannmartin: it's possible e42efbda10adff107c373998383c310f84b417f4

s3tezsky (approved):

sslt (approved):

s3tezsky (approved):

PR #2096, opened 2020-10-30
fix cleaning of old redis cache

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| On our project, after upgrading to `9.0.2`, old redis cache is not removed after calling `shopsys:redis:clean-cache-old`. Removing the build version from the prefix fixes the problem, but TBH, I am not sure whether this is ok and whether this modification does not break anything else
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/project-base/config/packages/snc_redis.yaml
+++ b/project-base/config/packages/snc_redis.yaml
@@ -24,6 +24,7 @@ snc_redis:
             dsn: 'redis://%redis_host%'
             options:
                 prefix: '%env(REDIS_PREFIX)%%build-version%:cache:framework:annotations:'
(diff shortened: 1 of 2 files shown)


vitek-rostislav:
btw, how come `RedisVersionsFacadeTest` did not reveal this problem? :thinking:

grossmannmartin:
> btw, how come `RedisVersionsFacadeTest` did not reveal this problem? 🤔

This test does not work with the actual configuration of clients (from `snc_redis.yaml`) but imitates the behavior in the test itself (a version is not obtained from the configuration, where it was changed).

s3tezsky (approved):

sslt (approved):

PR #2100, opened 2020-11-05
[framework] added support for customer user registration to Frontend API

malyMiso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| added support for customer user registration to Frontend API
|New feature| Yes
|[BC breaks]| No 
|Fixes issues| ...

### DIFF
--- a/packages/frontend-api/apiary.apib
+++ b/packages/frontend-api/apiary.apib
@@ -2020,6 +2020,38 @@ Returns customer user new personal data
         }
     }
 
+## Register a new customer user [/graphql{?register}]
+
+### POST [POST]
+
+Register and login a new customer user and return access and refresh tokens
+
+- Request (application/json)
+
+    - Body
+        mutation {
+            Register(input: {
+                firstName: "John"
+                lastName: "Doe"
+                email: "new-no-reply@shopsys.com"
+                password: "user123"
+            }) {
+                accessToken,
+                refreshToken
+            }
+        }
+
+- Response 200 (application/json; charset=utf-8)
+
+    {
+        "data": {
+            "Register": {
+                "accessToken": "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJpc3MiOiJodHRwOlwvXC9hcHBsaWNhdGlvbjo4MDAwIiwiYXVkIjoiaHR0cDpcL1wvYXBwbGljYXRpb246ODAwMCIsImlhdCI6MTYwNDU5MTkzNywiZXhwIjoxNjA0NTkyMjM3LCJkZXZpY2VJZCI6IjQwZGNjNmE0LTllZDAtNGM1YS04ZDc4LWMxOGMwMTllYWY3MCIsInV1aWQiOiJlZWZmYTM0OC1mYTg0LTQyZmQtOTAwMy1kYTkzOWZlODllYmIiLCJlbWFpbCI6Im5ldy1uby1yZXBseUBzaG9wc3lzLmNvbSIsImZ1bGxOYW1lIjoiRG9lIEpvaG4iLCJyb2xlcyI6WyJST0xFX0xPR0dFRF9DVVNUT01FUiJdfQ.MWNJIiI2C_DNn4D3-X99TYZZpOK6_QrylOnihMy7IB0l_B3BSSFJ-A34PJjdYtdp1D1PN8IUrFeykOhS3EWB_Zmdp7oMwEk5GK13-K-C0qva1H3R8qeBw-vMUp-LLv4jXpb8yOxQGojY2CJtPslNzFDTEqsyWY1qvnEDzTkzUHpJKAXeUDSvcBOcirZA0fUATeDSSzgiOX_RFYyfkcmgr39kQvVfMTP67oTSs6Re0X0cXURvh67JZwh719YMZ_plx88LB77s3liF8fqc4uIWNwL6LW8lVsN-k3U6Pd1I4QhYA7-9FbRxS47ashFSglhyBer0b5f8AYslZdZyiJ8wzw",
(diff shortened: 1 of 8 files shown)


packages/frontend-api/apiary.apib
```diff
@@ -2020,6 +2020,40 @@ Returns customer user new personal data
...
+- Request (application/json)
+
+    - Body
+        mutation {
+            Register(input: {
+                firstName: "John"
+                lastName: "Doe"
+      	        email: "new-no-reply@shopsys.com"
+      	        password: "user123"
+      	        privacyPolicy: true
```
grossmannmartin, line 2038: I'm not sure whether it is necessary to send privacyPolicy. 
In the end, it does not guarantee that the user will agree to the terms (it's based on the implementation of the client app)

project-base/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * @param string|null $email
+     * @param string|null $firstName
+     * @param string|null $lastName
+     * @param string|null $password
+     * @param string|null $privacyPolicy
+     * @return string
+     */
+    private static function getRegisterQuery(
```
grossmannmartin, line 122: is static intended? if so, the method should be called statically

project-base/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
```diff
@@ -0,0 +1,143 @@
...
+        ?string $email = self::DEFAULT_USER_EMAIL,
+        ?string $firstName = self::DEFAULT_USER_FIRST_NAME,
+        ?string $lastName = self::DEFAULT_USER_LAST_NAME,
+        ?string $password = self::DEFAULT_USER_PASSWORD,
+        ?string $privacyPolicy = self::DEFAULT_PRIVACY_POLICY
```
grossmannmartin, line 127: Are you passing null somewhere? I'm not sure if null will work

packages/frontend-api/src/Model/Mutation/Customer/User/CustomerUserMutation.php
```diff
@@ -68,7 +83,9 @@ public function __construct(
         CustomerUserRefreshTokenChainFacade $customerUserRefreshTokenChainFacade,
         TokenStorageInterface $tokenStorage,
         CustomerUserUpdateDataFactory $customerUserUpdateDataFactory,
-        CustomerUserFacade $customerUserFacade
+        CustomerUserFacade $customerUserFacade,
+        CustomerUserDataFactory $customerUserDataFactory,
+        TokenFacade $tokenFacade
```
grossmannmartin, line 88: Not considered a BC break because this class is not a part of a released version, good job 👍

grossmannmartin (changes requested):
Thanks! Nice work. 
Can you check my comments, please?

packages/frontend-api/apiary.apib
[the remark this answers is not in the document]
malyMiso: fixed

packages/frontend-api/src/Resources/config/graphql-types/RegistrationDecorator.types.yaml
```diff
@@ -0,0 +1,55 @@
+RegistrationDecorator:
```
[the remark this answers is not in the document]
malyMiso: I agree, fixed

packages/frontend-api/apiary.apib
malyMiso: I am not sure either. That's the reason of a standalone commit. I followed constraints implemented in registration form. 
User has to agree with privacy policy, when he registers within the registration form in SSFW. This is checked on the server side.

So it will be different in FE API than, if I remove this check. Is it ok ?

project-base/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
malyMiso: Nope, it's not. Removed

project-base/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
malyMiso: No I'm not, nullable option removed

packages/frontend-api/apiary.apib
malyMiso: Commit has been removed.

grossmannmartin (dismissed):
Thanks!

grossmannmartin:
I tested this PR and found the following

- fields are not trimmed (I think it's not probably related only with this PR, but it's a broader issue)
So, trying to register with email `"    dd@dd.dd"` (without quotes) returns "Internal server Error". A new user is created in the database, but I'm not able to login as this user on the frontend (in API the login method works even with spaces). 

Similarly, name and surname are trimmed when I perform registration on frontend, but not in FE API (and I don't know if it's a problem or not).

Can you check it out and create an issue for resolution if it's outside the scope of this PR?


Other than that I did not found any problems.

malyMiso:
> I tested this PR and found the following
> 
> * fields are not trimmed (I think it's not probably related only with this PR, but it's a broader issue)
>   So, trying to register with email `"    dd@dd.dd"` (without quotes) returns "Internal server Error". A new user is created in the database, but I'm not able to login as this user on the frontend (in API the login method works even with spaces).
> 
> Similarly, name and surname are trimmed when I perform registration on frontend, but not in FE API (and I don't know if it's a problem or not).
> 
> Can you check it out and create an issue for resolution if it's outside the scope of this PR?
> 
> Other than that I did not found any problems.

I agree, I've implemented `trim` function on customer user data creation. It solves only this PR, for other implemented features I'll create issue.

grossmannmartin (approved):
Nice one! Thanks.

sslt (dismissed):

sslt (approved):

grossmannmartin (approved):

PR #2107, opened 2020-11-07
[shopsys] move doctrine metadata and annotations cache to file instead of redis

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Every HTTP request do a lot of requests to Redis cache. This dramatically increases Redis CPU usage and may kill Redis. With OPcache configured for production, performance should be similar
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/project-base/config/packages/snc_redis.yaml
+++ b/project-base/config/packages/snc_redis.yaml
@@ -6,24 +6,12 @@ snc_redis:
             dsn: 'redis://%redis_host%'
             options:
                 prefix: '%env(REDIS_PREFIX)%%build-version%:cache:bestselling_products:'
-        doctrine_metadata:
-            type: 'phpredis'
-            alias: 'doctrine_metadata'
-            dsn: 'redis://%redis_host%'
-            options:
-                prefix: '%env(REDIS_PREFIX)%%build-version%:cache:doctrine:metadata:'
         doctrine_query:
             type: 'phpredis'
             alias: 'doctrine_query'
             dsn: 'redis://%redis_host%'
             options:
                 prefix: '%env(REDIS_PREFIX)%%build-version%:cache:doctrine:query:'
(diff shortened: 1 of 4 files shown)


project-base/config/packages/snc_redis.yaml
```diff
@@ -6,24 +6,12 @@ snc_redis:
...
-            alias: 'doctrine_metadata'
-            dsn: 'redis://%redis_host%'
-            options:
-                prefix: '%env(REDIS_PREFIX)%%build-version%:cache:doctrine:metadata:'
         doctrine_query:
             type: 'phpredis'
             alias: 'doctrine_query'
             dsn: 'redis://%redis_host%'
             options:
                 prefix: '%env(REDIS_PREFIX)%%build-version%:cache:doctrine:query:'
```
TomasLudvik, line 14: Doctrine query is not problematic? Why did it stay here?

TomasLudvik (changes requested):

project-base/config/packages/snc_redis.yaml
henzigo: The most problematic is `framework annotations` because lot of entries into cache on page load. I also moved `metadata cache` because it's generated once and they are not changing in running application. Both 

Doctrine queries are dynamically generated, so I prefer to have them in Redis cache because write to Redis is faster. 

Look at screen. There is count of queries grouped by cache type on stabil Homepage
Snímek obrazovky 2020-11-18 v 12 41 22

TomasLudvik: Ok, thank you for explanation

grossmannmartin (approved):
Really cool, thanks!

TomasLudvik (approved):

sslt (approved):

PR #2109, opened 2020-11-09
[shopsys] dropped support for PHP versions lower  than 7.4.1

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| From now on, only PHP 7.4.1 or higher is supported by Shopsys Framework. Explanation follows.
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| May 
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| No

In general, it's not a good idea to run an application on an unsupported version of the PHP.
PHP 7.2 is no longer actively supported and security support stops in a few days.
PHP 7.3 will stop being actively supported in nearly a month (even though security support will remain for another year).
(source https://www.php.net/supported-versions.php)

Many used libraries release new versions with PHP 7.4+ compatibility only, which forces us to keep older versions without important fixes or new features available. We count on the projects to use the newer version and drop support for lower versions by themselves, but it's not something we're happy about.
Because of all that makes maintaining the Shopsys Framework and mostly projects based on SSFW unnecessarily harder and without any added value (on the contrary in fact). 

Since v9.0.0 PHP 7.4 is a default version used if you use provided [Dockerfile](https://github.com/shopsys/shopsys/blob/v9.0.0/project-base/docker/php-fpm/Dockerfile), so your code should be more than ready.

### DIFF
--- a/docs/introduction/faq-and-common-issues.md
+++ b/docs/introduction/faq-and-common-issues.md
@@ -26,7 +26,7 @@ For more detailed information about the Shopsys Framework, please see [Shopsys F
 - [How can I create Front-end Breadcrumb navigation?](#how-can-i-create-front-end-breadcrumb-navigation)
 - [Do you have any tips how to debug emails during development in Docker?](#do-you-have-any-tips-how-to-debug-emails-during-development-in-docker)
 - [Can I see what is really happening in the Codeception acceptance tests when using Docker?](#can-i-see-what-is-really-happening-in-the-codeception-acceptance-tests-when-using-docker)
-- [Why is there a faked PHP 7.2 platform in the Composer config?](#why-is-there-a-faked-php-72-platform-in-the-composer-config)
+- [Why is there a faked PHP 7.4.1 platform in the Composer config?](#why-is-there-a-faked-php-741-platform-in-the-composer-config)
 - [How to make PHPStorm and PHPStan understand that I use extended classes?](#how-to-make-phpstorm-and-phpstan-understand-that-i-use-extended-classes)
 - [SMTP container cannot send email with error "Helo command rejected: need fully-qualified hostname"](#smtp-container-cannot-send-email-with-error-helo-command-rejected-need-fully-qualified-hostname)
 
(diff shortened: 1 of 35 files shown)


project-base/composer.json
```diff
@@ -19,7 +19,7 @@
         }
     },
     "require": {
-        "php": "^7.2",
+        "php": "^7.4.1",
```
s3tezsky, line 22: What is the reason of skipped 7.4.0? Am I missing any bug which was introduced?

project-base/composer.json
```diff
@@ -148,7 +148,7 @@
         "component-dir": "web/components",
         "sort-packages": true,
         "platform": {
-            "php": "7.2"
+            "php": "7.4.1"
```
s3tezsky, line 151: Do we still need platform to  be defined?

docs/introduction/start-building-your-application.md
```diff
@@ -138,28 +138,28 @@ If you're interested in why we use the forced PHP version in the first place, re
...
 ```
 After running `docker-compose up -d --build` you'll have the application running on the same PHP.
 
 Now you can modify the version in your `composer.json` as well so all packages will always be installed in a compatible version.
 ```diff
          "platform": {
--            "php": "7.2"
-+            "php": "7.2.19"
+-            "php": "7.4.1"
++            "php": "7.4.12"
```
s3tezsky, line 162: We are advicing use as platform version 7.4.12 but in our project-base is specified 7.4.1 - is there any reason for such a difference?

docs/introduction/faq-and-common-issues.md
```diff
@@ -152,9 +152,9 @@ See [Outgoing emails](https://github.com/djfarrelly/MailDev#outgoing-email) in t
 ## Can I see what is really happening in the Codeception acceptance tests when using Docker?
 Yes, you can! Check [the quick guide](./running-acceptance-tests.md#how-to-watch-what-is-going-on-in-the-selenium-browser).
 
-## Why is there a faked PHP 7.2 platform in the Composer config?
-As a general rule, packages and libraries that depend on PHP 7.2 will work as expected even on PHP 7.4 (any higher 7.x version), but not vice versa.
-Mainteiners of PHP are focusing on backward-compatibility (even if there were [some incompatible changes](https://www.php.net/manual/en/migration73.incompatible.php) introduced in PHP 7.3, in practice it doesn't cause issues).
+## Why is there a faked PHP 7.4.1 platform in the Composer config?
```
s3tezsky, line 155: Not sure if it is right or there should be `7.4.12` instead.

s3tezsky (changes requested):
Nice! Could you please check my comments?

project-base/composer.json
grossmannmartin: it's mainly because of roave/better-reflection package, so we can update this easy https://github.com/Roave/BetterReflection/blob/4.11.0/composer.json

project-base/composer.json
grossmannmartin: It unifies the version of the packages used for developers (freshly built image can have a different version than the older php-fpm image, or natively installed php).
I agree it's now much less needed as we use only one minor version, but I still want to keep it. It's not doing any harm and it's consistent with the documentation.

docs/introduction/start-building-your-application.md
grossmannmartin: yes, this paragraph shows the possibility to adjust (and lock) the versions with production installation. 
In general – we assume on the server is PHP 7.4.12, and to be consistent on local machines, we recommend setting the platform to 7.4.12 too.

docs/introduction/faq-and-common-issues.md
grossmannmartin: it's correct as we fake 7.4.1 in `composer.json`.

docs/introduction/start-building-your-application.md
s3tezsky: Ah, now I see. I got confused by all plus/minus signs!

docs/introduction/faq-and-common-issues.md
s3tezsky: As above, now it make sense to me!

s3tezsky (approved):
LGTM!

s3tezsky (changes requested):
Found issue while testing:
https://github.com/shopsys/shopsys/pull/2109/files?file-filters%5B%5D=.lock#diff-f2b42e29c9482076c57a611da9aafb5aadd308309e455c184262d56bf9116168R441
in symfony.lock there is still php in version 7.2 (the same is in `project-base/symfony.lock`)

grossmannmartin:
> in symfony.lock there is still php in version 7.2 (the same is in `project-base/symfony.lock`)

Thanks! I missed that. Fixed in b6029208974134db28bd99c328d47c75d46a0e51

s3tezsky (dismissed):

sslt (dismissed):

sslt (approved):

s3tezsky (approved):

PR #2122, opened 2020-11-14
requests are transactional only in admin

henzigo:
| Q             | A
| ------------- | ---
|Description, reason for the PR| On large projects, we need to use database replication. Currently, we need to remove Transactional requests for Frontend, because transactions are automatically queried to "Master" instance. For more information, you can read [this](https://github.com/doctrine/dbal/blob/2.12.x/lib/Doctrine/DBAL/Connections/PrimaryReadReplicaConnection.php#L19)
|New feature| Yes 
|Fixes issues| ... 


sslt:
Thanks for PR @henzigo , this is something we are thinking about and we want to address it thoroughly, As we believe this is major change and could affect  existing projects a lot, we will address it in next major release - v10.0.0.

pk16011990:
API requests with Query (not mutations) are no longer in transaction. I guess it is solved.

PR #2124, opened 2020-11-16
OrderPriceCalculation incorrect rounding for czech crowns

CarabineSK:
| Q             | A
| ------------- | ---
|Description, reason for the PR|`OrderPriceCalculation::calculateOrderRoundingPrice` method behaves incorrectly when you add `roundPriceWithVatByCurrency` rounding `none`. 
|New feature| No 
|Fixes issues| ... 

What happens is that we have added `none` rounding to `Rounding::roundPriceWithVatByCurrency` and this part of code behaves incorrectly. For example when I have price 100,47, `$roundedPriceWithVat` is 0,47 and then `$roundingPrice` type `none` is not taken into consideration.
```
$priceWithVat = $orderTotalPrice->getPriceWithVat();
$roundedPriceWithVat = $priceWithVat->round(0);

$roundingPrice = $this->rounding->roundPriceWithVatByCurrency($roundedPriceWithVat->subtract($priceWithVat), $currency);
```
 After I change how the `$roundedPriceWithVat` is calculated in the first place, this resolves the rounding strangeness.

Also, I have noticed that in method `calculateOrderRoundingPrice` is very strange if statement that behaves differently for some payments in CZK currency.
```
if (!$payment->isCzkRounding() || $currency->getCode() !== Currency::CODE_CZK) {
     return null;
}
```

### DIFF
--- a/packages/framework/src/Model/Order/OrderPriceCalculation.php
+++ b/packages/framework/src/Model/Order/OrderPriceCalculation.php
@@ -75,7 +75,7 @@ class OrderPriceCalculation
         }
 
         $priceWithVat = $orderTotalPrice->getPriceWithVat();
-        $roundedPriceWithVat = $priceWithVat->round(0);
+        $roundedPriceWithVat = $this->rounding->roundPriceWithVatByCurrency($priceWithVat, $currency);
 
         $roundingPrice = $this->rounding->roundPriceWithVatByCurrency(
             $roundedPriceWithVat->subtract($priceWithVat),


CarabineSK:
Used rounding type: `none`

Before the change:
image
After the change:
image

s3tezsky:
> Also, I have noticed that in method calculateOrderRoundingPrice is very strange if statement that behaves differently for some payments in CZK currency.

I have totally missed this information at first. Actually it is the whole point of this rounding.
When you pay with cache in czech crowns you cannot pay by smaller coins then 1 CZK.

PR #2131, opened 2020-11-23
[shopsys] single product is now loaded from Elasticsearch in FE API

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Load of single product from Elasticsearch provides better performance (less DB queries, better speed, better scalability) and data are consistent with the list. Some optimization introduced as part of this PR.
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes (see upgrade notes) 
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
+++ b/packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
@@ -63,4 +63,36 @@ class ParameterWithValuesFactory
 
         return $parametersWithValues;
     }
+
+    /**
+     * @param array $productData
+     * @return array
+     */
+    public function createParametersArrayFromProductArray(array $productData): array
+    {
+        $parametersWithValues = [];
+
+        foreach ($productData['parameters'] as $parameterArray) {
+            $parametersWithValues[] = $this->mapParameterArray($parameterArray);
+        }
+
+        return $parametersWithValues;
+    }
+
+    /**
+     * @param array $product
+     * @return array
+     */
+    protected function mapParameterArray(array $product): array
+    {
+        return [
+            'uuid' => $product['parameter_uuid'],
+            'name' => $product['parameter_name'],
+            'visible' => true,
+            'values' => [[
+                'uuid' => $product['parameter_value_uuid'],
+                'text' => $product['parameter_value_text'],
+            ]],
+        ];
+    }
 }
--- a/packages/frontend-api/src/Model/Resolver/Products/DataMapper/ProductArrayFieldMapper.php
+++ b/packages/frontend-api/src/Model/Resolver/Products/DataMapper/ProductArrayFieldMapper.php
@@ -1,0 +1,236 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Resolver\Products\DataMapper;
+
+use Shopsys\FrameworkBundle\Model\Category\CategoryFacade;
+use Shopsys\FrameworkBundle\Model\Product\Brand\Brand;
+use Shopsys\FrameworkBundle\Model\Product\Brand\BrandFacade;
+use Shopsys\FrameworkBundle\Model\Product\Flag\FlagFacade;
+use Shopsys\FrameworkBundle\Model\Product\ProductElasticsearchProvider;
+use Shopsys\FrontendApiBundle\Model\Parameter\ParameterWithValuesFactory;
+
+class ProductArrayFieldMapper
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Category\CategoryFacade
+     */
+    protected CategoryFacade $categoryFacade;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Product\Flag\FlagFacade
+     */
+    protected FlagFacade $flagFacade;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Product\Brand\BrandFacade
+     */
+    protected BrandFacade $brandFacade;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Product\ProductElasticsearchProvider
+     */
+    protected ProductElasticsearchProvider $productElasticsearchProvider;
+
+    /**
+     * @var \Shopsys\FrontendApiBundle\Model\Parameter\ParameterWithValuesFactory
+     */
+    protected ParameterWithValuesFactory $parameterWithValuesFactory;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Category\CategoryFacade $categoryFacade
+     * @param \Shopsys\FrameworkBundle\Model\Product\Flag\FlagFacade $flagFacade
+     * @param \Shopsys\FrameworkBundle\Model\Product\Brand\BrandFacade $brandFacade
+     * @param \Shopsys\FrameworkBundle\Model\Product\ProductElasticsearchProvider $productElasticsearchProvider
+     * @param \Shopsys\FrontendApiBundle\Model\Parameter\ParameterWithValuesFactory $parameterWithValuesFactory
+     */
+    public function __construct(
+        CategoryFacade $categoryFacade,
+        FlagFacade $flagFacade,
+        BrandFacade $brandFacade,
+        ProductElasticsearchProvider $productElasticsearchProvider,
+        ParameterWithValuesFactory $parameterWithValuesFactory
+    ) {
+        $this->categoryFacade = $categoryFacade;
+        $this->flagFacade = $flagFacade;
+        $this->brandFacade = $brandFacade;
+        $this->productElasticsearchProvider = $productElasticsearchProvider;
+        $this->parameterWithValuesFactory = $parameterWithValuesFactory;
+    }
+
+    /**
+     * @param array $data
+     * @return string|null
+     */
+    public function getShortDescription(array $data): ?string
+    {
+        return $data['short_description'];
+    }
+
+    /**
+     * @param array $data
+     * @return string
+     */
+    public function getLink(array $data): string
+    {
+        return $data['detail_url'];
+    }
+
+    /**
+     * @param array $data
+     * @return \Shopsys\FrameworkBundle\Model\Category\Category[]
+     */
+    public function getCategories(array $data): array
+    {
+        return $this->categoryFacade->getByIds($data['categories']);
+    }
+
+    /**
+     * @param array $data
+     * @return \Shopsys\FrameworkBundle\Model\Product\Flag\Flag[]
+     */
+    public function getFlags(array $data): array
+    {
+        return $this->flagFacade->getByIds($data['flags']);
+    }
+
+    /**
+     * @param array $data
+     * @return string[]
+     */
+    public function getAvailability(array $data): array
+    {
+        return ['name' => $data['availability']];
+    }
+
+    /**
+     * @param array $data
+     * @return string[]
+     */
+    public function getUnit(array $data): array
+    {
+        return ['name' => $data['unit']];
+    }
+
+    /**
+     * @param array $data
+     * @return int|null
+     */
+    public function getStockQuantity(array $data): ?int
+    {
+        return $data['stock_quantity'];
+    }
+
+    /**
+     * @param array $data
+     * @return bool
+     */
+    public function isUsingStock(array $data): bool
+    {
+        return $data['is_using_stock'];
+    }
+
+    /**
+     * @param array $data
+     * @return \Shopsys\FrameworkBundle\Model\Product\Brand\Brand|null
+     */
+    public function getBrand(array $data): ?Brand
+    {
+        if ((int)$data['brand'] > 0) {
+            return $this->brandFacade->getById((int)$data['brand']);
+        }
+
+        return null;
+    }
+
+    /**
+     * @param array $data
+     * @return bool
+     */
+    public function isSellingDenied(array $data): bool
+    {
+        return $data['calculated_selling_denied'];
+    }
+
+    /**
+     * @param array $data
+     * @return array
+     */
+    public function getAccessories(array $data): array
+    {
+        return $this->productElasticsearchProvider->getSellableProductArrayByIds($data['accessories']);
+    }
+
+    /**
+     * @param array $data
+     * @return string|null
+     */
+    public function getDescription(array $data): ?string
+    {
+        return $data['description'];
+    }
+
+    /**
+     * @param array $data
+     * @return array
+     */
+    public function getParameters(array $data): array
+    {
+        return $this->parameterWithValuesFactory->createParametersArrayFromProductArray($data);
+    }
+
+    /**
+     * @param array $data
+     * @return string|null
+     */
+    public function getSeoH1(array $data): ?string
+    {
+        return $data['seo_h1'];
+    }
+
+    /**
+     * @param array $data
+     * @return string|null
+     */
+    public function getSeoTitle(array $data): ?string
+    {
+        return $data['seo_title'];
+    }
+
+    /**
+     * @param array $data
+     * @return string|null
+     */
+    public function getSeoMetaDescription(array $data): ?string
+    {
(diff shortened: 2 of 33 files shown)


packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
```diff
@@ -264,85 +303,87 @@ protected function mapProduct(): array
     {
         return [
```
s3tezsky, line 304: Had you considered to introduce some kind of serializer from `Product` to array which would have the same structure as elasticsearch index? Then you could check what data came into this only once and prepare array which is same for both ways. 🤔 
When I push this idea further it could result in view object (rather than magic array) like it is in `ReadModel` what do you think? It should not take too much of effort.

packages/frontend-api/src/Model/Resolver/Products/ProductByUuidResolver.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->productElasticsearchProvider = $productElasticsearchProvider;
+    }
+
+    /**
+     * @param string $uuid
+     * @return array
+     */
+    public function resolver(string $uuid): array
+    {
+        if (Uuid::isValid($uuid) === false) {
```
s3tezsky, line 36: This check should not be neccessary in case you have it defined in `*.type.yaml`

packages/frontend-api/src/Model/Resolver/Products/ProductResolver.php
```diff
@@ -13,6 +13,9 @@
 use Shopsys\FrameworkBundle\Model\Product\Product;
 use Shopsys\FrameworkBundle\Model\Product\ProductFacade;
 
+/**
+ * @deprecated Use ProductByUuidResolver instead
```
s3tezsky, line 17: I am missing reason why is this class renamed. It is just obout naming or it has some other reason I do not see now?

packages/frontend-api/src/Resources/config/graphql-types/QueryDecorator.types.yaml
```diff
@@ -17,7 +17,7 @@ QueryDecorator:
                 description: "Returns list of ordered products that can be paginated using `first`, `last`, `before` and `after` keywords"
             product:
                 type: 'Product'
-                resolve: "@=resolver('product', [args['uuid']])"
+                resolve: "@=resolver('productByUuid', [args['uuid']])"
                 args:
                     uuid:
                         type: "Uuid!"
```
s3tezsky, line 23: This check will not let pass anything except UUID 😉

upgrade/UPGRADE-v9.1.0-dev.md
```diff
@@ -231,3 +231,9 @@ There you can find links to upgrade notes for other versions too.
 
 - add automatic string trimming and new Password type in frontend API ([#2127](https://github.com/shopsys/shopsys/pull/2127))
     - see #project-base-diff to update your project
+
+- use product by uuid from Elasticsearch in FE API (([#2131](https://github.com/shopsys/shopsys/pull/2131)))
```
s3tezsky, line 235: I am not sure if I like this line. Would not do something like this a bit better?

```
- FE API provides single product from Elasticsearch (([#2131](https://github.com/shopsys/shopsys/pull/2131)))
```

s3tezsky (changes requested):
Nice work.

I wrote down some points, which I belive would be nice to process. Also the ProductResolver is not really satisfying to me now.

project-base/src/Resources/definition/product/1.json
```diff
@@ -243,7 +243,8 @@
         }
       },
       "uuid": {
-        "type": "text"
+        "type": "keyword",
+        "index": true
```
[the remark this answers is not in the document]
grossmannmartin: oh, cool. Thanks! fixed in 50387b906 and 4c8a8d8b2

project-base/src/Resources/definition/product/1.json
```diff
@@ -199,12 +199,20 @@
           "parameter_id": {
             "type": "integer"
           },
+          "parameter_uuid": {
+            "type": "keyword",
+            "index": true
```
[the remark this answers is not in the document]
grossmannmartin: fixed in 50387b906 and 4c8a8d8b2

packages/frontend-api/src/Model/Resolver/Products/ProductByUuidResolver.php
grossmannmartin: oh, nice catch! fixed in c16720fe9

packages/frontend-api/src/Model/Resolver/Products/ProductResolver.php
grossmannmartin: The main reason is that I need different dependencies – keeping the current resolver would mean I need to deprecate current dependencies and add the new one in the BC manner. 
New resolver allows me to have dependencies clear, I can deprecate tho whole class, which is more understandable and I can switch to the previous resolver really easy if necessary.

packages/frontend-api/src/Resources/config/graphql-types/QueryDecorator.types.yaml
grossmannmartin: yup, fixed in c16720fe9

upgrade/UPGRADE-v9.1.0-dev.md
grossmannmartin: I don't think so. 
in upgrade notes should be instructions. I tried to rephrase the line in e365b3867

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
grossmannmartin: Well, I see a problem with this approach. 
When I map a product to the array, I need to query all data and create the complete array. 

If I don't query categories (for example), the anonymous function in the map is not executed and this behavior cannot be achieved when I create the whole array beforehand.

Currently, I need to maintain both (ES and DB mapping) because, for example, PromotedProducts are still obtained from DB.

grossmannmartin: refactored into multiple methods in 77e74b0cf

grossmannmartin: refactored again into multiple classes in fd642fdf072893055899e162879a73a409ef1a1c

packages/frontend-api/src/Model/Resolver/Products/ProductResolver.php
```diff
@@ -43,6 +46,15 @@ public function __construct(
...
+        @trigger_error(
+            sprintf(
+                'The "%s" class is deprecated and will be removed in the next major. Use "%s" instead.',
+                self::class,
+                ProductByUuidResolver::class
+            ),
+            E_USER_DEPRECATED
+        );
+
         if (Uuid::isValid($uuid) === false) {
```
grossmannmartin, line 58: validation of uuids is in separate (dependent) PR #2140

packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
```diff
@@ -63,4 +63,27 @@ public function createMultipleForProduct(Product $product): array
 
         return $parametersWithValues;
     }
+
+    /**
+     * @param array $productData
+     * @return array
+     */
+    public function createParameterArrayFromProductArray(array $productData): array
```
s3tezsky, line 71: Not really sure about the method name. Should not it by in plural `createParametersArrayFromProductArray`?

Singular should be used to map single parameter which could be used for assigning into returning array. Maybe cosider renaming it and refactoring?

packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
```diff
@@ -63,4 +63,27 @@ public function createMultipleForProduct(Product $product): array
...
+    {
+        $parametersWithValues = [];
+
+        foreach ($productData['parameters'] as $product) {
+            $parametersWithValues[] = [
```
s3tezsky, line 76: From previous comment: Please consider extracting to `createParameterArrayFromProductArray`.

packages/frontend-api/src/Model/Resolver/Products/DataMapper/ProductArrayFieldMapper.php
```diff
@@ -0,0 +1,236 @@
...
+        return $data['is_using_stock'];
+    }
+
+    /**
+     * @param array $data
+     * @return \Shopsys\FrameworkBundle\Model\Product\Brand\Brand|null
+     */
+    public function getBrand(array $data): ?Brand
+    {
+        if ((int)$data['brand'] > 0) {
```
s3tezsky, line 140: `(int)[2, 3]` will return `1`, so it passes the condition too. I am not sure if it is OK or not 🤔. Probably it should be handled from type definitions?

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
```diff
@@ -4,190 +4,93 @@
...
 use Shopsys\FrameworkBundle\Model\Product\Collection\ProductCollectionFacade;
 use Shopsys\FrameworkBundle\Model\Product\Flag\FlagFacade;
 use Shopsys\FrameworkBundle\Model\Product\Product;
-use Shopsys\FrameworkBundle\Model\Product\ProductFacade;
-use Shopsys\FrontendApiBundle\Model\Parameter\ParameterWithValuesFactory;
-use Shopsys\FrontendApiBundle\Model\Product\ProductAccessoryFacade;
+use Shopsys\FrontendApiBundle\Model\Resolver\Products\DataMapper\ProductArrayFieldMapper;
+use Shopsys\FrontendApiBundle\Model\Resolver\Products\DataMapper\ProductEntityFieldMapper;
 
 class ProductResolverMap extends ResolverMap
```
s3tezsky, line 21: Why is not the whole class deprecated and new class `ProductResolverByUuidMap` introduced? Would not make it more sence, not just because of tons of changes here?

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
```diff
@@ -4,190 +4,93 @@
...
      * @required
-     * @param \Shopsys\FrameworkBundle\Model\Product\ProductFacade $productFacade
+     * @param \Shopsys\FrontendApiBundle\Model\Resolver\Products\DataMapper\ProductArrayFieldMapper $productArrayFieldMapper
      * @internal This function will be replaced by constructor injection in next major
      */
-    public function setProductFacade(ProductFacade $productFacade): void
+    public function setProductArrayFieldMapper(ProductArrayFieldMapper $productArrayFieldMapper): void
     {
-        if ($this->productFacade !== null && $this->productFacade !== $productFacade) {
+        if ($this->productArrayFieldMapper !== null && $this->productArrayFieldMapper !== $productArrayFieldMapper) {
```
s3tezsky, line 88: I know it is ok for our standards as it fits so far but could you split it to multiline as it is the way how we are using it almost everywhere?

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
```diff
@@ -258,91 +162,40 @@ protected function map(): array
...
-            },
-            'isUsingStock' => function ($data) {
-                return $data instanceof Product ? $data->isUsingStock() : $data['is_using_stock'];
-            },
-            'brand' => function ($data) {
-                if ($data instanceof Product) {
-                    return $data->getBrand();
-                }
+            self::RESOLVE_FIELD => function ($value, ArgumentInterface $args, ArrayObject $context, ResolveInfo $info) {
+                $methodName = 'get' . ucfirst($info->fieldName);
```
s3tezsky, line 171: I have internall struggle with using only methods prefixed by `get`. E.g. I would like to use `isSellingDenied()` but I cannot without changing this callback.

Could the getting and validating method name be extracted into method?

s3tezsky (changes requested):
Wow, good job down here. It is getting bigger but I see a light!

Could you please walk through my minor issues.

packages/frontend-api/src/Model/Resolver/Products/DataMapper/ProductArrayFieldMapper.php
grossmannmartin: Well, the behavior is not changed, the code was just moved.
Not changed as each product can have only one brand (see ProductExportRepository)

s3tezsky: Understand. The condition just seems weird to me. It could be `$data['brand'] !== ''` which could make it more obvious about incoming data. I will leave it up to you.

packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
grossmannmartin: okay, in 9be1288de

packages/frontend-api/src/Model/Parameter/ParameterWithValuesFactory.php
grossmannmartin: okay, in 9be1288de

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
grossmannmartin: okay, in 01a4cd41fa1b21027952092c8132130c5e95ab78

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
grossmannmartin: allowed to use `get`, `is`, or no prefix in 01a4cd41fa1b21027952092c8132130c5e95ab78

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
grossmannmartin: That's not so easy... This map is not used just for ByUuid resolver, but for any work with the product in FE API (promoted product, lists), so the name is not ideal.

Changes are not so terrifying as it may seem from the PR. Changes between the already released version (9.0.4) and the current changeset are in deprecation of the three fields and three methods.
I'm not sure whether it's worth the new class and deprecate the old one.

packages/frontend-api/src/Model/Resolver/Products/DataMapper/ProductArrayFieldMapper.php
```diff
@@ -126,7 +126,7 @@ public function getStockQuantity(array $data): ?int
      * @param array $data
      * @return bool
      */
-    public function getIsUsingStock(array $data): bool
+    public function issUsingStock(array $data): bool
```
s3tezsky:
```suggestioin
    public function isUsingStock(array $data): bool
```

s3tezsky: Ah, it has been fixed in later commit!

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
s3tezsky: Nice!

packages/frontend-api/src/Model/Resolver/Products/ProductResolverMap.php
s3tezsky: > That's not so easy... This map is not used just for ByUuid resolver, but for any work with the product in FE API (promoted product, lists), so the name is not ideal.

Makes sence, did not realize it before. Thanks!

s3tezsky (approved):

sslt (approved):

PR #2133, opened 2020-11-25
[framework] Enabled caching for DomainRouter and LocalizedRouter

pk16011990:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Static and localized routes are now cached into file.
|New feature| No 
|Fixes issues| closes #2106  

### DIFF
--- a/packages/framework/src/Component/Router/DomainRouterFactory.php
+++ b/packages/framework/src/Component/Router/DomainRouterFactory.php
@@ -2,15 +2,18 @@
 
 namespace Shopsys\FrameworkBundle\Component\Router;
 
+use BadMethodCallException;
 use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Domain\Exception\InvalidDomainIdException;
+use Shopsys\FrameworkBundle\Component\Environment\EnvironmentType;
 use Shopsys\FrameworkBundle\Component\Router\Exception\RouterNotResolvedException;
 use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRouterFactory;
+use Symfony\Bundle\FrameworkBundle\Routing\Router;
 use Symfony\Component\Config\Loader\LoaderInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Routing\RequestContext;
-use Symfony\Component\Routing\Router;
 
 class DomainRouterFactory
 {
@@ -30,6 +33,7 @@ class DomainRouterFactory
     protected $domain;
 
     /**
+     * @deprecated This loader is deprecated and will be removed in the next major. Use Symfony\Bundle\FrameworkBundle\Routing\Router instead of Symfony\Component\Routing\Router without this dependency.
      * @var \Symfony\Component\Config\Loader\LoaderInterface
      */
     protected $configLoader;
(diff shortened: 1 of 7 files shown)


packages/framework/src/Component/Router/DomainRouterFactory.php
```diff
@@ -49,28 +52,80 @@ class DomainRouterFactory
...
         $routerConfiguration,
-        LoaderInterface $configLoader,
+        ?LoaderInterface $configLoader,
         LocalizedRouterFactory $localizedRouterFactory,
         FriendlyUrlRouterFactory $friendlyUrlRouterFactory,
         Domain $domain,
-        RequestStack $requestStack
+        RequestStack $requestStack,
+        ?ContainerInterface $container = null,
+        ?string $cacheDir = null
```
pk16011990, line 83: Why is it passed there? It can be taken from the container directly.

s3tezsky: It can be taaken from container, as is in fallback in costructor. I just do not like the way I am just picking some configuration argument directly from container an want to have it strictly passed via constructor.

packages/framework/src/Component/Router/LocalizedRouterFactory.php
```diff
@@ -72,4 +131,13 @@ protected function getLocaleRouterResourceByLocale(string $locale): string
...
         return str_replace('*', $locale, $this->localeRoutersResourcesFilepathMask);
     }
+
+    /**
+     * @param string $locale
+     * @return string
+     */
+    protected function getRoutingCacheDir(string $locale): string
+    {
+        return $this->cacheDir . '/routing/localized/' . $locale;
```
grossmannmartin, line 141: We usually put paths in the specialized file (paths.yaml), or in services file directly. 
Having the directory configuration here is confusing as cacheDir is passed via constructor, but it's not the final cacheDir for routing.

packages/framework/src/Component/Router/DomainRouterFactory.php
grossmannmartin: Almost seems like ParameterBag here could solve it nicely.

grossmannmartin (changes requested):

packages/framework/src/Component/Router/DomainRouterFactory.php
s3tezsky: You are right. To me it is like an antipattern which I would love to avoid when possible.

packages/framework/src/Component/Router/DomainRouterFactory.php
```diff
@@ -49,28 +52,80 @@ class DomainRouterFactory
...
+        $this->container = $container;
+
+        if ($cacheDir === null) {
+            $deprecationMessage = sprintf(
+                'The argument "$cacheDir" is not provided by constructor "%s()". In the next major it will be required.',
+                __METHOD__
+            );
+            @trigger_error($deprecationMessage, E_USER_DEPRECATED);
+
+            $cacheDir = $this->container->getParameter('kernel.cache_dir');
```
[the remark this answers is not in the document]
s3tezsky: Good catch. Falback was moved right to `getBasicRouter()`.

packages/framework/src/Component/Router/LocalizedRouterFactory.php
```diff
@@ -26,15 +29,69 @@ class LocalizedRouterFactory
...
+        $this->container = $container;
+
+        if ($cacheDir === null) {
+            $deprecationMessage = sprintf(
+                'The argument "$cacheDir" is not provided by constructor "%s()". In the next major it will be required.',
+                __METHOD__
+            );
+            @trigger_error($deprecationMessage, E_USER_DEPRECATED);
+
+            $cacheDir = $this->container->getParameter('kernel.cache_dir');
```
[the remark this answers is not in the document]
s3tezsky: Fixed too. Fallback is in `getRoutingCacheDir()`

packages/framework/src/Resources/config/services.yaml
```diff
@@ -476,7 +476,9 @@ services:
             - { name: router, priority: 70 }
 
     Shopsys\FrameworkBundle\Component\Router\DomainRouterFactory:
-        arguments: ['%router.resource%']
+        arguments:
+            $routerConfiguration: '%router.resource%'
+            $cacheDir: '%kernel.cache_dir%'
```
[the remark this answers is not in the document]
s3tezsky: now it is!

packages/framework/src/Component/Router/LocalizedRouterFactory.php
s3tezsky: moved to `paths.yaml`

packages/framework/src/Component/Router/DomainRouterFactory.php
grossmannmartin: Okay, I see. Just a final thought about this – does it make sense to swap the order of the arguments? 
First $cacheDir and container as the last. Because in the next major we probably remove the dependency on the container as it's not necessary when $cacheDir is provided and it seems to me it will be a little more clear in the future upgrade notes

upgrade/UPGRADE-v9.1.0-dev.md
```diff
@@ -231,3 +231,12 @@ There you can find links to upgrade notes for other versions too.
 
 - add automatic string trimming and new Password type in frontend API ([#2127](https://github.com/shopsys/shopsys/pull/2127))
     - see #project-base-diff to update your project
+
+- provide cache directory to DomainRouterFactory and LocalizedRouterFactory ([#2133](https://github.com/shopsys/shopsys/pull/2133))
+    - the step is necessary in case you have extended following classes:
+        - `Shopsys\FrameworkBundle\Component\Router\DomainRouterFactory`
+        - `Shopsys\FrameworkBundle\Component\Router\LocalizedRouterFactory`
+    - pass cache directory argument from your `services.yaml`
```
grossmannmartin, line 239:
```suggestion
    - pass cache directory argument in your `services.yaml` as `$cacheDir` argument
```

packages/framework/src/Component/Router/DomainRouterFactory.php
s3tezsky: Not really. The container is required and also will be in the future as a dependency for `Symfony\Bundle\FrameworkBundle\Routing\Router`.

grossmannmartin (commented):
I tested the behavior and I found one little problem. 

When I change the localized router file (for example `routing_front_en.yaml`), in dev environment the change is not propagated until I clear the cache manually. 

This can cause a little headache in project development.

packages/framework/src/Component/Router/DomainRouterFactory.php
grossmannmartin: My bad, thanks for explanation

grossmannmartin (approved):

sslt (approved):

PR #2144, opened 2020-12-02
[framework] FeedExportFactory: fixed casting of $lastSeekId to nullable int

### LINKED ISSUE #2039: Error casting $lastSeekId variable in FeedExportFactory
### What is happening

the begin block is not generated in xml feeds.

This is caused by recasting the $ lastSeekId variable to an integer in
an instance of the FeedExport class

```
return new FeedExport(
            $feed,
            $domainConfig,
            $feedRenderer,
            $this->filesystem,
            $this->localFilesystem,
            $this->mountManager,
            $this->em,
            $feedFilepath,
            $feedLocalFilepath,
            (int)$lastSeekId
        );
```
in the generateBatch function, this variable is then tested for null, but it never occurs and therefore the begin block is never inserted into the resulting xml feed. The variable has a value of 0 after recasting.

```
public function generateBatch(): void
    {
       ...
        if ($this->lastSeekId === null) {
            $this->writeToFeed($this->feedRenderer->renderBegin($this->domainConfig));
        }
       ...

```

### Expected result

the expected result is the insertion of a begin block at the beginning of the xml feed.

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR fixes retyping `$lastSeekId` from `string\|null` to `int\|null`. At this moment generating feeds don't add begin to feeds (eg. [this](https://github.com/shopsys/shopsys/blob/master/packages/product-feed-heureka-delivery/src/Resources/views/feed.xml.twig#L1-L4)) because [FeedExport class render begin only if $lastSeekId is null](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Feed/FeedExport.php#L141-L143) and `$lastSeekId` is always retyped to int so null never comes and begin is never rendered. This is issue in 9.1 version so I recommend you add this there. Thanks
|New feature| No
|Fixes issues| fixes #2039  

### DIFF
--- a/packages/framework/src/Model/Feed/FeedExportFactory.php
+++ b/packages/framework/src/Model/Feed/FeedExportFactory.php
@@ -86,6 +86,7 @@ class FeedExportFactory
         $feedRenderer = $this->feedRendererFactory->create($feed);
         $feedFilepath = $this->feedPathProvider->getFeedFilepath($feed->getInfo(), $domainConfig);
         $feedLocalFilepath = $this->feedPathProvider->getFeedLocalFilepath($feed->getInfo(), $domainConfig);
+        $lastSeekId = $lastSeekId !== null ? (int)$lastSeekId : $lastSeekId;
 
         return new FeedExport(
             $feed,
@@ -97,7 +98,7 @@ class FeedExportFactory
             $this->em,
             $feedFilepath,
             $feedLocalFilepath,
-            (int)$lastSeekId
+            $lastSeekId
         );
     }
 }


sslt:
Thank you @sspooky13, we will include your PR into upcoming SSFW 9.1

s3tezsky:
Related to #2039

s3tezsky (approved):
Thanks @sspooky13!

grossmannmartin (approved):
Good catch! Reviewed and tested.

sslt (approved):

PR #2146, opened 2020-12-03
[shopsys] cache friendly url slugs for generator into redis

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Each friendly URL created in twig recieved slug from database. This PR adds slugs to be cached in redis to reduce queries.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGenerator.php
+++ b/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGenerator.php
@@ -2,6 +2,7 @@
 
 namespace Shopsys\FrameworkBundle\Component\Router\FriendlyUrl;
 
+use BadMethodCallException;
 use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
 use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\Exception\FriendlyUrlNotFoundException;
 use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\Exception\MethodGenerateIsNotSupportedException;
@@ -12,6 +13,7 @@ use Symfony\Component\Routing\RequestContext;
 use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
 use Symfony\Component\Routing\RouteCompiler;
+use Symfony\Contracts\Cache\CacheInterface;
 
 class FriendlyUrlGenerator extends BaseUrlGenerator
 {
@@ -21,16 +23,40 @@ class FriendlyUrlGenerator extends BaseUrlGenerator
     protected $friendlyUrlRepository;
 
     /**
+     * @var \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlCacheKeyProvider|null
+     */
+    protected ?FriendlyUrlCacheKeyProvider $friendlyUrlCacheKeyProvider;
+
+    /**
+     * @var \Symfony\Contracts\Cache\CacheInterface|null
+     */
+    protected ?CacheInterface $mainFriendlyUrlSlugCache;
+
+    /**
      * @param \Symfony\Component\Routing\RequestContext $context
      * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRepository $friendlyUrlRepository
+     * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlCacheKeyProvider|null $friendlyUrlCacheKeyProvider
+     * @param \Symfony\Contracts\Cache\CacheInterface|null $mainFriendlyUrlSlugCache
      */
     public function __construct(
         RequestContext $context,
-        FriendlyUrlRepository $friendlyUrlRepository
+        FriendlyUrlRepository $friendlyUrlRepository,
+        ?FriendlyUrlCacheKeyProvider $friendlyUrlCacheKeyProvider = null,
+        ?CacheInterface $mainFriendlyUrlSlugCache = null
     ) {
+        if ($mainFriendlyUrlSlugCache === null) {
+            $deprecationMessage = sprintf(
+                'The argument "$mainFriendlyUrlSlugCache" is not provided by constructor in "%s". In the next major it will be required.',
+                self::class
+            );
+            @trigger_error($deprecationMessage, E_USER_DEPRECATED);
+        }
+
         parent::__construct(new RouteCollection(), $context, null);
 
         $this->friendlyUrlRepository = $friendlyUrlRepository;
+        $this->friendlyUrlCacheKeyProvider = $friendlyUrlCacheKeyProvider;
+        $this->mainFriendlyUrlSlugCache = $mainFriendlyUrlSlugCache;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGenerator.php
```diff
@@ -60,18 +76,15 @@ public function generateFromRouteCollection(
         $entityId = $parameters['id'];
         unset($parameters['id']);
 
-        try {
-            $friendlyUrl = $this->friendlyUrlRepository->getMainFriendlyUrl(
```
grossmannmartin, line 64: Why not add caching into friendlyUrlRepository? 

right now it's not used while exporting, and in the future, cannot be used in frontend API without hassle.

project-base/config/packages/snc_redis.yaml
```diff
@@ -23,6 +23,13 @@ snc_redis:
             type: 'phpredis'
             alias: 'session'
             dsn: 'redis://%redis_host%'
+        friendly_url_slugs:
+            type: 'phpredis'
+            alias: 'friendly_url_slugs'
+            dsn: 'redis://%redis_host%'
+            logging: true
```
grossmannmartin, line 30: can you explain why logging is here?

grossmannmartin (commented):
I would suggest a different commit message
something like `main friendly url slugs are now cached in redis`

I'm not sure about the approach and I would like to discuss this PR personally

packages/framework/src/Component/Router/FriendlyUrl/CachedFriendlyUrlSlugFacade.php
```diff
@@ -0,0 +1,74 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Router\FriendlyUrl;
+
+use Doctrine\Common\Cache\CacheProvider;
+use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\Exception\FriendlyUrlNotFoundException;
+
+class CachedFriendlyUrlSlugFacade
+{
+    protected const LIFETIME = 60;
```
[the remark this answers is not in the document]
s3tezsky: This one was removed, now it is cached permanently

packages/framework/src/Component/Router/FriendlyUrl/CachedFriendlyUrlSlugFacade.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Router\FriendlyUrl;
+
+use Doctrine\Common\Cache\CacheProvider;
+use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\Exception\FriendlyUrlNotFoundException;
+
+class CachedFriendlyUrlSlugFacade
```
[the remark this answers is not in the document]
s3tezsky: Good point with renaming those. Anyway this class was removed at all in the end.

packages/framework/src/Component/Router/FriendlyUrl/CachedFriendlyUrlSlugFacade.php
```diff
@@ -0,0 +1,74 @@
...
+     * @param string $routeName
+     * @param int $entityId
+     * @return string
+     */
+    public function getSlug(int $domainId, string $routeName, int $entityId): string
+    {
+        $friendlyUrl = $this->cacheProvider->fetch($this->getCacheId($routeName, $domainId, $entityId));
+
+        if ($friendlyUrl === false) {
+            throw new FriendlyUrlNotFoundException();
```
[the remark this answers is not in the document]
s3tezsky: Removed.

packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGenerator.php
s3tezsky: After discussion we are continuing this way in a bit different way. Thanks to Caching component from Symfony it will be easy to use in other use cases too.

project-base/config/packages/snc_redis.yaml
s3tezsky: Yeah, I forgot this one from testing. Now it is removed.

packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGenerator.php
```diff
@@ -20,17 +22,41 @@ class FriendlyUrlGenerator extends BaseUrlGenerator
...
      * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRepository $friendlyUrlRepository
+     * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlCacheKeyProvider|null $friendlyUrlCacheKeyProvider
+     * @param \Symfony\Contracts\Cache\CacheInterface|null $mainFriendlyUrlSlugCache
      */
     public function __construct(
         RequestContext $context,
-        FriendlyUrlRepository $friendlyUrlRepository
+        FriendlyUrlRepository $friendlyUrlRepository,
+        ?FriendlyUrlCacheKeyProvider $friendlyUrlCacheKeyProvider = null,
+        ?CacheInterface $mainFriendlyUrlSlugCache = null
```
grossmannmartin, line 45: This class is available as a service (thanks to autodiscovery for classes with `Generator` suffix) and also by the implemented setter injection via `setFriendlyUrlCacheKeyProvider`.

However, `$mainFriendlyUrlSlugCache` does not have the same cache pool if I request this class as a service and when used inside `FriendlyUrlFacade`. 

<details>
  <summary>Click to see dumped content</summary>
  
 
Snímek obrazovky 2020-12-08 v 17 01 58


</details>

I suggest either remove setter injection (and maybe remove this class from autowiring in project base so it's not a bc break), or configure the service with proper cache pool)

packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlRouterFactory.php
```diff
@@ -48,10 +66,44 @@ public function createRouter(DomainConfig $domainConfig, RequestContext $context
         return new FriendlyUrlRouter(
             $context,
             $this->configLoader,
-            new FriendlyUrlGenerator($context, $this->friendlyUrlRepository),
+            new FriendlyUrlGenerator(
```
grossmannmartin, line 69: Isn't possible to use generator as a service here?
I think it would be more clear and we can inject just the dependency we need for this class (right now I need to inject two dependencies just to pass it to another constructor)

project-base/config/services.yaml
```diff
@@ -123,3 +123,12 @@ services:
     Psr\Http\Message\ServerRequestInterface:
         factory: ['@Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory', 'createRequest']
         arguments: ['@=service("request_stack").getMasterRequest()']
+
+    Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlRouterFactory:
+        arguments:
+            $friendlyUrlRouterResourceFilepath: '%shopsys.router.friendly_url_router_filepath%'
+            $mainFriendlyUrlSlugCache: '@cache.main_friendly_url_slugs'
```
grossmannmartin, line 130: I'm thinking about using Symfony cache pool behavior.
The proper cache is auto-wirable if the argument is named similarly to cache pool name
 e.g. if you name pool `main_friendly_url_slug_cache` in `framework.yaml`  (without `cache.` prefix)

 it can be seen as magic but in the end, it's friendlier to the developers
- you can remove both service definitions (no need to specify the service manually)
- developers don't need to specify the correct cache when extending ( right now it's error-prone if I extend the class and forget about argument)
- it solves the problem with the generator as a service (my first comment)

see 
https://symfony.com/doc/current/cache.html#creating-custom-namespaced-pools

grossmannmartin (changes requested):

project-base/config/services.yaml
s3tezsky: Thakns for teaching me a new symfony magic!

Magic was applied.

packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlGenerator.php
s3tezsky: I have changed the name of cache pool.

packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlRouterFactory.php
s3tezsky: No it is not.

It would be nice but there is `$context` needed which is passed during factoring the `FriendlyUrlRouter`.

grossmannmartin (approved):

grossmannmartin:
rebased to master to re-trigger travis build

sslt (approved):

PR #2174, opened 2021-01-05
[project-base] validate search text is printable

### LINKED ISSUE #2063: 500 on search spacial character
### What is happening
Page returns 500 response code when search request looks like `https://example.com/hledani?q=%c1`.

It fails on:
```
CRITICAL - Uncaught PHP Exception Doctrine\DBAL\Exception\DriverException: "An exception occurred while executing 'SELECT c0_.id AS id_4, c0_.uuid AS uuid_5, c0_.level AS level_6, c0_.lft AS lft_7, c0_.rgt AS rgt_8, c1_.name AS name_9, c1_.id AS id_10, c1_.locale AS locale_11, c0_.parent_id AS parent_id_12, c1_.translatable_id AS translatable_id_13 FROM categories c0_ INNER JOIN category_domains c2_ ON (c2_.category_id = c0_.id) INNER JOIN category_translations c1_ ON c0_.id = c1_.translatable_id AND (c1_.locale = ?) WHERE c0_.parent_id IS NOT NULL AND c2_.domain_id = ? AND c2_.visible = true AND normalize(c1_.name) LIKE normalize(?) ORDER BY c0_.lft ASC' with params ["cs", 1, "\x25\xc1\x25"]:  SQLSTATE[22021]: Character not in repertoire: 7 ERROR:  invalid byte sequence for encoding "UTF8": 0xc1 0x25" at /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractPostgreSQLDriver.php line 83
```

### Expected result
Response is without error :)

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| #2063 
|New feature| No 
|Fixes issues| fixes #2063 

_This PR needs to be rebased on 9.0 branch and have added upgrade notes after release of v9.1.0_

### DIFF
--- a/packages/framework/src/Component/String/TransformString.php
+++ b/packages/framework/src/Component/String/TransformString.php
@@ -126,4 +126,13 @@ class TransformString
 
         return preg_replace($search, $replace, $subject, $limit);
     }
+
+    /**
+     * @param string $string
+     * @return string
+     */
+    public static function replaceInvalidUtf8CharactersByQuestionMark(string $string): string
+    {
+        return mb_convert_encoding($string, 'UTF-8', 'UTF-8');
+    }
 }
--- a/project-base/src/Controller/Front/ProductController.php
+++ b/project-base/src/Controller/Front/ProductController.php
@@ -6,6 +6,7 @@ namespace App\Controller\Front;
 
 use App\Form\Front\Product\ProductFilterFormType;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\String\TransformString;
 use Shopsys\FrameworkBundle\Model\Category\Category;
 use Shopsys\FrameworkBundle\Model\Category\CategoryFacade;
 use Shopsys\FrameworkBundle\Model\Module\ModuleFacade;
@@ -26,6 +27,7 @@ use Symfony\Component\HttpFoundation\Request;
 class ProductController extends FrontBaseController
 {
     public const SEARCH_TEXT_PARAMETER = 'q';
+    private const SEARCH_TEXT_DEFAULT_VALUE = '';
     public const PAGE_QUERY_PARAMETER = 'page';
     public const PRODUCTS_PER_PAGE = 12;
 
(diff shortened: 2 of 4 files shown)


project-base/src/Controller/Front/ProductController.php
```diff
@@ -259,7 +260,10 @@ public function listByBrandAction(Request $request, $id)
      */
     public function searchAction(Request $request)
     {
-        $searchText = $request->query->get(self::SEARCH_TEXT_PARAMETER, '');
+        $searchText = (string)$request->query->get(self::SEARCH_TEXT_PARAMETER, self::SEARCH_TEXT_DEFAULT_VALUE);
+        if ($this->isRequestSearchTextValid($searchText)) {
```
TomasLudvik, line 264: That method actually checks, if the search text is invalid.

TomasLudvik: In invalid state, you should redirect user to homepage and display flash message saying something like: "Search text does not include printable characters".

That is because right now you actually redirect user to page that displays products with space in name and that is weird.

TomasLudvik (changes requested):

project-base/src/Controller/Front/ProductController.php
s3tezsky: > That method actually checks, if the search text is invalid.

> In invalid state, you should redirect user to homepage

both fixed, thanks for good points 👍 

---

> and display flash message saying something like: "Search text does not include printable characters".

not fixed, @sslt agreed with quiet redirect to homepage

packages/framework/src/Component/String/TransformString.php
```diff
@@ -126,4 +126,13 @@ public static function replaceOccurences(string $search, string $replace, string
 
         return preg_replace($search, $replace, $subject, $limit);
     }
+
+    /**
+     * @param string $string
+     * @return string
+     */
+    public static function replaceNonPrintableCharactersByQuestionMark(string $string): string
```
TomasLudvik, line 134: I think this method should have some more accurate name, as replacing non printable characters is not exactly what it does. Maybe replaceNonUTF8CharactersByQuestionMar

s3tezsky: Ok, replaced to `replaceInvalidUtf8CharactersByQuestionMark` as we agreed 👍

TomasLudvik (approved):

sslt (approved):

PR #2185, opened 2021-01-12
[framework] fixed cart when one of cart item is not sellable

### LINKED ISSUE #1991: Excluding product from sale by admin causes error in existing cart
shopsys/framework": "9.0.0"

### What is happening
I add items to cart as customer. Then admin excludes product from sale. If I have this excluded product already in cart, it causes error 500, if exluded item is not last in my cart (in array index).

1. Add product to cart
2. Add another product to cart 
3. Sign in as admin and exclude product from sale (with lower array index)
4. load cart page

class CartWatcherFacade uses checkNotListableItems function to remove not listable items. The process is handled by removeItemById function which causes array index mismatch. Items in carts are indexed as:
[0 => Product1, 1 => Product2, 2 => Product3] If Product1 is removed by CartWatcherFacade, the indexes for other products stay same [1 => Product2, 2 => Product3]. That causes mismatch in CartController in indexAction for $cartItems vs. $cartItemPrices

$cartItems: [**1** => Product2, **2** => Product3]
$cartItemPrices: [**0** => Product2Price, **1** => Product3Price]

### Expected result
function removeItemById should rearrange array indexes after any element is removed

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| see #1991
|New feature| No 
|Fixes issues| fixes #1991  

### DIFF
--- a/packages/framework/src/Model/Cart/Cart.php
+++ b/packages/framework/src/Model/Cart/Cart.php
@@ -102,7 +102,7 @@ class Cart
      */
     public function getItems()
     {
-        return $this->items->toArray();
+        return array_values($this->items->toArray());
     }
 
     /**


project-base/src/Controller/Front/CartController.php
```diff
@@ -110,8 +110,7 @@ public function __construct(
     public function indexAction(Request $request)
     {
         $cart = $this->cartFacade->findCartOfCurrentCustomerUser();
-        $cartItems = $cart === null ? [] : $cart->getItems();
-
+        $cartItems = $cart === null ? [] : array_values($cart->getItems());
```
TomasLudvik, line 113: I have been thinking about this. On first look it might appear to someone like not necessary code and might want to delete it in future. I would add method getValues to Cart like:
```
public function getValues()
{
    $this->items->getValues();
}
```

and use that method here.

s3tezsky: I dissagree. The method you are suggesting is not helping at all. I would rather add comment on line above with an explanation of doing so. Then I will pray for reworking CartItems to contain even prices - having it separated is really hell.

TomasLudvik (approved):

TomasLudvik (approved):

sslt (approved):

PR #2196, opened 2021-01-18
[docs] use `CategoryResolverMap` as an example of `ResolverMap` in FE API introduction

malyMiso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| `ProductResolverMap::map()` has been changed, since introducing `ProductEntityFieldMapper` and `ProductArrayFieldMapper` in  SSFW 9.1. There is also custom section called `ProductResolverMap`.I think it would be better to use `CategoryResolverMap` as an example of `ResolverMap` for better understanding how resolver mappers work (Anyone can look at current implementation of `CategoryResolverMap`).
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/docs/frontend-api/introduction-to-frontend-api.md
+++ b/docs/frontend-api/introduction-to-frontend-api.md
@@ -203,7 +203,7 @@ ResolverMap can be created as a child of `Overblog\GraphQLBundle\Resolver\Resolv
 Example of `ResolverMap`:
 
 ```php
-class ProductResolverMap extends ResolverMap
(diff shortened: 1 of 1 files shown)


TomasLudvik (approved):

grossmannmartin (approved):

sslt (approved):

s3tezsky (changes requested):
Please change the target to 9.1 instead of master 🙏 .

s3tezsky (approved):

PR #2200, opened 2021-01-20
[shopsys] prevent to pass null to not-nullable argument to ListedProductViewFactory::createFromArray()

### LINKED ISSUE #2098: Passing nullable value to required argument in ListedProductViewFactory.php
Found in https://github.com/shopsys/shopsys/pull/2090#discussion_r512077387

### What is happening

When creating `ListedProductView` there may be passed `null` (actually it cannot, because of our recalculations and so on) to argument requiring `ProductPrice`.

https://github.com/shopsys/shopsys/blob/0d77eb9c5e6309df3e91eeaacf60b3c20419e784/packages/read-model/src/Product/Listed/ListedProductViewFactory.php#L127

### Expected result

The edgecase without found price is handled properly.

_E.g.: There could be raised and hadled exception when the `ProductPrice` is not available instead of returning null._

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| See linked issue
|New feature| No 
|Fixes issues| fixes #2098  

### DIFF
--- a/packages/read-model/src/Product/Listed/ListedProductViewElasticFacade.php
+++ b/packages/read-model/src/Product/Listed/ListedProductViewElasticFacade.php
@@ -8,6 +8,7 @@ use BadMethodCallException;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Paginator\PaginationResult;
 use Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser;
+use Shopsys\FrameworkBundle\Model\Pricing\Exception\NoProductPriceForPricingGroupException;
 use Shopsys\FrameworkBundle\Model\Product\Accessory\ProductAccessoryFacade;
 use Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData;
 use Shopsys\FrameworkBundle\Model\Product\Product;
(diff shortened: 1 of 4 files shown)


packages/read-model/src/Product/Listed/ListedProductViewFactory.php
```diff
@@ -144,12 +145,20 @@ public function createFromProduct(Product $product, ?ImageView $imageView, Produ
      */
     public function createFromArray(array $productArray, ?ImageView $imageView, ProductActionView $productActionView, PricingGroup $pricingGroup): ListedProductView
     {
+        $productPrice = $this->priceFactory->createProductPriceFromArrayByPricingGroup(
+            $productArray['prices'],
+            $pricingGroup
+        );
+        if ($productPrice === null) {
+            throw new NoProductPriceForPricingGroupException($productArray['id'], $pricingGroup->getId());
```
grossmannmartin, line 153: I'm thinking about throwing this exception directly in the factory. Because from a create method I expect creating an object and not return null. What do you think?

packages/read-model/src/Product/Listed/ListedProductViewFactory.php
```diff
@@ -144,12 +145,20 @@ public function createFromProduct(Product $product, ?ImageView $imageView, Produ
      */
     public function createFromArray(array $productArray, ?ImageView $imageView, ProductActionView $productActionView, PricingGroup $pricingGroup): ListedProductView
     {
+        $productPrice = $this->priceFactory->createProductPriceFromArrayByPricingGroup(
```
grossmannmartin, line 148: can you please add deprecation notice into method `createProductPriceFromArrayByPricingGroup` when trying to return null as we discussed?

packages/read-model/src/Product/Listed/ListedProductViewFactory.php
s3tezsky: I agree but we cannot do it in minor version as discussed.

packages/read-model/src/Product/Listed/ListedProductViewFactory.php
s3tezsky: Sure. It has been done in 822e9bbdbf1f19011910b9675d4b963e2d785040

packages/read-model/src/Product/Listed/ListedProductViewElasticFacade.php
```diff
@@ -269,12 +270,16 @@ protected function createFromArray(array $productsArray): array
...
-            );
+            try {
+                $listedProductViews[$productId] = $this->listedProductViewFactory->createFromArray(
+                    $productArray,
+                    $imageViews[$productId],
+                    $this->productActionViewFactory->createFromArray($productArray),
+                    $this->currentCustomerUser->getPricingGroup()
+                );
+            } catch (NoProductPriceForPricingGroupException $exception) {
+                continue;
```
grossmannmartin, line 281: I got feedback on this line – silently dropped exceptions can be confusing when you're solving some unexpected behavior. 
I'm not really sure how to deal with this information.

grossmannmartin (approved):

sslt (approved):

PR #2202, opened 2021-01-22
[shopsys] entites are refreshed after product visibility recalculations

### LINKED ISSUE #2108: OrderProductFacade does not refresh product entities after recalculations
### What is happening
[OrderProduct::runRecalculationsAfterStockQuantityChange](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Order/Item/OrderProductFacade.php) runs product recalculations (visibility, selling denied, availability, etc.). All recalculations are run as native SQL queries.
After that run product entity is not refreshed.  What it means during active/running request is that the product is still visible even when it was sold out.

In SSFW itself it causes no problems. 
It can be a problem after extending functionality. It is hard to find out, why product is still visible, even when it was sold out. In source code, everything looks fine. It is a bit mysterious, isn't it ? 

Because of this mysterious behavior, another issue was discovered and solved in #2084 

### Expected result
All product data become actual after recalculations during running request.

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| entities are, by default, not refreshed in identity map after update with native query. This PR add a manual refresh of loaded entities, so it's safe to work with those entites after recalculations (=native queries).
|New feature| No 
|Fixes issues| closes #2108 

### DIFF
--- a/packages/framework/src/Model/Product/ProductVisibilityRepository.php
+++ b/packages/framework/src/Model/Product/ProductVisibilityRepository.php
@@ -15,7 +15,7 @@ use Shopsys\FrameworkBundle\Model\Product\Exception\ProductVisibilityNotFoundExc
 class ProductVisibilityRepository
 {
     /**
-     * @var \Doctrine\ORM\EntityManagerInterface
+     * @var \Shopsys\FrameworkBundle\Component\EntityExtension\EntityManagerDecorator
      */
     protected $em;
 
@@ -30,7 +30,7 @@ class ProductVisibilityRepository
     protected $pricingGroupRepository;
 
     /**
-     * @param \Doctrine\ORM\EntityManagerInterface $em
+     * @param \Shopsys\FrameworkBundle\Component\EntityExtension\EntityManagerDecorator $em
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
      * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroupRepository $pricingGroupRepository
      */
@@ -54,6 +54,10 @@ class ProductVisibilityRepository
         $this->hideMainVariantsWithoutVisibleVariants($onlyMarkedProducts);
         $this->refreshGlobalProductVisibility($onlyMarkedProducts);
         $this->markAllProductsVisibilityAsRecalculated();
+
+        // refresh entities after native query calls
+        $this->em->refreshLoadedEntitiesByClassName(Product::class);
+        $this->em->refreshLoadedEntitiesByClassName(ProductVisibility::class);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


upgrade/UPGRADE-v9.1.1-dev.md
```diff
@@ -13,3 +13,6 @@ There you can find links to upgrade notes for other versions too.
 
 - update annotations for EntityExtensionTest.php ([#2197](https://github.com/shopsys/shopsys/pull/2197))
     - see #project-base-diff to update your project
+
+- add test to check if entities are refreshed after order is completed and after recalculation ([#2202](https://github.com/shopsys/shopsys/pull/2202))
```
grossmannmartin, line 17: missing info about changed interface in ProductVisibilityRepository

packages/framework/src/Model/Product/ProductVisibilityRepository.php
```diff
@@ -330,4 +341,48 @@ protected function hideMainVariantsWithoutVisibleVariants($onlyMarkedProducts)
...
+        $identityMap = $this->em->getUnitOfWork()->getIdentityMap();
+
+        if (!array_key_exists($className, $identityMap)) {
+            return;
+        }
+
+        foreach ($identityMap[$className] as $entity) {
+            $this->em->refresh($entity);
+        }
+    }
```
TomasLudvik, line 369: I think this method should be moved to `EntityManagerDecorator` including resolving entity names

TomasLudvik (changes requested):

upgrade/UPGRADE-v9.1.1-dev.md
grossmannmartin: The interface remains the same in the end.

packages/framework/src/Model/Product/ProductVisibilityRepository.php
grossmannmartin: Cool idea, thanks!
Done.

malyMiso:
Hi,

I was asked to check if this PR resolves scenario described in #2108. I have a question about performance. Couldn't it have an impact on performance while processing large amount of data, e.g. `ProductVisibilityMidnightCronModule`, or custom cron modules e.g transport data from IS. 

However, I like this solution with provided test and it solves described problem. :+1:

TomasLudvik (approved):

sslt (approved):

grossmannmartin:
> Hi,
> 
> I was asked to check if this PR resolves scenario described in #2108. I have a question about performance. Couldn't it have an impact on performance while processing large amount of data, e.g. `ProductVisibilityMidnightCronModule`, or custom cron modules e.g importing data from IS.
> 
> However, I like this solution with provided test and it solves described problem. 👍

Thank you for looking at this!
It's true that this can have a negative impact on the performance – most notably on the database (new queries are performed).

I'm afraid it's the price for updated data.

malyMiso:
> > Hi,
> > I was asked to check if this PR resolves scenario described in #2108. I have a question about performance. Couldn't it have an impact on performance while processing large amount of data, e.g. `ProductVisibilityMidnightCronModule`, or custom cron modules e.g importing data from IS.
> > However, I like this solution with provided test and it solves described problem. +1
> 
> Thank you for looking at this!
> It's true that this can have a negative impact on the performance – most notably on the database (new queries are performed).
> 
> I'm afraid it's the price for updated data.

I understand. I'm thinking about giving an option to disable refreshing entities on purpose. Adding parameter to `refreshProductsVisibility`  could solve it (or new function ?). It would always refresh entities by default and PRG can change it explicitly

In cron modules, entities won't be refreshed
In a request, entities will be refreshed

Does it make sense ?

PR #2206, opened 2021-01-27
[framework] empty strings are stored as null if possible

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We use nullable datatype in our database schema but actually we are saving empty string on many places. This PR unifies saving those empty strings (each of them is converted to null) to prevent confusions.
|New feature| No 
|Fixes issues| closes #1735  

### DIFF
--- a/packages/framework/src/Component/String/TransformString.php
+++ b/packages/framework/src/Component/String/TransformString.php
@@ -34,6 +34,19 @@ class TransformString
     }
 
     /**
+     * @param string|null $value
+     * @return string|null
+     */
+    public static function getTrimmedStringOrNullOnEmpty(?string $value): ?string
+    {
+        if ($value === null) {
+            return null;
+        }
+
+        return static::emptyToNull(trim($value));
+    }
+
+    /**
      * @param string $string
      * @return string
      * @link http://php.vrana.cz/vytvoreni-pratelskeho-url.php
--- a/packages/framework/src/Migrations/Version20210127085903.php
+++ b/packages/framework/src/Migrations/Version20210127085903.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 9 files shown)


TomasLudvik (changes requested):
Hey, nice job, it is only missing one thing - trimming of the text, so we are sure, there are no names with white space. We have agreed on this here https://github.com/shopsys/shopsys/pull/1735#issuecomment-756656009

packages/framework/src/Component/String/TransformString.php
```diff
@@ -33,6 +33,19 @@ public static function emptyToNull($value)
         return $value === '' ? null : $value;
     }
 
+    /**
+     * @param string|null $value
+     * @return string|null
+     */
+    public static function sanitizeInput(?string $value): ?string
```
TomasLudvik, line 40: From the name it is not clear what is being sanitized - that it is empty or whitespace string

packages/framework/src/Migrations/Version20210127085903.php
```diff
@@ -0,0 +1,41 @@
...
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20210127085903 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->sql('UPDATE category_translations SET name = NULL WHERE name = \'\'');
```
TomasLudvik, line 17: You need to trim even here in SQL like:
```suggestion
        $this->sql('UPDATE category_translations SET name = NULL WHERE TRIM(name) = \'\'');
```

TomasLudvik (changes requested):

TomasLudvik (approved):

sslt (approved):

PR #2211, opened 2021-01-28
[shopsys] replaced sensiolabs/security-checker with enlightn/security-checker

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| API on security.symfony.com nearing the end of service, so this PR introduces the replacement
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -87,6 +87,7 @@
         "doctrine/migrations": "^1.6.0",
         "doctrine/persistence": "^1.3.7",
         "elasticsearch/elasticsearch": "^7.6.1",
+        "enlightn/security-checker": "^1.3",
         "fp/jsformvalidator-bundle": "^1.6.1",
         "friendsofphp/php-cs-fixer": "^2.14.0",
         "friendsofsymfony/ckeditor-bundle": "^2.1",
@@ -118,7 +119,6 @@
         "psr/log": "^1.0",
         "ramsey/uuid": "^3.8",
         "roave/better-reflection": "^3.5",
-        "sensiolabs/security-checker": "^6.0",
         "shopsys/doctrine-orm": "^2.7.2",
         "shopsys/ordered-form": "^4.0",
         "shopsys/postgres-search-bundle": "^0.2",
@@ -197,11 +197,12 @@
             "App\\Environment::checkEnvironment",
             "@auto-scripts"
         ],
+        "security-check": "security-checker security:check",
         "auto-scripts": {
             "php phing clean": "script",
             "shopsys:domains-urls:configure": "symfony-cmd",
             "ckeditor:install --clear=skip --release=full --tag=4.5.11": "symfony-cmd",
-            "security-checker security:check": "script"
+            "composer security-check": "script"
         }
     },
     "config": {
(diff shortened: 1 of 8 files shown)


project-base/composer.json
```diff
@@ -139,9 +139,9 @@
         "auto-scripts": {
             "php phing clean": "script",
             "shopsys:domains-urls:configure": "symfony-cmd",
-            "ckeditor:install --clear=skip --release=full --tag=4.5.11": "symfony-cmd",
-            "security-checker security:check": "script"
-        }
+            "ckeditor:install --clear=skip --release=full --tag=4.5.11": "symfony-cmd"
+        },
+        "security-check": "security-checker security:check"
```
s3tezsky, line 144: Should not it be part of `auto-scripts` too?

```
        "auto-scripts": {
            "php phing clean": "script",
            "shopsys:domains-urls:configure": "symfony-cmd",
            "ckeditor:install --clear=skip --release=full --tag=4.5.11": "symfony-cmd"
            "composer security-check": "script"
        },
        "security-check": "security-checker security:check"
```

composer.json
```diff
@@ -197,6 +197,7 @@
             "App\\Environment::checkEnvironment",
             "@auto-scripts"
         ],
+        "security-check": "security-checker security:check",
```
s3tezsky, line 200: not conencted to this line:

```
"auto-scripts": {
            "php phing clean": "script",
            "shopsys:domains-urls:configure": "symfony-cmd",
            "ckeditor:install --clear=skip --release=full --tag=4.5.11": "symfony-cmd",
            "security-checker security:check": "script"  ### this line bothers me
        }
```

Could we use `"composer security-check": "script"` to prevent DRY?

s3tezsky (changes requested):
Please consider the way how is the security checker called.

project-base/composer.json
grossmannmartin: in project-base it's separate intentionally as new projects very often remove the security check from auto-scripts and add it as a check on CI performed every midnight.
Because it's better to run a security check on a time basis instead of push (projects can be without active development for a while, but check for known security issues should be performed anyway) and to prevent blocking of CI server for sending many requests.

composer.json
grossmannmartin: I'm not sure. If the composer is not available globally (for any reason), this auto-script will fail then

s3tezsky: I do not understand it now. If it might not be reliable why are we using the composer script? What I am suggesting as the same way as it is called from phing.

project-base/composer.json
s3tezsky: Well, what about projects which do not have the cron set? Now their build will be passing even on security issue.

I do mind we are not only changing the scipt which is checking it, but also change the bahavior of checking - which is really hidden in it.

composer.json
grossmannmartin: I see. You're right it's used in Phing anyway. fixed in ed8fbe8c2

project-base/composer.json
grossmannmartin: changed to run only after an update (require), as we discussed in ed8fbe8

TomasLudvik (approved):

s3tezsky (approved):

PR #2218, opened 2021-01-31
[framework] CsvReader class is now deprecated

### LINKED ISSUE #2171: Should not CsvReader use abstract filesystem?
https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Component/Csv/CsvReader.php uses functions like `file_exists` or `is_readable` - but these are unusable when the files are stored in an external filesystem, e.g. S3 that we use in our cloud.

Uploading a CSV to S3 is a common use-case when migrating data from an old eshop.

Should not this component work with `FilesystemInterface` instance for the filesystem functions? :thinking:

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Class CsvReader should not be used and is deprecated in this PR.
|New feature| No 
|Fixes issues| closes #2171  

### DIFF
--- a/packages/framework/src/Component/Csv/CsvReader.php
+++ b/packages/framework/src/Component/Csv/CsvReader.php
@@ -4,6 +4,9 @@ namespace Shopsys\FrameworkBundle\Component\Csv;
 
 use Symfony\Component\Filesystem\Exception\FileNotFoundException;
 
+/**
+ * @deprecated Class is obsolete and will be removed in the next major. Use SplFileObject::fgetcsv() instead.
+ */
 class CsvReader
 {
     /**
@@ -13,6 +16,14 @@ class CsvReader
      */
     public function getRowsFromCsv($filename, $delimiter = ';')
     {
+        @trigger_error(
+            sprintf(
+                'The %s() method is deprecated and will be removed in the next major. Use SplFileObject::fgetcsv() instead.',
+                __METHOD__
+            ),
+            E_USER_DEPRECATED
+        );
+
         if (!file_exists($filename) || !is_readable($filename)) {
             throw new FileNotFoundException();
         }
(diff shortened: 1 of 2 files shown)


packages/framework/src/Component/Csv/CsvReader.php
```diff
@@ -4,6 +4,9 @@
 
 use Symfony\Component\Filesystem\Exception\FileNotFoundException;
 
+/**
+ * @deprecated Class is obsolete and will be removed in the next major
```
pk16011990, line 8: I like when `@deprecated` contains how to resolve it - eg. "use `SplFileObject::fgetcsv()` instead". The information in upgrade notes is usefull only during upgrade.

TomasLudvik: There is info in the method and also in upgrade notes.

upgrade/UPGRADE-v9.2.0-dev.md
```diff
@@ -8,3 +8,5 @@ There you can find links to upgrade notes for other versions too.
 ## Application
 - use different css classes for javascript and tests ([#2179](https://github.com/shopsys/shopsys/pull/2179))
     - see #project-base-diff to update your project
+
```
TomasLudvik, line 11:
```suggestion
```

TomasLudvik (approved):

packages/framework/src/Component/Csv/CsvReader.php
grossmannmartin: Thanks for the point. I don't see any reason why not add additional info into depreciation. 
I added the hint "use SplFileObject::fgetcsv() instead". 
Do you agree @TomasLudvik ?

sslt (approved):

PR #2232, opened 2021-02-09
[framework] ImageGenerator now saves right encoded image instead of original one

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Image in ImageGenerator was encoded, but then was still saved original image. This has been now fixed.
|New feature| No 
|Fixes issues| ... 


TomasLudvik:
Calling getEncoded() is not necessary, in both cases is returned encoded image

PR #2183, opened 2021-01-12
[coding-standards] added support for symfony5

pesektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| added support for symfony5 for coding standards package
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| No 
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #2255, opened 2021-03-03
[coding-standards] allowed installing coding-standards with symfony 5 and higher

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In #2183 was added posibility to install `coding-standard` with symfony 5.0. Unforutenatelly there is missing support for higher versions.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/coding-standards/composer.json
+++ b/packages/coding-standards/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "conflict": {
-        "symfony/dependency-injection": ">=4.4.19"
+        "symfony/dependency-injection": ">=4.4.19 <5.0.0 || 5.1.11 || >=5.2.2"
     },
     "require": {
         "php": "^7.2",
@@ -23,7 +23,7 @@
         "squizlabs/php_codesniffer": "^3.5.0",
         "symplify/package-builder": "^7.3.18",
         "symplify/easy-coding-standard": "^7.3.18",
-        "symfony/finder": "^4.4|5.0"
+        "symfony/finder": "^4.4|^5.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.0",


packages/coding-standards/composer.json
```diff
@@ -23,7 +23,7 @@
         "squizlabs/php_codesniffer": "^3.5.0",
         "symplify/package-builder": "^7.3.18",
         "symplify/easy-coding-standard": "^7.3.18",
-        "symfony/finder": "^4.4|5.0"
```
grossmannmartin, line 26: Hi, I found out during testing the package cannot be properly installed on symfony 5+ app due to conflict section in composer.json.
```
  "conflict": {
        "symfony/dependency-injection": ">=4.4.19"
    },
```

This should probably be closed interval in this package at least. 
(Try creating website skeleton of Symfony 5 project and require shopsys/coding-standards there)

grossmannmartin (changes requested):

packages/coding-standards/composer.json
s3tezsky: Actually it cannot be closed interval becasuse of `symplify/easy-coding-standard`. In `symfony/dependency-injection` was changed some private properties to constans (if i remember it right) which lead to BC break in ECS.

I see this PR is probably named wrong but we cannot do much about it now.

s3tezsky: Good point, thanks! Fixed in 24dac18835aec8d3142f7c0d4b96313d537524d5

packages/coding-standards/composer.json
```diff
@@ -11,7 +11,7 @@
         }
     ],
     "conflict": {
-        "symfony/dependency-injection": ">=4.4.19"
+        "symfony/dependency-injection": ">=4.4.19 <5.0.0 || 5.1.11 || >=5.2.2"
```
grossmannmartin, line 14: What if symfony release version 5.1.12?

s3tezsky: According to [symfony release calendar](Symfony Releases Calendar) the 5.1 version reached the end of life so I expect no further patches.

grossmannmartin (approved):

sslt (approved):

PR #2260, opened 2021-03-09
[frontend-api] TokenAuthenicator: use HEADER_AUTHORIZATION consistently

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Now changing just the value of the constant can make an consistent change in the behavior of TokenAuthenticator. For example, we wanted to change the used header to avoid conflict with HTTP authorization on a non-public deployment of our app (thx to @milanStano for the idea). Even better would be to use DIC parameters instead of both constants, but that would introduce a BC break into SSFW.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/frontend-api/src/Model/Token/TokenAuthenticator.php
+++ b/packages/frontend-api/src/Model/Token/TokenAuthenticator.php
@@ -43,7 +43,7 @@ class TokenAuthenticator extends AbstractGuardAuthenticator
     public function supports(Request $request): bool
     {
         return $request->headers->has(static::HEADER_AUTHORIZATION) && strpos(
-            $request->headers->get('Authorization'),
+            $request->headers->get(static::HEADER_AUTHORIZATION),
             static::BEARER
         ) === 0;
     }


PetrHeinz:
I might have an idea how to use DIC parameters make it without the BC break...

```php
    /** @deprecated */
    protected const HEADER_AUTHORIZATION = 'Authorization';

    /** @deprecated */
    protected const BEARER = 'Bearer ';

    /**
     * @var \Shopsys\FrontendApiBundle\Model\Token\TokenFacade
     */
    protected $tokenFacade;

    protected string $headerAuthorization = 'Authorization ';

    protected string $bearer = 'Bearer ';

    /**
     * @param \Shopsys\FrontendApiBundle\Model\Token\TokenFacade $tokenFacade
     * @param string $headerAuthorization
     * @param string $bearer
     */
    public function __construct(TokenFacade $tokenFacade, string $headerAuthorization = 'Authorization ', string $bearer = 'Bearer ')
    {
        $this->tokenFacade = $tokenFacade;
        $this->headerAuthorization = $headerAuthorization;
        $this->bearer = $bearer;
        if ($headerAuthorization === 'Authorization' && $headerAuthorization !== static::HEADER_AUTHORIZATION) {
            @trigger_error('Don't override constants, use DIC property instead. Constants will be removed.'), E_USER_DEPRECATED);
            $this->headerAuthorization = static::HEADER_AUTHORIZATION;
        }
        if ($bearer === 'Bearer' && $bearer !== static::BEARER) {
            @trigger_error('Don't override constants, use DIC property instead. Constants will be removed.'), E_USER_DEPRECATED);
            $this->bearer = static::BEARER;
        }
        // Now we can safely use properties in this class as they will use the overridden constants in users' code (if they are extending this class and changing the constants). Plus they totally have the option to use DIC parameters instead right now. And it will not create potential problems in the rest of the app, as the constants are protected.
    }
```

Does it make sense to you? Would it be preferable to the original PR implementation?

s3tezsky (approved):
Hi @PetrHeinz thanks for the PR! 🎉 

I am OK to merge the PR as is in 9.1.1, but your suggestion about using DIC should be done in minor version - besides I would like to rename the properties to be more descriptive.

grossmannmartin (approved):

sslt (approved):

PR #2266, opened 2021-03-11
[framework] only visible parameters are exported to Elasticsearch

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We use Elasticsearch only for displaying data on frontend, so exporting parameters that are hidden on frontend is not necessary.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
+++ b/packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
@@ -446,7 +446,7 @@ class ProductExportRepository
     protected function extractParameters(string $locale, Product $product): array
     {
         $parameters = [];
-        $productParameterValues = $this->parameterRepository->getProductParameterValuesByProductSortedByName(
+        $productParameterValues = $this->parameterRepository->getVisibleProductParameterValuesByProductSortedByName(
             $product,
             $locale
         );
--- a/packages/framework/src/Model/Product/Parameter/ParameterRepository.php
+++ b/packages/framework/src/Model/Product/Parameter/ParameterRepository.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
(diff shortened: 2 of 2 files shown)


grossmannmartin (changes requested):
Hi, thanks for the PR.

I'm afraid it's not a good idea to accept this proposal.

PR title and description are about exporting to Elastic and I see the change in `ProductCachedAttributesFacade`, but export to elastic does not use this class, so parameters with visibility set to false are still exported. And I think that's a good thing, because...

I believe it's necessary to export "invisible" parameters. 
The field is named incorrectly because it's not about parameter visibility, but about the possibility to filter by the parameter. The parameter should be displayed on product detail regardless of "visibility". See parameters administration page, "visible" field is labeled as "Filter by"

Proposed change brings inconsistency between DB layer and ES layer of frontend API and read-model.
If you set `visible` to `false` (any parameter) and export data to elasticsearch, then with a default ES layer, the parameter is not visible in the filter (correct), is displayed on product detail (correct), and returned with Frontend API (correct). 
If you switch to the DB layer, the parameter is not returned in Frontend API, nor displayed on product detail. 

_note: switch to db layers is possible with the following changes:_ 
```diff
 // packages/frontend-api/src/Resources/config/graphql-types/QueryDecorator.types.yaml
    product:
        type: 'Product'
-       resolve: "@=resolver('productDetail', [args['uuid'], args['urlSlug']])"
+       resolve: "@=resolver('product', [args['uuid']])"
```

```diff
 // project-base/config/services.yaml

     Shopsys\ReadModelBundle\Product\Detail\ProductDetailViewFacadeInterface:
-        class: Shopsys\ReadModelBundle\Product\Detail\ProductDetailViewElasticsearchFacade
+        class: Shopsys\ReadModelBundle\Product\Detail\ProductDetailViewFacade
```

PR #2268, opened 2021-03-12
[shopsys] improved image lazy loading

### LINKED ISSUE #1980: Invalid HTML output due to image lazyload
### What is happening
When using twig image macro with lazyload, i.e. `{{ image(product, { itemprop: 'image', alt: product.fullname }) }}` which is defined in `vendor/shopsys/framework/src/Twig/ImageExtension.php` in method `getImageHtmlByEntityName`- and when you using lazyload (attr lazy) then **src is set to empty string**, but this output is **invalid according to W3C**, src cannot be empty. For lazyload I'm solving this by using base64 data-uri image, but i cannot change this file and i cannot set src attribute in image macro also (it's overriden in getImageHtmlByEntityName method).

### Expected result
Possibly two options:

1. Add some image placeholder by default (in src - so src attribute will not be empty) - base 64 data-uri seems to be an option (make this possible to be overriden please).

1. Allow us to define src attribute in twig macro (now it's always set to empty string, when you using lazy attr) - i.e. `{{ image(product, { src: 'placeholder.svg' }) }}` - just modify the condition for lazyload in `getImageHtmlByEntityName`, thanks.

Srnka392:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Browsers with native image lazy load support now have actual image in src, for others a placeholder image is rendered until loaded with javascript. Improved lazyloading behavior.
|New feature| No
|[BC breaks]| No
|Fixes issues| closes #1980


TomasGottvald (approved):

grossmannmartin (dismissed):
Nice work! Thanks

grossmannmartin (dismissed):
Hi, thanks for the work!

I noticed a failed tests and when I wanted to fix them, I notice strange behavior. 
When I try to render an entity without an image (so no-image should be used instead),
following HTML is generated
```
<img alt="" class="image-noimage" data-src="http://127.0.0.1/public/frontend/images/noimage.png" itemprop="image" loading="lazy" src="http://127.0.0.1/public/frontend/images/placeholder.gif" title=""/>
```

Does it make sense to have a placeholder in this case? Shouldn't be no-image.png in src directly? 

Can you check the failing tests please?

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -244,4 +264,59 @@ protected function getImageHtmlByEntityName(array $attributes, $entityName, $add
...
         ]);
     }
+
+    /**
+     * @param array $attributes
+     * @return bool
+     */
+    protected function shouldUseLazyLoad(array $attributes): bool
+    {
+        // no-image should not be lazy loaded
```
TomasLudvik, line 274: Do you think that this is not clear? If so move that condition into method with name similar to this comment

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -244,4 +264,59 @@ protected function getImageHtmlByEntityName(array $attributes, $entityName, $add
...
+
+    /**
+     * @param array $attributes
+     * @return array
+     */
+    protected function extractHtmlAttributesFromAttributes(array $attributes): array
+    {
+        $htmlAttributes = $attributes;
+
+        unset($htmlAttributes['type'], $htmlAttributes['size'], $htmlAttributes['lazy']);
```
TomasLudvik, line 290: nitpick: I am thinking about declaring not html attributes in constant and then unsetting this in loop

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -244,4 +264,59 @@ protected function getImageHtmlByEntityName(array $attributes, $entityName, $add
             'imageCssClass' => $this->getImageCssClass($entityName, $attributes['type'], $attributes['size']),
         ]);
     }
+
+    /**
+     * @param array $attributes
+     * @return bool
+     */
+    protected function shouldUseLazyLoad(array $attributes): bool
```
TomasLudvik, line 272: I am not sure about this method name. What about `isLazyLoaded` or `isLazyLoadEnabled`?

project-base/src/Twig/ImageExtension.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
TomasLudvik, line 1: Why is this happening in project-base and not in framework?

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
```diff
@@ -72,7 +80,7 @@ public function testGetImageHtml(): void
             $this->getCurrentUrl()
         );
         $expected .= sprintf(
-            '    <img alt="" class="image-product" itemprop="image" data-src="%s/content-test/images/product/default/1.jpg" title="" src="" loading="lazy"/>',
+            '    <img alt="" class="image-product" itemprop="image" data-src="%s/content-test/images/product/default/1.jpg" title="" src="%1$s/content-test/images/product/default/1.jpg" loading="lazy"/>',
```
TomasLudvik, line 83: Test for browsers without native lazyload support would be nice :)

project-base/src/Twig/ImageExtension.php
TomasLudvik: This should be mentioned as deprecated right? So in next major this code would be moved to framework. Same for services.yaml

project-base/assets/js/frontend/components/lazyLoadInit.js
```diff
@@ -1,17 +1,10 @@
...
+});
 
-export function lazyLoadInit (container) {
-    $(container).find('[loading=lazy]').each(function () {
-        $(this).attr('src', $(this).data('src')).addClass('loaded');
-    });
-}
-
-(new Register()).registerCallback(lazyLoadInit);
+export default lazyload;
```
TomasLudvik, line 10: Hmm, why have you decided to go this way instead of registerCallback? I like registerCallback more because it is called on new content automatically. Do you think that it is not necessary to apply lazyload on new contents?

TomasLudvik (changes requested):
Hey, nice change! I have just few points.

project-base/src/Twig/ImageExtension.php
grossmannmartin: You're right! fixed in 73e9ce93a

I did not add trigger error deprecation as it's no way, how a user can solve the deprecation.

packages/framework/src/Twig/ImageExtension.php
grossmannmartin: renamed to `isLazyLoadEnabled` in 73e9ce93a

packages/framework/src/Twig/ImageExtension.php
grossmannmartin: Good idea, changed in 73e9ce93a

packages/framework/src/Twig/ImageExtension.php
grossmannmartin: Removed the comment, I believe it's clear enough. In 73e9ce93a

project-base/assets/js/frontend/components/lazyLoadInit.js
grossmannmartin: RegisterCallback was called after page load and all `src` elements were replaced (so it was not lazy-loaded - every image was loaded after page load). Then it was completely unnecessary to have Minilazyload library. 

Because of the way, how Minilazyload library is implemented I rewrote it to be compatible with the idea of library – new content can be updated via `lazyload.update()` call (source: https://github.com/VelociraptorCZE/MiniLazyload#update) 

by the way, according to the previous documentation, it was still necessary to call `lazyLoadInit($container);` manually on new content.

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
grossmannmartin: I agree... I added simple test in 8f6f2a1a8. Do you think it's ok that way?

project-base/tests/ReadModelBundle/Functional/Twig/ImageExtensionTest.php
```diff
@@ -90,6 +92,36 @@ public function testGetImageHtml(): void
         libxml_clear_errors();
     }
 
+    public function testGetImageHtmlWithoutNativeLazyLoad(): void
+    {
+        $productId = 1;
+        $entityName = 'product';
+        $fileExtension = 'jpg';
```
TomasLudvik, line 99: This is used more than once in this class - maybe it could be refactored to constants?

TomasLudvik: Or maybe even better - creation of ImageView into method

project-base/assets/js/frontend/components/lazyLoadInit.js
TomasLudvik: BTW: in previous documentation there were two possible ways to do it - as you say and registerNewContent. 

I am not sure, if this is working right now. because I do not know how export works but only setting that new instance in constant does not initialize that right? Or am I wrong? I see they are describing it like that in docs, but I am not sure.

The part about calling javascripts after page load makes my head spin. When should you actually call that JS? I thought at the end of page is right - so all HTML is loaded and you should apply it then. But you are probably right, that by then there are already some images downloading or downloaded. Also we should solve the situation with new content. It should be automatic and if registerNewContent is not sufficient, we should find some better way.

grossmannmartin: You're right, it was not working for ajax requests without extra `update` call (for example `Load more` button).
Thanks for your help debugging it and how to make it work seamlessly!

TomasLudvik (approved):

sslt (approved):

PR #2270, opened 2021-03-15
deprecated MultidomainMigrationTrait.php

s3tezsky:
| Q             | A
| ------------- | ---
|New feature| Yes/No 
|Fixes issues| ... 


packages/framework/src/Migrations/AbstractMigration.php
```diff
@@ -0,0 +1,72 @@
...
+            }
+        );
+
+        return array_shift($domainConfigByDomainId)[DomainsConfigDefinition::CONFIG_LOCALE];
+    }
+
+    /**
+     * @return array
+     */
+    protected function getParsedDomainConfigs(): array
```
TomasLudvik, line 67: When you have container loaded here - would not be better to use Domain class instead of parsing yaml files?

packages/framework/src/Migrations/AbstractMigration.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return array_map(function ($domainConfig) {
+            return $domainConfig[DomainsConfigDefinition::CONFIG_ID];
+        }, $this->getParsedDomainConfigs());
+    }
+
+    /**
+     * @return array
+     */
+    protected function getCreatedDomainIds(): array
```
TomasLudvik, line 42: I would name it something like `getDomainIdsFromSettingValues` to be more describing.

packages/framework/src/Migrations/AbstractMigration.php
```diff
@@ -0,0 +1,72 @@
+<?php
```
TomasLudvik, line 1: I would update commit message to something like:
`replaced MultidomainMigrationTrait trait by AbstractMigration class`

TomasLudvik: Also those two commits should be probably squashed as now the first commit violates YAGNI and use is added in second one.

packages/framework/src/Migrations/AbstractMigration.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function setContainer(?ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * @return array
+     */
+    protected function getAllDomainIds(): array
```
TomasLudvik, line 32: This method violates YAGNI now - it is not used. I thought it will be only method for retrieving domain ids here.

TomasLudvik (changes requested):

s3tezsky:
Thanks @TomasLudvik for a review but I am closing this PR after discussion we had.

As we aggred this change will not introduce such a good change so we rather describe current behavior better.

Created #2276 and closing this one.

PR #2271, opened 2021-03-15
[framework] fix homepage slider items not ordered by position

CarabineSK:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Slider items on homepage can be ordered in administration, that change `position` attribute, but when displaying slider items on homepage, this position information is lost and the items are not ordered by it.
|New feature| No 
|Fixes issues| No 

### DIFF
--- a/packages/framework/src/Controller/Admin/SliderController.php
+++ b/packages/framework/src/Controller/Admin/SliderController.php
@@ -78,7 +78,8 @@ class SliderController extends AdminBaseController
             ->from(SliderItem::class, 's')
             ->where('s.domainId = :selectedDomainId')
             ->setParameter('selectedDomainId', $this->adminDomainTabsFacade->getSelectedDomainId())
-            ->orderBy('s.position');
+            ->orderBy('s.position')
+            ->addOrderBy('s.id');
         $dataSource = new QueryBuilderDataSource($queryBuilder, 's.id');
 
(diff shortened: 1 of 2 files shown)


CarabineSK:
This is not yet correct solution of the problem. When you create new SliderItem, it will appear first in administration, but will be last on homepage.

s3tezsky:
Thanks for the PR! 🎉 

We definitely should fix sorting results for slider items.

Besides we should investigate all entities which use `position` to fix the nullable state in this column - IMO it should be automatically recalculated.

packages/framework/src/Controller/Admin/SliderController.php
```diff
@@ -78,7 +78,8 @@ public function listAction()
             ->from(SliderItem::class, 's')
             ->where('s.domainId = :selectedDomainId')
             ->setParameter('selectedDomainId', $this->adminDomainTabsFacade->getSelectedDomainId())
-            ->orderBy('s.position');
+            ->orderBy('s.position')
+            ->addOrderBy('s.id');
```
grossmannmartin, line 82: is it necessary to order by id as a second ordering? 
this will take place only when several new slider items are created without reordering (position is then null). 

Wouldn't be better to recalculate position after creating a new item? 
(right now items with null in position behave differently in administration and on frontend)

grossmannmartin (approved):
nice 👍

sslt (approved):

PR #2274, opened 2021-03-15
[shopsys] replaced <i> tag with twig icon function

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Icon extension adds ability to change icon representation in one place. It brings possibility to add new icon types or even combine them.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
+++ b/packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
@@ -1,0 +1,8 @@
+{% if type == 'svg-font' %}
+    <i class="svg svg-{{ name }}{% if attr.class is not null %} {{ attr.class|trim }}{% endif %}"
+        {% if attr.title is not null %}title="{{ attr.title }}"{% endif %}
+        {% for key, value in attr.data %}
+            data-{{ key }}="{{ value }}"
+        {% endfor %}
+        ></i>
+{% endif %}
--- a/packages/framework/src/Twig/IconExtension.php
+++ b/packages/framework/src/Twig/IconExtension.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Twig;
+
+use Twig\Environment;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+class IconExtension extends AbstractExtension
+{
+    /**
+     * @var \Twig\Environment
+     */
+    protected $twigEnvironment;
+
+    /**
+     * @param \Twig\Environment $twigEnvironment
+     */
+    public function __construct(
+        Environment $twigEnvironment
+    ) {
+        $this->twigEnvironment = $twigEnvironment;
+    }
+
+    /**
+     * @return \Twig\TwigFunction[]
+     */
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('icon', [$this, 'renderIcon'], ['is_safe' => ['html']]),
+        ];
+    }
+
+    /**
+     * @param string $iconName
+     * @param array $attributes
+     * @param string|null $iconType
+     * @return string
+     */
+    public function renderIcon(string $iconName = '', array $attributes = [], ?string $iconType = 'svg-font'): string
+    {
+        $attributes['title'] = $attributes['title'] ?? null;
+        $attributes['class'] = $attributes['class'] ?? null;
+        $attributes['data'] = $attributes['data'] ?? [];
+
+        return $this->twigEnvironment->render(
+            '@ShopsysFramework/Common/Inline/Icon/icon.html.twig',
+            [
+                'name' => $iconName,
+                'attr' => $attributes,
+                'type' => $iconType,
+            ]
+        );
+    }
+}
--- a/project-base/templates/Front/Content/Category/panel.html.twig
+++ b/project-base/templates/Front/Content/Category/panel.html.twig
@@ -15,7 +15,10 @@
                 <a href="{{ url('front_product_list', { id: categoryWithLazyLoadedVisibleChildren.category.id }) }}" class="list-menu__item__link list-menu__item__link--level-{{ categoryWithLazyLoadedVisibleChildren.category.level }} {% if isCurrentCategory %}current{% endif %}">
                     {{ categoryWithLazyLoadedVisibleChildren.category.name }}
                     {% if categoryWithLazyLoadedVisibleChildren.hasChildren %}
-                        <i class="list-menu__item__control svg svg-arrow js-category-collapse-control" data-url="{{ url('front_category_branch', { parentCategoryId: categoryWithLazyLoadedVisibleChildren.category.id }) }}"></i>
+                        {{ icon('arrow', {
+                            class: 'list-menu__item__control js-category-collapse-control',
+                            data: { url: url('front_category_branch', { parentCategoryId: categoryWithLazyLoadedVisibleChildren.category.id }) }
+                        }) }}
                     {% endif %}
                 </a>
                 {% if categoryWithLazyLoadedVisibleChildren.category in openCategories %}
--- a/docs/frontend/icon-function.md
+++ b/docs/frontend/icon-function.md
@@ -1,0 +1,29 @@
+# Icon Function
+Icon twig function allows you to easily change icons html throughout the application.
+It is default for svg font icons.
+It also allows you to change or add new types without re-writing whole frontend templates.
+
+## Usage
+
+```twig
+{{ icon('warning') }}
+```
+
+## Params
+* name - first param is name of icon
+* class - optional - adds html attribute class with string inside (you can see above)
+* type - optional - is used type (svg-font (default), image, svg-sprite, svg-inline)
+
+## How to define new type
+Override template from framework `src/Resources/Common/Inline/Icon/icon.html.twig` and copy or edit first condition with all content.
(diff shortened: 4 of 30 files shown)


Srnka392 (approved):
Nice work, everything looks good.

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
```diff
@@ -0,0 +1,8 @@
+{% if type == 'svg-font' %}
+    <i class="svg svg-{{ name }} {{ attr.class|default('')}}"
```
s3tezsky, line 2: What about to create some default attributes right in `renderIcon` method when it will be merged with given arguments to prevent using `|default('')` everywhere?

s3tezsky, line 2: I am wondering why the `attr.class` is string. IMHO the list will do much better and will be more intuitive to use.

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
```diff
@@ -0,0 +1,8 @@
...
+        {% if attr.title|default('') %}title ="{{ attr.title }}"{% endif %}
+        {% for key, value in attr.data|default('') %}
```
s3tezsky, line 4: I would definitely set default values in PHP method. Iterating on empty string seems really weird. I am amazed it is working withou throwing any exception 😲 .

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
```diff
@@ -0,0 +1,8 @@
...
+            data-{{ key }} ="{{ value }}"
```
s3tezsky, line 5: I belive the space is not intentional here.

```suggestion
            data-{{ key }}="{{ value }}"
```

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
s3tezsky, line 2: Could we get rid of the space between `svg-{{ name }}` and `{{ attr.class|default('')}}"`. I am not sure if it has any impact (e.g. w3c standards), but is weird for my eyes.
When we rework classes to be list we could solve it by using `|join()` 😉.

packages/framework/src/Twig/IconExtension.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param string|null $iconType
+     * @return string
+     */
+    public function renderIcon(string $iconName = '', array $attributes = [], ?string $iconType = 'svg-font'): string
+    {
+        return $this->twigEnvironment->render(
+            '@ShopsysFramework/Common/Inline/Icon/icon.html.twig',
+            [
+                'name' => $iconName,
+                'attr' => $attributes,
```
s3tezsky, line 49: Personally I do not really like attr shortcut. Besides to be clear it should be named as `attrs` - what about not using shortcuts and name it `attributes`? 🤔

project-base/templates/Front/Content/Category/panel.html.twig
```diff
@@ -15,7 +15,7 @@
                 <a href="{{ url('front_product_list', { id: categoryWithLazyLoadedVisibleChildren.category.id }) }}" class="list-menu__item__link list-menu__item__link--level-{{ categoryWithLazyLoadedVisibleChildren.category.level }} {% if isCurrentCategory %}current{% endif %}">
                     {{ categoryWithLazyLoadedVisibleChildren.category.name }}
                     {% if categoryWithLazyLoadedVisibleChildren.hasChildren %}
-                        <i class="list-menu__item__control svg svg-arrow js-category-collapse-control" data-url="{{ url('front_category_branch', { parentCategoryId: categoryWithLazyLoadedVisibleChildren.category.id }) }}"></i>
+                        {{ icon('arrow', { class: 'list-menu__item__control js-category-collapse-control', data: { url: url('front_category_branch', { parentCategoryId: categoryWithLazyLoadedVisibleChildren.category.id }) } } ) }}
```
s3tezsky, line 18: This line is hard to read. Cannot we try to apply any better formating of it - it would be lovely? (it applies also to others ocurences)

s3tezsky (changes requested):
Nice! I like the way how it has changed.

Anyway I found some small issues which should not be missed to make it even better. Please check it out.

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
TomasLudvik: Because class is one of attributes and all attributes are string. Even in forms you set attributes as strings.

TomasLudvik: Well you need space between those strings, otherwise they would be joined.

packages/framework/src/Twig/IconExtension.php
TomasLudvik: It is consistent with naming attributes in Twig/Symfony so I would keep it.

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
TomasLudvik: changed

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
TomasLudvik: changed

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
TomasLudvik: fixed

project-base/templates/Front/Content/Category/panel.html.twig
TomasLudvik: I have updated long lines

packages/framework/src/Twig/IconExtension.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @param string $iconName
+     * @param array $attributes
+     * @param string|null $iconType
+     * @return string
+     */
+    public function renderIcon(string $iconName = '', array $attributes = [], ?string $iconType = 'svg-font'): string
+    {
+        $attributes['title'] = $attributes['title'] ?? '';
```
s3tezsky, line 45: Why use empty string as default instead of null?

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
s3tezsky: I am missing a default value for `attr.data`

packages/framework/src/Twig/IconExtension.php
TomasLudvik: updated to null

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
```diff
@@ -0,0 +1,8 @@
+{% if type == 'svg-font' %}
+    <i class="svg svg-{{ name }} {{ attr.class}}"
+        {% if attr.title is not same as('') %}title ="{{ attr.title }}"{% endif %}
+        {% for key, value in attr.data %}
+            data-{{ key }}="{{ value }}"
```
s3tezsky, line 5: If we want to be so consistent with symfony forms we should remove the `data-` prefix out of here and let it to be set via function. What do you think?

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
TomasLudvik: added

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
s3tezsky: Actually it is not.

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
s3tezsky: Cool!

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
s3tezsky: I udestand that. Anyway you do not need it when there is no `attr.class` specified. Then it will result in:

```
 <i class="svg svg-doge "></i>
```

packages/framework/src/Twig/IconExtension.php
s3tezsky: Well, if it is so, it should behave more like attributes in twig. See https://github.com/shopsys/shopsys/pull/2274#discussion_r595820933

packages/framework/src/Resources/views/Common/Inline/Icon/icon.html.twig
TomasLudvik: Sorry I do not follow, can you elaborate on that?

project-base/templates/Front/Content/Category/panel.html.twig
s3tezsky: It is much better! Thanks 🎉

s3tezsky (approved):

sslt (approved):

PR #2277, opened 2021-03-17
[framework] Add commonly used methods to Price class

PetrHeinz:
- `multiply`, `isZero` and `equals` are safe to use in `Price` and are commonly needed
- comparing `Price` objects is unsafe as there are two independent comparable parts in each object (same goes to `isPositive`/`isNegative`)
- dividing `Price` objects is unsafe as there might be unforeseen rounding issues
- rounding `Price` objects would need providing two `$scale` arguments and it's not needed often in my experience

| Q             | A
| ------------- | ---
|Description, reason for the PR| Commonly needed functions in Price class are missing and we are forced to manipulate with Prices in a rather awkward ways...
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Pricing/Price.php
+++ b/packages/framework/src/Model/Pricing/Price.php
@@ -89,13 +89,40 @@ class Price
     }
 
     /**
+     * @param int|string $multiplier
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\Price
+     */
+    public function multiply($multiplier): self
+    {
+        return new self(
+            $this->priceWithoutVat->multiply($multiplier),
+            $this->priceWithVat->multiply($multiplier)
+        );
+    }
+
+    /**
      * @return \Shopsys\FrameworkBundle\Model\Pricing\Price
      */
     public function inverse(): self
     {
-        return new self(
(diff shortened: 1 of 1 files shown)


PetrHeinz:
As a simple example I can provide these two snippets:

```php
if (!$singleOrderPreviewData->getTotalPrice()->getPriceWithoutVat()->equals($orderPreview->getTotalPrice()->getPriceWithoutVat()) ||
    !$singleOrderPreviewData->getTotalPrice()->getPriceWithVat()->equals($orderPreview->getTotalPrice()->getPriceWithoutVat()) ||
    !$singleOrderPreviewData->getTotalPrice()->getVatAmount()->equals($orderPreview->getTotalPrice()->getVatAmount())
) {
    throw new TotalPriceMismatchException($singleOrderPreviewData, $orderPreview);
}
```

```php
if (!$singleOrderPreviewData->getTotalPrice()->equals($orderPreview->getTotalPrice())) {
    throw new TotalPriceMismatchException($singleOrderPreviewData, $orderPreview);
}
```

The first snippet is pretty error-prone as well. Did you spot the mistake when you read it? :slightly_smiling_face:

PR #2279, opened 2021-03-23
[frontend-api] FEAPI authentication is configurable by parameters now

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Configuring authentication of FEAPI was possible only by overrigind defined constants. This PR allows you to configure it by DIC parameters. See https://github.com/shopsys/shopsys/pull/2260#issuecomment-793605731
|New feature| Yes 
|Fixes issues| ... 

There is one issue to deal with:
- Personally I do not really like the extra space after authentication type (`'Bearer '`) - am I the only one? Does it make sence to remove it and implement it without it?


sslt:
We discussed this as team and decided this is not the correct way to deal with the problem, so we are closing this PR as we will not implement it.

PR #2308, opened 2021-05-05
[framework] Enhancement of check / fix method annotations

### LINKED ISSUE #2273: MethodAnnotationsFactory::getMethodParameterNamesWithTypes() cannot generate default values
## Describe the bug

Phing target `annotations-fix` cannot generate @method annotations with default values. This can be considered an edge case, but it may produce warnings if you override a particular class (see below).

### To Reproduce

1. Override `UploadedFileDataFactory` service and `UploadedFileData` data object.
2. Run phing target `annotations-fix`.
3. See `@method \App\Component\UploadedFile\UploadedFileData createByEntity(object $entity, string $type)` appear in your new class, even though the `string $type` is optional.
4. Call the method without the second parameter elsewhere in your project.
5. Run phing target `phpstan`
6. Build fails.

### Workaround
Add a ignore into your `phpstan.neon` and ignore warnings in your IDE:
```neon
        -
            # Annotations check is unable to comprehend @method annotations with optional parameter.
            message: '#^Method App\\Component\\UploadedFile\\UploadedFileDataFactory::createByEntity\(\) invoked with 1 parameter, 2 required\.$#'
            path: *
```

Or override the method in question without really doing anything:
```php
    /**
     * @param object $entity
     * @param string $type
     * @return \App\Component\UploadedFile\UploadedFileData
     */
    public function createByEntity(
        object $entity, 
        string $type = UploadedFileTypeConfig::DEFAULT_TYPE_NAME
    ): BaseUploadedFileData {
        return parent::createByEntity($entity, $type);
    }
```

### Expected behavior

Default parameter values should be marked as such in `@method` annotations.

This can be achieved with such modification of the `MethodAnnotationsFactory::getMethodParameterNamesWithTypes()`:
```php
    /**
     * @param \Roave\BetterReflection\Reflection\ReflectionMethod $reflectionMethod
     * @return string
     */
    protected function getMethodParameterNamesWithTypes(ReflectionMethod $reflectionMethod): string
    {
        $methodParameterNamesWithTypes = [];
        foreach ($reflectionMethod->getParameters() as $methodParameter) {
            $methodParameterNamesWithTypes[] = sprintf(
                '%s $%s%s',
                $this->annotationsReplacer->replaceInParameterType($methodParameter),
                $methodParameter->getName(),
                $methodParameter->isDefaultValueAvailable() ? '=' . json_encode($methodParameter->getDefaultValue()) : ''
            );
        }

        return implode(', ', $methodParameterNamesWithTypes);
    }
```

But this will duplicate the `@method` annotations after running `annotations-fix`, eg.:
```diff
  * @method __construct(\App\Component\UploadedFile\UploadedFileFacade $uploadedFileFacade)
  * @method \App\Component\UploadedFile\UploadedFileData create()
  * @method \App\Component\UploadedFile\UploadedFileData createByEntity(object $entity, string $type)
+ * @method \App\Component\UploadedFile\UploadedFileData createByEntity(object $entity, string $type="default")
  */
 class UploadedFileDataFactory extends BaseUploadedFileDataFactory
 {
```

Seems that the command cannot parse the annotations correctly so it appends the annotation instead of replacing it. That needs fixing as well.

### Additional context

If you can get the fix into next patch releases, I would find the time to make a PR with a fix. I would add test cases to cover this new functionality.

This issue is probably there since the `annotation-fix` phing target was introduced.

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixes phing targets `annotations-fix` and `annotations-check` for cases when parameters have default values and cases when there already was a `@method` annotation but it needs an update
|New feature| No 
|Fixes issues| fixes #2273

In essence, after extending `\Framework\ClassA` with `\App\ClassA`, `phing annotaitons-fix` now does:

```diff
  /*
-  * @method existingMethod(\Framework\ClassA $objectA, \App\ClassB $objectB)
+  * @method existingMethod(\App\ClassA $object, \App\ClassB $objectB)
+  * @method newMethodWithDefaultValue(\App\ClassA $object = null)
   */
  class MyClass {
```

instead of 

```diff
  /*
   * @method existingMethod(\Framework\ClassA $object, \App\ClassB $objectB)
+  * @method newMethodWithDefaultValue(\App\ClassA $object)
+  * @method existingMethod(\App\ClassA $object)
   */
  class MyClass {
```

Which is great for static analysis and for longer work on projects where methods with many parameters have duplicated annotations (as you iteratively extend different parameters' classes).

### DIFF
--- a/packages/framework/src/Component/ClassExtension/AnnotationsAdder.php
+++ b/packages/framework/src/Component/ClassExtension/AnnotationsAdder.php
@@ -53,14 +53,45 @@ class AnnotationsAdder
     }
 
     /**
-     * @param string $docBlock
-     * @param string $annotationLinesBlock
+     * Appends second annotation block, annotation lines with colliding "name" will get replaced instead
+     *
+     * @see extractPropertyOrMethodAnnotationName() for explanation of how the "name" works
+     * @param string $annotation
+     * @param string $annotationToAdd
      * @return string
      */
-    private function replaceAnnotationsInExistingDocBlock(string $docBlock, string $annotationLinesBlock): string
+    protected function replaceAnnotationsInExistingDocBlock(string $annotation, string $annotationToAdd): string
     {
-        $annotationLinesToAdd = array_filter(explode("\n", $annotationLinesBlock));
+        $annotationLinesByName = [];
 
-        return str_replace(' */', implode("\n", $annotationLinesToAdd) . "\n */", $docBlock);
+        $annotationLines = explode("\n", $annotation);
+        $annotationStart = array_shift($annotationLines);
+        $annotationEnd = array_pop($annotationLines);
+        foreach ($annotationLines as $annotationLine) {
+            $annotationLinesByName[$this->extractPropertyOrMethodAnnotationName($annotationLine)] = $annotationLine;
+        }
+        $annotationLinesToAdd = array_filter(explode("\n", $annotationToAdd));
+        foreach ($annotationLinesToAdd as $annotationLine) {
+            $annotationLinesByName[$this->extractPropertyOrMethodAnnotationName($annotationLine)] = $annotationLine;
+        }
+
+        return implode("\n", [$annotationStart, ...array_values($annotationLinesByName), $annotationEnd]);
+    }
+
+    /**
+     * For property or method annotations returns just their name, eg. "method-setName" or "property-annotationsAdder"
+     * Otherwise it will return the whole annotation line, eg " * AnnotationsAdder constructor"
+     *
+     * @see \Tests\FrameworkBundle\Unit\Component\ClassExtension\AnnotationsAdderTest::testExtractPropertyOrMethodAnnotationName()
+     * @param string $annotationLine
+     * @return string
+     */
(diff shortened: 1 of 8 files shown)


PetrHeinz:
After implementation, I've testing this version of `annotations-fix` on a rather large project with FE API. There was no significant performance decrease (both versions ran for ~49 seconds) and there were zero unintended changes.

Seems to work OK in the real world as well as in tests :tada: 

---

I've provided failing tests for the functionality I intended to fix. These are passing now.

I've provided a unit test with a sizable data provider for the regular expression used in the implementation.

There is no BC break according to [the backward compatibility promise](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/) and no tests were changed in the process.

RobinDvorak:
Hello @PetrHeinz,
this won't get merged, because there won't be any new releases in 9.1. I cherry picked your commits and created https://github.com/shopsys/shopsys/pull/2434 so your changes will get to master and hopefully to new Major version.

grossmannmartin:
closed with #2434

PR #2309, opened 2021-05-07
Service alias misconfiguration detector

PetrHeinz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| You can easily forget to declare service aliases on your project. This PR adds an automated detector that should help you in such cases. It will even recommend a service config update that should fix the issues in most cases.
|New feature| Yes 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| No  (The check is optional, depending on configuration, and it's disabled by default)
|Fixes issues| ... 

This approach could be used to generate service configuration files and even configure the aliases during container compilation in the long run. That would completely eliminate the need for defining aliases manually if classes are extended by the docs (Convention over configuration approach).

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project name="Shopsys Framework" default="list">
 
+    <property name="check-alias-configuration" value="true"/>
     <property name="check-and-fix-annotations" value="true"/>
     <property name="npm.framework" value="@shopsys/framework"/>
     <property name="path.root" value="${project.basedir}/project-base"/>
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project name="shopsys_framework" default="list">
 
+    <property name="check-alias-configuration" value="false"/>
     <property name="check-and-fix-annotations" value="false"/>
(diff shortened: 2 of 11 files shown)


PetrHeinz:
This check proved useful during a project implementation.

More work would be needed to cover it with automated tests - specifically extracting the actual logic from `DetectMissingServiceAliasesCommand` (i.e. `getAppServicesWithParentsAndInterfaces()`, `getMissingDirectAliases()`, `getPossibleMissingAliasedServices()`, `filterUnambiguousPossibleMissingAliasedServices()`) and leaving only conversion to human-readable output along with the recommendations in the command.

This extracted class should be then covered by unit tests with manually created (or mocked) `ServiceAliasContainerInfoRegistry`.

Also, the compiler pass should be covered by tests similarly as other compiler passes in Symfony are (eg. https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/DependencyInjection/Tests/Compiler/RemoveUnusedDefinitionsPassTest.php).

PetrHeinz:
I guess the `ServiceAliasContainerInfoRegistry` could be simplified by using constructor for passing the variables (this would require changing the compiler pass as well) and declaring the registry service as lazy (to avoid issues with autowiring when not in `dev` environment).

PetrHeinz:
Let's close this one 🙌

PR #2313, opened 2021-05-12
[shopsys] bumped version of composer dependency composer/composer

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| due to reported security issue GHSA-h5h8-pc6h-jvvx
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
         "arvenil/ninja-mutex": "^0.4.1",
         "barryvdh/elfinder-flysystem-driver": "^0.2",
         "commerceguys/intl": "^1.0.0",
-        "composer/composer": "^1.6.0",
+        "composer/composer": "^1.10.22",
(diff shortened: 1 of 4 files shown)


grossmannmartin (commented):
I looked at this and I don't think we can remove this dependency. 

- it does not solve anything as composer/composer is still installed as a transient dependency
- composer is used in \App\Environment class

s3tezsky:
> I looked at this and I don't think we can remove this dependency.
> 
>     * it does not solve anything as composer/composer is still installed as a transient dependency
> 
>     * composer is used in \App\Environment class

@grossmannmartin  Thanks for great catch! I have force pushed a new solution.

grossmannmartin (approved):

sslt (approved):

PR #2325, opened 2021-05-20
[shopsys] fix calculation of transport and payment visibilities

### LINKED ISSUE #2302: FE API  return hidden transports and payments
## Describe the bug
If call FE query for transports with payments:
```
{
    query: transports {
        uuid
        payments {
            uuid
            name
        }
    }
}
```

return all payments where assigned to transport, regardless of whether they are visible for domain or hidden.

This problem is in payments query for transport:

```
{
    query: payments {
        uuid
        name
        transports {
            uuid
            name
        }
    }
}
```
return all transports where assigned to payment, regardless of whether they are visible for domain or hidden.

### Expected behavior
Payments and transport return only not hidden data, which can use in create order

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Attribute deleted of transport and payment was not used in consideration during calculation of their visibilities. Also transport and payment frontend API resolvers were not taking visibility in consideration.
|New feature| No 
|Fixes issues| closes #2302 

### DIFF
--- a/packages/framework/src/Model/Payment/PaymentRepository.php
+++ b/packages/framework/src/Model/Payment/PaymentRepository.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Model\Payment;
 
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Query\Expr\Join;
 use Shopsys\FrameworkBundle\Model\Payment\Exception\PaymentNotFoundException;
 use Shopsys\FrameworkBundle\Model\Transport\Transport;
 
@@ -106,6 +107,31 @@ class PaymentRepository
     public function getOneByUuid(string $uuid): Payment
     {
         $payment = $this->getPaymentRepository()->findOneBy(['uuid' => $uuid]);
+
+        if ($payment === null) {
+            throw new PaymentNotFoundException('Payment with UUID ' . $uuid . ' does not exist.');
+        }
+
+        return $payment;
+    }
+
+    /**
+     * @param string $uuid
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Payment\Payment
+     */
+    public function getEnabledOnDomainByUuid(string $uuid, int $domainId): Payment
+    {
+        $queryBuilder = $this->getPaymentRepository()->createQueryBuilder('p')
+            ->join(PaymentDomain::class, 'pd', Join::WITH, 'p.id = pd.payment AND pd.domainId = :domainId')
+            ->setParameter('domainId', $domainId)
+            ->where('p.uuid = :uuid')
+            ->setParameter('uuid', $uuid)
+            ->andWhere('p.deleted = false')
+            ->andWhere('pd.enabled = true')
+            ->andWhere('p.hidden = false');
+
+        $payment = $queryBuilder->getQuery()->getOneOrNullResult();
 
         if ($payment === null) {
             throw new PaymentNotFoundException('Payment with UUID ' . $uuid . ' does not exist.');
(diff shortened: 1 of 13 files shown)


packages/frontend-api/src/Model/Resolver/Payment/PaymentResolver.php
```diff
@@ -41,7 +41,7 @@ public function __construct(PaymentFacade $paymentFacade, Domain $domain)
     public function resolver(string $uuid): Payment
     {
         try {
-            return $this->paymentFacade->getByUuid($uuid);
+            return $this->paymentFacade->getVisibleByUuid($uuid);
```
grossmannmartin, line 44: shouldn't be `PaymentCanBeUsedValidator` updated too?

packages/frontend-api/src/Model/Resolver/Transport/TransportResolver.php
```diff
@@ -41,7 +41,7 @@ public function __construct(TransportFacade $transportFacade, Domain $domain)
     public function resolver(string $uuid): Transport
     {
         try {
-            return $this->transportFacade->getByUuid($uuid);
+            return $this->transportFacade->getVisibleByUuid($uuid);
```
grossmannmartin, line 44: Shouldn't be `TransportCanBeUsedValidator` updated too?

grossmannmartin (changes requested):
Is the base set as "master" correct? I see the milestone set to 9.1.2

packages/frontend-api/src/Model/Resolver/Payment/PaymentResolver.php
TomasLudvik: Yes, nice catch!

packages/frontend-api/src/Model/Resolver/Transport/TransportResolver.php
TomasLudvik: Yes, nice catch!

grossmannmartin:
While testing I found two scenarios that may be problematic:

- when I disable transport/payment on the domain, I still can obtain such transport/payment with graphql query
- when I try to use deleted transport/payment to create an order, I get a different validation error than on non-existing

deleted transport:
```
"validation": {
    "input": [
        {
            "message": "Please choose a valid combination of transport and payment",
            "code": "46ccd6d3-61e7-4a34-a42a-b13b92291e28"
        }
    ],
    "input.transport": [
        {
            "message": "Transport bfa48e3a-007d-4e19-97ab-5692f8cff1c5 doesn't exist",
            "code": "1f2d316b-3edd-4869-bba6-b234856a7783"
        }
    ]
}
```

non-existing:

```
"validation": {
    "input.transport": [
        {
            "message": "Transport bfa48e3a-007c-4e19-97ab-5692f8cff1c5 doesn't exist",
            "code": "1f2d316b-3edd-4869-bba6-b234856a7783"
        }
    ]
},
```

packages/framework/src/Model/Payment/PaymentRepository.php
```diff
@@ -113,4 +114,30 @@ public function getOneByUuid(string $uuid): Payment
...
+     */
+    public function getEnabledOnDomainByUuid(string $uuid, int $domainId): Payment
+    {
+        $queryBuilder = $this->getPaymentRepository()->createQueryBuilder('p')
+            ->join(PaymentDomain::class, 'pd', Join::WITH, 'p.id = pd.payment AND pd.domainId = :domainId')
+            ->setParameter('domainId', $domainId)
+            ->where('p.uuid = :uuid')
+            ->setParameter('uuid', $uuid)
+            ->andWhere('p.deleted = false')
+            ->setParameter('uuid', $uuid)
```
grossmannmartin, line 131: duplicated line :)

grossmannmartin (approved):

sslt (approved):

PR #2328, opened 2021-05-26
[framework] optimized category sorting in admin

### LINKED ISSUE #1993: Unusable CategoryFacade::editOrdering for larger data sets
### What is happening
When there are several hundreds of categories, it is quite impossible (takes too much time and consumes too much memory) to use drag&drop ordering in admin (`CategoryFacade::editOrdering`). The Gedmo/NestedSet extension is not suitable for that, so on a few projects, there is a kind of hack for solving this: 

There is a new attribute `Category::$position` that is set in the `CategoryFacade::editOrdering`:
```diff
- $this->em->flush($category);
- $this->categoryRepository->moveDown($category, CategoryRepository::MOVE_DOWN_TO_BOTTOM);
+ $category->setPosition($position++);
```

Then, the attribute is used for ordering, basically instead of `c.lft` in query builders:
```diff
- ->orderBy('c.lft')
+ ->orderBy('c.position, c.lft')
```

I do not like the fact that we need to maintain another attribute to be able to get the tree properly (that is what the `lft` and `rgt` are for), and it might be kind of confusing that lft and rgt attributes are not taking the order into account, however, to make the drag and drop usable, this "hack" is probably justified.

### Expected result
It is possible to use categories D&D in admin even for larger data sets and the existence of from the first point of view unnecessary `$position` attribute is documented (if you decide to implement it this way :slightly_smiling_face: )

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Category sorting is now much more performant on a large number of categories and it's not dependent on `max_input_vars` configuration. New command for nested set recalculation is introduced along with the helper method to calculate the nested set from the adjacency list. Documentation was added too on this topic.
|New feature| Yes 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| potentially yes 
|Fixes issues| closes #1993 

### Before

Snímek obrazovky 2021-05-26 v 17 06 30
Snímek obrazovky 2021-05-26 v 17 07 07

### After

Snímek obrazovky 2021-05-26 v 17 02 28
Snímek obrazovky 2021-05-26 v 17 02 37


sslt:
Simply outstanding :)

TomasLudvik (changes requested):

packages/framework/src/Model/Category/CategoryFacade.php
```diff
@@ -242,6 +252,31 @@ public function editOrdering($parentIdByCategoryId)
...
+            UPDATE ' . Category::class . ' c 
+            SET c.parent = :parent, c.level = :level, c.lft = :lft, c.rgt = :rgt 
+            WHERE c.id = :id
+        ');
+
+        foreach ($categoriesOrderingData as $categoryOrderingData) {
+            $query->execute([
+                'id' => (int)$categoryOrderingData['id'],
+                'parent' => $categoryOrderingData['parent_id'] ? (int)$categoryOrderingData['parent_id'] : $rootCategoryId,
+                // it's necessary to add 1 as root category is not included in source array
```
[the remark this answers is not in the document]
grossmannmartin: Well, I think it's good for the future understanding of the behavior. 
It's necessary to increase the numbers to get correct calculations and here I see the comment completely justified. 

I tried to refactor it, but the only thing I came up with is a method returning the whole modified parameter array with reeeeeeallyyyyyy long name `getParametersCalculatedWithMissingRootCategoryInSourceData` and right now I prefer a comment 😄

docs/model/how-to-sort-categories.md
```diff
@@ -0,0 +1,99 @@
...
+    ],
+    // ...
+]
+```
+
+!!! note
+    Categories with `parent_id` set to null will be at the main level, placed under hidden root category (see `CategoryFacade::getRootCategory()`).  
+    This root category must not be present in the data passed to reordering method.
+
+Usually, you don't have a complete nested set model, but it's much easier to obtain a sorted adjacency list (for example from the information system).
```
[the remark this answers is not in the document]
grossmannmartin: added in f3f1f7c92

packages/framework/src/Model/Category/CategoryFacade.php
```diff
@@ -242,6 +252,31 @@ public function editOrdering($parentIdByCategoryId)
         $this->em->flush();
     }
 
+    /**
+     * @param array<string, array<string, int|string|null>> $categoriesOrderingData
```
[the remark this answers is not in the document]
grossmannmartin: fixed in f9bceb1e1 and 7a42053b8

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,198 @@
+# Import categories
```
TomasLudvik, line 1:
```suggestion
# Import Categories
```

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,198 @@
...
+
+Category import is one of the specific imports.
+Due to the tree structure it's really easy to create ineffective script.
+In this cookbook we highlight some concepts and how to improve category import performance.
+
+## Task
+
+> _As an eshop owner I want to import categories from IS, so I don't need to edit categories on the two places._
```
TomasLudvik, line 9:
```suggestion
> _As an eshop owner I want to import categories from IS, so I don't need to edit categories in two places._
```

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,198 @@
...
+    If you're not familiar with the concepts of creating data imports, take a look at the [Basic data import cookbook](./basic-data-import.md)
+
+The main working loop is straightforward
+
+```php
+public function run(): void
+{
+    foreach ($this->getCategoryDataFromIs() as $importData) {
+        $this->importCategory($importData);
+    }
```
TomasLudvik, line 44: I am not sure, if this is right thing to do. Because:

1. You have to load all 10 000 categories, but you can load them in batch (I hope) so it would be better practice
2. If it is not possible to batch them - I would maybe start with creating some type of queue (like saving raw data in database), because what will you do, when 9 999 product will fail? 
3. I would prefer to remove importedCategory from source array to better work with memory.

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,198 @@
...
+ private array $parentIdByCategoryId;
+
+ private function importCategory(array $importData): void
+ {
+     // ...
+
+-    $this->categoryRepository->persistAsLastChild($category);
+-    $this->em->flush();
++    $parentCategoryIdOrNull = $importData['parent_uuid'] ? $this->categoryFacade->getByUuid($importData['parent_uuid'])->getId(): null;
++    $this->$parentIdByCategoryId[$category->getId()] = $parentCategoryIdOrNull;
```
TomasLudvik, line 168:
```suggestion
+    $this->parentIdByCategoryId[$category->getId()] = $parentCategoryIdOrNull;
```

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,198 @@
...
+     foreach ($this->getCategoryDataFromIs() as $importData) {
+         $this->importCategory($importData);
+     }
+     
++    $categoriesSortingData = CategoryNestedSetCalculator::calculateNestedSetFromAdjacencyList($this->parentIdByCategoryId);
++    $this->categoryFacade->reorderByNestedSetValues($categoriesSortingData);
+ }
+```
+
+> _Import finished in 7 minutes and 50 seconds_
```
TomasLudvik, line 194: We have 16 mins against 20 minutes, just want to mention

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,198 @@
...
+    //       'description_1' => 'Description of Electronics',
+    //       'description_2' => 'Description of Electronics in Czech',
+    //   ];
+}
+```
+
+Method returns data sorted exactly as they should be on the eshop, meaning category with the same parent will be first, if it comes first from the method.  
+And to make matters worse, IS can only return all 10000 categories.
+
+## First import
```
TomasLudvik, line 32: It was bit confusing for me, that I am following this guide and later I found, that there is better way. Maybe some info about it would be nice.

TomasLudvik (changes requested):

docs/cookbook/import-categories.md
grossmannmartin: Sorry, I don't follow, can you please explain it a little bit further?

packages/framework/src/Model/Category/CategoryFacade.php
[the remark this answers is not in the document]
grossmannmartin: As we decided, I have replaced the number with constant and removed the comment in 31e5a1fb1

docs/cookbook/import-categories.md
grossmannmartin: 🤔 that's right, it's not the best solution in the first paragraph, but I tried to put it in a "improve the code on the way", it's not "now delete everything and writ this", or?

grossmannmartin: I have added the info in d20c184e8

docs/cookbook/import-categories.md
grossmannmartin: Thanks for the ideas, 

1. I believe the exact implementation of the data loader is not important. I declared it as iterable, so you can easily return the generator and don't be bothered about batching here
2. related to 1., and I think it's above the scope of this cookbook
3. related to 1., I don't think it's necessary to have data as an array loaded in memory. Also, it's a good practice to remove elements from the array we currently iterate?

docs/model/how-to-sort-categories.md
```diff
@@ -0,0 +1,103 @@
...
+
+When you're adding a complete new category, you usually don't need to sort the whole tree again.  
+You can leverage the fact the `CategoryRepository` extends the `NestedTreeRepository`,
+so you have really helpful methods at your disposal to place a category to first or last place, or after a specific other category.
+
+For the full list, check out the [official documentation](https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/doc/tree.md#basic-usage-examples)
+
+## Import categories
+
+If you're interested to know the best practices about importing categories, you can take a look at the import categories cookbook.
```
vitek-rostislav, line 86: missing link :slightly_smiling_face:

vitek-rostislav: and maybe a link from the cookbook to this article would be nice as well :slightly_smiling_face:

vitek-rostislav (approved):
Hi, I have reviewed the articles, love them :heart_eyes: I have added just a few suggestions, please check them, I do not need to see it again :wink:

docs/cookbook/import-categories.md
TomasLudvik: Ok, maybe add some note about best practices :)

docs/model/how-to-sort-categories.md
grossmannmartin: fixed in fb9516b87

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+> _Import finished in 20 minutes and 10 seconds_
+
+Apart the increased import time, we also had to add two new dependencies, `CategoryRepository` and `EntityManager`.
+
+## Better way
+
+One of the reasons, why the import take so long, is the category tree structure.
+After each flush, the nested set is recalculated for the tree and this take some time.
```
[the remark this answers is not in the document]
grossmannmartin: fixed in fb9516b87

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,199 @@
...
+ }
+```
+
+> _Import finished in 20 minutes and 10 seconds_
+
+Apart the increased import time, we also had to add two new dependencies, `CategoryRepository` and `EntityManager`.
+
+## Better way
+
+One of the reasons, why the import take so long, is the category tree structure.
```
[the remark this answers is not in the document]
grossmannmartin: fixed in fb9516b87

docs/cookbook/import-categories.md
```diff
@@ -0,0 +1,199 @@
+# Import Categories
+
+Category import is one of the specific imports.
+Due to the tree structure it's really easy to create ineffective script.
+In this cookbook we highlight some concepts and how to improve category import performance.
+
+## Task
+
+> _As an eshop owner I want to import categories from IS, so I don't need to edit categories in two places._
```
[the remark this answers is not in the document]
grossmannmartin: fixed in the whole file in fb9516b87

docs/cookbook/import-categories.md
grossmannmartin: added in 401638197 :)

TomasLudvik (approved):

sslt (approved):

PR #2338, opened 2021-06-03
[framework] company name is return when is not null

malyMiso:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Company name is return only when is not null. Inspired by @RostislavKreisinger 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Customer/User/CustomerUser.php
+++ b/packages/framework/src/Model/Customer/User/CustomerUser.php
@@ -283,7 +283,7 @@ class CustomerUser implements UserInterface, TimelimitLoginInterface, Serializab
     public function getFullName()
     {
         if ($this->getCustomer()->getBillingAddress()->isCompanyCustomer()) {
-            return $this->getCustomer()->getBillingAddress()->getCompanyName();
+            return (string)$this->getCustomer()->getBillingAddress()->getCompanyName();
         }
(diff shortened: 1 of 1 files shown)


grossmannmartin:
This should not occur without an error in custom implementation. The company name is a mandatory field when a customer user is a company.
Can you please provide more info on where the current implementation causes trouble?

RostislavKreisinger:
> This should not occur without an error in custom implementation. The company name is a mandatory field when a customer user is a company.
> Can you please provide more info on where the current implementation causes trouble?

The error occurs in the constructor of the FrontendApiUser class. The reason why compamny name is NULL was not detected. My personal guess is probably migration. But I don't think that changes the fact that the getFullName function should only return a string.

grossmannmartin:
Hi, @malyMiso, @RostislavKreisinger
After the discussion, I have changed the implementation, so now it's always returned string even for the company name set to null, which I believe was the main problem.

Reasons for this change are:
- the behavior was different for company name set to null and empty string (for empty string it was also an empty string, for null a combination of first and last name
- it was confusing sometimes to be able to see a combination of first and last name for the company customer

s3tezsky (approved):

sslt (approved):

PR #2348, opened 2021-06-07
[framework] admin: fix label alignment in multidomain fields

### LINKED ISSUE #1783: Label of MultidomainType is not showing when entries are TextareaType
### What is happening

When I add `MultidomainType` field to a form in admin and I set `entry_type` to `TextareaType`, its label is hidden behind background of the first textarea's `form-line` and remaining textareas are missing left margin. 
image
```
    /**
     * @param \Symfony\Component\Form\FormBuilderInterface $builder
     * @param array $options
     */
    {
        $builder
            ->add('multidomainTextareas', MultidomainType::class, [
                'entry_type' => TextareaType::class,
                'label' => t('Label'),
                'required' => false,
            ])
            ->add('save', SubmitType::class);
    }
```
### Expected result

Label should definitely be visible. It could be located above the first textarea and then textareas themselves wouldn't have left margin. Or it could be located where it is now but not covered by white background of another element and then textareas for other domains would have to have the same margin as the first.

Srnka392:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Labels in the administration were bad align when the multidomain fields were added. This PR fixes this.
|New feature| No 
|Fixes issues| closes #1783

### DIFF
--- a/packages/framework/src/Resources/views/Admin/Form/theme.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Form/theme.html.twig
@@ -169,7 +169,7 @@
             </div>
             {{ form_errors(form) }}
         </div>
-    {% elseif multidomain is defined %}
+    {% elseif multidomain is defined and multidomain %}
         <div class="form-line__side form-line__side--multidomain">
             <div class="form-line__item">
                 <div class="input__wrap">
@@ -217,7 +217,7 @@
 {%- endblock text_row %}
 
 {% block money_row %}
-    {% if multidomain is defined %}
+    {% if multidomain is defined and multidomain %}
         <div class="form-line__side form-line__side--multidomain">
             <div class="form-line__item">
                 <div class="input__wrap">
@@ -269,8 +269,8 @@
             {% if js_container is not null %}
                 <div class="{{ js_container.container_class }}" data-type="{{ js_container.data_type }}">
             {% endif %}
-            <div class="form-line{% if multidomain is defined and isMultidomain() %} form-line--no-horizontal-gap{% endif %}">
-            {% if multidomain is defined %}
+            <div class="form-line{% if multidomain is defined and multidomain and isMultidomain() %} form-line--no-horizontal-gap{% endif %}">
+            {% if multidomain is defined and multidomain %}
                 {% if isMultidomain() %}
                     {{ form_label(form, getDomainName(domainId)) }}
                 {% endif %}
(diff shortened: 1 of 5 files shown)


packages/framework/src/Resources/views/Admin/Form/theme.html.twig
```diff
@@ -269,22 +269,32 @@
...
+                        <div class="input__wrap">
+                            {{- form_widget(form, { attr: { class: 'input--domain' }}) -}}
+                            <span class="input__domain">
+                                {{ domainIcon(domainId) }}
+                            </span>
+                        </div>
                     </div>
-                </div>
+                {% else %}
+                    <div>
```
TomasGottvald, line 292: It is necessary this div without any class?

packages/framework/src/Form/Admin/Country/CountryFormType.php
```diff
@@ -106,6 +110,61 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                 'required' => false,
                 'label' => t('Priority'),
             ])
+
+            // @todo: only for testing, remove before merge
```
TomasGottvald, line 114: You should fix it?

vitek-rostislav (approved):
Hi, you need to add 3.json and 4.json so it is possible to test it. Otherwise, looks ok to me, you can send it to test afterward, I do not need to see it again. But please, do not forget to remove the "DROP" commit before merging :smile:

packages/framework/src/Resources/views/Admin/Form/theme.html.twig
Srnka392: Fixed

packages/framework/src/Form/Admin/Country/CountryFormType.php
Srnka392: I need this for testing, after that, I will remove this code.

TomasGottvald (approved):

grossmannmartin:
Hi,
I found two problems 

In Safari expaded choice have some weird wrap
Snímek obrazovky 2021-06-08 v 14 34 33

Wysiwyg has domain icon over the toolbar
Snímek obrazovky 2021-06-08 v 14 36 09

packages/framework/src/Resources/views/Admin/Form/theme.html.twig
```diff
@@ -269,22 +269,32 @@
             {% if js_container is not null %}
                 <div class="{{ js_container.container_class }}" data-type="{{ js_container.data_type }}">
             {% endif %}
-            <div class="form-line">
+            <div class="form-line{% if multidomain is defined and isMultidomain() %} form-line--no-horizontal-gap{% endif %}">
```
vitek-rostislav, line 272: in the code above, there is a different condition (`if multidomain is defined and multidomain`), however, here, and in the other usages, there is just `if multidomain is defined and isMultidomain()` so the value of `multidomain` attribute has no impact on the condition evaluation, is that intended?

vitek-rostislav (commented):
Hi, I have one question, we can discuss it tomorrow if necessary :wink:

vitek-rostislav (approved):

sslt (approved):

PR #2350, opened 2021-06-10
Fix persist on null object in create delivery address

stanoMilan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Error on create delivery address, if $data->addressFilled is set to false
|New feature| No 
|Fixes issues| ... 

Fix error 500 on create deliveryAddress when $data->addressFilled set to false, because persists null object.

### DIFF
--- a/packages/framework/src/Model/Customer/DeliveryAddressFacade.php
+++ b/packages/framework/src/Model/Customer/DeliveryAddressFacade.php
@@ -34,11 +34,15 @@ class DeliveryAddressFacade
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddressData $deliveryAddressData
-     * @return \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddress
+     * @return \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddress|null
      */
-    public function create(DeliveryAddressData $deliveryAddressData): DeliveryAddress
+    public function createIfAddressFilled(DeliveryAddressData $deliveryAddressData): ?DeliveryAddress
     {
         $deliveryAddress = $this->deliveryAddressFactory->create($deliveryAddressData);
+
+        if ($deliveryAddress === null) {
+            return null;
+        }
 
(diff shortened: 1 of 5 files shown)


packages/framework/src/Model/Customer/DeliveryAddressFacade.php
```diff
@@ -52,14 +52,15 @@ public function edit(int $deliveryAddressId, DeliveryAddressData $deliveryAddres
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddressData $deliveryAddressData
-     * @return \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddress
+     * @return \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddress|null
      */
-    public function create(DeliveryAddressData $deliveryAddressData): DeliveryAddress
+    public function createIfAddressFilled(DeliveryAddressData $deliveryAddressData): ?DeliveryAddress
     {
         $deliveryAddress = $this->deliveryAddressFactory->create($deliveryAddressData);
```
TomasLudvik, line 59: This should be changed to early return like:
```
if (!$deliveryAddressData->addressFilled) { 
    return null;
}

TomasLudvik (changes requested):

packages/framework/src/Model/Customer/DeliveryAddressFacade.php
vitek-rostislav: hm that is debatable - you can not be sure  $this->deliveryAddressFactory->create does not return null (the internal logic of the method might change)

vitek-rostislav (approved):

packages/framework/src/Model/Customer/DeliveryAddressFacade.php
vitek-rostislav: fixed

TomasLudvik (approved):

pk16011990 (approved):

PR #2366, opened 2021-06-29
exclude monolog HTTP exception code 410

CarabineSK:
| Q             | A
| ------------- | ---
|Description, reason for the PR| `ProductNotFoundException` extends `GoneHttpException`. That means you get 410 instead of 404 when the product is hidden and you access the URL of the product. While 404 errors are not logged in monolog, 410 are still logged and error log is then full of 410 errors. I think we should not log 410 the same as we aren't logging 404.
|New feature| No 

Current:   
https://symfony.com/doc/current/logging/monolog_regex_based_excludes.html  
Proposed:   
https://symfony.com/doc/current/logging/monolog_exclude_http_codes.html

There might be some other things to adjust like file `config/packages/test/monolog.yaml`

### DIFF
--- a/project-base/config/packages/monolog.yaml
+++ b/project-base/config/packages/monolog.yaml
@@ -6,8 +6,7 @@ monolog:
             buffer_size: 1000
             action_level: warning
             handler: nested
-            excluded_404s:
-                - ^/
+            excluded_http_codes: [404, 410]
         nested:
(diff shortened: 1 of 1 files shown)


grossmannmartin:
Configuration is now different, and exclusion is already done on the another level

PR #2369, opened 2021-07-27
WIP: suppressed throwing exception when phpdoc is not valid

s3tezsky:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Annotation fixer may fail when there is extended method annotated originaly by a modern phpdoc. This PRs is hot fix which catch these kind of exceptions and transform it into warnings. There should be investigated a better solution for this problem.
|New feature| No 
|Fixes issues| ... 

Examples of unparsable annotations with threw error message (see the weird format of message):

message: `"\Shopsys\FrameworkBundle\Model\Category\array<int, array{id: string" is not a valid Fqsen.`
line: https://github.com/shopsys/shopsys/blob/ab69a2d961025b6e494d3929bcbdf7c2a079e99a/packages/framework/src/Model/Category/CategoryFacade.php#L259

---

message: `"\Shopsys\FrameworkBundle\Model\Product\Listing\array<string, string>" is not a valid Fqsen.`
line: https://github.com/shopsys/shopsys/blob/ab69a2d961025b6e494d3929bcbdf7c2a079e99a/packages/framework/src/Model/Product/Listing/ProductListOrderingModeForListFacade.php#L58

---

message: `"\Shopsys\FrameworkBundle\Controller\Admin\node_modules/@shopsys/framework/js/admin/components/CategoryTreeSorting.js" is not a valid Fqsen.`
line: https://github.com/shopsys/shopsys/blob/ab69a2d961025b6e494d3929bcbdf7c2a079e99a/packages/framework/src/Controller/Admin/CategoryController.php#L224

### DIFF
--- a/packages/framework/src/Command/ExtendedClassesAnnotationsCommand.php
+++ b/packages/framework/src/Command/ExtendedClassesAnnotationsCommand.php
@@ -131,6 +131,13 @@ class ExtendedClassesAnnotationsCommand extends Command
             }
         }
         $filesForAddingPropertyOrMethodAnnotations = $this->addPropertyAndMethodAnnotationsToProjectClasses($isDryRun);
+
+        if (count($this->methodAnnotationsAdder->warningBag) > 0) {
+            foreach ($this->methodAnnotationsAdder->warningBag as $exception) {
+                $symfonyStyle->warning($exception->getMessage());
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


vitek-rostislav:
The problems should be already resolved after upgrading libraries in https://github.com/shopsys/shopsys/pull/2430#issuecomment-1110040344.

PR #2405, opened 2022-02-17
[shopsys] changed entity extension subscriber class

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Package `joschi127/doctrine-entity-override-bundle` replaced with the custom implementation of entity extensibility.
|New feature| No 
|Fixes issues| ... 


packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
```diff
@@ -0,0 +1,306 @@
...
+    {
+        $this->setEntityExtensionMap($entityExtensionMap);
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getSubscribedEvents(): array
+    {
+        return [Events::loadClassMetadata];
```
vitek-rostislav, line 62: weird to see a constant that is not all caps :smile: 

I would like to take the opportunity here to pay homage to hip hop legend MF DOOM, aka Madvillain, RIP. https://www.youtube.com/watch?v=gSJeHDlhYls

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * @var array<class-string, class-string>
+     */
+    protected array $allParentClasses = [];
+
+    /**
+     * @var class-string[]|null
+     */
+    protected ?array $allRegisteredEntitiesCache = null;
```
vitek-rostislav, line 42: TBH, I am not a big fan of nullable arrays. I find an empty array more sexy than `null`. Is there a particular reason why it is nullable?

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
```diff
@@ -0,0 +1,306 @@
...
+        foreach ($this->entityExtensionMap as $parentEntity => $extendedEntity) {
+            Assert::classExists($parentEntity);
+            Assert::classExists($extendedEntity);
+
+            $parentClasses = class_parents($extendedEntity);
+
+            if (!array_key_exists($parentEntity, $parentClasses)) {
+                throw new LogicException(
+                    sprintf(
+                        'Invalid entity extension mapping. Entity "%s" is not child of "%s"',
```
vitek-rostislav, line 89: nitpick
```suggestion
                        'Invalid entity extension mapping. Entity "%s" is not a child of "%s"',
```

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * @var bool
+     */
+    protected bool $isEntityMapLoaded = false;
+
+    /**
+     * @var array<class-string, array<class-string, class-string>>
+     */
+    protected array $parentClassesByClass = [];
```
vitek-rostislav, line 32: while I am reviewing the code, I am thinking - is it even necessary to keep an array of all the parents? Would not it be sufficient the keep just these parents that are registered entities? Let's discuss it, please.

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
```diff
@@ -0,0 +1,306 @@
...
+     * @param class-string $parentClass
+     * @return class-string
+     */
+    protected function ensureAbsoluteClassName(string $classNameCandidate, string $parentClass): string
+    {
+        if (class_exists($classNameCandidate) || str_contains($classNameCandidate, '\\')) {
+            return $classNameCandidate;
+        }
+
+        $reflClass = new ReflectionClass($parentClass);
```
vitek-rostislav, line 291: nitpick - we should not use shortcuts, I would rather see `$reflectionClass` variable name here :innocent:

vitek-rostislav (changes requested):
Hi, simply beautiful :heart:  but definitely not simple :smile:

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
```diff
@@ -0,0 +1,306 @@
...
+        return [Events::loadClassMetadata];
+    }
+
+    /**
+     * @param array<class-string, class-string> $entityExtensionMap
+     */
+    protected function setEntityExtensionMap(array $entityExtensionMap): void
+    {
+        $this->entityExtensionMap = $entityExtensionMap;
+        $this->isEntityMapLoaded = false;
```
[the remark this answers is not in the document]
grossmannmartin: The method `setEntityExtensionMap` only sets the input data and the true load of the extension map is done in `loadEntityExtensionMap` method. 
Method `setEntityExtensionMap` is called also in tests in `OverwritableEntityExtensionMap`, and there we want to be sure that entity extension maps will be loaded again to prevent stale data, thus `false` here.

This all is done for performance reasons – the constructor is called upon each request, so it only set one property. The heavy lifting is done in method `loadClassMetadata` and this method is called only once.

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
grossmannmartin: It is weird I agree :D 

Thanks for the sharing, cool ♫

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
grossmannmartin: In this case, it distinguishes `not-loaded state` and `loaded empty state`. 

In practice, it's unexpected that there will be no registered entity, so If you find it more understandable, I can change the nullability and load registered entities on an empty array.

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
grossmannmartin: true indeed :)

packages/framework/src/Component/EntityExtension/EntityExtensionSubscriber.php
vitek-rostislav: Nah, you can keep it as is, the explanation is good enough for me, thank you :slightly_smiling_face:

PR #2413, opened 2022-03-04
[shopsys] replaced fzaninotto/Faker with FakerPHP/Faker

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| fzaninotto/Faker is abandoned, dependency is replaced with FakerPHP/Faker
|New feature| No 
|Fixes issues| ... 


project-base/symfony.lock
```diff
@@ -1181,11 +1067,5 @@
     },
     "zalas/phpunit-injector": {
         "version": "v1.2.7"
-    },
```
vitek-rostislav, line 1184: those many changes in `symfony.lock` are confusing and a little disturbing :smile: How come there are so many of them? :thinking: all the removed packages were dependencies of the original faker?

vitek-rostislav (approved):
Hi, smooth :ok_hand: I am just a little bit concerned about the changes in the `symfony.lock` files

project-base/symfony.lock
grossmannmartin: actually, I just run composer update and commit changes... I assumed it was something previously forgotten... 
I can limit changes in symfony(ies).lock, but the I would like to sync them so we don't have changes after each composer update

vitek-rostislav: ah, yeah, that could be the reason :+1: Thanks for the explanation :blush:

PR #2418, opened 2022-03-31
[migrations] add psr/log dependency

vitek-rostislav:
- fixed AbstractMigrationLockTestCase so it creates mock from LoggerInterface instead of Monolog\Logger

| Q             | A
| ------------- | ---
|Description, reason for the PR| the package itself should require all the dependencies it uses, otherwise, it might end non-funcioning, see https://github.com/shopsys/migrations/runs/5770698635?check_suite_focus=true
|New feature| No 
|Fixes issues| ... 


grossmannmartin (approved):
thanks! reviewed and tested

PR #2419, opened 2022-04-07
[frontend-api] updated lcobucci/jwt to version 4.1.5 

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| new version allows upgrading to PHP 8.1 and resolve deprecation notices
|New feature| No 
|Fixes issues| ... 


packages/frontend-api/src/Model/Token/TokenFacade.php
```diff
@@ -4,25 +4,28 @@
...
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser;
 use Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserFacade;
 use Shopsys\FrontendApiBundle\Model\Token\Exception\ExpiredTokenUserMessageException;
 use Shopsys\FrontendApiBundle\Model\Token\Exception\InvalidTokenUserMessageException;
 use Shopsys\FrontendApiBundle\Model\Token\Exception\NotVerifiedTokenUserMessageException;
 use Shopsys\FrontendApiBundle\Model\User\FrontendApiUser;
-use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 use Throwable;
+use function date_default_timezone_get;
```
vitek-rostislav, line 28: this import should be unnecessary, the function should work either way, or?

build.xml
```diff
@@ -342,6 +342,12 @@
             <arg value="${path.packages}/backend-api/phpunit.xml"/>
             <arg value="${path.packages}/backend-api/tests/"/>
         </exec>
+        <exec executable="${path.bin}/phpunit" logoutput="true" passthru="true" checkreturn="true">
```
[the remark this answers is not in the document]
vitek-rostislav: ah, now I see this change should have been probably part of the second commit, right?

packages/frontend-api/tests/Unit/Model/Token/TokenFacadeTest.php
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * @return \Shopsys\FrontendApiBundle\Model\Token\TokenFacade
+     */
+    private function createTokenFacade(): TokenFacade
+    {
+        $domain = $this->createDomain();
+
+        $customerUserFacade = $this->createMock(CustomerUserFacade::class);
+        $customerUserFacade->method('addRefreshTokenChain');
```
vitek-rostislav, line 96: is this necessary when you do not set any expected outcome from the method?

vitek-rostislav (commented):
Hi, nice work there :ok_hand: I have just a few questions :slightly_smiling_face:

build.xml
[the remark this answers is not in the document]
vitek-rostislav: I do not insist on changing the history though :wink:

packages/frontend-api/src/Model/Token/TokenFacade.php
grossmannmartin: it will work. 
The only difference is that without the import parser will look for the function in the current namespace first. With the import it will go for the standard function immediately.

packages/frontend-api/tests/Unit/Model/Token/TokenFacadeTest.php
grossmannmartin: Oh, you're right. It is not. 
Fixed in 573338ddd

packages/frontend-api/src/Model/Token/TokenFacade.php
vitek-rostislav: ah, ok, thanks for the explanation :ok_hand:

vitek-rostislav (approved):

build.xml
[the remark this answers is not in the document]
grossmannmartin: Ouch... yes it should be part of the second commit. 
I think unit tests should be run regardless of the frontend api enabling as they are independent (and run only in monorepo or split package)

vitek-rostislav (approved):

PR #2422, opened 2022-04-12
[backend-api] backend api is now put under review for future changes

grossmannmartin:
- package is now not tested with the new versions

| Q             | A
| ------------- | ---
|Description, reason for the PR| Current state of the Backend API does not comply with our vision of the product. The package has been put under review.
|New feature| No 
|Fixes issues| ... 


composer.json
```diff
@@ -121,6 +118,7 @@
         "psr/log": "^1.0",
         "ramsey/uuid": "^3.8",
         "roave/better-reflection": "^3.5",
+        "sensio/framework-extra-bundle": "^5.2",
```
vitek-rostislav, line 121: why is this package added, please?

vitek-rostislav (approved):

vitek-rostislav:
I am thinking, should not we mention this in the upgrade notes? :thinking: If anyone used the BE API, they would probably need to fork it and deal with PHP 8.1 support themselves during the upgrade to the new version of the framework.

vitek-rostislav (changes requested):
I am thinking, should not we mention this in the upgrade notes? :thinking: If anyone used the BE API, they would probably need to fork it and deal with PHP 8.1 support themselves during the upgrade to the new version of the framework.

composer.json
grossmannmartin: It's used in project-base and the dependency was in monorepo installed as a transitive dependency with now removed fos/rest-bundle

grossmannmartin:
> I am thinking, should not we mention this in the upgrade notes? 🤔 If anyone used the BE API, they would probably need to fork it and deal with PHP 8.1 support themselves during the upgrade to the new version of the framework.

Is it ok? 131e20049

PR #2426, opened 2022-04-21
[framework] Remove usage of deprecated methods from roave/better-reflection

RobinDvorak:
| Q             | A
| ------------- | ---
|Description, reason for the PR|Removes all usages of methods that are no longer available in roave/better-reflection v5.x so that we can upgrade to version 5 in the future without problems. After uppgrade of roave/better-reflection phpdocumentor/reflection-docblock must be added as new dependency to composer.json
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/tests/Unit/Component/ClassExtension/DocBlockParserTest.php
+++ b/packages/framework/tests/Unit/Component/ClassExtension/DocBlockParserTest.php
@@ -1,0 +1,156 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\FrameworkBundle\Unit\Component\ClassExtension;
+
+use phpDocumentor\Reflection\Fqsen;
+use phpDocumentor\Reflection\Types\Array_;
+use phpDocumentor\Reflection\Types\Compound;
+use phpDocumentor\Reflection\Types\Integer;
+use phpDocumentor\Reflection\Types\Null_;
+use phpDocumentor\Reflection\Types\Object_;
+use PHPUnit\Framework\TestCase;
+use Roave\BetterReflection\Reflection\ReflectionObject;
+use Roave\BetterReflection\Reflection\ReflectionParameter;
+use Roave\BetterReflection\Reflection\ReflectionProperty;
+use Shopsys\FrameworkBundle\Component\ClassExtension\DocBlockParser;
+use Tests\FrameworkBundle\Unit\Component\ClassExtension\Source\DummyClassForAnnotationsReplacerTest;
+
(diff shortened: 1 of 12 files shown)


grossmannmartin:
can you please also check the usage of `ClassReflector` in `\Shopsys\FrameworkBundle\Component\ClassExtension\ClassExtensionRegistry::getFqcn()`? 
This class is also removed in v5

RobinDvorak:
> can you please also check the usage of `ClassReflector` in `\Shopsys\FrameworkBundle\Component\ClassExtension\ClassExtensionRegistry::getFqcn()`? This class is also removed in v5

Sure! Totally mised that usage...

packages/framework/tests/Unit/Component/ClassExtension/DocBlockParserTest.php
```diff
@@ -0,0 +1,163 @@
...
+class DocBlockParserTest extends TestCase
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\ClassExtension\DocBlockParser
+     */
+    private DocBlockParser $docBlockParser;
+
+    protected function setUp(): void
+    {
+        $this->docBlockParser = new DocBlockParser();
```
vitek-rostislav, line 29: just a nitpick - I do not insist on changing this, I am ok with the current solution, however, we use https://packagist.org/packages/zalas/phpunit-injector so you could just add `@inject` annotation to the `$docBlockParser` property and you would not need the `setUp` method at all. Or do you need a new instance of `DocBlockParser` to be created before each of the tests?

RobinDvorak:
> can you please also check the usage of `ClassReflector` in `\Shopsys\FrameworkBundle\Component\ClassExtension\ClassExtensionRegistry::getFqcn()`? This class is also removed in v5

Unfortunately this is not possible in v4. From upgrade notes:
> ClassReflector, FunctionReflector and ConstantReflector have been removed. Use DefaultReflector to reflect all types.

In v5 they replaced all 3 classes with `DefaultReflector` and this class is not available in v4. So this replacement must be done at the same time as upgrading package.

packages/framework/tests/Unit/Component/ClassExtension/DocBlockParserTest.php
vitek-rostislav: Ah, I did not notice it is a unit test, and what is more important, I did not know it is not possible to use in unit tests :smile: Thanks for the info :+1:

vitek-rostislav (approved):

PR #2430, opened 2022-04-26
[shopsys] Upgrade to PHP 8.1

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| From now on, only PHP 8.1 or higher is supported by Shopsys Framework. Explanation follows.
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| May 
|Fixes issues| ... 

In general, it's not a good idea to run an application on an unsupported version of PHP.
PHP 7.4 is no longer actively supported and security support stops in a few months.

Many used libraries release new versions with PHP 8+ compatibility only, which forces us to keep older versions without important fixes or new features available.


RobinDvorak (approved):
<3

packages/form-types-bundle/.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: php
 
 php:
-    - 7.4
+    - 8.1
```
vitek-rostislav, line 4: I thought we have already removed the Travis checks and replaced them with the Github actions :thinking: :smile:

vitek-rostislav:
after local testing, it seems like all the problems mentioned as examples in https://github.com/shopsys/shopsys/pull/2369 are resolved now, i.e. I feel like we do not need to collect the warnings in `MethodAnnotationsAdder` anymore

composer.json
```diff
@@ -48,7 +48,7 @@
         }
     },
     "require": {
-        "php": "^7.4.1",
+        "php": "^8.1",
```
s3tezsky, line 51: Many developers will want to burn you alive for this change. It is really pitty there is no middle step with support of both PHP versions (`^7.4.1|^8.1`)

vitek-rostislav: Hi @s3tezsky :blush: Yeah, it would be nice to have the middle step, but you know, PHP 7.4 will end the support soon one way or another. Moreover, we would have to come up with a workaround for Roave/Better-reflection which has no middle step as well so in our codebase, we would have to require both versions and check the existence of particular classes, etc. I would say that does not worth it, the next release will be definitely a major, basically just with the upgraded PHP and related dependencies so I believe nobody will want to burn us alive :slightly_smiling_face:

s3tezsky: Burning alive was just a hyperbolic meaning 😁. Just remember how even smaller upgrades can get you mad sometimes... With this change you switch your PHP version and pray for better tomorows and painless upgrade.
I just wanted to write it is a pitty. You guys know what you are doing 😉.
So the SSFW X is comming soon?

vitek-rostislav: > Burning alive was just a hyperbolic meaning

I know I know :slightly_smiling_face: 

> So the SSFW X is comming soon?

Oh, what a cool name for the next version :grimacing: we should definitely use it somehow :smile: And yes, we have done almost all prerequisites so we should start the release process soon :wink:

composer.json
```diff
@@ -111,15 +111,15 @@
         "object-calisthenics/phpcs-calisthenics-rules": "^3.1",
         "overblog/graphiql-bundle": "^0.2",
         "overblog/graphql-bundle": "^0.13",
-        "phing/phing": "^2.16.1",
+        "phing/phing": "dev-oldstable",
```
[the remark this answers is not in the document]
grossmannmartin: added into commit message :)

packages/form-types-bundle/.travis.yml
grossmannmartin: You're right. I removed all Travis configurations and file a new user-story to use GitHub Actions in releaser

composer.json
s3tezsky: > Sadly the middle step is not achievable easily.

If it were easy, everyone do it 😁.

> All in all too much work for a PHP version that doesn't have an active support and stops receiving security updates in a few months.

I was pointing mainly on developer experience during upgrading. Some extra time of hacking could save amount of time and gray hair of them.

PS: I respect your decision. Nice job with the upgrade 👍

vitek-rostislav (approved):

Issue #2432, opened 2022-04-29
PHP Fatal error:  Uncaught ArgumentCountError


qualeo:
## Describe the bug
**Fatal Error**
```
PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function Symfony\Component\Config\Definition\Builder\TreeBuilder::__construct(), 0 passed in /var/www/html/packages/framework/src/Component/Domain/Config/DomainsConfigDefinition.php on line 22 and at least 1 expected in /var/www/html/vendor/symfony/config/Definition/Builder/TreeBuilder.php:26
```

**Stack trace**
```
Stack trace:
#0 /var/www/html/packages/framework/src/Component/Domain/Config/DomainsConfigDefinition.php(22): Symfony\Component\Config\Definition\Builder\TreeBuilder->__construct()
#1 /var/www/html/vendor/symfony/config/Definition/Processor.php(46): Shopsys\FrameworkBundle\Component\Domain\Config\DomainsConfigDefinition->getConfigTreeBuilder()
#2 /var/www/html/packages/framework/src/Component/Domain/Config/DomainsConfigLoader.php(136): Symfony\Component\Config\Definition\Processor->processConfiguration(Object(Shopsys\FrameworkBundle\Component\Domain\Config\DomainsConfigDefinition), Array)
#3 /var/www/html/packages/framework/src/Component/Domain/Config/DomainsConfigLoader.php(34): Shopsys\FrameworkBundle\Component\Domain\Config\DomainsConfigLoader->getProcessedConfig('/var/www/html/p...', Object(Shopsys\FrameworkBundle\Component\Domain\Config\DomainsConfigDefinition))
#4 /var/www/html/packages/framework/src/Component/Domain/DomainFactory.php(38): Shopsys\FrameworkBundle\Component\Domain\Config\DomainsConfigLoader->loadDomainConfigsFromYaml('/var/www/html/p...', '/var/www/html/p...')
#5 /var/www/html/project-base/var/cache/test/ContainerXxl5Wrt/srcApp_KernelTestContainer.php(1170): Shopsys\FrameworkBundle\Component\Domain\DomainFactory->create('/var/www/html/p...', '/var/www/html/p...')
#6 /var/www/html/project-base/var/cache/test/ContainerXxl5Wrt/srcApp_KernelTestContainer.php(1952): ContainerXxl5Wrt\srcApp_KernelTestContainer->getDomainService()
[...]

PR #2433, opened 2022-04-29
Updated TreeBuilder Instantiations

qualeo:
Fixes https://github.com/shopsys/shopsys/issues/2432


grossmannmartin:
Thanks for the PR. As mentioned in #2432, this is not the proper way of fixing it at the moment.

We cannot guarantee functionality with some Symfony packages upgraded to version 5. Please update symfony/flex package to version 1.18.7 to solve this problem. 

We plan to update all Symfony dependencies soon, stay tuned.

PR #2434, opened 2022-05-02
[framework] Enhancement of check / fix method annotations

RobinDvorak:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixes phing targets annotations-fix and annotations-check for cases when parameters have default values and cases when there already was a @method annotation but it needs an update. <br />https://github.com/shopsys/shopsys/pull/2308 won't be merged as v9.1 won't get any new versions, so I cherry picked commits from @PetrHeinz (Thank you!) and added correct behaviour for constants as default parameter. 
|New feature| No 
|Fixes issues| Closes https://github.com/shopsys/shopsys/issues/2273 

### DIFF
--- a/packages/framework/src/Component/ClassExtension/AnnotationsAdder.php
+++ b/packages/framework/src/Component/ClassExtension/AnnotationsAdder.php
@@ -53,14 +53,45 @@ class AnnotationsAdder
     }
 
     /**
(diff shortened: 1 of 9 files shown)


grossmannmartin (approved):
nice work, thanks 👍

PR #2435, opened 2022-05-02
[shopsys] Upgrade easy-coding-standard to v10

RobinDvorak:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Update ECS and allow parallel run
|New feature| No 
|Fixes issues| ~12 times faster check on Ryzen 5600X 

Before:
Screenshot from 2022-05-02 19-48-47

After:
Screenshot from 2022-05-02 19-36-25


packages/coding-standards/composer.json
```diff
@@ -11,22 +11,23 @@
...
-        "object-calisthenics/phpcs-calisthenics-rules": "^3.1",
-        "php-parallel-lint/php-parallel-lint": "^1.3.1",
-        "slevomat/coding-standard": "^6.3.5",
-        "squizlabs/php_codesniffer": "^3.5.0",
-        "symplify/easy-coding-standard": "^8.3.48",
         "symfony/finder": "^4.4|^5.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.5.20"
+        "friendsofphp/php-cs-fixer": "^3.3.1",
```
janatjak, line 23: nemam paru, jak funguje symplify monorepu builder, ale treba na uxf/fan jsem se na tyhle dev deps uplne vykaslal a jsou jen v root composer.json

packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
```diff
@@ -4,10 +4,10 @@
 
 namespace Shopsys\CodingStandards\Sniffs;
 
+use function array_reverse;
```
janatjak, line 7: na tohle mam takovej ugly hack, aby to bylo "kontextove u sebe"

viz. https://gitlab.com/uxf/fan/-/blob/main/ecs.php#L28

packages/framework/src/Command/TranslationReplaceSourceCommand.php
```diff
@@ -132,7 +132,9 @@ private function extractReplacementsFromPoFile(SplFileInfo $file)
         $translationSourceReplacements = [];
         $item = $defaults;
 
-        while ($line = fgets($stream)) {
+        $line = fgets($stream);
```
janatjak, line 135: takova pocitovka, ale vsechny tyhle while bych spis hodil jako:

```php
while (true) {
    $line = fgets($stream);

   if ($line === false) {
      break;
  }
}
```

ale klidne na to pec

packages/coding-standards/composer.json
RobinDvorak: In this case packages should be defined because we have custom rules.

composer.json
```diff
@@ -161,7 +158,10 @@
         "codeception/module-db": "^1.0",
         "codeception/module-webdriver": "^1.0",
         "codeception/phpunit-wrapper": "^9.0.7",
+        "composer/composer": "^2.2.12",
```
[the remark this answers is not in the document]
RobinDvorak: I was afraid to delete this :D

packages/framework/tests/Unit/Component/ClassExtension/Source/AnnotationsAdderTest/DummyClassWithAnAnnotation.php
```diff
@@ -4,6 +4,9 @@
 
 namespace Tests\FrameworkBundle\Unit\Component\ClassExtension\Source\AnnotationsAdderTest;
 
+/**
```
RobinDvorak, line 7: This is rollback from automatic fix...

composer.json
```diff
@@ -170,15 +170,16 @@
         "phpunit/phpunit": "^9.5.20",
         "psr/event-dispatcher": "^1.0.0",
         "slevomat/coding-standard": "^6.3.5",
+        "squizlabs/php_codesniffer": "^3.6.2",
```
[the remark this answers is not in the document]
grossmannmartin: Actually, classes from this package are directly used, therefore should not be left only as a transitive dependency.

composer.json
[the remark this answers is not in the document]
grossmannmartin: Classes are used in the codebase (few autoload.php files, Environment.php)

packages/backend-api/ecs.php
```diff
@@ -17,12 +17,9 @@
     $services = $containerConfigurator->services();
     $parameters = $containerConfigurator->parameters();
 
-    $parameters->set(
-        Option::SETS,
-        [
-            SetList::PSR_12,
-        ]
-    );
+    $parameters->set(Option::PARALLEL, true);
```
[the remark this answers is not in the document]
grossmannmartin: After the split, checks for each package are run independently. Common configuration should be in the `coding-standard` package, but I'm not sure how it works with the parameters

packages/coding-standards/src/Sniffs/ConstantVisibilityRequiredSniff.php
grossmannmartin: I thought we have the rule for grouping types of `use` applied...

RobinDvorak: I don't think there was any rule handling this. This is some internal change in ECS or CS Fixer new version... It's not the first time I see this change after upgrading ECS.

I'll try @janatjak suggestion and check the results. This sorting triggers me, but I am not in position to decide if it's ok or not.

grossmannmartin: I prefer to have imports sorted and grouped to `class`, `function`, `constant` parts (I believe it should not generate many changes in that case)

RobinDvorak: @janatjak We had the rule set like this:
```php
    $services->set(OrderedImportsFixer::class)
        ->call('configure', [['sort_algorithm' => 'alpha', 'imports_order' => ['class', 'function', 'const']]]);
```
But it was overrident by some set. Imports was a bit messy :D

project-base/ecs.php
```diff
@@ -143,80 +105,13 @@
...
-    // @deprecated This will be moved from project-base to coding-standards package in next major version
-    $services->set(DocCommentSpacingSniff::class)
-        ->property('linesCountBetweenDifferentAnnotationsTypes', 0);
-
-    // @deprecated This will be moved from project-base to coding-standards package in next major version
-    $services->set(UselessIfConditionWithReturnSniff::class);
-
-    $services->set(PhpdocToPropertyTypeFixer::class);
-
     $containerConfigurator->import(__DIR__ . '/vendor/shopsys/coding-standards/ecs.php', null, true);
```
RobinDvorak, line 116: Maybe this import should be at the start of the file, because it includes sets which could override rules? Same for the project-base and other packages.

RobinDvorak: What do qou think? @grossmannmartin

grossmannmartin (changes requested):

composer.json
```diff
@@ -170,15 +168,16 @@
...
         "symfony/var-dumper": "^4.4.0",
-        "symplify/easy-coding-standard": "^8.3.48",
+        "symplify/easy-coding-standard": "^10.2.2",
         "symplify/monorepo-builder": "^8.3.48",
         "zalas/phpunit-injector": "^2.0"
     },
     "conflict": {
         "symfony/symfony": "*",
-        "friendsofphp/php-cs-fixer": "2.19.0",
+        "friendsofphp/php-cs-fixer": ">=3.3.2",
```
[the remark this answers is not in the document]
RobinDvorak: Sure, forgot about this :D `friendsofphp/php-cs-fixer@3.3.1` is the last version supporting Symfony 4. As you can see here: https://packagist.org/packages/friendsofphp/php-cs-fixer#v3.3.2 in dependencies, Symfony 5 is required. 

This confict will be irrelevant after Symfony update :)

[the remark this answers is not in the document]
RobinDvorak: Info added to commit description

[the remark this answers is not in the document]
grossmannmartin: 🤔 in that case a new version will never be installed even without a conflict, or?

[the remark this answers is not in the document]
RobinDvorak: It didn't work for me. Without conflict it was always trying to install `3.4.0` as it satisfies `^3.3.1`. I checked that again to be sure

grossmannmartin (approved):

PR #2439, opened 2022-05-04
[shopsys] Use getValues() instead of toArray() on Doctrine Collections

### LINKED ISSUE #2409: ArrayCollections: use getValues() instead of toArray()
We use Doctrine `ArrayCollection` for collections in entities and often call `toArray` in the getters, e.g. see https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Customer/Customer.php#L114 - this might be problematic as after deleting items from the collection, `toArray` might return bad keys.

## Suggested solution
We should use `ArrayCollection::getValues` instead of `toArray()`

### Additional context
I see we use a slightly different approach in `Cart::getItems` - `array_values($this->collection->toArray())` (see https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Cart/Cart.php#L105) which might be also legit solution for the problem, however, `getValues()` is more straightforward :slightly_smiling_face:

RobinDvorak:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Using `toArray()` on Doctrine Collections might be problematic as described in #2409  
|New feature| No
|[BC breaks]| No
|Fixes issues| closes #2409 

### DIFF
--- a/packages/framework/src/Model/Administrator/Administrator.php
+++ b/packages/framework/src/Model/Administrator/Administrator.php
@@ -259,7 +259,7 @@ class Administrator implements UserInterface, UniqueLoginInterface, TimelimitLog
      */
     public function getAdministratorRoles(): array
     {
-        return $this->roles->toArray();
+        return $this->roles->getValues();
     }
 
     /**
@@ -348,7 +348,7 @@ class Administrator implements UserInterface, UniqueLoginInterface, TimelimitLog
     {
         $roles = [];
         /** @var \Shopsys\FrameworkBundle\Model\Administrator\Role\AdministratorRole $role */
-        foreach ($this->roles->toArray() as $role) {
+        foreach ($this->roles->getValues() as $role) {
             $roles[] = $role->getRole();
         }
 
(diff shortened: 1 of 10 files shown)


grossmannmartin (approved):

pk16011990:
Please, add this replacing code to the upgrade notes for collections in project-base. I think, that this issue is common for new project entities in project base or in overridden entities methods.

pk16011990 (changes requested):
Please, add this replacing code to the upgrade notes for collections in project-base. I think, that this issue is common for new project entities in project base or in overridden entity methods.

RobinDvorak:
> Please, add this replacing code to the upgrade notes for collections in project-base. I think, that this issue is common for new project entities in project base or in overridden entity methods.

@pk16011990 I didn't found any upgrade notes for Collections in project-base. I added it to classic upgrade notes under Application. You can check the result here: https://github.com/shopsys/shopsys/blob/1ab1faa878ffcf71cb3e2f6ed88ae8b6608ccaff/upgrade/UPGRADE-v9.2.0-dev.md

pk16011990 (approved):

PR #2459, opened 2022-05-19
[framework] CronModuleRepository: fix "Warning: array_pop(): Argument #1 ($array) must be passed by reference, value given"

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| running `cron` fails on the warning message
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Cron/CronModuleRepository.php
+++ b/packages/framework/src/Component/Cron/CronModuleRepository.php
@@ -59,7 +59,7 @@ class CronModuleRepository
             'SELECT cm.serviceId FROM ' . CronModule::class . ' cm WHERE cm.scheduled = TRUE'
         );
 
-        return array_map('array_pop', $query->getScalarResult());
+        return $query->getSingleColumnResult();
     }
 
     /**


packages/framework/src/Component/Cron/CronModuleRepository.php
```diff
@@ -59,7 +59,12 @@ public function getAllScheduledCronModuleServiceIds()
             'SELECT cm.serviceId FROM ' . CronModule::class . ' cm WHERE cm.scheduled = TRUE'
         );
 
-        return array_map('array_pop', $query->getScalarResult());
+        $allScheduledCronModuleServiceIds = [];
+        foreach ($query->getScalarResult() as $row) {
+            $allScheduledCronModuleServiceIds[] = array_pop($row);
+        }
+
+        return $allScheduledCronModuleServiceIds;
```
grossmannmartin, line 67: What about using 

```
return $query->getSingleColumnResult();
```

instead of the re-mapping altogether?

grossmannmartin (changes requested):

packages/framework/src/Component/Cron/CronModuleRepository.php
vitek-rostislav: ah, that is a great idea :ok_hand: :smile:

vitek-rostislav: push forced :innocent:

grossmannmartin (approved):

pk16011990 (approved):

PR #2468, opened 2022-06-02
[shopsys] resolve several Symfony deprecations

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to upgrade Symfony to 5 LTS version so we need to get rid of 4.4 deprecations
|New feature| No 
|Fixes issues| ... 


packages/framework/src/Resources/views/Admin/Content/BestsellingProduct/listGrid.html.twig
```diff
@@ -14,4 +14,4 @@
 {% block grid_pager_totalcount %}
     {% set entityName = 'category'|trans %}
     {{ parent() }}
-{% endblock %}
\ No newline at end of file
+{% endblock %}
```
grossmannmartin, line 17: do you know where is this template used? I did not find the usage...

phpstan.neon
```diff
@@ -9,9 +9,6 @@ parameters:
 
     ignoreErrors:
         # shopsys/framework - don't forget to add these rules to phpstan.neon in framework
-        -
-            message: '#^Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\NodeDefinition::children\(\)\.$#'
-            path: %currentWorkingDirectory%/packages/framework/src/Component/Domain/Config/DomainsUrlsConfigDefinition.php
```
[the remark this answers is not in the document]
vitek-rostislav: ah, yeah, it was there with a different pattern so I missed it the first time, fixed in 1e312257c :+1:

packages/framework/src/Resources/views/Admin/Content/BestsellingProduct/listGrid.html.twig
vitek-rostislav: that is a very good question :smile: even the content is somewhat weird, I removed it in 88aba0c18

grossmannmartin (approved):
nice work

PR #2469, opened 2022-06-10
migrations: fixed namespace selection when only one is available

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When only one namespace is available (eg. when in the context of the project), the first namespace should be selected from available namespaces, not from every registered namespaces
|New feature| No 
|Fixes issues| ... 


vitek-rostislav (approved):
Thanks for the fix! :heart:

pk16011990 (approved):

PR #2476, opened 2022-07-11
jj fw950 test pro kandidata

JorgeSquared:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The work done in this branch should introduce a few purposeful errors into the framework that are in compliance [with this document.](https://docs.google.com/spreadsheets/d/1HRAqxIWY5_1mPT9K9lOXYd1ouO5pH3E2zhsjeDH-So8/edit#gid=47460244)
|New feature| No, at least not in the standard sense.
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| It depends: this work is not intended for release in any public form, therefore it is not clear whether [BC promise](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise) applies here.
|Fixes issues| No known issues.
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| No

### DIFF
--- a/packages/framework/src/Model/Article/ArticleFacade.php
+++ b/packages/framework/src/Model/Article/ArticleFacade.php
@@ -145,6 +145,10 @@ class ArticleFacade
 
         $this->dismantleArticleData($articleData);
 
+        if (count($articleData->urls->toDelete[1]) > 0) {
+            throw new Exception();
+        }
+
         $article->edit($articleData);
         $this->friendlyUrlFacade->saveUrlListFormData('front_article_detail', $article->getId(), $articleData->urls);
 
--- a/packages/framework/src/Model/Article/ArticleFacade.php
+++ b/packages/framework/src/Model/Article/ArticleFacade.php
@@ -145,10 +145,6 @@ class ArticleFacade
 
         $this->dismantleArticleData($articleData);
 
-        if (count($articleData->urls->toDelete[1]) > 0) {
-            throw new Exception();
-        }
-
         $article->edit($articleData);
         $this->friendlyUrlFacade->saveUrlListFormData('front_article_detail', $article->getId(), $articleData->urls);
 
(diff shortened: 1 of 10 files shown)


vitek-rostislav:
> Good point, I got the feeling that when there is nothing marked for deletion, the index exists and its value is `0`. @vitek-rostislav If this is the case, is it still needed to treat it somehow, like use `isset` or something like that?

yeah, I would probably go with the `isset` solution :slightly_smiling_face:

packages/framework/src/Model/Article/ArticleFacade.php
```diff
@@ -143,12 +143,21 @@ public function edit($articleId, ArticleData $articleData)
...
         $articleData->seoMetaDescription = null;
 
-        if (count($articleData->urls->toDelete[1]) > 0) {
+        // error 500 thrown when trying to save changes after a certain url was marked for deletion
+        if (isset($articleData->urls->toDelete[1]) && count($articleData->urls->toDelete[1]) > 0) {
             throw new Exception();
         }
 
+        // disable text alignment in the wysiwyg editor
+        $articleData->text = preg_replace(
```
vitek-rostislav, line 155: I think you should also put the same logic into the `create` method - otherwise, the intended bug will not be there when a new article is created. (And please, DRY, a private method for this would be sexy :grin:) 

The same applies to the meta description modification (you are making it null only here, not in the edit method). It is true that this is solved in `Article::getSeoMetaDescription` method, but it is still inconsistent:
- when creating a new article, SEO meta description is saved into the database, but `null` is returned from the getter
- when editing an article, null is saved into the database
So maybe we can just keep the getter and do not modify the input if we do not mind the value in the database, or we should modify the value in both `create` and `edit` methods :slightly_smiling_face: 

Long story short, it would be nice to unify the approach - if we do not mind the original data in the database, I think the easiest way is just modifying the getters. We can discuss it further if you want, let me know :wink:

vitek-rostislav (changes requested):
Please check my comment :slightly_smiling_face:

packages/framework/src/Resources/views/Admin/Content/Article/list.html.twig
```diff
@@ -1,7 +1,13 @@
 {% extends '@ShopsysFramework/Admin/Layout/layoutWithPanel.html.twig' %}
 
 {% block title %}- {{ 'Articles overview'|trans }}{% endblock %}
-{% block h1 %}{{ 'Articles overview'|trans }} <br> ?{% endblock %}
+{% block h1 %}{{ 'Articles overview'|trans }}
```
vitek-rostislav, line 4: maybe it would be better to put the question mark at the top of `{% block main_content %}` - because now, you have `<h6>` tag inside `<h1>` tag which feels really weird :slightly_smiling_face:

PR #2490, opened 2022-07-21
[shopsys] do not mock Events in tests

RobinDvorak:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Event classes are final in Symfony 5
|New feature| No
|Fixes issues| Allow upgrade to Symfony 5

### DIFF
--- a/packages/framework/tests/Unit/Component/Router/Security/RouteCsrfProtectorTest.php
+++ b/packages/framework/tests/Unit/Component/Router/Security/RouteCsrfProtectorTest.php
@@ -8,6 +8,7 @@ use Shopsys\FrameworkBundle\Component\Router\Security\RouteCsrfProtector;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\ControllerEvent;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\Security\Csrf\CsrfToken;
 use Symfony\Component\Security\Csrf\CsrfTokenManager;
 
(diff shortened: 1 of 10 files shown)


packages/framework/tests/Unit/Component/Router/Security/RouteCsrfProtectorTest.php
```diff
@@ -18,35 +19,31 @@ public function testSubRequest(): void
...
+        $event = new ControllerEvent(
+            $this->createMock(HttpKernelInterface::class),
+            new ControllerProtected(),
+            new Request(),
+            HttpKernelInterface::SUB_REQUEST,
+        );
 
         $routeCsrfProtector = new RouteCsrfProtector($annotationReader, $tokenManagerMock);
-        $routeCsrfProtector->onKernelController($eventMock);
+        $routeCsrfProtector->onKernelController($event);
```
vitek-rostislav, line 30: there are no assertions in this test now :thinking: I was surprised there is nothing like `assertExceptionIsNotThrown` in PHPUnit :smile: I discussed this with @grossmannmartin and we agreed that we would vote for adding `assertTrue(true)` at the end of the test, even thought we are not happy about it :smile:

packages/framework/tests/Unit/Component/Router/Security/RouteCsrfProtectorTest.php
```diff
@@ -18,35 +19,31 @@ public function testSubRequest(): void
...
+        $event = new ControllerEvent(
+            $this->createMock(HttpKernelInterface::class),
+            new ControllerNotProtected(),
+            new Request(),
+            HttpKernelInterface::MASTER_REQUEST,
+        );
 
         $routeCsrfProtector = new RouteCsrfProtector($annotationReader, $tokenManagerMock);
-        $routeCsrfProtector->onKernelController($eventMock);
+        $routeCsrfProtector->onKernelController($event);
```
vitek-rostislav, line 46: this is the same story as `testSubRequest`, so could you please assert true here as well? :slightly_smiling_face:

vitek-rostislav (changes requested):
Hi, good job there, I like the outcome! I have added just two notes (it is actually one duplicated note :smile: )

packages/framework/tests/Unit/Component/Router/Security/RouteCsrfProtectorTest.php
RobinDvorak: Fixed here: https://github.com/shopsys/shopsys/pull/2490/commits/2e480f6a69cef675227ae735ee86e90ae4461a27

packages/framework/tests/Unit/Component/Router/Security/RouteCsrfProtectorTest.php
RobinDvorak: Fixed here: https://github.com/shopsys/shopsys/pull/2490/commits/2e480f6a69cef675227ae735ee86e90ae4461a27

vitek-rostislav (approved):

PR #2492, opened 2022-07-21
[releaser] check GitHub Actions instead of Travis builds for the packages

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We replaced Travis with GitHub Action a while ago
|New feature| Yes 
|Fixes issues| ... 


grossmannmartin (approved):

utils/releaser/src/GithubActions/GithubActionsStatusReporter.php
```diff
@@ -2,26 +2,28 @@
...
     private const EXTRA_PACKAGES = [];
 
     /**
-     * Packages that are not tested on Travis - old packages or forks
+     * Packages that are not tested on GitHub Actions - old packages or forks
      *
      * @var string[]
      */
     private const EXCLUDED_PACKAGES = [
+        'shopsys/jsformvalidator-bundle', // until https://github.com/shopsys/JsFormValidatorBundle/pull/1 is resolved
```
RobinDvorak, line 25: this is already merged :)

RobinDvorak (approved):

utils/releaser/src/GithubActions/GithubActionsStatusReporter.php
vitek-rostislav: In the end, I had to keep it excluded as we get the package name from the packagist, and the name is then put into the Github API URL. However, in the URL, there must be the name as it is used in the package github URL, so we from packagist we got:
```
shopsys/jsformvalidator-bundle
```
but to the API, we should use 
```
shopsys/JsFormValidatorBundle
```

PR #2493, opened 2022-07-24
FE API: all user errors now provide user error code

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The API clients need to know more information when a user error is returned from API. The HTTP code and user error code provide such information and thanks to that, the client can react appropriately to the error state.
|New feature| Yes 
|Fixes issues| ... 


packages/frontend-api/src/Model/Mutation/BaseTokenMutation.php
```diff
@@ -31,14 +31,14 @@ protected function runCheckUserIsLogged(): FrontendApiUser
         $token = $this->tokenStorage->getToken();
 
         if ($token === null) {
-            throw new UserError('Token is not valid.');
+            throw new InvalidTokenUserError('Token is not valid.');
```
grossmannmartin, line 34: Tokens are a little confusing to me right now. We have also `TokenUserMessageException` and its descendants that also prints an error to the user. Wouldn't be nice to unify it somehow? 

For example `TokenFacade::getTokenByString()` also returns "Token is not valid" message, but without code.

packages/frontend-api/src/Model/Mutation/Customer/User/Exception/InvalidAccountOrPasswordUserError.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Mutation\Customer\User\Exception;
+
+use GraphQL\Error\UserError;
+use Shopsys\FrontendApiBundle\Model\Error\UserErrorWithCodeInterface;
+
+class InvalidAccountOrPasswordUserError extends UserError implements UserErrorWithCodeInterface
```
grossmannmartin, line 10: Does it make sense to throw this exception instead of `LoginFailedUserError` in `LoginMutation`?

grossmannmartin (changes requested):

packages/frontend-api/src/Model/Resolver/Article/ArticleResolver.php
```diff
@@ -90,7 +91,7 @@ public function termsAndConditionsArticle(): Article
         $article = $this->legalConditionsFacade->findTermsAndConditions($this->domain->getId());
 
         if ($article === null) {
-            throw new UserError('Terms and condition article was not found');
+            throw new ArticleNotFoundUserError('Terms and condition article was not found', 'terms-and-conditions');
```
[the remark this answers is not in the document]
vitek-rostislav: I created a user story for that and then I tried to implement it and end up with a solution using `SlevomatCodingStandard\Sniffs\PHP\ForbiddenClassesSniff` which is good enough from my point of view :slightly_smiling_face: So I deleted the US and here is the implementation - c6fc606e0 :smile:

packages/frontend-api/src/Model/Mutation/Customer/User/Exception/InvalidAccountOrPasswordUserError.php
vitek-rostislav: yes! :+1: a28452ac1

packages/frontend-api/src/Model/Mutation/BaseTokenMutation.php
vitek-rostislav: good point, what about this? 35feb68ce

grossmannmartin (approved):

pk16011990 (changes requested):
Are upgrade notes needed for custom errors for project API code?

packages/frontend-api/src/Model/Error/InvalidArgumentUserError.php
```diff
@@ -0,0 +1,20 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Error;
+
+use GraphQL\Error\UserError;
+
+class InvalidArgumentUserError extends UserError implements UserErrorWithCodeInterface
+{
+    protected const CODE = 'invalid-argument';
```
pk16011990, line 11: Most of codes are indentical to exception name without postfix `Exception`, but here not and it sounds too generic. In my mind, it invokes "Hey, create some kind of helper with `camelCaseToKebabCase(removeExceptionPostfix(self::class))` and use it in most implementations of `getUserErrorCode()`". What do you thing about it?

pk16011990: Oh, the helper above brings magic and codes will not be searchable :(

vitek-rostislav: yeah, it is a good idea and I like it, however, it would probably worsen the code readability and understanding so for now, I will keep it as-is :slightly_smiling_face:

PR #2495, opened 2022-07-26
[shopsys] Elasticsearch: defined structure for ID

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| If I understand elastic structure right, we use `_id` and `id` where `_id` comes [from array index](https://github.com/shopsys/shopsys/blob/16da913a4dab9cef20132520bfd3c5d384972bf7/packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php#L154) and `id` comes [from classic bulk data](https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php#L224). If my theory is correct, `id` has no defined structure and he should have it. I understand is little thing but I would like to see all data defined in structure, for sure. If I am wrong, you can close this PR. Thanks 🤝
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Product/Search/ProductElasticsearchConverter.php
+++ b/packages/framework/src/Model/Product/Search/ProductElasticsearchConverter.php
@@ -14,6 +14,7 @@ class ProductElasticsearchConverter
     {
         $result = $product;
 
+        $result['id'] = $product['id'] ?? 0;
         $result['availability'] = $product['availability'] ?? '';
         $result['catnum'] = $product['catnum'] ?? '';
         $result['description'] = $product['description'] ?? '';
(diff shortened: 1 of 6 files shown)


packages/framework/src/Model/Product/Search/ProductElasticsearchConverter.php
```diff
@@ -14,6 +14,7 @@ public function fillEmptyFields(array $product): array
     {
         $result = $product;
 
+        $result['id'] = $product['id'] ?? 0;
```
vitek-rostislav, line 17: what about this? :slightly_smiling_face: 
```suggestion
        $result['id'] = $product['id'] ?? $product['_id'] ?? 0;
```

vitek-rostislav (approved):

pk16011990 (approved):

PR #2497, opened 2022-07-26
[frontend-api] avoid session starting in FE API

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| FE API is stateless hence it does not use the session. The problem is the session is started for each GraphQL query anyway. In our code, the session is started even in the FE API for two reasons: 1) autowiring of `FlashBagInterface` Symfony service anywhere in the application -> the session is started during the build of the container (see https://github.com/symfony/symfony/pull/36063); 2) `CartMigrationFacade` is defined as an event listener that listens on the `kernel.controller` event, therefore it is triggered even in the API requests. The facade was originally used in the Twig frontend only so it uses the session, but in the API requests, it does not work (and does not make any sense there). Besides the fact the session is useless in the API, it can be also dangerous as Redis can get overwhelmed by the huge number of pointless session entries. 
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/FlashMessage/FlashBagProvider.php
+++ b/packages/framework/src/Component/FlashMessage/FlashBagProvider.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\FlashMessage;
+
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
+
+class FlashBagProvider
+{
+    /**
+     * @var \Symfony\Component\HttpFoundation\RequestStack
+     */
+    protected RequestStack $requestStack;
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+     */
+    public function __construct(RequestStack $requestStack)
+    {
+        $this->requestStack = $requestStack;
+    }
+
+    /**
+     * @return \Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface|null
+     */
+    public function getFlashBag(): ?FlashBagInterface
+    {
+        $currentRequest = $this->requestStack->getCurrentRequest();
+        if ($currentRequest === null) {
+            return null;
+        }
+
+        /** @var \Symfony\Component\HttpFoundation\Session\Session $session */
+        $session = $currentRequest->getSession();
+
+        return $session->getFlashBag();
+    }
+}
--- a/packages/framework/src/Model/Cart/CartMigrationFacade.php
+++ b/packages/framework/src/Model/Cart/CartMigrationFacade.php
@@ -88,6 +88,15 @@ class CartMigrationFacade
      */
     public function onKernelController(ControllerEvent $event): void
     {
+        $frontendApiControllerClass = 'Shopsys\FrontendApiBundle\Controller\FrontendApiController';
+        $controller = $event->getController();
+        if (is_array($controller) && isset($controller[0])) {
+            $controller = $controller[0];
+        }
+
+        if ($controller instanceof $frontendApiControllerClass) {
+            return;
+        }
         $session = $event->getRequest()->getSession();
 
         $previousCartIdentifier = $session->get(static::SESSION_PREVIOUS_CART_IDENTIFIER);
--- a/packages/framework/src/Component/Error/LogoutExceptionSubscriber.php
+++ b/packages/framework/src/Component/Error/LogoutExceptionSubscriber.php
@@ -5,11 +5,11 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Component\Error;
 
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\FlashMessage\FlashBagProvider;
 use Shopsys\FrameworkBundle\Component\FlashMessage\FlashMessage;
 use Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
-use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\Routing\RouterInterface;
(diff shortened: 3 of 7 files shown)


packages/framework/src/Model/Cart/CartMigrationFacade.php
```diff
@@ -88,6 +88,11 @@ public function mergeCurrentCartWithCart(Cart $cart): void
      */
     public function onKernelController(ControllerEvent $event): void
     {
+        $controller = $event->getController();
+        $frontendApiControllerName = 'Shopsys\FrontendApiBundle\Controller\FrontendApiController';
+        if (class_exists($frontendApiControllerName) && isset($controller[0]) && get_class($controller[0]) === $frontendApiControllerName) {
```
grossmannmartin, line 93: is the check for the existence of the class important here? 

Also `isset($controller[0])` will fail when someone creates an invokable controller (https://symfony.com/doc/current/controller/service.html#invokable-controllers) 

I suggest rewriting it like 

```php
if (is_array($controller)) {
    $controller = $controller[0];
}

if (get_class($controller) === $frontendApiControllerClass) {
```

packages/framework/src/Model/Cart/CartMigrationFacade.php
```diff
@@ -88,6 +88,11 @@ public function mergeCurrentCartWithCart(Cart $cart): void
      */
     public function onKernelController(ControllerEvent $event): void
     {
+        $controller = $event->getController();
+        $frontendApiControllerName = 'Shopsys\FrontendApiBundle\Controller\FrontendApiController';
```
grossmannmartin, line 92: what about `$frontendApiControllerClass`?

grossmannmartin (changes requested):

packages/framework/src/Component/FlashMessage/FlashBagProvider.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * @var \Symfony\Component\HttpFoundation\Session\Session
+     */
+    protected Session $session;
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Session\Session $session
+     */
+    public function __construct(SessionInterface $session)
```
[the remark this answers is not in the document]
RobinDvorak: Request stack also should not be used in constuctor as you can see in paid inspections in Symfony plugin: https://espend.de/phpstorm/plugin/symfony - but I did not dig why

packages/framework/src/Model/Cart/CartMigrationFacade.php
RobinDvorak: @grossmannmartin wouldn't `$controller = $controller[0]` fail if the array is empty?

packages/framework/src/Model/Cart/CartMigrationFacade.php
RobinDvorak, line 92: why not `FrontendApiController::class`?

packages/framework/src/Component/FlashMessage/FlashBagProvider.php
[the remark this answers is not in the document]
vitek-rostislav: ah damn :smile: I will look into it :slightly_smiling_face:

packages/framework/src/Model/Cart/CartMigrationFacade.php
vitek-rostislav: ok, I will rewrite it + check for the empty array, thanks! :+1:

packages/framework/src/Model/Cart/CartMigrationFacade.php
vitek-rostislav: Because the API is not installed in the framework, project-base has the power to install and uninstall it so I can not be sure whether the class is available :confused:

packages/framework/src/Model/Cart/CartMigrationFacade.php
grossmannmartin: can be the array even empty?

packages/framework/src/Component/FlashMessage/FlashBagProvider.php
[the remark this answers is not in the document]
grossmannmartin: That's weird... I would really want to know the reason... Symfony docs mention the usage with requestStack in constructor https://symfony.com/doc/current/session.html#basic-usage 🤔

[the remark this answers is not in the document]
grossmannmartin: oh I see it now... you should not call methods on request stack inside constructor. You can autowire it and pass to property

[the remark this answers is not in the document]
vitek-rostislav: Ok, I have rewritten it with `RequestStack` in 14d52e333. From the discussion below the blog article, I understand that the request stack will not be a service as well one day (when there will be, as Fabien wrote, "a better session management system"), but at the moment, it is really the recommended way of accessing the session. Regarding the paid inspections - it says the request stack must not be used in the constructor but I believe at the moment the simple initialization is ok (when you look closer at the example provided, the method is called on the request stack directly in the constructor, which is IMHO a bad pattern in most of the cases with any service because of possible side effects :slightly_smiling_face: )

packages/framework/src/Model/Cart/CartMigrationFacade.php
vitek-rostislav: > is the check for the existence of the class important here?

as mentioned below, the API is not installed in the framework, project-base has the power to install and uninstall it so I can not be sure whether the class is available

vitek-rostislav: ah, ok, but now I see the `instanceof` does the trick :ok_hand: 33e5f73ca

packages/framework/src/Model/Cart/CartMigrationFacade.php
vitek-rostislav: 33e5f73ca

packages/framework/src/Component/FlashMessage/FlashBagProvider.php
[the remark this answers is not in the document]
vitek-rostislav: ....and fixed tests in 4e7cfe3fc :smile:

grossmannmartin (approved):

pk16011990 (approved):

vitek-rostislav:
Hi @pk16011990, on a project, we encountered a very weird behavior - there was added a session entry in the Redis after doing a request on the storefront. The checker did not reveal this error because, at the time of its execution, the session was not started yet. This makes me think we should ensure the checker is called later than it is called now.

On the project, there was `LogoutExceptionSubscriber` service set as `lazy`. The subscriber was not called but I think that thanks to the lazy loading it was probably initialized after `SessionChecker::onKernelResponse()` was called so we could not be aware of the problem with the started session.

In this PR, the problem is resolved thanks to the usage of `FlashBagProvider` in `LogoutExceptionSubscriber` but I believe we should investigate the behavior described above and make the checker more reliable even in such edge-case scenarios.

vitek-rostislav:
this is already included in the current version of Shopsys Platform

PR #2507, opened 2022-08-26
[framework] bin/console base phing targets are now more verbose

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When there is an error during DB fixtures loading, sometimes, you need more information to find out where the problem is (see the images below)
|New feature| Yes 
|Fixes issues| ... 

Output without the option (old state):
Screenshot from 2022-08-26 22-12-01

Output with the option (new state):
Screenshot from 2022-08-26 22-12-20


vitek-rostislav:
Maybe it is worth adding to all relevant phing targets? :thinking:

PR #2520, opened 2022-12-09
[framework] introduced new phing target demo-data that wipes and completely installs demo data including export to Elasticsearch

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| introduced new phing target `demo-data` that wipes and completely installs demo data including export to Elasticsearch
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/build.xml
+++ b/packages/framework/build.xml
@@ -116,13 +116,13 @@
 
     <target name="build" depends="build-deploy-part-1-db-independent,build-deploy-part-2-db-dependent" description="Builds application for production preserving your DB."/>
 
-    <target name="build-demo" depends="production-protection,wipe,build-version-generate,composer-prod,redis-check,dirs-create,assets,npm,db-demo,elasticsearch-index-recreate,elasticsearch-export,error-pages-generate,warmup,clean-redis-old" description="Builds application for production with clean demo DB."/>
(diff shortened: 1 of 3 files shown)


packages/framework/build.xml
```diff
@@ -309,6 +311,8 @@
         </exec>
     </target>
 
+    <target name="demo-data" depends="production-protection,db-wipe-public-schema,db-import-basic-structure,db-migrations,domains-data-create,db-fixtures-demo,plugin-demo-data-load,friendly-urls-generate,domains-urls-replace,elasticsearch-export" description="Wipes Postgres DB, recreates structure, fills with demo data and then exports data to Elasticsearch"/>
```
pk16011990, line 314: Now there is no separate target that will only allow running demo data. I see this as limiting for common scenarios.

The other thing is that with Postgres, structure creation and migrations are included, but not for Elasticsearch, which is not consistent.

TomasLudvik: I find it correct, that there is no target for running demo data creation only in Postgres, as it leads into faulty application, as you show different data on frontend than the data present in database.

`phing elasticsearch-export` does also migrations if needed see `IndexFacade::export` for more information.

grossmannmartin: I would also appreciate the separate (even hidden) target for load data to database only. 
I use it on purpose when I know I don't need elastic. 

But the final decision is yours of course.

TomasLudvik: I have hide db-demo, but kept it as it was

pk16011990 (approved):

PR #2523, opened 2022-12-15
[framework] remove LoggingMiddleware from connection in data fixtures for faster demodata generation

RostislavKreisinger:
| Q             | A
| ------------- | ---
|Description, reason for the PR| remove LoggingMiddleware from connection in data fixtures for faster demo data generation
|New feature| Yes
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/DataFixture/AbstractReferenceFixture.php
+++ b/packages/framework/src/Component/DataFixture/AbstractReferenceFixture.php
@@ -5,6 +5,9 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Component\DataFixture;
 
 use Doctrine\Common\DataFixtures\FixtureInterface;
+use Doctrine\DBAL\Driver\Middleware as MiddlewareInterface;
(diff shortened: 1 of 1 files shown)


packages/framework/src/Component/DataFixture/AbstractReferenceFixture.php
```diff
@@ -13,6 +16,30 @@ abstract class AbstractReferenceFixture implements FixtureInterface
      */
     protected $persistentReferenceFacade;
 
+    /**
+     * @var bool
+     */
+    protected bool $isLoggerTemporarilyDisabled = false;
```
TomasLudvik, line 22: Why is this necessary? The function is called when the instance is created so this value will be always false, right?

TomasLudvik (changes requested):

packages/framework/src/Component/DataFixture/AbstractReferenceFixture.php
RostislavKreisinger: you are right, I have consulted with the author and we will remove the review

PR #2530, opened 2023-01-04
[framework] remove product variant urls from sitemap

RostislavKreisinger:
| Q             | A
| ------------- | ---
|Description, reason for the PR| the sitemap now contains only the url addresses of the common product and the main product variants. Links to product variants do not have their own landing page, so they are not in the sitemap.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Sitemap/SitemapFacade.php
+++ b/packages/framework/src/Model/Sitemap/SitemapFacade.php
@@ -80,11 +80,11 @@ class SitemapFacade
      * @param \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig $domainConfig
      * @return \Shopsys\FrameworkBundle\Model\Sitemap\SitemapItem[]
      */
-    public function getSitemapItemsForVisibleProducts(DomainConfig $domainConfig)
+    public function getSitemapItemsForListableProducts(DomainConfig $domainConfig)
     {
         $pricingGroup = $this->pricingGroupSettingFacade->getDefaultPricingGroupByDomainId($domainConfig->getId());
 
-        return $this->sitemapRepository->getSitemapItemsForVisibleProducts($domainConfig, $pricingGroup);
+        return $this->sitemapRepository->getSitemapItemsForListableProducts($domainConfig, $pricingGroup);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


packages/framework/src/Model/Sitemap/SitemapFacade.php
```diff
@@ -80,11 +80,11 @@ public function generateForAllDomains()
      * @param \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig $domainConfig
      * @return \Shopsys\FrameworkBundle\Model\Sitemap\SitemapItem[]
      */
-    public function getSitemapItemsForVisibleProducts(DomainConfig $domainConfig)
+    public function getSitemapItemsForListableProducts(DomainConfig $domainConfig)
```
TomasLudvik, line 83: Renaming of public method in frameworks means you are introducing BC break (if anyone extended this method in theirs project, this will cause their application to fail on non existing method). This change should be mentioned in upgrade notes so developers are informed about this and can change theirs application accordingly. Same for SitemapRepository:: getSitemapItemsForListableProducts

TomasLudvik (changes requested):

packages/framework/src/Model/Sitemap/SitemapFacade.php
RostislavKreisinger: ok, i add upgrade note

upgrade/UPGRADE-v11.0.0-dev.md
```diff
@@ -863,7 +863,10 @@ There you can find links to upgrade notes for other versions too.
         ```
     - `transchoice()` method was removed from `\Shopsys\FrameworkBundle\Component\Translation\Translator`, use `trans()` with `count` parameter instead
     - static `tc()` method was removed from `\Shopsys\FrameworkBundle\Component\Translation\Translator`, use `t()` with `count` parameter instead
-    - static `tc()` function was removed from global namespace, use `t()` with `count` parameter instead 
+    - static `tc()` function was removed from global namespace, use `t()` with `count` parameter instead
+- remove product variant urls from sitemap ([#2530](https://github.com/shopsys/shopsys/pull/2530))
+    - Replace `SitemapFacade::getSitemapItemsForVisibleProducts()` with `SitemapFacade::getSitemapItemsForListableProducts()`
+    - Replace `SitemapRepository::getSitemapItemsForVisibleProducts()` with `SitemapRepository::getSitemapItemsForListableProducts()`
```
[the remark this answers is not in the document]
RostislavKreisinger: ok

TomasLudvik (approved):

PR #2537, opened 2023-01-12
[shopsys] update session and cart lifetime

RostislavKreisinger:
| Q             | A
| ------------- | ---
|Description, reason for the PR|  session is valid for a year instead of relation only and the cart is now deleted after 130 days from the user's last activity
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Cart/CartFacade.php
+++ b/packages/framework/src/Model/Cart/CartFacade.php
@@ -18,8 +18,7 @@ use Shopsys\FrameworkBundle\Model\Product\ProductRepository;
 
 class CartFacade
 {
-    protected const DAYS_LIMIT_FOR_UNREGISTERED = 60;
-    protected const DAYS_LIMIT_FOR_REGISTERED = 120;
+    protected const DAYS_LIMIT = 130;
 
     /**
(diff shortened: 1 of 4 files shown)


packages/framework/src/Model/Cart/CartFacade.php
```diff
@@ -252,6 +251,9 @@ public function findCartByCustomerUserIdentifier(CustomerUserIdentifier $custome
 
                 return null;
             }
+
+            $cart->setModifiedNow();
+            $this->em->flush();
```
TomasLudvik, line 256: I am wondering if this is necessary for loading cart. I think that changing modified just during update of cart would be sufficient. Same for `getCartByCustomerUserIdentifierCreateIfNotExists`

TomasLudvik (changes requested):

packages/framework/src/Model/Cart/CartFacade.php
RostislavKreisinger: in acceptance criterie is "since the user's last activity" , this i uderstand as visit any page and not just add/remove item or clean cart.

TomasLudvik (approved):

PR #2555, opened 2023-02-08
[shopsys] Product sorting by availability

RostislavKreisinger:
| Q             | A
| ------------- | ---
|Description, reason for the PR| product ordering in all product lists has been changed, now are products ordered primary by availability dispatch time. In stock as first and out of stock as last.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
+++ b/packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
@@ -244,6 +244,7 @@ class ProductExportRepository
             'calculated_selling_denied' => $product->getCalculatedSellingDenied(),
             'selling_denied' => $product->isSellingDenied(),
             'availability' => $product->getCalculatedAvailability()->getName($locale),
+            'availability_dispatch_time' => $product->getCalculatedAvailability()->getDispatchTime(),
             'is_main_variant' => $product->isMainVariant(),
             'is_variant' => $product->isVariant(),
             'detail_url' => $detailUrl,
--- a/packages/framework/src/Model/Product/Search/FilterQuery.php
+++ b/packages/framework/src/Model/Product/Search/FilterQuery.php
@@ -70,67 +70,70 @@ class FilterQuery
     {
         $clone = clone $this;
 
+        $clone->sorting = [];
+
+        if ($orderingModeId === ProductListOrderingConfig::ORDER_BY_RELEVANCE) {
+            $clone->sorting['_score'] = 'desc';
+
+            return $clone;
+        }
+
+        $clone->sorting['availability_dispatch_time'] = [
+            'order' => 'asc',
+            'missing' => '_last',
+        ];
+
         if ($orderingModeId === ProductListOrderingConfig::ORDER_BY_PRIORITY) {
-            $clone->sorting = [
-                'ordering_priority' => 'desc',
-                'name.keyword' => 'asc',
-            ];
+            $clone->sorting['ordering_priority'] = 'desc';
+            $clone->sorting['name.keyword'] = 'asc';
 
             return $clone;
         }
 
         if ($orderingModeId === ProductListOrderingConfig::ORDER_BY_NAME_ASC) {
-            $clone->sorting = [
-                'name.keyword' => 'asc',
-            ];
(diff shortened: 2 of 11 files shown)


packages/framework/src/Form/Admin/Product/Availability/AvailabilityFormType.php
```diff
@@ -35,6 +36,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'required' => false,
                 'invalid_message' => 'Number of days to expedite must be whole number.',
                 'constraints' => [
+                    new Constraints\NotBlank(['message' => 'Please enter number of days to dispatch. For out of stock fill 65535 as maximal value.']),
```
TomasLudvik, line 39: This is really weird to have out of stock availability set to 179,55 years. Why not keep null? You can work with null in order by https://www.postgresql.org/docs/current/queries-order.html

packages/framework/src/Model/Product/Search/FilterQuery.php
```diff
@@ -70,67 +70,66 @@ public function applyOrdering(string $orderingModeId, PricingGroup $pricingGroup
     {
         $clone = clone $this;
 
+        $clone->sorting = [];
+        $clone->sorting['availability_dispatch_time'] = 'asc';
```
TomasLudvik, line 74: I have mind problem with this as default for all orderings. I think, it should be applied only for default ordering but not for all. Consider I am ordering alphabetically from A to Z and I am looking for Apple iPhone 13 Pro, but it is not on stock yet, because it just has been released and manufacturer cannot keep up with demand, so currently it is out of stock and I do not see it under A, because it is somewhere on 13th page after on stock products with Z at beginning of name.

What do you think @pk16011990 ?

upgrade/UPGRADE-v11.0.0-dev.md
```diff
@@ -881,6 +881,11 @@ There you can find links to upgrade notes for other versions too.
     - see #project-base-diff to update your project
 - old phpRedisAdmin has been replaced with new Redis Commander ([#2550](https://github.com/shopsys/shopsys/pull/2550))
     - see #project-base-diff to update your project
+- product ordering in all product lists has been changed, now are products ordered primary by availability dispatch time. 
+    In stock as first and out of stock as last.  ([#2555](https://github.com/shopsys/shopsys/pull/2555))
+  - A prerequisite for modifying the sorting is canceling the nullable property for dispatch time in availability. 
+  - Now the maximum value of 65535 is assigned to out of stock availability.
+  - see #project-base-diff to update your project
```
TomasLudvik, line 888: We are supposed to write UPGRADE notes as directives what to do, you are describing changes here and it should not be part of UPGRADE notes.

packages/framework/src/Form/Admin/Product/Availability/AvailabilityFormType.php
RostislavKreisinger: in the framework we have an unbeatable generic setting in SortableNullsWalker. After agreement with PK, I chose this solution.

RostislavKreisinger: found more pretty solution

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -244,6 +244,7 @@ protected function extractResult(Product $product, int $domainId, string $locale
             'calculated_selling_denied' => $product->getCalculatedSellingDenied(),
             'selling_denied' => $product->isSellingDenied(),
             'availability' => $product->getCalculatedAvailability()->getName($locale),
+            'availability_dispatch_time' => $product->getCalculatedAvailability()->getDispatchTime(),
```
TomasLudvik, line 247: You are still sending dispatch time to elastic.

TomasLudvik (changes requested):

packages/framework/src/Model/Product/Search/FilterQuery.php
RostislavKreisinger: this is bussiness decision

TomasLudvik (approved):

PR #2559, opened 2023-02-10
[releaser] Several fixes of releaser

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| During release of 10.0.5 I have found several problems, that are fixed in this PR.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/utils/releaser/src/ReleaseWorker/AbstractVerifyInitialBranchReleaseWorker.php
+++ b/utils/releaser/src/ReleaseWorker/AbstractVerifyInitialBranchReleaseWorker.php
@@ -13,7 +13,7 @@ abstract class AbstractVerifyInitialBranchReleaseWorker extends AbstractShopsysR
      * @param string $initialBranchName
      * @return string
      */
-    public function getDescription(Version $version, string $initialBranchName = 'master'): string
+    public function getDescription(Version $version, string $initialBranchName = AbstractShopsysReleaseWorker::MAIN_BRANCH_NAME): string
     {
         return 'Verify that you\'re releasing on the proper branch';
     }
@@ -22,7 +22,7 @@ abstract class AbstractVerifyInitialBranchReleaseWorker extends AbstractShopsysR
      * @param \PharIo\Version\Version $version
      * @param string $initialBranchName
      */
-    public function work(Version $version, string $initialBranchName = 'master'): void
+    public function work(Version $version, string $initialBranchName = AbstractShopsysReleaseWorker::MAIN_BRANCH_NAME): void
     {
         $this->symfonyStyle->note(
             'It is important to perform the release process on the proper branch, i.e. on 7.3 branch when releasing 7.3.x patch version, and on master when releasing new minor version for the current release.'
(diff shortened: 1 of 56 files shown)


utils/releaser/src/ReleaseWorker/StageWorkerInterface.php
```diff
@@ -0,0 +1,34 @@
...
+     * @param \PharIo\Version\Version $version
+     * @return string
+     */
+    public function getDescription(Version $version): string;
+
+    /**
+     * @param \PharIo\Version\Version $version
+     * @param string $initialBranchName
+     */
+    public function work(Version $version, string $initialBranchName = 'master'): void;
```
RostislavKreisinger, line 26: initialBrancheName is required, so default value may not be here ?

utils/releaser/src/ReleaseWorker/AfterRelease/RemoveLockFilesReleaseWorker.php
```diff
@@ -21,22 +21,23 @@ public function getDescription(Version $version): string
...
-    public function work(Version $version): void
+    public function work(Version $version, string $initialBranchName = 'master'): void
     {
         $this->processRunner->run('git rm project-base/composer.lock');
         $this->processRunner->run('git rm project-base/package-lock.json');
         $this->processRunner->run('git rm project-base/migrations-lock.yml');
         $this->commit('removed locked versions of dependencies for unreleased version');
 
-        if ($this->initialBranchName === 'master') {
+        if ($this->currentBranchName === 'master') {
```
RostislavKreisinger, line 35: master could be a constant

utils/releaser/src/ReleaseWorker/AbstractVerifyInitialBranchReleaseWorker.php
```diff
@@ -10,9 +10,10 @@ abstract class AbstractVerifyInitialBranchReleaseWorker extends AbstractShopsysR
 {
     /**
      * @param \PharIo\Version\Version $version
+     * @param string $initialBranchName
      * @return string
      */
-    public function getDescription(Version $version): string
+    public function getDescription(Version $version, string $initialBranchName = 'master'): string
```
RostislavKreisinger, line 16: The default value should not be needed here. This is in other classes to.

utils/releaser/src/ReleaseWorker/StageWorkerInterface.php
TomasLudvik: Sadly interface says, that this method has only one argument, so in order to add another argument, it has to have default value set

utils/releaser/src/ReleaseWorker/AbstractVerifyInitialBranchReleaseWorker.php
TomasLudvik: Same as previous answer

utils/releaser/src/ReleaseWorker/AfterRelease/RemoveLockFilesReleaseWorker.php
TomasLudvik: I have added new constant

utils/releaser/src/ReleaseWorker/StageWorkerInterface.php
RostislavKreisinger: oki, in first view looks like new interface instead of ReleaseWorkerInterface

RostislavKreisinger (approved):

PR #2566, opened 2023-02-20
[shopsys] added name to images

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Image name can be used for image ALT etc.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Component/Image/ImageFacade.php
+++ b/packages/framework/src/Component/Image/ImageFacade.php
@@ -121,10 +121,14 @@ class ImageFacade
                 array_shift($orderedImages);
                 $this->deleteImages($entity, $orderedImages);
             }
-            $this->uploadImage($entity, $uploadedFiles, $type);
+            $this->uploadImage($entity, $imageUploadData->uploadedFilenames, $uploadedFiles, $type);
+            if (count($uploadedFiles) === 0) {
+                $this->saveImagesPathnames($imageUploadData);
+            }
         } else {
             $this->saveImageOrdering($orderedImages);
-            $this->uploadImages($entity, $uploadedFiles, $type);
+            $this->saveImagesPathnames($imageUploadData);
+            $this->uploadImages($entity, $imageUploadData->uploadedFilenames, $uploadedFiles, $type);
         }
 
         $this->deleteImages($entity, $imageUploadData->imagesToDelete);
@@ -132,13 +136,17 @@ class ImageFacade
 
     /**
      * @param object $entity
-     * @param array $temporaryFilenames
+     * @param array<int, array<string,string>> $namesIndexedByImageIdAndLocale
+     * @param array<int, string> $temporaryFilenamesIndexedByImageId
      * @param string|null $type
      */
-    protected function uploadImage($entity, $temporaryFilenames, $type): void
-    {
-        if (count($temporaryFilenames) > 0) {
-            $entitiesForFlush = [];
+    protected function uploadImage(
+        object $entity,
+        array $namesIndexedByImageIdAndLocale,
+        array $temporaryFilenamesIndexedByImageId,
+        ?string $type
+    ): void {
+        if (count($temporaryFilenamesIndexedByImageId) > 0) {
             $imageEntityConfig = $this->imageConfig->getImageEntityConfig($entity);
             $entityId = $this->getEntityId($entity);
             $oldImage = $this->imageRepository->findImageByEntity(
@@ -149,17 +157,16 @@ class ImageFacade
 
             if ($oldImage !== null) {
                 $this->em->remove($oldImage);
-                $entitiesForFlush[] = $oldImage;
             }
 
             $newImage = $this->imageFactory->create(
                 $imageEntityConfig->getEntityName(),
                 $entityId,
+                array_pop($namesIndexedByImageIdAndLocale),
+                array_pop($temporaryFilenamesIndexedByImageId),
                 $type,
-                array_pop($temporaryFilenames)
             );
             $this->em->persist($newImage);
-            $entitiesForFlush[] = $newImage;
 
             $this->em->flush();
         }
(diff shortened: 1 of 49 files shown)


packages/framework/src/Component/Image/ImageFacade.php
```diff
@@ -529,4 +542,32 @@ public function getImagesByEntityId(int $id, string $entityClass): array
 
         return $this->getImagesByEntityIdAndNameIndexedById($id, $entityName, null);
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\FileUpload\ImageUploadData $imageUploadData
+     * @param object $entity
+     * @param string|null $type
+     */
+    public function setImageUploadDataForEntityAndType(ImageUploadData $imageUploadData, object $entity, ?string $type = null): void
```
RostislavKreisinger, line 551: I was thining before why we don't have an ImageUploadDataFactory to build ImageUploadData and instead it builds sequentially in three places (eg: TransportData, TransportDataFactory, ImageFacade). Could we dedicate the whole thing to the factory?

packages/framework/src/Component/Image/ImageFacade.php
```diff
@@ -529,4 +542,32 @@ public function getImagesByEntityId(int $id, string $entityClass): array
...
+    {
+        $orderedImages = $this->getImagesByEntityIndexedById($entity, $type);
+        $imageUploadData->orderedImages = $orderedImages;
+
+        foreach ($orderedImages as $orderedImage) {
+            $imageUploadData->currentFilenamesIndexedByImageIdAndLocale[$orderedImage->getId()] = $orderedImage->getNames();
```
RostislavKreisinger, line 557:
```suggestion
            $imageUploadData->namesIndexedByImageIdAndLocale[$orderedImage->getId()] = $orderedImage->getNames();
```
i don't understant why it's named currentFilenames. It's name of image?

packages/framework/src/Component/Image/ImageFacade.php
```diff
@@ -529,4 +542,32 @@ public function getImagesByEntityId(int $id, string $entityClass): array
...
+        }
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\FileUpload\ImageUploadData $imageUploadData
+     */
+    protected function saveImagesPathnames(ImageUploadData $imageUploadData): void
+    {
+        foreach ($imageUploadData->currentFilenamesIndexedByImageIdAndLocale as $imageId => $filenamesIndexedByLocale) {
```
RostislavKreisinger, line 566:
```suggestion
        foreach ($imageUploadData->namesIndexedByImageIdAndLocale as $imageId => $namesIndexedByLocale) {
```

packages/framework/src/Component/Image/ImageFactory.php
```diff
@@ -43,36 +43,41 @@ public function __construct(
...
+     * @param array $names
      * @param string $temporaryFilename
+     * @param string|null $type
+     * @throws \Shopsys\FrameworkBundle\Component\Image\Processing\Exception\FileIsNotSupportedImageException
      * @return \Shopsys\FrameworkBundle\Component\Image\Image
      */
     public function create(
         string $entityName,
         int $entityId,
+        array $names,
```
RostislavKreisinger, line 55: it could be more descriptive like `namesIndexedByLocale`

packages/framework/src/Component/Image/ImageFactory.php
```diff
@@ -81,8 +86,8 @@ public function createMultiple(
         }
 
         $images = [];
-        foreach ($temporaryFilenames as $temporaryFilename) {
-            $images[] = $this->create($imageEntityConfig->getEntityName(), $entityId, $type, $temporaryFilename);
+        foreach ($temporaryFilenames as $key => $temporaryFilename) {
```
RostislavKreisinger, line 89: $key is a generic name. What is actually in it? I'm guessing something like image id.

packages/framework/src/Component/Image/ImageFactory.php
```diff
@@ -81,8 +86,8 @@ public function createMultiple(
...
+            $images[] = $this->create($imageEntityConfig->getEntityName(), $entityId, $names[$key] ?? [], $temporaryFilename, $type);
```
RostislavKreisinger, line 90: I am wondering if the coalescence makes sense and if such a case can occur, if it is not some kind of programming error. It would then be hidden by that coalescence. Wouldn't it be better to throw the exception?

packages/framework/src/Component/Image/ImageRepository.php
```diff
@@ -75,8 +75,9 @@ public function getImageByEntity($entityName, $entityId, $type)
     public function getImagesByEntityIndexedById($entityName, $entityId, $type)
     {
         $queryBuilder = $this->em->createQueryBuilder()
-            ->select('i')
+            ->select('i, it')
```
RostislavKreisinger, line 78: eager loading ❤️

packages/framework/src/Form/ImageUploadType.php
```diff
@@ -124,6 +126,39 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                        'constraints' => [
+                            new Assert\Length([
+                                'max' => 245,
+                                'maxMessage' => 'File name cannot be longer than {{ limit }} characters',
+                            ]),
+                        ],
+                    ],
+                ],
+            ])->add(
+                $builder->create('currentFilenamesIndexedByImageIdAndLocale', CollectionType::class, [
```
RostislavKreisinger, line 146:
```suggestion
                $builder->create('namesIndexedByImageIdAndLocale', CollectionType::class, [
```
in the task specification it is ALT

project-base/src/DataFixtures/Demo/ImageDataFixture.php
```diff
@@ -301,7 +315,7 @@ private function moveFilesFromLocalFilesystemToFilesystem(string $origin, string
     private function truncateImagesFromDb()
     {
         $this->em->getConnection()->executeStatement(
-            'TRUNCATE TABLE ' . self::IMAGES_TABLE_NAME
+            'TRUNCATE TABLE ' . self::IMAGES_TABLE_NAME . ', ' . self::IMAGES_TRANSLATIONS_TABLE_NAME
```
RostislavKreisinger, line 318: shouldn't the CASCADE setting solve this?

packages/framework/src/Component/Image/ImageFactory.php
TomasLudvik: Well it is array key generated by PHP

packages/framework/src/Component/Image/ImageFactory.php
TomasLudvik: That array can be empty, I do not want to force it so I solved it like this, feel free to suggest better solution.

packages/framework/src/Form/ImageUploadType.php
TomasLudvik: Yes, it is alternative for image, when it is not loaded. We use name for this by default.

project-base/src/DataFixtures/Demo/ImageDataFixture.php
TomasLudvik: It probably does, but I do not see problem with declaring it explicitly here.

packages/framework/src/Component/Image/ImageFacade.php
TomasLudvik: changed

packages/framework/src/Component/Image/ImageFacade.php
TomasLudvik: changed

packages/framework/src/Component/Image/ImageFactory.php
TomasLudvik: changed

packages/framework/src/Component/Image/ImageFacade.php
TomasLudvik: Well I have reworked it as you wished

RostislavKreisinger (approved):

RostislavKreisinger (approved):

PR #2583, opened 2023-03-20
[shopsys] Adverts on product list can now be set only for range of categories

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Administrators need to set adverts in product list only for some categories.
|New feature| Yes 
|Fixes issues| ... 


packages/framework/src/Model/Advert/Advert.php
```diff
@@ -172,4 +191,12 @@ public function getPositionName()
     {
         return $this->positionName;
     }
+
+    /**
+     * @return \Doctrine\Common\Collections\Collection|\Shopsys\FrameworkBundle\Model\Category\Category[]
+     */
+    public function getCategories(): Collection|array
```
LuKillman, line 198: I think it doesn't return `Collection`.

packages/framework/src/Model/Advert/AdvertRepository.php
```diff
@@ -40,26 +41,41 @@ public function findById($advertId)
...
             ->select('a')
             ->from(Advert::class, 'a')
             ->where('a.positionName = :positionName')->setParameter('positionName', $positionName)
             ->andWhere('a.hidden = FALSE')
             ->andWhere('a.domainId = :domainId')->setParameter('domainId', $domainId);
+
+        if ($category !== null) {
+            $queryBuilder
+                ->leftJoin('a.categories', 'c')
+                ->andWhere('c IN(:category) OR c IS NULL')
```
LuKillman, line 63: 1. Space is missing after `IN`. 
2. In case of single category is it good to use `IN` rather than `c = :category`?
3. I would swap the order of those conditions to be sure it's not `NULL` before comparing it. 

```suggestion
                ->andWhere('c IS NULL OR c = :category')
```

packages/frontend-api/src/Model/Advert/AdvertRepository.php
```diff
@@ -53,13 +55,23 @@ protected function getAdvertRepository()
...
     {
-        return $this->getVisibleAdvertsQueryBuilder($domainId)
+        $queryBuilder = $this->getVisibleAdvertsQueryBuilder($domainId)
             ->andWhere('a.positionName = :positionName')
             ->setParameter('positionName', $positionName);
+
+        if ($category !== null) {
+            $queryBuilder
+                ->leftJoin('a.categories', 'c')
+                ->andWhere('c IN(:category) OR c IS NULL')
```
LuKillman, line 70: Same comment as in framework/AdvertRepository in case that one gets accepted. :slightly_smiling_face:

packages/framework/assets/js/admin/validation/form/validationAdvert.js
```diff
@@ -19,6 +19,26 @@ export default function validationAdvert () {
     $advertForm.find('input[name="advert_form[settings][type]"]').change(initAdvertForm);
     initAdvertForm();
 
+    const getPositionName = function () {
+        return $('select[name="advert_form[settings][positionName]"]').val();
+    };
+
+    const initAdvertProductList = function () {
+        console.log(getPositionName());
```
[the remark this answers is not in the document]
TomasLudvik: Thank you, I have foolishly thought we have standards for that :)

packages/framework/src/Model/Advert/AdvertRepository.php
TomasLudvik: Good point

LuKillman (approved):

packages/framework/src/Model/Advert/Advert.php
```diff
@@ -70,6 +72,13 @@ class Advert
      */
     protected $hidden;
 
+    /**
+     * @var \Shopsys\FrameworkBundle\Model\Category\Category[]|\Doctrine\Common\Collections\Collection
```
vitek-rostislav, line 76:
```suggestion
     * @var \Doctrine\Common\Collections\Collection<int, \Shopsys\FrameworkBundle\Model\Category\Category>
```

packages/framework/src/Model/Advert/Advert.php
```diff
@@ -70,6 +72,13 @@ class Advert
...
+     * @ORM\ManyToMany(targetEntity="Shopsys\FrameworkBundle\Model\Category\Category")
+     * @ORM\JoinTable(name="advert_categories")
+     */
+    protected $categories;
```
vitek-rostislav, line 80: why do not we use strict typehints? (even in `AdvertData, there is a new property without the typehint) :innocent: with the suggestion above, you can add `Collection` typehint here.

packages/framework/src/Model/Advert/Advert.php
TomasLudvik: I have updated it

packages/framework/src/Model/Advert/Advert.php
TomasLudvik: I have added typehints as you suggested

PR #2588, opened 2023-04-04
[shopsys] checking if order status update email should be sent is done directly in OrderMailFacade instead in multiple places

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Checking if email can be sent was done on multiple places and developers could forget to do such check. Using of sendMail method should be now more straightforward.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/project-base/src/Controller/Front/OrderController.php
+++ b/project-base/src/Controller/Front/OrderController.php
@@ -6,7 +6,6 @@ namespace App\Controller\Front;
 
 use App\Form\Front\Order\DomainAwareOrderFlowFactory;
 use App\Model\Order\FrontOrderData;
-use App\Model\Order\Order;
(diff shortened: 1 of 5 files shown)


project-base/src/Controller/Front/OrderController.php
```diff
@@ -421,12 +421,6 @@ private function getTermsAndConditionsResponse(): Response
      */
     private function sendMail(Order $order): void
```
LuKillman, line 422: I think this method might no longer be necessary now that it contains only 1 line.

TomasLudvik: removed

LuKillman (approved):

pk16011990 (approved):

PR #2605, opened 2023-04-25
[framework] RegisterProjectFrameworkClassExtensionCompilerPass now ensures that both class and interface alias are registered in Symfony services

### LINKED ISSUE #2400: Annotations fixer not working when extending class by its interface
## Describe the bug

Annotation fixer does not fix my extended class declared like this in services.yaml
```
Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserDataFactoryInterface:
    alias: App\Model\Customer\User\CustomerUserDataFactory
```
but will fix when extended like this:
```
Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserDataFactory:
    alias: App\Model\Customer\User\CustomerUserDataFactory
```
### Expected behavior

Annotation fixer works in both cases

### LINKED ISSUE #2539: annotation fixer ignores services extended with `class` keyword
## Describe the bug
When service is extended with the `class` keyword (for example when tagging is needed) 

```yaml
    Shopsys\FrameworkBundle\Model\Product\ProductVisibilityFacade:
        class: App\Model\Product\ProductVisibilityFacade
        tags:
            # must be ran after CategoryVisibilityRepository and ProductPriceRecalculator
            - {
                name: kernel.event_listener,
                event: kernel.response,
                method: onKernelResponse,
                priority: 10,
                }
```

Annotation fixer do not change annotations at all.  

When `alias` is used

```yaml
    Shopsys\FrameworkBundle\Model\Product\ProductVisibilityFacade:
        alias: App\Model\Product\ProductVisibilityFacade
```

Everything work as expected.

### Expected behavior
Annotations for services extended with `class` keyword are replaced the same way as they are with `alias`, or the behavior is documented, if it's intentional.

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| All variants of service needs to be registered in class extension for annotation fixer to work correctly.
|New feature| No 
|Fixes issues| closes #2400 closes #2539  

### DIFF
--- a/packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
+++ b/packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
@@ -4,9 +4,15 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\DependencyInjection\Compiler;
 
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflection\ReflectionObject;
+use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
 use Shopsys\FrameworkBundle\Component\ClassExtension\ClassExtensionRegistry;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
 
 class RegisterProjectFrameworkClassExtensionsCompilerPass implements CompilerPassInterface
 {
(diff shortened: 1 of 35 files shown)


packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
```diff
@@ -16,11 +20,26 @@ class RegisterProjectFrameworkClassExtensionsCompilerPass implements CompilerPas
...
             if ($this->isFrameworkClassWithAlias($container, $serviceId)) {
                 $aliasId = (string)$container->getAlias($serviceId);
+
+                try {
+                    $serviceClass = $container->findDefinition($serviceId)->getClass();
+                } catch (ServiceNotFoundException) {
+                    $serviceClass = null;
+                }
+
+                if ($serviceClass !== null && !$this->isProjectClass($aliasId) && $this->isProjectClass($serviceClass)) {
```
grossmannmartin, line 34: This check is done only if the service is already aliased... meaning this condition can be met only if
- current service is a framework class
- current service has the alias defined
- the alias is not a project class
- current service has the class set as a project class

how should look the definition to meet these conditions?

packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
```diff
@@ -44,4 +63,29 @@ private function isProjectClass(string $serviceId): bool
...
+     * @param \Symfony\Component\DependencyInjection\Definition $classExtensionRegistryDefinition
+     * @param string $aliasId
+     */
+    private function addAllVariantsOfServiceToClassExtension(
+        ReflectionClass $frameworkClassBetterReflection,
+        Definition $classExtensionRegistryDefinition,
+        string $aliasId
+    ): void {
+        if ($frameworkClassBetterReflection->isInterface()) {
+            $className = rtrim($frameworkClassBetterReflection->getName(), 'Interface');
```
grossmannmartin, line 78: rtrim does not work as presumed here. 

for example  
`rtrim('Myface', 'Interface')` results in `My` 
`rtrim('MyFacadeInterface', 'Interface')` result in `MyFacad`

Maybe `preg_replace('/Interface$/', '', $string);` would be a better choice

packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
```diff
@@ -44,4 +63,29 @@ private function isProjectClass(string $serviceId): bool
...
+
+            if (class_exists($className)) {
+                $classExtensionRegistryDefinition->addMethodCall('addExtendedService', [$className, $aliasId]);
+            }
+        } else {
+            $interfaceName = $frameworkClassBetterReflection->getName() . 'Interface';
+
+            if (interface_exists($interfaceName)) {
```
grossmannmartin, line 86: Wouldn't be nice also to check whether the class implements the interface?

If I just create the `Shopsys\FrameworkBundle\Model\Cart\Watcher\CartWatcherFacadeInterface` interface, the link between this interface and ordinary CartWatcherFacade will be created in the service extension map, even though it does not make sense at this point

grossmannmartin (changes requested):

packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
TomasLudvik: This happens when class is extended in services.yaml instead of aliasing.

Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserUpdateDataFactory:
    class: App\Model\Customer\User\CustomerUserUpdateDataFactory

packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
TomasLudvik: Changed

packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
TomasLudvik: Added

packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php
grossmannmartin: This class will work, because it implements the interface and this interface has defined alias in services.yaml (https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Resources/config/services.yaml#L756) 

I tried it this way: 
- create new `\App\Model\Product\TopProduct\TopProductFacade` class extending the framework's one
- create new `\App\Controller\Admin\TopProductController`class extending the framework's one
- extend it in services.yaml with the `class`
    ```
    Shopsys\FrameworkBundle\Model\Product\TopProduct\TopProductFacade:
        class: App\Model\Product\TopProduct\TopProductFacade
    ```
- run `php phing annotations-fix`

In the current implementation, only `@method \App\Model\Product\Product[] getProductsForDomain(int $domainId)` annotation is added. 

Expected is also the annotation for the constructor and property `$topProductFacade`

grossmannmartin: I have tested it on our internal project and I got the `You have requested a non-existent service "App\Model\Mail\MailTemplateConfiguration".` error.

Can you check it, please?

grossmannmartin (approved):

PR #2619, opened 2023-05-29
[infrastructure] added review stage

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to have everything possible built on GitHub Actions so we are slowly moving our internal CI functions, like this one.
|New feature| No 
|Fixes issues| ... 


.github/workflows/docker-build.yaml
```diff
@@ -16,67 +16,99 @@ jobs:
...
-            -   name: Login to Docker Hub
-                uses: docker/login-action@v1
+            -   name: Login to GitHub Container Registry
+                uses: docker/login-action@v2
                 with:
-                    username: ${{ secrets.DOCKER_USERNAME }}
-                    password: ${{ secrets.DOCKER_PASSWORD }}
+                    registry: ghcr.io
+                    username: ${{ github.actor }}
+                    password: ${{ secrets.GITHUB_TOKEN }}
```
grossmannmartin, line 33: Does it need some settings for other team members? 
The actor is the one who pushed, right (I see TomasLudvik in the output), so when somebody else with the maintainer permission creates a PR, will it be accepted with the token?

.github/workflows/docker-build.yaml
```diff
@@ -229,10 +261,47 @@ jobs:
...
+            -   name: Extract branch name to ENV variable BRANCH_NAME
+                run: echo "BRANCH_NAME=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_ENV
+            -   name: Create branch directory
+                run: |
+                    rm -rf ${BRANCH_NAME}
+                    mkdir ${BRANCH_NAME}
+            -   name: Set ENV variables required for review
+                run: |
+                    HOST=${BRANCH_NAME}.${{ vars.REVIEW_STAGE_URL }}
+                    echo "HOSTS=${HOST}, en.${HOST}" >> $GITHUB_ENV
```
grossmannmartin, line 286: scond domain should be `en`? in traefik hosts I see `cz`.

.github/set-urls-for-review.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -e
+
+setUrlsToDomainsUrls () {
+    DOMAINS=${HOSTS}
+    ITERATOR=1
+
+    for DOMAIN in ${DOMAINS//,/ } ; do
+        docker compose exec -T php-fpm sed -i "s/http:\/\/127.0.0.${ITERATOR}:8000/https:\/\/${DOMAIN}/g" project-base/config/domains_urls.yaml
```
grossmannmartin, line 8: maybe it could be rewritten without docker compose as a script that is meant to be run inside the container. 
Then it would not be necessary to transfer the file through artifacts

docker/conf/docker-compose.github-actions.review.yml.dist
```diff
@@ -0,0 +1,91 @@
...
+            - -c
+            - config_file=/var/lib/postgresql/conf/postgres.conf
+        networks:
+            - default
+        labels:
+            - traefik.enable=false
+
+    webserver:
+        restart: always
+        image: nginx:1.13-alpine
```
grossmannmartin, line 24: Can we use newer version?

.github/workflows/cancel-old-reviews.yaml
```diff
@@ -0,0 +1,17 @@
...
+name: Cancel reviews older than seven days
+jobs:
+    rebuild-master-images:
+        name: Cancel reviews older than seven days
+        runs-on: [ self-hosted, linux, review-stage ]
+        steps:
+            -   name: GIT checkout branch - master
+                uses: actions/checkout@v3
+                with:
+                    ref: 'refs/heads/master'
```
grossmannmartin, line 13: What about just curl the required file instead of the full checkout of the repository?

.github/workflows/cancel-old-reviews.yaml
```diff
@@ -0,0 +1,17 @@
...
+            -   name: Cancel reviews older than seven days
+                run: |
+                    chmod +x ./.github/cancel-old-reviews.sh
+                    ./.github/cancel-old-reviews.sh
```
grossmannmartin, line 17: you can run the file with `bash ./.github/cancel-old-reviews.sh`. That way it's not necessary to set the executable bit.

.github/workflows/cancel-old-reviews.yaml
```diff
@@ -0,0 +1,17 @@
+on:
+    schedule:
+        -   cron: '0 2 * * *'
+name: Cancel reviews older than seven days
+jobs:
+    rebuild-master-images:
```
grossmannmartin, line 6: shouldn't be this job named differently?

.github/workflows/cancel-old-reviews.yaml
grossmannmartin: 🤔  the `manually_cancel_review` workflow is written without checkout or curl... is it necessary here at all?

grossmannmartin (commented):
really nice work 💚

.github/workflows/docker-build.yaml
```diff
@@ -16,67 +16,99 @@ jobs:
     build-docker-images:
         if: github.event.pull_request.head.repo.full_name == 'shopsys/shopsys'
         name: Build PHP-FPM and Elasticsearch images
-        needs: cancel
         runs-on: ubuntu-20.04
+        permissions:
+            contents: read
+            packages: write
```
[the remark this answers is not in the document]
TomasLudvik: No, this sets permissions for GITHUB_TOKEN

.github/workflows/docker-build.yaml
TomasLudvik: Yes, permissions says that GITHUB_TOKEN has to be created with write permissions for packages, so nothing else is required for other maintainers.

.github/workflows/docker-build.yaml
TomasLudvik: Second should be cz - fixed

.github/set-urls-for-review.sh
TomasLudvik: It is dockerignored, that is why I need to move it

docker/conf/docker-compose.github-actions.review.yml.dist
TomasLudvik: We can but it is version we use in all of our docker-compose files

.github/workflows/cancel-old-reviews.yaml
TomasLudvik: Well, I need that file :D do you have a better idea?

.github/workflows/cancel-old-reviews.yaml
TomasLudvik: Good catch!

.github/workflows/manually-cancel-review.yaml
```diff
@@ -0,0 +1,19 @@
...
+    workflow_dispatch:
+        inputs:
+            branch_name:
+                description: 'Branch name'
+                required: true
+                type: string
+name: Manually cancel review
+jobs:
+    cancel-review:
+        if:  ${{ inputs.branch_name }} != 'master'
```
grossmannmartin, line 11: Here you should probably somehow obtain the cancel-review.sh file

TomasLudvik: fixed

grossmannmartin (approved):

grossmannmartin (approved):

PR #2636, opened 2023-06-22
Don't create brand view if brand id is null from elastic data

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| If product doesn't have brand then to elastic is exported empty string (`''`) and so then will create `BrandView` with data: `id = 0, name = '', mainUrl = ''` and if you wanna check Brand exists you have to check `id !== 0` or something like that. I recommend returning `null` instead of "empty" `BrandView` [like creating BrandView from product entity](https://github.com/shopsys/shopsys/blob/master/packages/read-model/src/Brand/BrandViewFacade.php#L52-L63). Thanks 🤝. P.S.: you have defined brand as integer bud if product doesn't have brand you sent empty string (`''`) and it's not correct (I think)
|New feature| Yes
|Fixes issues| ... 


pk16011990:
The current version does not contain a read-model, and FE-API does not load brands from Elasticsearch.

PR #2659, opened 2023-07-11
improvements to Urql client and query error handling 

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| After the first part of connectors removal, there were still some issues (over-mapping images, the need to wrap all queries with useQueryError). This PR solves them, and additionally takes care of refactoring the Urql client.
|New feature| No
|Fixes issues|


tvikito (approved):

project-base/storefront/components/Layout/LoadingHandler.tsx
```diff
@@ -1,26 +1,9 @@
...
-        showSuccessMessage(t('Successfully logged in'));
-
-        if (loginLoading === 'loading-with-cart-modifications') {
-            showInfoMessage(t('Your cart has been modified. Please check the changes.'));
-        }
-
-        updateGeneralState(null);
-    }, []);
+    useLoginLoader();
+    usePageLoader();
```
tvikito, line 6: This component doesn't seem to fulfill it's purpose. Wouldn't be better idea to move those two hooks directly into it's parrent AppPageContent?

project-base/storefront/components/Pages/App/CartReloader.tsx
```diff
@@ -0,0 +1,7 @@
+import { useReloadCart } from 'hooks/cart/useReloadCart';
+
+export const CartReloader: FC = () => {
+    useReloadCart();
+
+    return null;
+};
```
tvikito, line 7: Is it necessary to have component for this? I don't see much of the point to have those "microcomponents" with one hook. Why not use it directly in AppPageContent?

project-base/storefront/components/Pages/App/PageHeadScript.tsx
tvikito, line 1: I am not sure if this has changed in another PR, but can we move content from this component into AppPageContent? Since links with fonts are removed there is only GtmHeadScript left.

project-base/storefront/package.json
```diff
@@ -20,7 +20,7 @@
         "typecheck": "tsc --pretty --noEmit --project tsconfig.json"
     },
     "dependencies": {
-        "@floating-ui/react-dom-interactions": "0.9.1",
+        "@floating-ui/react": "^0.24.6",
```
tvikito, line 23: Let's put specific version 

`"@floating-ui/react": "0.24.6",`

project-base/storefront/package.json
```diff
@@ -55,17 +55,17 @@
...
     },
     "devDependencies": {
-        "@graphql-codegen/cli": "^3.2.2",
-        "@graphql-codegen/fragment-matcher": "^4.0.1",
-        "@graphql-codegen/introspection": "^3.0.1",
-        "@graphql-codegen/typescript-operations": "^3.0.2",
+        "@graphql-codegen/cli": "^4.0.1",
+        "@graphql-codegen/fragment-matcher": "^5.0.0",
+        "@graphql-codegen/introspection": "^4.0.0",
+        "@graphql-codegen/typescript-operations": "^4.0.1",
```
tvikito, line 68: Let's put specific version

project-base/storefront/components/Layout/LoadingHandler.tsx
sebaholesz: Done

project-base/storefront/components/Pages/App/AppPageContent.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { usePersistStore } from 'store/zustand/usePersistStore';
+import { Error500ContentWithBoundary } from '../ErrorPage/Error500Content';
+import { Error503Content } from '../ErrorPage/Error503Content';
+import { PageHeadScripts } from './PageHeadScript';
+import { CartReloader } from './CartReloader';
+
+type AppPageContentProps = {
+    Component: NextComponentType<NextPageContext, any, any>;
+    pageProps: ServerSidePropsType;
+    err?: any;
```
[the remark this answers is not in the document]
sebaholesz: Its typed like this since the beginning, I really dont know what type it is :cry:

project-base/storefront/helpers/errors/ignoredErrors.ts
```diff
@@ -1,7 +1,8 @@
-import { ApplicationErrors, ApplicationErrorsType } from './applicationErrors';
+import { ApplicationErrors } from './applicationErrors';
 
 // errors we don't want to display to users
-export const ApplicationIgnoredErrors: readonly ApplicationErrorsType[] = [
+export const ApplicationIgnoredErrors = [
     ApplicationErrors['no-result-found-for-slug'],
+    ApplicationErrors['seo-page-not-found'],
     ApplicationErrors['invalid-token'],
```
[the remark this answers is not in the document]
sebaholesz: Because if we do not do it, every time a refresh happens in the background, the user gets a message, which is not desired.

project-base/storefront/package.json
sebaholesz: I can see this is the current state, probably changed in a later commit.

project-base/storefront/package.json
sebaholesz: I can see this is the current state, probably changed in a later commit.

project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx
```diff
@@ -60,10 +61,10 @@ const BlogArticleDetailPage: NextPage = () => {
     );
 };
 
-export const getServerSideProps = getServerSidePropsWithRedisClient((redisClient) => async (context) => {
-    const domainConfig = getDomainConfig(context.req.headers.host!);
+export const getServerSideProps = getServerSidePropsWithRedisClient((redisClient, domainConfig) => async (context) => {
     const ssrCache = ssrExchange({ isClient: false });
-    const client = createClient(context, domainConfig.publicGraphqlEndpoint, ssrCache, redisClient);
+    const t = await getT(domainConfig.defaultLocale, 'common');
```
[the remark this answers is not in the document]
Comment (sebaholesz) [reply project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx:64]: We need the `createClient` function to be synchronous as we use it to create the client in `_app.tsx`

image

I am not sure what should be the workaround. Do you have an idea?

project-base/storefront/components/Pages/App/PageHeadScript.tsx
sebaholesz: It made more sense to me to do it [here](https://github.com/shopsys/shopsys/pull/2665/commits/5f5f68ee7066c32015c61f27efd991cf5c5a2380). Tell me if thats alright with you.

project-base/storefront/components/Pages/App/AppPageContent.tsx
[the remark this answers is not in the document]
tvikito: Ok, let's not waste time on this then.

project-base/storefront/components/Pages/App/PageHeadScript.tsx
tvikito: Yes, sure. I was already lost where those changes were placed.

project-base/storefront/package.json
```diff
@@ -87,21 +79,18 @@
         "@types/uuid": "8.3.1",
         "@typescript-eslint/eslint-plugin": "5.59.1",
         "@typescript-eslint/parser": "5.59.1",
-        "autoprefixer": "10.4.13",
+        "autoprefixer": "^10.4.14",
```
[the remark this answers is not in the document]
tvikito: I can still see allowance for minor upgrades. Let's fix the version for all deps.

project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx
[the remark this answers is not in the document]
Comment (tvikito) [reply project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx:64]: I have pushed commit with possible refactoring. It has only downside, that we have to revert part of the cleaning `_app` component. So I will let you decide if to implement this all over the app.

project-base/storefront/package.json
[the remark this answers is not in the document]
sebaholesz: Done

project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx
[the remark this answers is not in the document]
Comment (sebaholesz) [reply project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx]: I checked your commit and noticed several things you've done:

- `domainConfig` is not created in `initServerSideProp`, but injected from outside
- `t` function is not created in every page file, but in one place
- `ssrExchange` is not created in multiple places, but in one

And I have taken these ideas and added some of mine:

- `_app` is not wrapped with a weird, pseudo code which creates a client with a non-existent URL and without a fetcher
- `initServerSideProps` has its parameters typed in a better way so either a client or a set of building blocks with which a client can be built are passed
- type the return type of `createClient` better so it does not have to be checked if the client is `null`

With these ideas in mind I did what you can see in this [commit](https://github.com/shopsys/shopsys/pull/2659/commits/c85ba8dbe1487a217fa5591ecc64ce7c5ea85290). I hope we can find a middle ground here. I must say I like the logic of a wrapper making some stuff available for the wrapped function. Seems like a neat way of using a decorator. Let me know what you think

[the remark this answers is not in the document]
Comment (tvikito) [reply project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx]: Superb! Very nice improvement with both ideas 💯

PR #2661, opened 2023-07-12
blog article detail page now uses SEO H1 as heading when available

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Blog articles on blog article detail page should use the SEO H1 title when available. This had been forgotten on FE and only implemented on the API. This PR fixes the bug by implementing it on FE as well.
|New feature| No
|Fixes issues|


project-base/storefront/components/Pages/BlogArticle/BlogArticleDetailContent.tsx
```diff
@@ -17,7 +17,7 @@ export const BlogArticleDetailContent: FC<BlogArticleDetailContentProps> = ({ bl
 
     return (
         <Webline>
-            <ArticleTitle dataTestId={TEST_IDENTIFIER + 'title'}>{blogArticle.name}</ArticleTitle>
+            <ArticleTitle dataTestId={TEST_IDENTIFIER + 'title'}>{blogArticle.seoH1 ?? blogArticle.name}</ArticleTitle>
```
tvikito, line 20: Better put this. It's because when `seoH1` would be for some reason empty string, so we are safe that this page will have some title.

```suggestion
            <ArticleTitle dataTestId={TEST_IDENTIFIER + 'title'}>{blogArticle.seoH1 || blogArticle.name}</ArticleTitle>
```

tvikito: But actually I am wondering if we really want to display seo title instead of name...? Are you sure of this change? 🤔

tvikito (approved):

pk16011990 (approved):

PR #2662, opened 2023-07-13
[shopsys] fixed split packages builds

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| On split packages we run checks and those checks were failing on 13.0 branch. This PR fixes that.
|New feature| Yes 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/ecs.php
+++ b/packages/framework/ecs.php
@@ -141,6 +141,7 @@ return static function (ECSConfig $ecsConfig): void {
         PropertyTypeHintSniff::class => [
             __DIR__ . '/tests/Unit/Component/ClassExtension/Source/*/*.php',
             __DIR__ . '/tests/Unit/Component/ClassExtension/Source/*.php',
+            ...explode("\n", trim(shell_exec("grep -rl '\\* @ORM\\\\Entity' " . escapeshellarg(__DIR__ . '/src')) ?? ''))
         ],
         PhpdocToPropertyTypeFixer::class => [
             __DIR__ . '/src/*',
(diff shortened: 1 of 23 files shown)


packages/framework/ecs.php
```diff
@@ -141,6 +141,7 @@
         PropertyTypeHintSniff::class => [
             __DIR__ . '/tests/Unit/Component/ClassExtension/Source/*/*.php',
             __DIR__ . '/tests/Unit/Component/ClassExtension/Source/*.php',
+            ...explode("\n", trim(shell_exec("grep -rl '\\* @ORM\\\\Entity' " . escapeshellarg(__DIR__ . '/src')) ?? ''))
```
TomasLudvik, line 144: I am not able to say what this does, do we really want it like that?

packages/framework/tests/Test/Provider/TestCurrencyProvider.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Tests\App\Functional\Model\Pricing\Currency;
+namespace Tests\FrameworkBundle\Test\Provider;
```
TomasLudvik, line 5: This change should be mentioned in UPGRADE, right?

packages/framework/ecs.php
grossmannmartin: I expected the comment on this line :D 

We cannot ignore `__DIR__ . '/src/*'` because individual configs are merged together when running from monorepo. So this check is then disabled completely for framework package.

The meaning of this line is: 
- find all files containing `* @ORM\Entity` in the `__DIR__ . '/src'` directory, convert the (string) output to an array, and all items from the array spread here.

I was not able to come up with a better solution.

packages/framework/tests/Test/Provider/TestCurrencyProvider.php
grossmannmartin: Added, thanks

TomasLudvik (approved):

pk16011990 (approved):

PR #2669, opened 2023-07-18
[shopsys] fix storefront file permissions for user with uid different from 1000

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Linux user with different UID than 1000 was not able to run storefront container in dev environment due to wrong ownership of files from the volume. This PR adds auto fixing of the permissions and removes no longer necessary chown in github actions builds.
|New feature| No 
|Fixes issues| ... 


TomasLudvik (approved):
Good job!

PR #2673, opened 2023-07-23
errors on SF can now have different verbosity

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We want to be able to differentiate between different error message verbosity in order to specify which errors should be showed to users, which logged in sentry, and which should be completely ignored. With this setting it's now possible.
|New feature| Yes
|Fixes issues| ...

### DIFF
--- a/project-base/storefront/helpers/errors/applicationErrors.ts
+++ b/project-base/storefront/helpers/errors/applicationErrors.ts
@@ -1,31 +1,55 @@
-export const ApplicationErrors = {
-    default: 'default',
-    'cart-not-found': 'cart-not-found',
-    'max-allowed-limit': 'max-allowed-limit',
-    'packetery-address-id-invalid': 'packetery-address-id-invalid',
-    'invalid-credentials': 'invalid-credentials',
-    'invalid-refresh-token': 'invalid-refresh-token',
-    'order-emails-not-sent': 'order-emails-not-sent',
-    'order-empty-cart': 'order-empty-cart',
-    'personal-data-request-type-invalid': 'personal-data-request-type-invalid',
-    'blog-category-not-found': 'blog-category-not-found',
-    'image-type-invalid': 'image-type-invalid',
-    'image-size-invalid': 'image-size-invalid',
-    'order-not-found': 'order-not-found',
-    'personal-data-hash-invalid': 'personal-data-hash-invalid',
-    'product-price-missing': 'product-price-missing',
-    'no-result-found-for-slug': 'no-result-found-for-slug',
-    'store-not-found': 'store-not-found',
-    'invalid-token': 'invalid-token',
-    'product-not-found': 'product-not-found',
-    'handling-with-logged-customer-comparison': 'handling-with-logged-customer-comparison',
-    'comparison-not-found': 'comparison-not-found',
-    'compared-item-not-found': 'compared-item-not-found',
-    'compared-item-already-exists': 'compared-item-already-exists',
-    'wishlist-not-found': 'wishlist-not-found',
-    'wishlist-item-already-exists': 'wishlist-item-already-exists',
-    'wishlist-item-not-found': 'wishlist-item-not-found',
-    'seo-page-not-found': 'seo-page-not-found',
+export type ApplicationErrorVerbosityLevel = 'flash-message' | 'no-flash-message' | 'no-log';
+
+const ApplicationErrors = {
+    default: 'flash-message',
+    'cart-not-found': 'flash-message',
+    'max-allowed-limit': 'flash-message',
+    'packetery-address-id-invalid': 'flash-message',
+    'invalid-credentials': 'flash-message',
+    'invalid-refresh-token': 'flash-message',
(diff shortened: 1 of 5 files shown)


project-base/storefront/helpers/errors/applicationErrors.ts
```diff
@@ -1,28 +1,50 @@
+export type ApplicationErrorVerbosityLevel = 'flash-message' | 'no-flash-message' | 'no-log';
+
 export const ApplicationErrors = {
```
tvikito, line 3: Since we have those nice typings everywhere, would be great to have types also for type of error.

```suggestion
export const ApplicationErrors: Record<string, ApplicationErrorVerbosityLevel > = {...}
```

project-base/storefront/helpers/errors/errorMessageMapper.ts
```diff
@@ -1,52 +1,36 @@
...
-        [ApplicationErrors['personal-data-hash-invalid']]: t('Invalid hash.'),
-        [ApplicationErrors['product-price-missing']]: t('Product price is missing.'),
-        [ApplicationErrors['store-not-found']]: t('Store not found.'),
-        [ApplicationErrors['product-not-found']]: t('Product not found.'),
-        [ApplicationErrors['handling-with-logged-customer-comparison']]: t('Product not found.'),
-        [ApplicationErrors['comparison-not-found']]: t('Comparison not found.'),
-        [ApplicationErrors['compared-item-not-found']]: t('Compared product not found.'),
-        [ApplicationErrors['compared-item-already-exists']]: t('Compared product is already compared.'),
+const getErrorMessageTranslationString = (errorCode: FlashMessageKeys, t: Translate): string | undefined => {
+    const ERROR_MESSAGES: Record<FlashMessageKeys, string | undefined> = {
```
tvikito, line 5: I guess we don't want the error to have `undefined` message

```suggestion
    const ERROR_MESSAGES: Record<FlashMessageKeys, string> = {
```

project-base/storefront/helpers/errors/errorMessageMapper.ts
```diff
@@ -1,52 +1,36 @@
...
     };
 
     return ERROR_MESSAGES[errorCode];
 };
 
-export const hasErrorMessage = (errorCode: string, t: Translate): boolean => {
-    return getErrorMessageTranslationString(errorCode as ApplicationErrorsWithoutIgnoredErrorsType, t) !== undefined;
-};
-
 export const getErrorMessage = (errorCode: string, t: Translate): string => {
```
tvikito, line 32:
```suggestion
export const getErrorMessage = (errorCode: FlashMessageKeys & NoFlashMessageKeys, t: Translate): string => {
```

project-base/storefront/helpers/errors/errorMessageMapper.ts
```diff
@@ -1,52 +1,36 @@
...
-    const translationString = getErrorMessageTranslationString(
-        errorCode as ApplicationErrorsWithoutIgnoredErrorsType,
-        t,
-    );
+    const translationString = getErrorMessageTranslationString(errorCode as FlashMessageKeys & NoFlashMessageKeys, t);
```
tvikito, line 33:
```suggestion
    const translationString = getErrorMessageTranslationString(errorCode, t);
```

project-base/storefront/helpers/errors/errorMessageMapper.ts
sebaholesz: Nice catch, fixed.

project-base/storefront/helpers/errors/errorMessageMapper.ts
sebaholesz: I think what you meant is `FlashMessageKeys | NoFlashMessageKeys`, because `FlashMessageKeys & NoFlashMessageKeys` is `never`

project-base/storefront/helpers/errors/errorMessageMapper.ts
sebaholesz: Removed.

project-base/storefront/helpers/errors/applicationErrors.ts
sebaholesz: I had this typed as you suggested, but after doing so, all three subtypes (`FlashMessageKeys`, `NoFlashMessageKeys`, `NoLogKeys`) are `never`. I dont know why. I could not fix it.

tvikito: Maybe it was some TS server error, but on my side it's working. 

image

tvikito: Also please remove `export` from this `ApplicationErrors` constant. It is used only internally in this file.

tvikito: I was curious so I tried but no success neither. So let's leave it since those typings for `getErrorMessageTranslationString` is great.

PR #2689, opened 2023-08-02
router.push and router.replace now navigate correctly between dynamic…

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In some places our approach to pushing and replacing the URL with next router was invalid. It caused unnecessary reloads and SSR calls. This PR fixes that issue.
|New feature| No
|Fixes issues| ...

### DIFF
--- a/project-base/storefront/hooks/useQueryParams.ts
+++ b/project-base/storefront/hooks/useQueryParams.ts
@@ -319,12 +319,26 @@ export const useQueryParams = () => {
         // remove queries which are not set or removed
         const filteredQueries = getFilteredQueries(queries);
 
+        const asPathname = router.asPath.split('?')[0];
+        const dynamicPageQueryKey = getDynamicPageQueryKey(router.pathname);
+
+        let filteredQueriesWithDynamicParam = filteredQueries;
+        if (dynamicPageQueryKey) {
+            filteredQueriesWithDynamicParam = {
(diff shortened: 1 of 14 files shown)


project-base/storefront/hooks/useQueryParams.ts
```diff
@@ -317,14 +317,23 @@ export const useQueryParams = () => {
...
         router[isPush ? 'push' : 'replace'](
             {
-                pathname: pathnameOverride || router.asPath.split('?')[0],
+                pathname: router.pathname,
                 query: filteredQueries,
             },
-            undefined,
+            getUrlFromPathnameAndQueries(pathnameOverride || asPathname, filteredQueries),
             { shallow: true },
         );
```
tvikito, line 338: After discussion it seems like we don't need to pass `filteredQueries` into those to arguments. `filteredQueries` are already passed alongside with `pathname`, so Next should already pass the queries.

This way we probably don't need this `getUrlFromPathnameAndQueries` util at all. Since on other places is it simply returning just pathname. If this idea is correct let's remove `getUrlFromPathnameAndQueries` and pass directly pathname everywhere this function is used.

sebaholesz: Refactored

tvikito (approved):

pk16011990 (approved):

PR #2695, opened 2023-08-03
load more feature

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| With this feature, products can be dynamically loaded and added to the previously loaded images.
|New feature| Yes
|Fixes issues| ... 

----
  - https://sh-fwcc-1246-load-more-feature.odin.shopsys.cloud
  - https://cz.sh-fwcc-1246-load-more-feature.odin.shopsys.cloud


project-base/storefront/components/Pages/BrandDetail/helpers.ts
```diff
@@ -0,0 +1,113 @@
...
+            currentPage,
+            currentLoadMore,
+            readBrandProductsFromCache,
+        );
+
+        const pageSize = previouslyQueriedProductsFromCache
+            ? DEFAULT_PAGE_SIZE
+            : DEFAULT_PAGE_SIZE * (currentLoadMore + 1);
+        const isMoreThanOnePage = pageSize > DEFAULT_PAGE_SIZE;
+        const endCursor = isMoreThanOnePage ? getEndCursor(currentPage) : getEndCursor(currentPage, currentLoadMore);
```
tvikito, line 68: I am not sure if this is better readable, but I will leave on you

```suggestion
        const endCursor = getEndCursor(currentPage, isMoreThanOnePage ? undefined :  currentLoadMore);
```

The same would be applied to other `helpers`.

project-base/storefront/components/Pages/BrandDetail/helpers.ts
```diff
@@ -0,0 +1,113 @@
...
+    endCursor: string,
+    pageSize: number,
+) =>
+    client.readQuery<BrandProductsQueryApi, BrandProductsQueryVariablesApi>(BrandProductsQueryDocumentApi, {
+        urlSlug,
+        orderingMode,
+        filter,
+        endCursor,
+        pageSize,
+    })?.data ?? undefined;
```
tvikito, line 113:
```suggestion
    })?.data || undefined;
```

More of philosophy than practical benefit, but if you would be deciding whether to use `??` or `||` the way I take it is that you don't want to let pass values such as empty string or boolean, in such a cases you rather want to return `undefined` so better to use `||`. You decide.

project-base/storefront/components/Pages/CategoryDetail/CategoryDetailProductsWrapper.tsx
```diff
@@ -46,76 +31,16 @@ export const CategoryDetailProductsWrapper: FC<CategoryDetailProps> = ({ categor
...
                 fetching={fetching}
+                loadMoreFetching={loadMoreFetching}
                 category={category}
                 gtmMessageOrigin={GtmMessageOriginType.other}
             />
             <Pagination
                 paginationScrollTargetRef={paginationScrollTargetRef}
                 totalCount={category.products.totalCount}
+                isWithLoadMore
+                hasNext={categoryProductsData?.products.pageInfo.hasNextPage}
```
tvikito, line 42: Wouldn't be better prop name?

```suggestion
                isWithNextPage={categoryProductsData?.products.pageInfo.hasNextPage}
```

project-base/storefront/components/Pages/BrandDetail/helpers.ts
```diff
@@ -0,0 +1,113 @@
...
+    const [brandProductsData, setBrandProductsData] = useState<undefined | BrandProductsQueryApi>(
+        readBrandProductsFromCache(
+            client,
+            urlSlug,
+            sort,
+            mappedFilter,
+            getEndCursor(currentPage),
+            DEFAULT_PAGE_SIZE * (currentLoadMore + 1),
+        ),
+    );
```
tvikito, line 41: If `readCategoryDetailFromCache` is typed properly, you don't need to put type here

```suggestion
    const [brandProductsData, setBrandProductsData] = useState(
        readBrandProductsFromCache(
            client,
            urlSlug,
            sort,
            mappedFilter,
            getEndCursor(currentPage),
            DEFAULT_PAGE_SIZE * (currentLoadMore + 1),
        ),
    );
```

The same would be applied to other `helpers`.

project-base/storefront/helpers/pagination/loadMore.ts
```diff
@@ -0,0 +1,75 @@
...
+        | FlagProductsQueryApi
+        | undefined,
+) => {
+    newProductsData.products = {
+        ...newProductsData.products,
+        edges: mergeProductEdges(previouslyQueriedProductsFromCache?.products.edges, newProductsData.products.edges),
+    };
+
+    return newProductsData;
+};
```
tvikito, line 27: You shouldn't modify existing data from outside of the function. But I have bigger idea, what about to cut `edges` here and simply return array of edges content? Since we don't need anything else besides `edges`, it would make sense to me.

Then we would have instead of this 
image

this
image

and you can remove this function

project-base/storefront/helpers/pagination/parseLoadMoreFromQuery.ts
```diff
@@ -0,0 +1,4 @@
+export const parseLoadMoreFromQuery = (query: string | string[] | undefined): number => {
+    const parsedLoadMore = Number(query);
+    return isNaN(parsedLoadMore) ? 0 : parsedLoadMore;
+};
```
tvikito, line 4: I guess this is mainly for remove undefined value and replace it with `0`. But cannot we simply use `undefined` in case you don't have loadmore?

project-base/storefront/hooks/gtm/productList/useGtmPaginatedProductListViewEvent.ts
```diff
@@ -11,17 +11,31 @@ export const useGtmPaginatedProductListViewEvent = (
...
+    const lastViewedStringifiedProducts = useRef<string>();
+    const { currentPage, currentLoadMore } = useQueryParams();
+    const previousLoadMoreRef = useRef(currentLoadMore);
     const { url } = useDomainConfig();
+    const stringifiedProducts = JSON.stringify(paginatedProducts);
 
     useEffect(() => {
-        if (paginatedProducts !== undefined && lastViewedProducts.current !== paginatedProducts) {
-            lastViewedProducts.current = paginatedProducts;
+        if (paginatedProducts !== undefined && lastViewedStringifiedProducts.current !== stringifiedProducts) {
```
tvikito, line 21:
```suggestion
        if (paginatedProducts && lastViewedStringifiedProducts.current !== stringifiedProducts) {
```

project-base/storefront/hooks/useQueryParams.ts
```diff
@@ -273,12 +280,22 @@ export const useQueryParams = () => {
...
 
+    const pushQueryLoadMore = (loadMore: number) => {
+        const newQuery: UrlQueries = {
+            ...query,
+            [LOAD_MORE_QUERY_PARAMETER_NAME]: loadMore > 0 ? loadMore.toString() : undefined,
+        } as const;
+
+        pushQueries(newQuery, true);
+    };
+
```
tvikito, line 298: Since this is used only on one place cannot we use it directly there? I don't see that this would improve anything, just splitting the code to another place.

```suggestion
const loadMore = () => {
        const newQuery: UrlQueries = {
            ...query,
            [LOAD_MORE_QUERY_PARAMETER_NAME]: loadMore > 0 ? loadMore.toString() : undefined,
        } as const;

        pushQueries(newQuery, true);
    };
```

project-base/storefront/components/Pages/BrandDetail/helpers.ts
sebaholesz: No problem :)

project-base/storefront/components/Pages/BrandDetail/helpers.ts
sebaholesz: No problem :)

project-base/storefront/components/Pages/CategoryDetail/CategoryDetailProductsWrapper.tsx
sebaholesz: Decided to rename it to `hasNextPage`

project-base/storefront/components/Pages/BrandDetail/helpers.ts
sebaholesz: Good catch, removed the types.

project-base/storefront/helpers/pagination/parseLoadMoreFromQuery.ts
sebaholesz: We apply arithmetic operations to the result of this function, so I would rather not.

project-base/storefront/hooks/gtm/productList/useGtmPaginatedProductListViewEvent.ts
sebaholesz: Done

project-base/storefront/hooks/useQueryParams.ts
sebaholesz: You're right, let's do it as you say.

project-base/storefront/helpers/pagination/loadMore.ts
sebaholesz: We discussed this on a call and I changed the behavior in https://github.com/shopsys/shopsys/pull/2695/commits/4665f68195be1d49c3197e2c4c2b249820388f40

tvikito: Cool! I think it was worth the time. Thanks for your effort! 🙏

tvikito (approved):

project-base/storefront/helpers/pagination/loadMore.ts
```diff
@@ -61,3 +66,55 @@ export const getPreviouslyQueriedProductsFromCache = (
...
+
+    const pageOffset = Math.ceil(loadedProductsDifference / pageSize);
+
+    return {
+        updatedPage: currentPage + pageOffset,
+        updatedLoadMore: currentLoadMore - pageOffset,
+    };
+};
+
+export const getRedirectWithOffsetPageIfTooManyLoadedProducts = (
```
tvikito: Cannot we shorten the names of those two functions? Those long names are counterproductive in my opinion. Since until I read function name I would rather simply open this function and maybe faster read code itself to understand what it does. Maybe `getRedirectWithOffsetPage` would do the job as well? And `getRedirectWithOffsetPage`? Since I don't have to know about how exactly is this offset or redirect calculated.

tvikito (approved):

project-base/storefront/helpers/pagination/loadMore.ts
sebaholesz: Renamed it to getOffsetPageAndLoadMore and getRedirectWithOffsetPage.

project-base/storefront/helpers/pagination/loadMore.ts
```diff
@@ -98,22 +101,23 @@ export const getRedirectWithOffsetPage = (
...
+    if (updatedQueries.updatedLoadMore > 0) {
+        updatedQuery[LOAD_MORE_QUERY_PARAMETER_NAME] = JSON.stringify(updatedQueries.updatedLoadMore);
+    } else {
+        delete updatedQuery[LOAD_MORE_QUERY_PARAMETER_NAME];
     }
 
     return {
         redirect: {
-            destination: updatedUrl,
+            destination: `${currentSlug}?${stringify(updatedQuery)}`,
```
tvikito: image

tvikito (changes requested):

project-base/storefront/helpers/pagination/loadMore.ts
```diff
@@ -122,23 +130,205 @@ export const getRedirectWithOffsetPage = (
...
+        if (!value) {
+            continue;
+        }
+
+        if (Array.isArray(value)) {
+            value.forEach((v) => searchParams.append(key, v));
+        } else {
+            searchParams.set(key, value);
+        }
+    }
```
tvikito, line 145: Niiice! I guess much better than use legacy library or an another library in general for such a small case. Thanks for your effort 🙏

tvikito (approved):

PR #2710, opened 2023-08-10
refactor Sliders

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| remove keen-slider and replace it with custom solution with CSS
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1259-refactor-sliders.odin.shopsys.cloud
  - https://cz.tv-fw-1259-refactor-sliders.odin.shopsys.cloud


project-base/storefront/components/Blocks/Banners/BannersSliderItem.tsx
```diff
@@ -1,35 +1,45 @@
...
 };
 
-export const BannersSliderItem: FC<BannersSliderItemProps> = ({
-    item: { webMainImage, mobileMainImage, link, name },
-}) => {
-    const { width } = useGetWindowSize();
-    const image = width > desktopFirstSizes.tablet ? webMainImage : mobileMainImage;
+export const BannersSliderItem = forwardRef<HTMLAnchorElement, BannersSliderItemProps>(
+    ({ item: { webMainImage, mobileMainImage, link, name } }, ref) => {
+        const { width } = useGetWindowSize();
```
sebaholesz: Do I understand it correctly that we are still depending on client-side JS here? If so, cannot we rewrite it to pure CSS?

project-base/storefront/public/svg/IconsSvg.tsx
```diff
@@ -365,8 +365,8 @@ export const HeartFull = () => (
 export const EmptyCart = () => (
     <svg width="572" height="512" viewBox="0 0 572 512" fill="none" xmlns="http://www.w3.org/2000/svg">
         <path
-            fill-rule="evenodd"
-            clip-rule="evenodd"
+            fillRule="evenodd"
+            clipRule="evenodd"
```
sebaholesz, line 369: Haha, thanks  for this :laughing: I know I F-ed up...

project-base/storefront/components/Blocks/Banners/BannersSlider.tsx
```diff
@@ -1,114 +1,116 @@
...
+            className="flex flex-col gap-6 vl:flex-row"
+            data-testid={dataTestId}
+            onMouseEnter={handleHoverEnter}
+            onMouseLeave={handleHoverLeave}
+            onTouchStart={handleHoverEnter}
+            onTouchEnd={handleHoverLeave}
+        >
             <div
+                className={twJoin(
+                    "grid h-80 snap-mandatory rounded-xl [-ms-overflow-style:'none'] [scrollbar-width:'none']",
```
sebaholesz: Not sure if these styles work as you'd expect them...
[Screencast from 11.8.2023 08:11:46.webm](https://github.com/shopsys/shopsys/assets/55181354/7e0a715c-78e7-44ff-9095-98801f6c3fbd)

project-base/storefront/components/Blocks/Banners/BannersSliderItem.tsx
```diff
@@ -1,35 +1,45 @@
...
-    const imageSize: ImageSizeFragmentApi | null = image?.sizes.find((i) => i.size === 'default') ?? null;
+        return (
+            <a href={link} className="h-80 min-w-full max-w-full basis-full snap-center" draggable={false} ref={ref}>
+                {width < 0 ? (
+                    <div className="flex h-full w-full items-center justify-center">
+                        <Loader className="w-14" />
+                    </div>
+                ) : (
+                    imageElement
+                )}
```
sebaholesz: I understand the logic with a spinner and I must say I like the UX of it, as it looks much better than a jumpy slider, but isn't it possible to handle it with just CSS?

sebaholesz (changes requested):

project-base/storefront/components/Blocks/Banners/BannersSliderItem.tsx
tvikito: Yup, JS is still needed. What exactly do you mean by pure CSS in this case of useGetWindowSize? Only CSS would be possible if we wouldn't have two types of images (mobile and desktop) to show. Otherwise it's messing the DOM.

project-base/storefront/components/Blocks/Banners/BannersSliderItem.tsx
tvikito: As I wrote in the previous comment, only possible when we have only one type (size) of the images.

project-base/storefront/components/Blocks/Banners/BannersSlider.tsx
tvikito: Yes, Firefox was buggy in some scenarios. If I've fixed it bug in Chrome appeared. So I decided to pull back to previous solution and use this refactoring there.

project-base/storefront/components/Blocks/Banners/BannersSlider.tsx
```diff
@@ -15,16 +17,16 @@ export const BannersSlider: FC<BannersSliderProps> = ({ sliderItems, dataTestId
     const [pause, setPause] = useState(false);
     const timer = useRef<NodeJS.Timer | null>(null);
     const sliderBoxRef = useRef<HTMLDivElement>(null);
+
+    const { width: windowWidth } = useGetWindowSize();
+    const isDesktop = windowWidth > desktopFirstSizes.tablet;
+    const isRecognizingWindowWidth = windowWidth < 0;
```
sebaholesz, line 23: I would say this name is incorrect, isn't it? If the width is below 0, meaning -1, that's when the window width is not recognized, right? Shouldn't it be named the other way around?

sebaholesz (changes requested):

project-base/storefront/components/Blocks/Banners/BannersSlider.tsx
tvikito: That's why it is in present continuous with "ing" which is state exactly for what is needed - state when window width is not recognized yet, but still recognizing. Are we on the same page now?

sebaholesz (approved):

PR #2714, opened 2023-08-11
[project-base] changed robots.txt datafixture to avoid encouraging inappropriate practices

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| /admin should not be mentioned in the robots.txt because it reveals the url of administration to everyone. Then it makes no sense to change the admin url
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-better-robots-datafixture.odin.shopsys.cloud
  - https://cz.mg-better-robots-datafixture.odin.shopsys.cloud

### DIFF
--- a/project-base/app/src/DataFixtures/Demo/SettingValueDataFixture.php
+++ b/project-base/app/src/DataFixtures/Demo/SettingValueDataFixture.php
@@ -127,7 +127,7 @@ class SettingValueDataFixture extends AbstractReferenceFixture implements Depend
             );
             $this->setting->setForDomain(
                 SeoSettingFacade::SEO_ROBOTS_TXT_CONTENT,
-                'Disallow: /admin',
+                'Disallow: *?filter=',
                 $domainId,
             );
             $this->setting->setForDomain(


project-base/app/src/DataFixtures/Demo/SettingValueDataFixture.php
```diff
@@ -127,7 +127,7 @@ public function load(ObjectManager $manager)
             );
             $this->setting->setForDomain(
                 SeoSettingFacade::SEO_ROBOTS_TXT_CONTENT,
-                'Disallow: /admin',
+                'Disallow: *?filter=',
```
TomasLudvik, line 130: This way admin can now be revealed to the public right? Would not be better to use something like:
/admin*

TomasLudvik (changes requested):

project-base/app/src/DataFixtures/Demo/SettingValueDataFixture.php
grossmannmartin: actually, no... robots.txt is merely a recommendation. 
so having the URL of the administration here actually exposes the URL to the world (to anyone, who opens this file). 
From this point of view, I think the URL of the administration must not be in the robots.txt file at all. 

Besides, the /admin route itself already contains the proper `robots` meta tag (`<meta name="robots" content="noindex, nofollow">`)

_I think robots.txt should be renamed to hacker_must_visit.txt_

TomasLudvik (approved):

PR #2716, opened 2023-08-14
[SSP-1579] - last visited products

TomasGottvald:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Adding last visited functionality.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tg-fw-1579-last-visited.odin.shopsys.cloud
  - https://cz.tg-fw-1579-last-visited.odin.shopsys.cloud


project-base/storefront/components/Layout/CommonLayout.tsx
```diff
@@ -22,6 +24,7 @@ export const CommonLayout: FC<CommonLayoutProps> = ({ children, description, tit
         </Webline>
         <Adverts positionName="header" withGapBottom withWebline />
         {children}
+        {showLastVisitedProducts && <LastVisitedProducts />}
```
sebaholesz: Do we want to have it as a part of the CommonLayout? I think it would be more suitable to add it in pages where it's desired. :thinking:

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
```diff
@@ -24,6 +24,7 @@ import { useRouter } from 'next/router';
 import { Fragment, useRef } from 'react';
 import { ProductWishlistButton } from 'components/Blocks/Product/ButtonsAction/ProductWishlistButton';
 import { useWishlist } from 'hooks/useWishlist';
+import { SetLastVisitedProductCatalogNumbers } from 'helpers/products/LastVisitedProducts';
```
sebaholesz: Both `SetLastVisitedProductCatalogNumbers` and `LastVisitedProducts` should be camelCase, not PascalCase.

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
```diff
@@ -41,6 +42,8 @@ export const ProductDetailContent: FC<ProductDetailContentProps> = ({ product, f
 
     useGtmProductDetailViewEvent(product, getUrlWithoutGetParameters(router.asPath), fetching);
 
+    SetLastVisitedProductCatalogNumbers(product.catalogNumber);
```
sebaholesz: If written like this, we call it on every render and re-render, which is definitely not what we want. The goal is to call it once, ideally inside a one-time useEffect...

project-base/storefront/helpers/products/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,34 @@
+import { usePersistStore } from 'store/zustand/usePersistStore';
+
+const LAST_VISITED_MAX_ITEMS = 20;
+
+export const GetLastVisitedProductCatalogNumbers = (): string[] => {
```
sebaholesz: This should be camelCase, not PascalCase.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,34 @@
...
+    const lastVisitedProducts = usePersistStore((store) => store.lastVisitedProducts);
```
sebaholesz: You should not call hooks inside normal functions. Either change it to a hook, or find another way around.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return lastVisitedProducts ? JSON.parse(lastVisitedProducts) : [];
+};
+
+export const SetLastVisitedProductCatalogNumbers = (newLastVisitedItemCatalogNumber: string): void => {
+    let olderItemCatalogNumbers = GetLastVisitedProductCatalogNumbers();
+
+    const updateUserState = usePersistStore((store) => store.updateUserState);
```
sebaholesz: Again, you should not call hooks inside normal functions.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,34 @@
...
+        (oldItemCatalogNumber) => oldItemCatalogNumber === newLastVisitedItemCatalogNumber,
+    );
+
+    if (!alreadyAdded) {
+        olderItemCatalogNumbers = [newLastVisitedItemCatalogNumber, ...olderItemCatalogNumbers];
+    } else if (olderItemCatalogNumbers.length > 0) {
+        olderItemCatalogNumbers = [
+            newLastVisitedItemCatalogNumber,
+            ...olderItemCatalogNumbers.filter(
+                (oldItemCatalogNumber) => oldItemCatalogNumber !== newLastVisitedItemCatalogNumber,
```
sebaholesz: This is not compliant with the AC:

_pokud uživatel znovu přijde na produkt, který je již na tomto seznamu, nepřidává se duplicitně a jeho pořadí se nemění_

project-base/storefront/helpers/products/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,34 @@
...
+            ),
+        ];
+    }
+
+    const lastVisitedProductCatalogNumbers = JSON.stringify(olderItemCatalogNumbers.slice(0, LAST_VISITED_MAX_ITEMS));
```
sebaholesz: What do we gain by parsing it from and to string?

project-base/storefront/helpers/products/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return lastVisitedProducts ? JSON.parse(lastVisitedProducts) : [];
+};
+
+export const SetLastVisitedProductCatalogNumbers = (newLastVisitedItemCatalogNumber: string): void => {
+    let olderItemCatalogNumbers = GetLastVisitedProductCatalogNumbers();
+
+    const updateUserState = usePersistStore((store) => store.updateUserState);
+    const alreadyAdded = olderItemCatalogNumbers.some(
+        (oldItemCatalogNumber) => oldItemCatalogNumber === newLastVisitedItemCatalogNumber,
+    );
```
sebaholesz: If it's just an array of strings, you can use `includes`, as it is more semantically clear and the performance should also be a tiny bit better, though this is negligible for sure in this case.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    if (!alreadyAdded) {
+        olderItemCatalogNumbers = [newLastVisitedItemCatalogNumber, ...olderItemCatalogNumbers];
```
sebaholesz: Pushing does not work in this scenario? Have you tried?

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ProductsSlider } from 'components/Blocks/Product/ProductsSlider';
+import { GetLastVisitedProductCatalogNumbers } from 'helpers/products/LastVisitedProducts';
+import { Webline } from 'components/Layout/Webline/Webline';
+import { useTypedTranslationFunction } from 'hooks/typescript/useTypedTranslationFunction';
+import { GtmProductListNameType } from 'types/gtm/enums';
+import { useProductsByCatnumsApi } from 'graphql/generated';
+
+const LastVisitedProducts: FC = () => {
+    const lastVisitedProductCatalogNumbers = GetLastVisitedProductCatalogNumbers();
+    const [result] = useProductsByCatnumsApi({ variables: { catnums: lastVisitedProductCatalogNumbers } });
```
sebaholesz: I don't see this query on server-side. This means it's only loaded on the client. I understand it might not be relevant to crawlers and robots, but the loading state should therefore be handled with a skeleton or similar.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,29 @@
...
+                <Heading type="h2">{t('Last visited products')}</Heading>
+                <ProductsSlider
+                    products={lastVisitedProducts}
+                    gtmProductListName={GtmProductListNameType.last_visited_products}
+                />
+            </Webline>
+        );
+    }
+
+    return null;
```
sebaholesz: I would say turning this condition around is more readable and is the standard we are trying to follow.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export default LastVisitedProducts;
```
sebaholesz: Please, rewrite this to a named export

sebaholesz (changes requested):
Hey :wave: Thanks for the PR. I have added some comments and change requests. Please, check them out and let me know what you think. Good luck.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,29 @@
+import { Heading } from 'components/Basic/Heading/Heading';
+import { ProductsSlider } from 'components/Blocks/Product/ProductsSlider';
+import { GetLastVisitedProductCatalogNumbers } from 'helpers/products/LastVisitedProducts';
+import { Webline } from 'components/Layout/Webline/Webline';
+import { useTypedTranslationFunction } from 'hooks/typescript/useTypedTranslationFunction';
+import { GtmProductListNameType } from 'types/gtm/enums';
+import { useProductsByCatnumsApi } from 'graphql/generated';
+
+const LastVisitedProducts: FC = () => {
+    const lastVisitedProductCatalogNumbers = GetLastVisitedProductCatalogNumbers();
```
tvikito: Typo with capital letter...?

```suggestion
    const lastVisitedProductCatalogNumbers = getLastVisitedProductCatalogNumbers();
```

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const [result] = useProductsByCatnumsApi({ variables: { catnums: lastVisitedProductCatalogNumbers } });
+    const lastVisitedProducts = result.data?.productsByCatnums ?? [];
+    const t = useTypedTranslationFunction();
+    if (lastVisitedProducts.length !== 0) {
```
tvikito: Please put empty line before if

project-base/storefront/public/locales/sk/common.json
```diff
@@ -159,6 +159,7 @@
     "Item {{ itemName }} has been sold out.": "Item {{ itemName }} has been sold out.",
     "Last name": "Priezvisko",
     "Last name must be at most {{ max }} characters": "",
+    "Last visited products": "",
```
tvikito: Maybe we can fill also this translation?

project-base/storefront/types/gtm/enums.ts
```diff
@@ -40,6 +40,7 @@ export enum GtmProductListNameType {
     autocomplete_search_results = 'autocomplete_search_results',
     wishlist = 'wishlist',
     sharedWishlist = 'sharedWishlist',
+    last_visited_products = 'last visited products',
```
tvikito: Do we have this enum implemented anywhere? I don't see it, if it is so, please remove this and let's add it once it needed.

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
TomasGottvald: Fixed, setLastVisited has been moved to useLastVisited.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
TomasGottvald: Camelcase fixed by moving to useLastVisited.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
TomasGottvald: True, moved to useLastVisited.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Fixed

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Fixed by moving it to useLastVisited, bcs it cant be called like this.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: added

project-base/storefront/public/locales/sk/common.json
TomasGottvald: A lot of translations is missing, but filled.

project-base/storefront/types/gtm/enums.ts
TomasGottvald: I think it is better to use is unified this way. Next developer can use it without searching if it is already added.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
TomasGottvald: True fixed, in very old version it was array of all information. Then changed to catnum only.

project-base/storefront/helpers/products/LastVisitedProducts.tsx
TomasGottvald: True fixed, in very old version it was array of all information. Then changed to catnum only.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Hmmm... but when there will be no last visited... then it will jump from skeleton to "no content". I will discuss it with other developers

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: I dont know what you mean ); Please give me more info (;

project-base/storefront/components/Layout/CommonLayout.tsx
TomasGottvald: I thought it will be easier to switch on/off by one prop. But ok, changed (;

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
sebaholesz: Fixed

project-base/storefront/helpers/products/LastVisitedProducts.tsx
sebaholesz: Fixed

project-base/storefront/helpers/products/LastVisitedProducts.tsx
sebaholesz: Fixed

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
sebaholesz: I dont think its about discussing it with devs 🤔 Its a business decision. Either we want a loader or we dont. We are able to implement it gracefully.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
sebaholesz: Fixed, check 2 last commits

project-base/storefront/types/gtm/enums.ts
sebaholesz: It is used, I am resolving this thread

sebaholesz (approved):

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: We have discussed it. First try was, let it without loader. Second was, lets save hasLastVisited:boolean to cookies. Then check and if it is true, show loader. This approach should avoid content jumping (;

packages/framework/src/Model/Product/ProductElasticsearchProvider.php
```diff
@@ -76,4 +76,34 @@ public function getSellableProductArrayByUuids(array $productUuids): array
             $this->filterQueryFactory->createSellableProductsByProductUuidsFilter($productUuids),
         );
     }
+
+    /**
+     * @param int[][] $productsIds
+     * @return array
+     */
+    public function getBatchedVisibleByProductIds(array $productsIds): array
```
TomasLudvik: Batching is used only in frontend API, so I think it should not be present in framework. This applies for all batching methods introduced to framework

project-base/app/src/Model/Product/ProductRepository.php
```diff
@@ -371,13 +371,15 @@ protected function applyOrdering(
...
             ->from(Product::class, 'p')
             ->where('p.catnum IN (:catnums)')
             ->setParameter('catnums', $catnums)
             ->getQuery()
             ->getScalarResult();
 
-        return array_column($result, 'id');
+        $catnumToIdMap = array_column($result, 'id', 'catnum');
+
+        return array_values(array_filter(array_map(static fn (string $catnum) => $catnumToIdMap[$catnum] ?? null, $catnums)));
```
TomasLudvik: This is really badly readable, can you refactor it?

UPGRADE-14.0.md
```diff
@@ -113,6 +113,18 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
...
+    ```diff
+        public function __construct(
+            protected readonly ProductFacade $productFacade,
+            protected readonly ProductFilterFacade $productFilterFacade,
+            protected readonly ProductConnectionFactory $productConnectionFactory,
+    +       protected readonly DataLoaderInterface $productsVisibleAndSortedByIdsBatchLoader,
+    +       protected readonly ProductRepository $productRepository,
+        )
+    ```
+    - see #project-base-diff to update your project
```
TomasLudvik: There are multiple changes in framework, that are not reflected here.

TomasLudvik (changes requested):

project-base/app/src/Model/Product/ProductRepository.php
vitek-rostislav: b6194ee40 - is it better now? :slightly_smiling_face:

UPGRADE-14.0.md
vitek-rostislav: yeah, however, the rest of the changes are from the cherry-picked commit - I will add the upgrade notes for this commit in its PR (https://github.com/shopsys/shopsys/pull/2901) - I hope it is ok :innocent:

packages/framework/src/Model/Product/ProductElasticsearchProvider.php
vitek-rostislav: that was not easy at all :smile: 3d46bd6a8

project-base/app/src/Model/Product/ProductRepository.php
vitek-rostislav: ...and second iteration of refactoring :smile: 2865179aa

UPGRADE-14.0.md
TomasLudvik: This is very confusing - if you have cherry-picked it here and it will be committed with this change, I think it should be added here, as someone can update to this change only.

UPGRADE-14.0.md
```diff
@@ -113,6 +113,18 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   see #project-base-diff to update your project
 -   replace custom application bootstrapping with symfony/runtime ([#2914](https://github.com/shopsys/shopsys/pull/2914))
     -   see #project-base-diff to update your project
+-   move productsByCatnums endpoint from project-base to frontend-api package ([#2716](https://github.com/shopsys/shopsys/pull/2716))
+    -   constructor of `Shopsys\FrontendApiBundle\Model\Resolver\Products\ProductsQuery` has cahnged:
```
TomasLudvik:
```suggestion
    -   constructor of `Shopsys\FrontendApiBundle\Model\Resolver\Products\ProductsQuery` has changed:
```

TomasLudvik (changes requested):

UPGRADE-14.0.md
vitek-rostislav: ok, makes sense, after refactoring the logic, there was just one thing left for mentioning in the upgrade notes, added in 58a387fe2

TomasLudvik (approved):

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Fixed now we store info about last visited products in cookies again.

TomasLudvik (approved):

project-base/storefront/package.json
```diff
@@ -45,6 +45,7 @@
         "next-translate": "2.0.0",
         "next-translate-plugin": "2.0.0",
         "next-urql": "5.0.2",
+        "nookies": "^2.5.2",
```
tvikito: We don't use `nookies` but `cookies-next`. Let's remove this.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    const lastVisitedProducts = result?.productsByCatnums ?? [];
+    const { t } = useTranslation();
+
+    if (fetching && lastVisitedProductsFromCookies.length) {
+        return <SkeletonModuleLastVisitedProducts />;
+    }
+
+    if (lastVisitedProducts.length !== 0) {
```
tvikito: We've already agreed we will use shorten way 

```suggestion
    if (lastVisitedProducts.length) {
```

project-base/storefront/helpers/products/lastVisitedProducts.tsx
tvikito, line 1: Since this helper is related only for the module, let's move this code into `helper.ts` and place it next to the module itself.

project-base/storefront/helpers/products/lastVisitedProducts.tsx
```diff
@@ -0,0 +1,45 @@
+import nookies from 'nookies';
+
+const LAST_VISITED_MAX_ITEMS = 20;
+
+export const useLastVisited = () => {
+    const getLastVisitedProductCatalogNumbers = (): string[] => {
+        const cookies = nookies.get();
+        if ('lastVisitedProducts' in cookies) {
```
tvikito: Space before missing

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
```diff
@@ -40,9 +41,12 @@ export const ProductDetailContent: FC<ProductDetailContentProps> = ({ product, f
     const { isProductInComparison, toggleProductInComparison, isPopupCompareOpen, setIsPopupCompareOpen } =
         useComparison();
     const { toggleProductInWishlist, isProductInWishlist } = useWishlist();
+    const { setLastVisitedProductCatalogNumbers } = useLastVisited();
 
     useGtmProductDetailViewEvent(product, getUrlWithoutGetParameters(router.asPath), fetching);
 
+    setLastVisitedProductCatalogNumbers(product.catalogNumber);
+
```
tvikito: No way. You can't call function like this 🙈 You must use useEffect for this. Otherwise it will run with every rerender.

project-base/storefront/pages/products/[productSlug].tsx
```diff
@@ -46,6 +47,7 @@ const ProductDetailPage: NextPage = () => {
             {product?.__typename === 'MainVariant' && (
                 <ProductDetailMainVariantContent fetching={fetching} product={product} />
             )}
+            <LastVisitedProducts removeCatalogNumber={product?.catalogNumber} />
```
Comment (tvikito) [inline project-base/storefront/pages/products/[productSlug].tsx]: Why don't we handle also main variant? 🤔

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
```diff
@@ -40,9 +41,12 @@ export const ProductDetailContent: FC<ProductDetailContentProps> = ({ product, f
     const { isProductInComparison, toggleProductInComparison, isPopupCompareOpen, setIsPopupCompareOpen } =
         useComparison();
     const { toggleProductInWishlist, isProductInWishlist } = useWishlist();
+    const { setLastVisitedProductCatalogNumbers } = useLastVisited();
```
tvikito: Why are we using this hook here and not in root of the page (`[productSlug].tsx`)? So it can handle also main variant.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const LastVisitedProducts: FC<LastVisitedProductsProps> = ({ removeCatalogNumber }) => {
+    const { getLastVisitedProductCatalogNumbers } = useLastVisited();
+    let lastVisitedProductsFromCookies = getLastVisitedProductCatalogNumbers();
+
+    if (removeCatalogNumber && !!lastVisitedProductsFromCookies.length) {
+        lastVisitedProductsFromCookies = lastVisitedProductsFromCookies.filter(
+            (item: string) => item !== removeCatalogNumber,
+        );
+    }
```
tvikito: You cannot handle code in a component like regular function... 🙈🙈 This needs to be handled properly by `useEffect` and `useState`. But in this case this is not good idea anyway. You can wrap this filter into useMemo and return final array.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    const [{ data: result, fetching }] = useProductsByCatnumsApi({
+        variables: { catnums: lastVisitedProductsFromCookies },
+    });
```
tvikito: I see the API is called even though we don't have any lastVisitedProducts. Let's fix this.

tvikito (changes requested):

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,49 @@
+import { ProductsSlider } from 'components/Blocks/Product/ProductsSlider';
```
tvikito, line 1: The app is broken because of this feature ❌

image

project-base/storefront/package.json
TomasGottvald: New for me, fixed

project-base/storefront/helpers/products/lastVisitedProducts.tsx
TomasGottvald: Fixed

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Fixed

project-base/storefront/helpers/products/lastVisitedProducts.tsx
TomasGottvald: Fixed.

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
TomasGottvald: Good point, I forgot do that.

project-base/storefront/pages/products/[productSlug].tsx
Comment (TomasGottvald) [reply project-base/storefront/pages/products/[productSlug].tsx]: I dont understand this part. It is part of Common Layout on bottom section. MainVariant content and Regular product content is above. I dont think this is restricted for any type of product.

Comment (TomasGottvald) [reply project-base/storefront/pages/products/[productSlug].tsx]: Maybe I got it. Last visited setter moved to productSlug, so it will set cookies for all product types.

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
TomasGottvald: Maybe I got it. Last visited setter moved to productSlug, so it will set cookies for all product types.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: True, it is better solution. Fixed.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Added pause param.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Fixed

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const [{ data: result, fetching }] = useProductsByCatnumsApi({
+        variables: { catnums: reducedVisitedProductsCatnums },
+        pause: !allLastVisitedProductsFromCookies,
+    });
+
+    const lastVisitedProducts = result?.productsByCatnums ?? [];
+    const { t } = useTranslation();
+
+    if (fetching && !!lastVisitedProductsFromCookies?.length) {
+        return <SkeletonModuleLastVisitedProducts />;
```
tvikito, line 32: Can we put proper layout for this skeleton? So it is not wider then it actually is. 

image

project-base/storefront/helpers/serverSide/initServerSideProps.ts
```diff
@@ -149,6 +153,7 @@ export const initServerSideProps = async ({
             // JSON.parse(JSON.stringify()) fix of https://github.com/vercel/next.js/issues/11993
             urqlState: JSON.parse(JSON.stringify(currentSsrCache.extractData())),
             isMaintenance,
+            lastVisitedProductsFromCookies,
```
tvikito: We can discuss this, but I don't like that currently this is logic runs even on the pages which don't need to handle last visited products. I would like to make some generic solution. For example return parsed cookies from here and handle data in the place where they need to be used.

project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx
```diff
@@ -13,13 +14,13 @@ import { isRedirectedFromSsr } from 'helpers/isRedirectedFromSsr';
...
 import { getServerSidePropsWrapper } from 'helpers/serverSide/getServerSidePropsWrapper';
-import { initServerSideProps } from 'helpers/serverSide/initServerSideProps';
+import { ServerSidePropsType, initServerSideProps } from 'helpers/serverSide/initServerSideProps';
 import { NextPage } from 'next';
 import { useRouter } from 'next/router';
 import { OperationResult } from 'urql';
 import { createClient } from 'urql/createClient';
 
-const BlogArticleDetailPage: NextPage = () => {
+const BlogArticleDetailPage: NextPage<ServerSidePropsType> = ({ lastVisitedProductsFromCookies }) => {
```
Comment (tvikito) [inline project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx]: Here you would extract only `{cookies}` and send needed information to `LastVisitedProducts` component. Then you can use `getLastVisitedProductCatalogNumbers` there. This way we would have page/component specific logic in this specific place.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedHelpers.ts
```diff
@@ -0,0 +1,47 @@
+import { getCookie, setCookie } from 'cookies-next';
+import { IncomingMessage, ServerResponse } from 'http';
+
+const LAST_VISITED_MAX_ITEMS = 20;
```
tvikito: We currently display only 4 items. Why do we have 20? Let's save only 4 according to current UI. Projects can easily adjust according to their needs afterward.

project-base/storefront/pages/products/[productSlug].tsx
```diff
@@ -75,6 +87,17 @@ export const getServerSideProps = getServerSidePropsWrapper(
                     };
                 }
 
+                if (
```
Comment (tvikito) [inline project-base/storefront/pages/products/[productSlug].tsx]: Why do you need to set this during SSR? Especially when it is set in browser again...? 🤔

Comment (TomasGottvald) [reply project-base/storefront/pages/products/[productSlug].tsx]: Finally deleted in your commit. I thing it was mistake when i was fighting with ssr and csr hydrating.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedHelpers.ts
TomasGottvald: Fixed.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedProducts.tsx
TomasGottvald: Fixed 

obrazek

project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx
Comment (TomasGottvald) [reply project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx]: Changed to cookies, then cookies.lastVisitedProductsFromCookies sent to component.

project-base/storefront/helpers/serverSide/initServerSideProps.ts
TomasGottvald: Changed to return cookies - at first time I thought about it in first time and have in mind than we dont want to use cookies so much, so decided to return only needed ones. Finally changed, returning all cookies and parsed on place when we need it.

project-base/storefront/components/Blocks/Product/LastVisitedProducts/LastVisitedHelpers.ts
```diff
@@ -15,7 +15,10 @@ export const getLastVisitedProductCatalogNumbers = (context?: GetServerSideProps
 
 export const setLastVisitedProductCatalogNumbers = (newLastVisitedProductsCatalogNumber: string): void => {
     const currentLastVisitedProductsCatalogNumbers = getLastVisitedProductCatalogNumbers() || [];
-    currentLastVisitedProductsCatalogNumbers.unshift(newLastVisitedProductsCatalogNumber);
+
+    if (currentLastVisitedProductsCatalogNumbers.indexOf(newLastVisitedProductsCatalogNumber) === -1) {
```
tvikito: What about 

`if (currentLastVisitedProductsCatalogNumbers.includes(newLastVisitedProductsCatalogNumber)) {
        currentLastVisitedProductsCatalogNumbers.unshift(newLastVisitedProductsCatalogNumber);
    }`

I would consider it as better readable.

TomasGottvald: Yes it is little bit better. Fixed.

tvikito (approved):

PR #2717, opened 2023-08-14
storefront restructuring

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There were some issues with the folder and code structure of storefront. Some were handled in this PR. For more details check individual commit messages.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-fw-1547-sf-restructuring.odin.shopsys.cloud
  - https://cz.sh-fw-1547-sf-restructuring.odin.shopsys.cloud


project-base/storefront/components/Basic/Flag/Flag.tsx
```diff
@@ -1,4 +1,4 @@
-import { twMergeCustom } from 'utils/twMerge';
+import { twMergeCustom } from 'helpers/twMerge';
```
tvikito, line 1: Now when I am seeing this, I wonder if it would be better rather use `utils` folder instead `helpers`. Since `utils` seems more generic where we can put everything, not just `helpers`, but vice versa seems to me like not every utility which we need is considered as helper. Though `helpers` and `utils` are almost synonyms. But just something to thing about twice to be sure this is correct step.

project-base/storefront/components/Blocks/Product/ButtonsAction/ProductComparePopup.tsx
```diff
@@ -13,15 +13,15 @@ type ProductComparePopupProps = {
 export const ProductComparePopup: FC<ProductComparePopupProps> = ({ onCloseCallback }) => {
     const t = useTypedTranslationFunction();
     const { url } = useDomainConfig();
-    const [productsComparisonUrl] = getInternationalizedStaticUrls(['/products-comparison'], url);
+    const [productComparisonUrl] = getInternationalizedStaticUrls(['/product-comparison'], url);
```
tvikito, line 16: To me it seems like it is comparing products not single product, so I am wondering why it needed to be changed? 🤔

sebaholesz: I would refer to Cambridge Dictionary:
https://dictionary.cambridge.org/dictionary/english/product-comparison

Furthermore, I apply the same logic as with other groups of items, such as:

- product list - list of products
- product category - category of products
etc.

project-base/storefront/components/Basic/Flag/Flag.tsx
sebaholesz: I think this is an unnecessary change which will only cause conflicts. If we decide to do it after all, we can do it separately

project-base/storefront/config/staticRewritePaths.js
```diff
@@ -25,7 +25,7 @@ const STATIC_REWRITE_PATHS = {
         '/cookie-consent': '/cookie-consent',
         '/abandoned-cart/:cartUuid': '/abandoned-cart/:cartUuid',
         '/grapesjs-template': '/grapesjs-template',
-        '/products-comparison': '/products-comparison',
+        '/product-comparison': '/product-comparison',
```
TomasGottvald, line 28: is this change necessary? I think that products - is even better - bcs we compare more than one product. This change only have more cons then pluses.

sebaholesz: Answered here:
https://github.com/shopsys/shopsys/pull/2717#discussion_r1293280420

PR #2727, opened 2023-08-16
added iframe and image to grapejs

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Added iframe and image to grapejs custom blocks.
|New feature| Yes
|Fixes issues| ... 

----
  - https://sh-grapejs-images-and-iframes.odin.shopsys.cloud
  - https://cz.sh-grapejs-images-and-iframes.odin.shopsys.cloud


project-base/storefront/components/Basic/UserText/GrapeJsParser.tsx
```diff
@@ -1,48 +1,60 @@
...
+import { GtmMessageOriginType, GtmProductListNameType } from 'gtm/types/enums';
+import { parseCatnums } from 'helpers/parsing/grapesJsParser';
 import { replaceAll } from 'helpers/replaceAll';
 import { memo } from 'react';
-import { GtmProductListNameType } from 'gtm/types/enums';
-import { GJS_PRODUCTS_SEPARATOR, parseCatnums } from 'helpers/parsing/grapesJsParser';
+import { UserText } from './UserText';
 
-type GrapesJsParserProps = {
+type GrapeJsParserProps = {
```
tvikito: In the beginning I was mistaken as well, but library is called GrapesJS not GrapeJS https://grapesjs.com/

project-base/storefront/components/Basic/UserText/GrapeJsParser.tsx
```diff
@@ -1,48 +1,60 @@
...
+        }
+
+        const products = replaceAll(part, /\[gjc-comp-ProductList&#61;|\]/g, '')
+            .split(',')
+            .map((product) =>
+                allProductsResponse.data?.productsByCatnums.find(
+                    (blogArticleProduct) => blogArticleProduct.catalogNumber === product,
+                ),
+            )
+            .filter(Boolean) as ListedProductFragmentApi[];
```
tvikito: Sometimes it can help to do type assertion, but in this case it seems to me like there is better assertion to do. Maybe better can be to put non-null assertion to the values which we are sure we have.

```suggestion
        const products = replaceAll(part, /\[gjc-comp-ProductList&#61;|\]/g, '')
            .split(',')
            .map((product) =>
                allProductsResponse.data!.productsByCatnums.find(
                    (blogArticleProduct) => blogArticleProduct.catalogNumber === product,
                )!,
            )
            .filter(Boolean);
```

This way you won't loose exact return type and only mark data which you will have for sure. `allProductsResponse.data` is checked in previous lines and `.find` will always find value since you are data are based on the same list.

tvikito (changes requested):

project-base/storefront/components/Basic/UserText/GrapeJsParser.tsx
sebaholesz: Without it, the code does not understand the type :cry: However, with a little rewrite, the assert is not necessary anymore.

project-base/storefront/components/Basic/UserText/GrapeJsParser.tsx
sebaholesz: Thanks for that... I was too lazy to check it myself :laughing:

project-base/storefront/components/Basic/UserText/GrapeJsParser.tsx
tvikito: The way I wrote my suggestion it was working, but nevermind. Nice that it's working this way.

tvikito (approved):

pk16011990 (approved):

PR #2730, opened 2023-08-17
added block with bestsellers to category detail

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Added block with bestsellers to category detail. This allows the users to see the most sold products from a specific category.
|New feature| Yes
|Fixes issues| ... 

----
  - https://sh-category-bestsellers.odin.shopsys.cloud
  - https://cz.sh-category-bestsellers.odin.shopsys.cloud


project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellersListItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+    product,
+    gtmProductListName,
+    listIndex,
+}) => {
+    const formatPrice = useFormatPrice();
+    const { url } = useDomainConfig();
+
+    return (
+        <div className="flex flex-wrap items-center gap-y-4 border-t border-greyLight py-4 first-of-type:border-0 lg:flex-nowrap lg:gap-5">
+            <div className="flex w-full items-center lg:w-7/12">
```
tvikito: With flexbox it's better to use `basis`.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
```diff
@@ -0,0 +1,63 @@
...
+                                      count: products.length - NUMBER_OF_VISIBLE_ITEMS,
+                                  })
+                                : t('View the next {{ count }} best sellers', {
+                                      count: products.length - NUMBER_OF_VISIBLE_ITEMS,
+                                  })}
+                        </span>
+                    </Button>
+                </div>
+            )}{' '}
+        </div>
```
tvikito: Let's put some spacing here so the sections are better visible.

```suggestion
        <div className="mb-8">
            <div className="mb-3 break-words font-bold text-dark lg:text-lg">
                {t('Do not want to choose? Choose certainty')}
            </div>
            
            <div className="mb-5">
                {products.map((product, index) => {
                    return index <= NUMBER_OF_VISIBLE_ITEMS - 1 || isCollapsed ? (
                        <CategoryBestsellersListItem
                            key={product.uuid}
                            product={product}
                            gtmProductListName={GtmProductListNameType.bestsellers}
                            listIndex={index}
                        />
                    ) : null;
                })}
            </div>
            
            {products.length > NUMBER_OF_VISIBLE_ITEMS && (
                <div className="text-center">
                    <Button type="button" size="small" onClick={() => setIsCollapsed((prev) => !prev)}>
                        <span>
                            {isCollapsed
                                ? t('Hide the next {{ count }} best sellers', {
                                      count: products.length - NUMBER_OF_VISIBLE_ITEMS,
                                  })
                                : t('View the next {{ count }} best sellers', {
                                      count: products.length - NUMBER_OF_VISIBLE_ITEMS,
                                  })}
                        </span>
                    </Button>
                </div>
            )}
        </div>
```

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
```diff
@@ -0,0 +1,63 @@
...
+                <div className="text-center">
+                    <Button type="button" size="small" onClick={() => setIsCollapsed((prev) => !prev)}>
+                        <span>
+                            {isCollapsed
+                                ? t('Hide the next {{ count }} best sellers', {
+                                      count: products.length - NUMBER_OF_VISIBLE_ITEMS,
+                                  })
+                                : t('View the next {{ count }} best sellers', {
+                                      count: products.length - NUMBER_OF_VISIBLE_ITEMS,
+                                  })}
```
tvikito:
```suggestion
                            {t(isCollapsed ? 'Hide the next {{ count }} best sellers' : 'View the next {{ count }} best sellers', {
                                      count: products.length - NUMBER_OF_VISIBLE_ITEMS,
                                  })}
```

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
```diff
@@ -0,0 +1,63 @@
...
+                        </span>
+                    </Button>
+                </div>
+            )}{' '}
```
tvikito:
```suggestion
            )}
```

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    return (
+        <div className="mb-8">
+            <div className="mb-3 break-words font-bold text-dark lg:text-lg">
+                {t('Do not want to choose? Choose certainty')}
+            </div>
+            <div className="mb-5">
+                {products.map((product, index) => {
+                    return index <= NUMBER_OF_VISIBLE_ITEMS - 1 || isCollapsed ? (
```
tvikito: Maybe that's only my own perspective but isn't this more intuitive?

```suggestion
                    return (index + 1) <= NUMBER_OF_VISIBLE_ITEMS || isCollapsed ? (
```

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
tvikito: Actually I am wondering, was this text required? Maybe just simple "Show more"/"Show less" would be enough for this feature?

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const t = useTypedTranslationFunction();
+    const router = useRouter();
+    const [isCollapsed, setIsCollapsed] = useState(false);
+
+    useComponentUpdate(() => setIsCollapsed(false), [router.asPath]);
+    useGtmSliderProductListViewEvent(products, GtmProductListNameType.bestsellers);
+
+    if (products.length === 0) {
+        return null;
+    }
```
tvikito: Since you are passing data from parent doesn't make much sense to me to run logic until these lines. What about put this check to don't even show this component in case we don't have any data?

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
```diff
@@ -0,0 +1,63 @@
...
+type CategoryBestsellersProps = {
+    products: ListedProductFragmentApi[];
+};
+
+export const CategoryBestsellers: FC<CategoryBestsellersProps> = ({ products }) => {
+    const t = useTypedTranslationFunction();
+    const router = useRouter();
+    const [isCollapsed, setIsCollapsed] = useState(false);
+
+    useComponentUpdate(() => setIsCollapsed(false), [router.asPath]);
```
tvikito: This doesn't seem to work at any scenarios which I've tested. If you need to persist this state, let's remove this and save value in Zustand.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
tvikito: I am not sure if it's just a bug on my side but I don't see the `count` variable shown in the translation string.

image

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
tvikito: I am not sure about this but what about prepare/filter data for this loop in advance? I understand that you see here algorithms here and yes it's one small loop more, but to me it feels like correct way since ideally I want to work in the this JSX part with prepared data and don't put logic itself here. Up to you.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellersListItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+                />
+            </div>
+
+            <div className="w-2/6 text-right lg:w-2/12" data-testid={TEST_IDENTIFIER + 'totalprice'}>
+                <div>
+                    <div className="text-h6 font-bold leading-5 text-primary" data-testid={TEST_IDENTIFIER + 'price'}>
+                        {formatPrice(product.price.priceWithVat)}
+                    </div>
+                </div>
+            </div>
```
tvikito: I understand that until it's being decided if remove or not remove `data-testid` in scenarios where aren't those values used you want to follow the convention but is it really necessary to include them here in new components?

tvikito: Let's refactor this to usage only for one `div` no need for those three.

project-base/storefront/public/locales/cs/common.json
```diff
@@ -141,6 +142,10 @@
     "Go to cart": "Přejít do košíku",
     "Great choice! We have added your item to the cart": "Skvělá volba! Vaše zboží jsme přidali do košíku",
     "Hide results": "Skrýt výsledky",
+    "Hide the next {{ count }} best sellers_one": "Skrýt {{ count }} následující produkt",
+    "Hide the next {{ count }} best sellers_few": "Skrýt {{ count }} následující produkty",
+    "Hide the next {{ count }} best sellers_many": "Skrýt {{ count }} následujících produktů",
+    "Hide the next {{ count }} best sellers_other": "Skrýt {{ count }} následujících produktů",
```
tvikito: This seems to me very overwhelming. If it is not required by business, I don't see any beneficial here. Just bunch of additional unnecessary data on client. If you agree let's put just simple "Show more"/"Show less" as suggested in previous comment.

tvikito (changes requested):

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellersListItem.tsx
sebaholesz: Thanks for the suggestion, changed.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: Alright, done

project-base/storefront/public/locales/cs/common.json
sebaholesz: Yeah, TBH I agree with you :laughing:

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: I completely removed the count, but I left the ternary operator outside of the function call. Let me know WYT

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: Nice, thanks

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: Yeah, lets write it your way.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: You're right

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: Good point, rewritten.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: Tbh I took some of the things pretty brainlessly from a project and this is the effect of it :laughing: The point of this line of code was to hide the products every time you go from a category to category. Therefore, the point is not to remember, but correctly forget the state. However, after some consideration and testing, I realized it's not needed at all, so I just removed it :slightly_frowning_face:

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: This is an issue with translates which pisses me off. Not a problem with code, but since I removed the dynamic translations, I don't think we need to take care of it rn.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
sebaholesz: Yeah. in this case where we have at most 10 products, I think everybody's PC can handle that :laughing:

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellersListItem.tsx
sebaholesz: Removed data test IDs and also those redundant divs

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
tvikito: Perfect, I have nothing to complain about :)

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellers.tsx
tvikito: One more thing which I had on my mind. I thought about separate those filtered items to the outside variable like

```
const shownProducts = products.filter((_, index) => index + 1 <= NUMBER_OF_VISIBLE_ITEMS || !isCollapsed)
```

I'll leave it up to you.

tvikito (approved):

pk16011990 (approved):

PR #2742, opened 2023-08-22
[framework] tests are now multilingual

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When first domain was switched to some other language, many tests would fail on error. This PR solves all issues in Czech language and should mitigate most of the errors in other languages.
|New feature| No 
|Fixes issues| ... 

----
  - https://tl-multidomain-tests.odin.shopsys.cloud
  - https://cz.tl-multidomain-tests.odin.shopsys.cloud


packages/framework/src/Component/Domain/Domain.php
```diff
@@ -207,4 +207,12 @@ public function getDateTimeZone(): DateTimeZone
     {
         return $this->getCurrentDomainConfig()->getDateTimeZone();
     }
+
+    /**
+     * @return \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig[]
+     */
+    public function getAllDomainConfigs(): array
```
grossmannmartin: it seems like we already have the method with the same implementation.  
`getAllIncludingDomainConfigsWithoutDataCreated`

packages/framework/src/Migrations/MultidomainMigrationTrait.php
```diff
@@ -4,30 +4,66 @@
...
      * @param int $domainId
      * @return string
      */
-    protected function getDomainLocale($domainId)
+    protected function getDomainLocale(int $domainId): string
     {
-        return $this
-            ->sql('SELECT get_domain_locale(:domainId)', ['domainId' => $domainId])
-            ->fetchOne();
+        return $this->getDomain()->getDomainConfigById($domainId)->getLocale();
```
grossmannmartin: I noticed the other occurrences remain in other migrations. Why was it necessary to change it specifically here?
The `get_domain_locale` seems to do the same thing as the implementation here.
Can you explain it to me, please?

packages/framework/src/Migrations/Version20160512152113.php
```diff
@@ -80,6 +82,14 @@ public function up(Schema $schema): void
...
     }
 
+    /**
+     * @param \Symfony\Component\DependencyInjection\ContainerInterface|null $container
+     */
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
```
grossmannmartin: isn't possible to have this method directly in the multidomainMigrationTrait?

packages/framework/src/Migrations/MultidomainMigrationTrait.php
```diff
@@ -65,5 +65,8 @@ protected function getAllLocales(): array
         return $domainLocales;
     }
 
-    abstract public function setContainer(ContainerInterface $container = null);
+    /**
+     * @param \Symfony\Component\DependencyInjection\ContainerInterface|null $container
+     */
+    abstract public function setContainer(?ContainerInterface $container = null);
```
grossmannmartin: Why is the nullability required?

project-base/app/src/Migrations/Version20200221155940.php
```diff
@@ -5,41 +5,75 @@
...
+final class Version20200221155940 extends AbstractMigration implements ContainerAwareInterface
 {
+    use MultidomainMigrationTrait;
+
     /**
      * @param \Doctrine\DBAL\Schema\Schema $schema
      */
     public function up(Schema $schema): void
     {
+        $translationId = 1;
```
grossmannmartin, line 22: ID is autogenerated, is it necessary to manually set IDs?

project-base/app/translations/tests.en.po
```diff
@@ -0,0 +1,8 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: cs\n"
```
grossmannmartin:
```suggestion
"Language: en\n"
```

packages/framework/tests/Unit/Component/ArrayUtils/ArraySorterTest.php
```diff
@@ -0,0 +1,65 @@
...
+            ['l' => 'x'],
+            ['l' => 'y'],
+            ['l' => 'ý'],
+            ['l' => 'z'],
+            ['l' => 'ž'],
+        ];
+
+        $arrayCzechLettersExpected = $arrayCzechLetters;
+
+        ArraySorter::sortArrayAlphabeticallyByValue('l', $arrayCzechLetters);
```
grossmannmartin: I'm not sure about the ArraySorter and its test. 
Currently it work only with cs_CZ collation, but that fact is hidden in the implementation. 
Shouldn't the collation be configurable, or at least mentioned in the method/class name? 

What happens if the parameter in the English locale starts with Ch? Will it be properly sorted? 

Also, the current test will pass for the implementation 

```
public static function sortArrayAlphabeticallyByValue(string $valueKey, array &$array): void {
    return $array;
}
```

That leads me to the thought, that the test is not sufficient. I think it would be nice to truly test the sorting of the array

project-base/app/tests/FrontendApiBundle/Functional/Product/PaginatedProductsTest.php
```diff
@@ -106,7 +97,7 @@ private function getQueryWithOrderingMode(string $orderingMode): string
         return '
             query {
                 products (
-                    first: 5
+                    first: 4
```
grossmannmartin, line 100: Why the results were not the same for the 5 products? And will it be okay, when the domain will not be Czech, but German for example?

project-base/app/tests/FrontendApiBundle/Functional/Product/Flag/FlagsTest.php
```diff
@@ -19,34 +20,38 @@ public function testFlags(): void
...
+        ArraySorter::sortArrayAlphabeticallyByValue('name', $flags);
+
+        $arrayExpected = [
+            'data' => [
+                'flags' => $flags,
+            ],
+        ];
 
-        $this->assertQueryWithExpectedJson($query, $jsonExpected);
+        $this->assertQueryWithExpectedJson($query, json_encode($arrayExpected, JSON_THROW_ON_ERROR));
```
grossmannmartin: maybe `assertQueryWithExpectedArray` could be used instead of the encoding json, what do you think?

project-base/app/tests/FrontendApiBundle/Functional/Product/Flag/FlagTest.php
```diff
@@ -109,90 +111,82 @@ public function testFlagByUuidFilteredByAnotherFlag(): void
...
-        $this->assertQueryWithExpectedJson($query, $jsonExpected);
+        ArraySorter::sortArrayAlphabeticallyByValue('name', $products);
+
+        $productsWithNodes = [];
+
+        for ($i = 0; $i < $limit; $i++) {
+            $productsWithNodes[] = [
+                'node' => $products[$i],
+            ];
+        }
```
grossmannmartin, line 155: why is necessary to prepare 10 products and then use only 5?

grossmannmartin (changes requested):

packages/framework/src/Component/Domain/Domain.php
TomasLudvik: Oh, that I overlooked, fixed

packages/framework/src/Migrations/MultidomainMigrationTrait.php
TomasLudvik: I have unified functions in the trait to use domain configs, I did not intend to update migrations not to use current DB functions.

packages/framework/src/Migrations/Version20160512152113.php
TomasLudvik: Nope as it is dependent on the ContainerAwareInterface.

packages/framework/src/Migrations/MultidomainMigrationTrait.php
TomasLudvik: It was defined same as in the interface, but you are right, it is not necessary here

TomasLudvik: It is used for adding translations. ID was manually set here before, I only added foreach

project-base/app/src/Migrations/Version20200221155940.php
TomasLudvik: It is used for adding translations. ID was manually set here before, I only added foreach

project-base/app/translations/tests.en.po
TomasLudvik: fixed

packages/framework/tests/Unit/Component/ArrayUtils/ArraySorterTest.php
TomasLudvik: I removed the test, as at the time of writing I was not using internal PHP functions, so I wrote the test, but now it is not necessary anymore.

Also, I have added locale as a required parameter to function so it should be no longer a problem.

project-base/app/tests/FrontendApiBundle/Functional/Product/PaginatedProductsTest.php
TomasLudvik: That is a problematic topic - the last four include a brand name at the beginning of the name, so I hope it will be sufficient for most, but as I wrote in the description of this PR - I am trying to fix most, but there is no 100% solution.

project-base/app/tests/FrontendApiBundle/Functional/Product/Flag/FlagsTest.php
TomasLudvik: Changed

project-base/app/tests/FrontendApiBundle/Functional/Product/Flag/FlagTest.php
TomasLudvik: Again - I hope that the first ten will be sufficient for the test to work in most of the languages.

packages/framework/src/Migrations/Version20160512152113.php
grossmannmartin: I understand that the Migration itself has to implement the `ContainerAwareInterface`, but the implementation of the `setContainer` method could be in the trait, or?  
That way we will not be forced to copy-paste the method every time we need to use the `MultidomainMigrationTrait`

TomasLudvik: Ok I have changed it

grossmannmartin (approved):

pk16011990 (changes requested):
Missing upgrade notes for addin `implements ContainerAwareInterface` :)

pk16011990 (approved):

PR #2748, opened 2023-08-24
[monorepo-tools] let the user decide if he wants to proceed with one repo

tolik518:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We are in need of using only 1 repo as an argument
|New feature| Yes 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| No, but could break if script is used in automatization where only one repo was provided 
|Fixes issues| ... 

We're using the [monorepo-tools](https://github.com/shopsys/monorepo-tools) quite frequently and it makes sense, in our context, to build a monorepo with only one repo and add other repos later.
We are currently commenting out the exit to use the tools to fit our needs, but better practice would be to let the user decide if he wants to use only 1 repo.

I had created a [PR](https://github.com/shopsys/monorepo-tools/pull/2) in the monorepo-tools repo before realizing it had to be opened here.


grossmannmartin:
Hi @tolik518, thank you for your contribution.
We'll review this PR as soon as possible :)

grossmannmartin (approved):

pk16011990 (approved):

tolik518:
Will the readonly repo also be updated - as the PR is now merged? @grossmannmartin

PR #2761, opened 2023-08-31
[project-base] reduced image url redis cache size

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Only information about the source image is saved into the Redis cache as nothing else is necessary.
|New feature| No 
|Fixes issues| ... 

----
  - https://tl-reduce-redis-cache-size.odin.shopsys.cloud
  - https://cz.tl-reduce-redis-cache-size.odin.shopsys.cloud

### DIFF
--- a/project-base/app/src/Component/Image/ImageFacade.php
+++ b/project-base/app/src/Component/Image/ImageFacade.php
@@ -271,7 +271,6 @@ class ImageFacade extends BaseImageFacade
         $cacheId = $this->getCacheIdForImageUrl(
             $image->getId(),
             $domainConfig->getId(),
-            $additionalSizeIndex,
         );
 
         $friendlyUrlSeoEntityName = $this->cache->get(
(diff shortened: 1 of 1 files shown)


project-base/app/src/Component/Image/ImageFacade.php
```diff
@@ -411,35 +411,21 @@ protected function saveImageOrdering($orderedImages): void
...
         ?int $additionalIndex = null,
     ): string {
         return sprintf(
-            'ImageUrl_imageId-%d_domainId-%d_type-%s_size-%s_additionalIndex-%s',
+            'ImageUrl_imageId-%d_domainId-%d_type-%s',
             $imageId,
             $domainId,
-            $type,
-            $sizeName,
             $additionalIndex,
```
grossmannmartin: The last variable in the string key is named `type`, but the `$additionalIndex` is passed.

Also, has the distinguishing by $additionalIndex any meaning? From the code, it seems like the additionalSizes will have always the same slug saved in the Redis. Isn't possible to reduce the data further and make the code a little less complex as the bonus?

Snímek obrazovky 2023-08-31 v 14 59 54

grossmannmartin (changes requested):

project-base/app/src/Component/Image/ImageFacade.php
TomasLudvik: I have updated it

grossmannmartin (approved):

pk16011990 (approved):

PR #2762, opened 2023-09-01
[shopsys] introduced php-fpm base image

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| introduced a base docker image for php-fpm container.
|New feature| No 
|Fixes issues| ... 

**Read-only Github repository with Dockerfile to build the image:**  
https://github.com/shopsys/php-image
- the repository is split automatically with the rest of the packages
- branches specified in Github Action are built automatically every day at 3 am
- tagged versions are built once on push tag to ensure the same working environment as it's on release
  - if necessary, the tagged version may be rebuilt manually by triggering the appropriate workflow

**Built image on Dockerhub:**  
https://hub.docker.com/repository/docker/shopsys/php-image/general

For testing changes in dockerfile, the `Use PHP Dockerfile instead of built image` label may be used on PR, to build php-image container directly from the Dockerfile during the CI build.

**Improvements:**
- much faster first build
- the same base image is used in the local environment, GitHub Actions, Gitlab CI
- simplified Github Action workflow

**Local development**

For the project, there is no change. Any necessary adjustments may be put into the `app/docker/php-fpm/Dockerfile` and the changes are propagated when the image is rebuilt.

For monorepo, especially when introducing changes in the `packages/php-image/Dockerfile`, several approaches may be used:
- verify changes by putting them into the `project-base/app/docker/php-fpm/Dockerfile` and, when satisfied, move those changes to `packages/php-image/Dockerfile`. Optionally test on CI with the `Use PHP Dockerfile instead of built image` label on the PR
- make changes in the `packages/php-image/Dockerfile` and build the image manually (preferably with any distinguished name/tag. That requires to temporarily change the `project-base/app/docker/php-fpm/Dockerfile` (the first `FROM` section)
  - this process may be automated with the script, but I don't find it necessary to introduce the script now as the changes in the Dockefile are quite rare 

----
  - https://mg-base-image.odin.shopsys.cloud
  - https://cz.mg-base-image.odin.shopsys.cloud


utils/releaser/src/ReleaseWorker/Release/TagPhpPlatformReleaseWorker.php
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * @param \PharIo\Version\Version $version
+     * @param string $initialBranchName
+     */
+    public function work(
+        Version $version,
+        string $initialBranchName = AbstractShopsysReleaseWorker::MAIN_BRANCH_NAME,
+    ): void {
+        $tempDirectory = trim($this->processRunner->run('mktemp -d -t shopsys-release-XXXX'));
+        $packageName = 'php-platform';
```
TomasLudvik: Introduce some constant for this as it is used in multiple files

utils/releaser/src/ReleaseWorker/Release/TagPhpPlatformReleaseWorker.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $this->processRunner->run(
+            sprintf(
+                'cd %s/%s && git checkout %s && git tag %s',
+                $tempDirectory,
+                $packageName,
+                $initialBranchName,
+                $versionString,
+            ),
+        );
```
TomasLudvik: Why is this separated here and not part of monorepo-split?

utils/releaser/src/ReleaseWorker/Release/TagPhpPlatformReleaseWorker.php
```diff
@@ -0,0 +1,122 @@
...
+
+
+        $this->processRunner->run('rm -r ' . $tempDirectory);
+        $this->symfonyStyle->note([
+            sprintf('Wait for Github Actions to build a tagged version of %s (approx 1 hour)', $packageName),
+            sprintf('You can track progress on https://github.com/shopsys/%s/actions', $packageName),
+        ]);
+        $this->confirm(
+            sprintf('Confirm that there are new version of %s on Docker Hub (https://hub.docker.com/r/shopsys/%1$s/tags)', $packageName),
```
TomasLudvik, line 98: You mean GitHub repository, right?

packages/php-platform/.github/workflows/build-tagged-version.yaml
```diff
@@ -0,0 +1,28 @@
...
+        permissions:
+            contents: read
+        steps:
+            -   name: GIT checkout tag - ${{ github.ref }}
+                uses: actions/checkout@v3
+            -   name: Login to DockerHub
+                uses: docker/login-action@v2
+                with:
+                    username: ${{ secrets.DOCKERHUB_USERNAME }}
+                    password: ${{ secrets.DOCKERHUB_PASSWORD }}
```
TomasLudvik: Why are you using DockerHub instead of GitHub Package Registry? Is it because you do not want to use ghcr.io in images names?

packages/php-platform/.github/workflows/build-tagged-version.yaml
```diff
@@ -0,0 +1,28 @@
...
+            -   name: Set up QEMU
+                uses: docker/setup-qemu-action@v2
+            -   name: Docker Setup Buildx
```
TomasLudvik: Why is this necessary?

packages/php-platform/.github/workflows/build-tagged-version.yaml
```diff
@@ -0,0 +1,28 @@
+on:
+    push:
+        tags:
+            - '**'
+    workflow_dispatch:
```
TomasLudvik: I think you do not want to build this manually as you are missing TAG here

packages/php-platform/.github/workflows/rebuild-dev-image.yaml
```diff
@@ -0,0 +1,33 @@
...
+        -   cron: '0 3 * * *'
+    workflow_dispatch:
+name: Rebuild PHP Platform Docker image
+jobs:
+    rebuild-php-fpm-images:
+        name: Rebuild ${{ matrix.branches }} branch
+        runs-on: ubuntu-22.04
+        strategy:
+            matrix:
+                branches: [ '13.0' ]
```
TomasLudvik: We currently support versions 11 and 12 also right?

.ci/monorepo_functions.sh
```diff
@@ -22,6 +22,7 @@ get_all_packages() {
         form-types-bundle \
         migrations \
         monorepo-tools \
+        php-platform \
```
TomasLudvik: I feel like `php-platform` name is not clear, at first it feels like some PHP-based platform project.

`php-fpm-docker-image` for example is much more apparent to me

.github/workflows/docker-build.yaml
```diff
@@ -145,6 +145,11 @@ jobs:
                     registry: ghcr.io
                     username: ${{ github.actor }}
                     password: ${{ secrets.GITHUB_TOKEN }}
+            -   name: Build custom base image if requested
+                if: "contains(github.event.pull_request.labels.*.name, 'build-php')"
```
TomasLudvik: We wanted labels to be nice and clear and you started to introduce new ones in "programmer" style. Is there any reason why to not name label like this: "Force custom PHP-FPM Docker Image build"?

TomasLudvik (changes requested):

utils/releaser/src/ReleaseWorker/Release/TagPhpPlatformReleaseWorker.php
grossmannmartin: This tagging has to be done prior to the monorepo-split (the one that splits tags) to ensure the `FROM` instruction can be safely updated and committed and the application can be built after the split.

packages/php-platform/.github/workflows/build-tagged-version.yaml
grossmannmartin: I decided to use the default Docker registry for several reasons: 
- I still find it the "industry standard" and I find it easier to write the `docker run` or FROM instructions from the top of my head
- Dockerhub offers some additional nice tooling – for example, Docker Scout for image analysis
- I find the user interface on dockerhub more user-friendly

packages/php-platform/.github/workflows/build-tagged-version.yaml
grossmannmartin: buildx is necessary to be able to build image for multiple platforms

packages/php-platform/.github/workflows/build-tagged-version.yaml
grossmannmartin: 🤔  I wanted to be able to manually run the workflow for the already existing tag to trigger the rebuild of the image (for example when some security issue is found and fixed in the base image).

Should it be written differently?

packages/php-platform/.github/workflows/rebuild-dev-image.yaml
grossmannmartin: true, but that would require backporting this change into the previous versions. 

I did not find it necessary because: 
- As versions 11 and 12 are not developed actively, but merely bug-fixed, I think it's a waste of computer resources to periodically build images for those branches
- To leverage the benefits to the end-user it would require us to release a version – and 11 is now only bug-fixed

Generally speaking, the 13.0 version of the base image may be used even for the 11.0 and the 12.0 versions 🤔

.ci/monorepo_functions.sh
grossmannmartin: It was a struggle to come up with a nice name, tbh.

I had the following criteria:
- It should not be similar to the existing official php image, as the versioning will be completely different 
- It should describe that it's not just another php-fpm image, but contains more specific stuff for the desired usage
- It should be short and easy to remember
- It should at least loosely remind the Shopsys Platform
- Preferably it should not contain any unnecessary technical terms (such as fpm) so we can use the best technology at the time without worrying about the name

So the `shopsys/php-platform` seemed like the best choice.

.github/workflows/docker-build.yaml
grossmannmartin: I wanted to distinguish between the "functional labels" and the mere markers. 

But I don't have any strong feelings about the label names.
The suggested one is really long and emphasizes the importance of the label at the expense of others, more frequently used, while in reality this one (and the review-isolated) will be used rarely.

Snímek obrazovky 2023-09-05 v 13 23 14

utils/releaser/src/ReleaseWorker/Release/TagPhpPlatformReleaseWorker.php
grossmannmartin: better? 9a4f543e6d4caaa87a40a587c1b13c4e53885d54

.ci/monorepo_functions.sh
TomasLudvik: Okay, but I miss info about it being a Docker image. So maybe something like `platform-php-docker` or `plaform-php-docker-image` so when you read it out loud it is `Shopsys Platform PHP Docker Image`? :) 
Also interesting choices as searching on Docker Hub would be easy are:
`shopsys-php-image` `shopsys-php-docker-image`

.github/workflows/docker-build.yaml
TomasLudvik: Well, at least I can see what it does, as the description is trimmed 😃 

What about `Use PHP Dockerfile instead of built image`? It will be at the end and descriptive.

TomasLudvik: Same then could be used for dependencies :)

.ci/monorepo_functions.sh
grossmannmartin: changed to php-image as we discussed

.github/workflows/docker-build.yaml
grossmannmartin: I have renamed the labels to 
`Use PHP Dockerfile instead of built image` 
`Use docker-compose defined services for review`

note: The `dependencies` label has no special meaning and probably was just named with the first case lowered.

TomasLudvik (approved):

pk16011990 (approved):

PR #2765, opened 2023-09-04
fix SimpleNavigation slider

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| on smaller viewports we had broken slider, now it's fixed
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fwcc-1699-fix-categories-slider.odin.shopsys.cloud
  - https://cz.tv-fwcc-1699-fix-categories-slider.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/components/Pages/CategoryDetail/AdvancedSeoCategories.tsx
+++ b/project-base/storefront/components/Pages/CategoryDetail/AdvancedSeoCategories.tsx
@@ -6,17 +6,19 @@ type AdvancedSeoCategoriesProps = {
     readyCategorySeoMixLinks: CategoryDetailFragmentApi['readyCategorySeoMixLinks'];
 };
 
+const simpleNavigationItemTwClass = 'lg:justify-center text-center';
+
 export const AdvancedSeoCategories: FC<AdvancedSeoCategoriesProps> = ({ readyCategorySeoMixLinks }) => {
     const { t } = useTranslation();
-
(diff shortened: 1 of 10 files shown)


project-base/storefront/components/Pages/CategoryDetail/AdvancedSeoCategories.tsx
```diff
@@ -6,17 +6,19 @@ type AdvancedSeoCategoriesProps = {
     readyCategorySeoMixLinks: CategoryDetailFragmentApi['readyCategorySeoMixLinks'];
 };
 
+const simpleNavigationItemTwClass = 'lg:justify-center text-center';
+
```
sebaholesz, line 10: I do not see many benefits of having this in a separate variable. I will leave this to you, but I would just use it as-is.

sebaholesz (approved):

project-base/storefront/components/Pages/CategoryDetail/AdvancedSeoCategories.tsx
tvikito: The purpose of having this in separate variable is to be able to use IntelliSense to recognize the "TwClass" regex. We agreed on this with other FE guys. Maybe you missed this meeting, but this is only way currently to tell VS Code that this is meant for Tailwind classes. However you need to setup this regex on your side in your editor.

pk16011990 (approved):

PR #2770, opened 2023-09-05
[github-actions] fixed checks on fork build

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR fixes build from fork repositories on Github Actions
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -485,15 +485,11 @@ jobs:
                 if: needs.variables.outputs.DOCKER_STOREFRONT_CI_IMAGE_EXISTS == 0
                 run: .github/build-storefront-ci-image.sh ${{ needs.variables.outputs.DOCKER_STOREFRONT_CI_REPOSITORY_TAG }}
             -   name: Build PHP-FPM image
-                run: .github/build-php-fpm-image.sh php-fpm:base base
-            -   name: Prepare docker-compose.yml file from docker-compose.github-actions.yml.dist
-                run: |
-                    sh .github/prepare-docker-compose-file.sh \
-                    "./docker/conf/docker-compose.github-actions.yml.dist" \
-                    "./docker-compose.yml" \
-                    "php-fpm:base" \
-                    "${{ needs.variables.outputs.DOCKER_STOREFRONT_CI_REPOSITORY_TAG }}" \
-                    "${{ needs.variables.outputs.DOCKER_ELASTICSEARCH_REPOSITORY_TAG }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -570,12 +570,22 @@ jobs:
...
             -   name: Stop previous running application
                 run: docker compose down
+            -   name: Prepare docker-compose.yml file from docker-compose.github-actions.cypress.yml.dist
+                run: |
+                    sh .github/prepare-docker-compose-file.sh \
+                    "./docker/conf/docker-compose.github-actions.cypress.yml.dist" \
+                    "./docker-compose.yml" \
+                    "${{ needs.variables.outputs.DOCKER_PHP_FPM_BASE_REPOSITORY_TAG }}" \
+                    "${{ needs.variables.outputs.DOCKER_STOREFRONT_REPOSITORY_TAG }}" \
+                    "${{ needs.variables.outputs.DOCKER_ELASTICSEARCH_REPOSITORY_TAG }}"
```
TomasLudvik: I see one problem with this - before docker-composes were in artifacts, so in case of need to debug something, you just downloaded them and used them.

Also you did not removed all docker composes from artifacts

.github/workflows/docker-build.yaml
```diff
@@ -570,12 +570,22 @@ jobs:
...
             -   name: Build application
                 run: |
-                    mv docker-compose.cypress.yml docker-compose.yml
                     docker compose up -d --wait
-                    docker compose exec php-fpm php phing -D production.confirm.action=y db-create frontend-api-generate-new-keys build-demo-dev-quick error-pages-generate test-db-create test-db-demo test-elasticsearch-index-recreate test-elasticsearch-export
-                    docker compose exec php-fpm php phing -D production.confirm.action=y -D change.environment=acc environment-change
+                    docker cp ./ shopsys-framework-php-fpm:/var/www/html
```
TomasLudvik, line 556: Why use this instead of volume?

TomasLudvik (changes requested):

.github/workflows/docker-build.yaml
grossmannmartin: docker composes in artifacts are still used in the default build.
The problem with the fork build is, that we cannot use the docker-compose from artifacts because the step with the build image is not run, therefore the image in the docker-compose in the artifacts does not exist.

.github/workflows/docker-build.yaml
grossmannmartin: So I can use the same docker-compose file

TomasLudvik (approved):

pk16011990 (approved):

TomasLudvik (approved):

PR #2771, opened 2023-09-05
add Storefront documentation

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| add Storefront documentation to the root docs folder and add missing documentation
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1632-add-fe-documentation.odin.shopsys.cloud
  - https://cz.tv-fw-1632-add-fe-documentation.odin.shopsys.cloud


docs/navigation.yml
```diff
@@ -1,13 +1,14 @@
...
-    - administration
-    - frontend-api
-    - extensibility
-    - automated-testing
-    - contributing
+  - index.md
+  - installation
+  - introduction
+  - model
+  - cookbook
```
grossmannmartin: please use 4 spaces indent in all yml files

tvikito: Not possible with usage of Prettier. We need to figure out how to combine IDE setup which you use on BE with our setup on FE side. Seems like Prettier doesn't care about .editorconfig and use its own setup.

tvikito: Should be resolved now

PR #2781, opened 2023-09-11
[framework] added auto rendered uuid for entities in administration

### LINKED ISSUE #2157: UUID displayed in administration
I believe it would be nice to display UUID in administration for related entities. 
For example product, category, article, etc.

It could be achieved really easily by adding displayOnlyType with UUID, similarly to the already present ID, or for example adding as a small text under the heading or something like that.

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| If rendered page contains an object with the UUID property, this UUID is automatically rendered into the `h1` block. 
|New feature| Yes 
|Fixes issues| closes #2157  

https://github.com/shopsys/shopsys/assets/1177414/38a5c48c-d5ed-44b9-bf59-fc8d8f750830

----
  - https://mg-display-uuid.odin.shopsys.cloud
  - https://cz.mg-display-uuid.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
@@ -1,0 +1,6 @@
+{% set keys = _context|keys %}
+{% if (keys[1] is defined) and (attribute(_context[keys[1]], 'uuid') is defined) %}
+    <span class="edit-uuid">
+        {{ _context[keys[1]].uuid }}
+    </span>
+{% endif %}
--- a/packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
+++ b/packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
@@ -1,6 +1,9 @@
 {% set keys = _context|keys %}
 {% if (keys[1] is defined) and (attribute(_context[keys[1]], 'uuid') is defined) %}
-    <span class="edit-uuid">
-        {{ _context[keys[1]].uuid }}
-    </span>
+    {% set uuid = _context[keys[1]].uuid %}
+    <span class="edit-uuid cursor-pointer js-copy-to-clipboard"
+          data-content="{{ uuid }}"
+          data-placement="right"
+          title="{{ uuid }}"
+    >UUID {{ icon('duplicate') }}</span>
 {% endif %}
--- a/packages/framework/assets/js/common/components/copyToClipboard.js
+++ b/packages/framework/assets/js/common/components/copyToClipboard.js
@@ -1,0 +1,35 @@
+import Register from '../utils/Register';
+import '../../common/bootstrap/tooltip';
+import Translator from 'bazinga-translator';
+
+export default class CopyToClipboard {
+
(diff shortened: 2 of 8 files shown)


packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
```diff
@@ -0,0 +1,6 @@
+{% set keys = _context|keys %}
+{% if (keys[1] is defined) and (attribute(_context[keys[1]], 'uuid') is defined) %}
+    <span class="edit-uuid">
+        {{ _context[keys[1]].uuid }}
+    </span>
+{% endif %}
```
TomasLudvik: It is weird to me to have one of many attributes of an entity displayed somewhere else than others, but I feel it is not my decision - @pk16011990

TomasLudvik (approved):

packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
pk16011990: I agree, It's weird. However, I understand that it's managed from one place.
Idea: Replace it with the text "UUID" with the tooltip (shows UUID hash) and icon "Copy to clipboard". And with a smaller margin-left, e.g. 1em.

packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
```diff
@@ -0,0 +1,6 @@
+{% set keys = _context|keys %}
+{% if (keys[1] is defined) and (attribute(_context[keys[1]], 'uuid') is defined) %}
```
pk16011990, line 2: `key[1]` assumes an entity passed to the template as the first variable? Or why 1?

grossmannmartin: It's something magical in the Symfony land... It does not matter in what order I have passed the variables, the entity was second in the context array 🤷 

And when it's only informative I decided to make this assumption. In the worst-case scenario, the UUID will not be displayed somewhere, and then we can adjust this "cleveristic".

packages/framework/src/Resources/views/Admin/Layout/displayUuid.html.twig
grossmannmartin: Well, the UX of the administration is far from ideal, and displaying more fields in the already long enough list of attributes certainly does not help. Especially when some of these fields are for the common administrator irrelevant (ID, UUID)

I like the idea of the icon – changed as you described, now it's quite elegant in my opinion :)

TomasLudvik: My point still stands. It is nice that it is managed from a single point, but I would prefer to add it to each form. 

The argument that the current UX is not ideal is valid, but adding a new flowing element is in my opinion less perfect than adding an attribute to an already long list, even if the administrator does not need it.

TomasLudvik (approved):

pk16011990 (approved):

PR #2806, opened 2023-09-15
[project-base] order is updated in GoPay only if configuration is set

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| There were calls on GoPay with 409 results. We do not want to spam GoPay servers with incorrect calls so we made this fix.
|New feature| No 
|Fixes issues| ... 

----
  - https://tl-gopay-export-fix.odin.shopsys.cloud
  - https://cz.tl-gopay-export-fix.odin.shopsys.cloud

### DIFF
--- a/project-base/app/.env.test
+++ b/project-base/app/.env.test
@@ -1,3 +1,11 @@
 # define your env variables for the test environment here
 
 DATABASE_NAME=shopsys-test
+
+GOPAY_IS_PRODUCTION_MODE=false
+GOPAY_EN_GOID=test-data
+GOPAY_EN_CLIENTID=test-data
+GOPAY_EN_CLIENTSECRET=test-data
+GOPAY_CS_GOID=test-data
+GOPAY_CS_CLIENTID=test-data
+GOPAY_CS_CLIENTSECRET=test-data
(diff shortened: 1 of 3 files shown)


project-base/app/config/packages/gopay.yaml
```diff
@@ -1,15 +1,10 @@
...
-        cs:
-            goid: '%env(GOPAY_CS_GOID)%'
-            clientId: '%env(GOPAY_CS_CLIENTID)%'
-            clientSecret: '%env(GOPAY_CS_CLIENTSECRET)%'
-            language: CS
+        #        en:
+        #            goid: '%env(GOPAY_EN_GOID)%'
+        #            clientId: '%env(GOPAY_EN_CLIENTID)%'
+        #            clientSecret: '%env(GOPAY_EN_CLIENTSECRET)%'
+        #            language: EN
```
grossmannmartin: With this configuration commented, we will not have Gopay enabled on our devel/production server, where the Gopay keys for both cs and en domains are set.

I would like to keep the env variables and gopay config as it is, and just check whether the values are set or empty to determine whether the Gopay request should be made or not.

grossmannmartin (changes requested):

project-base/app/config/packages/gopay.yaml
TomasLudvik: Ok, I have added it back

grossmannmartin (changes requested):
can you please check the failed tests?

grossmannmartin (approved):

pk16011990 (approved):

PR #2807, opened 2023-09-17
Transport and payment page fetching state refactoring

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Logic on the transport and payment page was unclear and caused weird bugs, especially when loading things from the previous order. This PR tries to focus on the refactoring of the logic and making it a little bit more clear. It tries not to touch many things around as this should be handled in a more general PR at a later stage. It also solved a bug with incorrect loading of transport and payment from the previous order which happened if the previous transport was not available for the current one.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-tnp-fetching-states-and-skeletons.odin.shopsys.cloud
  - https://cz.sh-tnp-fetching-states-and-skeletons.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
+++ b/project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
@@ -7,13 +7,22 @@ import { NotificationBars } from 'components/Layout/NotificationBars/Notificatio
 import { Webline } from 'components/Layout/Webline/Webline';
 import useTranslation from 'next-translate/useTranslation';
 import { useDomainConfig } from 'hooks/useDomainConfig';
+import { OrderLayoutContentSkeleton } from './OrderLayoutContentSkeleton';
+import { Footer } from '../Footer/Footer';
+import { ReactNode } from 'react';
 
 type OrderLayoutProps = {
     activeStep: number;
     isTransportOrPaymentLoading?: boolean;
+    contentSkeleton?: ReactNode;
 };
 
-export const OrderLayout: FC<OrderLayoutProps> = ({ activeStep, isTransportOrPaymentLoading, children }) => {
+export const OrderLayout: FC<OrderLayoutProps> = ({
+    activeStep,
+    isTransportOrPaymentLoading,
+    children,
+    contentSkeleton,
+}) => {
     const { t } = useTranslation();
     const { url } = useDomainConfig();
 
@@ -26,17 +35,25 @@ export const OrderLayout: FC<OrderLayoutProps> = ({ activeStep, isTransportOrPay
             </Webline>
             <Adverts positionName="header" withGapBottom withWebline />
             <Webline>
-                <OrderSteps activeStep={activeStep} domainUrl={url} />
-            </Webline>
-            <Webline>
-                <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
-                    <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
-                    <div className="w-full vl:max-w-md">
-                        <OrderSummary isTransportOrPaymentLoading={isTransportOrPaymentLoading} />
-                    </div>
-                </div>
+                {contentSkeleton ? (
+                    <OrderLayoutContentSkeleton>{contentSkeleton}</OrderLayoutContentSkeleton>
+                ) : (
+                    <>
+                        <OrderSteps activeStep={activeStep} domainUrl={url} />
+
+                        <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
+                            <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
+                            <div className="w-full vl:max-w-md">
+                                <OrderSummary isTransportOrPaymentLoading={isTransportOrPaymentLoading} />
+                            </div>
+                        </div>
+                    </>
+                )}
             </Webline>
             <Adverts positionName="footer" withGapBottom withWebline />
+            <Webline type="dark">
+                <Footer simpleFooter />
+            </Webline>
         </>
     );
 };
--- a/project-base/storefront/components/Layout/OrderLayout/OrderLayoutContentSkeleton.tsx
+++ b/project-base/storefront/components/Layout/OrderLayout/OrderLayoutContentSkeleton.tsx
@@ -1,0 +1,16 @@
+import Skeleton from 'react-loading-skeleton';
+
+export const OrderLayoutContentSkeleton: FC = ({ children }) => {
+    return (
+        <>
+            <Skeleton className="h-10 w-full" />
+
+            <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
+                <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
+                <div className="h-60 w-full vl:max-w-md">
+                    <Skeleton className="h-full w-full" />
+                </div>
+            </div>
+        </>
+    );
+};
--- a/project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
+++ b/project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
@@ -1,117 +1,67 @@
 import { TransportAndPaymentSelect } from './TransportAndPaymentSelect/TransportAndPaymentSelect';
 import { OrderAction } from 'components/Blocks/OrderAction/OrderAction';
 import { useCurrentCart } from 'connectors/cart/Cart';
-import {
-    LastOrderFragmentApi,
-    ListedStoreFragmentApi,
-    TransportWithAvailablePaymentsAndStoresFragmentApi,
-    useStoreQueryApi,
-} from 'graphql/generated';
+import { useTransportsQueryApi } from 'graphql/generated';
 import { hasValidationErrors } from 'helpers/errors/hasValidationErrors';
-import { getGtmPickupPlaceFromLastOrder, getGtmPickupPlaceFromStore } from 'gtm/helpers/mappers';
 import { getInternationalizedStaticUrls } from 'helpers/getInternationalizedStaticUrls';
-import { ChangePaymentHandler } from 'hooks/cart/useChangePaymentInCart';
-import { ChangeTransportHandler } from 'hooks/cart/useChangeTransportInCart';
+import { useChangePaymentInCart } from 'hooks/cart/useChangePaymentInCart';
+import { useChangeTransportInCart } from 'hooks/cart/useChangeTransportInCart';
 import useTranslation from 'next-translate/useTranslation';
 import { useDomainConfig } from 'hooks/useDomainConfig';
 import dynamic from 'next/dynamic';
 import { useRouter } from 'next/router';
-import { useMemo, useState } from 'react';
+import { useState } from 'react';
 import { usePersistStore } from 'store/usePersistStore';
 import { GtmMessageOriginType } from 'gtm/types/enums';
+import { OrderLayout } from 'components/Layout/OrderLayout/OrderLayout';
+import {
+    getTransportAndPaymentValidationMessages,
+    useHandleTransportAndPaymentLoadingAndRedirect,
+    useLoadTransportAndPaymentFromLastOrder,
+} from './helpers';
+import Skeleton from 'react-loading-skeleton';
 
 const ErrorPopup = dynamic(() => import('components/Forms/Lib/ErrorPopup').then((component) => component.ErrorPopup));
 
-type TransportAndPaymentContentProps = {
-    transports: TransportWithAvailablePaymentsAndStoresFragmentApi[] | undefined;
-    lastOrder: LastOrderFragmentApi | null;
-    changeTransportInCart: ChangeTransportHandler;
-    changePaymentInCart: ChangePaymentHandler;
-    isTransportSelectionLoading: boolean;
-    isPaymentSelectionLoading: boolean;
-};
-
-type TransportAndPaymentErrorsType = {
-    transport: {
-        name: 'transport';
-        label: string;
-        errorMessage: string | undefined;
-    };
-    payment: {
-        name: 'payment';
-        label: string;
-        errorMessage: string | undefined;
-    };
-    goPaySwift: {
-        name: 'goPaySwift';
-        label: string;
-        errorMessage: string | undefined;
-    };
-};
-
-export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = ({
-    transports,
-    lastOrder,
-    changePaymentInCart,
-    changeTransportInCart,
-    isPaymentSelectionLoading,
-    isTransportSelectionLoading,
-}) => {
+export const TransportAndPaymentContent: FC = () => {
     const router = useRouter();
     const { url } = useDomainConfig();
     const { t } = useTranslation();
+    const cartUuid = usePersistStore((store) => store.cartUuid);
     const { transport, pickupPlace, payment, paymentGoPayBankSwift } = useCurrentCart();
     const [isErrorPopupVisible, setErrorPopupVisibility] = useState(false);
-    const packeteryPickupPoint = usePersistStore((store) => store.packeteryPickupPoint);
-
     const [cartUrl, contactInformationUrl] = getInternationalizedStaticUrls(
         ['/cart', '/order/contact-information'],
         url,
     );
 
-    const [{ data: pickupPlaceData }] = useStoreQueryApi({
-        pause: lastOrder?.transport.transportType.code === 'packetery' || !lastOrder?.pickupPlaceIdentifier,
-        variables: { uuid: lastOrder?.pickupPlaceIdentifier ?? null },
+    const [changeTransportInCart, isTransportSelectionLoading] = useChangeTransportInCart();
+    const [changePaymentInCart, isPaymentSelectionLoading] = useChangePaymentInCart();
+    const [{ data: transportsData, fetching: areTransportsLoading }] = useTransportsQueryApi({
+        variables: { cartUuid },
+        requestPolicy: 'network-only',
     });
 
-    const transportAndPaymentValidationMessages = useMemo(() => {
-        const errors: Partial<TransportAndPaymentErrorsType> = {};
+    const [isLoadingTransportAndPaymentFromLastOrder, lastOrderPickupPlace] = useLoadTransportAndPaymentFromLastOrder(
+        changeTransportInCart,
+        changePaymentInCart,
+    );
 
-        if (!transport) {
-            errors.transport = {
-                name: 'transport',
-                label: t('Choose transport'),
-                errorMessage: t('Please select transport'),
-            };
-        } else {
-            if (transport.isPersonalPickup && pickupPlace?.identifier === undefined) {
-                errors.transport = {
-                    name: 'transport',
-                    label: t('Choose transport'),
-                    errorMessage: t('Please select transport with a personal pickup place'),
-                };
-            }
-            if (!payment) {
-                errors.payment = {
-                    name: 'payment',
-                    label: t('Choose payment'),
-                    errorMessage: t('Please select payment'),
-                };
-            }
-            if (payment?.goPayPaymentMethod?.identifier === 'BANK_ACCOUNT' && !paymentGoPayBankSwift) {
-                errors.goPaySwift = {
-                    name: 'goPaySwift',
-                    label: t('Choose your bank'),
-                    errorMessage: t('Please select your bank'),
-                };
-            }
-        }
+    const validationMessages = getTransportAndPaymentValidationMessages(
+        transport,
+        pickupPlace,
+        payment,
+        paymentGoPayBankSwift,
+        t,
+    );
 
-        return errors;
-    }, [transport, payment, paymentGoPayBankSwift, pickupPlace?.identifier, t]);
+    const isTransportAndPaymentSkeletonVisible = useHandleTransportAndPaymentLoadingAndRedirect(
+        areTransportsLoading,
+        isLoadingTransportAndPaymentFromLastOrder,
+    );
 
     const onSelectTransportAndPaymentHandler = () => {
-        if (hasValidationErrors(transportAndPaymentValidationMessages)) {
+        if (hasValidationErrors(validationMessages)) {
             setErrorPopupVisibility(true);
 
             return;
@@ -120,58 +70,46 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
         router.push(contactInformationUrl);
     };
 
-    const lastOrderPickupPlace: ListedStoreFragmentApi | null = useMemo(() => {
-        if (!lastOrder?.pickupPlaceIdentifier) {
-            return null;
-        }
-
-        if (packeteryPickupPoint?.identifier === lastOrder.pickupPlaceIdentifier) {
-            return packeteryPickupPoint;
-        }
-
-        if (pickupPlaceData?.store) {
-            return getGtmPickupPlaceFromStore(lastOrder.pickupPlaceIdentifier, pickupPlaceData.store);
-        }
-
-        return getGtmPickupPlaceFromLastOrder(lastOrder.pickupPlaceIdentifier, lastOrder);
-    }, [lastOrder, packeteryPickupPoint, pickupPlaceData?.store]);
-
     return (
         <>
-            {transports && (
-                <TransportAndPaymentSelect
-                    transports={transports}
-                    lastOrderPickupPlace={lastOrderPickupPlace}
-                    lastOrderTransportUuid={lastOrder?.transport.uuid ?? null}
-                    lastOrderPaymentUuid={lastOrder?.payment.uuid ?? null}
-                    changeTransportInCart={changeTransportInCart}
-                    changePaymentInCart={changePaymentInCart}
-                    isTransportSelectionLoading={isTransportSelectionLoading}
-                />
-            )}
+            <OrderLayout
+                activeStep={2}
+                isTransportOrPaymentLoading={isTransportSelectionLoading || isPaymentSelectionLoading}
+                contentSkeleton={isTransportAndPaymentSkeletonVisible ? <Skeleton className="h-64 w-full" /> : null}
+            >
+                {!!transportsData?.transports.length && (
+                    <TransportAndPaymentSelect
+                        transports={transportsData.transports}
+                        lastOrderPickupPlace={lastOrderPickupPlace}
+                        changeTransportInCart={changeTransportInCart}
+                        changePaymentInCart={changePaymentInCart}
+                        isTransportSelectionLoading={isTransportSelectionLoading}
+                    />
+                )}
 
-            <OrderAction
-                buttonBack={t('Back')}
-                buttonNext={t('Contact information')}
-                hasDisabledLook={
-                    hasValidationErrors(transportAndPaymentValidationMessages) ||
-                    isTransportSelectionLoading ||
-                    isPaymentSelectionLoading
-                }
-                isLoading={(isTransportSelectionLoading || isPaymentSelectionLoading) && !!transport && !!payment}
-                withGapTop
-                withGapBottom
-                buttonBackLink={cartUrl}
-                nextStepClickHandler={onSelectTransportAndPaymentHandler}
-            />
-
-            {isErrorPopupVisible && (
-                <ErrorPopup
-                    onCloseCallback={() => setErrorPopupVisibility(false)}
-                    fields={transportAndPaymentValidationMessages}
-                    gtmMessageOrigin={GtmMessageOriginType.transport_and_payment_page}
+                <OrderAction
+                    buttonBack={t('Back')}
+                    buttonNext={t('Contact information')}
+                    hasDisabledLook={
+                        hasValidationErrors(validationMessages) ||
+                        isTransportSelectionLoading ||
+                        isPaymentSelectionLoading
+                    }
+                    isLoading={(isTransportSelectionLoading || isPaymentSelectionLoading) && !!transport && !!payment}
+                    withGapTop
+                    withGapBottom
+                    buttonBackLink={cartUrl}
+                    nextStepClickHandler={onSelectTransportAndPaymentHandler}
                 />
-            )}
+
+                {isErrorPopupVisible && (
+                    <ErrorPopup
+                        onCloseCallback={() => setErrorPopupVisibility(false)}
+                        fields={validationMessages}
+                        gtmMessageOrigin={GtmMessageOriginType.transport_and_payment_page}
+                    />
+                )}
+            </OrderLayout>
         </>
     );
 };
--- a/project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
+++ b/project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
@@ -13,23 +13,15 @@ import {
     TransportWithAvailablePaymentsAndStoresFragmentApi,
     useGoPaySwiftsQueryApi,
 } from 'graphql/generated';
-import { logException } from 'helpers/errors/logException';
-import { mapPacketeryExtendedPoint, packeteryPick } from 'helpers/packetery';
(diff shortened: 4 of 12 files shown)


project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
```diff
@@ -25,18 +27,42 @@ export const OrderLayout: FC<OrderLayoutProps> = ({ activeStep, isTransportOrPay
...
+                        </div>
+                    </Webline>
+                </>
+            )}
+            <Adverts positionName="footer" withGapBottom withWebline />
+        </>
+    );
+};
+
+const OrderLayoutContentSkeleton: FC = ({ children }) => {
```
tvikito: Since we have folder for it and all skeletons are in separate file, let's move this to separate file as well.

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
```diff
@@ -25,18 +27,42 @@ export const OrderLayout: FC<OrderLayoutProps> = ({ activeStep, isTransportOrPay
...
                 <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
                     <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
-                    <div className="w-full vl:max-w-md">
-                        <OrderSummary isTransportOrPaymentLoading={isTransportOrPaymentLoading} />
+                    <div className="h-60 w-full vl:max-w-md">
+                        <Skeleton className="h-full w-full" />
                     </div>
                 </div>
             </Webline>
-            <Adverts positionName="footer" withGapBottom withWebline />
```
tvikito, line 39: Is this removal really wanted?

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
```diff
@@ -25,18 +27,42 @@ export const OrderLayout: FC<OrderLayoutProps> = ({ activeStep, isTransportOrPay
...
+const OrderLayoutContentSkeleton: FC = ({ children }) => {
+    return (
+        <>
             <Webline>
-                <OrderSteps activeStep={activeStep} domainUrl={url} />
+                <Skeleton className="h-10 w-full" />
             </Webline>
             <Webline>
                 <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
                     <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
```
tvikito: Why do you need children here? If I understand correctly you don't want to display content until it's loaded, right?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
-            {isErrorPopupVisible && (
-                <ErrorPopup
-                    onCloseCallback={() => setErrorPopupVisibility(false)}
-                    fields={transportAndPaymentValidationMessages}
-                    gtmMessageOrigin={GtmMessageOriginType.transport_and_payment_page}
-                />
-            )}
+            <OrderLayout
+                activeStep={2}
+                isTransportOrPaymentLoading={Boolean(isTransportSelectionLoading) || isPaymentSelectionLoading}
```
tvikito: I know it's not your change, but can we remove Boolean? Since it's already boolean

```suggestion
                isTransportOrPaymentLoading={isTransportSelectionLoading || isPaymentSelectionLoading}
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+            >
+                {isTransportAndPaymentSkeletonVisible ? (
+                    <Skeleton className="h-64 w-full" />
+                ) : (
+                    <>
+                        {!!transportsData?.transports.length && (
+                            <TransportAndPaymentSelect
+                                transports={transportsData.transports}
+                                lastOrderTransportUuid={lastOrderData?.lastOrder?.transport.uuid ?? null}
+                                lastOrderPaymentUuid={lastOrderData?.lastOrder?.payment.uuid ?? null}
```
tvikito: Can we work only with `undefined` or defined value? 

 ```suggestion
                                lastOrderTransportUuid={lastOrderData?.lastOrder?.transport.uuid}
                                lastOrderPaymentUuid={lastOrderData?.lastOrder?.payment.uuid}
```

File: project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx [hunk 6]
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+                            buttonNext={t('Contact information')}
+                            hasDisabledLook={
+                                hasValidationErrors(validationMessages) ||
+                                isTransportSelectionLoading ||
+                                isPaymentSelectionLoading
+                            }
+                            isLoading={
+                                (isTransportSelectionLoading || isPaymentSelectionLoading) &&
+                                transport !== null &&
+                                payment !== null
```
tvikito:
```suggestion
                                transport &&
                                payment
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
         </>
     );
 };
+
+const getLastOrderPickupPlace = (
+    lastOrder: LastOrderFragmentApi | undefined | null,
+    packeteryPickupPoint: ListedStoreFragmentApi | null,
+    pickupPlaceData: StoreQueryApi | undefined,
+): ListedStoreFragmentApi | null => {
+    if (lastOrder?.pickupPlaceIdentifier === undefined || lastOrder.pickupPlaceIdentifier === null) {
```
tvikito:
```suggestion
    if (!lastOrder?.pickupPlaceIdentifier) {
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+        return null;
+    }
+
+    if (packeteryPickupPoint?.identifier === lastOrder.pickupPlaceIdentifier) {
+        return packeteryPickupPoint;
+    }
+
+    if (pickupPlaceData?.store !== undefined && pickupPlaceData.store !== null) {
```
tvikito:
```suggestion
    if (!pickupPlaceData?.store) {
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+const getTransportAndPaymentValidationMessages = (
+    transport: Maybe<TransportWithAvailablePaymentsAndStoresFragmentApi>,
+    pickupPlace: Maybe<ListedStoreFragmentApi>,
+    payment: Maybe<SimplePaymentFragmentApi>,
+    paymentGoPayBankSwift: Maybe<string>,
+    t: Translate,
+) => {
+    const errors: Partial<TransportAndPaymentErrorsType> = {};
+
+    if (transport === null) {
```
tvikito:
```suggestion
    if (!transport) {
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+        errors.transport = {
+            name: 'transport',
+            label: t('Choose transport'),
+            errorMessage: t('Please select transport'),
+        };
+    } else {
+        if (transport.isPersonalPickup && pickupPlace?.identifier === undefined) {
```
tvikito:
```suggestion
        if (transport.isPersonalPickup && !pickupPlace?.identifier) {
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+            errors.transport = {
+                name: 'transport',
+                label: t('Choose transport'),
+                errorMessage: t('Please select transport with a personal pickup place'),
+            };
+        }
+        if (payment === null) {
```
tvikito:
```suggestion
        if (!payment) {
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+            errors.payment = {
+                name: 'payment',
+                label: t('Choose payment'),
+                errorMessage: t('Please select payment'),
+            };
+        }
+        if (payment?.goPayPaymentMethod?.identifier === 'BANK_ACCOUNT' && paymentGoPayBankSwift === null) {
```
tvikito:
```suggestion
        if (payment?.goPayPaymentMethod?.identifier === 'BANK_ACCOUNT' && !paymentGoPayBankSwift) {
```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +95,122 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
+                            />
+                        )}
+                    </>
+                )}
+            </OrderLayout>
         </>
     );
 };
+
+const getLastOrderPickupPlace = (
```
tvikito: Can we move these functions into `helpers.tsx`?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
```diff
@@ -309,3 +201,13 @@ const ResetButton: FC<ResetButtonProps> = ({ text, dataTestId, onClick }) => (
         <ArrowIcon className="ml-2" />
     </button>
 );
+
+const getPickupPlaceDetail = (
```
tvikito: Maybe better place would be in `helpers.tsx`?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
```diff
@@ -309,3 +201,13 @@ const ResetButton: FC<ResetButtonProps> = ({ text, dataTestId, onClick }) => (
...
+    selectedTransport: Maybe<TransportWithAvailablePaymentsAndStoresFragmentApi>,
+    selectedPickupPlace: ListedStoreFragmentApi | null,
+    transportItem: TransportWithAvailablePaymentsAndStoresFragmentApi,
+) =>
+    selectedTransport?.uuid === transportItem.uuid &&
+    transportItem.stores?.edges?.some((storeEdge) => storeEdge?.node?.identifier === selectedPickupPlace?.identifier)
+        ? selectedPickupPlace
+        : null;
```
tvikito: Can we work with `undefined` and `null` use only when it's necessary for the types which usually only comes from BE API?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSkeletonContext.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const initialContextValue: TransportAndPaymentSkeletonContextType = {
+    isTransportAndPaymentSkeletonVisible: true,
+    // eslint-disable-next-line no-empty-function, @typescript-eslint/no-empty-function
+    setIsLoadingInitialTransportAndPaymentData: () => {},
+    // eslint-disable-next-line no-empty-function, @typescript-eslint/no-empty-function
+    setIsLoadingTransportAndPaymentFromLastOrder: () => {},
+};
+
+const TransportAndPaymentSkeletonContext = createContext<TransportAndPaymentSkeletonContextType>(initialContextValue);
```
tvikito: Why use Context and not hooks with Zustand? I would avoid using Context until there's reason for it. And even though I tried to find reason for it, I couldn't find any. Also Zustand usage is much cleaner then using Context.

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
sebaholesz: Sure 😇 moved...

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
sebaholesz: Its not removed, git just doesn't understand me moving it:

image

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
sebaholesz: Its just a layout and each page will probably require a different skeleton (transport and payment vs. contact information). I think it's an elegant solution 🤔

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Yep, changed...

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Changed

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Changed.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Changed.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Changed.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Changed.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Changed.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Changed.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: We can, moved.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
sebaholesz: Hmmm... could be. I moved it. Let's see if we like it or not 👍

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
sebaholesz: `pickupPlace` itself can be null. To accommodate for only working with undefined, I had to add the exclamation mark:

image

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: After doing the change (and also required subsequent changes), I dropped it and decided not to do it right now. I think we need to handle this more generally.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSkeletonContext.tsx
sebaholesz: To be honest I dont have a clear candidate. The reason for me was that this way it is clearly separated from the rest of the application. However, I agree that a global store is much easier to work with and it does not require component restructuring just to make the context available. I rewrote it to Zutand, then. However, let's still discuss it in a FE meeting.

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
```diff
@@ -25,17 +27,23 @@ export const OrderLayout: FC<OrderLayoutProps> = ({ activeStep, isTransportOrPay
...
+                    <Webline>
+                        <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
+                            <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
+                            <div className="w-full vl:max-w-md">
+                                <OrderSummary isTransportOrPaymentLoading={isTransportOrPaymentLoading} />
+                            </div>
+                        </div>
+                    </Webline>
+                </>
+            )}
```
tvikito: Maybe you can remove Webline from skeleton and from content shown after loading is finished and simply wrap to Webline all of this?

```suggestion
            <Webline>
            {isLoading ? (
                <OrderLayoutContentSkeleton>{children}</OrderLayoutContentSkeleton>
            ) : (
                <>
                        <OrderSteps activeStep={activeStep} domainUrl={url} />
                        
                        <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
                            <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
                            <div className="w-full vl:max-w-md">
                                <OrderSummary isTransportOrPaymentLoading={isTransportOrPaymentLoading} />
                            </div>
                        </div>
                </>
            )}
            </Webline>
```

project-base/storefront/components/Layout/OrderLayout/OrderLayoutContentSkeleton.tsx
```diff
@@ -0,0 +1,20 @@
...
+            </Webline>
+            <Webline>
+                <div className="mb-24 flex w-full flex-col flex-wrap vl:mt-7 vl:mb-16 vl:flex-row">
+                    <div className="mb-16 w-full vl:mb-0 vl:min-h-[61vh] vl:flex-1 vl:pr-10">{children}</div>
+                    <div className="h-60 w-full vl:max-w-md">
+                        <Skeleton className="h-full w-full" />
+                    </div>
+                </div>
+            </Webline>
+        </>
```
tvikito: Let's remove all those Webline and either use one or even better, according to my previous comment, put it in parent component.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -1,117 +1,64 @@
...
-                    name: 'goPaySwift',
-                    label: t('Choose your bank'),
-                    errorMessage: t('Please select your bank'),
-                };
-            }
-        }
+    const validationMessages = useMemo(
+        () => getTransportAndPaymentValidationMessages(transport, pickupPlace, payment, paymentGoPayBankSwift, t),
+        [transport, pickupPlace, payment, paymentGoPayBankSwift],
+    );
```
tvikito: I think useMemo doesn't help anything here. Since it's being run after every change which can happen on the page anyway, or?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -120,58 +67,52 @@ export const TransportAndPaymentContent: FC<TransportAndPaymentContentProps> = (
...
-                />
-            )}
+            <OrderLayout
+                activeStep={2}
+                isTransportOrPaymentLoading={isTransportSelectionLoading || isPaymentSelectionLoading}
+                isLoading={isTransportAndPaymentSkeletonVisible}
+            >
+                {isTransportAndPaymentSkeletonVisible ? (
+                    <Skeleton className="h-64 w-full" />
+                ) : (
```
tvikito: I still think this can be improved. What about this?

```suggestion
            <OrderLayout
                activeStep={2}
                isTransportOrPaymentLoading={isTransportSelectionLoading || isPaymentSelectionLoading}
                isLoading={isTransportAndPaymentSkeletonVisible}
                skeleton={<Skeleton className="h-64 w-full" />}
            >
               // content itself
```

And then adjust proper Skeleton there

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
```diff
@@ -40,139 +37,21 @@ const TEST_IDENTIFIER = 'pages-order-';
...
-    const getPickupPlaceDetail = (transportItem: TransportWithAvailablePaymentsAndStoresFragmentApi) =>
-        transport?.uuid === transportItem.uuid &&
-        transportItem.stores?.edges?.some((storeEdge) => storeEdge?.node?.identifier === pickupPlace?.identifier)
-            ? pickupPlace
-            : null;
+    const { changePayment, changeGoPaySwift, resetPaymentAndGoPayBankSwift } =
+        usePaymentChangeInSelect(changePaymentInCart);
+    const { preSelectedTransport, changeTransport, changePickupPlace, closePickupPlacePopup } =
+        useTransportChangeInSelect(transports, lastOrderPickupPlace, changeTransportInCart, {
+            onTransportReset: async () => changePayment(null),
```
tvikito: Something seems not right here. Can you please recheck all those async calls?

I think if you remove async here it will still function the same since you are not awaiting `changePayment` which is asynchronous. Either remove async are await changePayment. Or refactor this logic.

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
```diff
@@ -0,0 +1,356 @@
...
+
+        if (!updatedTransport) {
+            await changeTransportHandler(null, null);
+
+            if (callbacks.onTransportReset) {
+                await callbacks.onTransportReset();
+            }
+
+            return false;
+        }
```
tvikito: I guess it makes more sense to switch those two checks since it's better to check if `updatedTransport` is present and then compare it

```suggestion
        if (!updatedTransport) {
            await changeTransportHandler(null, null);

            if (callbacks.onTransportReset) {
                await callbacks.onTransportReset();
            }

            return false;
        }
        
        if (updatedTransport.uuid === currentTransport?.uuid) {
            return false;
        }
```

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
```diff
@@ -0,0 +1,356 @@
...
+        useState<TransportWithAvailablePaymentsAndStoresFragmentApi | null>(null);
+    const clearPacketeryPickupPoint = usePersistStore((store) => store.clearPacketeryPickupPoint);
+    const setPacketeryPickupPoint = usePersistStore((store) => store.setPacketeryPickupPoint);
+    const { transport: currentTransport, pickupPlace: currentPickupPlace } = useCurrentCart();
+
+    const changeTransport = async (updatedTransportUuid: string | null): Promise<boolean> => {
+        const updatedTransport = transports?.find((transport) => transport.uuid === updatedTransportUuid);
+
+        if (updatedTransport?.uuid === currentTransport?.uuid) {
+            return false;
```
tvikito: This doesn't make much sense to me. Why do we return boolean for handler function...? Why do we need to return anything? 🤔 I tried to understand from the code but seems even dirtier 🙈

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
```diff
@@ -0,0 +1,356 @@
...
+        if (payment?.goPayPaymentMethod?.identifier === 'BANK_ACCOUNT' && !paymentGoPayBankSwift) {
+            errors.goPaySwift = {
+                name: 'goPaySwift',
+                label: t('Choose your bank'),
+                errorMessage: t('Please select your bank'),
+            };
+        }
+    }
+
+    return errors;
```
tvikito: Maybe better readable? I will leave it on you.

```
    const errors: Partial<TransportAndPaymentErrorsType> = {};

    if (!transport) {
        errors.transport = {
            name: 'transport',
            label: t('Choose transport'),
            errorMessage: t('Please select transport'),
        };
        
        return errors;
    } 
    
        if (transport.isPersonalPickup && !pickupPlace?.identifier) {
            errors.transport = {
                name: 'transport',
                label: t('Choose transport'),
                errorMessage: t('Please select transport with a personal pickup place'),
            };
        }
        if (!payment) {
            errors.payment = {
                name: 'payment',
                label: t('Choose payment'),
                errorMessage: t('Please select payment'),
            };
        }
        if (payment?.goPayPaymentMethod?.identifier === 'BANK_ACCOUNT' && !paymentGoPayBankSwift) {
            errors.goPaySwift = {
                name: 'goPaySwift',
                label: t('Choose your bank'),
                errorMessage: t('Please select your bank'),
            };
        }
   

    return errors;
```

project-base/storefront/pages/order/transport-and-payment.tsx
```diff
@@ -41,28 +19,10 @@ const TransportAndPaymentPage: FC<ServerSidePropsType> = () => {
...
-                    )}
-                </OrderLayout>
-                <Webline type="dark">
-                    <Footer simpleFooter />
-                </Webline>
-            </EmptyCartWrapper>
+            <TransportAndPaymentContent />
+            <Webline type="dark">
+                <Footer simpleFooter />
+            </Webline>
```
tvikito: Shouldn't this be in OrderLayout? 🤔

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Yeah, I think you are right. After analysing how much time it actually takes, I think we can omit it.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
sebaholesz: I had to provide the handler directly, not as a callback, for reasons related to some of the other threads. Because of that this code is not here anymore. But for a future reference, I am adding a lengthy explanation:

I am not sure what you mean. It is true that keeping or removing async is the same, but I left the async to be more clear and transparent about it being an asynchronous operation.

If you check the type of the callback:
image

Then you can see that this is exactly the type of the provided function:
image


The reason, why it is not awaited here, is because of IoC, as the handler is responsible for dispatching and handling the callbacks. From the outside, I can only adhere to the API.

This function is therefore provided, called, awaited:
image

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
sebaholesz: Could be. I switched them.

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
sebaholesz: I confused the handlers. You are right, the one which is defined here does not need to return anything.

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
sebaholesz: Yeah, could be. Changed.

project-base/storefront/pages/order/transport-and-payment.tsx
sebaholesz: Yeah, I think it should. I moved it.

project-base/storefront/components/Layout/OrderLayout/OrderLayoutContentSkeleton.tsx
sebaholesz: I think I did what you requested.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Though it's not the way I would do it (I believe a skeleton is a content as any other) I tried to follow your suggeestions. I also removed the `isLoading` prop and only base my decision on the `contentSkeleton` prop (if its there, I show it, otherwise I dont)

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
```diff
@@ -8,14 +8,21 @@ import { Webline } from 'components/Layout/Webline/Webline';
...
 import { useDomainConfig } from 'hooks/useDomainConfig';
 import { OrderLayoutContentSkeleton } from './OrderLayoutContentSkeleton';
+import { Footer } from '../Footer/Footer';
+import { ReactNode } from 'react';
 
 type OrderLayoutProps = {
     activeStep: number;
     isTransportOrPaymentLoading?: boolean;
-    isLoading?: boolean;
+    contentSkeleton?: ReactNode | null;
```
tvikito: Do we need to put another null type here? Isn't enough just `contentSkeleton?: ReactNode;`?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
tvikito: Nice, I think since you have the skeleton for the page in OrderLayout, the content skeleton can be part of this main skeleton. Especially when it makes the code cleaner.

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
tvikito: I can't find any place which would need to use returned value. Everywhere it is just called without need of the return. If I remove this return there can be some logic refactored. Can you please check again?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
tvikito: OK. Thanks for explanation!

project-base/storefront/components/Layout/OrderLayout/OrderLayout.tsx
sebaholesz: Done

tvikito (approved):

tvikito (approved):

PR #2826, opened 2023-09-22
[monorepo] split monorepo is now done by Github Actions

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Split monorepo to individual repositories is now done by GitHub Actions. Also, the internal tool to rewrite history was changed from `git filter-branch` to `git-filter-repo` (https://github.com/newren/git-filter-repo). See note below. Split feature branches are automatically deleted after the PR is closed.
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes (changes commit hashes in split packages) 
|Fixes issues| ... 

_Note about changing the rewrite history tool:_
Replacing `git filter-branch` with `git-filter-repo` brings a significant performance leap (from 52 minutes to 1 minute). And the resulting history does not contain the empty merge commits (like [this one](https://github.com/shopsys/project-base/commit/e87d28b6cac0bab9dc2daa885e36706acd036151) or [the other one](https://github.com/shopsys/coding-standards/commit/4ae9a211fdb458e16cf7decf4ade5a19fbea7f23))
On the other side, it will be necessary to force-push the protected branches in the individual packages, because the new tool generates different commit hashes and the history is slightly different due to missing empty merge commits.

----
  - https://mg-split-repo.odin.shopsys.cloud
  - https://cz.mg-split-repo.odin.shopsys.cloud


.github/monorepo/split-repositories.sh
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+echo -e "${BLUE}Splitting branch '$SPLIT_BRANCH'...${NC}"
+
+WORKSPACE=`pwd`
+
+if [[ "$FORCE" == true ]]; then
+    PUSH_OPTS="--force"
+else
+    PUSH_OPTS="--tags"
```
TomasLudvik, line 28: Are you sure you want to push all tags always? And that you do not want to push during force?

.github/workflows/clean-up-after-branch-merge.yaml
```diff
@@ -0,0 +1,46 @@
...
+                run: |
+                    BRANCH_NAME=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}
+                    echo "BRANCH_NAME=${BRANCH_NAME}" >> $GITHUB_OUTPUT
+    cancel-review-after-branch-merge:
+        name: Cancel review after branch merge
+        runs-on: [self-hosted, linux, review-stage]
+        needs: variables
+        steps:
+            -   name: GIT checkout branch - ${{ vars.MAIN_BRANCH_NAME }}
+                uses: actions/checkout@v3
```
TomasLudvik: It would be nice to unite version of checkout, ideally through all workflows

.github/monorepo/split-repositories.sh
```diff
@@ -0,0 +1,52 @@
...
+set -u
+
+# Import functions
+. $(dirname "$0")/monorepo_functions.sh
+
+assert_split_branch_variable
+assert_remote_template_variable
+
+if [[ "$FORCE" == true ]]; then
+    assert_split_branch_is_not_protected
```
TomasLudvik, line 18: We use to force split for example 12.0 during release. This will disable it, right?

TomasLudvik (changes requested):

.github/monorepo/split-repositories.sh
grossmannmartin: It will not be necessary anymore to force split those branches as they will be split automatically

.github/monorepo/split-repositories.sh
grossmannmartin: Quite. This option separates two workflows – feature branches (mg-fix-bug, ...) and any of main branches (master, 13.0, ...). 

If we split any of the main branches, we push the tags. When no new tag is introduced, nothing extra happens. When a new tag is created in the process (during the release process), this new tag is propagated. 

If we split the feature branch, we don't expect new tags, but we do not care about the history of branches in packages  – thus `--force`

.github/workflows/clean-up-after-branch-merge.yaml
grossmannmartin: now we use version 4

TomasLudvik (approved):

pk16011990 (approved):

PR #2829, opened 2023-09-24
Fixes regarding default delivery address

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Various things regarding default delivery address were fixed. Check commit messages for more info.
|New feature| No
|Fixes issues| ...

----
  - https://sh-default-delivery-address-in-order.odin.shopsys.cloud
  - https://cz.sh-default-delivery-address-in-order.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/hooks/user/useCurrentUserContactInformation.ts
+++ b/project-base/storefront/hooks/user/useCurrentUserContactInformation.ts
@@ -21,7 +21,10 @@ const mergeContactInformation = (
         const filteredProperty = filteredContactInformationFromStore[key as keyof ContactInformation];
 
         const isEmptyString = typeof filteredProperty === 'string' && filteredProperty.length === 0;
-        if (isEmptyString && key in contactInformationFromApi) {
+        if (
+            (isEmptyString || filteredProperty === null || filteredProperty === undefined) &&
+            key in contactInformationFromApi
+        ) {
             delete filteredContactInformationFromStore[key as keyof ContactInformation];
         }
     }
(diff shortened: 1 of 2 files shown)


project-base/storefront/hooks/user/useCurrentUserContactInformation.ts
```diff
@@ -21,7 +21,7 @@ const mergeContactInformation = (
         const filteredProperty = filteredContactInformationFromStore[key as keyof ContactInformation];
 
         const isEmptyString = typeof filteredProperty === 'string' && filteredProperty.length === 0;
-        if (isEmptyString && key in contactInformationFromApi) {
+        if ((isEmptyString || filteredProperty === null) && key in contactInformationFromApi) {
```
tvikito: Můžeme odebrat check pro `null`?

```suggestion
        if ((isEmptyString || !filteredProperty) && key in contactInformationFromApi) {
```

tvikito (approved):

project-base/storefront/hooks/user/useCurrentUserContactInformation.ts
sebaholesz: I don't think it makes sense. We do not want to filter out all falsy values, just empty strings and nulls.

tvikito: Ah, ok. It was not obvious from code here. But I've checked and I see also `undefined` as possible value, don't you want to include it as well?

sebaholesz: I did, thanks for the suggestion

pk16011990 (approved):

PR #2830, opened 2023-09-25
[shopsys] graphql-bundle classes are now dumped so they can use composer autoload

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| graphql-bundle suggests to use composer autoload for loading its classes - https://github.com/overblog/GraphQLBundle/blob/master/docs/index.md#composer-autoloader-configuration-optional
|New feature| No 
|Fixes issues| ... 

----
  - https://tl-overblog-definitions.odin.shopsys.cloud
  - https://cz.tl-overblog-definitions.odin.shopsys.cloud

### DIFF
--- a/project-base/app/config/packages/overblog_graphql.yaml
+++ b/project-base/app/config/packages/overblog_graphql.yaml
@@ -1,0 +1,21 @@
+overblog_graphql:
+    definitions:
+        schema:
+            query: Query
+            mutation: Mutation
+        mappings:
+            types:
+                -   type: yaml
+                    dir: "%kernel.project_dir%/config/graphql/types"
+        builders:
+            args:
+                -   alias: "PaginatorArgumentsBuilder"
+                    class: App\FrontendApi\Component\Arguments\PaginatorArgumentsBuilder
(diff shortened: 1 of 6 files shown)


project-base/app/config/packages/overblog_graphql.yaml
```diff
@@ -0,0 +1,6 @@
+overblog_graphql:
```
grossmannmartin, line 1: this configuration key is already in the file `project-base/app/config/packages/shopsys_frontend_api.yaml`
It would be nice to unify it

grossmannmartin (changes requested):

project-base/app/config/packages/overblog_graphql.yaml
TomasLudvik: I have split shopsys_frontend_api.yaml to separate files so it is more consistent

TomasLudvik:
test

grossmannmartin (approved):

pk16011990 (approved):

PR #2840, opened 2023-09-27
fixed redis cache on storefront

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The directive name was changed from _redisCache back to redisCache because the former was completely omitted, even on Storefront. Now, a custom createClient is used everywhere with a custom fetch function. This custom function is important as it takes care of directive handling. During Server Side Rendering (SSR), it utilizes the directive and subsequently cleans it on the client. Additionally, two new ESLint forbidden imports were added specifically for default client creation functions. This addition is crucial as using those functions would result in errors with our directive. The term "await" was removed from initServerSideProps where it was previously used for client creation, but it was found to be unnecessary. Lastly, if the Redis client is not provided on the server, an exception is captured (but not thrown) to Sentry. Capturing this exception will assist us in discovering bugs in the logic in the future.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-redis-cache-fix.odin.shopsys.cloud
  - https://cz.sh-redis-cache-fix.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/urql/fetcher.ts
+++ b/project-base/storefront/urql/fetcher.ts
@@ -3,33 +3,48 @@ import md5 from 'crypto-js/md5';
 import { isServer } from 'helpers/isServer';
 import { RedisClientType, RedisModules, RedisScripts } from 'redis';
 
-const CACHE_REGEXP = `@_redisCache\\(\\s?ttl:\\s?([0-9]*)\\s?\\)`;
-const QUERY_NAME_REGEXP = `query\\s([A-z]*)(\\([A-z:!0-9$,\\s]*\\))?\\s@_redisCache`;
+const CACHE_REGEXP = `@redisCache\\(\\s?ttl:\\s?([0-9]*)\\s?\\)`;
+const QUERY_NAME_REGEXP = `query\\s([A-z]*)(\\([A-z:!0-9$,\\s]*\\))?\\s@redisCache`;
 const REDIS_PREFIX_PATTERN = `${process.env.REDIS_PREFIX}:fe:queryCache:`;
 
+const removeDirectiveFromQuery = (query: string) => query.replace(new RegExp(CACHE_REGEXP), '');
+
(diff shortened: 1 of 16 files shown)


project-base/storefront/urql/fetcher.ts
```diff
@@ -3,33 +3,48 @@ import md5 from 'crypto-js/md5';
 import { isServer } from 'helpers/isServer';
 import { RedisClientType, RedisModules, RedisScripts } from 'redis';
 
-const CACHE_REGEXP = `@_redisCache\\(\\s?ttl:\\s?([0-9]*)\\s?\\)`;
-const QUERY_NAME_REGEXP = `query\\s([A-z]*)(\\([A-z:!0-9$,\\s]*\\))?\\s@_redisCache`;
-const REDIS_PREFIX_PATTERN = `${process.env.REDIS_PREFIX}:fe:queryCache:`;
+const CACHE_REGEXP = `@redisCache\\(\\s?ttl:\\s?([0-9]*)\\s?\\)`;
+const QUERY_NAME_REGEXP = `query\\s([A-z]*)(\\([A-z:!0-9$,\\s]*\\))?\\s@redisCache`;
+const REDIS_PREFIX_PATTERN = () => `${process.env.REDIS_PREFIX}:fe:queryCache:`;
```
tvikito: Why this needs to be function? 🤔

tvikito: If this need to be function let's put proper function name like `getRedisPrefixPattern` otherwise it doesn't make sense to me why I need to call this (`REDIS_PREFIX_PATTERN`) constant or in this case just something what suppose to be constant but is function actually.

sebaholesz: I renamed it according to your suggestion.

tvikito (approved):

pk16011990 (approved):

PR #2851, opened 2023-10-03
fix Zásilkovna order delivery

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| in case user choose Zásilkovna delivery type (personal pickup) and then wanted to include different delivery information (only phone, name and surname is allowed since delivery place is choosed accroding to user choosen Zásilkovna place) he got validation error for state, city, etc. which were not even shown in the form.
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1857-fix-zisilkovna-order-delivery.odin.shopsys.cloud
  - https://cz.tv-fw-1857-fix-zisilkovna-order-delivery.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
+++ b/project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
@@ -51,7 +51,6 @@ export const ContactInformationDeliveryAddress: FC = () => {
 
                 setValue(formMeta.fields.deliveryFirstName.name, formValues.firstName, { shouldValidate: true });
                 setValue(formMeta.fields.deliveryLastName.name, formValues.lastName, { shouldValidate: true });
-                setValue(formMeta.fields.deliveryCompanyName.name, formValues.companyName, { shouldValidate: true });
                 setValue(formMeta.fields.deliveryTelephone.name, formValues.telephone, { shouldValidate: true });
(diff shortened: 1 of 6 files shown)


project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
```diff
@@ -51,7 +51,6 @@ export const ContactInformationDeliveryAddress: FC = () => {
 
                 setValue(formMeta.fields.deliveryFirstName.name, formValues.firstName, { shouldValidate: true });
                 setValue(formMeta.fields.deliveryLastName.name, formValues.lastName, { shouldValidate: true });
-                setValue(formMeta.fields.deliveryCompanyName.name, formValues.companyName, { shouldValidate: true });
```
sebaholesz, line 54: Why was it removed? I think it would be nice adding motivation for changes to commit messages, as then I would not have to return this back to you to get an answer 🤔

sebaholesz (changes requested):

project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
tvikito: Because we don't need (we don't want) this information to be set. It's for the case when user choose pick up place delivery.

sebaholesz (approved):

pk16011990 (approved):

PR #2855, opened 2023-10-04
multiple tabs logged out user behavior

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Reload the page in case user is logged in or logged out in different tab. When user log out in one tab in others he is still marked as logged in, since we rely on API endpoint for this check, but the problem is other tabs don't know that user has logged out. In case cached result from API is that user is logged in/logged out but we don't/do have tokens in cookies me we refresh the page so the log in status is correct.
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1708-multiple-tabs-logged-out-user-behavior.odin.shopsys.cloud
  - https://cz.tv-fw-1708-multiple-tabs-logged-out-user-behavior.odin.shopsys.cloud


project-base/storefront/hooks/cart/useReloadCart.ts
```diff
@@ -16,12 +18,22 @@ export const useReloadCart = (): void => {
...
+
+            if ((isUserLoggedIn && !isWithUserTokens) || (!isUserLoggedIn && isWithUserTokens)) {
+                router.reload();
+            } else if (cartUuid || isUserLoggedIn) {
+                client.query(CartQueryDocumentApi, { cartUuid }, { requestPolicy: 'network-only' }).toPromise();
+            }
+        };
+
+        func();
+    }, [slug]);
```
sebaholesz: Seems like it doesnt have to be async and can be thus simplified

sebaholesz (changes requested):

project-base/storefront/hooks/cart/useReloadCart.ts
tvikito: Yup, I forgot this after some debugging. Refactored.

PR #2863, opened 2023-10-06
Add instant skeletons

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Add instant skeletons to the pages. After click on the link there is instant response for proper page skeleton. 
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1711-add-instant-skeletons.odin.shopsys.cloud
  - https://cz.tv-fw-1711-add-instant-skeletons.odin.shopsys.cloud


project-base/storefront/components/Blocks/Product/ProductsList/ProductsList.tsx
```diff
@@ -3,7 +3,7 @@ import { DEFAULT_PAGE_SIZE } from 'config/constants';
 import { CategoryDetailFragmentApi, ListedProductFragmentApi } from 'graphql/generated';
 import { createEmptyArray } from 'helpers/arrayUtils';
 import { GtmMessageOriginType, GtmProductListNameType } from 'gtm/types/enums';
-import { ProductListItemSkeleton } from './ProductListItemSkeleton';
+import { SkeletonProductListItem } from '../../Skeleton/SkeletonProductListItem';
```
sebaholesz: I think we should not use this type of relative imports. I thought that we had a rule enforcing this, but it was removed (god knows why).

project-base/storefront/components/Blocks/Skeleton/SkeletonPageStore.tsx
```diff
@@ -1,7 +1,7 @@
 import { Webline } from 'components/Layout/Webline/Webline';
 import Skeleton from 'react-loading-skeleton';
 
-export const StorePageSkeleton: FC = () => (
+export const SkeletonPageStore: FC = () => (
```
sebaholesz, line 5: I must say that some of these names have a weird word order. What is the key we are using here?

sebaholesz: Because for example I see `SkeletonProductListItem`, which makes complete sense to me. It starts with skeleton and is followed by a description of the actual skeleton target. But here it is counterintuitive for me.

project-base/storefront/components/Pages/BlogCategory/BlogCategoryArticlesWrapper.tsx
```diff
@@ -6,7 +6,7 @@ import { createEmptyArray } from 'helpers/arrayUtils';
 import { mapConnectionEdges } from 'helpers/mappers/connection';
 import { useQueryParams } from 'hooks/useQueryParams';
 import { RefObject, useMemo } from 'react';
-import { BlogArticleSkeleton } from '../BlogArticle/BlogArticleSkeleton';
+import { SkeletonArticleBlog } from '../../Blocks/Skeleton/SkeletonArticleBlog';
```
sebaholesz: Again with the relative import

project-base/storefront/components/Pages/Search/SearchContent.tsx
```diff
@@ -7,7 +7,7 @@ import { mapConnectionEdges } from 'helpers/mappers/connection';
 import useTranslation from 'next-translate/useTranslation';
 import { useRouter } from 'next/router';
 import { useMemo } from 'react';
-import { CategoryDetailPageSkeleton } from '../CategoryDetail/CategoryDetailPageSkeleton';
+import { SkeletonPageCategoryDetail } from '../../Blocks/Skeleton/SkeletonPageCategoryDetail';
```
sebaholesz: And here

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -2,16 +2,32 @@ import { SLUG_TYPE_QUERY_PARAMETER_NAME } from 'helpers/queryParamNames';
...
+export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
+    children,
+    href,
+    type,
+    queryParams,
+    as,
+    onClick,
+    ...props
+}) => {
+    const isStatic = type === 'static' || type === 'homepage' || type === 'stores';
```
sebaholesz, line 29: I would maybe use the type you created above `STATIC_PAGES.includes(type)`

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }, [isPageLoading]);
+
+    if (!isPageLoading && !isFetchingData) {
+        return <>{children}</>;
+    }
+
+    switch (redirectPageType) {
+        case 'homepage':
+            // TODO add proper skeleton
```
sebaholesz: Could we maybe solve it here and not leave as a TODO?

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
```diff
@@ -0,0 +1,50 @@
...
+        case 'brand':
+        case 'flag':
+            return <SkeletonPageCategoryDetail />;
+        case 'article':
+        case 'blogArticle':
+            return <SkeletonPageArticle />;
+        case 'blogCategory':
+            return <SkeletonPageBlogCategory />;
+        case 'stores':
+            // TODO add proper skeleton
```
sebaholesz: Maybe here as well?

project-base/storefront/store/slices/createPageStateSlice.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type PageStateSlice = {
+    isPageLoading: boolean | null;
+    isFetchingData: boolean | null;
+
+    updateIsPageLoading: (value: boolean | null) => void;
+    updateIsFetchingData: (value: boolean | null) => void;
+};
+
+export const createPageStateSlice: StateCreator<PageStateSlice> = (set) => ({
```
sebaholesz: This is not used anywhere...

sebaholesz: And even if it was, isn't it confusing to have pageLoading in two slices? Same with fetching data... I think you actually wanted to remove one of them.

project-base/storefront/pages/categories/[categorySlug].tsx
```diff
@@ -64,12 +62,9 @@ const CategoryDetailPage: NextPage = () => {
...
                 breadcrumbsType="category"
+                isFetchingData={isSkeletonVisible}
             >
-                {isSkeletonVisible ? (
-                    <CategoryDetailPageSkeleton />
-                ) : (
-                    !!categoryData && <CategoryDetailContent category={categoryData} />
-                )}
+                {!!categoryData && <CategoryDetailContent category={categoryData} />}
             </CommonLayout>
```
Comment (sebaholesz) [inline project-base/storefront/pages/categories/[categorySlug].tsx]: You can easily solve your problem with loading by implementing the changes seen below. However, I think the same problem might be also in other places (where we show skeleton based on fetching but then have another condition on the actual visibility, such as here). In my opinion, those places were not discovered yet only because there is no client-side link to it in the app. I think it should be handled in this PR to get rid of this double problem.


image

sebaholesz (changes requested):

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
tvikito: Yup, this was my first implementation, but unfortunately TS is not smart enough here... 😏 It complains about this way later in the code because it can't properly exclude values which are not static.

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
tvikito: Ah, I forgot to push commits for this 🙈 It's there now.

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
tvikito: Ah, I forgot to push commits for this 🙈 It's there now.

project-base/storefront/store/slices/createPageStateSlice.ts
tvikito: Ah, nice catch. Forgotten after playing around with some other solutions. Removed.

project-base/storefront/pages/categories/[categorySlug].tsx
Comment (tvikito) [reply project-base/storefront/pages/categories/[categorySlug].tsx]: Nope. Unfortunately flashing is still happening...

project-base/storefront/components/Blocks/Skeleton/SkeletonPageStore.tsx
tvikito: My idea behind this naming is not about proper english, rather it's about better visibility for what is page skeleton to not mix it. Other solution can be split then into separate folders. But in case we don't want folders I wouldn't care about proper english ordering since is help with grouping types of skeletons. We can discuss this with other guys.

sebaholesz: Yeah, okay. TBH in this I do not care that much, so if you have a reason behind it, that's okay.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
sebaholesz: OMG... okay 😆

project-base/storefront/pages/categories/[categorySlug].tsx
Comment (sebaholesz) [reply project-base/storefront/pages/categories/[categorySlug].tsx]: Yeah, in the end there was one more issue. But should be solved now.

project-base/storefront/components/Blocks/BlogPreview/BlogPreview.tsx
```diff
@@ -46,15 +49,19 @@ export const BlogPreview: FC = () => {
                 )}
             </div>
 
-            <div className="flex flex-col gap-16 vl:flex-row  vl:justify-between">
-                {!!blogMainItems && (
-                    <div className="flex flex-1 gap-6 vl:gap-16">
-                        <BlogPreviewMain articles={blogMainItems} />
-                    </div>
-                )}
+            {!!(fetchingArticles || fetchingBlogUrl) && <SkeletonMagazine />}
```
sebaholesz: This seems like a boolean already. Why do you need to cast it?

project-base/storefront/components/Blocks/BlogPreview/BlogPreview.tsx
```diff
@@ -46,15 +49,19 @@ export const BlogPreview: FC = () => {
...
 
-                {!!blogSideItems && <BlogPreviewSide articles={blogSideItems} />}
-            </div>
+            {!fetchingArticles && !fetchingBlogUrl && !!(blogMainItems || blogSideItems) && (
```
sebaholesz, line 54: First part of the condition seems unnecessary as it should be caught with the condition above.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -43,7 +43,10 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
             }
             onClick={(e) => {
                 onClick?.(e);
-                updatePageLoadingState({ isPageLoading: true, redirectPageType: type });
+
+                if (type !== 'static' && !e.ctrlKey && !e.metaKey) {
```
sebaholesz: Is this here to disallow this for "open in new tab" clicks? If so, maybe taking the condition out to a variable like "isNotOpenInNewTab" would be better for future reference.

sebaholesz (changes requested):

project-base/storefront/store/slices/createPageStateSlice.ts
sebaholesz: Not removed, still there.

tvikito: Aha, glitch in the Matrix. Removed now.

project-base/storefront/components/Blocks/BlogPreview/BlogPreview.tsx
tvikito: Actually nope. This is not conditional rendering where it is either this or that. If I remove this condition it will display skeleton and content during fetching.

project-base/storefront/components/Blocks/BlogPreview/BlogPreview.tsx
tvikito: Ah, sure. Refactored.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
tvikito: Yup, I guess that's improvement. Refactored. I've added one more condition to not even run onClick function from the props, which I've realized is also expected behavior.

sebaholesz (approved):

PR #2865, opened 2023-10-06
[monorepo] UPGRADE and CHANGELOG files are only in recent version

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Having all historical changelog entries and upgrade files makes the repository more and more bloated. That makes the search for the relevant information hard. This PR leverages the fact that we started to develop the version in separate branches (e.g., version `12.1.0-dev` is developed in branch `12.1`, version `13.0.0-dev` in branch `13.0`) and leaves only info we need to have for the branch. It also strips the need to synchronize the files between  branches\*.
|New feature| No 
|Fixes issues| ... 

\* for example when we did a new patch version in the 12.1 branch, we had to add the appropriate upgrade file into the 13.0 branch, along with the changelog entries, and keep them in sync if some changes occur.


UPGRADE-12.1.md
```diff
@@ -0,0 +1,47 @@
+# UPGRADING FROM 12.0 to 12.1
```
TomasLudvik, line 1: I thought we want to keep upgrade from previous major and all current minors.

TomasLudvik (changes requested):

UPGRADE-12.1.md
grossmannmartin: added back

TomasLudvik (approved):

pk16011990 (approved):

PR #2868, opened 2023-10-09
[infrastructure] added translations dump check

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| We want to test, that every new change added to our source has its translations dumped.
|New feature| No 
|Fixes issues| ... 

----
  - https://tl-translations-dump-check.odin.shopsys.cloud
  - https://cz.tl-translations-dump-check.odin.shopsys.cloud


.github/workflows/docker-build.yaml
```diff
@@ -463,6 +463,44 @@ jobs:
...
+                    name: docker-compose
+            -   name: Build application
+                run: docker compose up -d php-fpm storefront
+            -   name: Check translations are dumped
+                run: |
+                    docker compose exec -T php-fpm composer install --optimize-autoloader --no-interaction
+                    docker compose exec -T php-fpm php phing translations-dump
+                    docker compose cp php-fpm:/var/www/html/ .
+                    cp -a ./html/. ./
+                    rm -rf ./html
```
grossmannmartin, line 494: I assume this is due to fact that we don't have a volume in the docker-compose file? 
Can you explain please the reason for second cp and rm?

grossmannmartin (approved):

.github/workflows/docker-build.yaml
TomasLudvik: We have discussed this via Slack. There is no way to copy only the content of the directory using docker compose cp, so it is necessary to do it like this.

pk16011990 (approved):

PR #2871, opened 2023-10-11
[FW-1706] Refactoring of various error-related matters on SF

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR mainly aims to improve how we work with status codes and how we understand error handling. 
|New feature| No
|Fixes issues| ... 

----
  - https://sh-serverside-error-codes.odin.shopsys.cloud
  - https://cz.sh-serverside-error-codes.odin.shopsys.cloud

### DIFF
--- a/UPGRADE-14.0.md
+++ b/UPGRADE-14.0.md
@@ -117,3 +117,9 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   added missing skeletons for Homepage and Stores
     -   adjustments to current skeletons to match the design of a page better
     -   see #project-base-diff to update your project
+-   refactoring of various error-related matters on SF ([#2871](https://github.com/shopsys/shopsys/pull/2871))
+    -   the goal was to shine light on some of the not-well-understood places in regard of error handling on SF
+    -   for you to get the most out of this PR, you should check `error-handling.md` in SF docs, which is a direct result of this PR
+    -   it contains explanations and tips on how to improve error handling in your SF code as well
+    -   for not-discussed changes, see #project-base-diff to update your project
+
--- a/docs/storefront/error-handling.md
+++ b/docs/storefront/error-handling.md
@@ -1,0 +1,148 @@
+# Error handling on Storefront
+
+## The `logException` function
+
+This function will be your friend while logging exceptions anywhere in the app. It checks the current environemnt and based on it logs it to the console (development) and sends the error to Sentry. You should use it to make sure the errors are correctly displayed both in the console and in Sentry.
+
+## Run-time error on the server (`getServerSideProps` or `getInitialProps`)
+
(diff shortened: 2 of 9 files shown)


tvikito (approved):

vitek-rostislav:
Feedback from testing:

- [ ] `resolve-friendly-url` script can hypothetically return any return code so I suggest taking this into account in `middleware.ts` - I would respect the returned code and pass it in `MIDDLEWARE_STATUS_CODE_KEY`. For 4XX codes, it can still return the message 'Friendly URL page not found' 
- [ ] `middleware.ts` - It is not safe to pass the `JSON.stringify(e)` in the `MIDDLEWARE_STATUS_MESSAGE_KEY` header as it might expose sensitive data. I would rather use the `logException` function.
- [ ] `middleware.ts` - the `JSON.stringify(e)` piece of code mentioned above does not work as expected - in the logs, I see just `Middleware runtime error ({}})` - but this should be probably resolved by the point above
- [ ] It would be nice to describe the `logException` function in the documentation and motivate developers to use it

vitek-rostislav (changes requested):
Feedback from testing:

- [x] `resolve-friendly-url` script can hypothetically return any return code so I suggest taking this into account in `middleware.ts` - I would respect the returned code and pass it in `MIDDLEWARE_STATUS_CODE_KEY`. For 4XX codes, it can still return the message 'Friendly URL page not found' 
- [x] `middleware.ts` - It is not safe to pass the `JSON.stringify(e)` in the `MIDDLEWARE_STATUS_MESSAGE_KEY` header as it might expose sensitive data. I would rather use the `logException` function.
- [x] `middleware.ts` - the `JSON.stringify(e)` piece of code mentioned above does not work as expected - in the logs, I see just `Middleware runtime error ({}})` - but this should be probably resolved by the point above
- [x] It would be nice to describe the `logException` function in the documentation and motivate developers to use it

tvikito (approved):

vitek-rostislav:
Hi, @sebaholesz, thanks for the fixes :ok_hand: I noticed one more thing that might be addressed (but I do not insist on that, I suggest you discuss this with @pk16011990. We have the following code in `_error.tsx`:
```
const ErrorPage: NextPage<ErrorPageProps> = ({ hasGetInitialPropsRun, err, statusCode }): ReactElement => {
    if (!hasGetInitialPropsRun && err) {
        logException(err);
    }

    return statusCode === 404 ? <Error404Content /> : <Error500Content />;
};
```

That means, even 410 errors (which now can be returned from the middleware) will be rendered as 500 error page. So the page currently looks like this:
image

It is essential to mention that the HTTP code is correctly set to 410.
For me, a good enough solution would be:
```diff
- return statusCode === 404 ? <Error404Content /> : <Error500Content />;
+ const notFoundStatusCodes = [404, 410];
    
+ return notFoundStatusCodes.includes(statusCode) ? <Error404Content /> : <Error500Content />;
```

However, the question is, what about the other status codes, do we want to care about them? :slightly_smiling_face:

pk16011990 (approved):

pk16011990 (changes requested):
missing upgrade notes

pk16011990 (approved):

vitek-rostislav (approved):

PR #2872, opened 2023-10-12
[shopsys] fixed project-base on packagist

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| after moving the php app into the `app/` subfolder, the packagist was not able to find the composer.json file. This PR adds a symlink to it, while preventing running the `composer install` from the root
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-fix-project-base-packagist.odin.shopsys.cloud
  - https://cz.mg-fix-project-base-packagist.odin.shopsys.cloud

### DIFF
--- a/project-base/app/check-composer-json.sh
+++ b/project-base/app/check-composer-json.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+if [ -h "composer.json" ]; then
+    echo 'Please run composer install in the app/ subfolder.'
+    exit 1
+fi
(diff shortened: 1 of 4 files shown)


vitek-rostislav (approved):
nice :ok_hand: just a question - why do I see the `app/composer.json` marked with red? :smile: I thought it meant that it is a symlink but the script is not marked, so I am confused :thinking: 
image

grossmannmartin:
> nice 👌 just a question - why do I see the `app/composer.json` marked with red? 😄 I thought it meant that it is a symlink but the script is not marked, so I am confused 🤔 image

I'm not quite sure... my guess is because the file is named *.json and the content is not a json... but the content seems all right. for example bin symlink has similar content https://github.com/shopsys/shopsys/blob/14.0/bin

TomasLudvik (approved):

pk16011990 (approved):

PR #2875, opened 2023-10-13
[infrastructure] commands not used during installation, standards checking and running tests are now run on GitHub Actions to ensure all our commands are passing

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| We want to test as much of our application automatically as possible. This PR ensures that all our commands are passing.
|New feature| No 
|Fixes issues| ... 

----
  - https://tl-phing-targets-ci.odin.shopsys.cloud
  - https://cz.tl-phing-targets-ci.odin.shopsys.cloud

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -506,6 +506,36 @@ jobs:
                         echo "All translations are dumped";
                         exit 0;
                     fi
+    check-console-commands:
+        name: Check that all console commands that are not used during installation are passing
+        needs: build-php-fpm-image
+        if: |
+            always() && !failure() && !cancelled() &&
+            (
+                github.event.pull_request.head.repo.full_name == 'shopsys/shopsys' ||
+                github.ref_protected == true
+            ) &&
+            (needs.build-php-fpm-image.result == 'success' || needs.build-php-fpm-image.result == 'skipped')
+        runs-on: ubuntu-22.04
+        steps:
+            -   name: GIT checkout branch - ${{ github.ref }}
+                uses: actions/checkout@v4
+                with:
+                    ref: ${{ github.ref }}
+            -   name: Download docker-compose.yml from artifacts
+                uses: actions/download-artifact@v3
+                with:
+                    name: docker-compose
+            -   name: Build application
+                run: |
+                    docker compose up -d
+                    docker compose exec -T php-fpm composer install --optimize-autoloader --no-interaction --dev
+                    docker compose exec -T php-fpm php phing -D production.confirm.action=y db-create frontend-api-generate-new-keys build-demo-dev-quick error-pages-generate elasticsearch-index-recreate elasticsearch-export
+            -   name: Check all commands are passing
+                run: |
+                    docker compose exec -T php-fpm php phing -D production.confirm.action=y cron-list cron-run-all-serially cron-watch maintenance-on maintenance-off elasticsearch-export-changed
+                    docker compose exec -T php-fpm php bin/console shopsys:categories:recalculate
+                    docker compose exec -T php-fpm php bin/console shopsys:cdn-domain-url:replace
     build-fork-docker-images:
         if: |
             github.event.pull_request.head.repo.full_name != 'shopsys/shopsys' && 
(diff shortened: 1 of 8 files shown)


build.xml
```diff
@@ -38,10 +38,6 @@
         </exec>
     </target>
 
-    <target name="cron" description="Disabled - use new cron instances instead.">
```
vitek-rostislav, line 41: I would remove the "Should be executed periodically by system Cron every 5 minutes." from the description in the framework package as it is not true anymore IMHO :slightly_smiling_face:

docs/introduction/cron.md
```diff
@@ -31,8 +31,13 @@ The instance of cron is actually a named group of cron jobs.
 
 You can learn how to set up multiple cron instances in [Working with Multiple Cron Instances](../cookbook/working-with-multiple-cron-instances.md) cookbook.
 
+!!! note
+    By default it is not possible to run all crons at once.
```
vitek-rostislav: I would rewrite the whole note :innocent: 
```
There is a special phing target `run-all-crons-serially` that (as the name suggests) allows you to run all the CRON modules serially. This is rather for testing reasons (we use it on our CI to check that no CRON is failing) and there is no point in using this target in production.
```

.github/workflows/docker-build.yaml
```diff
@@ -558,6 +591,14 @@ jobs:
...
                         exit 0;
                     fi
+            -   name: Check all not previously used commands are passing
+                run: |
+                    docker compose exec -T php-fpm php phing -D production.confirm.action=y frontend-api-enable cron-list cron-run-all-serially cron-watch maintenance-on maintenance-off elasticsearch-export-changed
+                    docker compose exec -T php-fpm php bin/console shopsys:categories:recalculate
+                    docker compose exec -T php-fpm php bin/console shopsys:cdn-domain-url:replace
+                    docker compose exec -T php-fpm php bin/console shopsys:clean-opcache
+                    docker compose exec -T php-fpm php bin/console shopsys:database:dump dump.sql
+                    docker compose exec -T php-fpm php bin/console shopsys:domains:info
```
vitek-rostislav: is not it possible to wrap all these lines into a file or script so you do need to repeat yourself? ... now I see there is one difference - `frontend-api-enable` - this is on purpose?

vitek-rostislav (changes requested):

vitek-rostislav:
One more thing - should not we add upgrade notes for removing the overridden `cron` target from `project-base/app/build.xml`? :slightly_smiling_face:

build.xml
TomasLudvik: Good catch

docs/introduction/cron.md
TomasLudvik: Changed to something more straightforward

.github/workflows/docker-build.yaml
TomasLudvik: I know, that our GitHub actions workflows can be lengthy. Still, in this case, I prefer to have that code directly visible when editing such files, because click-through is not present there, and searching and opening included files feels to be annoying.

vitek-rostislav (approved):

vitek-rostislav (changes requested):
Hi, while testing, I noticed there are a few commands that are already used elsewhere so I think we do not need to test them here separately:

- `shopsys:domains:info` (used in translations dump)
- `shopsys:database:dump` (used in acceptance tests build)
- `shopsys:clean-opcache` (used as a dependency of the `clean-cache` phing target which is used e.g. in check of DB schema)

vitek-rostislav (approved):

UPGRADE-14.0.md
```diff
@@ -96,3 +96,5 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
             -   react/self-closing-comp
             -   react/jsx-sort-props
             -   no-relative-import-paths/no-relative-import-paths
+-   add cron phing target ([#2875](https://github.com/shopsys/shopsys/pull/2875))
```
pk16011990: Is not removal?

pk16011990 (changes requested):

UPGRADE-14.0.md
TomasLudvik: I have reworded it

pk16011990 (approved):

PR #2876, opened 2023-10-13
[project-base] repeat order improvements

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR does not allow to pay order if it was already paid, or too many transactions is created. Also extends the amount of information returned from CheckPaymentStatusMutation. It also implements repeated payment on SF plus minor things regarding payments (see commit messages).
|New feature| Yes 
|Fixes issues| ... 

----
  - https://mg-repeat-order-impro.odin.shopsys.cloud
  - https://cz.mg-repeat-order-impro.odin.shopsys.cloud


project-base/app/src/FrontendApi/Mutation/Payment/Exception/OrderCannotBePaidUserError.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\FrontendApi\Mutation\Payment\Exception;
+
+use Overblog\GraphQLBundle\Error\UserError;
+use Shopsys\FrontendApiBundle\Model\Error\UserErrorWithCodeInterface;
+
+class OrderCannotBePaidUserError extends UserError implements UserErrorWithCodeInterface
```
TomasLudvik: I feel like this error is missing reason why not

project-base/app/src/Model/Order/Order.php
```diff
@@ -200,6 +201,18 @@ public function getPaymentTransactions(): array
...
     }
 
+    /**
+     * @return bool
+     */
+    public function canBePaid(): bool
+    {
+        if ($this->paymentTransactions->count() >= 2) {
+            return false;
+        }
```
TomasLudvik: For me it is not clear why. Can you please elaborate?

project-base/app/src/FrontendApi/Mutation/Payment/PaymentMutation.php
```diff
@@ -47,17 +47,20 @@ public function payOrderMutation(Argument $argument): PaymentSetupCreationData
 
     /**
      * @param \Overblog\GraphQLBundle\Definition\Argument $argument
-     * @return bool
+     * @return array{isPaid: bool, transactionCount: int}
      */
-    public function checkPaymentStatusMutation(Argument $argument): bool
+    public function checkPaymentStatusMutation(Argument $argument): array
```
TomasLudvik: I feel like check method should return bool, so can you change name to get?

TomasLudvik (changes requested):

project-base/app/src/FrontendApi/Mutation/Payment/PaymentMutation.php
sebaholesz: If we change it to get, it should not be a mutation then... Or maybe it should, as it probably checks with the external service?

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
```diff
@@ -1,21 +1,29 @@
...
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
+import { Button } from 'components/Forms/Button/Button';
+import useTranslation from 'next-translate/useTranslation';
+import { SpinnerIcon } from 'components/Basic/Icon/IconsSvg';
 
 type GoPayGatewayProps = {
     orderUuid: string;
+    requiresAction?: boolean;
+    initButtonText?: string;
```
tvikito: Maybe `initialButtonText` would be better naming? Since `init` can also be verb which is not the case.

project-base/storefront/components/Pages/Order/PaymentConfirmation/PaymentConfirmationElements.tsx
```diff
@@ -14,8 +14,14 @@ export const PaymentWrapper: FC = ({ children }) => (
...
+    if (message) {
+        return (
+            <div className="text-center lg:text-left" dangerouslySetInnerHTML={{ __html: message }}>
+                {children}
+            </div>
+        );
+    }
+
+    return <div className="text-center lg:text-left">{children}</div>;
+};
```
tvikito: What about

```suggestion
export const Message: FC<MessageProps> = ({ children, message }) => {
        return (
            <div className="text-center lg:text-left" dangerouslySetInnerHTML={{ __html: message || '' }}>
                {children}
            </div>
        );
 };
+

```

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
```diff
@@ -102,6 +102,24 @@ export const TransportAndPaymentSelect: FC<TransportAndPaymentSelectProps> = ({
                     />
                 }
             />
+            {paymentItem.uuid === payment?.uuid &&
+                payment.type === 'goPay' &&
+                payment.goPayPaymentMethod?.identifier === 'BANK_ACCOUNT' && (
```
tvikito: Can we have separated variable for this so JSX is be better readable?

project-base/storefront/pages/order-payment-confirmation.tsx
```diff
@@ -8,45 +9,64 @@ import { getServerSidePropsWrapper } from 'helpers/serverSide/getServerSideProps
...
-                !gtmCreateOrderEventUserPart
-            ) {
-                return;
-            }
+    const checkPaymentStatus = async () => {
+        const checkPaymentStatusActionResult = await checkPaymentStatusAction({ orderUuid: orderUuidParam });
+
+        const { gtmCreateOrderEventOrderPart, gtmCreateOrderEventUserPart } = getGtmCreateOrderEventFromLocalStorage();
+        if (
+            checkPaymentStatusActionResult.data?.CheckPaymentStatus === undefined ||
```
tvikito:
```suggestion
            !checkPaymentStatusActionResult.data?.CheckPaymentStatus ||
```

project-base/app/src/FrontendApi/Mutation/Payment/Exception/OrderCannotBePaidUserError.php
grossmannmartin: I'm not sure about it. That way we should either change error name, or create a new one when a new reason why order cannot be paid occurs.

project-base/app/src/Model/Order/Order.php
grossmannmartin: It's requested to work that way - see the acceptance criteria

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
sebaholesz: Good idea, changed.

project-base/storefront/components/Pages/Order/PaymentConfirmation/PaymentConfirmationElements.tsx
sebaholesz: Unfortunately no. If we provide both `dangerouslySetInnerHTML` and `children`, the app crashes. https://react.dev/reference/react-dom/components/common#common-caveats

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentSelect/TransportAndPaymentSelect.tsx
sebaholesz: I also put the select in a separate variable.

project-base/storefront/pages/order-payment-confirmation.tsx
sebaholesz: Fixed

project-base/app/src/FrontendApi/Mutation/Payment/PaymentMutation.php
grossmannmartin: I agree with the Seba. It's a mutation, so I would not name it with the get prefix

TomasLudvik: In that case `updatePaymentStatusMutation`?

project-base/app/src/FrontendApi/Mutation/Payment/Exception/OrderCannotBePaidUserError.php
TomasLudvik: My point is: user does not know what is wrong and cannot act accordingly.

project-base/app/src/FrontendApi/Mutation/Payment/PaymentMutation.php
grossmannmartin: renamed to updatePaymentStatusMutation along with the usages

project-base/app/src/FrontendApi/Mutation/Payment/Exception/OrderCannotBePaidUserError.php
grossmannmartin: I have separated errors

TomasLudvik (approved):

tvikito (approved):

pk16011990 (changes requested):
Order with successful payment by credit card has an empty page with Unknown error. However the order is marked as paid in administration and I can't reproduce it - see https://mg-repeat-order-impro.odin.shopsys.cloud/order-payment-confirmation?orderIdentifier=610dd2b2-4651-40e8-8040-f3ec6b6c3db1&id=3211652548 https://mg-repeat-order-impro.odin.shopsys.cloud/admin/order/edit/42
image

grossmannmartin:
> Order with successful payment by credit card has an empty page with Unknown error. However the order is marked as paid in administration and I can't reproduce it

I looked at the container's logs, but there's nothing there. 
however, I tried it locally and on the production server (with the same URL) and I got this error 

Snímek obrazovky 2023-10-31 v 21 16 26

project-base/app/src/Model/Order/Order.php
```diff
@@ -200,6 +201,14 @@ public function getPaymentTransactions(): array
         return $this->paymentTransactions->getValues();
     }
 
+    /**
+     * @return bool
+     */
+    public function isMaxTransactionCountReached(): bool
+    {
+        return $this->paymentTransactions->count() >= 2;
```
TomasLudvik: Can we convert the int to constant please?

project-base/app/config/graphql/types/ModelType/Payment/PaymentStatus.types.yaml
```diff
@@ -0,0 +1,10 @@
+PaymentStatus:
+    type: object
+    config:
+        fields:
+            isPaid:
+                type: "Boolean!"
+                description: "Whether the order is already paid or not"
+            transactionCount:
+                type: "Int!"
+                description: "Count of already processed transactions"
```
TomasLudvik, line 10: Would not be better to also send info about the limit of maximal counts, so the storefront does not try to do another payment, if it is no longer possible?

packages/framework/src/Model/Order/OrderRepository.php
```diff
@@ -97,6 +97,25 @@ public function getById($id)
...
+     * @param string $uuid
+     * @return \Shopsys\FrameworkBundle\Model\Order\Order
+     */
+    public function getByUuid(string $uuid): Order
+    {
+        $order = $this->createOrderQueryBuilder()
+            ->andWhere('o.uuid = :uuid')
+            ->setParameter(':uuid', $uuid)
+            ->getQuery()
+            ->getOneOrNullResult();
```
TomasLudvik: What is the benefit of doing it this way and not to use findOneBy?

packages/framework/src/Migrations/Version20231101182132.php
```diff
@@ -0,0 +1,67 @@
...
+     * @param string $pageContent
+     */
+    private function setPageContent(string $settingName, string $pageContent): void
+    {
+        $paymentSuccessfulPageContent = $this->sql(
+            'SELECT COUNT(*) FROM setting_values WHERE name = :settingName AND domain_id = 1;',
+            [
+                'settingName' => $settingName,
+            ],
+        )->fetchOne();
```
TomasLudvik: What about the other domains?

TomasLudvik (changes requested):

project-base/app/config/graphql/types/ModelType/Payment/PaymentStatus.types.yaml
grossmannmartin: I can't tell. It's no problem to add the field with the max count, but I'm not sure if it's useful on the storefront. 

@sebaholesz can you please let me know whether I should implement this?

packages/framework/src/Migrations/Version20231101182132.php
grossmannmartin: other domains are now handled in 60cd31f0b4

packages/framework/src/Model/Order/OrderRepository.php
grossmannmartin: fixed in 28ace46c68

project-base/app/src/Model/Order/Order.php
grossmannmartin: Yes. Yes, we can. 10fb9145a8

project-base/app/config/graphql/types/ModelType/Payment/PaymentStatus.types.yaml
TomasLudvik: It is, @sebaholesz has already implemented his own constant in order-payment-confirmations.tsx for this purpose, so I think it would be better to set it in one place and sent via API

sebaholesz: Yeah, it would make sense. Having it in two places seems not correct.

UPGRADE-14.0.md
```diff
@@ -105,13 +105,40 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 -   add consumers and RabbitMQ to deployed application ([#2904](https://github.com/shopsys/shopsys/pull/2904))
     -   set new environment variables `RABBITMQ_DEFAULT_USER`, `RABBITMQ_DEFAULT_PASS`, `RABBITMQ_IP_WHITELIST` in your deployment tool (with use of the default config it will be Gitlab CI)
     -   see #project-base-diff to update your project
--   re-enable phing target cron ([#2875](https://github.com/shopsys/shopsys/pull/2875))
-    -   see #project-base-diff to update your project
+-   re-enable phing target cron ([#2875](https://github.com/shopsys/shopsys/pull/2875)) - see #project-base-diff to update your project
+    <<<<<<< HEAD
```
tvikito: Seems like some fault during merge

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
```diff
@@ -0,0 +1,43 @@
...
+        return <img alt={t('Your payment was successful')} src="/public/frontend/images/sent-cart.svg" />;
+    })();
+
+    const HeadingText = (() => {
+        if (type === 'order-created') {
+            return t('Your order was created');
+        } else if (type === 'payment-failed') {
+            return t('Your payment was not successful');
+        }
+        return t('Your payment was successful');
```
tvikito: What about use switch here?

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const ConfirmationPageContent: FC<ConfirmationPageContentProps> = ({ type, content, AdditionalContent }) => {
+    const { t } = useTranslation();
+
+    const Image = (() => {
+        if (type === 'order-created') {
+            return <img alt={t('Your order was created')} src="/public/frontend/images/sent-cart.svg" />;
+        } else if (type === 'payment-failed') {
+            return <img alt={t('Your payment was not successful')} src="/public/frontend/images/sent-cart.svg" />;
+        }
+        return <img alt={t('Your payment was successful')} src="/public/frontend/images/sent-cart.svg" />;
```
tvikito: What about use switch here?

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    return (
+        <div className="mt-16 mb-10 flex flex-col items-center justify-center lg:mt-16 lg:mb-24 lg:flex-row">
+            <div className="mb-0 w-40 lg:mr-32">{Image}</div>
+            <div>
+                <div className="h1 mb-3">{HeadingText}</div>
+                {!!content && (
+                    <div className="text-center lg:text-left" dangerouslySetInnerHTML={{ __html: content }} />
+                )}
+                {!!content && !!AdditionalContent && AdditionalContent}
```
tvikito: Why do we display this component at all if we don't have `content`? Can we add condition in parent component?

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
```diff
@@ -1,21 +1,29 @@
...
+    const [initiatedPaymentGate, setInitiatedPaymentGate] = useState(!requiresAction);
     const [goPayPaymentSetup, setGoPayPaymentSetup] = useState<GoPayCreatePaymentSetupApi | undefined>(undefined);
     const [, payOrder] = usePayOrderMutationApi();
+    const { t } = useTranslation();
+    const wasPaidRef = useRef(false);
 
     useEffect(() => {
+        if (wasPaidRef.current || !initiatedPaymentGate) {
+            return;
+        }
```
tvikito: As I've already explain to you somewhere else this way of handling `useEffect` function is good practice in my opinion. Since useEffect is not classic function where you would use this pattern it should only return cleanup function. If you can wrap the whole code into the condition why not to wrap it? I will leave it on your choice.

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
```diff
@@ -1,21 +1,29 @@
...
+
         payOrder({ orderUuid }).then((payOrderResult) => {
```
tvikito: What will happen in case `payOrder` will throw an error? Don't we want to handle those cases as well?

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
```diff
@@ -38,21 +46,20 @@ export const GoPayGateway: FC<GoPayGatewayProps> = ({ orderUuid }) => {
...
-                    {t('Pay with GoPay')}
-                </Link>
+            {initiatedPaymentGate && !!goPayPaymentSetup && (
+                <Script
+                    id="go-pay-embedded-js"
+                    src={goPayPaymentSetup.embedJs}
+                    onLoad={initGoPayCheckout(goPayPaymentSetup.gatewayUrl)}
+                />
+            )}
+            {requiresAction && (
```
tvikito, line 56: Tiny thing. Can we separate this with one line, please? Since all FE agreed on this would be great if you would have it on your mind also for other places. Visibility is much better, at least for most of us.

project-base/storefront/pages/order-payment-confirmation.tsx
```diff
@@ -1,72 +1,70 @@
...
-                return;
-            }
+    const { isPaymentStatusFetching, paymentStatusData } = useUpdatePaymentStatus(
+        orderUuid,
+        orderPaymentStatusPageValidityHashParam,
+    );
+    const { content, isContentFetching } = usePaymentConfirmationContent(
+        orderUuid,
+        paymentStatusData?.UpdatePaymentStatus,
+    );
```
tvikito: I we can't have those two hooks in one hook? Since they are connected together, it seems to me like it is not necessary to have them separated. I guess it would also simplify the logic here.

tvikito (commented):
Honestly, I can't say I like the solution with PaymentFail and PaymentSuccess components and connected logic. Seems to me like one big net which is not solved the best possible way. But if you feel comfortable about the logic and code I will trust you that it was not possible to solve it better way, since I would need to experiment on my own to make sure my point is correct.

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
sebaholesz: Done

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
sebaholesz: Done

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
sebaholesz: I am not sure what you mean. The error is handled right below. If you are asking about the `.catch` clause, then from what I understand, the mutation function cannot directly throw. If they do, it is not a request error.

image

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
sebaholesz: I completely removed it as a part of refactoring

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
sebaholesz: I completely removed it as a part of refactoring

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
sebaholesz: So the point is to show something even after 5 minutes. The user has 5 minutes to display the sensitive content (with order details etc) and then he can display a very generic message afterwards.

UPGRADE-14.0.md
sebaholesz: Thanks, removed.

project-base/storefront/pages/order-payment-confirmation.tsx
sebaholesz: I combined them and removed isPaymentStatusFetching as it is allways shorter-lived than isContentFetching

tvikito (approved):

project-base/app/config/graphql/types/ModelType/Payment/PaymentStatus.types.yaml
vitek-rostislav: done in https://github.com/shopsys/shopsys/pull/2876/commits/c06e35f10f18b88a67a6c85ad4dd1a2e1b3a4dc6
@TomasLudvik @sebaholesz could you please review it?

vitek-rostislav: done in c86b1abc8
@sebaholesz @TomasLudvik could you please review it?

project-base/storefront/pages/order-payment-confirmation.tsx
```diff
@@ -4,20 +4,20 @@ import { Webline } from 'components/Layout/Webline/Webline';
...
 import Skeleton from 'react-loading-skeleton';
 
-const MAX_ALLOWED_PAYMENT_TRANSACTIONS = 2;
-
 const OrderPaymentConfirmationPage: FC<ServerSidePropsType> = () => {
     const { t } = useTranslation();
     const { orderIdentifier, orderPaymentStatusPageValidityHash } = useRouter().query;
     const orderUuid = getStringFromUrlQuery(orderIdentifier);
     const orderPaymentStatusPageValidityHashParam = getStringFromUrlQuery(orderPaymentStatusPageValidityHash);
+    const [{ data }] = useSettingsQueryApi({ requestPolicy: 'cache-first' });
```
sebaholesz: `cache-first` is the default value, you do not need to provide it explicitly.

project-base/storefront/components/Pages/Order/PaymentConfirmation/helpers.ts
```diff
@@ -77,11 +62,11 @@ export const usePaymentConfirmationContent = (
     };
 
     useEffect(() => {
-        if (!paymentStatusData) {
-            return;
+        if (!wasUpdatedPaymentStatusRef.current) {
+            updatePaymentStatus();
+            wasUpdatedPaymentStatusRef.current = true;
```
tvikito: Why do we have this ref here? I am guessing it is connected with strict mode double rendering...?

project-base/storefront/components/Pages/Order/PaymentConfirmation/helpers.ts
```diff
@@ -30,6 +32,7 @@ export const useUpdatePaymentStatus = (orderUuid: string, orderPaymentStatusPage
         ) {
             return;
         }
+        fetchContent(updatePaymentStatusActionResult.data.UpdatePaymentStatus.isPaid);
```
tvikito: Why can't we fetch data in with genereted hooks in each Payment `PaymentSuccess` and `PaymentFail` componenent? I guess it is made as generic here because of loading state, right? Why not handle loading state properly in those components?

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
```diff
@@ -20,22 +20,20 @@ export const GoPayGateway: FC<GoPayGatewayProps> = ({ orderUuid, requiresAction,
...
 
-                return;
-            }
-
-            setGoPayPaymentSetup(payOrderResult.data?.PayOrder.goPayCreatePaymentSetup ?? undefined);
-        });
-        wasPaidRef.current = true;
+                setGoPayPaymentSetup(payOrderResult.data?.PayOrder.goPayCreatePaymentSetup ?? undefined);
+            });
+            wasPaidRef.current = true;
```
tvikito, line 35: Why do we need ref here?

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
```diff
@@ -20,22 +20,20 @@ export const GoPayGateway: FC<GoPayGatewayProps> = ({ orderUuid, requiresAction,
     const wasPaidRef = useRef(false);
 
     useEffect(() => {
-        if (wasPaidRef.current || !initiatedPaymentGate) {
-            return;
-        }
+        if (!wasPaidRef.current && initiatedPaymentGate) {
+            payOrder({ orderUuid }).then((payOrderResult) => {
+                if (payOrderResult.error?.graphQLErrors !== undefined) {
```
tvikito: if (payOrderResult.error?.graphQLErrors) {

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
```diff
@@ -38,21 +44,20 @@ export const GoPayGateway: FC<GoPayGatewayProps> = ({ orderUuid }) => {
...
+                    id="go-pay-embedded-js"
+                    src={goPayPaymentSetup.embedJs}
+                    onLoad={initGoPayCheckout(goPayPaymentSetup.gatewayUrl)}
+                />
+            )}
+            {requiresAction && (
+                <Button className={className} onClick={() => setInitiatedPaymentGate(true)}>
+                    {initiatedPaymentGate && <SpinnerIcon className="w-5" />}
+                    {initiatedPaymentGate ? t('You are being redirected...') : initialButtonText}
+                </Button>
```
tvikito:
```suggestion
                <Button className={className} onClick={() => setInitiatedPaymentGate(true)}>
                    {initiatedPaymentGate 
                      ? <>
                           <SpinnerIcon className="w-5" />
                           t('You are being redirected...')
                      </> 
                     : initialButtonText}
                </Button>
```

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
```diff
@@ -0,0 +1,24 @@
...
+        <div className="mt-16 mb-10 flex flex-col items-center justify-center lg:mt-16 lg:mb-24 lg:flex-row">
+            <div className="mb-0 w-40 lg:mr-32">
+                <img alt={heading} src="/public/frontend/images/sent-cart.svg" />
+            </div>
+            <div>
+                <div className="h1 mb-3">{heading}</div>
+                {!!content && (
+                    <div className="text-center lg:text-left" dangerouslySetInnerHTML={{ __html: content }} />
+                )}
+                {!!content && !!AdditionalContent && AdditionalContent}
```
tvikito:
```suggestion
                {!!content && AdditionalContent}
```

No need for check since undefined will not be shown in the DOM.

TomasLudvik (approved):

project-base/storefront/components/Pages/Order/PaymentConfirmation/helpers.ts
sebaholesz: I did it this way, because we need to control one skeleton while loading the status and then a second one while loading the content. This way we can handle it in one place. There are other approaches I have considered, such as:

- show a skeleton for status and a different one for content (they would look the same, but would be controlled from a different component)
- have one loading state which would be controlled from two components (by callbacks or something)

Do you consider any of them to be better?

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
sebaholesz: It creates a payment transaction on BE every time it is called (the `payOrder` mutation). This way it is only called once.

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
sebaholesz: Fixed

project-base/storefront/components/Pages/Order/PaymentConfirmation/Gateways/GoPayGateway.tsx
sebaholesz: Why not, changed.

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
sebaholesz: Done and moved under the previous condition.

project-base/storefront/components/Pages/Order/PaymentConfirmation/helpers.ts
sebaholesz: Yes. There are some events which happen when the status is updated. I tried to rewrite the code so it can be run twice.

project-base/storefront/components/Pages/Order/PaymentConfirmation/helpers.ts
tvikito: I personally would vote for third option which is handling those two loading state separately. Since it actually is two different states and I don't like this `content` prop. Actual solution is causing complexity and is more difficult to debug. 

The loading transition shouldn't be even visible in my understanding. Since after load the page fetching for content should start immediately with the component render. And big benefit is that we can use generated urql hook and don't care about handling those states at all.

project-base/storefront/components/Blocks/ConfirmationPage/ConfirmationPageContent.tsx
tvikito: Nice

project-base/storefront/components/Pages/Order/PaymentConfirmation/helpers.ts
sebaholesz: Fixed according to our discussion on a call.

tvikito: Nice, me personally I would call this as nice improvement :)

pk16011990 (approved):

PR #2888, opened 2023-10-18
[FW-1879] removed unnecessary default value for domainConfig

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The value is now initially set to undefined, not to first-domain data. If the value is not available when used, an exception is logged (to Sentry). The value is null-assured inside the useDomainConfig hook as we know we are setting it correctly before.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-remove-default-domain-config.odin.shopsys.cloud
  - https://cz.sh-remove-default-domain-config.odin.shopsys.cloud


tvikito (approved):

pk16011990 (changes requested):
Please add upgrade notes.

pk16011990 (approved):

PR #2895, opened 2023-10-20
[FW-1678] search on search page is now not called if search query is empty

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Search on search page is now not called if search query is empty, which is required because the API treats empty string as null and thus throws an error.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-empty-search-query-bug.odin.shopsys.cloud
  - https://cz.sh-empty-search-query-bug.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/public/locales/cs/common.json
+++ b/project-base/storefront/public/locales/cs/common.json
@@ -360,6 +360,7 @@
     "The transport you selected is no longer available.": "Doprava, kterou jste si zvolili, již není dostupná.",
     "There are no products in the shared wishlist.": "Sdílený seznam oblíbených produktů je prázdný.",
     "There are no products in the wishlist. Add some first.": "Seznam oblíbených produktů je prázdný. Nejprve přidejte nějaké produkty do oblíbených.",
+    "There are no results as you have searched with an empty query...": "Vaše vyhledávání nevrátilo žádné výsledky, protože jste nezadali žádný dotaz...",
     "There was an error while adding a promo code to the order.": "Během přidávání slevového kupónu do objednávky se vyskytla chyba.",
     "There was an error while changing your password": "Při změně Vašeho hesla došlo k chybě",
     "There was an error while deleting your delivery address": "Při mazání dodací adresy došlo k chybě",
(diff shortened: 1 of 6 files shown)


project-base/storefront/public/locales/cs/common.json
```diff
@@ -360,6 +360,7 @@
     "The transport you selected is no longer available.": "Doprava, kterou jste si zvolili, již není dostupná.",
     "There are no products in the shared wishlist.": "Sdílený seznam oblíbených produktů je prázdný.",
     "There are no products in the wishlist. Add some first.": "Seznam oblíbených produktů je prázdný. Nejprve přidejte nějaké produkty do oblíbených.",
+    "There are no results as you have searched with an empty query...": "Vaše vyhledávání nevrátilo žádné výsledky, protože jste nezadali žádný dotaz...",
```
tvikito, line 363: Is it really necessary to have another text for this? Don't we have and cannot we show a regular message "No search results"?

tvikito (approved):

project-base/storefront/public/locales/cs/common.json
sebaholesz: After the discussion and consideration, I decided to leave the message. I base my decision on the following things:

1. `<SearchContent />` accepts a `searchData` object, which we would have to mock in order to achieve an empty result (this is pretty extensive, as "empty products" also need an info on filters)
2. Because the empty products also need info about filters, it made me think about why... It is because if we search for some products (also using filters), it might not return any results for that specific combination of search query and filters and then we want to show the invalid query and filters to the user. In this case, there are no filters. The user just searched with an invalid query and there is no info about the filters, so it doesnt make sense to show the filter panel at all
3. You asked if we need a special message for this case and I think it makes sense, because this is indeed a special case. It is the only `pause` condition and so I think it makes sense to treat it in a special way.image
 

Let me know if you agree 🙂

tvikito: No problem, agree :)

pk16011990 (approved):

PR #2901, opened 2023-10-24
[framework] [frontend-api] universal product list

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR|  At the moment, there are two product list functionalities - wishlist and comparison. Both are implemented slightly differently but basically, they do the same thing. We want to unify the implementation and make it general so it is easier to add new product lists in the future. Also, having one common codebase for the product lists makes it easier to maintain. 
|New feature| Kind of :slightly_smiling_face:  
|Fixes issues| ... 

----
  - https://rv-fw-1880-product-list.odin.shopsys.cloud
  - https://cz.rv-fw-1880-product-list.odin.shopsys.cloud


docs/storefront/caching/graphcache.md
```diff
@@ -0,0 +1,156 @@
...
+-   Details how cache **updates** should occur after certain mutations.
+-   Uses **optimistic updates** to immediately reflect changes in the UI before a server's response is received, providing a smoother user experience.
+
+### Official Documentation
+
+For a comprehensive understanding and in-depth details on Graphcache, its benefits, and its API, please refer to the [official URQL documentation](https://formidable.com/open-source/urql/docs/graphcache/). Some of the things, however, are not explained in the best way and might seem chaotic. Especially when it comes to manual updates. For tips on how to manually update the cache after mutations, read the part of this documentation dedicated to it.
+
+## Keys
+
+-   `keyNull`, `keyWishlist`, `keyUuid`, `keyName`, `keyCode`, `keyUrl`, `keyComparison`: These are keying functions used to generate keys for various GraphQL types in the cache. They are based on properties like `uuid`, `name`, `code`, `url`, etc.
```
vitek-rostislav: I would omit the `keyComparison` as it does not exist anymore

docs/storefront/caching/graphcache.md
```diff
@@ -0,0 +1,156 @@
...
+If you can read from the cache this way, you can just mimic this behavior by directly reading and not loopic through all queries.
+
+## Optimistic Cache Updates (located in `omptimistic.ts`)
+
+### Examples
+
+-   `getOptimisticChangeTransportInCartResult` & `getOptimisticChangePaymentInCartResult`: These functions provide optimistic results for the respective mutations.
+-   `getPaymentFromTransport`: Helper function to get a specific payment method from a transport based on its UUID.
+
+```
```
vitek-rostislav: this adds an empty code block - was there supposed to be something there?

project-base/storefront/graphql/requests/productLists/fragments/ComparisonFragment.graphql
```diff
@@ -1,4 +1,4 @@
-fragment ProductComparisonFragment on Comparison{
+fragment ProductComparisonFragment on ProductList {
```
tvikito: I am not sure, if this is the best way to think about this topic. I know the way we think about this is that Comparison needs some more information compared to ProductList so we do have two fragments. But is it so different so we can't have only fragment and same structure for every list? I would probably rather have one structure for every list.

project-base/storefront/graphql/requests/productLists/mutations/AddProductToComparisonMutation.graphql
```diff
@@ -0,0 +1,5 @@
+mutation AddProductToComparisonMutation($input: ProductListUpdateInput!) {
```
tvikito: Then we can have also only one mutation. Can't we have only one mutation for ProductList? 🤔

project-base/storefront/hooks/productLists/comparison/useComparison.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const useComparison = () => {
+    const { t } = useTranslation();
+    const isUserLoggedIn = useIsUserLoggedIn();
+
+    const [, addProductToListMutation] = useAddProductToComparisonMutationApi();
+    const [, removeProductFromListMutation] = useRemoveProductFromComparisonMutationApi();
+    const [, cleanListMutation] = useCleanProductListMutationApi();
+
+    const comparisonUuid = usePersistStore((store) => store.comparisonUuid);
```
tvikito: That's pretty big question, maybe I am missing some understanding. Seems like we are complicating things on FE. Since on BE we have only one API and on FE we are basically splitting it into two APIs/logics back again. 

Why don't we have type next to uuid coming from API and then we just filter out proper type and use its uuid? Then we can have only one value in store like `productListUuids` and store uuids with proper type. I guess it's connected with having two different fragments/structures, but seems to me like huge improvement in case we share with BE only one API because now we need to consider the type almost everywhere... So to me it seems like we have done half of the possible improvements as was done on BE.

project-base/storefront/hooks/productLists/comparison/useComparison.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }, [comparisonData?.productList?.uuid]);
+
+    const { cleanList, isProductInList, toggleProductInList } = useProductList(
+        ProductListTypeEnumApi.ComparisonApi,
+        comparisonUuid,
+        comparisonData,
+        {
+            cleanList: cleanListMutation,
+            removeProductFromList: removeProductFromListMutation,
+            addProductToList: addProductToListMutation,
```
tvikito: The same topic which I've pointed out in previous comments. This is just breaking DRY principle. This can be implemented into `useProductList`. Since it's the same on both Wishlist and Comparison.

project-base/storefront/hooks/productLists/comparison/useComparison.tsx
```diff
@@ -0,0 +1,78 @@
...
+                updateComparisonUuid(null);
+            },
+            removeError: () => showErrorMessage(t('Unable to remove product from comparison.')),
+            removeSuccess: (result) => {
+                if (!result) {
+                    updateComparisonUuid(null);
+                }
+
+                showSuccessMessage(t('Product has been removed from your comparison.'));
+            },
```
tvikito: This can be also implemented within `useProductList`. We know the type so why not implement it there....?

project-base/storefront/hooks/productLists/useProductListUuids.ts
```diff
@@ -0,0 +1,31 @@
...
+        return productListsUuids;
+    };
+
+    const removeAllProductListUuids = () => {
+        updateWishlistUuid(null);
+        updateComparisonUuid(null);
+    };
+
+    return { getAllProductListUuids, removeAllProductListUuids };
+};
```
tvikito: Again, logic split here. I know we probably won't add more list in the near future, but imagine we have 10 types of the list, how many stores/ifs will be here? 🙈

In case of one ProductListUuids store we can skip all this logic.

project-base/storefront/urql/cache/updates.ts
```diff
@@ -0,0 +1,134 @@
...
+    cache.updateQuery({ query: query, variables: { input: args } }, (data) => ({
+        ...data,
+        __typename: 'ProductList',
+        productList: null,
+    }));
+};
+
+const manuallyUpdateProductListQuery = (input: ProductListInputApi, result: DataField, cache: Cache) => {
+    const query =
+        input.type === ProductListTypeEnumApi.WishlistApi ? WishlistQueryDocumentApi : ComparisonQueryDocumentApi;
```
tvikito: Nope. This is not good approach in order of this refactoring. We don't want to refactor this after adding third or next type. Let's refactor this so it is prepared for next list.

project-base/storefront/graphql/requests/productLists/fragments/ComparisonFragment.graphql
sebaholesz: As discussed, I agree with you. After going through it once again, I moved the layer of generic code up, now making basically everything generic. You can check [the commit, in which I do it](https://github.com/shopsys/shopsys/pull/2901/commits/40f0a750f24c4d5c83d4bfc417cf203a2727053d) and the modified upgrade notes.

project-base/storefront/graphql/requests/productLists/mutations/AddProductToComparisonMutation.graphql
sebaholesz: As discussed, I agree with you. After going through it once again, I moved the layer of generic code up, now making basically everything generic. You can check [the commit, in which I do it](https://github.com/shopsys/shopsys/pull/2901/commits/40f0a750f24c4d5c83d4bfc417cf203a2727053d) and the modified upgrade notes.

project-base/storefront/hooks/productLists/comparison/useComparison.tsx
sebaholesz: As discussed, I agree with you. After going through it once again, I moved the layer of generic code up, now making basically everything generic. You can check [the commit, in which I do it](https://github.com/shopsys/shopsys/pull/2901/commits/40f0a750f24c4d5c83d4bfc417cf203a2727053d) and the modified upgrade notes.

project-base/storefront/hooks/productLists/comparison/useComparison.tsx
sebaholesz: As discussed, I agree with you. After going through it once again, I moved the layer of generic code up, now making basically everything generic. You can check [the commit, in which I do it](https://github.com/shopsys/shopsys/pull/2901/commits/40f0a750f24c4d5c83d4bfc417cf203a2727053d) and the modified upgrade notes.

project-base/storefront/hooks/productLists/comparison/useComparison.tsx
sebaholesz: I kept this as the only thing specific to the implementation. I feel like that is the correct approach.

project-base/storefront/hooks/productLists/useProductListUuids.ts
sebaholesz: This was completely removed 👍

project-base/storefront/urql/cache/updates.ts
sebaholesz: This was completely removed 👍

packages/framework/src/Migrations/Version20231020101214.php
```diff
@@ -0,0 +1,63 @@
...
+class Version20231020101214 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->sql('
+            CREATE TABLE product_list_items (
+                uuid UUID NOT NULL,
```
TomasLudvik: Currently, we are using integer IDs as primary keys everywhere. We also had some internal discussion, where I heard that not everyone is fine with using UUIDs as primary keys. In such case I would suggest to have integer as primary key and UUID as unique key.

packages/framework/src/Model/Product/List/ProductList.php
```diff
@@ -0,0 +1,87 @@
...
+ * @ORM\Entity
+ */
+class ProductList
+{
+    /**
+     * @var string
+     * @ORM\Column(type="guid", unique=true)
+     * @ORM\Id
+     */
+    protected string $uuid;
```
TomasLudvik: We do not want to use type hints in Entities and DataObjects, as those cannot be changed in project-base.

packages/framework/src/Model/Product/List/ProductListRepository.php
```diff
@@ -0,0 +1,97 @@
...
+
+        if ($uuid !== null) {
+            $queryBuilder->andWhere('pl.uuid = :uuid');
+            $parameters['uuid'] = $uuid;
+        }
+
+        return $queryBuilder
+            ->setParameters($parameters)
+            ->getQuery()
+            ->getOneOrNullResult();
```
TomasLudvik: Any reason to not use findOneBy?

packages/frontend-api/src/Resources/config/graphql-types/QueryDecorator.types.yaml
```diff
@@ -154,3 +154,10 @@ QueryDecorator:
                 type: 'Settings'
                 resolve: "@=query('settingsQuery')"
                 description: "Returns current settings"
+            productList:
+                type: 'ProductList'
+                resolve: "@=query('productListQuery', args)"
+                args:
+                    input:
+                        type: "ProductListInput!"
+                description: "FInd product list by uuid and type or if customer is logged, try find the the oldest list of the given type for the logged customer. The logged customer can also optionally pass the UUID of his product list."
```
TomasLudvik: You have typo in first word

Also you use different formats for uuid.

You are saying oldest list - that makes me think, there should be only one list maximum per type, should it not?

packages/framework/src/Model/Product/List/ProductListRepository.php
```diff
@@ -71,6 +71,21 @@ public function findProductListByTypeAndCustomerUser(
             ->getOneOrNullResult();
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\List\ProductListTypeEnum $productListType
+     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser $customerUser
+     * @return \Shopsys\FrameworkBundle\Model\Product\List\ProductList[]
+     */
+    public function getProductListsByTypeAndCustomerUser(
```
TomasLudvik, line 70: This confuses me as those lists have nothing for a user to differentiate between them, like names.

packages/frontend-api/src/Model/Mutation/ProductList/ProductListMutation.php
```diff
@@ -0,0 +1,72 @@
...
+        $input = $argument['input'];
+        $productUuid = $input['productUuid'];
+        $productListInput = $input['productListInput'];
+        $productListType = $productListInput['type'];
+        $productListUuid = $productListInput['uuid'];
+        $customerUser = $this->currentCustomerUser->findCurrentCustomerUser();
+
+        $productList = null;
+        if ($customerUser !== null) {
+            $productList = $this->productListFacade->findProductListByTypeAndCustomerUser($productListType, $customerUser, $productListUuid);
```
TomasLudvik, line 57: As I mentioned my confusion earlier - this is another reason for it as you are unable to add to any other lists of yours than the last one.

project-base/app/src/DataFixtures/Demo/ProductListDataFixture.php
```diff
@@ -73,6 +73,7 @@ private function createProductList(
 
         foreach ($products as $product) {
             $this->productListFacade->addProductToList($productList, $product);
+            sleep(2);
```
TomasLudvik: This does not feel ok. This will take at least eight seconds to install these fixtures, that is a pretty lot.

packages/frontend-api/src/Resources/config/graphql-types/MutationDecorator.types.yaml
```diff
@@ -75,3 +75,10 @@ MutationDecorator:
                     input:
                         type: ProductListUpdateInput!
                 resolve: "@=mutation('removeProductFromListMutation', args)"
+            CleanProductList:
+                type: ProductList
+                description: "Removes the product list"
```
TomasLudvik: I feel like this method should be called RemoveProductList and also connected mutation

packages/framework/src/Model/Product/List/ProductListRepository.php
```diff
@@ -102,6 +103,19 @@ public function findAnonymousProductListByTypeAndUuid(
         ]);
     }
 
+    public function removeOldProductLists(): void
+    {
+        $removeDate = new DateTimeImmutable('-31day');
```
TomasLudvik: I would add info to the method name, that it is removing old product lists for unlogged users. 

Also I would move that date to the cron.

TomasLudvik (changes requested):

project-base/storefront/components/Pages/ProductComparison/ProductComparisonContent.tsx
```diff
@@ -88,8 +88,8 @@ export const ProductComparisonContent: FC<ProductComparisonContentProps> = ({ pr
                     >
                         <ProductComparisonHead productsCompare={productsCompare} />
                         <ProductComparisonBody
+                            comparedProducts={productsCompare}
```
tvikito: Maybe we can stay unified so the prop can be named the same `comparedProducts`?

project-base/storefront/hooks/productLists/useUpdateProductListUuid.ts
```diff
@@ -0,0 +1,23 @@
+import { ProductListTypeEnumApi } from 'graphql/generated';
+import { usePersistStore } from 'store/usePersistStore';
+
+export const useUpdateProductListUuid = (productListType: ProductListTypeEnumApi) => {
```
tvikito, line 4: I am bit confused now. I've seen that you started use the patter for those helper files which are placed next to module/feature. Is this following pattern which you've used already?

project-base/storefront/hooks/productLists/useProductList.ts
```diff
@@ -1,39 +1,54 @@
...
+    const isUserLoggedIn = useIsUserLoggedIn();
+
+    const [, addProductToListMutation] = useAddProductToListMutationApi();
+    const [, removeProductFromListMutation] = useRemoveProductFromListMutationApi();
+    const [, cleanListMutation] = useCleanProductListMutationApi();
+
+    const [{ data: productListData, fetching }] = useProductListQueryApi({
+        variables: {
+            input: {
+                type: ProductListTypeEnumApi.WishlistApi,
```
tvikito: We probably don't want this here. Does those features even work with this? 🧐

project-base/storefront/components/Pages/ProductComparison/ProductComparisonContent.tsx
sebaholesz: Yeah, I didnt want to fix it everywhere, but I did it. It is renamed now.

project-base/storefront/hooks/productLists/useUpdateProductListUuid.ts
sebaholesz: If your question is why is this file not near the "product list" feature, I dont where that would be. We would probably have to add ProductList folder to components/Pages, which I am not sure if it sounds so good.

If you are asking if this is the first time I added a helpers file next to a feature, then it is not. The first time I did it was more than 4 months ago. In some cases it makes sense, as we discussed in some FE meeting quite some time ago.

project-base/storefront/hooks/productLists/useProductList.ts
sebaholesz: We definitely dont. It would all behave as a wishlist. And I mean... this is why we are doing CR, right?

tvikito: Yes, definitely! :) I was just wondering how it behaved.

project-base/storefront/hooks/productLists/useUpdateProductListUuid.ts
tvikito: Ah, ok. I didn't realize the fact about multiple places. Let's leave it here.

packages/framework/src/Migrations/Version20231020101214.php
vitek-rostislav: fixed in 3ee5dad3f and 6f44300c1

packages/framework/src/Model/Product/List/ProductList.php
vitek-rostislav: ah, true, I forgot about that, thanks, fixed in 6f44300c17d992adc489b94df17d89898d1d6e70

packages/framework/src/Model/Product/List/ProductListRepository.php
vitek-rostislav: good catch, rewritten in 92ac5bea3

packages/frontend-api/src/Resources/config/graphql-types/QueryDecorator.types.yaml
vitek-rostislav: fixed in c66b6ea6f.

Regarding the number of lists per type - the user story requires the possibility of adding multiple lists of the same type. In our implementation, there is only one list of each of the types, however, there must be no limitation on that, so basically, I can not count on the fact the database will return a unique entry when I ask for customer user's product list of a given type.

packages/framework/src/Model/Product/List/ProductListRepository.php
vitek-rostislav: yes, they can have names - it is prepared for such a modification if a particular project needs that.

packages/frontend-api/src/Model/Mutation/ProductList/ProductListMutation.php
vitek-rostislav: You can add a product to any of your lists if you provide the list UUID :wink:

project-base/app/src/DataFixtures/Demo/ProductListDataFixture.php
vitek-rostislav: Ok, fair point, I changed it to 1 second in 3e493ec3d. I am not comfortable with adding a logic for setting the creation date for the product list item that would be there just for testing purposes. If you feel differently, let's discuss it, please.

packages/frontend-api/src/Resources/config/graphql-types/MutationDecorator.types.yaml
vitek-rostislav: yeah, I do agree (I kept the naming from the original wishlist/comparison implementation). As changing this in history would complicate the rebase, I changed it in a new commit - 67727a52d

packages/framework/src/Model/Product/List/ProductListRepository.php
vitek-rostislav: makes sense, fixed in 54a6e779c

project-base/app/src/DataFixtures/Demo/ProductListDataFixture.php
TomasLudvik: Can you update it to use order by createdAt and ID

TomasLudvik (approved):

pk16011990 (approved):

PR #2902, opened 2023-10-24
fix set default delivery address country

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Set default delivery address country after user access Contact Information page
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1793-fix-set-default-delivery-address-country.odin.shopsys.cloud
  - https://cz.tv-fw-1793-fix-set-default-delivery-address-country.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
+++ b/project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
@@ -84,7 +84,11 @@ export const ContactInformationDeliveryAddress: FC = () => {
                     setValue(formMeta.fields.deliveryPostcode.name, deliveryAddress.postcode, { shouldValidate: true });
                 }
             }
-        } else if (deliveryAddressUuid === '') {
+        } else {
+            const selectedCountryOption = countriesAsSelectOptions.find(
+                (option) => option.value === user?.country.code,
+            );
+
             setValue(formMeta.fields.deliveryFirstName.name, '');
             setValue(formMeta.fields.deliveryLastName.name, '');
             setValue(formMeta.fields.deliveryCompanyName.name, '');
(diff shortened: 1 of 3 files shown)


project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
```diff
@@ -84,7 +84,7 @@ export const ContactInformationDeliveryAddress: FC = () => {
                     setValue(formMeta.fields.deliveryPostcode.name, deliveryAddress.postcode, { shouldValidate: true });
                 }
             }
-        } else if (deliveryAddressUuid === '') {
+        } else {
```
sebaholesz, line 87: Looking at this condition and what it did the condition on line 66, I think we can change that one from 
`if (user && deliveryAddressUuid) {` to `if (deliveryAddressUuid) {`, because only signed-in users can have saved delivery addresses.

sebaholesz (changes requested):
Missing upgrade notes

project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationDeliveryAddress.tsx
tvikito: I am not sure if this is good improvement. Just for better type check on line 67 I would leave it. But also since this logic is pretty complex I don't think that connecting those two values here is good idea. I would rather keep them both in the check.

sebaholesz (approved):

pk16011990 (changes requested):
When I'm logged as no-reply@shopsys.com, billing country is Slovakia, I have still pre-filled country Czechia in last order step with billing information.

pk16011990 (approved):

PR #2903, opened 2023-10-25
[shopsys] increase speed of Product creation

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| There were some unnecessary flushes fired during the creation of the Product. These flushes were moved after all data of each product-related entity were created (e.g. domain data). This significantly improves the speed of the Product creation. For big batches, like performance data creation of 40000 products this is up to 24 times faster.
|New feature| No 
|Fixes issues| ... 

----
  - https://tl-improve-performance-data.odin.shopsys.cloud
  - https://cz.tl-improve-performance-data.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
+++ b/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
@@ -49,8 +49,10 @@ class FriendlyUrlFacade
 
         foreach ($friendlyUrls as $friendlyUrl) {
             $locale = $this->domain->getDomainConfigById($friendlyUrl->getDomainId())->getLocale();
-            $this->resolveUniquenessOfFriendlyUrlAndFlush($friendlyUrl, $namesByLocale[$locale]);
+            $this->resolveUniquenessOfFriendlyUrl($friendlyUrl, $namesByLocale[$locale]);
         }
+
+        $this->em->flush();
     }
 
     /**
@@ -64,15 +66,17 @@ class FriendlyUrlFacade
         $friendlyUrl = $this->friendlyUrlFactory->createIfValid($routeName, $entityId, (string)$entityName, $domainId);
 
         if ($friendlyUrl !== null) {
-            $this->resolveUniquenessOfFriendlyUrlAndFlush($friendlyUrl, $entityName);
+            $this->resolveUniquenessOfFriendlyUrl($friendlyUrl, $entityName);
         }
+
+        $this->em->flush();
     }
 
     /**
      * @param \Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrl $friendlyUrl
      * @param string $entityName
      */
-    protected function resolveUniquenessOfFriendlyUrlAndFlush(FriendlyUrl $friendlyUrl, $entityName)
+    protected function resolveUniquenessOfFriendlyUrl(FriendlyUrl $friendlyUrl, $entityName)
     {
         $attempt = 0;
 
@@ -108,7 +112,6 @@ class FriendlyUrlFacade
         }
 
         $this->em->persist($friendlyUrl);
-        $this->em->flush();
         $this->setFriendlyUrlAsMain($friendlyUrl);
     }
 
@@ -214,8 +217,6 @@ class FriendlyUrlFacade
         }
         $mainFriendlyUrl->setMain(true);
         $this->renewMainFriendlyUrlSlugCache($mainFriendlyUrl);
-
-        $this->em->flush();
     }
 
     /**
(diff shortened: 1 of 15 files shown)


packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
```diff
@@ -108,7 +112,6 @@ protected function resolveUniquenessOfFriendlyUrlAndFlush(FriendlyUrl $friendlyU
         }
 
         $this->em->persist($friendlyUrl);
-        $this->em->flush();
```
vitek-rostislav, line 111: the method should not be called `...AndFlush` anymore

packages/framework/src/Model/Product/Pricing/ProductManualInputPriceFacade.php
```diff
@@ -41,6 +41,5 @@ public function refresh(Product $product, PricingGroup $pricingGroup, ?Money $in
             $manualInputPrice->setInputPrice($inputPrice);
         }
         $this->em->persist($manualInputPrice);
-        $this->em->flush();
```
vitek-rostislav, line 44: I feel a little bit concerned about removing `flush` from the public facade methods. Basically, it means you need to keep it in mind when using the method and call it "manually" wherever you use it (which can be already seen - you had to add the `flush` call into `CartWatcherTest::testGetModifiedPriceItemsAndUpdatePrices()`

vitek-rostislav: I can live with that in the protected methods (e.g. you removed it from `GoogleProductDomainFacade::saveGoogleProductDomain` and I would be ok with that if the visibility was changed to protected -  it is not used outside the class)

vitek-rostislav: ...the same applies to the other `*ProductDomainFacade::save*` methods (`HeurekaProductDataFixture` and `ZboziPluginDataFixture` need to be adjusted in such a case)

project-base/app/src/DataFixtures/Performance/ProductDataFixture.php
```diff
@@ -115,9 +113,7 @@ public function load(OutputInterface $output)
                 $this->productsByCatnum[$product->getCatnum()] = $product;
             }
 
-            if ($this->countImported % self::BATCH_SIZE === 0) {
-                $this->cleanAndLoadReferences();
-            }
+            $this->cleanAndLoadReferences();
```
vitek-rostislav, line 115: is not that an overkill to do so after each product? I would expect it would be better just to decrease the `BATCH_SIZE` value, or? Have you tried changing it to `10`, `100` maybe?

project-base/app/src/DataFixtures/Performance/ProductDataFixture.php
```diff
@@ -196,6 +192,8 @@ private function makeProductDataUnique(ProductData $productData)
      */
     private function getUniqueIndex()
     {
+        $this->demoDataIterationCounter++;
```
vitek-rostislav, line 194: could you please extract this into a separate commit with a descriptive commit message? I do not understand the change (the property is already increased in the `load` method, I can imagine this change is because of the variants creation but I am not sure at the moment :thinking:

vitek-rostislav (changes requested):

project-base/app/src/DataFixtures/Performance/ProductDataFixture.php
TomasLudvik: I have removed the increment in the load function. The problem was that increment did not work and I was getting errors about creating the same catnum products. I did not look into it anymore as it was not important to me and this fixed the problem. I do not see the need to create a separate commit for that.

project-base/app/src/DataFixtures/Performance/ProductDataFixture.php
TomasLudvik: This leads to much faster generation than cleaning after any batch. I have tried several scenarios and this wast the fastest one.

packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
TomasLudvik: changed

packages/framework/src/Model/Product/Pricing/ProductManualInputPriceFacade.php
TomasLudvik: I have changed methods to protected ones

vitek-rostislav (approved):
Thank you for the fixes!

project-base/app/src/DataFixtures/Performance/ProductDataFixture.php
vitek-rostislav: ok, thanks for the info

pk16011990 (approved):

PR #2905, opened 2023-10-26
fix Comparison for not logged in users

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Besides fix Comparison for not logged in users it was done added unified code for Comparison and Wishlist and refactored Zustand store to use only one store (User Store) for all cartUuid, wishlistUuid and comparisonUuid
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1905-fix-comparison.odin.shopsys.cloud
  - https://cz.tv-fw-1905-fix-comparison.odin.shopsys.cloud


project-base/storefront/hooks/comparison/useComparison.tsx
```diff
@@ -18,28 +18,23 @@ export const useComparison = () => {
...
 
     useEffect(() => {
-        if (initialLoadRef.current) {
-            initialLoadRef.current = true;
-        } else {
-            setIsLoadingVisible(fetching);
-        }
-    }, [fetching]);
+        setIsFetchingPaused(!comparisonUuid && !isUserLoggedIn);
+    }, [comparisonUuid]);
```
sebaholesz, line 32: Because there was a question regarding this on Slack, I am just asking to make sure I understand your motivation. This is done so that the fetching paused is updated AFTER the first render through the useEffect and the initial value is true, right? Can you elaborate why that is necessary?

tvikito: Yup, motivation is exactly this. Since comparisonUuid is in persist store in browser this information is not accessible to Next server. So it would cause Next complaining about different rendered UI.

pk16011990 (approved):

PR #2906, opened 2023-10-26
add additional skeletons

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| added skeletons for Wishlist, Comparison and My orders and Order detail page
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-fw-1896-add-additional-skeletons.odin.shopsys.cloud
  - https://cz.tv-fw-1896-add-additional-skeletons.odin.shopsys.cloud


project-base/storefront/components/Pages/ProductComparison/ProductComparison.tsx
```diff
@@ -1,40 +1,40 @@
...
-
     useGtmSliderProductListViewEvent(comparison?.products, GtmProductListNameType.product_comparison_page);
 
     return (
         <Webline>
-            {fetching ? (
-                <div className="flex w-full justify-center py-10">
-                    <Loader className="w-10" />
+            <Heading className="flex w-full items-end lg:mb-8 lg:w-auto lg:flex-1" type="h1">
+                {`${t('Product comparison')}${comparison?.products.length ? ` (${comparison.products.length})` : ''}`}
```
sebaholesz: This would benefit from being in a variable...

project-base/storefront/components/Pages/ProductComparison/ProductComparison.tsx
```diff
@@ -1,40 +1,40 @@
...
+            </Heading>
+
+            {fetching && <SkeletonModuleComparison />}
+
+            {!!comparison?.products && !fetching && <ProductComparisonContent productsCompare={comparison.products} />}
```
sebaholesz: Cannot these two be joined into a ternary operator? Why does one condition have `!!comparison?.products` and the second does not?

project-base/storefront/components/Pages/Wishlist/Wishlist.tsx
```diff
@@ -31,59 +31,65 @@ export const Wishlist: FC = () => {
...
-    if (fetching) {
-        return <LoaderWithOverlay />;
-    }
-
     return (
         <>
-            <Heading className="!text-big lg:!text-h1" type="h1">
-                {t('Wishlist')}
+            <Heading type="h1">
+                {`${t('Wishlist')}${wishlist?.products.length ? ` (${wishlist.products.length})` : ''}`}
```
sebaholesz: Same with the title here. Not sure how much it would help but I feel like it would make it more clear.

project-base/storefront/components/Pages/Wishlist/Wishlist.tsx
```diff
@@ -31,59 +31,65 @@ export const Wishlist: FC = () => {
...
-                            onChange={() => {
-                                return false;
-                            }}
-                        />
-                        <SubmitButton
-                            className="mt-2 lg:ml-2 lg:mt-0"
+
+            {fetching && <SkeletonModuleWishlist />}
+
+            {wishlist?.products.length && !fetching && (
```
sebaholesz: This is a different condition than in comparison. Here you check the length, there you only check the products array. Is this on purpose?

project-base/storefront/components/Pages/Wishlist/Wishlist.tsx
```diff
@@ -31,59 +31,65 @@ export const Wishlist: FC = () => {
...
+                        </div>
+
+                        <div className="flex w-full flex-col items-center lg:w-1/2 lg:flex-row">
+                            <TextInput
+                                id="copyUrl-input"
+                                label={t('Send a link to this wishlist')}
+                                value={buildShareUrl()}
+                                onChange={() => {
+                                    return false;
+                                }}
```
sebaholesz: Is there a reason for not writing it in shorthand?

sebaholesz (changes requested):

project-base/storefront/components/Pages/ProductComparison/ProductComparison.tsx
tvikito: Yeah, I guess that would make JSX better readable. Refactored.

project-base/storefront/components/Pages/ProductComparison/ProductComparison.tsx
tvikito: You mean why does second condition have `!!comparison?.products.length`, right? I unified those conditions. Also for Wishlist.

project-base/storefront/components/Pages/Wishlist/Wishlist.tsx
tvikito: Sure. Refactored.

project-base/storefront/components/Pages/Wishlist/Wishlist.tsx
tvikito: Refactored, according to previous comment.

project-base/storefront/components/Pages/Wishlist/Wishlist.tsx
tvikito: I have no idea, this was not included by my changes. But looks like it can be removed and doesn't break anything. So it's removed.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -26,7 +26,15 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
...
-    const isStatic = type === 'static' || type === 'homepage' || type === 'stores';
+    const isStatic =
+        type === 'static' ||
+        type === 'homepage' ||
+        type === 'stores' ||
+        type === 'wishlist' ||
+        type === 'comparison' ||
+        type === 'orders' ||
+        type === 'order' ||
+        type === 'variants';
```
sebaholesz: This does not make much sense to me conceptually. Static types are those, which cannot be returned by the BE script and are not friendly URL => variants are friendly URL, they should be returned by the BE script. The script should be able to return something as "product_regular" and "product_main_variant". I understand this PR is about skeletons but you are just patching holes. First, the script should be able to differentiate between these types, then it would implicitly appear as part of the `FriendlyPagesTypesKeys` type and you would not need to handle this like this.

sebaholesz: I would not use the name `variants`. We call it `Main Variant` almost everywhere and I would stick to that.

project-base/storefront/components/Blocks/Product/ProductAction.tsx
```diff
@@ -18,29 +18,16 @@ const TEST_IDENTIFIER = 'blocks-product-action';
...
-                        )
-                    }
-                >
-                    {t('Choose variant')}
-                </Button>
+                <ExtendedNextLink href={product.slug} type="variants">
+                    <Button className="w-full" dataTestId={TEST_IDENTIFIER + '-choose-variant'} name="choose-variant">
+                        {t('Choose variant')}
+                    </Button>
+                </ExtendedNextLink>
```
sebaholesz: Now, if you click on the list item (product tile) you get a different skeleton than if you click on the product action, that's confusing. It should be the same skeleton.

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
```diff
@@ -32,6 +37,8 @@ export const SkeletonManager: FC<SkeletonManagerProps> = ({ isFetchingData, isPa
             return <SkeletonPageHome />;
         case 'product':
             return <SkeletonPageProductDetail />;
+        case 'variants':
+            return <SkeletonPageProductVariants />;
```
sebaholesz: I did not realize it in previous reviews, but the skeletons should mirror the pages, thus `ProductDetailMainVariant` and `ProductDetail`.

project-base/storefront/components/Layout/Header/MenuIconic/MenuIconicElements.tsx
```diff
@@ -10,14 +10,14 @@ export const MenuIconicItem: FC<{ title?: string }> = ({ children, className, da
...
-export const MenuIconicSubItemLink: FC<MenuIconicItemLinkProps> = ({ children, href, onClick, dataTestId }) => {
+export const MenuIconicSubItemLink: FC<MenuIconicItemLinkProps> = ({ children, href, onClick, type, dataTestId }) => {
     if (href) {
         return (
             <ExtendedNextLink
                 className="block py-3 px-5 text-sm text-dark no-underline"
                 data-testid={dataTestId}
                 href={href}
-                type="static"
+                type={type || 'static'}
```
sebaholesz: I think giving a default value to the prop makes more sense. Just a suggestion.

project-base/storefront/components/Pages/App/AppPageContent.tsx
```diff
@@ -33,6 +34,10 @@ export const AppPageContent: FC<AppPageContentProps> = ({ Component, pageProps }
     const { url } = pageProps.domainConfig;
     const userConsent = usePersistStore((store) => store.userConsent);
 
+    useEffect(() => {
+        usePersistStore.persist.rehydrate();
+    }, []);
```
sebaholesz: In the manner of continuing with abstracting the functionality into separate hooks, could you also extract this one? I know it's pretty small, but will make it more clear. If we keep this trend, we will be more bound to do that with future code as well, whereas if we keep adding small useEffects and code bits, we will end up with a super crowded component as `_app.tsx` previously

project-base/storefront/hooks/comparison/useComparison.tsx
```diff
@@ -65,15 +54,13 @@ export const useComparison = () => {
 
         if (removeProductFromComparisonResult.error) {
             const { applicationError } = getUserFriendlyErrors(removeProductFromComparisonResult.error, t);
-            if (applicationError?.message) {
-                showErrorMessage(applicationError.message);
-            } else {
-                showErrorMessage(t('Unable to remove product from comparison.'));
-            }
```
sebaholesz, line 72: I am not sure what is the reason for removing this? cannot `applicationError.message` contain something interesting?

project-base/storefront/store/broadcast.ts
```diff
@@ -0,0 +1,58 @@
...
+        const state = Object.entries(updated).reduce((obj, [key, val]) => {
+            let newObj = { ...obj };
+            if (previous[key] !== val) {
+                newObj = { ...newObj, [key]: val };
+            }
+            return newObj;
+        }, {} as Item);
+
+        channel.postMessage(state);
+    };
```
sebaholesz, line 34: Are you sure this code (the entire module) can handle more complex structures as sets and maps which we use in our store? Dont we need some more sophisticated (de)serialization?

project-base/storefront/store/broadcast.ts
```diff
@@ -0,0 +1,58 @@
...
+    Mps extends [StoreMutatorIdentifier, unknown][] = [],
+    Mcs extends [StoreMutatorIdentifier, unknown][] = [],
+>(
+    f: StateCreator<T, Mps, Mcs>,
+    name: string,
+) => StateCreator<T, Mps, Mcs>;
+
+type BroadcastImpl = <T>(f: StateCreator<T, [], []>, name: string) => StateCreator<T, [], []>;
+
+const broadcastImpl: BroadcastImpl = (f, name) => (set, get, store) => {
```
sebaholesz, line 14: I do not see the piece of code which synchronizes a new tab with what is present in other tabs. Is that not necessary?

sebaholesz, line 14: Are we cleaning/closing the BroadcastChannel when needed?

project-base/storefront/store/broadcast.ts
sebaholesz, line 34: Can it happen that the state actually does not change? Can we not send a message in such case?

project-base/storefront/hooks/cart/useRemoveFromCart.ts
```diff
@@ -1,26 +1,36 @@
...
     const [{ fetching }, removeItemFromCart] = useRemoveFromCartMutationApi();
     const { url, currencyCode } = useDomainConfig();
     const cartUuid = usePersistStore((store) => store.cartUuid);
+    const isUserLoggedIn = useIsUserLoggedIn();
+
+    const [, refetchCart] = useCartQueryApi({
+        variables: { cartUuid },
+        pause: !cartUuid && !isUserLoggedIn,
+        requestPolicy: 'network-only',
+    });
```
sebaholesz: Now you call CartQuery twice on every page which has this hook, I do not think that is desirable.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
```diff
@@ -39,8 +49,6 @@ export const useRemoveFromCart = (gtmProductListName: GtmProductListNameType): [
                 url,
             );
         }
-
-        return removeItemFromCartActionResult.data?.RemoveFromCart ?? null;
```
sebaholesz: I do not like this change. These hooks (all in the folder `cart`) are written to return cart. I would like to keep that. It makes sense to have it that way since using the function's return is the best thing when you want to react synchronously.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
```diff
@@ -1,26 +1,36 @@
...
     const updateCartUuid = usePersistStore((store) => store.updateCartUuid);
 
     const removeItemFromCartAction = async (cartItem: CartItemFragmentApi, listIndex: number) => {
         const removeItemFromCartActionResult = await removeItemFromCart({
             input: { cartUuid, cartItemUuid: cartItem.uuid },
         });
 
+        if (removeItemFromCartActionResult.error) {
+            refetchCart();
+        }
```
sebaholesz: I am not saying I disagree, just what exactly was the motivation to add this? I could make no sense of the reason from the commit message.

project-base/storefront/hooks/auth/useAuth.tsx
```diff
@@ -19,14 +19,7 @@ export type LoginHandler = (
...
 
-export type LogoutHandler = () => Promise<
-    OperationResult<
-        LogoutApi,
-        Exact<{
-            [key: string]: never;
-        }>
-    >
->;
+type LogoutHandler = () => Promise<void>;
```
sebaholesz, line 23: You do not need this type at all, then.

project-base/storefront/hooks/app/useAuthLoader.ts
```diff
@@ -1,37 +1,43 @@
...
+                authLoading === 'login-loading-with-cart-modifications'
+            ) {
+                showInfoMessage(t('Your cart has been modified. Please check the changes.'));
+            }
 
-        updateAuthLoadingState(null);
-    }, []);
+            updateAuthLoadingState(null);
+        }
+    }, [authLoading]);
```
sebaholesz: I am really sad to see this 😢 The "beauty" of it was in the fact that we knew it was going to run only once. Now we have this added complexity of checking the initial load and then reacting to authLoading... 

Is this because of zustand hydration? Cannot we react to authLoading and differentiate between not-initialized and not-set? We would still have the dependency on authLoading, but without this initLoading logic.

Or maybe is there a pointer given by zustand to check if it initialized/hydrated?

project-base/storefront/components/Layout/Header/DropdownMenu/SubMenu.tsx
```diff
@@ -46,10 +48,24 @@ export const SubMenu: FC = () => {
...
+    const { onMenuToggleHandler } = useContext(DropdownMenuContext);
+
     if (href) {
         return (
-            <ExtendedNextLink passHref className="mb-5 px-8 text-sm text-dark no-underline" href={href} type="static">
+            <ExtendedNextLink
+                passHref
+                className="mb-5 px-8 text-sm text-dark no-underline"
+                href={href}
+                type={type || 'static'}
```
sebaholesz: I think you could put static as a default value for type.

UPGRADE-14.0.md
```diff
@@ -190,4 +190,8 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 -   fix Comparison for not logged in users ([#2905](https://github.com/shopsys/shopsys/pull/2905))
     -   unified code for Comparison and Wishlist
     -   refactored Zustand store to use only one store (User Store) for all cartUuid, wishlistUuid and comparisonUuid
+-   added skeletons for Wishlist, Comparison and My orders, Order detail and Product Variants page ([#2906](https://github.com/shopsys/shopsys/pull/2906))
+    -   see #project-base-diff to update your project
+-   fix Add to cart/wishlist/comparison feature when using multiple tabs ([#2906](https://github.com/shopsys/shopsys/pull/2906))
```
sebaholesz: I am not sure about this. Should there be two entries for one PR? I understand those are multiple US, but one PR 🤔

UPGRADE-14.0.md
```diff
@@ -190,4 +190,8 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
...
+    -   As fresh unlogged user without `cartUuid` I open multiple tabs, then in each tab I add product to cart, I am experiencing that each tab contains different products in cart, even after refresh. Same applies to `Wishlist` and `Comparison`. This was solved by adding `BroadcastChannel` middleware to Zustand store, so now it uses always only first returned `uuid`.
```
sebaholesz: I would use a different formulation. Something like "There was an error". Now when a person reads this, he reads a problem description, without knowing this is a past, solved error.

sebaholesz: I think there are several more things we should add here:

- there is no redirect to homepage after logout
- useRemoveProductFromCart changed its API, this is a breaking change and can break people's apps (here, I would still prefer to not change the API as described below)
- the new broadcast channel can possibly have problems with some more complex data structures. This can potentially break people's apps and they should know about it
- I would mention upgrade of a package (zustand)
- we now rehydrate zustand on the client, so we should tell them that they do not have to check for the initial state in their apps (you removed it for wishlist and comparison, so you can mention it as an example to follow)
- and the last point, which isnt super strong, but still worth consideration, I would mention that there are new skeleton types. It is possible that they have already implemented their own or dont have some of the pages. In such case, they should make an informed decision about it

sebaholesz (changes requested):

project-base/storefront/components/Layout/Header/MenuIconic/MenuIconicElements.tsx
tvikito: I personally prefer this way. There is no like bad way here but I rather see on one place what can be send to the prop. Also it's implemented on other places this way, so I would rather leave it like this if you don't have some pragmatic reason.

project-base/storefront/components/Layout/Header/DropdownMenu/SubMenu.tsx
tvikito: I personally prefer this way. There is no like bad way here but I rather see on one place what can be send to the prop. Also it's implemented on other places this way, so I would rather leave it like this if you don't have some pragmatic reason.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
tvikito: Yeah, naming could be better matching. Renamed to `productMainVariant`.

project-base/storefront/components/Blocks/Product/ProductAction.tsx
tvikito: Ah, nice catch. I missed this. Refactored.

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
tvikito: Yes, it won't hurt to have better consistency. Refactored.

project-base/storefront/components/Pages/App/AppPageContent.tsx
tvikito: I am not sure if I like this, since it's one more function call for 3 lines of code. If that would be bigger chunk of code I would separate the code. But ok, if you prefer it this way, I don't have problem with this approach. Refactored.

project-base/storefront/hooks/comparison/useComparison.tsx
tvikito: You have it refactored on the line bellow :)

project-base/storefront/store/broadcast.ts
tvikito: I haven't found any issues. If you have some doubts, please let me know what you wanna test or test on your own. I don't see any reason to add more complexity for now.

project-base/storefront/store/broadcast.ts
tvikito: That's purpose of this fix. And that's what BroadcastChannel functionality does.

tvikito: When we would need it?

project-base/storefront/store/broadcast.ts
tvikito: What would we need this for? I cannot find any reason to implement this.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
tvikito: It is not calling CartQuery twice, because it's using cached data for second call. We have two calls of useCartQueryApi which is nothing what we wouldn't already used in our app. But I forgot I've already exported `refetchCart` from `useCurrentCart`. So I am using this exported function and this hook is removed.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
tvikito: We don't need this anywhere. It is more confusing then beneficial. It was not consistent with a rest of cart helpers. After we find place we can easily implement.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
tvikito: You can test it on your own. This implementation I would consider it as a bugfix. When you add to cart some product, then open another tab, remove the product from cart, go to other tab and try to remove it there as well. It won't do anything. Because of that it make sense to refetch cart so user get the fresh results.

project-base/storefront/hooks/auth/useAuth.tsx
tvikito: What do you mean? It's used in logout function. 

image

If you meant it like that it not some complex type so we don't need to type it at all, the same idea you can implement to every hook return types, so we can remove, since we will find out later on when return type is used incorrectly.

project-base/storefront/hooks/app/useAuthLoader.ts
tvikito: In this case that's great that you were sad about this. Actually I have found the way how to check if store is hydrated. I like it more as well. So it's refactored.

UPGRADE-14.0.md
tvikito: After my yesterday understanding it's not important how many PR/commits we do have, but having it documented here is. So I think about this that it is better readable than having all those changes under one group. I will ask in the channel to be sure.

UPGRADE-14.0.md
tvikito: I don't see this so understandable but I used better formulation.

tvikito: Yup, thanks for help with summary. Just few notes.

I won't include warning about BroadcastChannel, as I've answered with previous comment - I don't see any reason why it would break. And about additional skeletons there is another note in the root tree.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
tvikito: I don't see any point to change the BE script. But I agree with you that this component got a little bit out of clean implementation. That's why I refactored it. Please see new commit.

UPGRADE-14.0.md
tvikito: Ok, it was agreed that after we finish CR I will split those changes to separate PRs.

project-base/storefront/components/Layout/Header/MenuIconic/MenuIconicElements.tsx
sebaholesz: As I said, just a suggestion. No problem with keeping it this way 🙂

project-base/storefront/components/Layout/Header/DropdownMenu/SubMenu.tsx
sebaholesz: As I said, just a suggestion. No problem with keeping it this way 🙂

project-base/storefront/hooks/cart/useRemoveFromCart.ts
sebaholesz: It is calling cart twice, still. See the video


https://github.com/shopsys/shopsys/assets/55181354/729ae134-b560-4d12-aeff-0597af3cbbe6

UPGRADE-14.0.md
```diff
@@ -195,8 +195,25 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
...
-    -   see #project-base-diff to update your project
+
+    -   Before this change if I open multiple tabs as fresh unlogged user without `cartUuid`, then in each tab I add product to cart, I am experiencing that each tab contains different products in cart, even after refresh. Same applies to `Wishlist` and `Comparison`. This is solved now by adding `BroadcastChannel` middleware to Zustand store, so now it uses always only first returned `uuid`.
+    -   removed redirect to homepage after logout, this was not good approach
+    -   `removeItemFromCartAction ` in `useRemoveProductFromCart` is now `void` type (doesn't return `CartFragmentApi`) and in case of error is returned from API it refect cart so user gets latest cart updates
+    -   Zustand package is upgraded to latest version
+    -   now we hydrate Zustand store manually during first load, which means that now we don't have do overcomes for matching server rendered UI with client like
```
sebaholesz: Phrase _which means that now we don't have do overcomes for matching server rendered UI with client_ seems incorrect. Try changing it to something more appropriate, such as:

_We are now manually hydrating the Zustand store during the initial page load. This means that we no longer need to resolve mismatches between the UI rendered by the server and the UI managed on the client side_.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
sebaholesz, line 1: This change (everything about it) is great, I like it. What I am thinking, though, is that you should add upgrade notes for it. Things like:

- remove static type from your links
- if you have custom types, we moved it from STATIC_PAGES to CUSTOM_PAGES

sebaholesz (changes requested):

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
sebaholesz: You are right, I am dumb. Those are actually all rendered on the same page. I'm resolving this.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
sebaholesz: Okay. Should we apply this logic to other cart mutations as well? Just a thought, you can resolve this thread as you wish...

project-base/storefront/store/broadcast.ts
sebaholesz: Just performance optimization. I guess it would add more unneeded complexity. Just a thought...

project-base/storefront/store/broadcast.ts
sebaholesz: I was just thinking about a memory leak. Seems that cannot happen.

sebaholesz: I meant stuff like this (video). What I meant is that it seems we are only syncing state, but not the behaviour, which will cause issues.

https://github.com/shopsys/shopsys/assets/55181354/4087ae2a-fb3f-4092-a5fd-64abcb7c5ea6

project-base/storefront/hooks/comparison/useComparison.tsx
sebaholesz: Oh, sheesh. Right. Sorry.

project-base/storefront/store/broadcast.ts
sebaholesz: I thought about SEO categories, but they are in session storage. As for the most complex thing in persist store (packetery) it works okay. I guess there are no issues with this.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
sebaholesz: I am not sure what you mean by it was not consistent. These are the previous types (simplified):

```ts
type RemovePromoCodeHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type RemoveFromCartHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type ChangeTransportHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type ChangePaymentHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type ApplyPromoCodeHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type AddToCartAction = (...params) => Promise<AddToCartMutationApi['AddToCart'] | null>;
```

And these are the current:

```ts
type RemovePromoCodeHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type RemoveFromCartHandler = (...params) => Promise<void>;
type ChangeTransportHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type ChangePaymentHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type ApplyPromoCodeHandler = (...params) => Promise<CartFragmentApi | undefined | null>;
type AddToCartAction = (...params) => Promise<AddToCartMutationApi['AddToCart'] | null>;
```

What does it help us to remove it. Why cannot we have a unified API for the cart hooks?

project-base/storefront/store/broadcast.ts
sebaholesz, line 1: Seeing the broadcast now for what it is (a message bus), I think we should approach it from a different way, rather than just a state sync mechanism. It should actually do things like:

- trigger events across tabs (loadings, reloads, refetches)
- make things in sync (right now we have some loading mismatches as shown)

I understand that this might be something for the future. We just need to be straight with ourselves and say that RN it does not fully work. I will leave it up to you, if you want to fix it here, or later. If the latter is what you choose, let me know and I will create a US for it.

sebaholesz (changes requested):

project-base/storefront/hooks/cart/useRemoveFromCart.ts
tvikito: I haven't found another case where this would be needed. In case of Add to cart it will just add one more to quantity and return correct result. In case of changing quantity, it's the same result. If you have some improvements on your mind, we can create separate task. This is just something which was sort of connected to this synchronization.

project-base/storefront/store/broadcast.ts
tvikito: It's not so simple as we talked on call. There are more problems to answers. Let's discuss this elsewhere.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
tvikito: Ah, yes I forgot about adding notes for this refactoring. I still don't understand how this is helpful 🙈 But ok. It's filled in.

UPGRADE-14.0.md
tvikito: Done.

project-base/storefront/store/broadcast.ts
tvikito: Merging carts (topic of this task) is correct, just UI needs to be reloaded. This needs to be solved in different task.

project-base/storefront/hooks/cart/useRemoveFromCart.ts
tvikito: Solved now. Thanks for your help!

project-base/storefront/hooks/cart/useRemoveFromCart.ts
tvikito: Hmm, ok. I've probably taken wrong closures. It's reverted.

sebaholesz (approved):

UPGRADE-14.0.md
```diff
@@ -244,3 +244,5 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 -   ExtendedNextLink was refactored ([#2906](https://github.com/shopsys/shopsys/pull/2906))
     -   remove static type from your links
     -   if you have custom types, we moved it from STATIC_PAGES to CUSTOM_PAGES
+-   use BroadcastChannel to refresh cart ([#2906](https://github.com/shopsys/shopsys/pull/2906))
+    -   whereever you use modifications for cart there should be also implemented the same behavior
```
sebaholesz: whereever => wherever

project-base/storefront/hooks/cart/useReloadCart.ts
```diff
@@ -1,39 +1,21 @@
...
-
-        if ((isUserLoggedIn && !isWithUserTokens) || (!isUserLoggedIn && isWithUserTokens)) {
-            router.reload();
-        } else if (cartUuid || isUserLoggedIn) {
-            client.query(CartQueryDocumentApi, { cartUuid }, { requestPolicy: 'network-only' }).toPromise();
-        }
-    }, [slug]);
+    useBroadcastChannel('refetchCart', () => {
+        refetchCart();
+    });
```
sebaholesz: Just an idea to change 

`const { modifications, refetchCart } = useCurrentCart(false);`

to

`const { modifications, refetchCart } = useCurrentCart();`

and then call it like `refetchCart({request-policy: "network-only"});`

sebaholesz (approved):

project-base/storefront/hooks/cart/useReloadCart.ts
tvikito: It was already implemented this way with disabled cache, so I wouldn't risk to break anything. Since testing changing this is not focus of this task which is already taking very long.

UPGRADE-14.0.md
tvikito: Fixed

pk16011990 (changes requested):
I found an issue with cart.

* unregistered user
* Hello Kitty product in comparison and wishlist
* client side action to cart does not a show cart - only loading icon (no error in console and no request to API)
* refresh cart page (SSR) works properly

project-base/storefront/pages/order/contact-information.tsx
```diff
@@ -87,6 +87,12 @@ const ContactInformationPage: FC<ServerSidePropsType> = () => {
         }
     }, []);
 
+    useEffect(() => {
+        if (currentCart.isCartEmpty || !currentCart.transport || !currentCart.payment) {
+            router.replace(currentCart.isCartEmpty ? cartUrl : transportAndPaymentUrl);
+        }
+    }, [currentCart.payment, currentCart.transport, currentCart.isCartEmpty]);
```
sebaholesz: I would vote for a better split of conditions. This seems a little bit weird to first check in the if and then conditionally change arguments. What do you think?

sebaholesz (approved):
Very nice 👍 Just a small suggestion.

pk16011990 (changes requested):
The last problematic use-case. All changes in the cart are taken into account in the other tabs, except for one. Completing an order empties the cart, but the other tabs don't know about it. The problem is if I have order creation open in another tab: I create it in the first one so the second one for example in steps 2 and 3 doesn't work until page is refreshed

project-base/storefront/pages/order/contact-information.tsx
tvikito: Sure! Good idea.

pk16011990 (approved):

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -64,10 +64,10 @@ export const TransportAndPaymentContent: FC = () => {
     };
 
     useEffect(() => {
-        if (isCartEmpty) {
+        if (!cart?.items.length) {
             router.replace(cartUrl);
         }
-    }, [isCartEmpty]);
+    }, [cart]);
```
sebaholesz: If you add the length as dependency instead of the entire cart, you will reduce the number of runs, right?

project-base/storefront/pages/order/contact-information.tsx
```diff
@@ -86,6 +87,16 @@ const ContactInformationPage: FC<ServerSidePropsType> = () => {
         }
     }, []);
 
+    useEffect(() => {
+        if (!currentCart.cart?.items.length) {
+            router.replace(cartUrl);
```
sebaholesz: I would add a return after this one to not overwrite this replace with the following one 🤔

sebaholesz (changes requested):

romankuna:
After create order as logged user, when i click on Shopss logo, or some category in menu, i see skeletons and in GraphQL Network console i see call CartQuery
VIDEO: 

https://github.com/shopsys/shopsys/assets/60608926/2a3dcca9-9fc5-4be0-b1a8-5da04ec26a17


image

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
tvikito: Actually it is not good idea to check length, which can be the same even if items change. So I improved this by adding `cart.items` as dependecy.

project-base/storefront/pages/order/contact-information.tsx
tvikito: This wouldn't happen. But ok, it's improved. I don't like those returns in useEffect since it's always just hacking useEffect which should only return function called when unmount not use return for end of the function run. But I split this logic into two separate useEffects.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: Okay, correct 👍

sebaholesz (approved):

romankuna:
2. I am in cart with some product. When i log in, empty cart will be displayed for a moment

https://github.com/shopsys/shopsys/assets/60608926/62080bcb-55df-44ab-8ee7-23fd5d4f476e

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -64,7 +64,7 @@ export const TransportAndPaymentContent: FC = () => {
     };
 
     useEffect(() => {
-        if (!cart?.items.length) {
+        if (cart !== undefined && !cart?.items.length) {
             router.replace(cartUrl);
         }
     }, [cart?.items]);
```
sebaholesz: Since you check for undefined, you do not need to optionally chain it:

```ts
if (cart !== undefined && !cart.items.length) {
    router.replace(cartUrl);	        
}
```

and since you have to check for undefined, you could even explicitly check for the length, hmm?

project-base/storefront/pages/order/contact-information.tsx
```diff
@@ -88,16 +89,14 @@ const ContactInformationPage: FC<ServerSidePropsType> = () => {
...
-    useEffect(() => {
-        if (!currentCart.transport || !currentCart.payment) {
-            router.replace(transportAndPaymentUrl);
+        if (currentCart.cart !== undefined) {
+            if (!currentCart.cart?.items.length) {
+                router.replace(cartUrl);
+            } else if (!(currentCart.transport && currentCart.payment)) {
+                router.replace(transportAndPaymentUrl);
+            }
         }
```
sebaholesz: I do not like nested conditions like this. I am not sure what your source for early returns from useEffects has been, but based on the type definition for useEffects, I do not see anything wrong with using an early return given that there is no destructor provided. If I base my comment on the fact that I have not read anything about it being wrong, I would be glad if you could unwrap the conditions.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
tvikito: I need to specificaly check for undefined because I use `null` and `undefined` for purpose to distinguish if cart has been fetched or exist. In case it exist but it's empty, it's null. Otherwise it's undefined. This allows me to show empty cart message or loading state. Length was already discussed.

project-base/storefront/pages/order/contact-information.tsx
tvikito: We will discuss this on FE meeting. For now, skipping this.

PR #2907, opened 2023-10-30
[shopsys] added core for dispatch/consume system

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Prepares the message dispatch/consume for future usage. More description follows
|New feature| Yes 
|Fixes issues| ... 

What is done:
- `AbstractMessageDispatcher` should be used as a base class for all future dispatchers. Provides the necessary dependencies and is configured automatically
- messages are truly sent to the queue at the end of the request/console-command/processed-message – after the master transaction is committed
- services using internal cache now implement `ResetInterface`, so they are automatically cleared on each message (prevent stale in-memory data)
- when entity manager is closed, the worker is automatically restarted

----
  - https://mg-messenger-messages.odin.shopsys.cloud
  - https://cz.mg-messenger-messages.odin.shopsys.cloud

### DIFF
--- a/UPGRADE-14.0.md
+++ b/UPGRADE-14.0.md
@@ -107,6 +107,10 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   see #project-base-diff to update your project
 -   re-enable phing target cron ([#2875](https://github.com/shopsys/shopsys/pull/2875))
     -   see #project-base-diff to update your project
+-   prepare core for dispatch/consume system ([#2907](https://github.com/shopsys/shopsys/pull/2907))
+    -   your custom classes that utilize internal array caching or need to be reset between message consumption should now implement the `\Symfony\Contracts\Service\ResetInterface` interface
+    -   method `Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPriceRecalculationScheduler::cleanScheduleForImmediateRecalculation()` has been renamed to `reset()`
+    -   see #project-base-diff to update your project
 
 ### Storefront
 
--- a/packages/framework/src/Component/Messenger/ExampleDispatcher.php
+++ b/packages/framework/src/Component/Messenger/ExampleDispatcher.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Messenger;
+
+class ExampleDispatcher extends AbstractMessageDispatcher
+{
+    /**
+     * @param string $value
+     */
+    public function dispatchExampleMessage(string $value): void
+    {
+        $this->messageBus->dispatch(new ExampleMessage($value));
+    }
+}
--- a/packages/product-feed-heureka/src/Model/FeedItem/HeurekaProductDataBatchLoader.php
+++ b/packages/product-feed-heureka/src/Model/FeedItem/HeurekaProductDataBatchLoader.php
@@ -47,6 +47,8 @@ class HeurekaProductDataBatchLoader
             $domainConfig,
         );
 
+        $this->loadedProductCpcs = [];
+
         foreach ($products as $product) {
             $key = $this->getKey($product, $domainConfig);
             $productId = $product->getId();
--- a/composer.json
+++ b/composer.json
@@ -160,6 +160,7 @@
         "symfony/proxy-manager-bridge": "^5.4",
         "symfony/rate-limiter": "^5.4",
         "symfony/security-bundle": "^5.4",
+        "symfony/service-contracts": "^2.5.2",
         "symfony/stopwatch": "^5.4",
         "symfony/translation": "^5.4",
         "symfony/twig-bundle": "^5.4",
--- a/docker/conf/docker-compose.github-actions.cypress.yml.dist
+++ b/docker/conf/docker-compose.github-actions.cypress.yml.dist
@@ -46,6 +46,24 @@ services:
         volumes:
             - web-volume:/var/www/html/project-base/app/web
 
+    php-consumer:
+        restart: always
+        image: php-fpm-image
+        container_name: shopsys-framework-php-consumer
+        volumes:
+            - web-volume:/var/www/html/project-base/app/web
+        depends_on:
+            - php-fpm
+            - rabbitmq
+        entrypoint:
+            - /bin/sh
+            - -c
+            - sleep 5 && project-base/app/docker/php-fpm/consumer-entrypoint.sh 600
+
+    rabbitmq:
+        image: rabbitmq:3.12-management-alpine
+        container_name: shopsys-framework-rabbitmq
+
     redis:
         image: redis:5.0-alpine
         container_name: shopsys-framework-redis
--- a/packages/framework/composer.json
+++ b/packages/framework/composer.json
@@ -93,6 +93,7 @@
         "symfony/proxy-manager-bridge": "^5.4",
         "symfony/rate-limiter": "^5.4",
         "symfony/security-bundle": "^5.4",
+        "symfony/service-contracts": "^2.5.2",
         "symfony-cmf/routing": "^2.0.3",
         "symfony-cmf/routing-bundle": "^2.0.3",
         "symfony/polyfill-php80": "^1.24.0",
--- a/packages/framework/src/Component/Elasticsearch/AbstractExportScheduler.php
+++ b/packages/framework/src/Component/Elasticsearch/AbstractExportScheduler.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Component\Elasticsearch;
 
-abstract class AbstractExportScheduler
+use Symfony\Contracts\Service\ResetInterface;
+
+abstract class AbstractExportScheduler implements ResetInterface
 {
     /**
      * @var int[]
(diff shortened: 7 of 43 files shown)


packages/framework/src/Component/Messenger/ExampleDispatcher.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Shopsys\FrameworkBundle\Component\Messenger;
+
+class ExampleDispatcher extends AbstractMessageDispatcher
+{
+    /**
+     * @param string $value
+     */
+    public function dispatchExampleMessage(string $value): void
+    {
+        if ($this->transportDsn === '') {
```
vitek-rostislav: I have a painful memory from developing my last dispatcher on the previous project. I forgot to put this check in my dispatch method and it was quite hard to debug why my code was not working at some point.

I am wondering whether there is a chance to ease this while reducing the DRY violations in the future code (every dispatcher requires this check in its `dispatch*` method). Is not it possible to ensure the check is performed without actually copying it from one dispatcher to another? :smile: :thinking: Let's discuss it, please

packages/framework/src/Component/Messenger/DeferredEnvelope/DeferEnvelopeMiddleware.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Messenger\DeferredEnvelope;
+
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
+use Symfony\Component\Messenger\Middleware\StackInterface;
+use Symfony\Component\Messenger\Stamp\ReceivedStamp;
+
+class DeferEnvelopeMiddleware implements MiddlewareInterface
```
vitek-rostislav: I am fine with the `Defer` naming, however, I find `Delay` more "predictable" - if I was looking for such functionality, I would expect it to be named with `Delayed` prefix. But maybe it is just me, I definitely do not insist on the change, I just wanted to share my thoughts :slightly_smiling_face:

packages/framework/src/Component/Messenger/DeferredEnvelope/DispatchCollectedEnvelopesSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+    ) {
+    }
+
+    public function handleCollectedMessageEnvelopes(): void
+    {
+        foreach ($this->deferredEnvelopesCollector->popEnvelopes() as $envelope) {
+            try {
+                $this->messageBus->dispatch($envelope);
+            } catch (HandlerFailedException $e) {
+                foreach ($e->getNestedExceptions() as $nested) {
```
vitek-rostislav: This may be extracted into a protected method with a description in the comment (or at least a descriptive name). I mean, it is a pretty specific piece of code in otherwise a general functionality.

packages/framework/src/Component/Messenger/DeferredEnvelope/DispatchCollectedEnvelopesSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            KernelEvents::RESPONSE => 'handleCollectedMessageEnvelopes',
```
vitek-rostislav: How do we ensure this is called after the transaction is committed? Shouldn't we set the priority here?

packages/framework/src/Component/Messenger/DeferredEnvelope/DispatchCollectedEnvelopesSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+            ConsoleEvents::TERMINATE => 'handleCollectedMessageEnvelopes',
```
vitek-rostislav: As we discussed the `KernelEvents::TERMINATE`, what about this one? Is it triggered when the console command fails? If so, do we want to dispatch the message in such a case? I am thinking we could add the event as an argument for `handleCollectedMessageEnvelopes`. If the event is an instance of `ConsoleTerminateEvent`, we can get its return code and dispatch the message only if the return code is OK (just an idea :thinking: )

packages/product-feed-heureka/src/Model/FeedItem/HeurekaProductDataBatchLoader.php
```diff
@@ -47,6 +47,8 @@ public function loadForProducts(array $products, DomainConfig $domainConfig): vo
             $domainConfig,
         );
 
+        $this->loadedProductCpcs = [];
```
vitek-rostislav, line 50: Does not this class need to implement `ResetInterface`?

UPGRADE-14.0.md
```diff
@@ -107,6 +107,10 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   see #project-base-diff to update your project
 -   re-enable phing target cron ([#2875](https://github.com/shopsys/shopsys/pull/2875))
     -   see #project-base-diff to update your project
+-   prepare core for dispatch/consume system ([#2907](https://github.com/shopsys/shopsys/pull/2907))
+    -   your custom classes using internal array caching or should be somehow reset between consuming messages should now implement `\Symfony\Contracts\Service\ResetInterface` interface
```
vitek-rostislav: the sentence is somehow broken :smile:

vitek-rostislav (changes requested):
Hi, nice job, I have added just a few notes

packages/framework/src/Component/Messenger/ExampleDispatcher.php
grossmannmartin: I have created the `NullMessageBus` that does not dispatch the message. When the transportDsn is empty, this message bus is used instead of the default one to prevent errors and code duplication. 

in 45caff6488 and c4124c4efe

packages/framework/src/Component/Messenger/DeferredEnvelope/DeferEnvelopeMiddleware.php
grossmannmartin: good call, renamed in 2e1d353cc8

packages/framework/src/Component/Messenger/DeferredEnvelope/DispatchCollectedEnvelopesSubscriber.php
grossmannmartin: extracted, nice idea 👍 – c77e384e34

packages/framework/src/Component/Messenger/DeferredEnvelope/DispatchCollectedEnvelopesSubscriber.php
grossmannmartin: Yes. Yes, we should. – 50f02b34c2

packages/framework/src/Component/Messenger/DeferredEnvelope/DispatchCollectedEnvelopesSubscriber.php
grossmannmartin: This is the main concern I have about the master transaction listener... It's used only in the request-response flow. Console and event flow do not use the master transaction, so the transaction may be committed even on error and in that case I believe we want to dispatch the message

packages/product-feed-heureka/src/Model/FeedItem/HeurekaProductDataBatchLoader.php
grossmannmartin: The `loadForProducts` is the entrypoint for getting the correct data. So when it's cleared here, it's not necessary to reset the service extra.

UPGRADE-14.0.md
grossmannmartin: the sentence is somehow fixed - d540de7636

vitek-rostislav:
feedback from testing:

- [x] We should respect the transactional master request - when there is an exception, data in DB are rollbacked. In such a case, we should not process the dispatched messages

:heavy_check_mark: When there is an error that causes the closing of the entity manager, the message processing is stopped immediately, and the message is not re-enqueued.

:heavy_check_mark: the messages are actually dispatched on kernel response, on console terminate, and after another message is processed - so it is ensured the data are already committed in the database when handling the message.

:heavy_check_mark: the internal service cache is cleared after a message is processed.

One thing that surprised me - when the message processing fails on a regular exception, it is then enqueued 3 more times - I believe that is ok. For a reason, I was expecting 3 attempts in total but actually, there are 4 attempts to process the message.

grossmannmartin:
> feedback from testing:
> 
> * [ ]  We should respect the transactional master request - when there is an exception, data in DB are rollbacked. In such a case, we should not process the dispatched messages
 
Thanks for testing. I have improved this scenario :) 

> One thing that surprised me - when the message processing fails on a regular exception, it is then enqueued 3 more times - I believe that is ok. For a reason, I was expecting 3 attempts in total but actually, there are 4 attempts to process the message.

🤔  hmmm... I'm not sure. According to this

Snímek obrazovky 2023-11-05 v 20 12 24

it seems like it should be retried 3 times (that makes the 4 attempts in general)

but then this 

Snímek obrazovky 2023-11-05 v 20 14 34

talks about three failings... 

well, all in all, we can adjust the behavior with the max_retries configuration

vitek-rostislav:
> Thanks for testing. I have improved this scenario :)

Awesome, thank you, I have re-tested it, works like a charm, so I am moving the task further in the process :wink:

vitek-rostislav (approved):

PR #2917, opened 2023-11-09
[shopsys] product recalculation and export is now done by queue

### LINKED ISSUE #1171: Elasticsearch isolation while running tests
Writing tests to check some functionality bound to Elasticsearch, more specifically modifying data in Elasticsearch is now nearly impossible because the state of data in Elasticsearch is not isolated between each test.

Consider the following use-case (data fixtures are loaded):
1. Write test `ProductIsDeletedFromSearch`. 
  1.1 Product with `ID=1` is deleted from DB
  1.2 Data are exported to Elasticsearch
  1.3 Some `Assert::notExists` assertion is executed and pass
  1.4 Test is finished and DB transaction is rolled back
2. Write test `ProductCanBeSearched`
  2.1 Query to Elasticsearch tries to find product with `ID=1`
  2.2 Some `Assert::exists` is executed and **fail!**

Because even when the product is still in the database thanks to rollbacking the transaction, Changes in Elasticsearch are permanent.

_Elasticsearch does not support transactions the same way as PostgreSQL._

### Expected result
Running tests should not affect each other when data in Elasticsearch is modified in one of them or in the executed code.

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Product recalculations (visibility, availability, prices, ...) and export to Elasticsearch were scattered across the codebase and done differently in many places. This PR aims to unify the approach no matter from where a product should be recalculated (cron, console, request) and to ease the necessary steps to make the product right (what has to be called and in which order)
|New feature| Yes 
|Fixes issues| closes #1171  

This PR adds a few new concepts that should be known:
- most of the calculated fields were removed and the usage was either no longer necessary or changed
- at any place where a product is somehow changed, only, and the *ONLY* `ProductRecalculationDispatcher::dispatch*()` method has to be called
  - this method dispatches a message for the message broker and async handler then immediately recalculate the product and export to elastic
- in tests, the real queue is not used – instead where necessary (when testing product changes), the `Tests\App\Test\WebTestCase::handleDispatchedRecalculationMessages()` method should be called to run scheduled recalculations – that way we are sure the code works in real life – the message is truly dispatched, thus product recalculated
  - calling this method also creates a snapshot in Elasticsearch before any changes are exported and restores it afterward – tests are no longer dependent on the changes made, nor the order of run tests
- variants are handled automatically, so it's enough to dispatch a single variant, or main variant to make the whole group right
- new command `shopsys:dispatch:recalculations` allows manual dispatch message(s) for recalculation
- when the whole catalog should be dispatched, only a single message "dispatch all" is dispatched and then the async handler takes care of dispatching all product IDs – that way we may dispatch all products to recalculation during request without worrying about the size of the catalog

----
  - https://mg-messenger-recalculation.odin.shopsys.cloud
  - https://cz.mg-messenger-recalculation.odin.shopsys.cloud


packages/framework/src/Command/DispatchRecalculationMessageCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        return $this->executeIds($productIds, $symfonyStyle);
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Style\SymfonyStyle $symfonyStyle
+     * @return int
+     */
+    private function executeAll(SymfonyStyle $symfonyStyle): int
```
vitek-rostislav: this should be protected

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
```diff
@@ -0,0 +1,105 @@
...
+
+        return Command::SUCCESS;
+    }
+
+    /**
+     * @param int[] $productIds
+     * @param \Symfony\Component\Console\Style\SymfonyStyle $symfonyStyle
+     * @return int
+     */
+    private function executeIds(array $productIds, SymfonyStyle $symfonyStyle): int
```
vitek-rostislav: this should be protected

packages/framework/src/Model/Product/Recalculation/DispatchAllProductsMessageHandler.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Shopsys\FrameworkBundle\Model\Product\ProductFacade;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+class DispatchAllProductsMessageHandler implements MessageHandlerInterface
```
vitek-rostislav, line 10: as we agreed recently, we should use the configuration of handlers via the attribute

packages/framework/src/Model/Product/Recalculation/DispatchAllProductsMessage.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+class DispatchAllProductsMessage
```
vitek-rostislav, line 7: At first, I wondered whether we needed a message to dispatch other messages, I thought that `ProductRecalculationDispatcher::dispatchAllProducts` could fetch all the IDs and then directly use `dispatchProductIds`, the current solution seemed a little over-engineered. 

However, the more I think about it, the more it makes sense to me - this way you do not need any resources during the synchronous request to dispatch the message as the "get all IDs" logic is in the handler (i.e. the async part), is that the motivation behind the solution? I do not think it is over-engineered anymore, I think it is an awesome approach :slightly_smiling_face:

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -0,0 +1,42 @@
...
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\IndexFacade $indexFacade
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\IndexRegistry $indexRegistry
+     * @param \Shopsys\FrameworkBundle\Component\Elasticsearch\IndexDefinitionLoader $indexDefinitionLoader
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     */
+    public function __construct(
+        private readonly IndexFacade $indexFacade,
+        private readonly IndexRegistry $indexRegistry,
+        private readonly IndexDefinitionLoader $indexDefinitionLoader,
+        private readonly Domain $domain,
```
vitek-rostislav: protected?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationMessage.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+class ProductRecalculationMessage
+{
+    /**
+     * @param int $productId
+     */
+    public function __construct(
+        private readonly int $productId,
```
vitek-rostislav: what about public readonly? :innocent: 

Also, I would like to talk about the "1 message = 1 product ID" vs "1 message = multiple product IDs" topic, please. Maybe we have already discussed it, I just do not remember the arguments here.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationMessageHandler.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Messenger\Handler\Acknowledger;
+use Symfony\Component\Messenger\Handler\BatchHandlerInterface;
+use Symfony\Component\Messenger\Handler\BatchHandlerTrait;
+use Throwable;
+
+class ProductRecalculationMessageHandler implements BatchHandlerInterface
```
vitek-rostislav, line 13: hm, is there an attribute alternative for this interface? :smile: It seems like there is no such thing :confused:

vitek-rostislav: btw I think I am starting to understand why you dispatch messages with just one product ID :smile:

packages/framework/src/Model/Product/Elasticsearch/ProductIndex.php
```diff
@@ -6,9 +6,8 @@
 
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractIndex;
-use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexSupportChangesOnlyInterface;
 
-class ProductIndex extends AbstractIndex implements IndexSupportChangesOnlyInterface
```
vitek-rostislav, line 11: I wondered whether we could get rid of the interface completely, however, it is used for blog articles and their further implementation is something I see the first time in my life, there is a queue in Redis apparently, so ok, this is something I can look forward to in my task :smile:

packages/framework/src/Model/Product/Elasticsearch/ProductExportCronModule.php
```diff
@@ -4,25 +4,33 @@
...
-use Shopsys\FrameworkBundle\Component\Domain\Domain;
-use Shopsys\FrameworkBundle\Component\Elasticsearch\AbstractExportCronModule;
-use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexDefinitionLoader;
-use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexFacade;
+use Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDispatcher;
+use Shopsys\Plugin\Cron\SimpleCronModuleInterface;
+use Symfony\Bridge\Monolog\Logger;
 
-class ProductExportCronModule extends AbstractExportCronModule
+class ProductExportCronModule implements SimpleCronModuleInterface
```
vitek-rostislav: Do you have any particular use case in mind when this CRON module is useful? I feel like it is a good safety net to have all the products recalculated once a day, however, maybe it is not needed (provided all relevant actions dispatch the recalculation messages) :thinking: it is true that one will probably never cover all the places, so I suppose the existence of this module is justified

If we agree on keeping the module, I would suggest renaming it to `ProductRecalculationCronModule`

packages/framework/src/Model/Category/CategoryEvent.php
```diff
@@ -0,0 +1,45 @@
...
+
+class CategoryEvent extends Event
+{
+    /**
+     * The CREATE event occurs once a category was created.
+     *
+     * This event allows you to run jobs dependent on the category creation.
+     */
+    public const CREATE = 'category.create';
+    /**
```
vitek-rostislav, line 17: I would expect an empty line here, standards are ok with that?

packages/framework/src/Model/Category/CategoryEvent.php
```diff
@@ -0,0 +1,45 @@
...
+     * The UPDATE event occurs once a category was changed.
+     *
+     * This event allows you to run jobs dependent on the category change.
+     */
+    public const UPDATE = 'category.update';
+    /**
```
vitek-rostislav, line 23: same here, I would expect an empty line here, standards are ok with that?

packages/framework/src/Model/Category/CategoryEvent.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Contracts\EventDispatcher\Event;
+
+class CategoryEvent extends Event
+{
+    /**
+     * The CREATE event occurs once a category was created.
+     *
+     * This event allows you to run jobs dependent on the category creation.
+     */
+    public const CREATE = 'category.create';
```
vitek-rostislav, line 16: Instead of the constants, it would be nice to have an enum, however, I do not insist on that now, we can do that for all these events later :wink:

vitek-rostislav: the truth is, it could be a little weird for usage, we would have to use something like `Enum::EVENT_NAME->value` as the dispatcher expects string for the event name...

packages/framework/src/Model/Product/AffectedProductsFacade.php
```diff
@@ -0,0 +1,77 @@
...
+namespace Shopsys\FrameworkBundle\Model\Product;
+
+use Shopsys\FrameworkBundle\Model\Category\Category;
+use Shopsys\FrameworkBundle\Model\Product\Availability\Availability;
+use Shopsys\FrameworkBundle\Model\Product\Brand\Brand;
+use Shopsys\FrameworkBundle\Model\Product\Flag\Flag;
+use Shopsys\FrameworkBundle\Model\Product\Parameter\Parameter;
+use Shopsys\FrameworkBundle\Model\Product\Unit\Unit;
+
+class AffectedProductsFacade
```
vitek-rostislav, line 13: There are quite a lot of new concepts, we definitely should not forget about the docs before releasing the version :smile:  I can even imagine a webinar or something like that

vitek-rostislav: btw why do I think we have already had the "affected" facades?

vitek-rostislav: btw why do I think we have already had the "affected" facades?

packages/framework/src/Model/Product/AffectedProductsRepository.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Availability\Availability $availability
+     * @return int[]
+     */
+    public function getProductIdsWithAvailability(Availability $availability): array
+    {
+        $queryBuilder = $this->createQueryBuilder()
+            ->where('p.calculatedAvailability = :availability')
```
vitek-rostislav: should not here be `OR p.availability = :availability`? But I do not see the method in the resulting code at all, so I will probably find out later :slightly_smiling_face:

packages/framework/src/Model/Product/AffectedProductsRepository.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Flag\Flag $flag
+     * @return int[]
+     */
+    public function getProductIdsWithFlag(Flag $flag): array
+    {
+        $queryBuilder = $this->createQueryBuilder()
+            ->leftJoin('p.flags', 'f')
```
vitek-rostislav: this will be probably in conflict with Tom's task - he was moving flags to product domains, if I remember it correctly

vitek-rostislav: https://github.com/shopsys/shopsys/pull/2939/commits/a3dbc46a35c01b0a23ac7abdc3b1f73528bc9088#diff-41cbda7848fbd86f06d6d510b6dace9f3b20ff05967afd731fed86e9653a2c5e

packages/framework/src/Model/Product/Recalculation/DispatchAffectedProductsSubscriber.php
```diff
@@ -0,0 +1,128 @@
...
+
+        $this->productRecalculationDispatcher->dispatchProductIds($productIds);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Stock\StockEvent $stockEvent
+     */
+    public function dispatchAllProductsIfStockDomainsChanged(StockEvent $stockEvent): void
+    {
+        if ($stockEvent->hasChangedDomains()) {
```
vitek-rostislav, line 78: On the way I noticed `StockEvent::$stock` is actually unused, maybe we can remove it?

project-base/app/src/Model/Category/CategoryFacade.php
```diff
@@ -125,35 +131,9 @@ public function edit($categoryId, CategoryData $categoryData): BaseCategory
         $this->categoryParameterFacade->saveRelation($category, $categoryData->parametersPosition, $categoryData->parametersCollapsed);
         $this->linkedCategoryFacade->updateLinkedCategories($category, $categoryData->linkedCategories);
 
-        $this->scheduleProductsToExportByCategory($category);
```
vitek-rostislav, line 128: something to think about - when the parent method dispatches an event, is not it dangerous to extend the method this way? I mean, the category is modified, the event is dispatched and handled by the subscriber, but the extended method then further modifies the category (and does not dispatch the event).

I am not sure this is something we can/should solve here, but I would definitely like to talk about this.

packages/framework/src/Model/Product/ProductRepository.php
```diff
@@ -536,26 +536,4 @@ public function getAllOfferedProducts(int $domainId, PricingGroup $pricingGroup)
     {
         return $this->getAllOfferedQueryBuilder($domainId, $pricingGroup)->getQuery()->execute();
     }
-
-    /**
-     * @param \Shopsys\FrameworkBundle\Model\Product\Product[] $products
-     */
-    public function markProductsForExport(array $products): void
```
vitek-rostislav, line 528: I believe this task is a great opportunity to solve our "win-win upgrade automation" issue. It seems like solving the issue will take less time than writing down all the upgrade notes for the changes in this pull request :pray:

packages/framework/build.xml
```diff
@@ -322,7 +322,8 @@
     <target name="db-demo" depends="production-protection,frontend-api-generate-new-keys,db-wipe-public-schema,clean,clean-redis,db-import-basic-structure,db-migrations,domains-data-create,db-fixtures-demo,plugin-demo-data-load,friendly-urls-generate,friendly-url-entity-mapping-check,domains-urls-replace" description="Creates DB and fills it with demo data" hidden="true"/>
 
     <target name="db-fixtures-demo" depends="production-protection" description="Loads demo data fixtures." hidden="true">
-        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+        <exec executable="MESSENGER_TRANSPORT_DSN=" passthru="true" checkreturn="true">
```
vitek-rostislav, line 325: just making a note here, I would like to talk about the commit message:
> - after datafixtures are loaded, other tasks have to be done so it's pointless to let consumer export products

packages/framework/src/Resources/config/services_test.yaml
```diff
@@ -28,3 +28,5 @@ services:
         arguments:
             $indexDefinitionsDirectory: '%shopsys.elasticsearch.structure_dir%'
             $indexPrefix: 'test_%env(ELASTIC_SEARCH_INDEX_PREFIX)%'
+
+    Tests\FrameworkBundle\Test\ProductIndexBackupFacade: ~
```
vitek-rostislav, line 32: you can close https://github.com/shopsys/shopsys/issues/1171  with this pull request, right? :tada:

packages/framework/tests/Test/ProductIndexBackupFacade.php
```diff
@@ -0,0 +1,123 @@
...
+        $this->cloneIndexToIndex($indexName, $backupIndexName);
+        $this->isSnapshotCreated = true;
+    }
+
+    public function restoreSnapshotIfPreviouslyCreated(): void
+    {
+        $indexName = $this->getFirstDomainIndexName();
+        $backupIndexName = $this->getBackupIndexName();
+
+        if (!$this->isSnapshotCreated) {
```
vitek-rostislav, line 49: micro-optimization, this condition could be the first thing in this method

packages/framework/src/Model/Transport/Transport.php
```diff
@@ -386,4 +393,12 @@ public function getUuid(): string
     {
         return $this->uuid;
     }
+
+    /**
+     * @return int
+     */
+    public function getDaysUntilDelivery(): int
```
vitek-rostislav: as I learned recently, we should not add type hints to getters in entities :frowning_face:

project-base/app/src/Model/Transport/Transport.php
```diff
@@ -25,12 +25,6 @@
  */
 class Transport extends BaseTransport
 {
-    /**
-     * @var int
-     * @ORM\Column(type="integer")
-     */
-    private int $daysUntilDelivery;
```
vitek-rostislav, line 32: you said you would deal with the migrations later, right?

vitek-rostislav, line 32: you said you would deal with the migrations later, right?

packages/framework/src/Model/Product/ProductDomain.php
```diff
@@ -201,4 +207,20 @@ public function setVat(Vat $vat): void
     {
         $this->vat = $vat;
     }
+
+    /**
+     * @return bool
+     */
+    public function getSaleExclusion(): bool
```
vitek-rostislav: unwanted return type hint in entity getter

packages/framework/src/Model/Product/Product.php
```diff
@@ -975,11 +970,12 @@ public function getSeoMetaDescription(int $domainId)
     }
 
     /**
-     * @return bool
+     * @param int $domainId
+     * @return  bool
      */
-    public function isMarkedForVisibilityRecalculation()
+    public function getSaleExclusion(int $domainId): bool
```
vitek-rostislav: unwanted return type hint in entity getter...even thought here I can not imaging someone is changing the type, but they might want to make it nullable :shrug:

packages/framework/src/Model/Product/Availability/AvailabilityStatusEnum.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Availability;
+
+enum AvailabilityStatusEnum: string implements AvailabilityStatusEnumInterface
```
vitek-rostislav, line 7: :green_heart:

packages/framework/src/Model/Product/Availability/ProductStoreAvailabilityInformation.php
```diff
@@ -2,21 +2,21 @@
...
      * @param string $availabilityInformation
-     * @param \App\Model\Product\Availability\AvailabilityStatusEnum $availabilityStatus
+     * @param \Shopsys\FrameworkBundle\Model\Product\Availability\AvailabilityStatusEnum $availabilityStatus
      */
     public function __construct(
         private readonly string $storeName,
         private readonly int $storeId,
         private readonly string $availabilityInformation,
-        private readonly AvailabilityStatusEnum $availabilityStatus,
+        private readonly AvailabilityStatusEnumInterface $availabilityStatus,
```
vitek-rostislav: protected

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductSellingDeniedOnDomainTest.php
```diff
@@ -46,7 +40,7 @@ public function testSellingDeniedOnDomain(): void
 
         $this->dispatchFakeKernelResponseEventToTriggerImmediateRecalculations();
```
vitek-rostislav: is this line still needed? Ah, I see it is replaced in the WIP commit with `handleDispatchedRecalculationMessages`. So now, you call it twice in this test

vitek-rostislav (changes requested):

packages/framework/src/Command/RecalculationsCommand.php
```diff
@@ -65,9 +62,6 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         // Main variant is set for recalculations after change of variants visibility.
         $this->productPriceRecalculator->runAllScheduledRecalculations();
 
-        $output->writeln('<fg=green>Products availability.</fg=green>');
```
vitek-rostislav, line 68: let's discuss this further, please - I believe with this step, we should also remove the `$outOfStockAvailabilty` property, as well as `$usingStock` and `$outOfStockAction` properties. In the end, even the `$availability` field itself is not used now, right? 

I believe you have already taken this ride on the train of thoughts :smile: It probably deserves a separate task

packages/framework/src/Model/Product/Filter/ProductFilterRepository.php
```diff
@@ -43,10 +40,6 @@ public function applyFiltering(
             $productFilterData->maximalPrice,
             $pricingGroup,
         );
-        $this->filterByStock(
```
vitek-rostislav, line 46: is it ok that this is completely removed? there is still `ProductFilterData::$inStock` property so this removal will break the filtering by availability functionality, or? 
image

While I am writing this comment, I figured - the class is not used at all, right? I mean, we are using the Elasticsearch implementation and we can get rid of this along with `ProductFilterCountRepository` and `ProductOnCurrentDomainFacadeInterface`...I thought it might be removed in https://github.com/shopsys/shopsys/pull/2935 but for this, we need another task probably

project-base/app/src/Model/Product/Availability/AvailabilityFacade.php
```diff
@@ -9,30 +9,8 @@
 /**
  * @property \App\Model\Product\Availability\AvailabilityRepository $availabilityRepository
  * @property \App\Component\Setting\Setting $setting
- * @method __construct(\Doctrine\ORM\EntityManagerInterface $em, \App\Model\Product\Availability\AvailabilityRepository $availabilityRepository, \App\Component\Setting\Setting $setting, \Shopsys\FrameworkBundle\Model\Product\Availability\ProductAvailabilityRecalculationScheduler $productAvailabilityRecalculationScheduler, \Shopsys\FrameworkBundle\Model\Product\Availability\AvailabilityFactoryInterface $availabilityFactory, \Symfony\Component\EventDispatcher\EventDispatcherInterface $eventDispatcher)
+ * @method __construct(\Doctrine\ORM\EntityManagerInterface $em, \App\Model\Product\Availability\AvailabilityRepository $availabilityRepository, \App\Component\Setting\Setting $setting, \Shopsys\FrameworkBundle\Model\Product\Availability\AvailabilityFactoryInterface $availabilityFactory, \Symfony\Component\EventDispatcher\EventDispatcherInterface $eventDispatcher)
  */
 class AvailabilityFacade extends BaseAvailabilityFacade
```
vitek-rostislav: do we want to keep the empty extended classes in the project base? I mean, it has both pros and cons so I am just curious whether there is a consensus about this.

packages/framework/src/Model/Category/CategoryFacade.php
```diff
@@ -143,7 +143,7 @@ public function edit($categoryId, CategoryData $categoryData)
 
         $this->pluginCrudExtensionFacade->saveAllData('category', $category->getId(), $categoryData->pluginData);
 
-        $this->categoryVisibilityRecalculationScheduler->scheduleRecalculation($category);
```
vitek-rostislav, line 138: I was struggling for a while to figure this out but I believe I get it. You do not need to pass the category to the scheduler for the category recalculation (it was a violation of SRP - its responsibility was also marking the affected products for their visibility recalculation, right?). And this is not necessary as it is covered thanks to the dispatched `CategoryEvent`, yes?

packages/framework/src/Model/Product/Recalculation/DispatchAllProductsMessage.php
grossmannmartin: Yes, it's truly the reason behind it. That way we are no longer concerned about the size of the catalog while doing request-response related operations where we need to dispatch all products

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
grossmannmartin: it's changed in the later commits, I try to move it to proper commit

packages/framework/src/Model/Product/Product.php
```diff
@@ -659,61 +643,11 @@ public function getBrand()
...
-             * When the product is fetched from persistence, the mainVariant is only a proxy object,
-             * when we call something on this proxy object, Doctrine fetches it from persistence too.
-             *
-             * The problem is the Doctrine seems to not fetch the main variant when we only write something into it,
-             * but when we read something first, Doctrine fetches the object, and the use-case works correctly.
-             *
-             * If you think this is strange and it shouldn't work even before the code was moved to Product, you are right, this is strange.
-             * When the code is outside of Product, Doctrine does the job correctly, but once the code is inside of Product,
-             * Doctrine seems to not fetching the main variant.
-             */
```
vitek-rostislav, line 715: I am so glad this is gone :joy:

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
grossmannmartin: changed

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
grossmannmartin: changed

packages/framework/src/Model/Product/Recalculation/DispatchAllProductsMessageHandler.php
grossmannmartin: ~thanks! changed~

I have changed it, but then I realized (thanks to comment https://github.com/shopsys/shopsys/pull/2917#discussion_r1420430761) that it's not possible to use due to BatchHandlerInterface used in the final code :(

packages/framework/src/Model/Product/Recalculation/ProductRecalculationMessage.php
grossmannmartin: changed to public readonly :) good idea


About 1 message = 1 id, it seems more versatile to me at the moment. We can discuss it further :)

packages/framework/src/Model/Product/Recalculation/ProductRecalculationMessageHandler.php
grossmannmartin: sadly, the attribute for batch processing does not exists

packages/framework/src/Model/Product/Elasticsearch/ProductIndex.php
grossmannmartin: in the future it would be nice to remove this interface completely 👍

packages/framework/src/Model/Product/Elasticsearch/ProductExportCronModule.php
grossmannmartin: You nicely summarized my thoughts exactly 😄 
I would like to keep the cron module (at least for now). It may be easily disabled in a project if it becomes unnecessary. 

Good thinking about the renaming though. Done

packages/framework/src/Model/Category/CategoryEvent.php
grossmannmartin: Apparently 🤔 

It's the same in the other events (UnitEvent, ParameterEvent, BrandEvent, FlagEvent, ...)

packages/framework/src/Model/Category/CategoryEvent.php
grossmannmartin: same answer as above 🙂

project-base/app/tests/App/Functional/Model/Product/ProductVisibilityRepositoryTest.php
```diff
@@ -149,132 +151,82 @@ public function testIsVisibleOnAnyDomainWhenHidden()
...
         $productAgain = $this->em->getRepository(Product::class)->find($id);
 
+        return $this->getVisibilityForProduct($productAgain);
+    }
+
+    /**
+     * @param \App\Model\Product\Product $productAgain
+     * @return \Shopsys\FrameworkBundle\Model\Product\ProductVisibility
+     */
+    public function getVisibilityForProduct(Product $productAgain): ProductVisibility
```
vitek-rostislav: nitpick, can be private

packages/framework/src/Model/Category/CategoryEvent.php
grossmannmartin: We have to probably split the event and the event names into separate classes (class and enum) because the event now has a constructor and may return the value of the dispatched category (flag, brand, etc.). 

I'm not saying it's wrong, but I would like to evaluate the solution in a separate task 😇

packages/framework/src/Model/Product/AffectedProductsFacade.php
grossmannmartin: Yes, yes we should not. 

The `affected` word was used in a class in the previous project we worked on. 
And in the platform in `MarkProductForExportSubscriber` in method names (like `markAffectedByFlag`)

project-base/app/tests/App/Functional/Model/Product/ProductFacadeTest.php
```diff
@@ -47,13 +47,11 @@ class ProductFacadeTest extends TransactionFunctionalTestCase
      * @dataProvider getTestCalculationHiddenAndSellingDeniedDataProvider
      * @param mixed $hidden
      * @param mixed $sellingDenied
-     * @param mixed $calculatedHidden
      * @param mixed $calculatedSellingDenied
      */
     public function testCalculationHiddenAndSellingDenied(
         $hidden,
         $sellingDenied,
-        $calculatedHidden,
```
vitek-rostislav, line 58: maybe the word "hidden" can now be omitted from the name of the test (and data provider). Also, I am wondering whether the `handleDispatchedRecalculationMessages` call is not missing from this method. The truth is, it is not necessary here yet because the selling denied is not yet processed by the queue at the moment, right?

packages/framework/src/Model/Product/Product.php
```diff
@@ -551,9 +551,9 @@ public function setProductCategoryDomains(array $productCategoryDomains)
 
         foreach ($productCategoryDomains as $productCategoryDomain) {
             if ($this->isProductCategoryDomainInArray(
-                    $productCategoryDomain,
-                    $this->productCategoryDomains->getValues(),
-                ) === false) {
+                $productCategoryDomain,
```
vitek-rostislav: can be a fixup of [removed calculated hidden field](https://github.com/shopsys/shopsys/pull/2917/commits/39abd6494a1b82df2d3a66be3acc5226b8ad73c6)

phpstan.neon
```diff
@@ -66,6 +66,8 @@ parameters:
         - %currentWorkingDirectory%/packages/framework/tests/Unit/Component/ClassExtension/Source/*
         # Exclude coding standards for generated files
         - %currentWorkingDirectory%/project-base/app/tests/App/Test/Codeception/_generated/*
+        # Exclude ignored entity extension test
```
vitek-rostislav: the commit message is a little bit confusing, the test was skipped before, right? In this commit, you are just excluding the test from the coding standards checks

project-base/app/tests/FrontendApiBundle/Functional/Cart/AnonymousAddOrderItemsToCartTest.php
```diff
@@ -60,7 +60,7 @@ public function testOrderItemsAreAddedToNotExistingCart(bool $shouldMerge, ?stri
         );
         $data = $this->getResponseDataForGraphQlType($response, 'AddOrderItemsToCart');
         $this->assertArrayHasKey('items', $data);
-        $this->assertCartItems($expectedItems, $data['items']);
+        $this->assertArrayElements($expectedItems, $data['items']);
```
vitek-rostislav, line 63: I do not think this test should be modified, it is about cart items :thinking:

vitek-rostislav: Ah, I see, it is just a refactoring. Hm, I would say the order of the cart items is important and their order should be taken into account in tests, but that is another story

packages/framework/src/Model/Product/Product.php
```diff
@@ -587,6 +587,15 @@ protected function isProductCategoryDomainInArray(
         return false;
     }
 
+    /**
+     * @param int $domainId
+     * @return bool|null
+     */
+    public function isDomainHidden(int $domainId): ?bool
```
vitek-rostislav: should this have the return type hint? :grimacing:

packages/framework/src/Model/Product/ProductDomain.php
```diff
@@ -223,4 +229,20 @@ public function setSaleExclusion(bool $saleExclusion): void
...
+     */
+    public function setDomainHidden(bool $domainHidden): void
+    {
+        $this->domainHidden = $domainHidden;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isDomainHidden(): bool
```
vitek-rostislav: also here, probably an unwanted return type hint

packages/framework/src/Model/Product/ProductDomain.php
```diff
@@ -94,6 +94,12 @@ class ProductDomain
      */
     protected $saleExclusion;
 
+    /**
+     * @var bool
+     * @ORM\Column(type="boolean")
+     */
+    protected $domainHidden;
```
vitek-rostislav, line 145: a database migration will be added later, right?

packages/framework/src/Model/Category/CategoryParameter.php
```diff
@@ -56,15 +56,15 @@ public function __construct(Category $category, Parameter $parameter, bool $coll
     }
 
     /**
-     * @return \App\Model\Category\Category
+     * @return \Shopsys\FrameworkBundle\Model\Category\Category
      */
     public function getCategory(): Category
```
vitek-rostislav: after moving the class to the framework package, you should remove the return type hints from the getters (and I think we need an automated check for this)

packages/framework/src/Model/Pricing/Currency/CurrencyFacade.php
```diff
@@ -81,7 +81,8 @@ public function edit($currencyId, CurrencyData $currencyData)
             $currency->setExchangeRate($currencyData->exchangeRate);
         }
         $this->em->flush();
-        $this->productPriceRecalculationScheduler->scheduleAllProductsForDelayedRecalculation();
+
+        $this->productRecalculationDispatcher->dispatchAllProducts();
```
vitek-rostislav: I need to mention that this approach is slightly different from the other places where the events are used (e.g. `CategoryEvent`, `FlagEvent`, etc.

// edit: or is it not? :thinking: When it is necessary to dispatch all the products, we do not use the events (e.g. in `CategoryFacade::reorderByNestedSetValues()`*), which probably makes sense. I am not sure, if we are consistent in this at the moment, let's discuss this briefly, please.

\* On the other hand, there is `PricingGroupFacade::create()` that dispatches `PricingGroupEvent::CREATE` and then the subscriber dispatches all the products for recalculation

packages/framework/src/Model/Product/AffectedProductsRepository.php
grossmannmartin: this method was removed altogether

packages/framework/src/Model/Product/Recalculation/DispatchAffectedProductsSubscriber.php
grossmannmartin: I would like to keep it to make adjustments easier (when you need to do something in a project depending on certain stock change)

packages/framework/src/Model/Product/Pricing/ProductCalculatedPrice.php
```diff
@@ -4,72 +4,9 @@
...
-use Doctrine\ORM\Mapping as ORM;
 use Shopsys\FrameworkBundle\Component\Money\Money;
-use Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup;
-use Shopsys\FrameworkBundle\Model\Product\Product;
 
-/**
- * @ORM\Table(name="product_calculated_prices")
- * @ORM\Entity
- */
 class ProductCalculatedPrice
```
vitek-rostislav: this is left as an empty unused class, can we remove it?

vitek-rostislav: ...we should probably remove `ProductCalculatedPricesType` and `productCalculatedPrices.html.twig` as well

packages/framework/src/Model/Product/ProductVariantFacade.php
```diff
@@ -39,29 +35,21 @@ public function __construct(
      */
     public function createVariant(Product $mainProduct, array $variants)
     {
+        // parent method is not called intentionally
```
vitek-rostislav: there is no parent class :slightly_smiling_face:

vitek-rostislav: btw, do we still need the extended `ProductVariantFacade` in the project base? there is still some logic but I am not able to decode what is its real purpose

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -34,14 +38,75 @@ public function __construct(
...
                 $this->indexRegistry->getIndexByIndexName(ProductIndex::getName()),
                 $this->indexDefinitionLoader->getIndexDefinition(ProductIndex::getName(), $domainId),
-                $productIds,
+                $idsToRecalculate,
             );
         }
     }
+
+    /**
+     * @param array $productIds
```
vitek-rostislav: int[]

project-base/app/src/Model/Category/CategoryFacade.php
grossmannmartin: I was kinda hoping the delayed dispatch would take care of this problem... let's discuss it, please :)

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -34,14 +38,75 @@ public function __construct(
...
+     * @return array
```
vitek-rostislav: int[]

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -34,14 +38,75 @@ public function __construct(
...
+     * @return int[]
+     */
+    protected function getVariantIds(array $mainVariantsIds): array
+    {
+        $result = $this->em->createQuery('SELECT p.id FROM ' . Product::class . ' p WHERE p.mainVariant IN (:productIds) AND p.variantType = :variantTypeVariant')
+            ->setParameter('productIds', $mainVariantsIds)
+            ->setParameter('variantTypeVariant', Product::VARIANT_TYPE_VARIANT)
+            ->getResult();
+
+        return array_column($result, 'id');
```
vitek-rostislav: these methods fit a little bit more into `ProductRepository` IMHO, but I do not insist on that

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -34,14 +38,75 @@ public function __construct(
...
+
+    /**
+     * @param int[] $productIds
+     * @return int[]
+     */
+    protected function dropVariantIdsFromProductIds(array $productIds): array
+    {
+        $result = $this->em->createQuery('SELECT p.id FROM ' . Product::class . ' p WHERE p.id IN (:productIds) AND (p.variantType = :variantTypeMain OR p.variantType = :variantTypeNone)')
+            ->setParameter('productIds', $productIds)
+            ->setParameter('variantTypeMain', Product::VARIANT_TYPE_MAIN)
```
vitek-rostislav: I am wondering which approach is safer:
```
p.variantType = :variantTypeMain OR p.variantType = :variantTypeNone
```
versus
```
p.variantType != :variantTypeVariant
```

Maybe rather the second one? I mean, if someone adds another variant type, they will need to add the type here with the first option, right?

This is really hypothetical, I believe we will be ok even with the first solution, I do not remember anyone adding another variant type, so :shrug:

packages/framework/src/Model/Product/ProductRepository.php
grossmannmartin: it would be nice to solve it prior this issue... in a separate task 😄

packages/framework/build.xml
grossmannmartin: let's talk about the commit message

packages/framework/src/Resources/config/services_test.yaml
grossmannmartin: true, thanks. Added into PR description

project-base/app/tests/App/Functional/Model/Product/ProductOnCurrentDomainElasticFacadeCountDataTest.php
```diff
@@ -57,6 +57,7 @@ public function testCategory(): void
                 $this->domain->getId(),
```
vitek-rostislav: I have just noticed that the first parameter (`$domainId`) in `ProductFilterConfigFactory::createForCategory()` method is not used at all, could you please take a look at it?

packages/framework/tests/Test/ProductIndexBackupFacade.php
grossmannmartin: 👍  moved

project-base/app/src/Model/Transport/Transport.php
grossmannmartin: Yes. Yes, I will.

packages/framework/src/Model/Product/Availability/ProductStoreAvailabilityInformation.php
grossmannmartin: 🤔  why the standards are not complaining?

vitek-rostislav (changes requested):
Hi, awesome job there, I really like the way many places are simplified :ok_hand: I have added a few more notes so please check them out :wink:

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductSellingDeniedOnDomainTest.php
grossmannmartin: thanks, removed one call

packages/framework/src/Model/Product/Availability/ProductStoreAvailabilityInformation.php
grossmannmartin: Oh, that's why – https://github.com/shopsys/shopsys/pull/2944

packages/framework/src/Command/RecalculationsCommand.php
grossmannmartin: Yes, yes it deserves the separate task. Please 😄

packages/framework/src/Model/Category/CategoryFacade.php
grossmannmartin: Exactly 🙂

packages/framework/src/Model/Product/Product.php
grossmannmartin: Me too 😄

project-base/app/tests/App/Functional/Model/Product/ProductVisibilityRepositoryTest.php
grossmannmartin: changed, thanks!

packages/framework/src/Model/Product/ProductVariantFacade.php
grossmannmartin: 😅  the comment should be in the child method. removed from here. 

I believe the extended implementation does not copy the product into the main variant but makes the main variant directly from the product.
We may probably move this implementation into the framework as it makes much more sense.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
grossmannmartin: changed, thanks

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
grossmannmartin: changed, thanks

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
grossmannmartin: maybe a new repository for the new feature? 😇

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
grossmannmartin: your proposal is easier to read and make sense to me (based on the method name), so changed

project-base/app/tests/App/Functional/Model/Product/ProductOnCurrentDomainElasticFacadeCountDataTest.php
grossmannmartin: removed, thanks

packages/framework/src/Model/Product/Pricing/ProductCalculatedPrice.php
grossmannmartin: I have removed the ProductCalculatedPrice file, thanks!

About the ProductCalculatedPricesType and productCalculatedPrices.html.twig – they're still used in the administration. It just not use the product calculated prices, but calculates them itself

packages/framework/build.xml
grossmannmartin: I will rewrite the commit message to be more clear

project-base/app/tests/FrontendApiBundle/Functional/Cart/AnonymousAddOrderItemsToCartTest.php
grossmannmartin: Let's create a separate issue for that :)

packages/framework/src/Model/Pricing/Currency/CurrencyFacade.php
grossmannmartin: makes sense to unify it 👍  done

packages/framework/src/Model/Product/Product.php
grossmannmartin: fixuped :)  (and squashed 😇 )

packages/framework/build.xml
grossmannmartin: rewritten, better?

packages/framework/src/Model/Transport/Transport.php
grossmannmartin: removed return type

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
grossmannmartin: moved :)

packages/framework/src/Model/Product/ProductDomain.php
grossmannmartin: removed return type hint

packages/framework/src/Model/Product/Product.php
grossmannmartin: removed return type hint

project-base/app/src/Model/Product/Availability/AvailabilityFacade.php
grossmannmartin: I have removed the empty class as we agreed

project-base/app/tests/App/Functional/Model/Product/ProductFacadeTest.php
grossmannmartin: renamed. 
Yes, right now the selling denied is not processed by the queue (but it will be in a later commit)

phpstan.neon
grossmannmartin: this commit was dropped entirely. The exclusion from phpstan check is already done in the file itself

packages/framework/src/Model/Product/Product.php
grossmannmartin: removed return type

packages/framework/src/Model/Product/ProductDomain.php
grossmannmartin: removed return type

packages/framework/src/Model/Category/CategoryParameter.php
grossmannmartin: thanks, removed return typehint

packages/framework/src/Model/Product/Filter/ProductFilterRepository.php
grossmannmartin: I have removed the `ProductFilterCountRepository` along with the relevant classes as they're not used anymore

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
grossmannmartin: I have added a new repository :) it's now more readable, thanks

packages/framework/src/Model/Product/ProductVariantFacade.php
grossmannmartin: implementation moved to framework

packages/framework/src/Model/Product/AffectedProductsRepository.php
grossmannmartin: Thanks! I have updated the code

packages/framework/src/Model/Product/Availability/ProductStoreAvailabilityInformation.php
grossmannmartin: fixed

project-base/app/tests/App/Functional/Model/Product/ProductFacadeTest.php
grossmannmartin: calculated selling denied is still present, so I have updated the tests to use the dispatched recalculation messages

project-base/app/src/Model/Transport/Transport.php
grossmannmartin: migration for this column added – 3fcfd921f9

packages/framework/src/Model/Product/ProductDomain.php
```diff
@@ -132,6 +132,12 @@ class ProductDomain
      */
     protected $orderingPriority;
 
+    /**
+     * @var bool
+     * @ORM\Column(type="boolean")
+     */
+    protected $saleExclusion;
```
grossmannmartin, line 139: migration for this column added in 323df4050f

packages/framework/src/Model/Product/Product.php
```diff
@@ -160,19 +160,6 @@ class Product extends AbstractTranslatableEntity
      */
     protected $outOfStockAvailability;
 
-    /**
-     * @var \Shopsys\FrameworkBundle\Model\Product\Availability\Availability
-     * @ORM\ManyToOne(targetEntity="Shopsys\FrameworkBundle\Model\Product\Availability\Availability")
-     * @ORM\JoinColumn(name="calculated_availability_id", referencedColumnName="id", nullable=false)
-     */
-    protected $calculatedAvailability;
```
grossmannmartin, line 168: conditionally removed column (it does not exist on previous CC project, but it does on FW based projects) – 99c6c09bbf

packages/framework/src/Model/Product/ProductDomain.php
grossmannmartin: db migration here a076885179

packages/framework/src/Model/Category/CategoryParameter.php
```diff
@@ -14,36 +14,36 @@
 class CategoryParameter
```
grossmannmartin, line 14: migration for this change – 24c5b524a3

packages/framework/build.xml
vitek-rostislav: yes, thank you!

packages/framework/src/Command/RecalculationsCommand.php
vitek-rostislav: I have created the task in Jira

project-base/app/tests/FrontendApiBundle/Functional/Cart/AnonymousAddOrderItemsToCartTest.php
vitek-rostislav: created

packages/framework/src/Model/Product/ProductFacade.php
```diff
@@ -141,18 +150,24 @@ public function edit($productId, ProductData $productData)
...
             $product->refreshVariants($productData->variants);
         }
+
         $this->refreshProductAccessories($product, $productData->accessories);
         $this->em->flush();
         $this->productSellingDeniedRecalculator->calculateSellingDeniedForProduct($product);
+
         $this->imageFacade->manageImages($product, $productData->images);
         $this->friendlyUrlFacade->saveUrlListFormData('front_product_detail', $product->getId(), $productData->urls);
-        $this->createFriendlyUrlsWhenRenamed($product, $originalNames);
```
vitek-rostislav, line 172: after this method call was removed, it is not used anymore, I am a little bit concerned about the removal but if we really want to get rid of this behavior, we should remove the unused method

packages/framework/src/Model/Product/ProductFacade.php
```diff
@@ -326,6 +341,26 @@ public function getByUuid(string $uuid): Product
         return $this->productRepository->getOneByUuid($uuid);
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\ProductData $productData
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     */
+    public function editProductStockRelation(ProductData $productData, Product $product): void
```
vitek-rostislav, line 344: this method is public just because of two tests. Just saying, you can close this comment :smile:

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
```diff
@@ -0,0 +1,77 @@
...
+    public function __construct(
+        protected readonly EntityManagerInterface $em,
+    ) {
+    }
+
+    /**
+     * @param int[] $productIds
+     * @return int[]
+     */
+    public function getIdsToRecalculate(array $productIds): array
```
vitek-rostislav, line 24: It would be nice to have a test for this method. Or at least an issue for creating the test :smile:

vitek-rostislav (approved):

packages/framework/src/Model/Product/ProductFacade.php
grossmannmartin: well, removed :)

packages/framework/src/Model/Product/ProductFacade.php
grossmannmartin: Although the change wouldn't be hard, I think it may be handy to be able to edit just stocks... (I'm thinking about some stock-only related import for instance).
If you feel it differently, I'm open to discussion.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
grossmannmartin: added test in ba1943735d

project-base/app/src/Command/CleanStorefrontQueryCacheCommand.php
```diff
@@ -16,7 +16,7 @@
 class CleanStorefrontQueryCacheCommand extends Command
```
vitek-rostislav, line 16: now we can move the command as well, right? I do not insist on that :smile:

vitek-rostislav (approved):

vitek-rostislav (approved):

pk16011990 (approved):

PR #2921, opened 2023-11-14
[SSP-1960] added logic for orgering GTM events

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We needed a way to manage the order of GTM events. This PR implements such mechanism 
|New feature| Yes
|Fixes issues| ... 

----
  - https://sh-ssp-1960-gtm-events-order.odin.shopsys.cloud
  - https://cz.sh-ssp-1960-gtm-events-order.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/gtm/context/GtmProvider.tsx
+++ b/project-base/storefront/gtm/context/GtmProvider.tsx
@@ -1,0 +1,33 @@
+import { useRouter } from 'next/router';
+import React, { useState, createContext, useEffect } from 'react';
+
+export type GtmContextType = {
+    didPageViewRun: boolean;
+    setDidPageViewRun: (newState: boolean) => void;
+};
+
+const defaultState: GtmContextType = {
+    didPageViewRun: false,
+    setDidPageViewRun: () => undefined,
+};
+
+export const GtmContext = createContext(defaultState);
+
+export const GtmProvider: FC = ({ children }) => {
+    const [didPageViewRun, setDidPageViewRun] = useState(defaultState.didPageViewRun);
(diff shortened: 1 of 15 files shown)


project-base/storefront/gtm/context/GtmProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+const defaultState: GtmContextType = {
+    didPageViewRun: false,
+    setDidPageViewRun: () => undefined,
+};
+
+export const GtmContext = createContext(defaultState);
+
+export const GtmProvider: FC = ({ children }) => {
```
tvikito, line 16: Why we are coming back to old times with Context? Why don't we use Zustand and hook for this feature as we do for other stuff? Also then we don't have worry about using context within `Provider`.

sebaholesz: The motivation here is based on several things:

1. it is store-agnostic, so it will work on projects not using zustand. It does not require you to use any package
2. it allows you to very easily add and remove it to just parts of your app (you could say the same is true for zustand, but zustand is global and if you only want it for order process, you cannot do that)
3. the point is to make GTM as isolated as possible, this is based on that goal
4. I do not see anything old-timish about context. Basically all packages doing similar thing (providing some data to a part of the tree) use context, as it is native react.

tvikito: As discussed between FE guys. We are leaving Context for this case.

tvikito (approved):

PR #2924, opened 2023-11-17
[framework] handle image resizing by image proxy

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The whole image resizing BE functionality is unnecessary, we can now leverage Next Image with imgProxy powers. As a result, the framework codebase is easier as it does not have to care about the image resizing at all. Also, less disk space will be needed as only the original image sizes have to be kept now. This PR goes hand in hand with https://github.com/shopsys/deployment/pull/5 and https://github.com/shopsys/deployment/pull/6
|New feature| Yes
|Fixes issues| ... 

----
  - https://rv-images.odin.shopsys.cloud
  - https://cz.rv-images.odin.shopsys.cloud


packages/framework/src/Component/Image/ImageFacade.php
```diff
@@ -449,7 +337,6 @@ public function copyImages(object $sourceEntity, object $targetEntity): void
                 $this->mountManager->copy(
                     'main://' . $this->imageLocator->getAbsoluteImageFilepath(
                         $sourceImage,
-                        ImageConfig::ORIGINAL_SIZE_NAME,
```
grossmannmartin, line 452: I know it's not your change, but I wanted to point it out anyway 😇 

In the catch block, the UnableToDeleteFile exception is caught on the copy operation. Is that OK?

packages/framework/src/Component/Image/ImageLocator.php
```diff
@@ -112,46 +72,27 @@ public function getAbsoluteAdditionalImageFilepath(Image $image, int $additional
...
         }
 
-        if ($sizeName === null) {
-            $pathParts[] = ImageConfig::DEFAULT_SIZE_NAME;
-        } else {
-            $pathParts[] = $sizeName;
-        }
+        $pathParts[] = ImageConfig::ORIGINAL_SIZE_NAME;
 
         return implode('/', $pathParts) . '/';
```
grossmannmartin: Just a thought. Wouldn't be nice to unify the paths to be returned without the slash at the end?
It's common to return the paths without the trailing slash in Symfony world (https://symfony.com/doc/current/components/filesystem.html#canonicalization) 

Currently, it's more or less random and it may cause confusion sometimes.

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -124,16 +122,10 @@ public function getImageHtml($imageOrEntity, array $attributes = [])
         try {
             $image = $this->imageFacade->getImageByObject($imageOrEntity, $attributes['type']);
             $entityName = $image->getEntityName();
-            $attributes['src'] = $this->getImageUrl($image, $attributes['size'], $attributes['type']);
+            $attributes['src'] = $this->getImageUrl($image, $attributes['size']);
```
grossmannmartin: it would be nice to remove the lazy attribute along with the already unsupported sinergi/browser-detector package.

Maybe at least the separate US?

packages/framework/src/Twig/ImageExtension.php
```diff
@@ -211,10 +202,9 @@ protected function preventDefault(array &$attributes): void
     /**
      * @param array $attributes
      * @param string $entityName
-     * @param \Shopsys\FrameworkBundle\Component\Image\AdditionalImageData[] $additionalImagesData
      * @return string
      */
-    protected function getImageHtmlByEntityName(array $attributes, $entityName, $additionalImagesData = []): string
+    protected function getImageHtmlByEntityName(array $attributes, string $entityName): string
```
grossmannmartin, line 169: I still see several usage of the `size` word in this file. Is it ok?

packages/frontend-api/src/Model/Resolver/Image/ImagesQuery.php
```diff
@@ -80,120 +71,50 @@ public function imagesByAdvertQuery(Advert $advert, ?string $type, ?string $size
...
-
-    /**
-     * @param \Shopsys\FrameworkBundle\Model\Advert\Advert $advert
-     * @param string|null $type
-     * @param string|null $size
-     * @return \Shopsys\FrameworkBundle\Component\Image\Config\ImageSizeConfig[]
-     */
-    protected function getSizeConfigsForAdvert(Advert $advert, ?string $type, ?string $size): array
-    {
-        $entityName = static::IMAGE_ENTITY_ADVERT;
```
grossmannmartin, line 188: the IMAGE_ENTITY_ADVERT constant is now unused

project-base/app/src/Component/Image/ImageFacade.php
```diff
@@ -178,117 +174,10 @@ public function getImageUrlFromAttributes(
...
-                $seoEntityName = $this->getSeoNameByImageAndLocale($image, $domainConfig->getLocale());
-
-                return $this->getFriendlyUrlSlug($seoEntityName);
-            },
-        );
-
-        return $this->cdnFacade->resolveDomainUrlForAssets($domainConfig)
-            . $this->imageUrlPrefix
-            . $this->imageLocator->getRelativeAdditionalImageFilepathWithSlug($image, $additionalSizeIndex, $sizeName, $friendlyUrlSeoEntityName);
+        return $this->getImageUrl($domainConfig, $image, $type);
```
grossmannmartin, line 180: isn't possible to unify the ImageFacade from the framework package and from the project-base into a single class?

project-base/storefront/graphql/requests/products/fragments/ProductDetailInterfaceFragment.graphql
```diff
@@ -14,7 +14,10 @@ fragment ProductDetailInterfaceFragment on Product {
     ean
     description
     images {
-        ...ImageSizesFragment
+        ...ImageFragment
+    }
+    productMainImage: mainImage {
+        ...ImageFragment
```
grossmannmartin: This duplicates data that are already obtained from the `images` field. 
As it results in another database query and several cache checks I would prefer to not combine images and main image in one query at all.

docker/nginx/nginx.conf
```diff
@@ -81,12 +81,18 @@ server {
             location ~ ^\/content\/images\/(?<entity_name>\w+)(?<image_type>\/\w+)?\/(?<image_size>(default|original|galleryThumbnail|modal|list|thumbnail|thumbnailSmall|thumbnailExtraSmall|thumbnailMedium|header|footer|productList|productListSecondRow|cartPreview|productListMiddle|productListMiddleRetina|listAside|listGrid|searchThumbnail|listBig)\/)(?<add_image_id>\d+--)?(?<image_name>[\w\-]+_(?<image_id>\d+))\.(?<image_extension>jpg|jpeg|png|gif) {
                 expires 1w;
```
grossmannmartin, line 84: can you please also reformat this file? some blocks are not indented properly

project-base/app/web/imageResizer.php
```diff
@@ -0,0 +1,84 @@
...
+// change autoloading source for monorepo
+if (file_exists(__DIR__ . '/../../../parameters_monorepo.yaml')) {
+    $projectRootDirectory = __DIR__ . '/../../..';
+}
+
+require $projectRootDirectory . '/vendor/symfony/dotenv/Dotenv.php';
+(new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
+
+$CDN_API_KEY = $_ENV['CDN_API_KEY'] ?? null;
+$CD_API_SALT = $_ENV['CDN_API_SALT'] ?? null;
```
grossmannmartin: this variable name has a typo

UPGRADE-14.0.md
```diff
@@ -211,6 +211,189 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
...
+        -   method `getImageSizeConfigByEntity` was removed
+        -   method `getImageSizeConfigByEntityName` was removed
+        -   method `assertImageSizeConfigByEntityNameExists` was replaced by `assertImageConfigByEntityNameExists`
+        -   method `getImageSizeConfigByImage` was removed
+        -   method `getImageEntityConfig` is now strictly typed
+        -   method `hasImageConfig` is now strictly typed
+        -   method `getEntityConfigByEntityName` is now strictly typed and its visibility changed to `protected`
+        -   method `getAllImageEntityConfigsByClass` is now strictly typed
+        -   method `getImageEntityConfigByClass` is now strictly typed
+-   `Shopsys\FrameworkBundle\Component\Image\Config\ImageConfigDefinition` class has been changed:
```
grossmannmartin: the indentation does not seem right in this upgrade note (and the following ones)

UPGRADE-14.0.md
```diff
@@ -211,6 +211,189 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
...
+        ```diff
+            public function getImageUrlFromAttributes(
+                DomainConfig $domainConfig,
+                int $id,
+                string $extension,
+                string $entityName,
+                ?string $type,
+        -       ?string $sizeName = null,
+        ```
+    -   method `getAllImageEntityConfigsByClass` has been removed
```
grossmannmartin: usually, the method names are written with parenthesis (this and other occurrences)

packages/framework/src/Command/MigrateImagesCommand.php
```diff
@@ -0,0 +1,236 @@
...
+     */
+    private function deleteDirectory(string $directory, SymfonyStyle $style): void
+    {
+        $isDirectoryEmpty = $this->filesystem->listContents($directory)->toArray() === [];
+        $additionalInfo = $isDirectoryEmpty ? 'is empty' : 'with all its content';
+
+        $style->writeln(sprintf('Removing directory %s (%s)', $directory, $additionalInfo));
+
+        if ($isDirectoryEmpty) {
+            $this->filesystem->delete($directory . '/');
```
grossmannmartin: deleting the empty directory on the local filesystem will result in an error.
Snímek obrazovky 2023-11-29 v 17 15 56

I understand that the migration script is meant primarily for the S3 storage, but as it's a part of the deploy phing target, I don't think we can ignore that

UPGRADE-14.0.md
```diff
@@ -211,6 +211,189 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 -   upgrade Storefront docker image ([#2931](https://github.com/shopsys/shopsys/pull/2931))
     -   now we use Node.js version 20-alpine3.17 and PNPM version 8.10.5
     -   see #project-base-diff to update your project
+-   handle image resizing by image proxy ([#2924](https://github.com/shopsys/shopsys/pull/2924))
+    -   for local development without CDN, the `imageResizer.php` file has a setup for local imgProxy instance
+    -   if you are using VSH CDN, the image proxy is already set up for you in `imageResizer.php` file.
+    -   for any other CDN, you can use the `imageResizer.php` file as well, but you have to set up the proxy yourself
```
grossmannmartin, line 262: shouldn't be mentioned also the nginx.conf file? When someone has more image sizes than the platform, then they need to add a size into a regex so the redirection work properly

docs/frontend/configuring-image-sizes-for-individual-devices-width.md
```diff
@@ -1,82 +0,0 @@
-# Configuring Image Sizes for Individual Devices Width
-
-In Shopsys Platform you can configure image sizes for individual devices width. This is allowed by HTML element [Picture element (see more in Mozilla official documentation)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture).
-
-## Introduction
```
grossmannmartin, line 5: this file have to be removed from the docs/frontend/navigation.yml, or the documentation will not build. (see failed container locally)

grossmannmartin (changes requested):
Really nice work. I'm happy about the cleaner code 😍

packages/framework/src/Component/Image/ImageFacade.php
vitek-rostislav: 249ca26cb - it was probably supposed to be `UnableToCopyFile`, however, after further examination, there are multiple exceptions that might be thrown, so I am catching a general exception now. We can discuss it if you want :slightly_smiling_face:

packages/framework/src/Component/Image/ImageLocator.php
vitek-rostislav: hm, that is a good point, however, here the path is used to determine a folder. It is then used in various places where the filename is appended. So it is never the final path. Lets discuss this please

packages/framework/src/Twig/ImageExtension.php
vitek-rostislav: I did not feel like creating user-story :smile: f5563d56d

packages/framework/src/Twig/ImageExtension.php
vitek-rostislav: ah, good catch, fixed in c99929089 and 12178f4df

packages/frontend-api/src/Model/Resolver/Image/ImagesQuery.php
vitek-rostislav: fixed in a52532494 and ec71d1d49

project-base/app/src/Component/Image/ImageFacade.php
vitek-rostislav: Yesterday, I created SSP-1971 for this :slightly_smiling_face:

docker/nginx/nginx.conf
vitek-rostislav: 6c2aa20db - I hope I have not broken it :smile:

project-base/app/web/imageResizer.php
vitek-rostislav: good catch, fixed in a3d8eacbf

UPGRADE-14.0.md
vitek-rostislav: I am not able to spot it, let's take a look at it together, please :smile:

docs/frontend/configuring-image-sizes-for-individual-devices-width.md
vitek-rostislav: ah, thanks! 41f16c887

UPGRADE-14.0.md
vitek-rostislav: yes, good point! 6e3ac2c02

UPGRADE-14.0.md
vitek-rostislav: ok :slightly_smiling_face: d0b2eeb65

project-base/storefront/graphql/requests/products/fragments/ProductDetailInterfaceFragment.graphql
vitek-rostislav: makes sense, fixed in 839c49afa

packages/framework/src/Command/MigrateImagesCommand.php
vitek-rostislav: I am not happy about that but here is a workaround for the problem - 4443ed998

grossmannmartin: nice

grossmannmartin (approved):

packages/framework/src/Component/Image/ImageLocator.php
vitek-rostislav: 36223a220

UPGRADE-14.0.md
vitek-rostislav: dc7ad61a6

project-base/storefront/components/Blocks/Banners/BannersSliderItem.tsx
```diff
@@ -16,9 +16,7 @@ export const BannersSliderItem: FC<BannersSliderItemProps> = ({
             {!image ? (
                 <BannerImage alt="no image" src="images/optimized-noimage.webp" />
             ) : (
-                <picture>
-                    <BannerImage alt={image.name || name} src={image.url} />
-                </picture>
+                <BannerImage alt={image.name || name} src={image.url} />
```
vitek-rostislav, line 20: Shouldn't we specify the width here? There were sizes (and additionalSizes) specified in the `images.yaml`

project-base/storefront/components/Blocks/OrderSummary/SingleProduct.tsx
```diff
@@ -14,11 +14,14 @@ export const SingleProduct: FC<SingleProductProps> = ({ item }) => {
 
     return (
         <li className="flex items-center border-b border-creamWhite py-3">
-            <Image
-                alt={item.product.mainImage?.name || item.product.fullName}
-                className="mr-4 h-14 w-14"
-                image={item.product.mainImage}
-            />
+            <div className="mr-4 flex h-14 w-14 items-center">
```
vitek-rostislav: Would it make sense to add `wrapperClass` to the Image component and then use it like this?

```
<div class={wrapperClass}>
    <Image
        alt=...
        ...
    />
</div>
```

so you do not have to wrap the Image component into a div everywhere?

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellersListItem.tsx
```diff
@@ -34,12 +34,7 @@ export const CategoryBestsellersListItem: FC<CategoryBestsellersListItemProps> =
                         type="product"
                         onClick={() => onGtmProductClickEventHandler(product, gtmProductListName, listIndex, url)}
                     >
-                        <Image
-                            alt={product.fullName}
-                            className="max-h-[80px] max-w-[80px]"
-                            dataTestId={TEST_IDENTIFIER + 'image'}
-                            image={product.mainImage}
-                        />
+                        <Image alt={product.fullName} height={80} src={product.mainImage?.url} width={80} />
```
vitek-rostislav: the alt should be like this:
```
alt={product.mainImage?.name || product.fullName}
```

UPGRADE-14.0.md
```diff
@@ -610,3 +610,8 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 
     -   keep in mind that the implementation heavily depends on the graphcache in `cacheExchange.ts`. This cache is responsible for manually updating queries, which decreases the number of requests, but also makes the entire functionality to behave as expected. If you use the generic mutations, queries, and fragments, you do not need to touch it, as it should work out of the box. However, there is a chance that your implementation will be more customized and not directly follow the provided one, for example by having a custom query for your list. In such case, if implementing any other list in a more custom manner, follow the guidelines in the docs (`graphcache.md`).
     -   we also renamed `productsCompare` to `comparedProducts` across the files as it is a more suitable and better name.
+
+-   implement NextImage component ([#2924](https://github.com/shopsys/shopsys/pull/2924))
```
vitek-rostislav, line 629: I would add two instructions here (feel free to rephrase it :slightly_smiling_face: )
```
-   check all the places where you use the `Image` component and modify your useages accordingly as the component interface has chagned
-   in your `next.config.js`, add all your domains names into `images -> remotePatterns` setting
```

vitek-rostislav (commented):
Hi, I have added just a few notes, check them out, please. Also, could you please update the docs? I have tried something in 0fb7bfcc825811ae83e1b651a5b72fcfea7ff34c, however, I believe it still needs your update regarding the component props. Plus the mention about the picture element can be probably removed now. And maybe the article deserves adding the similar information that you have added into the upgrade notes (regarding the tw approaches - width/height vs fill prop, thanks!

project-base/storefront/components/Blocks/Banners/BannersSliderItem.tsx
tvikito: Hmm, nope, because we done use `NextImage` here. We use simple `img` tag. But I have replaced this place and others where `img` has been used with out `Image` component. So now it's even better optimized.

project-base/storefront/components/Blocks/OrderSummary/SingleProduct.tsx
tvikito: Nope, because you don't need wrapper everywhere. Also it would mess up custom classnames. Because then you have to have different prop name for the second (like we had `wrapperClassname`) and for this prop we don't have intelisense for Tailwind classes.

project-base/storefront/components/Pages/CategoryDetail/CategoryBestsellers/CategoryBestsellersListItem.tsx
tvikito: Ah, nice catch. Replaced.

UPGRADE-14.0.md
tvikito: No problem, thanks for those additions. Just grammar changed. Added.

project-base/storefront/components/Blocks/Banners/BannersSliderItem.tsx
vitek-rostislav: ah, I did not notice it is not the Image component :+1: 

> But I have replaced this place and others where img has been used with out Image component. So now it's even better optimized. 

:ok_hand:

project-base/storefront/components/Blocks/OrderSummary/SingleProduct.tsx
vitek-rostislav: ok, makes sense :+1:

PR #2934, opened 2023-11-29
[framework] annotation fixer: get property type from typehint when the annotation is missing

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When defining properties using constructor property promotion, there are no doc blocks from which to take the property type. Hence, we need to try to get the type from typehint. In the future, this might be the preferred way (after adding typehints everywhere :slightly_smiling_face: )
|New feature| Yes 
|Fixes issues| ... 

----
  - https://rv-annotation-fixer.odin.shopsys.cloud
  - https://cz.rv-annotation-fixer.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/ClassExtension/TypehintHelper.php
+++ b/packages/framework/src/Component/ClassExtension/TypehintHelper.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\ClassExtension;
+
+use Roave\BetterReflection\Reflection\ReflectionNamedType;
+use Roave\BetterReflection\Reflection\ReflectionProperty;
+
+class TypehintHelper
+{
+    /**
+     * @param \Roave\BetterReflection\Reflection\ReflectionProperty $reflectionProperty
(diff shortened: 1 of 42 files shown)


packages/framework/src/Component/ClassExtension/TypehintHelper.php
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * @param \Roave\BetterReflection\Reflection\ReflectionProperty $reflectionProperty
+     * @return string
+     */
+    public static function getPropertyTypeFromTypehint(ReflectionProperty $reflectionProperty): string
+    {
+        $type = $reflectionProperty->getType();
+
+        if (($type instanceof ReflectionNamedType) === false) {
+            return '';
```
grossmannmartin: I prefer to return `null` in this method and then use `$propertyTypeString ?? ''` in PropertyAnnotationFactory. 

Right now the  `replaceInPropertyType` method seems kinda weird with mixing comparison of null and empty string

grossmannmartin (approved):

pk16011990 (approved):

PR #2936, opened 2023-11-30
[framework] order info is now asynchronously sent to Heureka after order is created

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This is a functionality that was implemented in Shopsys Framework but it was omitted in FE API. Storefront part (checkbox with user agreement) needs to be implemented again as well in a separate pull request. Originally, the Heureka request was synchronous. Now with the queues implemented in Shopsys Platform, we are able to process the order asynchronously. 
|New feature| Yes 
|Fixes issues| ... 

----
  - https://rv-heureka.odin.shopsys.cloud
  - https://cz.rv-heureka.odin.shopsys.cloud


packages/framework/src/Model/Order/Messenger/HeurekaPlacedOrderMessageHandler.php
```diff
@@ -0,0 +1,44 @@
...
+            if ($orderSent === true) {
+                $this->logger->info('Order successfully sent to Heureka', [
+                    'orderId' => $placedOrderMessage->orderId,
+                ]);
+            }
+        } catch (Exception $exception) {
+            $this->logger->error('Sending order to Heureka failed', [
+                'orderId' => $placedOrderMessage->orderId,
+                'exception' => $exception,
+            ]);
```
TomasLudvik, line 42: There could be some network error or something, should we not try a few times to send it again after some time? Like half an hour or by cron?

packages/framework/src/Model/Order/Messenger/HeurekaPlacedOrderMessageHandler.php
```diff
@@ -0,0 +1,44 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Order\Messenger;
+
+use Exception;
+use Psr\Log\LoggerInterface;
+use Shopsys\FrameworkBundle\Model\Order\OrderFacade;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+class HeurekaPlacedOrderMessageHandler implements MessageHandlerInterface
```
TomasLudvik: Should this class be called Heureka? Other parts do not mention it and this could be universal for other actions like sending e-mail after order creation.

TomasLudvik: I have two things that we should discuss:
- would not be better to have handlers configured manually as described in https://symfony.com/doc/5.4/messenger.html#manually-configuring-handlers? I know most of our colleagues do not like "magic" and this is one of those cases.
- if not, should not we follow the recommended way to define Handler in PHP 8 as described here https://symfony.com/doc/5.4/messenger.html#creating-a-message-handler using The #[AsMessageHandler] PHP attribute

TomasLudvik (changes requested):

packages/framework/src/Model/Order/Messenger/HeurekaPlacedOrderMessageHandler.php
vitek-rostislav: The class is called Heureka on purpose - I have a general message that is dispatched after an order is created. Then, there could be any number of handlers that process the message. If I wanted to send mail using the queue, I would just create another specific message handler (e.g. `EmailPlaceOrderMessageHandler`) that would take care of the emails. Basically, I do not want to process all the asynchronous tasks after order creation in one generic handler.

I just want to mention that I have already tested the behaviour with multiple handlers to be sure it works this way (when you think about it, the approach is similar to events and event handlers/subscribers)

vitek-rostislav: ad your suggestions - I do not like the manual configuration of something that works without it but we can discuss it further in the team if you feel it differently.

Regarding the attribute, yes, thank you for pointing it out, I did not know about it as I learned the old way on the project we used the messenger. So I changed it in 7e4c27094

packages/framework/src/Model/Order/Messenger/HeurekaPlacedOrderMessageHandler.php
vitek-rostislav: With the messenger, we have now great powers :blush: 254428bbd

To explain that, there is an inbuilt mechanism - when an exception is thrown during the message processing, the message is re-enqueued three times by default with 1, 2, and 4-second delay (I have configured longer delays in my solution, see the [docs](https://symfony.com/doc/current/messenger.html#retries-failures)).

What is really great and works out of the box is when there are multiple handlers for one message and one of them fails, the message is still re-enqueued but it is then processed just by the handlers that failed (e.g. when sending to heureka fails but email is sent successfully, it will not try to send the email again). This is thanks to the [stamps system](https://symfony.com/doc/current/components/messenger.html#adding-metadata-to-messages-envelopes), `HandledStamp` here in particular.

Don't you just love it? :heart_eyes:

project-base/app/config/packages/messenger.yaml
```diff
@@ -15,6 +15,19 @@ framework:
...
+                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
+                options:
+                    vhost: '%env(MESSENGER_TRANSPORT_VHOST)%'
+                    exchange:
+                        name: placed_order_exchange
+                        type: direct
+                    queues:
+                        placed_orders: ~
+                retry_strategy:
+                    delay: 60000
```
TomasLudvik, line 29: I would add at least 5 - 6 max retries to ensure a day of trying

pk16011990 (approved):

TomasLudvik (approved):

PR #2952, opened 2023-12-18
[frontend-api] added ability to change a payment in an order

### LINKED ISSUE #2381: Frontend API - TransportResolver and TransportsResolver return even hidden payments related to transport
## Describe the bug
```
query transports {
    transports {
        uuid
        name
        payments {
            uuid
            name
        }
    }
}
```
This query will return payments not visible on current domain, because resolver uses relation directly from entity which does not take payments visibility into account.

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When goPay payment fails after order is created, the user should be able to change the payment to a different type
|New feature| Yes 
|Fixes issues| closes #2381

----
  - https://rv-change-order-payment.odin.shopsys.cloud
  - https://cz.rv-change-order-payment.odin.shopsys.cloud


project-base/app/tests/FrontendApiBundle/Functional/Payment/OrderPaymentsTest.php
```diff
@@ -0,0 +1,155 @@
...
+            ],
+        ];
+
+        yield 'order on first domain with Czech post transport' => [
+            'orderReferenceName' => OrderDataFixture::ORDER_PREFIX . 3,
+            'expectedPaymentReferenceNames' => [
+                PaymentDataFixture::PAYMENT_CASH_ON_DELIVERY,
+                PaymentDataFixture::PAYMENT_GOPAY_BANK_ACCOUNT,
+            ],
+        ];
```
grossmannmartin: cannot be used `yield from $this->getOrderPaymentsSingledomainDataProvider();` to prevent dry for first two scenarios?

packages/framework/assets/js/admin/components/Payment.js
```diff
@@ -12,7 +10,7 @@ import constant from '../utils/constant';
             const selectedType = $paymentType.val();
             const $goPayPaymentMethodFormLine = $('.js-payment-gopay-payment-method').closest('.form-line');
 
-            if (selectedType === constant('\\App\\Model\\Payment\\Payment::TYPE_GOPAY')) {
+            if (selectedType === 'goPay') {
```
grossmannmartin, line 13: 🤔  wouldn't be better to use constant?

packages/framework/src/Model/GoPay/Exception/GoPayPaymentDownloadException.php
```diff
@@ -0,0 +1,45 @@
...
+//                . "Request data:\n%s\n"
+//                . "Response data:\n%s\n",
+//                $actualCode,
+//                $expectedCode,
+//                $url,
+//                $method,
+//                print_r($requestData, true),
+//                print_r($responseData, true),
+//            ),
+//        );
```
grossmannmartin: do we want to keep the commented code here?

packages/framework/src/Model/GoPay/GoPayClient.php
```diff
@@ -14,20 +12,22 @@
 use GoPay\OAuth2;
 use GoPay\Token\CachedOAuth;
 use GoPay\Token\InMemoryTokenCache;
```
grossmannmartin, line 14: I believe with this change we also need to update the composer.json files (the library is used in the framework, but not in project-base)

packages/framework/src/Model/GoPay/GoPayOrderMapper.php
```diff
@@ -2,29 +2,29 @@
 
 declare(strict_types=1);
 
-namespace App\Model\GoPay;
+namespace Shopsys\FrameworkBundle\Model\GoPay;
 
-use App\Model\GoPay\PaymentMethod\GoPayPaymentMethodFacade;
-use App\Model\Order\Order;
 use League\ISO3166\ISO3166;
```
grossmannmartin, line 7: this dependency should be moved to framework package (in composer.json).
I would recommend splitting the monorepo and verify individual packages

packages/framework/src/Model/GoPay/PaymentMethod/GoPayPaymentMethodRepository.php
```diff
@@ -2,32 +2,32 @@
...
 use Doctrine\ORM\Query\Expr\Join;
+use Shopsys\FrameworkBundle\Model\GoPay\BankSwift\GoPayBankSwift;
 
 class GoPayPaymentMethodRepository
 {
     /**
      * @param \Doctrine\ORM\EntityManagerInterface $em
      */
-    public function __construct(private EntityManagerInterface $em)
+    public function __construct(protected EntityManagerInterface $em)
```
grossmannmartin: nitpick: this may be readonly

packages/framework/src/Model/Order/OrderRepository.php
```diff
@@ -293,4 +297,24 @@ protected function getOrderListQueryBuilder()
             ->join('o.currency', 'c')
             ->leftJoin('o.customerUser', 'cu');
     }
+
+    /**
+     * @param \DateTime $fromDate
+     * @return \Shopsys\FrameworkBundle\Model\Order\Order[]
+     */
+    public function getAllUnpaidGoPayOrders(DateTime $fromDate): array
```
grossmannmartin: wouldn't this method feel better in GopayRepository and GopayFacade?

packages/framework/src/Resources/config/services.yaml
```diff
@@ -1039,3 +1039,7 @@ services:
     Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationMessageHandler:
         arguments:
             $logger: '@monolog.logger.queue'
+
+    Shopsys\FrameworkBundle\Model\GoPay\GoPayClientFactory:
+        arguments:
+            - '%gopay_config%'
```
grossmannmartin, line 1068: default value for this parameter should be set in framework's config file, or?

project-base/app/config/services_test.yaml
```diff
@@ -37,7 +37,7 @@ services:
         arguments:
             - '%gopay_config%'
 
-    App\Model\GoPay\GoPayClientFactory:
+    Shopsys\FrameworkBundle\Model\GoPay\GoPayClientFactory:
```
grossmannmartin, line 40: isn't possible to have this config in the framework package?

project-base/app/src/DataFixtures/Demo/GoPayDataFixture.php
```diff
@@ -30,8 +30,8 @@ class GoPayDataFixture extends AbstractReferenceFixture implements DependentFixt
         [
             'name' => 'Aqua bank',
             'swift' => 'ABC123456',
-            'image_normal_url' => 'airbank image Url',
-            'image_large_url' => 'airbank large image Url',
+            'image_normal_url' => 'Aqua bank image Url',
+            'image_large_url' => 'Aqua bank large image Url',
```
grossmannmartin: Did we get a better offer for the promotion? 😆 

edit: oh, now I see the inconsistency :)

packages/frontend-api/src/Model/Mutation/Payment/Exception/MaxTransactionCountReachedUserError.php
```diff
@@ -2,14 +2,14 @@
...
-namespace App\FrontendApi\Mutation\Payment\Exception;
+namespace Shopsys\FrontendApiBundle\Model\Mutation\Payment\Exception;
 
 use Overblog\GraphQLBundle\Error\UserError;
 use Shopsys\FrontendApiBundle\Model\Error\UserErrorWithCodeInterface;
 
 class MaxTransactionCountReachedUserError extends UserError implements UserErrorWithCodeInterface
 {
-    private const CODE = 'max-transaction-count-reached';
+    public const CODE = 'max-transaction-count-reached';
```
grossmannmartin: I'm not sure about using the constant in the tests. 
Changing the value is not a safe operation – Storefront takes into account the value. So I think it is fair to let the test fail when the value is changed

packages/framework/src/Model/GoPay/OrderGoPayStatusUpdateCronModule.php
```diff
@@ -38,16 +38,20 @@ public function run(): void
         $twentyOneDaysAgo = $now->sub(DateInterval::createFromDateString('21 days'));
         $orders = $this->orderFacade->getAllUnpaidGoPayOrders($twentyOneDaysAgo);
 
-        $this->logger->info('Downloading status updates for `' . count($orders) . '` orders.');
+        $this->logger->info('Downloading status updates for orders.', [
+            'ordersCount' => count($orders),
+        ]);
```
grossmannmartin, line 42: 💚

packages/frontend-api/src/Resources/translations/validators.cs.po
```diff
@@ -37,6 +37,18 @@ msgstr "Staré heslo musí mít minimálně {{ limit }} znaků"
...
+
+msgid "Payment {{ paymentUuid }} cannot be used with swift {{ swift }}"
+msgstr ""
+
+msgid "Payment {{ paymentUuid }} cannot be used with transport {{ transportUuid }}"
+msgstr ""
+
+msgid "Payment {{ paymentUuid }} is not available"
+msgstr ""
+
```
grossmannmartin: shouldn't it be translated?

project-base/app/src/DataFixtures/Demo/GoPayDataFixture.php
```diff
@@ -22,152 +24,81 @@ class GoPayDataFixture extends AbstractReferenceFixture implements DependentFixt
...
             'image_normal_url' => 'airbank image Url',
             'image_large_url' => 'airbank large image Url',
             'is_online' => true,
         ],
         [
-            'name' => 'Aqua bank',
-            'swift' => 'ABC123456',
-            'image_normal_url' => 'Aqua bank image Url',
-            'image_large_url' => 'Aqua bank large image Url',
+            'name' => 'FIO banka',
```
grossmannmartin: shouldn't it be in english?

project-base/app/src/DataFixtures/Demo/GoPayDataFixture.php
```diff
@@ -22,152 +24,81 @@ class GoPayDataFixture extends AbstractReferenceFixture implements DependentFixt
...
             'image_normal_url' => 'payment_card',
             'image_large_url' => 'payment_card@2x',
             'payment_group' => 'card-payment',
         ],
         [
             'reference_name' => self::BANK_ACCOUNT_METHOD,
-            'currency' => CurrencyDataFixture::CURRENCY_CZK,
             'identifier' => GoPayPaymentMethod::IDENTIFIER_BANK_TRANSFER,
-            'name' => '[CS] Rychlý bankovní převod',
+            'name_pattern' => '[%s] Rychlý bankovní převod',
```
grossmannmartin: shouldn't it be in english by default and translated when necessary?

packages/framework/src/Component/Domain/Config/DomainConfig.php
```diff
@@ -62,6 +62,14 @@ public function getLocale()
         return $this->locale;
     }
 
+    /**
+     * @return string
+     */
+    public function getLocaleUppercase(): string
```
grossmannmartin: I would rather use everywhere the locale as is set. 
This can be confusing while creating a new feature (should I use `getLocale()` or `getLocaleUppercase()`?)

grossmannmartin (changes requested):

packages/framework/assets/js/admin/components/Payment.js
vitek-rostislav: nah, the constant function does not work unless you define the constant in `constant.js` file - one way or another, you must duplicate the constant value from the PHP code. From my point of view, using the string directly in JS is less confusing as with the `constant` function, one might think the constant value is taken from the PHP class which is misleading and caused me headaches in the past :smile: (I will create a user story for removing the constant function completely if you do not stop me)

packages/framework/src/Model/GoPay/Exception/GoPayPaymentDownloadException.php
vitek-rostislav: we certainly do not :slightly_smiling_face: thanks, removed in 1db50a4b40

project-base/app/tests/FrontendApiBundle/Functional/Payment/OrderPaymentsTest.php
vitek-rostislav: yes! Nice :ok_hand: dc8509eff1

packages/framework/src/Model/GoPay/GoPayClient.php
vitek-rostislav: makes sense :+1: 673326342a

packages/framework/src/Model/GoPay/GoPayOrderMapper.php
vitek-rostislav: makes sense :+1: 673326342a

packages/framework/src/Model/GoPay/PaymentMethod/GoPayPaymentMethodRepository.php
vitek-rostislav: :+1: 295dde153

packages/framework/src/Model/Order/OrderRepository.php
vitek-rostislav: can be :slightly_smiling_face: 7045e184ea

project-base/app/config/services_test.yaml
vitek-rostislav: Hmmm, I do not think so - `Tests\FrontendApiBundle\Functional\Payment\GoPay\GoPayClientFactory` is defined in the project-base - the functional tests are in project-base, so I am not sure here, we can discuss it

packages/framework/src/Resources/config/services.yaml
vitek-rostislav: I do not know - it seems like we always define the configs in the project-base, but, indeed, I do not feel ok about it, let's discuss it, please

packages/frontend-api/src/Model/Mutation/Payment/Exception/MaxTransactionCountReachedUserError.php
vitek-rostislav: Hmm, I have been doing it like this for a while now :smile: You have a valid point, however - if the test fails after they change the code, will they know they should fix it in the storefront file, or will they just adjust the value in the test? :thinking:

packages/frontend-api/src/Resources/translations/validators.cs.po
vitek-rostislav: I think it was committed by accident, the translations were then added later

project-base/app/src/DataFixtures/Demo/GoPayDataFixture.php
vitek-rostislav: indeed :slightly_smiling_face: eec09fd49e

project-base/app/src/DataFixtures/Demo/GoPayDataFixture.php
vitek-rostislav: indeed :slightly_smiling_face: eec09fd49e
I do not want to complicate it with the translator (if you are ok with that) - it is basically used just for the tests

packages/framework/src/Component/Domain/Config/DomainConfig.php
vitek-rostislav: yeah, I was also not sure about it, I wanted to keep the swift pattern with uppercase, as well as the original payment names in data fixtures, but the truth is, it might bring more troubles than benefits, removed in c60eb0882a

packages/framework/assets/js/admin/components/Payment.js
grossmannmartin: that makes sense. I didn't realize that, so let's create an US :)

project-base/app/config/services_test.yaml
grossmannmartin: oh, you're right

grossmannmartin (approved):

packages/framework/src/Resources/config/services.yaml
vitek-rostislav: resolved by f1fea3171b

packages/frontend-api/src/Model/Mutation/Payment/Exception/MaxTransactionCountReachedUserError.php
vitek-rostislav: changed in 9de72b2fdc

project-base/storefront/hooks/order/useChangePaymentInOrder.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (editedOrder) {
+            showSuccessMessage(t('Your payment has been successfully changed'));
+
+            if (isUserLoggedIn) {
+                router.push({
+                    pathname: customerOrderDetailUrl,
+                    query: { orderNumber: editedOrder.number },
+                });
+            } else {
+                router.push(`${orderByHashUrl}/${editedOrder.urlHash}`);
```
sebaholesz: Isn't this URL relative? Won't that cause issues?

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+type PaymentsInOrderSelectProps = {
+    orderUuid: string;
+    orderPaymentType: string | undefined;
+    canPaymentBeRepeated: boolean;
+};
+
+export const PaymentsInOrderSelect: FC<PaymentsInOrderSelectProps> = ({
+    orderUuid,
+    canPaymentBeRepeated,
+    orderPaymentType,
```
sebaholesz: Reading this code, it was unclear to me what `orderPaymentType` actually is (even though it was me who named it like that 😆). Could you rename it to something like `lastUsedOrderPaymentType`?

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+                    </TransportAndPaymentListItem>
+                    <GoPayGateway
+                        requiresAction
+                        className="mt-5"
+                        initialButtonText={t('Repeat payment')}
+                        isDisabled={selectedPaymentUuid !== currentOrderPayment.uuid}
+                        orderUuid={orderUuid}
+                    />
+                </div>
+            )}
```
sebaholesz: I am trying to understand this code. You use `currentOrderPayment` here. Is it sure that this is going to be go pay? Is it because if `orderPaymentType` is go pay, and we have `currentOrderPayment`, we know it is going to be go pay? I am not sure if I quite follow.

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    return (
+        <div className="mt-6 flex flex-col items-center gap-6">
+            {orderPaymentType === PaymentTypeEnum.GoPay && canPaymentBeRepeated && currentOrderPayment && (
+                <div className="flex w-full flex-col items-center">
+                    <TransportAndPaymentListItem
+                        key={currentOrderPayment.uuid}
+                        className="order-none"
+                        isActive={selectedPaymentUuid === currentOrderPayment.uuid}
```
sebaholesz: The naming of "selected" and "current" makes it seems like it is the same thing and might be somewhat confusing. Could you consider different naming?

sebaholesz (changes requested):

project-base/storefront/hooks/order/useChangePaymentInOrder.ts
```diff
@@ -0,0 +1,42 @@
+import { useChangePaymentInOrderMutationApi } from 'graphql/generated';
```
tvikito: Since this hook is used only in one place what about move it to the attached component folder to helpers file?

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+    canPaymentBeRepeated: boolean;
+};
+
+export const PaymentsInOrderSelect: FC<PaymentsInOrderSelectProps> = ({
+    orderUuid,
+    canPaymentBeRepeated,
+    orderPaymentType,
+}) => {
+    const { t } = useTranslation();
+    const { changePaymentInOrderFetching, changePaymentInOrderHandler } = useChangePaymentInOrder();
```
tvikito: We are trying to follow pattern of marking boolean variables starting with `is` like `isChangePaymentInOrderFetching`. Let's use it here as well.

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const [selectedPaymentUuid, setSelectedPaymentUuid] = useState<string | undefined>();
+
+    const [{ data: orderAvailablePaymentsData, fetching: orderAvailablePaymentsFetching }] =
```
tvikito:
```suggestion
    const [{ data: orderAvailablePaymentsData, fetching: isOrderAvailablePaymentsFetching }] =
```

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+        return null;
+    }
+
+    return (
+        <div className="mt-6 flex flex-col items-center gap-6">
+            {orderPaymentType === PaymentTypeEnum.GoPay && canPaymentBeRepeated && currentOrderPayment && (
+                <div className="flex w-full flex-col items-center">
+                    <TransportAndPaymentListItem
+                        key={currentOrderPayment.uuid}
+                        className="order-none"
```
tvikito: This `order-none` is here for nothing or? 🤔

project-base/storefront/hooks/order/useChangePaymentInOrder.ts
VaniaToper: Good question. I did a demo where I tested router.push on the /stores page
```
useEffect(() => {
    router.push('/order-url/test');
}, []);
```
And the resulting url was /order-url/test so I guess my solution ok :)

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
VaniaToper: I realized that I don't even need these props here, but I renamed it in the parent component

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
VaniaToper: Good point, I added a condition where I check if currentOrderPayment is gopay

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
VaniaToper: Maybe there was some overlap with the selectedPayment (in order or for change?). So I'm going to expand the naming to selectedPaymentUuidForChange, is that ok?

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
VaniaToper: Ok, I extended from the urql naming where it's just fetching, but your solution is good too

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
VaniaToper: I need to override the order-1 className which is used in TransportAndPaymentListItem

project-base/storefront/urql/cache/updates.ts
```diff
@@ -93,6 +93,9 @@ export const cacheUpdates: UpdatesConfig = {
         RemoveProductList(_result, args: RemoveProductListMutationVariablesApi, cache) {
             manuallyRemoveProductListQuery(cache, args.input);
         },
+        ChangePaymentInOrder(_result, _args, cache) {
+            invalidateFields(cache, ['order']);
+        },
```
sebaholesz, line 122: It is possible to do it in a more optimal way by taking the new payment and updating the order query with it. You save a query. Not sure how helpful that is in this case. I will leave it for you to decide.

project-base/storefront/hooks/order/useChangePaymentInOrder.ts
```diff
@@ -0,0 +1,48 @@
...
```
sebaholesz: Seems like you now have this file duplicated

project-base/storefront/components/PaymentsInOrderSelect/helper.ts
```diff
@@ -0,0 +1,48 @@
...
+
+        if (!editedOrder) {
+            showErrorMessage(t('An error occurred while changing the payment'));
+
+            return;
+        }
+
+        showSuccessMessage(t('Your payment has been successfully changed'));
+
+        if (withRedirectAfterChanging) {
```
sebaholesz: You can do an early return here, right?

sebaholesz (changes requested):

project-base/storefront/components/PaymentsInOrderSelect/helper.ts
VaniaToper: Sure, good point

project-base/storefront/hooks/order/useChangePaymentInOrder.ts
VaniaToper: My bad, didn't notice, thanks

project-base/storefront/urql/cache/updates.ts
VaniaToper: Hmm, you mean something like manuallyRemoveProductListQuery, but with an order?. Maybe this will be a cleaner solution, but I think it's a bit overkill because this is only for the order detail with gopay and the optimization will be some small numbers.
If you have other points about it, I am open to discussing 🙂

project-base/storefront/hooks/order/useChangePaymentInOrder.ts
tvikito: We have all helpers named as `helpers`, let's replace the name `helper`.

sebaholesz (approved):

UPGRADE-14.0.md
```diff
@@ -838,6 +838,7 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
         +       ?ArrayObject $context = null,
         +   ): Price
         ```
+    -   FE API: `PaymentStatusDecorator` type has been removed, `Order` type is returned from `UpdatePaymentStatus` instead
```
grossmannmartin: payment status decorator was introduced in this PR, so it's not necessary to mention it in the upgrade notes

project-base/storefront/components/Blocks/OrderStatus/OrderStatus.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const OrderStatus: FC<OrderStatusProps> = ({ order }) => {
+    const { t } = useTranslation();
+
+    return (
+        <>
+            {order.status}
+            {order.payment.type === PaymentTypeEnum.GoPay && (
+                <>
+                    {' '}
```
sebaholesz, line 17: Not sure if we want this spacing?

project-base/storefront/components/Pages/Customer/OrderDetailContent.tsx
```diff
@@ -95,6 +98,24 @@ export const OrderDetailContent: FC<OrderDetailContentProps> = ({ order }) => {
                                 </Table>
                             </Cell>
                         )}
+                        <Cell className="flex-1">
+                            <ColumnHeader>{t('Payment information')}</ColumnHeader>
+                            <Table className="border-0 p-0">
+                                <Row>
+                                    <CellMinor>{t('Status')}</CellMinor>
+                                    <Cell align="right" dataTestId={TEST_IDENTIFIER + 'status'}>
```
sebaholesz: If we do not need the identifier for an existing test, lets not add it.

project-base/storefront/hooks/cart/useIsPaymentWithCartAvailable.ts
```diff
@@ -0,0 +1,9 @@
+import { useSettingsQueryApi } from 'graphql/generated';
+
+export const useIsPaymentWithCartAvailable = (paymentTransactionCount: number) => {
```
sebaholesz: `canOrderBePaid` or `isPaymentAvailableForOrder` makes more sense to me.

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const [isGoPayVisible, setIsGoPayVisible] = useState(false);
+
+    const [{ data: orderAvailablePaymentsData, fetching: isOrderAvailablePaymentsFetching }] =
+        useOrderAvailablePaymentsQueryApi({
+            variables: { orderUuid },
+        });
+
+    const isPaymentWithCardAvailable = useIsPaymentWithCartAvailable(paymentTransactionCount);
+    const currentOrderPayment = orderAvailablePaymentsData?.orderPayments.currentPayment;
+    const isSelectedPaymentEqualsOrderPayment = selectedPaymentForChange?.uuid === currentOrderPayment?.uuid;
```
sebaholesz: Either `isSelectedPaymentEqualToOrderPayment` or `doesSelectedPaymentEqualOrderPayment`, to make it grammatically correct

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const filteredAvailablePayments = orderAvailablePaymentsData?.orderPayments.availablePayments.filter((payment) => {
+        return payment.type !== PaymentTypeEnum.GoPay || isPaymentWithCardAvailable;
+    });
```
sebaholesz: const filteredAvailablePayments = orderAvailablePaymentsData?.orderPayments.availablePayments.filter((payment) => payment.type !== PaymentTypeEnum.GoPay || isPaymentWithCardAvailable);

project-base/storefront/components/PaymentsInOrderSelect/PaymentsInOrderSelect.tsx
```diff
@@ -0,0 +1,125 @@
...
+            if (
+                selectedPaymentForChange.type === PaymentTypeEnum.GoPay &&
+                changePaymentInOrderData?.ChangePaymentInOrder
+            ) {
+                setIsGoPayVisible(true);
+            }
+        }
+    };
+
+    if (isOrderAvailablePaymentsFetching) {
```
sebaholesz: maybe rather `areOrderAvailablePaymentsFetching`

sebaholesz (changes requested):

project-base/storefront/components/Blocks/OrderStatus/OrderStatus.tsx
VaniaToper: Yes we want it :)

project-base/storefront/hooks/cart/useIsPaymentWithCartAvailable.ts
VaniaToper: But this condition only checks whether the order can be paid by card. if I rename it to canOrderBePaid it won't be relevant to what this hook has inside

sebaholesz: Then the name does not make sense even more. `useIsPaymentWithCartAvailable` sounds like a hook which tells me if a payment is available for a given CART. The name does not say anything about a CARD.

VaniaToper: Oh, I see, there's a typo, carT instead of carD, thank you

sebaholesz (approved):

packages/framework/src/Model/Payment/Service/PaymentServiceFacade.php
```diff
@@ -2,55 +2,59 @@
...
      * @param \Psr\Log\LoggerInterface $logger
      * @param \Symfony\Component\DependencyInjection\ContainerInterface $container
+     * @param \Shopsys\FrameworkBundle\Model\Payment\PaymentSetupCreationDataFactory $paymentSetupCreationDataFactory
      */
     public function __construct(
-        private readonly PaymentTransactionFacade $paymentTransactionFacade,
-        private readonly PaymentTransactionDataFactory $paymentTransactionDataFactory,
+        protected readonly PaymentTransactionFacade $paymentTransactionFacade,
+        protected readonly PaymentTransactionDataFactory $paymentTransactionDataFactory,
         GoPayFacade $goPayFacade,
```
pk16011990, line 44: Why isn't it marked as 'protected'?

pk16011990 (approved):

packages/framework/src/Model/Payment/Service/PaymentServiceFacade.php
vitek-rostislav: because it is not a property - I am not able to use constructor property promotion here. The `GoPayFacade` is wired via the constructor and stored in `$paymentServices` array (instances of PaymentServiceInterface). If I get it correctly, the `PaymentServiceFacade` class is ready to handle more payment services thanks to that (If we had e.g. Sofort, we would just add `SofortFacade` here

PR #2953, opened 2023-12-19
[project-base] messages are now consumed on gitlab ci review

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Application ready during the review stage has to has the message system up and running, so we're able to test the app properly.
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-gitlab-review-rabbit.odin.shopsys.cloud
  - https://cz.mg-gitlab-review-rabbit.odin.shopsys.cloud

### DIFF
--- a/.github/set-urls-for-review.sh
+++ b/.github/set-urls-for-review.sh
@@ -6,6 +6,7 @@ setUrlsToDomainsUrls () {
 
     for DOMAIN in ${DOMAINS//,/ } ; do
         docker compose exec -T php-fpm sed -i "s/http:\/\/127.0.0.${ITERATOR}:8000/https:\/\/${DOMAIN}/g" project-base/app/config/domains_urls.yaml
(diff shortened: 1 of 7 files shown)


vitek-rostislav (commented):
Hi, just a few things to consider:
- it would be nice to have the rabbit admin available on the review server, maybe it would be therefore good idea to have the rabbit service installed globally and use VHOSTS
- we should probably set proper domains for the consumer in `gitlab/scripts/set-urls-for-review.sh`
- I am not sure but we might need to generate API keys for the consumer container in `.gitlab-ci.yml` for the `review` stage (for the php-fpm container, we do so using `build-review-ci` phing target

vitek-rostislav:
Hello, thank you for addressing the notes :ok_hand: 
One more thing - the `remove` vhost function is not used anywhere but it probably should be used in `review:stop`, right?

grossmannmartin:
> Hello, thank you for addressing the notes 👌 One more thing - the `remove` vhost function is not used anywhere but it probably should be used in `review:stop`, right?

Yes, yes it should. On GitLab I have added it to review:stop stage. On GitHub it's already in cancel-review.sh script

vitek-rostislav (approved):

PR #2959, opened 2023-12-28
[shopsys] fixed feed generation during request

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Internal Symfony's service resetter cannot be used during the request as it clears additional classes necessary for a properly working app (profiler, stopwatch, session, ...). This PR introduces a custom service resetter that clears only classes implementing ResetInterface from Shopsys\ and App\ namespace. 
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-fix-feed-generation.odin.shopsys.cloud
  - https://cz.mg-fix-feed-generation.odin.shopsys.cloud


packages/framework/src/DependencyInjection/ServicesResetter.php
```diff
@@ -0,0 +1,41 @@
...
+use Symfony\Contracts\Service\ResetInterface;
+use Traversable;
+
+class ServicesResetter
+{
+    /**
+     * @param \Traversable<string, \Symfony\Contracts\Service\ResetInterface> $resettableServices
+     */
+    public function __construct(
+        protected readonly Traversable $resettableServices,
```
TomasLudvik, line 16:
```suggestion
        protected readonly Traversable $resetableServices,
```

packages/framework/src/Resources/config/services.yaml
```diff
@@ -16,7 +16,7 @@ services:
         public: false
 
     Shopsys\FrameworkBundle\:
-        resource: '../../**/*{Calculation,Collector,Dispatcher,Facade,Factory,Generator,Handler,InlineEdit,Listener,Loader,Mapper,Middleware,Parser,Provider,Recalculator,Registry,Repository,Resolver,Service,Scheduler,Subscriber,Transformer}.php'
```
TomasLudvik: What is the benefit of this change?

TomasLudvik (approved):

packages/framework/src/DependencyInjection/ServicesResetter.php
grossmannmartin: resettable seems right 
https://en.wiktionary.org/wiki/resettable
https://www.collinsdictionary.com/dictionary/english/resettable

(same name is also used in Symfony class)

packages/framework/src/Resources/config/services.yaml
grossmannmartin: When the Collector classes are autoconfigured, the debug-only services are created in the production env. 

That caused a problem because the ElasticsearchCollector service was passed as one of the dependencies to the service resetter. This class has a dependency on the service `ElasticsearchRequestCollection` and this service is configured only in a dev mode.

Long story short, most collector classes currently present in the shopsys/framework package are debug-only and should not be configured for production env.

TomasLudvik (approved):

PR #2975, opened 2024-01-08
move migration 20200219145345

stanoMilan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Move migration Version 20200219145345 has wrong namespace. Entity Adverts has properties $datetimeVisibleFrom and $datetimeVisibleTo in entity on framework package.
|New feature| No 
|Fixes issues| ... 

### DIFF
--- a/packages/framework/src/Migrations/Version20200219145345.php
+++ b/packages/framework/src/Migrations/Version20200219145345.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Migrations;
+namespace Shopsys\FrameworkBundle\Migrations;
 
 use Doctrine\DBAL\Schema\Schema;
 use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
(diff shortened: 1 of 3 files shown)


packages/framework/src/Migrations/Version20200219145345.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Migrations;
+namespace Shopsys\FrameworkBundle\Migrations;
 
 use Doctrine\DBAL\Schema\Schema;
 use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
```
grossmannmartin, line 8: I suggest wrapping the migration in `up` method with condition
```
if ($this->isAppMigrationNotInstalledRemoveIfExists('Version20200219145345')) {
    $this->sql('ALTER TABLE adverts ADD datetime_visible_from TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL');
    $this->sql('ALTER TABLE adverts ADD datetime_visible_to TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL');
}
```

that way the migration file may be safely deleted during upgrade and it's not necessary to check whether it's installed or not. 
So then the second bullet in the upgrade notes may be deleted

grossmannmartin (changes requested):

packages/framework/src/Migrations/Version20200219145345.php
sspooky13: hmm 🤔 that did not cross my mind. [Migration changed](https://github.com/shopsys/shopsys/pull/2975/commits/204d2ba7381481afc8093ec290a29a1089c3ab5a) and [upgrade log](https://github.com/shopsys/shopsys/pull/2975/commits/a5b6d30d9791638b02cbaac8dadd721ffc93d463) too

grossmannmartin (approved):

vitek-rostislav (approved):

vitek-rostislav (changes requested):
Hi @sspooky13, could you please target this PR against 14.0 branch so the fix is available in the version `v14.0`. After the merge, could you also please merge the `14.0` branch into `15.0`? :pray:

vitek-rostislav (approved):

PR #2977, opened 2024-01-08
[frontend-api] add display timezone to SettingsQuery

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| For proper formatting on the storefront, the API must provide the display timezone, plus all the dates must be stored with time.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://rv-display-timezone.odin.shopsys.cloud
  - https://cz.rv-display-timezone.odin.shopsys.cloud


packages/framework/src/Model/Store/ClosedDay/ClosedDay.php
```diff
@@ -36,7 +36,7 @@ class ClosedDay
 
     /**
      * @var \DateTimeImmutable
-     * @ORM\Column(type="date_immutable")
+     * @ORM\Column(type="datetime_immutable")
```
grossmannmartin: the date field in form also have the `'view_timezone' => DateTimeHelper::UTC_TIMEZONE,` set.

Maybe it would be nice to search the codebase for usage of the UTC timezone

packages/frontend-api/src/Model/Resolver/Settings/DisplayTimezoneQuery.php
```diff
@@ -22,6 +22,6 @@ public function __construct(
      */
     public function displayTimezoneQuery(): string
     {
-        return $this->displayTimeZoneProvider->getDisplayTimeZone()->getName();
+        return $this->domain->getDateTimeZone()->getName();
```
grossmannmartin: Shouldn't it be the other way around? I would expect the display timezone provider to take into account the timezone settings in domains

grossmannmartin (changes requested):

packages/frontend-api/src/Model/Resolver/Settings/DisplayTimezoneQuery.php
vitek-rostislav: re-worked in 60e23b277

vitek-rostislav: and 733dafcc0 :slightly_smiling_face:

packages/framework/src/Model/Store/ClosedDay/ClosedDay.php
vitek-rostislav: resolved in c45327a25

vitek-rostislav: resolved in c45327a25

UPGRADE-14.0.md
```diff
@@ -769,6 +769,21 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 -   add Category ID to CategoryHierarchyItem ([#2962](https://github.com/shopsys/shopsys/pull/2962))
     -   see #project-base-diff to update your project
 -   add display timezone to FE API SettingsQuery ([#2977](https://github.com/shopsys/shopsys/pull/2977))
+    -   `Shopsys\FrameworkBundle\Component\DateTimeHelper\DateTimeHelper` class was changed:
+        -   `UTC_TIMEZONE` constant was removed
+        -   `TIME_REGEX` constant was removed
+        -   `convertDatetimeStringFromDisplayTimeZoneToUtc()` method was removed
+        -   `convertDateTimeFromUtcToDisplayTimeZone()` method was removed
```
grossmannmartin: everything was added in this version, so I think the upgrade notes are not necessary

grossmannmartin (approved):

UPGRADE-14.0.md
vitek-rostislav: ah, good catch :ok_hand: fixed in f1a124537

packages/framework/src/Form/Admin/Store/StoreFormType.php
```diff
@@ -182,6 +197,7 @@ public function configureOptions(OptionsResolver $resolver): void
             ->setDefaults([
                 'data_class' => StoreData::class,
                 'attr' => ['novalidate' => 'novalidate'],
+                'constraints' => new Constraints\Callback($this->validateOpeningHours(...)),
```
grossmannmartin: this callback validator breaks any js validation, or?

packages/framework/src/Form/DatePickerType.php
```diff
@@ -15,9 +15,9 @@ class DatePickerType extends AbstractType
     public const FORMAT_JS = 'dd.mm.yy';
 
     /**
-     * @param \Shopsys\FrameworkBundle\Component\Localization\DisplayTimeZoneProviderInterface|null $displayTimeZoneProvider
+     * @param \Shopsys\FrameworkBundle\Component\Localization\DisplayTimeZoneProvider $displayTimeZoneProvider
```
grossmannmartin: I think it should be annotated with interface too instead of custom implementation

packages/framework/src/Form/DatePickerType.php
```diff
@@ -30,12 +30,9 @@ public function configureOptions(OptionsResolver $resolver)
             'widget' => 'single_text',
             'format' => static::FORMAT_PHP,
             'html5' => false,
+            'display_timezone' => $this->displayTimeZoneProvider->getDisplayTimeZoneForAdmin(),
```
grossmannmartin: display_timezone or view_timezone?

packages/framework/src/Migrations/Version20240130104139.php
```diff
@@ -0,0 +1,35 @@
...
+
+class Version20240130104139 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE stores ADD domain_id INT NOT NULL DEFAULT 1');
+        $storeDomainsData = $this->sql('SELECT store_id, domain_id FROM store_domains WHERE is_enabled = TRUE ORDER BY domain_id')->fetchAllAssociative();
```
grossmannmartin, line 18: storedomain db table should remain?

packages/framework/src/Migrations/Version20240131163656.php
```diff
@@ -0,0 +1,68 @@
...
+
+            if ($firstOpeningTime !== null && $firstClosingTime !== null) {
+                $this->insertIntoOpeningHours($firstOpeningTime, $firstClosingTime, $storeId, $dayOfWeek);
+            }
+
+            if ($secondOpeningTime !== null && $secondClosingTime !== null) {
+                $this->insertIntoOpeningHours($secondOpeningTime, $secondClosingTime, $storeId, $dayOfWeek);
+            }
+        }
+        $this->sql('DELETE FROM store_opening_hours WHERE opening_time IS NULL AND closing_time IS NULL');
```
grossmannmartin: this may be null at any point?

grossmannmartin (changes requested):

packages/framework/src/Form/Transformers/OpeningHoursCollectionTransformer.php
```diff
@@ -0,0 +1,38 @@
...
+        return OpeningHoursDataHelper::getOpeningHoursIndexedByDayNumber($value);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Store\OpeningHours\OpeningHoursData[][] $value
+     * @return \Shopsys\FrameworkBundle\Model\Store\OpeningHours\OpeningHoursData[]
+     */
+    public function reverseTransform($value)
+    {
+        $flattenedOpeningHours = [];
```
grossmannmartin: does it make sense to have this code also in helper?

packages/framework/src/Model/Store/StoreDataFactory.php
```diff
@@ -74,12 +73,14 @@ public function createFromStore(Store $store): StoreData
     }
 
     /**
+     * @param int $domainId
      * @return \Shopsys\FrameworkBundle\Model\Store\StoreData
      */
-    protected function createInstance(): StoreData
+    protected function createInstance(int $domainId): StoreData
```
grossmannmartin: createInstance should only create the object. Rest of the values set should be in setData or create/createfromstore methods. 
Currently, for example, data are set and are immediately overridden in createFromStore.

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
```diff
@@ -0,0 +1,209 @@
...
+    ) {
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Store\Store $store
+     * @return bool
+     */
+    public function isOpenNow(Store $store): bool
+    {
+        $now = (new DateTimeImmutable())->setTimezone($this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($store->getDomainId()));
```
grossmannmartin: nitpick:

may be written as 
```php
$now = new DateTimeImmutable(timezone: $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($store->getDomainId()));
```

grossmannmartin (changes requested):

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
```diff
@@ -0,0 +1,209 @@
...
+        4 => 'thursday',
+        5 => 'friday',
+        6 => 'saturday',
+        7 => 'sunday',
+    ];
+
+    /**
+     * @var \Spatie\OpeningHours\OpeningHours[]
+     */
+    protected array $openingHoursSetting = [];
```
grossmannmartin, line 28: Should this provider implement the ResetInterface? Or it's ok to keep this setting during whole message processing?

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
```diff
@@ -0,0 +1,209 @@
...
+     */
+    protected function addOpeningHourRangeToWeekSetting(
+        int $dayOfWeek,
+        ?string $openingTime,
+        ?string $closingTime,
+        array $weekSetting,
+    ): array {
+        if ($openingTime !== null && $closingTime !== null) {
+            $openingHoursRange = $openingTime . '-' . $closingTime;
+            $weekSetting[$this->getEnglishDayNameFromDayNumber($dayOfWeek)][] = $openingHoursRange;
```
grossmannmartin: Does it make sense to keep the opening and closing time nullable when they are ignored in that case? 
What would change if opening and closing times were non-nullable in entity (and in higher layers)

Then this method will maybe just `formatOpeningHourRange(string $openingTime, string $closingTime)`

and the usage will be 
```
$weekSetting[$this->getEnglishDayNameFromDayNumber($dayOfWeek)][] = formatOpeningHours()
```

I find it more understandable than passing array that is modified and returned

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
```diff
@@ -0,0 +1,209 @@
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Store\OpeningHours\OpeningHoursData[] $openingHoursData
+     * @return \Spatie\OpeningHours\OpeningHours
+     */
+    public function getOpeningHoursSettingFromData(array $openingHoursData): SpatieOpeningHours
+    {
+        $weekSetting = [];
+
+        foreach ($openingHoursData as $openingHourData) {
+            $weekSetting = $this->addOpeningHourRangeToWeekSetting($openingHourData->dayOfWeek, $openingHourData->openingTime, $openingHourData->closingTime, $weekSetting);
```
grossmannmartin: nitpick

I think the arguments may be on separate lines to improve readability

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Store\Store $store
+     * @return array[][]
+     */
+    protected function getExceptions(Store $store): array
+    {
+        $exceptions = [];
+        $closedDays = $this->closedDayFacade->getThisWeekClosedDaysNotExcludedForStore(
+            $this->domain->getId(),
```
grossmannmartin: Why current domain id? The whole provider seems independent of the application state. 
Couldn't it be `$store->getDomainId()`?

maybe then the repository method may be simplified too, I'm not sure now

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
```diff
@@ -0,0 +1,209 @@
...
+            $openingHourData = $this->openingHoursDataFactory->create();
+            $openingHourData->dayOfWeek = $this->getDayNumberFromEnglishDayName($dayName);
+
+            return [$openingHourData];
+        }
+
+        $openingHoursData = [];
+        /** @var \Spatie\OpeningHours\TimeRange $openingHour */
+        foreach ($openingHoursForDay->getIterator() as $openingHour) {
+            $openingHourData = $this->openingHoursDataFactory->create();
```
grossmannmartin: nitpick

Just thinking whether it would make sense to change signature of create method to 
`create(int $dayOfWeek)`

it seems like it has to be always constructed with the dayOfWeek or the object is invalid

packages/framework/src/Model/Store/StoreFacade.php
```diff
@@ -192,4 +178,47 @@ public function getStoresByIds(array $storeIds): array
...
+        $daysOfWeek = range(1, 7);
+        $missingDays = array_diff($daysOfWeek, $daysCovered);
+
+        foreach ($missingDays as $missingDay) {
+            $openingHourData = $this->openingHoursDataFactory->create();
+            $openingHourData->dayOfWeek = $missingDay;
+            $openingHours[] = $this->openingHoursFactory->create($openingHourData);
+        }
+
+        return array_map(
```
grossmannmartin: usually, I like array_map, but in this case wouldn't be this more readable?

```
foreach ($openingHours as $openingHour) {
    $openingHour->setStore($store);
}

return $openingHours;
```

maybe even better, the store may be set directly in both loops, or the openingHoursFactory may contain the method `createWithStore(OpeningHourData $data, Store $store)`

project-base/app/tests/FrontendApiBundle/Functional/Order/TransportInOrderValidationTest.php
```diff
@@ -108,28 +102,6 @@ public function testTransportPriceChanged(): void
         $this->assertTrue($response['data']['CreateOrder']['cart']['modifications']['transportModifications']['transportPriceChanged']);
     }
 
-    public function testDisabledPickupPlaceUnavailable(): void
```
grossmannmartin, line 111: do we want to test the pickup place from another domain? or it's tested elsewhere?

packages/framework/src/Resources/config/parameters_common.yaml
```diff
@@ -8,8 +8,7 @@ parameters:
     # Administration locale. Must be registered in domains.yaml
     shopsys.admin_locale: en
 
-    # All time values are displayed in this timezone (default is PHP timezone)
-    shopsys.display_timezone: ~
+    shopsys.admin_display_timezone: 'Europe/Prague'
```
grossmannmartin: I would like to have a comment with explanation of this setting 😇

project-base/app/config/parameters_common.yaml
```diff
@@ -21,7 +21,6 @@ parameters:
     container.autowiring.strict_mode: true
     container.dumper.inline_class_loader: true
     build-version: '0000000000000000_%kernel.environment%'
-    shopsys.display_timezone: Europe/Prague
     shopsys.cron_timezone: Europe/Prague
```
grossmannmartin, line 24: It seems kinda weird that we have one timezone set in the framework and the second (cron) is only defined in the framework and set in project-base.

packages/frontend-api/src/Model/Resolver/OpeningHours/OpeningHoursResolverMap.php
```diff
@@ -0,0 +1,102 @@
...
+     * @param \Shopsys\FrameworkBundle\Model\Store\OpeningHours\OpeningHoursDataFactory $openingHoursDataFactory
+     * @param \Shopsys\FrameworkBundle\Component\Localization\DisplayTimeZoneProvider $displayTimeZoneProvider
+     * @param \Shopsys\FrameworkBundle\Component\DateTimeHelper\DateTimeHelper $dateTimeHelper
+     * @param \Shopsys\FrameworkBundle\Model\Store\OpeningHours\StoreOpeningHoursProvider $storeOpeningHoursProvider
+     */
+    public function __construct(
+        protected readonly Domain $domain,
+        protected readonly ClosedDayFacade $closedDayFacade,
+        protected readonly OpeningHoursDataFactory $openingHoursDataFactory,
+        protected readonly DisplayTimeZoneProviderInterface $displayTimeZoneProvider,
```
grossmannmartin: those four dependencies seems unused now

packages/framework/src/Resources/translations/validators.cs.po
```diff
@@ -127,6 +127,9 @@ msgstr "Počet dní k expedici musí být větší nebo roven {{ compared_value
 msgid "Number of days to expedite must be whole number."
 msgstr "Počet dní k expedici musí být celé číslo"
 
+msgid "Opening hours setting is not valid"
+msgstr "Nastavení otevírací doby není validní"
```
grossmannmartin:
```suggestion
msgstr "Neplaté nastavení otevírací doby"
```

maybe?

packages/framework/src/Component/DateTimeHelper/DateTimeHelper.php
```diff
@@ -78,4 +51,25 @@ public static function getCurrentRoundedTimeForIntervalAndTimezone(
...
+    public static function createDateTimeFromTime(string $hoursAndMinutes): DateTimeImmutable
+    {
+        return new DateTimeImmutable(sprintf('1970-01-01 %s:00', $hoursAndMinutes));
+    }
+
+    /**
+     * @param int $domainId
+     * @return int
+     */
+    public function getDayOfWeek(int $domainId): int
```
grossmannmartin: maybe `getCurrentDayOfWeek`?

project-base/app/src/Form/Admin/PromoCodeFormTypeExtension.php
```diff
@@ -181,28 +180,20 @@ private function buildTimeValidationFormGroup(FormBuilderInterface $builder): vo
             'label' => t('Apply according to date and time limit'),
         ]);
 
-        $timeValidationGroup->add('dateValidFrom', DatePickerType::class, [
-            'view_timezone' => DateTimeHelper::UTC_TIMEZONE,
+        $timeValidationGroup->add('datetimeValidFrom', DateTimeType::class, [
```
grossmannmartin, line 183: Can you please notify the tester to test promo codes?

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Store/StoreFormType.php
vitek-rostislav: that is a good catch! Even more, the JS was already broken by another missing callback validation, fixed in 3be2397433

packages/framework/src/Form/DatePickerType.php
vitek-rostislav: true, fixed in 958cbcbe3e, 87d99c5dd4, and c759934ca9

packages/framework/src/Form/DatePickerType.php
vitek-rostislav: good catch, fixed in 1722d2beed

packages/framework/src/Migrations/Version20240131163656.php
vitek-rostislav: :ok_hand: 764c4cda64

packages/framework/src/Migrations/Version20240130104139.php
vitek-rostislav: :ok_hand: 764c4cda64

packages/framework/src/Migrations/Version20240131163656.php
vitek-rostislav: yeah, there might be entries entry with NULL opening hours for a day, see the screens below:
image
image

So after the migration, you end up with a row containing null values, however, now I am not sure why this was a problem :smile: I will test the migration one more time.

packages/framework/src/Form/Transformers/OpeningHoursCollectionTransformer.php
vitek-rostislav: well, unlike the other method, this code is used only here, but I am ok with moving it there (I actually like the result as it puts a name on the functionality that was otherwise anonymous in the `reverseTransform` method) 88fdeb53ef

packages/framework/src/Model/Store/StoreDataFactory.php
vitek-rostislav: fair point, fixed in 3aaee55ef9

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
vitek-rostislav: nice! 07c7a4f988

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
vitek-rostislav: I was thinking about that during the implementation and thought it was unnecessary here. On the other hand, now, with some time gap, I think it won't harm and maybe it will save someone from unexpected behavior in the future, so added in 5cee258c00

packages/framework/src/Form/DatePickerType.php
vitek-rostislav: aaand in 5d254979e1 :smile:

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
vitek-rostislav: great idea, reworked in 69985a44f9

packages/framework/src/Migrations/Version20240131163656.php
vitek-rostislav: solved by 69985a44f9

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
vitek-rostislav: the code has changed with the refactoring, but 9cbf232a34 :+1:

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
vitek-rostislav: makes sense - 093dad2fd8

packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
vitek-rostislav: great idea - 658d9162de

packages/framework/src/Model/Store/StoreFacade.php
vitek-rostislav: I like the `createWithStore` approach - b88577fa93

project-base/app/tests/FrontendApiBundle/Functional/Order/TransportInOrderValidationTest.php
vitek-rostislav: To me, it was an edge case - the admin would have to change the store domain in production when a customer had the store selected as a pickup store in his cart.

I could create a test for this behavior but honestly, I do not want to postpone the task with that anymore

packages/framework/src/Resources/config/parameters_common.yaml
vitek-rostislav: added in 6330541347

project-base/app/config/parameters_common.yaml
vitek-rostislav: lets discuss it please, if it was up to me, I would move the cron timezone value setting to the framework. But maybe it was [implemented](https://github.com/shopsys/shopsys/commit/262c9861) this way for a reason that I do not see at the moment.

packages/frontend-api/src/Model/Resolver/OpeningHours/OpeningHoursResolverMap.php
vitek-rostislav: :+1: I have removed them in 69985a44f9df3335fe416ec5f615bb790c69f383

packages/framework/src/Resources/translations/validators.cs.po
vitek-rostislav: thanks, that is definitely less developer-slangish :smile:

packages/framework/src/Component/DateTimeHelper/DateTimeHelper.php
vitek-rostislav: yes! :ok_hand: c360e0befa

project-base/app/src/Form/Admin/PromoCodeFormTypeExtension.php
vitek-rostislav: sure thing :+1:

packages/framework/src/Model/Store/OpeningHours/OpeningHoursRange.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public function getOpeningTime(): string
+    {
+        return $this->openingTime;
+    }
+
+    /**
+     * @return string
+     */
+    public function getClosingTime(): string
```
grossmannmartin: after rebase the check for not using types in entities should warn you about setters, getters, and properties in entity and data object. Can you please verify then?

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
```diff
@@ -0,0 +1,64 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Store\OpeningHours;
+
+use DateTimeImmutable;
+use Shopsys\FrameworkBundle\Model\Store\OpeningHours\StoreOpeningHoursProvider;
+use Shopsys\FrameworkBundle\Model\Store\Store;
+
+class StoreOpeningHoursApiProvider extends StoreOpeningHoursProvider
```
grossmannmartin: I don't think this provider should extend the one from framework. 
I believe the composition is a better design pattern in this case.

And maybe some methods may be moved to frontend api package... or maybe only one storeopeninghoursprovider in framework is sufficient 🤔  

let's discuss it please

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * @param \DateTimeImmutable $date
+     * @param \Shopsys\FrameworkBundle\Model\Store\Store $store
+     * @return \Shopsys\FrontendApiBundle\Model\Store\OpeningHours\OpeningHoursWithDateData
+     */
+    protected function getOpeningHoursDataForDate(DateTimeImmutable $date, Store $store): OpeningHoursWithDateData
+    {
+        $openingHoursForDay = $this->getOpeningHoursSetting($store)->forDate($date);
```
grossmannmartin: nitpick

`...ForDay` or `...ForDate`?

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Store\Store $store
+     * @return \Shopsys\FrontendApiBundle\Model\Store\OpeningHours\OpeningHoursWithDateData[]
+     */
+    public function getFollowingWeekOpeningHours(Store $store): array
+    {
+        $openingHoursData = [];
+
+        $today = new DateTimeImmutable('today', timezone: $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($store->getDomainId()));
```
grossmannmartin:
```suggestion
        $today = new DateTimeImmutable('today', $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($store->getDomainId()));
```

timezone is a second argument, so named argument is unnecessary

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
```diff
@@ -0,0 +1,64 @@
...
+
+        for ($i = 0; $i <= 6; $i++) {
+            $day = $today->modify("+${i} days");
```
grossmannmartin: in PHP 8.3 the proper way is this
```suggestion
            $day = $today->modify("+{$i} days");
```

grossmannmartin (changes requested):

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
vitek-rostislav: b61c2061f9

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
vitek-rostislav: yes! 3da408586a

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
vitek-rostislav: thanks, resolved in https://github.com/shopsys/shopsys/commit/b61c2061f9673808794d71538399a24e737ca785

packages/frontend-api/src/Model/Store/OpeningHours/StoreOpeningHoursApiProvider.php
vitek-rostislav: thank you! resolved in https://github.com/shopsys/shopsys/commit/b61c2061f9673808794d71538399a24e737ca785

project-base/app/config/parameters_common.yaml
vitek-rostislav: unified in 468a5a17c1 :+1:

grossmannmartin (approved):

TomasLudvik (approved):

pk16011990 (approved):

PR #2980, opened 2024-01-12
[SSP-2007] entity changes log

RostislavKreisinger:
| Q             | A
| ------------- | ---
|Description, reason for the PR| we wants log changeset from changed entity and show changes in admin edit form of logged entity
|New feature| Yes

----
  - https://rk-ssp-2007-entity-changes-log.odin.shopsys.cloud
  - https://cz.rk-ssp-2007-entity-changes-log.odin.shopsys.cloud

### DIFF
--- a/UPGRADE-14.0.md
+++ b/UPGRADE-14.0.md
@@ -1137,6 +1137,27 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 
     -   see #project-base-diff to update your project
 
+-   log entity changes ([#2980](https://github.com/shopsys/shopsys/pull/2980))
+
+    -   see #project-base-diff to update your project
+    -   if you have extended `OrderItem`, or `Order` don't forget to add the appropriate loggable attribute to keep the entity logged – see https://docs.shopsys.com/en/14.0/nodel/log-entity-changes/
+    -   `Shopsys\FrameworkBundle\Controller\Admin\OrderController::__construct()` changed its interface:
+
+    ```diff
+        public function __construct(
+            protected readonly OrderFacade $orderFacade,
+            protected readonly AdvancedSearchOrderFacade $advancedSearchOrderFacade,
+            protected readonly OrderItemPriceCalculation $orderItemPriceCalculation,
+            protected readonly AdministratorGridFacade $administratorGridFacade,
+            protected readonly GridFactory $gridFactory,
+            protected readonly BreadcrumbOverrider $breadcrumbOverrider,
+            protected readonly OrderItemFacade $orderItemFacade,
+            protected readonly Domain $domain,
+            protected readonly OrderDataFactoryInterface $orderDataFactory,
+            protected readonly AdminDomainFilterTabsFacade $adminDomainFilterTabsFacade,
+        +   protected readonly EntityLogGridFactory $entityLogGridFactory,
+    ```
+
 -   sent emails via async queue ([#2998](https://github.com/shopsys/shopsys/pull/2998))
     -   see #project-base-diff to update your project
 -   change the product catnums field type in GrapesJs to text (([#2994](https://github.com/shopsys/shopsys/pull/2994)))
--- a/packages/framework/src/Component/EntityLog/Attribute/EntityLogIdentify.php
+++ b/packages/framework/src/Component/EntityLog/Attribute/EntityLogIdentify.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\EntityLog\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_METHOD)]
+class EntityLogIdentify
+{
+    public const IS_LOCALIZED = true;
+
+    /**
+     * @param bool $isLocalized
+     */
+    public function __construct(
+        public bool $isLocalized = false,
+    ) {
+    }
+}
--- a/packages/framework/src/Component/EntityLog/ChangeSet/ChangeSetResolver.php
+++ b/packages/framework/src/Component/EntityLog/ChangeSet/ChangeSetResolver.php
@@ -1,0 +1,154 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet;
+
+use Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableEntityConfigFactory;
+use Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\DataTypeResolver\DataTypeResolverInterface;
+use Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade;
+use Webmozart\Assert\Assert;
+
+class ChangeSetResolver
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\DataTypeResolver\DataTypeResolverInterface[]
+     */
+    protected array $dataTypeResolvers;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\DataTypeResolver\DataTypeResolverInterface[] $dataTypeResolvers
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableEntityConfigFactory $loggableEntityConfigFactory
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade $entityLogFacade
+     */
+    public function __construct(
+        iterable $dataTypeResolvers,
+        protected readonly LoggableEntityConfigFactory $loggableEntityConfigFactory,
+        protected readonly EntityLogFacade $entityLogFacade,
+    ) {
+        $this->registerDataTypeResolvers($dataTypeResolvers);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\DataTypeResolver\DataTypeResolverInterface[] $dataTypeResolvers
+     */
+    protected function registerDataTypeResolvers(iterable $dataTypeResolvers): void
+    {
+        Assert::allIsInstanceOf($dataTypeResolvers, DataTypeResolverInterface::class);
+
+        $dataTypeResolversArray = [];
+
+        foreach ($dataTypeResolvers as $dataTypeResolver) {
+            $dataTypeResolversArray[] = $dataTypeResolver;
+        }
+
+        //sort data type resolvers from higher to lower priority
+        usort($dataTypeResolversArray, function (DataTypeResolverInterface $a, DataTypeResolverInterface $b) {
+            return $b->getPriority() - $a->getPriority();
+        });
+
+        $this->dataTypeResolvers = $dataTypeResolversArray;
+    }
+
+    /**
+     * @param array $changeSet
+     * @param object $entity
+     * @return array
+     */
+    public function resolveChangeSetForEntity(array $changeSet, object $entity): array
+    {
+        $loggableSetup = $this->loggableEntityConfigFactory->getLoggableSetupByEntity($entity);
+
+        $resolvedChangeSet = [];
+
+        foreach ($changeSet as $property => $changes) {
+            if (!$loggableSetup->isPropertyLoggable($property)) {
+                continue;
+            }
+            $resolvedChanges = $this->getResolvedChanges($changes);
+
+            if ($resolvedChanges) {
+                $resolvedChangeSet[$property] = $resolvedChanges;
+            }
+        }
+
+        return $resolvedChangeSet;
+    }
+
+    /**
+     * @param array $changes
+     * @return \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ResolvedChanges|null
+     */
+    protected function getResolvedChanges(array $changes): ?ResolvedChanges
+    {
+        foreach ($this->dataTypeResolvers as $dataTypeResolver) {
+            if ($dataTypeResolver->isResolvedDataTypeByChanges($changes)) {
+                $resolvedChanges = $dataTypeResolver->getResolvedChanges($changes);
+
+                if ($resolvedChanges->isOldValueSameAsNewValue()) {
+                    return null;
+                }
+
+                return $resolvedChanges;
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * @param \Doctrine\ORM\PersistentCollection[] $scheduledCollections
+     * @param object $entity
+     * @return array
+     */
+    public function resolveChangesOnCollectionForEntity(array $scheduledCollections, object $entity): array
+    {
+        $resolvedChangedCollections = [];
+
+        foreach ($scheduledCollections as $scheduledCollection) {
+            if ($scheduledCollection->getOwner() !== $entity) {
+                continue;
+            }
+
+            $collectionChanges = new CollectionChanges();
+
+            foreach ($scheduledCollection->getInsertDiff() as $insertEntity) {
+                $loggableSetup = $this->loggableEntityConfigFactory->getLoggableSetupByEntity($insertEntity);
+                $newReadableValue = $this->entityLogFacade->getEntityReadableIdentifierByEntityAndLoggableSetup(
+                    $insertEntity,
+                    $loggableSetup,
+                );
+
+                $collectionChanges->insertedItems[] = (new ResolvedChanges(
+                    EntityLogFacade::getEntityNameByEntity($insertEntity),
+                    null,
+                    null,
+                    $newReadableValue ?? $insertEntity->getId(),
+                    $insertEntity->getId(),
+                ));
+            }
+
+            foreach ($scheduledCollection->getDeleteDiff() as $deleteEntity) {
+                $loggableSetup = $this->loggableEntityConfigFactory->getLoggableSetupByEntity($deleteEntity);
+                $oldReadableValue = $this->entityLogFacade->getEntityReadableIdentifierByEntityAndLoggableSetup(
+                    $deleteEntity,
+                    $loggableSetup,
+                );
+
+                $collectionChanges->deletedItems[] = (new ResolvedChanges(
+                    EntityLogFacade::getEntityNameByEntity($deleteEntity),
+                    $oldReadableValue ?? $deleteEntity->getId(),
+                    $deleteEntity->getId(),
+                    null,
+                    null,
+                ));
+            }
+
+            $assoc = $scheduledCollection->getMapping();
+            $collectionFieldName = $assoc['fieldName'];
+            $resolvedChangedCollections[$collectionFieldName] = $collectionChanges;
+        }
+
+        return $resolvedChangedCollections;
+    }
+}
--- a/packages/framework/src/Component/EntityLog/ChangeSet/DataTypeResolver/DateTimeDataTypeResolver.php
+++ b/packages/framework/src/Component/EntityLog/ChangeSet/DataTypeResolver/DateTimeDataTypeResolver.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\DataTypeResolver;
+
+use DateTime;
+use DateTimeImmutable;
+use Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ResolvedChanges;
+use Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade;
+
+class DateTimeDataTypeResolver extends AbstractDataTypeResolver
+{
+    protected const DATE_TIME_FORMAT_WITH_TIMEZONE = 'c';
+    protected const DATE_TIME_FORMAT_FOR_HUMAN = 'd. m. Y H:i:s';
+
+    /**
+     * @param mixed $value
+     * @return bool
+     */
+    protected function isResolvedDataType(mixed $value): bool
+    {
+        return $value instanceof DateTime || $value instanceof DateTimeImmutable;
+    }
+
+    /**
+     * @param array{0: \DateTimeInterface|null, 1: \DateTimeInterface|null} $changes
+     * @return \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ResolvedChanges
+     */
+    public function getResolvedChanges(array $changes): ResolvedChanges
+    {
+        $oldDateTime = $changes[0];
+        $newDateTime = $changes[1];
+
+        return new ResolvedChanges(
+            EntityLogFacade::getEntityNameByEntity($oldDateTime ?? $newDateTime),
+            null,
+            $oldDateTime?->format(static::DATE_TIME_FORMAT_WITH_TIMEZONE),
+            null,
+            $newDateTime?->format(static::DATE_TIME_FORMAT_WITH_TIMEZONE),
+        );
+    }
+
+    /**
+     * @return int
+     */
+    public function getPriority(): int
+    {
+        return 1;
+    }
+}
--- a/packages/framework/src/Component/EntityLog/ChangeSet/Formatter/CollectionChangesFormatter.php
+++ b/packages/framework/src/Component/EntityLog/ChangeSet/Formatter/CollectionChangesFormatter.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\Formatter;
+
+class CollectionChangesFormatter
+{
+    /**
+     * @param array $changes
+     * @return string
+     */
+    public function formatChanges(array $changes): string
+    {
+        $formattedCollectionChanges = [];
+        $this->formatInsertedItems($changes['insertedItems'], $formattedCollectionChanges);
+        $this->formatDeletedItems($changes['deletedItems'], $formattedCollectionChanges);
+
+        return implode(',<br> ', $formattedCollectionChanges);
+    }
+
+    /**
+     * @param array $insertedChanges
+     * @param string[] $formattedCollectionChanges
+     */
+    protected function formatInsertedItems(array $insertedChanges, array &$formattedCollectionChanges): void
+    {
+        foreach ($insertedChanges as $insertedChange) {
+            $formattedCollectionChanges[] = t('Created %dataType%: "%readableValue%"', ['%dataType%' => $insertedChange['dataType'], '%readableValue%' => $insertedChange['newReadableValue']]);
+        }
+    }
+
+    /**
+     * @param array $deletedChanges
+     * @param string[] $formattedCollectionChanges
+     */
+    protected function formatDeletedItems(array $deletedChanges, array &$formattedCollectionChanges): void
+    {
+        foreach ($deletedChanges as $deletedChange) {
+            $formattedCollectionChanges[] = t('Removed %dataType%: "%readableValue%"', ['%dataType%' => $deletedChange['dataType'], '%readableValue%' => $deletedChange['oldReadableValue']]);
+        }
+    }
+}
--- a/packages/framework/src/Component/EntityLog/ChangeSet/ResolvedChanges.php
+++ b/packages/framework/src/Component/EntityLog/ChangeSet/ResolvedChanges.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet;
+
+class ResolvedChanges
+{
+    /**
+     * @param string $dataType
+     * @param mixed $oldReadableValue
+     * @param mixed $oldValue
+     * @param mixed $newReadableValue
+     * @param mixed $newValue
+     */
+    public function __construct(
+        public string $dataType,
+        public mixed $oldReadableValue,
+        public mixed $oldValue,
+        public mixed $newReadableValue,
+        public mixed $newValue,
+    ) {
+    }
+
+    /**
+     * @return bool
+     */
+    public function isOldValueSameAsNewValue(): bool
+    {
+        return $this->oldValue === $this->newValue;
+    }
+}
--- a/packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
+++ b/packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
@@ -1,0 +1,171 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\EntityLog\EventListener;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Event\PostFlushEventArgs;
+use Doctrine\ORM\Event\PostPersistEventArgs;
+use Doctrine\ORM\Event\PostUpdateEventArgs;
+use Doctrine\ORM\Event\PreRemoveEventArgs;
+use Psr\Log\LoggerInterface;
+use Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableEntityConfig;
+use Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableEntityConfigFactory;
+use Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ChangeSetResolver;
+use Shopsys\FrameworkBundle\Component\EntityLog\Enum\EntityLogActionEnum;
+use Shopsys\FrameworkBundle\Component\EntityLog\Enum\EntityLogActionEnumInterface;
+use Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade;
+use Symfony\Contracts\Service\ResetInterface;
+use Throwable;
+
+class EntityLogEventListener implements ResetInterface
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLog[]
+     */
+    protected array $logs = [];
+
+    /**
+     * @param \Doctrine\ORM\EntityManagerInterface $em
+     * @param \Psr\Log\LoggerInterface $monolog
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableEntityConfigFactory $loggableEntityConfigFactory
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ChangeSetResolver $changeSetResolver
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade $entityLogFacade
+     */
+    public function __construct(
+        protected readonly EntityManagerInterface $em,
+        protected readonly LoggerInterface $monolog,
+        protected readonly LoggableEntityConfigFactory $loggableEntityConfigFactory,
+        protected readonly ChangeSetResolver $changeSetResolver,
+        protected readonly EntityLogFacade $entityLogFacade,
+    ) {
+    }
+
+    /**
+     * @param \Doctrine\ORM\Event\PostFlushEventArgs $args
+     */
+    public function postFlush(PostFlushEventArgs $args): void
+    {
+        if (count($this->logs) <= 0) {
+            return;
+        }
+
+        $logCollectionNumber = uniqid('entityLog', true);
+
+        foreach ($this->logs as $log) {
+            $log->setLogCollectionNumber($logCollectionNumber);
+            $this->em->persist($log);
+        }
+
+        $this->logs = [];
+        $this->em->flush();
+    }
+
+    //create
+    /**
+     * @param \Doctrine\ORM\Event\PostPersistEventArgs $args
+     */
+    public function postPersist(PostPersistEventArgs $args): void
+    {
+        $entity = $args->getObject();
+        $this->log(EntityLogActionEnum::CREATE, $entity);
+    }
+
+    //update
+    /**
+     * @param \Doctrine\ORM\Event\PostUpdateEventArgs $args
+     */
+    public function postUpdate(PostUpdateEventArgs $args): void
+    {
+        $entity = $args->getObject();
+        $this->log(EntityLogActionEnum::UPDATE, $entity);
+    }
+
+    //delete
+    /**
+     * @param \Doctrine\ORM\Event\PreRemoveEventArgs $args
+     */
+    public function preRemove(PreRemoveEventArgs $args): void
+    {
+        $entity = $args->getObject();
+        $this->log(EntityLogActionEnum::DELETE, $entity);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Enum\EntityLogActionEnum $actionEnum
+     * @param object $entity
+     */
+    protected function log(EntityLogActionEnumInterface $actionEnum, object $entity): void
+    {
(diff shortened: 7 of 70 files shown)


packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
```diff
@@ -0,0 +1,171 @@
...
+
+use Doctrine\ORM\EntityManagerInterface;
+use ReflectionAttribute;
+use ReflectionClass;
+use Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver;
+use Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade;
+use Shopsys\FrameworkBundle\Model\Localization\Localization;
+
+// TODO v rámci téhle servisy se pro čtení atributů použítvá php reflectionXXX ale v projektu je BetterReflection balíček (https://github.com/Roave/BetterReflection), který by mohl zlepšit práci s atributy,
+// hlavně v případě kdy chceme z nějakého atributu vytáhnout jeho parametry pomocí jmén a né jen číslovaným indexem do pole.
```
TomasLudvik: As discussed on Slack - this TODO is out of date, please remove it.

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        $reflectionClass = new ReflectionClass($entityFullyQualifiedName);
+        $loggableAttributes = $reflectionClass->getAttributes(Loggable::class, ReflectionAttribute::IS_INSTANCEOF);
+        $isLoggable = count($loggableAttributes) > 0;
+
+        $loggableSetup = new LoggableSetup(
+            $entityName,
+            $entityFullyQualifiedName,
+            $isLoggable && method_exists($entityFullyQualifiedName, 'getId'), //TODO asi bude fajn říct programátorovi, že když nemá entita fci getId, nejde logovat :-D
```
TomasLudvik: Please add some mention to the docs so this is clear and TODO can be removed. If you do not want to add a whole new section about logging, you can add it at least to the FAQ.

packages/framework/src/Component/EntityLog/Attribute/LoggableSetup.php
```diff
@@ -0,0 +1,194 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\EntityLog\Attribute;
+
+class LoggableSetup
```
TomasLudvik: Setup is connected to installation, I think you wanted to use Config or some similar word.

As this class is connected strictly to Entities, I would name it something like: LoggableEntityConfig or LoggableEntityDefinition

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
```diff
@@ -0,0 +1,171 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+use ReflectionAttribute;
+use ReflectionClass;
+use Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver;
+use Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade;
+use Shopsys\FrameworkBundle\Model\Localization\Localization;
+
+// TODO v rámci téhle servisy se pro čtení atributů použítvá php reflectionXXX ale v projektu je BetterReflection balíček (https://github.com/Roave/BetterReflection), který by mohl zlepšit práci s atributy,
+// hlavně v případě kdy chceme z nějakého atributu vytáhnout jeho parametry pomocí jmén a né jen číslovaným indexem do pole.
+class AttributeReader
```
TomasLudvik: I think this class is Factory of LoggableSetup - I would rename it to something like LoggableEntityConfigFactory so its purpose is more clear.

packages/framework/src/Component/EntityLog/Attribute/EntityLogIdentify.php
```diff
@@ -0,0 +1,21 @@
...
+#[Attribute(Attribute::TARGET_METHOD)]
+class EntityLogIdentify
+{
+    public const IS_LOCALIZED = true;
+
+    /**
+     * @param bool $isLocalized
+     */
+    public function __construct(
+        public bool $isLocalized = false,
```
grossmannmartin, line 18: nitpick – may it be readonly?

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * @param object|string $objectOrClass
+     * @return \Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableSetup
+     */
+    public function getLoggableSetupByEntity(object|string $objectOrClass): LoggableSetup
+    {
+        $entityFullyQualifiedName = is_string($objectOrClass) ? $objectOrClass : get_class($objectOrClass);
+        $entityFullyQualifiedName = str_replace('Proxies\__CG__\\', '', $entityFullyQualifiedName);
```
grossmannmartin: similar code is used in `EntityLogFacade::getEntityNameByEntity()` method. 
I think it should be unified. 

Maybe even return a new data object
```
class EntityNameData {
    public function __construct(
        public readonly string $fullyQualifiedName,
        public readonly string $shorName,
    )
}
```

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
```diff
@@ -0,0 +1,177 @@
...
+    public function __construct(
+        protected readonly EntityManagerInterface $em,
+        protected readonly LoggerInterface $monolog,
+        protected readonly AttributeReader $attributeReader,
+        protected readonly ChangeSetResolver $changeSetResolver,
+        protected readonly EntityLogFacade $entityLogFacade,
+    ) {
+    }
+
+    public function enable(): void
```
grossmannmartin: It seems like it's no way of disabling it (the default value of the property is also `true`).
Maybe this method is not necessary?

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if (!$this->enabled) {
+            return;
+        }
+
+        if (count($this->logs) <= 0) {
+            return;
+        }
+
+        $logCollectionNumber = uniqid('entityLog');
```
grossmannmartin: I would add true as a second argument to increase likelihood of unique value – see warning in documentation https://www.php.net/manual/en/function.uniqid.php
(it would require also changing the size of the db field)

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
```diff
@@ -0,0 +1,177 @@
...
+use Doctrine\ORM\Event\PreRemoveEventArgs;
+use Exception;
+use Psr\Log\LoggerInterface;
+use Shopsys\FrameworkBundle\Component\EntityLog\Attribute\AttributeReader;
+use Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableSetup;
+use Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ChangeSetResolver;
+use Shopsys\FrameworkBundle\Component\EntityLog\Enum\EntityLogActionEnum;
+use Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade;
+
+class EntityLogEventListener
```
grossmannmartin: I think this listener should implement `Symfony\Contracts\Service\ResetInterface`, so the `$logs` array will be cleared during message processing even in case of error.

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
```diff
@@ -0,0 +1,177 @@
...
+            if (!$this->enabled || !$loggableSetup->isLoggable()) {
+                return;
+            }
+
+            match ($actionEnum) {
+                EntityLogActionEnum::UPDATE => $this->resolveUpdateEntity($entity, $loggableSetup),
+
+                default => $this->registerLog($entity, $loggableSetup, $actionEnum),
+            };
+        } catch (Exception $exception) {
```
grossmannmartin: Maybe `Throwable` instead?

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
```diff
@@ -0,0 +1,177 @@
...
+
+        try {
+            if (!$this->enabled || !$loggableSetup->isLoggable()) {
+                return;
+            }
+
+            match ($actionEnum) {
+                EntityLogActionEnum::UPDATE => $this->resolveUpdateEntity($entity, $loggableSetup),
+
+                default => $this->registerLog($entity, $loggableSetup, $actionEnum),
```
grossmannmartin: If we're deciding here what method to call according to $actionEnum and then we still have to pass the $actionEnum to the called method, then I believe it may be simplified. 

for example like 
```
protected function registerLog(
        object $entity,
        LoggableSetup $loggableSetup,
        EntityLogActionEnum $actionEnum,
    ): void {
        $changes = [];

        if ($actionEnum === EntityLogActionEnum::UPDATE) {
            $changes = $this->resolveUpdateEntity($entity);  // maybe rename to resolveUpdateChangeset

            if (count($changes) === 0) {
                return;
            }
        }

        $entityLog = $this->entityLogFacade->createEntityLog($entity, $loggableSetup, $actionEnum, $changes);

        $this->logs[] = $entityLog;
    }
```

in the `log` method, this will be called without the match, and the `resolveUpdateEntity()` just return the `$resolvedChangeSet`

packages/framework/src/Component/EntityLog/Model/EntityLogFacade.php
```diff
@@ -0,0 +1,135 @@
...
+            $loggableSetup->getParentEntityIdentityFunctionName() ? $entity->{$loggableSetup->getParentEntityFunctionName()}()->{$loggableSetup->getParentEntityIdentityFunctionName()}() : null,
+        );
+    }
+
+    /**
+     * @param object $entity
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableSetup $loggableSetup
+     * @return int
+     */
+    protected function getEntityIdentifierByEntityAndLoggableSetup(object $entity, LoggableSetup $loggableSetup): int
```
grossmannmartin: This method may accept only $entity and the exception may be like 
```
throw new NotLoggableException(sprintf('Entity %s without ID as primary key is not loggable.', $entity::class));
 ```
 
 or?

File: packages/framework/src/Component/EntityLog/Model/EntityLogFacade.php [hunk 13]
```diff
@@ -0,0 +1,135 @@
...
+
+        if ($administrator) {
+            $source = EntityLogSourceEnum::ADMIN;
+            $userIdentifier = $administrator->getUserIdentifier();
+        }
+
+        if ($currentCustomer) {
+            $source = EntityLogSourceEnum::USER;
+            $userIdentifier = $currentCustomer->getUserIdentifier();
+        }
```
grossmannmartin: I would like to separate the source and user detection into a separate class (or at least a method). 

The current problem is that an anonymous user in graphql is logged as a `system` event. 
Maybe we could set some value in FrontendApiController and check if the value is set here (or in a new class). 
admin detection seems fine

packages/framework/src/Resources/config/services.yaml
```diff
@@ -1039,3 +1039,18 @@ services:
...
+
+    Shopsys\FrameworkBundle\Component\EntityLog\EventListener\EntityLogEventListener:
+        tags:
+            - { name: doctrine.event_listener, event: postPersist, priority: 1 }
+            - { name: doctrine.event_listener, event: postUpdate, priority: 1 }
+            - { name: doctrine.event_listener, event: preRemove, priority: 1 }
+            - { name: doctrine.event_listener, event: postFlush, priority: 1 }
+
+    Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogRepository:
+        public: true
```
grossmannmartin: Why is necessary to make the service public?

packages/framework/src/Component/EntityLog/Model/EntityLogRepository.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        return $this->em->getRepository(EntityLog::class);
+    }
+
+    /**
+     * @param string $entityName
+     * @param int $entityId
+     * @return \Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLog[]
+     */
+    public function getEntityLogsByEntityNameAndEntityId(string $entityName, int $entityId): array
```
grossmannmartin: this method is used in the unused EntityLogFacade method and the test. 
I'm thinking about the meaning of the method. If it's used only in tests, then the test is testing some scenario that does not happen in the application, or?

edit: I probably understand the reason now, is it because the data are displayed only in the grid and the grid expects the querybuilder? If so, I would rather use the method returning querybuilder in tests. 
Currently I'm able to change implementation of the methods so tests will pass, but the grid will be broken

packages/framework/src/Component/EntityLog/ChangeSet/ChangeSetResolver.php
```diff
@@ -0,0 +1,157 @@
...
+        $resolvedChangeSet = [];
+
+        foreach ($changeSet as $property => $changes) {
+            if (!$loggableSetup->isPropertyLoggable($property)) {
+                continue;
+            }
+            $resolvedChanges = $this->getResolvedChanges($changes);
+
+            if ($resolvedChanges) {
+                $resolvedChangeSet[$property] = $resolvedChanges->toArray();
```
grossmannmartin: Is it necessary to make it an array here (and in resolveChangesOnCollectionForEntityMethod)? 
What if we just pass an array of objects through the layers and make the conversion to an array in the entity's setter? If I'm not mistaken, the array is necessary just to save the data properly to DB, or?

packages/framework/src/Component/EntityLog/ChangeSet/DataTypeResolver/DateTimeDataTypeResolver.php
```diff
@@ -0,0 +1,54 @@
...
+
+use DateTime;
+use DateTimeImmutable;
+use Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ResolvedChanges;
+use Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade;
+
+class DateTimeDataTypeResolver extends AbstractDataTypeResolver
+{
+    protected const DATE_TIME_FORMAT_WITH_TIMEZONE = 'c';
+    protected const DATE_TIME_FORMAT_FOR_HUMAN = 'd. m. Y H:i:s';
```
grossmannmartin, line 15: I'm not sure about the default readable format set here. Isn't possible to store just the ISO8601 value and make the conversion to readable format to the ResolverChangesFormatter? 

Also this does not take into account the display timezone, so when datetime field is changed, in the entity log will be displayed the time in the UTC, which will be different from the value the user set. 
(If we use the `DateTimeFormatter` or the `DateTimeFormatterExtension` to format the value, the value will be formatted in the appropriate display timezone)

packages/framework/src/Component/EntityLog/ChangeSet/ResolvedChanges.php
```diff
@@ -0,0 +1,40 @@
...
+     * @param string|null $oldValue
+     * @param string|null $newReadableValue
+     * @param string|null $newValue
+     */
+    public function __construct(
+        public string $dataType,
+        public mixed $oldReadableValue,
+        public mixed $oldValue,
+        public mixed $newReadableValue,
+        public mixed $newValue,
```
grossmannmartin, line 21: the annotation is string|null, shouldn't be the type the same? (`?string`)

packages/framework/src/Component/EntityLog/Model/EntityLog.php
```diff
@@ -0,0 +1,230 @@
...
+ */
+class EntityLog
+{
+    /**
+     * @var int
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="IDENTITY")
+     */
+    protected int $id;
```
grossmannmartin: just FYI, for the sake of extensibility, the framework entities do not use type hints (it's not possible to change the type in the extended entity – e.g. make it nullable). 
In this entity, I don't think it's a problem as I don't expect this to be extended – but I'll leave the final decision to you

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
```diff
@@ -0,0 +1,171 @@
...
+class AttributeReader
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Localization\Localization $localization
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Attribute\LoggableSetupCacheFacade $loggableSetupCacheFacade
+     * @param \Doctrine\ORM\EntityManagerInterface $em
+     * @param \Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver $entityNameResolver
+     */
+    public function __construct(
+        protected readonly Localization $localization,
```
grossmannmartin: this property seems unused

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
```diff
@@ -0,0 +1,171 @@
...
+        protected readonly LoggableSetupCacheFacade $loggableSetupCacheFacade,
+        protected readonly EntityManagerInterface $em,
+        protected readonly EntityNameResolver $entityNameResolver,
```
grossmannmartin: this property seems unused

packages/framework/src/Component/EntityLog/ChangeSet/ChangeSetResolver.php
```diff
@@ -0,0 +1,157 @@
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\DataTypeResolver\DataTypeResolverInterface[] $dataTypeResolvers
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Attribute\AttributeReader $attributeReader
+     * @param \Doctrine\ORM\EntityManagerInterface $em
+     * @param \Shopsys\FrameworkBundle\Component\EntityLog\Model\EntityLogFacade $entityLogFacade
+     */
+    public function __construct(
+        iterable $dataTypeResolvers,
+        protected readonly AttributeReader $attributeReader,
+        protected readonly EntityManagerInterface $em,
```
grossmannmartin: this property seems unused

packages/framework/src/Component/EntityLog/ChangeSet/Formatter/CollectionChangesFormatter.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * @param array $insertedChanges
+     * @param string[] $formattedCollectionChanges
+     */
+    protected function formatInsertedItems(array $insertedChanges, array &$formattedCollectionChanges): void
+    {
+        foreach ($insertedChanges as $insertedChange) {
+            $formattedCollectionChanges[] = t('Created %dataType%: "%readableValue%"', ['%dataType%' => $insertedChange['dataType'], '%readableValue%' => $insertedChange['newReadableValue']]);
```
grossmannmartin, line 29: can you please dump translations and add czech translations?

packages/framework/src/Resources/views/Admin/Content/EntityLog/listGrid.html.twig
```diff
@@ -0,0 +1,14 @@
...
+    {{ parent() }}
+{% endblock %}
+
+{% block grid_value_cell_id_changeSet %}
+    {{ value|raw }}
+{% endblock %}
+
+{% block grid_no_data %}
+    {{ 'No logs created so far.'|trans }}
+{% endblock %}
```
grossmannmartin: please ensure the line break at the end of the file

grossmannmartin (changes requested):

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
RostislavKreisinger: removed

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
RostislavKreisinger: removed, in docs is warning about this

packages/framework/src/Component/EntityLog/Attribute/LoggableSetup.php
RostislavKreisinger: ok, is renamed as LoggableEntityConfig.

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
RostislavKreisinger: ok, renamed

packages/framework/src/Component/EntityLog/Attribute/EntityLogIdentify.php
RostislavKreisinger: i dont know, as programmer you havent acces to instance of attribute class dicretctly. For resolving this attribute is user ReflectionAttribute class. I thing in this case readonly has no practical meaning.

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
RostislavKreisinger: ok, ill do it

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
RostislavKreisinger: ill remove whole enabling ability.

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
RostislavKreisinger: ok.

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
RostislavKreisinger: ok

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
RostislavKreisinger: ok

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
RostislavKreisinger: maybe :-)

packages/framework/src/Component/EntityLog/Model/EntityLogFacade.php
RostislavKreisinger: ok

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
RostislavKreisinger: removed

packages/framework/src/Component/EntityLog/ChangeSet/ChangeSetResolver.php
RostislavKreisinger: removed

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
RostislavKreisinger: removed

packages/framework/src/Component/EntityLog/Model/EntityLogRepository.php
RostislavKreisinger: ok

packages/framework/src/Resources/config/services.yaml
RostislavKreisinger: now isnt, removed.

packages/framework/src/Component/EntityLog/Model/EntityLogFacade.php
RostislavKreisinger: this problem we shoud discus.

packages/framework/src/Component/EntityLog/ChangeSet/ChangeSetResolver.php
RostislavKreisinger: Changeset serialization moved to LogEntity

packages/framework/src/Component/EntityLog/ChangeSet/DataTypeResolver/DateTimeDataTypeResolver.php
RostislavKreisinger: updated by suggestions

packages/framework/src/Component/EntityLog/ChangeSet/ResolvedChanges.php
RostislavKreisinger: fixed

packages/framework/src/Component/EntityLog/Model/EntityLog.php
RostislavKreisinger: i left it as it

packages/framework/src/Component/EntityLog/ChangeSet/Formatter/CollectionChangesFormatter.php
RostislavKreisinger: ok

packages/framework/src/Resources/views/Admin/Content/EntityLog/listGrid.html.twig
RostislavKreisinger: ok

RostislavKreisinger: ok

packages/framework/src/Component/EntityLog/Model/EntityLogFacade.php
RostislavKreisinger: changed according by discusion

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
RostislavKreisinger: whole functionality was removed

packages/framework/src/Component/EntityLog/Attribute/AttributeReader.php
RostislavKreisinger: good idea.

packages/framework/src/Component/EntityLog/Model/EntityNameData.php
```diff
@@ -0,0 +1,34 @@
...
+
+class EntityNameData
+{
+    /**
+     * @param string $fullyQualifiedName
+     * @param string $shorName
+     */
+    public function __construct(
+        protected readonly string $fullyQualifiedName,
+        protected readonly string $shorName,
```
grossmannmartin: I made a typo in a suggestion and this typo made its way here 😇 
Can you fix it pls?

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -116,7 +116,7 @@ public function createOrder(OrderData $orderData, OrderPreview $orderPreview, ?C
         );
 
         $this->em->persist($order);
-        $this->em->flush();
+//        $this->em->flush();
```
grossmannmartin: can you please remove the commented code?

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -1063,6 +1078,15 @@ msgstr "Zadaný zákazník nebyl nalezen, prosím překontrolujte zadané údaje
 msgid "Entered in days. Zero represents products in stock, empty means unknown availability."
 msgstr "Zadává se ve dnech. Nula reprezentuje zboží skladem, nevyplněné znamená, že dostupnost je neznámá."
 
+msgid "Entity"
+msgstr "Entita"
+
+msgid "Entity identifier"
+msgstr "Identifikárot entity"
```
grossmannmartin, line 1175:
```suggestion
msgstr "Identifikátor entity"
```

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -1651,6 +1675,9 @@ msgstr "Nebyly nalezeny žádné e-mailové šablony. Nejprve musíte nějaké v
 msgid "No flags found. You have to create some first."
 msgstr "Nebyly nalezeny žádné příznaky. Nejprve musíte nějaké vytvořit."
 
+msgid "No logs created so far."
+msgstr "Žádné protokly nejsou zatím vytvořeny."
```
grossmannmartin:
```suggestion
msgstr "Žádné protokoly nejsou zatím vytvořeny."
```

UPGRADE-14.0.md
```diff
@@ -629,6 +629,8 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   see #project-base-diff to update your project
 -   add failure transport for unprocessable messages ([#2958](https://github.com/shopsys/shopsys/pull/2958))
     -   see #project-base-diff to update your project
+-   log entity changes ([#2980](https://github.com/shopsys/shopsys/pull/2980))
+    -   see #project-base-diff to update your project
```
grossmannmartin, line 1142:
````suggestion
    -   if you have extended `Country`, `OrderItem`, `Order`, `OrderStatus`, `Payment`, or `Transport`, don't forget to add the appropriate loggable attribute to keep the entity logged – see https://docs.shopsys.com/en/14.0/cookbook/log-entity-changes/
    -   `Shopsys\FrameworkBundle\Controller\Admin\OrderController::__construct()` changed its interface:
    ```diff
        public function __construct(
            protected readonly OrderFacade $orderFacade,
            protected readonly AdvancedSearchOrderFacade $advancedSearchOrderFacade,
            protected readonly OrderItemPriceCalculation $orderItemPriceCalculation,
            protected readonly AdministratorGridFacade $administratorGridFacade,
            protected readonly GridFactory $gridFactory,
            protected readonly BreadcrumbOverrider $breadcrumbOverrider,
            protected readonly OrderItemFacade $orderItemFacade,
            protected readonly Domain $domain,
            protected readonly OrderDataFactoryInterface $orderDataFactory,
            protected readonly AdminDomainFilterTabsFacade $adminDomainFilterTabsFacade,
        +   protected readonly EntityLogGridFactory $entityLogGridFactory,
    ```
````

File: docs/cookbook/log-entity-changes.md [hunk 30]
```diff
@@ -0,0 +1,136 @@
+# Log entity changes
+
+The Shopsys platform allows you to log changes to entities.
+
+The logging system works to capture changes in the doctrine unit of work before they are flushed.
+
+After setting logging we used php attributes above class, property and method. (https://www.php.net/manual/en/language.attributes.overview.php)
```
grossmannmartin:
```suggestion
We used PHP attributes above class, property, and method for setting logging. (https://www.php.net/manual/en/language.attributes.overview.php)
```

docs/cookbook/log-entity-changes.md
```diff
@@ -0,0 +1,136 @@
...
+
+As an implemented sample, you can study the settings on Order, OrderItem, OrderStatus and Country entities. The principle of use will be described on these examples.
```
grossmannmartin: It would be nice to have each sentence on a separate line. It improves the readability of a markdown file (and diffs are nicer), while it will still be rendered as a single paragraph.

I don't insist on the change in this already written article (although I would not be against it 😄 ). 
Just for the future.

docs/cookbook/log-entity-changes.md
```diff
@@ -0,0 +1,136 @@
...
+
+The attributes `Loggable` and `LoggableChild` are used to mark the entity to be logged. Both of these attributes have logging strategy settings available. If I want to log all properties in the base, I use the `Loggable(Loggable::STRATEGY_INCLUDE_ALL)` strategy. If I would like to not log certain properties, I can mark them with the `ExcludeLog` attribute. Conversely, if I want to log only a few properties from an entity, it would be better to use the `Loggable(Loggable::STRATEGY_EXCLUDE_ALL)` strategy and then mark which properties I want to log using the `Log` attribute.
+!!! danger
+
+    **Now is possible log only entity with primary key `id` and method `getId()`**
```
grossmannmartin:
```suggestion
    **It is possible to log only entities with primary key `id` and method `getId()`**
```

docs/cookbook/log-entity-changes.md
```diff
@@ -0,0 +1,136 @@
...
+    public function getName()
+    {
+        return $this->name;
+    }
+
+    ...
+}
+```
+
+There are properties in the Order entity that are not only of scalar data type. For example, the status property is of data type OrderStatus or deliveryCountry is of data type Country (another entity). These entities themselves do not need to be logged in, but we want to see the human name on the order in the event of a status change. The `EntityLogIdentify` attribute is again used for this naming. So if I want to see in the order log that the status has changed from New to In Progress, I need to mark the getName function on the OrderStatus entity.
```
grossmannmartin:
```suggestion
There are properties in the Order entity that are not a simple scalar data type. For example, the status property is of data type OrderStatus or deliveryCountry is of data type Country (another entity). These entities themselves do not need to be logged, but we want to see the human name on the order in the event of a status change. The `EntityLogIdentify` attribute is again used for this naming. So if I want to see in the order log that the status has changed from New to In Progress, I need to mark the getName function on the OrderStatus entity.
```

grossmannmartin (changes requested):

packages/framework/src/Component/EntityLog/Model/EntityNameData.php
RostislavKreisinger: sure :-D

packages/framework/src/Model/Order/OrderFacade.php
RostislavKreisinger: removed

packages/framework/src/Resources/translations/messages.cs.po
RostislavKreisinger: fixed

UPGRADE-14.0.md
RostislavKreisinger: fixed

docs/cookbook/log-entity-changes.md
RostislavKreisinger: fixed

docs/cookbook/log-entity-changes.md
RostislavKreisinger: sure

docs/cookbook/log-entity-changes.md
RostislavKreisinger: sure

docs/cookbook/log-entity-changes.md
RostislavKreisinger: sure

grossmannmartin (approved):

docs/cookbook/log-entity-changes.md
```diff
@@ -0,0 +1,136 @@
+# Log entity changes
+
+The Shopsys platform allows you to log changes to entities.
```
vitek-rostislav:
```suggestion
The Shopsys platform allows you to log changes to entities. Thanks to the mechanism, you can answer the questions like "Who and when changed the order status?", "Why is the product price changed?", etc.
```

docs/cookbook/log-entity-changes.md
vitek-rostislav:
```suggestion
There are properties in the Order entity that are not only of scalar data type. For example, the status property is of data type OrderStatus or deliveryCountry is of data type Country (another entity). These entities themselves do not need to be logged, but we want to see the human name on the order in the event of a status change. The `EntityLogIdentify` attribute is again used for this naming. So if I want to see in the order log that the status has changed from "New" to "In Progress", I need to mark the getName function on the OrderStatus entity.
````

File: docs/cookbook/log-entity-changes.md [hunk 35]
```diff
@@ -0,0 +1,136 @@
...
+        return $this->name;
+    }
+
+    ...
+}
+```
+
+There are properties in the Order entity that are not only of scalar data type. For example, the status property is of data type OrderStatus or deliveryCountry is of data type Country (another entity). These entities themselves do not need to be logged in, but we want to see the human name on the order in the event of a status change. The `EntityLogIdentify` attribute is again used for this naming. So if I want to see in the order log that the status has changed from New to In Progress, I need to mark the getName function on the OrderStatus entity.
+
+```php
```
vitek-rostislav: the example does not match with the explanation above - the `OrderStatus` is mentioned there, so IMHO the code should show the `OrderStatus` entity, not `Country`

vitek-rostislav (changes requested):
Hi, I liked the docs, it is understandable. I have added a few suggestions, so check them out, please. Also, I am not sure about the location of the article in the "cookbook" section as technically, it is not a cookbook :upside_down_face: I suggest either creating a new section (called "entity changes log"? :thinking:), or moving the article into the existing section "model" that seems to me to be the closest match

docs/cookbook/log-entity-changes.md
RostislavKreisinger: ok

docs/cookbook/log-entity-changes.md
RostislavKreisinger: thx. fixed

UPGRADE-14.0.md
```diff
@@ -1136,6 +1136,24 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
         -   `ProductExtendedConnection` logic has been moved to `ProductConnection`
 
     -   see #project-base-diff to update your project
+-   log entity changes ([#2980](https://github.com/shopsys/shopsys/pull/2980))
+    -   see #project-base-diff to update your project
+    -   if you have extended `OrderItem`, or `Order` don't forget to add the appropriate loggable attribute to keep the entity logged – see https://docs.shopsys.com/en/14.0/cookbook/log-entity-changes/
```
vitek-rostislav:
```suggestion
    -   if you have extended `OrderItem`, or `Order` don't forget to add the appropriate loggable attribute to keep the entity logged – see https://docs.shopsys.com/en/14.0/model/log-entity-changes/
```

vitek-rostislav (approved):

UPGRADE-14.0.md
RostislavKreisinger: fixed

packages/framework/src/Component/EntityLog/ChangeSet/DataTypeResolver/DateTimeDataTypeResolver.php
```diff
@@ -0,0 +1,54 @@
...
+        return $value instanceof DateTime || $value instanceof DateTimeImmutable;
+    }
+
+    /**
+     * @param array $changes
+     * @return \Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ResolvedChanges
+     */
+    public function getResolvedChanges(array $changes): ResolvedChanges
+    {
+        /** @var \DateTimeImmutable|\DateTime|null $oldDateTime */
```
pk16011990: It can be written as `@param arrra{0: \DateTimeInterface|null, 1: \DateTimeInterface|null} $changes`

pk16011990 (approved):

Issue #2372, opened 2021-08-31
Annotations Fixer: extending classes from other than framework namespaces


vitek-rostislav:
## Describe the expected outcome of your requested feature
I noticed that in some classes in my project, the annotations are not fixed when running `php phing annotations-fix`. This is because I extend classes from [FE API Bundle](https://github.com/shopsys/frontend-api). I believe such namespaces should be included in the tool for fixing annotations as well (maybe all the Shopsys vendor namespaces?) or, it should be possible to configure the namespaces that I am extending in my project so the annotations are fixed properly then.

AFAIK, the `Shopsys\FrameworkBundle` namespace is hardcoded in 2-3 places now:
- https://github.com/shopsys/shopsys/blob/master/packages/framework/src/DependencyInjection/Compiler/RegisterProjectFrameworkClassExtensionsCompilerPass.php#L34
- https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Component/ClassExtension/ClassExtensionRegistry.php#L32
- https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Component/ClassExtension/ClassExtensionRegistry.php#L72

TomasLudvik:
Do not forget to update `\Tests\FrameworkBundle\Unit\Component\ClassExtension\AnnotationsReplacerTest` to ensure that this functionality is working correctly.

PR #2983, opened 2024-01-12
[shopsys] added Persoo bundle for product search

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We have added another way to search products - Persoo, that enables personalization and faster search.
|New feature| Yes 
|Fixes issues| closes #2380  

Notes for review: There are some YAGNI violations as unused constants etc. It is intentional as I discovered these values during development and I would not want to loose them.

----
  - https://tl-persoo-product-suggestions.odin.shopsys.cloud
  - https://cz.tl-persoo-product-suggestions.odin.shopsys.cloud


UPGRADE-14.0.md
```diff
@@ -768,6 +768,54 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   see #project-base-diff to update your project
 -   add Category ID to CategoryHierarchyItem ([#2962](https://github.com/shopsys/shopsys/pull/2962))
     -   see #project-base-diff to update your project
+-   add Persoo autocompletion
+    -   `Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterDataFactory` has been introduced so use it for creation of `ProductFilterData` instead of `new ProductFilterData()`
```
vitek-rostislav, line 936: the commit message should be fixed - it says `introduced ProductDataFactory in the framework` but it should be `introduced ProductFilterDataFactory in the framework`, right?

project-base/app/config/services.yaml
```diff
@@ -838,3 +838,6 @@ services:
 
     Shopsys\FrameworkBundle\Model\Product\ProductElasticsearchProvider:
         alias: App\Model\Product\ProductElasticsearchProvider
+
+    Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterDataFactory:
+        alias: App\Model\Product\Filter\ProductFilterDataFactory
```
vitek-rostislav, line 839: In `App\Model\Product\Filter\ProductFilterData`, there is just one method - `isFilterActive` which is not used anywhere. So we could get rid of it, as well as `App\Model\Product\Filter\ProductFilterDataFactory` :slightly_smiling_face:

project-base/app/src/FrontendApi/Model/Product/Filter/ProductFilterDataMapper.php
```diff
@@ -78,31 +61,4 @@ protected function getParametersAndValuesByUuids(array $parameterAndValueUuids):
 
         return $parametersFilterData;
     }
-
-    /**
-     * Method is extended because of https://github.com/shopsys/shopsys/pull/2380
```
vitek-rostislav, line 83: https://github.com/shopsys/shopsys/pull/2380 can be closed now

packages/framework/src/Component/ClassExtension/ClassExtensionRegistry.php
```diff
@@ -68,6 +68,22 @@ protected function getOtherClassesExtensionMap(): array
...
+        /** @var \Symfony\Component\Finder\SplFileInfo $file */
+        foreach ($finder as $file) {
+            $frontendApiClassFqcn = $this->getFqcn($file->getPathname());
+            $projectClassFqcn = str_replace('Shopsys\FrontendApiBundle', 'App\FrontendApi', $frontendApiClassFqcn);
+
+            if (class_exists($projectClassFqcn)) {
+                $otherClassesMap[$frontendApiClassFqcn] = $projectClassFqcn;
+            }
+        }
+
```
vitek-rostislav, line 86: does this (partly) resolve https://github.com/shopsys/shopsys/issues/2372? Could you please move it to a separate commit and extend tests with this new functionality?

packages/frontend-api/src/Model/Product/Connection/ProductConnection.php
```diff
@@ -6,31 +6,28 @@
...
+     * @param string|null $orderingMode
+     * @param null $totalCount
+     * @param string $defaultOrderingMode
      */
     public function __construct(
         array $edges,
         ?PageInfoInterface $pageInfo,
-        int $totalCount,
-        callable $productFilterOptionsClosure,
+        protected $productFilterOptionsClosure,
```
vitek-rostislav: could it be type-hinted with `Closure`? You would probably have to change the corresponding type hints in  `ProductConnectionFactory` from `callable` to `Closure` but that seems valid to me.

Also (nitpick), `$productFilterOptionsClosure`, `$orderingMode`, and `$defaultOrderingMode` properties in the constructor could be marked as `readonly`

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductsSearchTest.php
```diff
@@ -15,7 +15,7 @@ public function testSearchInAllProducts(): void
         $firstDomainLocale = $this->getFirstDomainLocale();
         $query = '
             query {
-                products (first: 5, search: "' . t('book', [], Translator::TESTS_TRANSLATION_DOMAIN, $firstDomainLocale) . '") {
```
vitek-rostislav, line 18: It is still possible to use search products via the `products` query on the API. It is defined in `ProductsQuery::productsQuery` and overridden in `project-base` in `productsWithOverlyingEntityQuery` method. I have a few notes on that:

- it now might be confusing which method should be used for the products search - `products` or `productSearch`?
- the original (well original, rather let's say the original overridden) query is not tested anymore as you changed these tests to use `productsSearch` query, is this ok?
- if we now just removed the "original" `products` query, we would probably lose the batch loading feature if I get the current implementation correctly

Makefile
```diff
@@ -1,10 +1,10 @@
...
-	docker cp /tmp/schema.graphql shopsys-framework-storefront:/home/node/app/schema.graphql
-	docker-compose exec -u root storefront chown node:node schema.graphql
-	docker-compose exec storefront pnpm run gql
-	docker-compose exec storefront rm -rf /home/node/app/schema.graphql
+	mutagen-compose exec php-fpm php phing frontend-api-generate-graphql-schema
+	mutagen-compose cp php-fpm:/var/www/html/project-base/app/schema.graphql /tmp/schema.graphql
+	mutagen-compose cp /tmp/schema.graphql storefront:/home/node/app/schema.graphql
+	mutagen-compose exec -u root storefront chown node:node schema.graphql
+	mutagen-compose exec storefront pnpm run gql
+	mutagen-compose exec storefront rm -rf /home/node/app/schema.graphql
```
vitek-rostislav: this was probably not intended to be committed, right? :slightly_smiling_face:

packages/frontend-api/src/DependencyInjection/RegisterProductsSearchResultsProvidersCompilerPass.php
```diff
@@ -0,0 +1,44 @@
...
+use Shopsys\FrontendApiBundle\Model\Resolver\Products\Search\ProductSearchResultsProviderResolver;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class RegisterProductsSearchResultsProvidersCompilerPass implements CompilerPassInterface
+{
+    /**
+     * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
+     */
+    public function process(ContainerBuilder $container)
```
vitek-rostislav:
```suggestion
    public function process(ContainerBuilder $container): void
```

packages/frontend-api/src/DependencyInjection/RegisterProductsSearchResultsProvidersCompilerPass.php
```diff
@@ -0,0 +1,44 @@
...
+            $priority = null;
+
+            foreach ($tags as $tag) {
+                if (array_key_exists('priority', $tag)) {
+                    $priority = $tag['priority'];
+                }
+            }
+
+            if (!is_int($priority)) {
+                throw new ProductSearchResultsProviderPriorityNotSetException(sprintf('Service "%s" has not defined required tag priority "%s" or its type is not integer.', $serviceId, implode(',', $tags)));
```
vitek-rostislav: the `tags` seems to be an array of arrays so when I removed the `priority` tag from `services.yaml`, I got an error:

```
In RegisterProductsSearchResultsProvidersCompilerPass.php line 32:
                                       
  [ErrorException]                     
  Warning: Array to string conversion 
```

Also, if it was fixed, what was the expected error message? Why would you like to implode the tags? The resulting sentence just do not make sense to me

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProvider.php
```diff
@@ -51,7 +51,16 @@ function ($offset, $limit) use ($search, $productFilterData) {
             $this->productFacade->getFilteredProductsCountOnCurrentDomain($productFilterData, $search),
             $argument,
             $productFilterData,
-            ProductsQuery::getOrderingModeFromArgument($argument),
+            ProductsQuery::getDefaultOrderingMode($argument),
```
vitek-rostislav: this means the ordering mode from the argument is never taken into account, is that ok?
I am the ordering is broken because of that

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProviderResolver.php
```diff
@@ -0,0 +1,75 @@
...
+        return $productSearchResultsProvidersOrderedByPriority;
+    }
+
+    /**
+     * @param string $serviceId
+     * @param int $priority
+     */
+    public function registerProductSearchResultsProvider(string $serviceId, int $priority): void
+    {
+        while (array_key_exists($priority, $this->productSearchResultsProvidersServiceIdByPriority)) {
```
vitek-rostislav: this feels quite unintuitive and may lead to unexpected behavior - maybe an error should be thrown when two services have the same priority.

One scenario that could be weird with my suggestion here - you have different services A and B, A is enabled for the 1st domain only, B is enabled for the second domain only. You set top priority for both but it won't work :thinking: Let's discuss it, please

packages/persoo/CONTRIBUTING.md
```diff
@@ -0,0 +1,10 @@
+# Contributing
+
+Thank you for your contributions to Shopsys Frontend API package.
```
vitek-rostislav:
```suggestion
Thank you for your contributions to Shopsys Persoo package.
```

packages/persoo/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+[![Downloads](https://img.shields.io/packagist/dt/shopsys/persoo.svg)](https://packagist.org/packages/shopsys/persoo)
+
+This bundle for [Shopsys Platform](https://www.shopsys.com) adds advanced search using [Persoo](https://www.persoo.ai).
+The bundle is dedicated for projects based on Shopsys Platform (i.e. created from [`shopsys/project-base`](https://github.com/shopsys/project-base)) exclusively.
+This repository is maintained by [shopsys/shopsys] monorepo, information about changes is in [monorepo CHANGELOG.md](https://github.com/shopsys/shopsys/blob/master/CHANGELOG.md).
+
+## Documentation
+
+[Documentation](https://docs.shopsys.com/en/latest/frontend-api/introduction-to-frontend-api/) can be found in Shopsys Platform Knowledge Base.
```
vitek-rostislav: is the link correct?

packages/persoo/README.md
```diff
@@ -0,0 +1,59 @@
...
+PERSOO_RECOMMENDATION_OFFER_ID // value provided by Persoo company
+PERSOO_RECOMMENDATION_LOCATION_ID // value provided by Persoo company
+PERSOO_SEARCH_ALGORITHM_ID // value provided by Persoo company
+PERSOO_SEARCH_OFFER_ID // value provided by Persoo company
+PERSOO_SEARCH_LOCATION_ID // value provided by Persoo company
+```
+
+## Contributing
+
+Thank you for your contributions to Shopsys Frontend API package.
```
vitek-rostislav:
```suggestion
Thank you for your contributions to Shopsys Persoo package.
```

packages/persoo/src/Component/Persoo/Client.php
```diff
@@ -0,0 +1,233 @@
...
+     */
+    public function getData(
+        string $query,
+        string $index,
+        string $action,
+        int $offset,
+        int $limit,
+        array $filter = [],
+        string $browserId = 'AAABQAg1a7sMJnpF6WlwIFuC',
+        string $requestingPage = 'http://127.0.0.1:8000/',
```
vitek-rostislav: the default values should be like this?

packages/persoo/src/Component/Persoo/Client.php
```diff
@@ -0,0 +1,233 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\PersooBundle\Component\Persoo;
+
+use Shopsys\PersooBundle\Component\Persoo\Exception\PersooIndexNotRecognizedException;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+class Client
```
vitek-rostislav: I would prefer naming it `PersooClient` as we are trying to avoid the general class names

packages/persoo/src/Component/Persoo/Client.php
```diff
@@ -0,0 +1,233 @@
...
+    {
+        if ($index === 'products') {
+            return 'id';
+        }
+
+        if ($index === 'categories') {
+            return 'categoryId';
+        }
+
+        throw new PersooIndexNotRecognizedException(sprintf('Persoo index "%s" is not recognized.', $index));
```
vitek-rostislav: what about articles?

packages/persoo/src/Component/Persoo/Result.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\PersooBundle\Component\Persoo;
+
+class Result
```
vitek-rostislav: rather `PersooResult`, please

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
```diff
@@ -0,0 +1,95 @@
...
+        $after = $argument['after'] ?? null;
+
+        $connectionBuilder = new ConnectionBuilder();
+        $offset = $connectionBuilder->getOffsetWithDefault($after, 0);
+        $persooFilter = $this->productFilterToPersooFilterMapper->mapForSearch($productFilterData, $this->domain);
+
+        $result = $this->client->getData($search, Client::PERSOO_INDEX_PRODUCTS, Client::PERSOO_ACTION_SEARCH, $offset, $limit, $persooFilter);
+
+        return $this->productConnectionFactory->createConnectionForAll(
+            function ($offset, $limit) use ($result) {
```
vitek-rostislav: you do not need to use offset and limit as the pagination is already done on the Persoo side, right? And the proper `endCursor` to offset is ensured by the `ConnectionBuilder` usage, right?

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
```diff
@@ -0,0 +1,95 @@
...
+                $filterQuery = $this->filterQueryFactory->createSellableProductsByProductIdsFilter($result->getIds());
```
vitek-rostislav: would not it be safer to pass the limit here?

project-base/app/config/packages/dev/monolog.yaml
```diff
@@ -9,3 +9,7 @@ monolog:
             path: "%shopsys.log_stream%"
             level: notice
             formatter: "monolog.formatter.line"
+        sentry:
```
vitek-rostislav, line 12: we use `sentry` handler in dev mode?

vitek-rostislav, line 12: we use `sentry` handler in dev mode?

vitek-rostislav (changes requested):

packages/frontend-api/src/DependencyInjection/RegisterProductsSearchResultsProvidersCompilerPass.php
TomasLudvik: Removed

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProvider.php
TomasLudvik: Fixed

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProviderResolver.php
TomasLudvik: I changed it, it will throw exception now

packages/persoo/src/Component/Persoo/Client.php
TomasLudvik: These are mandatory and in the current state, they are not available. These will be added to the frontend API in the next iterations.

packages/persoo/src/Component/Persoo/Client.php
TomasLudvik: I have no information on what name of the column should be used yet, so I could not implement it.

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
TomasLudvik: Yes

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
TomasLudvik: I have added it

project-base/app/config/packages/dev/monolog.yaml
TomasLudvik: Yes

TomasLudvik: Yes

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProvider.php
grossmannmartin: please consider extensibility. With the static method it is not possible to change the behavior in the project

packages/persoo/CONTRIBUTING.md
TomasLudvik: Nice catch

packages/persoo/README.md
TomasLudvik: Nice catch

packages/persoo/README.md
TomasLudvik: Nice catch

Makefile
TomasLudvik: Yeah, right

UPGRADE-14.0.md
TomasLudvik: I have changed it

project-base/app/config/services.yaml
TomasLudvik: I have removed ProductFilterData from App, but I have kept factory, as it creates data for CategorySeoMix that is still part of the project-base

packages/frontend-api/src/Model/Product/Connection/ProductConnection.php
TomasLudvik: I have added typehints

packages/frontend-api/src/DependencyInjection/RegisterProductsSearchResultsProvidersCompilerPass.php
TomasLudvik: Added

packages/persoo/src/Component/Persoo/Client.php
TomasLudvik: Renamed

packages/persoo/src/Component/Persoo/Result.php
TomasLudvik: Renamed

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductsSearchTest.php
TomasLudvik: I have updated code as we discussed it

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProvider.php
TomasLudvik: I have moved it into the trait

vitek-rostislav: hmm, but it is not possible to extend a trait, right? What if someone wants to change e.g. default ordering mode now? How can he overwrite `getDefaultOrderingMode`method?

vitek-rostislav: I believe it could be reverted into the previous state and just wrap the static call into a protected method in the classes where you use `ProductsQuery::getOrderingModeFromArgument($argument),`?

So you end up with `$this->getOrderingMode($argument)` and
```
protected function getOrderingMode($argument)
{
    return ProductsQuery::getOrderingModeFromArgument($argument);
}
```

and maybe this protected method can be static as well. Or maybe I misunderstood what @grossmannmartin meant?

vitek-rostislav: ah, my bad, this does not make sense. My original thought was creating new `ProductOrderingModeProvider` and use it via autowiring without static methods

grossmannmartin: When ProductsQuery is extended, the `ProductsQuery::getOrderingModeFromArgument($argument)` will always call the parent method in the shopsys/framework package context.

packages/frontend-api/src/Component/Arguments/ProductSearchPaginatorArgumentsBuilder.php
```diff
@@ -0,0 +1,27 @@
...
+     * @param array $config
+     * @return array
+     */
+    public function toMappingDefinition(array $config): array
+    {
+        $this->checkMandatoryFields($config);
+
+        $mappingDefinition = parent::toMappingDefinition($config);
+
+        unset($mappingDefinition['categorySlug'], $mappingDefinition['brandSlug'], $mappingDefinition['flagSlug']);
```
vitek-rostislav: IMHO this is too specific - if we added something into `ProductPaginatorArgumentsBuilder::toMappingDefinition`, it might break `ProductSearchPaginatorArgumentsBuilder`and no one might notice that. 

I suggest not extending one product builder from another but creating one more abstract class for product paginator builders that would extend the default abstract one, and then `ProductSearchPaginatorArgumentsBuilder` and `ProductPaginatorArgumentsBuilder` would extend that new abstract class, what do you think?

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProvider.php
```diff
@@ -0,0 +1,68 @@
...
+                    $offset,
+                    ProductListOrderingConfig::ORDER_BY_RELEVANCE,
+                    $productFilterData,
+                    $search,
+                );
+            },
+            $this->productFacade->getFilteredProductsCountOnCurrentDomain($productFilterData, $search),
+            $argument,
+            $productFilterData,
+            $this->getDefaultOrderingMode($argument),
```
vitek-rostislav: here should be also used `getOrderingModeFromArgument`

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchQuery.php
```diff
@@ -0,0 +1,49 @@
...
+     * @param \Shopsys\FrontendApiBundle\Model\Resolver\Products\Search\ProductSearchResultsProviderResolver $productSearchResultsProviderFacade
+     * @param \Shopsys\FrontendApiBundle\Model\Product\Filter\ProductFilterFacade $productFilterFacade
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     */
+    public function __construct(
+        protected readonly ProductSearchResultsProviderResolver $productSearchResultsProviderFacade,
+        protected readonly ProductFilterFacade $productFilterFacade,
+        protected readonly Domain $domain,
+    ) {
+        $this->productSearchResultsProvider = $this->productSearchResultsProviderFacade->getProductsSearchResultsProviderByDomainId($this->domain->getId());
```
vitek-rostislav: it is not a good idea to have "logic" in the constructor - it is considered a code smell as it might have side effects (here e.g. `$this->domain->getId()` might fail on `NoDomainSelectedException` etc.)

Could you please refactor it? You can get rid of the `$productSearchResultsProvider` property completely and just call 
```
$productSearchResultsProvider = $this->productSearchResultsProviderFacade->getProductsSearchResultsProviderByDomainId($this->domain->getId());
```
directly in `productsSearchQuery` method

vitek-rostislav (changes requested):
Hi, I have added just a few notes, please check them out.

Also, after the refactorings, there are now a few static calls of non-static methods in `ProductsQuery` in `project-base` so please take a look at them.

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProvider.php
TomasLudvik: I have introduced new service for this

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchResultsProvider.php
TomasLudvik: Fixed

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchQuery.php
TomasLudvik: Changed

packages/frontend-api/src/Component/Arguments/ProductSearchPaginatorArgumentsBuilder.php
TomasLudvik: I have introduced new Abstract class as requested

vitek-rostislav: thanks, so you can remove the line :slightly_smiling_face: 

```
unset($mappingDefinition['categorySlug'], $mappingDefinition['brandSlug'], $mappingDefinition['flagSlug']);
```

packages/frontend-api/src/Model/Resolver/Products/Search/ProductSearchQuery.php
```diff
@@ -0,0 +1,45 @@
...
+
+class ProductSearchQuery extends AbstractQuery
+{
+    /**
+     * @param \Shopsys\FrontendApiBundle\Model\Resolver\Products\Search\ProductSearchResultsProviderResolver $productSearchResultsProviderFacade
+     * @param \Shopsys\FrontendApiBundle\Model\Product\Filter\ProductFilterFacade $productFilterFacade
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     */
+    public function __construct(
+        protected readonly ProductSearchResultsProviderResolver $productSearchResultsProviderFacade,
```
vitek-rostislav: class type name vs property name mismatch, could you please rename the property to `$productSearchResultsProviderResolver`?

vitek-rostislav (approved):

UPGRADE-14.0.md
```diff
@@ -1036,6 +1036,14 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
         )
     ```
 
+    -   [features moved](#movement-of-features-from-project-base-to-packages) to the `framework` package:
+
+        -   `ProductFilterData` and part of related logic
```
vitek-rostislav, line 1131: I would add one more thing:

- `ClassExtensionRegistry::getOtherClassesExtensionMap()` logic was moved to its parent class

vitek-rostislav (approved):

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
```diff
@@ -77,21 +77,47 @@ public function getProductsSearchResults(
...
             $this->productOrderingModeProvider->getOrderingModeFromArgument($argument),
         );
     }
+
+    /**
+     * @param int $getOffsetWithDefault
+     * @param int $limit
+     * @return int
+     */
+    protected function getPageFromOffsetAndLimit(int $getOffsetWithDefault, int $limit): int
```
vitek-rostislav: `$getOffsetWithDefault` name does not make sense to me in the context of the method, IMHO it should be just `$offset`

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
```diff
@@ -77,21 +77,47 @@ public function getProductsSearchResults(
...
+    {
+        return (int)ceil($getOffsetWithDefault / $limit);
+    }
+
+    /**
+     * @param array $sortedProducts
+     * @param int $offset
+     * @return array
+     */
+    protected function appendEmptyArrayValuesForCorrectFunctionOfConnectionBuilder(
```
vitek-rostislav: this is really disturbing :thinking: I do not understand it, could we please discuss it?

vitek-rostislav (changes requested):

vitek-rostislav (approved):

PR #2990, opened 2024-01-16
[SSP-2071] add possibility to change SF error verbosity for development

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In development (or in special environments) we want to be able to see the actual errors happening, not the non-descriptive message "unknown error" or similar. This PR improves that, while also improving handling of 500 errors for friendly URL pages.
|New feature| Yes
|Fixes issues| ... 

----
  - https://sh-ssp-2071-sf-error-verbosity-for-development.odin.shopsys.cloud
  - https://cz.sh-ssp-2071-sf-error-verbosity-for-development.odin.shopsys.cloud


project-base/storefront/helpers/errors/handleServerSideErrorResponseForFriendlyUrls.ts
```diff
@@ -1,12 +1,20 @@
...
+const { publicRuntimeConfig } = getConfig();
 
 export const handleServerSideErrorResponseForFriendlyUrls = (
     graphQLErrors: GraphQLError[] | undefined,
     data: unknown,
     res: ServerResponse<IncomingMessage>,
 ) => {
     if (graphQLErrors?.[0]?.extensions.code === 500) {
+        if (publicRuntimeConfig.errorMessageVerbosity === 'devel') {
+            throw new Error(JSON.stringify(mapGraphqlErrorForDevelopment(graphQLErrors[0])));
```
tvikito, line 13: Would it makes sense to you prepare this also for a case when we have more error messages?

project-base/storefront/.env
```diff
@@ -10,3 +10,4 @@ REDIS_HOST=redis
 REDIS_PREFIX='app'
 GTM_ID=
 GRAPHQL_REDIS_CACHE=1
+ERROR_MESSAGE_VERBOSITY='user'
```
tvikito: Hmm, I wonder how somebody debugging this will know about other possible values? 🧐 I know it's written in docs, but we know how many people read docs. So what about make it more obvious and rather use something like

`ERROR_DEBUGGING=1` 

and then run this connected logic if ERROR_DEBUGGING === '1'?

project-base/storefront/components/Pages/ErrorPage/Error500Content.tsx
```diff
@@ -29,9 +34,13 @@ export const Error500Content: FC = () => {
                 <ErrorPage isWithoutImage>
                     <ErrorPageTextHeading>{t('Something went wrong.')}</ErrorPageTextHeading>
                     <ErrorPageTextMain>{t('Please try again later or contact us.')}</ErrorPageTextMain>
-
                     <ErrorPageButtonLink href="/">{t('Back to shop')}</ErrorPageButtonLink>
                 </ErrorPage>
+                {err && (
+                    <div className="my-3 flex justify-center">
+                        <CopyTextBlock showCopyTextButton={false} textToCopy={err} />
```
tvikito: This `showCopyTextButton` prop is confusing me since here is already used CopyTextBlock which should by name already include this functionality. 

My proposal would be to put condition here and either show `CopyTextBlock` which will allow user to copy or simple div.

project-base/storefront/helpers/toasts.tsx
```diff
@@ -1,20 +1,31 @@
...
-            toastId: message,
-        });
-    }
-    if (type === 'info') {
+        if (publicRuntimeConfig.errorMessageVerbosity === 'user') {
+            toast.error(() => <span dangerouslySetInnerHTML={{ __html: message }} data-testid="toast-error" />, {
+                toastId: message,
+            });
+        } else {
+            toast.error(() => <CopyTextBlock data-testid="toast-error" textToCopy={message} />, {
```
tvikito: no need for `data-testid="toast-error"`

project-base/storefront/helpers/toasts.tsx
```diff
@@ -1,20 +1,31 @@
...
+const { publicRuntimeConfig } = getConfig();
+
 const showMessage = (message: string, type: 'info' | 'error' | 'success'): void => {
     if (type === 'error') {
-        toast.error(() => <span dangerouslySetInnerHTML={{ __html: message }} data-testid="toast-error" />, {
-            toastId: message,
-        });
-    }
-    if (type === 'info') {
+        if (publicRuntimeConfig.errorMessageVerbosity === 'user') {
```
tvikito: Since we have a lot of places where this `errorMessageVerbosity` check is used, what about add new exported constant like `isWithErrorDebugging`?

project-base/storefront/helpers/errors/mapGraphqlErrorForDevelopment.ts
```diff
@@ -0,0 +1,11 @@
+import { GraphQLError } from 'graphql';
+
+export const mapGraphqlErrorForDevelopment = (graphqlError: GraphQLError) => ({
+    message: graphqlError.message,
+    location: graphqlError.locations,
+    path: graphqlError.path,
+    extensions: {
+        ...graphqlError.extensions,
+        trace: Array.isArray(graphqlError.extensions.trace) ? graphqlError.extensions.trace.slice(0, 3) : [],
```
tvikito, line 9: I wonder what are other possible types of `graphqlError.extensions.trace`? Cannot this be `graphqlError.extensions.trace?.slice(0, 3) || []`?

project-base/storefront/components/Basic/CopyTextBlock/CopyTextBlock.tsx
```diff
@@ -0,0 +1,36 @@
...
+            setIsCopySuccess(false);
+        }
+    };
+
+    return (
+        <div className="rounded bg-greyLighter">
+            <div className="flex justify-end rounded bg-greyDark px-3 py-2">
+                {showCopyTextButton && (
+                    <button className="text-white" onClick={handleCopy}>
+                        {isCopySuccess ? 'Copied' : 'Copy Text'}
```
tvikito: Hmm, in code above I see that this `isCopySuccess` is supposed to be used for error handling but here is used for differentiate between state if used have copied or not. 

If promise will fall into catch user is not notified and he do infinite number of click without getting info what is wrong. That's why I would suggest to rather have three states `undefined`, `true` and `false` and decide what to display according to those (copied, copy text, error while copy). Or think about different way of handling like display toast message and don't change button text.

project-base/storefront/urql/errorExchange.ts
```diff
@@ -23,46 +29,69 @@ export const getErrorExchange =
...
+    error.graphQLErrors
+        .map((graphqlError) => mapGraphqlErrorForDevelopment(graphqlError))
+        .forEach((simplifiedGraphqlError) => showErrorMessage(JSON.stringify(simplifiedGraphqlError)));
+};
 
-                    const isCartError = operation.query === CartQueryDocumentApi;
-                    if (isCartError) {
-                        handleCartError(parsedErrors);
+const handleErrorMessagesForUsers = (error: CombinedError, t: Translate | undefined, operation: Operation) => {
+    const parsedErrors = t ? getUserFriendlyErrors(error, t) : undefined;
```
tvikito: Hmm, if you go higher in the tree where is this function called I see that `t` is always being sent. Can we please adjust this so we don't have unnecessary conditions like this?

project-base/storefront/urql/errorExchange.ts
```diff
@@ -23,46 +29,69 @@ export const getErrorExchange =
...
-                        return;
-                    }
+    if (parsedErrors.userError) {
+        logException(parsedErrors.userError);
+    }
 
-                    if (!parsedErrors.applicationError) {
-                        return;
-                    }
+    const isCartError = operation.query === CartQueryDocumentApi;
```
tvikito: According to my previous comment this won't probably be the case anyway, but, I understand idea of not having variable defined before it's used because function can return earlier, but I don't see any benefit here. It's just simple boolean (not some heavy function call) so it get's readability down for basically no benefits. Let's rather define it with other variables on top. What do you think?

project-base/storefront/urql/errorExchange.ts
```diff
@@ -77,3 +106,11 @@ const handleCartError = ({ userError, applicationError }: ParsedErrors) => {
...
     }
 };
+
+const hasFriendlyUrlQueryFailedWith500 = (operation: Operation, error: CombinedError) =>
+    error.graphQLErrors.some((graphqlError) => graphqlError.extensions.code === 500) &&
+    operation.query.definitions.some(
+        (definition) =>
+            definition.kind === Kind.OPERATION_DEFINITION &&
+            definition.directives?.some((directiveDefinition) => directiveDefinition.name.value === 'friendlyUrl'),
+    );
```
tvikito: Maybe you can consider this way. Just idea.

```suggestion
    error.graphQLErrors.some(({extensions}) => extensions.code === 500) &&
    operation.query.definitions.some(
        ({kind, directives}) =>
            kind === Kind.OPERATION_DEFINITION &&
            directives?.some(({name}) => name.value === 'friendlyUrl'),
    );
```

UPGRADE-14.0.md
```diff
@@ -991,3 +991,8 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   fix Add to cart popup product link type
     -   fix Bestsellers product link type
 -   add scroll to cart order personal pick up place select modal ([#2979](https://github.com/shopsys/shopsys/pull/2979))
+
+-   add possibility to change SF error verbosity for developmen ([#2990](https://github.com/shopsys/shopsys/pull/2990))
```
grossmannmartin: typo

project-base/storefront/.env
sebaholesz: That sounds very reasonable 😆 I will do that.

project-base/storefront/components/Pages/ErrorPage/Error500Content.tsx
sebaholesz: Smart, let's do that. Just so you know, the reason why the button is not visible in some cases, is because Next.js does not allow button onClicks on the 500 page.

project-base/storefront/helpers/toasts.tsx
sebaholesz: True, removed.

project-base/storefront/helpers/toasts.tsx
sebaholesz: Done

project-base/storefront/helpers/errors/mapGraphqlErrorForDevelopment.ts
sebaholesz: Its unknown 🙃 So I think this is the best way to check it

project-base/storefront/components/Basic/CopyTextBlock/CopyTextBlock.tsx
sebaholesz: To make it simple, I'll just change the text 👍

project-base/storefront/urql/errorExchange.ts
sebaholesz: We always send it, but it can be undefined. However, I still moved the condition one level higher

project-base/storefront/urql/errorExchange.ts
sebaholesz: Im not sure if I understand you 100%, but if the idea is to have the variable defined on top, I can move it. I am more inclined towards defining the variable as close to where it is used, but in this particular case, it's a small difference.

project-base/storefront/urql/errorExchange.ts
sebaholesz: I changed the first one, as that was possible, but the second one was kept, as `directives` only exists on it if `kind === Kind.OPERATION_DEFINITION` is true. Because of that, I first need to check it.

project-base/storefront/helpers/errors/handleServerSideErrorResponseForFriendlyUrls.ts
sebaholesz: I changed the condition, but keep in mind, that after a discussion with a BE dev we found out there should always be just one of them.

project-base/storefront/urql/errorExchange.ts
tvikito: I don't see it can be undefined 🤔Maybe you haven't gone deep enough. If I go to place where this function (`getErrorExchange`) is used, I found only one place 

image

here it is marked that it can be undefined, ok. But If I see where this function (`getUrqlExchanges`) is used I found again only one place

image

where it is marked as defined. So there is no reason mark it as undefined. Let's refactor this please.

sebaholesz: Wow, you are right. Thanks for being so persistent. I removed it now.

tvikito (approved):

PR #2999, opened 2024-01-24
[shopsys] added personalization arguments to search

tvikito:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| We have added a new way to personalize user's search and product suggestions via Persoo. For its personalization to work we added new arguments in this PR.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tl-add-personalization-to-search.odin.shopsys.cloud
  - https://cz.tl-add-personalization-to-search.odin.shopsys.cloud


packages/frontend-api/src/Resources/config/graphql-types/QueryDecorator.types.yaml
```diff
@@ -14,6 +14,12 @@ QueryDecorator:
                 args:
                     search:
                         type: "String!"
+                    requestingPage:
```
vitek-rostislav, line 17: each search query now uses three input arguments - `search`, `requestingPage`, and `userIdentifier` - IMHO it would make sense to wrap them into an input object so we do not have to repeat ourselves.

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
```diff
@@ -75,12 +76,31 @@ public function getProductsSearchResults(
         $search = $argument['search'] ?? '';
         $limit = $argument['first'] ?? 0;
         $after = $argument['after'] ?? null;
+        $requestingPage = $argument['requestingPage'] ?? '';
```
vitek-rostislav, line 79: I believe the empty string can be set as a default value in `QueryDecorator.types.yaml`, would not it be better there?

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
```diff
@@ -75,12 +76,31 @@ public function getProductsSearchResults(
...
+        $userIdentifier = $argument['userIdentifier'] ?? '';
+
+        if (filter_var($requestingPage, FILTER_VALIDATE_URL) === false) {
+            throw new InvalidArgumentUserError('Provided argument \'requestingPage\' is not in correct URL format.');
```
vitek-rostislav, line 83: should not this be implemented as a validation (i.e. using constraint and validator)? I believe I can live without that but the design would be better with the validator IMHO.

One way or another, could you please add tests for this behavior?

(the same comment applies to `$userIdentifier` bellow)

vitek-rostislav (changes requested):

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
TomasLudvik: We decided we do not want to follow this path now

packages/persoo/src/Model/Product/ProductSearchResultsProvider.php
TomasLudvik: We decided we do not want to follow this path now

PR #3001, opened 2024-01-25
[shopsys] removed entity's and dataobject's property typehints and typehints from getter/setter

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| To ease extensibility, the Doctrine columns and association should not have type hinted property and their getter/setter method should not have a return type/property type. Otherwise, it's impossible, for example, to make the property nullable. This PR adds a PHPStan Rule to disallow type hints and return types on such elements and remove them where they were.
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-no-typehint-entities.odin.shopsys.cloud
  - https://cz.mg-no-typehint-entities.odin.shopsys.cloud


packages/framework/src/Component/Cron/CronModule.php
```diff
@@ -121,7 +121,7 @@ public function suspend()
     /**
      * @return bool
      */
-    public function isEnabled(): bool
+    public function isEnabled()
```
TomasLudvik: I think that `is` and `has` methods could have an exception as this naming is only used for the booleans.

packages/coding-standards/src/Phpstan/OrmPropertyGetterHasNoTypehintRule.php
```diff
@@ -0,0 +1,113 @@
...
+            return [];
+        }
+
+        if ($node->getReturnType() === null) {
+            return [];
+        }
+
+        $methodName = $node->name->toString();
+
+        if (!str_starts_with($methodName, 'get')) {
```
TomasLudvik: What about setters? I think that same rule applies there.

TomasLudvik (changes requested):

phpstan.neon
```diff
@@ -72,3 +72,6 @@ includes:
     - vendor/phpstan/phpstan-symfony/extension.neon
     - vendor/phpstan/phpstan-symfony/rules.neon
     - %currentWorkingDirectory%/packages/coding-standards/extension.neon
+rules:
+    -   \Shopsys\CodingStandards\Phpstan\OrmPropertyGetterHasNoTypehintRule
+    -   \Shopsys\CodingStandards\Phpstan\OrmPropertyHasNoTypehintRule
```
TomasLudvik: What about data objects? Same applies there

grossmannmartin: I have no idea how to safely determine a data object belonging to the entity...

packages/framework/src/Component/Cron/CronModule.php
grossmannmartin: boolean types are now ignored

packages/coding-standards/src/Phpstan/OrmPropertyGetterHasNoTypehintRule.php
grossmannmartin: added setter rule and applied

phpstan.neon
grossmannmartin: added data object rule and applied

packages/coding-standards/src/Phpstan/OrmPropertyGetterAndSetterHasNoTypehintRule.php
```diff
@@ -110,4 +172,20 @@ private function isMethodInAncestor(string $methodName, Scope $scope): bool
...
+     */
+    private function isParameterDataObjectOfCurrentEntity(string $parameterType, Scope $scope): bool
+    {
+        $currentClassName = $scope->getClassReflection()?->getName();
+
+        if ($currentClassName === null) {
+            return false;
+        }
+
+        return str_ends_with($parameterType, 'Data') && str_starts_with($parameterType, $currentClassName);
```
TomasLudvik:
```suggestion
        return str_starts_with($parameterType, $currentClassName . 'Data');
```

Your way it would pass for Product entity when ProductVideoData would be provided.

packages/coding-standards/src/Phpstan/OrmPropertyGetterAndSetterHasNoTypehintRule.php
```diff
@@ -70,6 +89,49 @@ public function processNode(Node $node, Scope $scope): array
...
+     */
+    private function checkSetter(
+        ClassMethod $node,
+        string $methodName,
+        Scope $scope,
+        PropertyReflection $propertyReflection,
+    ): array {
+        if (count($node->getParams()) !== 1) {
+            return [];
+        }
```
TomasLudvik, line 107: Why do you want to skip it for setters with two and more parameters?

packages/framework/src/Model/Feed/FeedModule.php
```diff
@@ -72,7 +72,7 @@ public function unschedule(): void
     /**
      * @return bool
      */
-    public function isScheduled(): bool
+    public function isScheduled()
```
TomasLudvik: pleaser return typehint here

packages/framework/src/Model/Store/Store.php
```diff
@@ -212,95 +210,95 @@ protected function setDomains(StoreData $storeData): void
...
+    public function getStock()
     {
         return $this->stock;
     }
 
     /**
      * @return bool
      */
-    public function isDefault(): bool
+    public function isDefault()
```
TomasLudvik: please return typehint here

TomasLudvik (changes requested):

packages/framework/src/Model/Store/Store.php
grossmannmartin: fixed

packages/framework/src/Model/Feed/FeedModule.php
grossmannmartin: fixed

packages/coding-standards/src/Phpstan/OrmPropertyGetterAndSetterHasNoTypehintRule.php
grossmannmartin: good catch, thanks! fixed in 5b97590957

packages/coding-standards/src/Phpstan/OrmPropertyGetterAndSetterHasNoTypehintRule.php
grossmannmartin: Because then it's not just a simple setter and I don't want to make any assumptions about such code

TomasLudvik (approved):

pk16011990 (approved):

PR #3004, opened 2024-01-27
[shopsys] all relevant entities are now created with factory

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Added PhpStan rule to ensure that each relevant entity has its own factory that uses entity name resolver. This rule is applied.
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-fix-extensibility.odin.shopsys.cloud
  - https://cz.mg-fix-extensibility.odin.shopsys.cloud

### DIFF
--- a/packages/coding-standards/src/Phpstan/EntityShouldHaveFactoryRule.php
+++ b/packages/coding-standards/src/Phpstan/EntityShouldHaveFactoryRule.php
@@ -1,0 +1,112 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\Phpstan;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\InClassNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use ReflectionClass;
+
+class EntityShouldHaveFactoryRule implements Rule
+{
+    private const CHECKED_NAMESPACE = 'Shopsys\\';
+    private const IGNORED_SUFFIXES = [
+        'Domain',
+        'Translation',
+    ];
+
+    /**
+     * @return string
+     */
+    public function getNodeType(): string
+    {
+        return InClassNode::class;
+    }
+
+    /**
+     * @param \PhpParser\Node $node
+     * @param \PHPStan\Analyser\Scope $scope
+     * @return array
+     */
+    public function processNode(Node $node, Scope $scope): array
+    {
+        if (!$node instanceof InClassNode) {
+            return [];
+        }
+
+        $entityClassName = $scope->getClassReflection()?->getName() ?? '';
+        $factoryClassName = $entityClassName . 'Factory';
+
+        if (!str_starts_with($entityClassName, self::CHECKED_NAMESPACE)) {
+            return [];
+        }
+
+        if (!$this->isCheckedEntity($entityClassName, $node)) {
+            return [];
+        }
+
+        if (class_exists($factoryClassName)) {
+            if ($this->factoryUsesEntityNameResolver($factoryClassName)) {
+                return [];
+            }
+
+            return [
+                RuleErrorBuilder::message(sprintf(
+                    'Factory %s do not use entity name resolver',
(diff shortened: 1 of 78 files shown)


packages/coding-standards/src/Phpstan/EntityShouldHaveFactoryRule.php
```diff
@@ -0,0 +1,112 @@
...
+            ))->build(),
+        ];
+    }
+
+    /**
+     * @param string $className
+     * @param \PHPStan\Node\InClassNode $node
+     * @return bool
+     */
+    private function isCheckedEntity(string $className, InClassNode $node): bool
```
vitek-rostislav, line 78: are we 100% no one will need to extend the class? :smile: (I can see other classes with private methods in the coding standards package so maybe this discussion was already resolved :thinking: )

packages/framework/src/Model/Category/CategoryParameterFactory.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Category\Category $category
+     * @param \Shopsys\FrameworkBundle\Model\Product\Parameter\Parameter $parameter
+     * @param bool $collapsed
+     * @param int $position
+     * @return \Shopsys\FrameworkBundle\Model\Category\CategoryParameter
+     */
+    public function create(Category $category, Parameter $parameter, bool $collapsed, int $position): CategoryParameter
+    {
+        $classData = $this->entityNameResolver->resolve(CategoryParameter::class);
```
vitek-rostislav: it is actually `$className` or `$class` IMHO, but I see we already use it like this in many existing factories :frowning_face:

packages/framework/src/Model/Store/StoreFactory.php
```diff
@@ -4,14 +4,26 @@
...
+    }
+
     /**
      * @param \Shopsys\FrameworkBundle\Model\Store\StoreData $storeData
      * @return \Shopsys\FrameworkBundle\Model\Store\Store
      */
     public function create(StoreData $storeData): Store
     {
-        return new Store($storeData);
+        $classData = $this->entityNameResolver->resolve(Store::class);
```
vitek-rostislav: if you think my previous comment makes sense, I would appreciate if you renamed the local variables in all the factories :pray:

packages/framework/src/Model/Product/List/ProductListItemFactory.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\List\ProductList $productList
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @return \Shopsys\FrameworkBundle\Model\Product\List\ProductListItem
+     */
+    public function create(ProductList $productList, Product $product): ProductListItem
+    {
+        $productListClassName = $this->entityNameResolver->resolve(ProductListItem::class);
```
vitek-rostislav, line 27: this is what I like :grin:

vitek-rostislav (approved):

packages/coding-standards/src/Phpstan/EntityShouldHaveFactoryRule.php
grossmannmartin: As this rule is applied only for monorepo, I believe it's not necessary to ease extensibility.

packages/framework/src/Model/Category/CategoryParameterFactory.php
grossmannmartin: I was surprised too with the naming, but then I just went with the flow 🤷

grossmannmartin: renamed to $entityClassName everywhere, better?

vitek-rostislav: yes yes, thank you! :slightly_smiling_face:

vitek-rostislav (approved):

PR #3023, opened 2024-02-12
[SSP-1669] cypress refactoring

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR combines multiple cypress improvements, such as visual regression tests, test splitting and atomization, implementation of typescript, removal of unnecessary test IDs, etc.
|New feature| Yes
|Fixes issues| ... 

----
  - https://sh-cypress-refactoring.odin.shopsys.cloud
  - https://cz.sh-cypress-refactoring.odin.shopsys.cloud


UPGRADE-14.0.md
```diff
@@ -1579,5 +1579,36 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
...
 -   added more verbose error messages when using logException on SF ([#3018](https://github.com/shopsys/shopsys/pull/3018))
+
     -   messages logged to sentry now contain more context
     -   when adding error logs using `logException`, make sure you always provide as much context as possible
+
+-   refactored cypress tests
+
+    -   rewrote cypress tests to typescript for better static control and easier working with the code
+        -   added type annotations, changed file type extensions, so yuo should do the same to all your existing files
+        -   added tsconfig specifically for cypress and excluded cypress files from the root tsconfig (except for the data test IDs file)
```
TomasLudvik: UPGRADE notes are not supposed to describe what has changed, but only provide instructions for easy and fast upgrade

docker/conf/docker-compose-mac.yml.dist
```diff
@@ -43,13 +43,36 @@ services:
         ports:
             - "3000:3000"
 
+    # This service is used exclusively for running Cypress tests
```
TomasLudvik: Do we really need this right above `storefront-cypress`? :)

project-base/Makefile
```diff
@@ -12,8 +12,21 @@ generate-schema-native:
 	cd storefront; npm run gql
 	rm -rf storefront/schema.graphql
 
-run-acceptance-tests:
-	docker compose exec php-fpm php phing -D production.confirm.action=y -D change.environment=acc environment-change
+define run_acceptance_tests
+	docker compose exec php-fpm php phing -D production.confirm.action=y -D change.environment=test environment-change
```
TomasLudvik, line 16: Why we do not run acceptance tests in acceptance environment?

TomasLudvik (changes requested):

project-base/Makefile
grossmannmartin: In the acceptance environment backend of the application overwrites its URL to webserver:8080, while the storefront has still the domain set to 127.0.0.1:8000. This causes improper functioning.

UPGRADE-14.0.md
sebaholesz: I have tried to follow your suggestion more closely, though I have left some of the change descriptions to give everybody the needed context

docker/conf/docker-compose-mac.yml.dist
sebaholesz: Probably not... I think the name is descriptive enough... I removed it.

TomasLudvik (approved):

project-base/storefront/components/Basic/Loader/LoaderWithOverlay.tsx
```diff
@@ -1,7 +1,11 @@
 import { SpinnerIcon } from 'components/Basic/Icon/IconsSvg';
+import { DataTestIds } from 'cypress/dataTestIds';
 
 export const LoaderWithOverlay: FC = ({ className }) => (
-    <div className="absolute inset-0 z-overlay flex h-full w-full items-center justify-center bg-greyLighter opacity-50">
+    <div
+        className="absolute inset-0 z-overlay flex h-full w-full items-center justify-center bg-greyLighter opacity-50"
+        data-testid={DataTestIds.loader_overlay}
```
tvikito: Since we do this ground up refactoring I have also major idea. What about to replace this prop `data-testid` with `dataTestId` so we are matching our (React recommended) props?

sebaholesz: Unfortunately, even though this was our original plan, this causes both a TS error and a runtime error:
image
image

sebaholesz: If you have an idea how to circumvent this, we can try it, but otherwise we have no other way.

tvikito: I see here two options here:

1) As it is suggested in error message = we can use `datatestid` for DOM property and the same also for our FC props
2) Ignore this property in ESlint rules

For both I would suggest to add declaration for React HTMLAttributes so we know the required type. I have added one WIP commit to this branch with changes for number 2. I will let you decide which one to use. If you are not sure we can discuss the best approach on FE meeting with other guys.

sebaholesz: We discussed and resolved this on a call. We renamed data test IDs to `tid` (everywhere)

PR #3032, opened 2024-02-19
[SSP-2067] [SSP-2139] order process fixes

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We have had many issues with access logic for order pages, as it often broke and did not work correctly. This PR fixes the scenarios where a user tries to access transport and payment, or contact information pages without completing previous necessary steps. Furthermore, it also fixes other issues, mostly related to the order process. You can read commit messages or UPGRADE.md for more details.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-ssp-2067-2139-order-page-access-fix.odin.shopsys.cloud
  - https://cz.sh-ssp-2067-2139-order-page-access-fix.odin.shopsys.cloud


project-base/storefront/hooks/cart/useOrderPagesAccess.ts
```diff
@@ -0,0 +1,36 @@
...
+        } else if (page === 'contact-information' && (!cart.transport || !cart.payment)) {
+            router.replace(transportAndPaymentUrl);
+        } else {
+            setCanContentBeDisplayed(true);
+        }
+    };
+
+    useEffect(() => {
+        handleOrderPagesAccess();
+    }, [cart, isFetching]);
```
tvikito: I understand the sentiment behind having cleaner useEffect, but I am not sure if this separation is beneficial (maybe even correct) in this place. The part which I am not sure about is that function is using data which are not available during SSR. With combination that this function is used only on one place I am not sure if it wouldn't be better have it directly in useEffect. Just a hint, I will let you decide.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
```diff
@@ -63,19 +65,15 @@ export const TransportAndPaymentContent: FC = () => {
         router.push(contactInformationUrl);
     };
 
-    useEffect(() => {
-        if (cart !== undefined && !cart?.items.length) {
-            router.replace(cartUrl);
-        }
-    }, [cart?.items]);
+    const canContentBeDisplayed = useOrderPagesAccess('transport-and-payment');
```
tvikito, line 68: Can we follow marking boolean values as `isWithPageAccess`?

project-base/storefront/cypress/tsconfig.json
```diff
@@ -6,9 +6,10 @@
         "baseUrl": ".",
         "skipLibCheck": true,
         "strict": true,
+        "jsx": "preserve",
         "forceConsistentCasingInFileNames": true,
         "incremental": true,
         "esModuleInterop": true
     },
-    "include": ["**/*.ts", "**/*.cy.ts"]
+    "include": ["**/*.ts", "**/*.cy.ts", "../graphql/generated/index.tsx"]
```
tvikito: What about have included both `ts` and `tsx` like we have in storefront folder `"**/*.ts", "**/*.tsx"`?

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
```diff
@@ -0,0 +1,36 @@
...
+import { getInternationalizedStaticUrls } from 'helpers/getInternationalizedStaticUrls';
+import { useDomainConfig } from 'hooks/useDomainConfig';
+import { useRouter } from 'next/router';
+import { useEffect, useState } from 'react';
+
+export const useOrderPagesAccess = (page: 'transport-and-payment' | 'contact-information') => {
+    const router = useRouter();
+    const { cart, isFetching } = useCurrentCart();
+    const { url } = useDomainConfig();
+    const [canContentBeDisplayed, setCanContentBeDisplayed] = useState(false);
```
tvikito: What about naming which would follow the purpose of the hook like `isWithPageAccess`?

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
```diff
@@ -0,0 +1,36 @@
...
+        } else {
+            setCanContentBeDisplayed(true);
+        }
+    };
+
+    useEffect(() => {
+        handleOrderPagesAccess();
+    }, [cart, isFetching]);
+
+    return canContentBeDisplayed;
```
tvikito, line 31: Hmm, if I understand the logic right in case cart is `undefined `it will display skeleton even if the is no fetching happening. I think we should implement 3 possible values return -  boolean and `undefined` and then distinguish proper state. If return `undefined` that would mean cart is still fetching, boolean would only return when fetching is resolved. Otherwise we are not able distinguish between fetching and without access state.

project-base/storefront/cypress/tsconfig.json
sebaholesz: Im not sure if that makes sense? Are we expecting more tsx files here?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: I do not agree with it. It is not any convention we follow, nor it is something strictly followed by any of the major technologies we use. For me the point is for the code to be as readable as possible. Ideally follow natural language.

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
sebaholesz: See my comment above.

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
sebaholesz: Sure. I moved it inside the hook 😇

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
sebaholesz: alright, but the skeleton would still be shown in case of a falsy value? I guess I have no problem with that, if it helps anything.

tvikito: Nope, skeleton would be shown only in case there is some fetching happening. This is simply missing handling for this case of undefined value. As user I don't want to see infinite skeleton in case action failed.

project-base/storefront/cypress/tsconfig.json
tvikito: In case there are some new files are you sure everyone will know about adding this file also to this config? I don't see reason why to add files individually.

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
tvikito: I think we have already agreed with other FE guys that we want to mark those boolean values with this pattern, or...? 🤔 I personally don't see it any difficult to read the code I rather see it as improved because I can quickly recognize boolean value without thinking about it.

If you want to have code better readable I don't see point of naming variable as `canContentBeDisplayed` if you use directly return from `useOrderPagesAccess`. Wouldn't it be rather better readable if you in JSX know why some element shouldn't be displayed? So why not at least variable `canContentBeAccessed` or as I wrote previously something like `IsWithAccess`?

project-base/storefront/cypress/tsconfig.json
sebaholesz: That is a fair point. I wanted to limit the leak of files from storefront to cypress, but if we do not care about that, then I have no problem with it.

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
sebaholesz: Could you describe this more closely? I still do not understand what should happen in case of undefined. Not display anything? Or?

project-base/storefront/components/Pages/Order/TransportAndPayment/TransportAndPaymentContent.tsx
sebaholesz: I do not remember any such agreement, and neither has Ivan. Regarding Tom's and Michal's opinion, I was not able to get it in time for this answer. Nevertheless, the opinion that is shared by Ivan and myself, also the BE part of our team, and the industry (check react or next.js code) is to allow all prefixes such as has, can, should, are...

sebaholesz: Furthermore, isWithAccess makes no sense to me. I do not even know what that means.

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
sebaholesz: As described above. isWithPageAccess makes no sense to me. I do not see any benefit of such name, only negatives.

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
tvikito: You are handling this logic. I am not responsible for this task neither about business decisions. I am just telling you about third case which you don't handle yet. But I guess in the end we should end up with some sort of message. Either specific or general error/fetching failed message.

project-base/storefront/cypress/tsconfig.json
```diff
@@ -11,5 +11,5 @@
         "incremental": true,
         "esModuleInterop": true
     },
-    "include": ["**/*.ts", "**/*.cy.ts", "../graphql/generated/index.tsx"]
+    "include": ["**/*.ts", "**/*.cy.ts", "../**/*.ts", "../**/*.tsx"]
```
tvikito: "include": ["**/*.ts", "../**/*.tsx"]

project-base/storefront/cypress/tsconfig.json
tvikito: Not yet there see my comment

project-base/storefront/hooks/cart/useOrderPagesAccess.ts
sebaholesz: We agreed on a call that this is okay

project-base/storefront/cypress/tsconfig.json
sebaholesz: I changed it to:

`"include": ["**/*.ts", "../**/*.tsx", "../**/*.ts"]`

we need `"../**/*.ts"` to also include SF files (we need to go one level up)

project-base/storefront/urql/cache/updates.ts
```diff
@@ -77,7 +77,7 @@ export const cacheUpdates: UpdatesConfig = {
             _args: ChangeTransportInCartMutationVariablesApi,
             cache,
         ) {
-            if (result.ChangeTransportInCart?.uuid) {
+            if (result.ChangeTransportInCart?.__typename) {
```
tvikito: Is it necessary to check for the `__typename`? Cannot we have `if (result.ChangeTransportInCart) {`?

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
```diff
@@ -317,12 +317,12 @@ export const useLoadTransportAndPaymentFromLastOrder = (
...
 
     useEffect(() => {
-        if (!isUserLoggedIn) {
+        if (!isUserLoggedIn || !cart) {
             return;
         }
 
         loadTransportAndPaymentFromLastOrder();
-    }, []);
+    }, [!cart]);
```
tvikito, line 327: This seems little strange isn't it? Do we really need to check for this negated value instead of just `[cart]`?

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
```diff
@@ -317,12 +317,12 @@ export const useLoadTransportAndPaymentFromLastOrder = (
     };
 
     useEffect(() => {
-        if (!isUserLoggedIn) {
+        if (!isUserLoggedIn || !cart) {
```
tvikito, line 322: Hmm, if I understand this issue properly what about solve it by using `useIsUserLoggedIn` hook which we would modify to return 3 states - true, false and undefined. And then distinguish between them to know if user API is already returned or if user is really not logged in?

project-base/storefront/components/Forms/TextInput/TextInputControlled.tsx
```diff
@@ -38,6 +38,17 @@ export const TextInputControlled: FC<TextInputControlledProps> = ({
         [field, textInputProps],
     );
 
+    const combinedOnChangeHandler: ChangeEventHandler<HTMLInputElement> = useCallback(
```
tvikito: Is there reason why we have `useCallback` here?

project-base/storefront/hooks/forms/useOnFinishHydrationDefaultValuesPrefill.ts
```diff
@@ -0,0 +1,19 @@
+import { isClient } from 'helpers/isClient';
+import { useEffect } from 'react';
+import { FieldValues, UseFormReturn } from 'react-hook-form';
+import { usePersistStore } from 'store/usePersistStore';
+
+export const useOnFinishHydrationDefaultValuesPrefill = <T extends FieldValues>(
+    defaultValues: T,
+    formProviderMethods: UseFormReturn<T>,
+) => {
+    const hasHydratedOnClient = isClient && usePersistStore.persist.hasHydrated();
```
tvikito: You can use `isStoreHydrated `helper function

project-base/storefront/cypress/tsconfig.json
tvikito: Ah, yes, I've missed that.

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
sebaholesz: Yeah, because cart changes with every cart change, whereas !cart changes only when cart boolean value changes. We may find a more elegant way. Do you know any?

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
sebaholesz: Well, what would that solve? I need to wait for the cart to be loaded. Not sure how your suggestion would fix that.

project-base/storefront/components/Forms/TextInput/TextInputControlled.tsx
sebaholesz: Probably not :) I will remove it for both.

project-base/storefront/hooks/forms/useOnFinishHydrationDefaultValuesPrefill.ts
sebaholesz: Oh, thats right 👍 thanks

project-base/storefront/urql/cache/updates.ts
sebaholesz: I don't know why I haven't thought of this before 🤦 thanks for the suggestion

project-base/storefront/components/Pages/Order/TransportAndPayment/helpers.ts
sebaholesz: I switched the hooks for the boolean regarding user's sign-in status, but will leave the reset as-is.

tvikito (approved):

project-base/storefront/components/Forms/Textarea/TextareaControlled.tsx
```diff
@@ -18,9 +18,17 @@ export const TextareaControlled: FC<TextareaControlledProps> = ({ name, render,
     } = useController({ name, control });
     const textareaId = formName + '-' + name;
 
+    const combinedOnBlurHandler: FocusEventHandler<HTMLTextAreaElement> = (event) => {
```
tvikito: What about more common naming such as `handleOnBlur`?

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
```diff
@@ -46,7 +43,7 @@ export const Radiobutton = forwardRef<HTMLInputElement, RadiobuttonProps>(
                     value={value}
                     onBlur={onBlur}
                     onChange={onChange}
-                    onClick={onClickHandler}
+                    onClick={onClickWithNull}
```
tvikito: I don't understand the motivation behind using this naming instead of general like `onClickHandler`. Here I don't get why it's necessary to expose to naming that function works with `null` value? Do you have some new mindset for naming those handlers?

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
```diff
@@ -12,25 +12,22 @@ export type RadiobuttonProps = NativeProps & {
     value: any;
     checked: InputHTMLAttributes<HTMLInputElement>['checked'];
     label: ReactNode;
-    onChangeCallback?: (newValue: string | null) => void;
+    onClick?: (newValue: string | null) => void;
```
tvikito, line 15: In commit message I see 

_"radiobutton component now does not accept onClick prop
anymore, so you should provide your actions and callbacks
via onChange"_

But here I see that it still accept this prop...?

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
```diff
@@ -12,25 +12,22 @@ export type RadiobuttonProps = NativeProps & {
...
-                if (checked) {
-                    onChangeCallback(null);
-                } else {
-                    onChangeCallback(event.currentTarget.value);
-                }
-            },
-            [checked, onChangeCallback],
-        );
+            if (checked) {
+                onClick(null);
```
tvikito, line 26: I am sure there will be some motivation behind this change, but I don't see it. Can you explain please?

project-base/storefront/components/Pages/Order/ContactInformation/contactInformationFormMeta.ts
```diff
@@ -56,55 +58,99 @@ export const useContactInformationForm = (): [UseFormReturn<ContactInformation>,
                 otherwise: Yup.string(),
             }),
             differentDeliveryAddress: Yup.boolean(),
-            deliveryFirstName: Yup.string().when('differentDeliveryAddress', {
-                is: true,
+            deliveryFirstName: Yup.string().when(['differentDeliveryAddress', 'deliveryAddressUuid'], {
+                is: (differentDeliveryAddress: boolean, deliveryAddressUuid: string) =>
```
tvikito, line 62: Wouldn't be better naming such as `newDeliveryAddress`? Just by looking I have no idea what does `different` mean.

project-base/storefront/components/Pages/Order/ContactInformation/contactInformationFormMeta.ts
```diff
@@ -56,55 +58,99 @@ export const useContactInformationForm = (): [UseFormReturn<ContactInformation>,
...
                 then: validateCountry(t),
             }),
             newsletterSubscription: Yup.boolean(),
-            deliveryAddressUuid: Yup.string().optional().nullable(),
+            deliveryAddressUuid: Yup.string().optional(),
             note: Yup.string().optional().nullable(),
         }),
     );
-    const defaultValues = contactInformationValues;
+    const defaultValues: ContactInformation = {
```
tvikito: What about proposal which Ivan had about put typing to those form hooks? Cannot we implement them with this occasion? This way we would already know the type from the `contactInformationValues` and not need to type it manually.

project-base/storefront/hooks/countries/useCountriesAsSelectOptions.ts
```diff
@@ -0,0 +1,18 @@
...
+    return useMemo(() => mapCountriesToSelectOptions(countriesData?.countries), [countriesData?.countries]);
+};
+
+const mapCountriesToSelectOptions = (countries: CountryFragmentApi[] | undefined): SelectOptionType[] =>
+    countries?.map((country) => {
+        return {
+            label: country.name,
+            value: country.code,
+        };
+    }) ?? [];
```
tvikito: Maybe simpler 

` countries?.map((country) => ({
            label: country.name,
            value: country.code,
        })
) ?? [];`

would be enough?

sebaholesz: Good, thanks. Changed.

project-base/storefront/components/Pages/Order/ContactInformation/contactInformationFormMeta.ts
sebaholesz: The type there is kind of unnecessary. Even without it, it works correctly (adding or removing property breaks TS). The only reason I added it is to see this:

image

Instead of this:

image

But maybe that is unnecessary. I removed the type completely, as it is redundant, but if you have a different suggestion, then let me know 😇

project-base/storefront/components/Pages/Order/ContactInformation/contactInformationFormMeta.ts
sebaholesz: Yeah, I see how that is confused. However, this is how it is named on the API. It is a boolean parameter saying if the delivery address is different from billing address, which means, that the name is actually so confusing, that it was even hard for you to see what the parameter is for, for which I don't blame you 😆 I created a [US](https://shopsys.atlassian.net/browse/SSP-2212) where we rename this from `differentDeliveryAddress` to `isDeliveryAddressDifferentFromBilling`, ok?

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
sebaholesz: Yeah, it was a process. I saw it being called `onChangeCallback`, which made no sense as it was used along with `onChange`, which would make the developer think it is the exact same thing. However, internally, `onClick` was used. Because of that, I thought it would be cleaner and more descriptive to show that it is actually `onClick`. Do you agree?

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
sebaholesz: Oh sheesh, you are right. This was done in a fixup and did not change the message. Superb work! Thanks for taking such a detailed look 💙 I rewrote the message so it actually matches what was done.

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
sebaholesz: No, not really. I will just change it back.

project-base/storefront/components/Forms/Textarea/TextareaControlled.tsx
sebaholesz: I would like to keep this naming of "combined" where they really are "combined", meaning where we combine two onEvent handlers to one. This is used where you want to provide both your own callback and the react hook form callback. It's been like that for quite some time and it still makes sense to me.

tvikito: Really? I can't find any other place where we would use it. I still consider this as confusing because we use approach with custom function such as `onBlurHandler` and inside you do whatever magic you want (including calling another callback).

project-base/storefront/components/Pages/Order/ContactInformation/contactInformationFormMeta.ts
tvikito: Ah, I see. Sure! Thanks.

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
tvikito: What about simply have this? 

```
const onClickHandler: MouseEventHandler<HTMLInputElement> = (event) => {
            if (onClick && !checked) {
                onClick(event.currentTarget.value);
            }
        };
```

sebaholesz: That won't work. We need it to act like a checkbox, meaning that if I click on an already selected radiobutton, I want to uncheck it. Your code would not do anything in that case.

project-base/storefront/components/Forms/Textarea/TextareaControlled.tsx
sebaholesz: There are 4 palces where we use it:

image

You have to combine the handlers somewhere. Would you rather do it inside the component, or everywhere you use it? Or are you only saying that you do not like the name? We can discuss it on a call if you want.

sebaholesz: As a developer using the component, I do not care about how the handlers are combined, I only care about my handler being called. However, If I did this:

image

Then the provided `onBlur` would be ignored. If I change the order of the props:

image


Then react hook form callbacks would not work. because of that, we need to combine the handlers into a single, _combined_ handler. Which is my approach

project-base/storefront/components/Forms/Radiobutton/Radiobutton.tsx
sebaholesz: agreed to keep

project-base/storefront/components/Forms/Textarea/TextareaControlled.tsx
sebaholesz: I was wrong, this was actually introduced here. I renamed it as you wanted...

PR #3035, opened 2024-02-21
refactor mobile menu

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| refactor/redesign mobile menu
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-ssp-1723-refactor-mobile-menu.odin.shopsys.cloud
  - https://cz.tv-ssp-1723-refactor-mobile-menu.odin.shopsys.cloud


project-base/storefront/components/Layout/Header/MobileMenu/MobileMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return (
+        <>
+            <div className="py-3 flex">
+                {!!historyMenuGroups?.length && (
+                    <button
+                        className="cursor-pointer flex w-9 justify-start gap-2 items-center text-sm uppercase text-dark"
+                        title={t('Back')}
+                        onClick={() => {
+                            setCurrentMenuItems([...historyMenuGroups].pop()!);
+                            setHistoryMenuGroups(historyMenuGroups.slice(0, -1));
```
sebaholesz: I am just thinking if there isn't a nicer way to write this...

Maybe the second row could use the current value provided by the anonymous function? Not sure. It's just for you to consider, you can resolve as you wish.

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+            </div>
+
+            <div>
+                {currentMenuItems.map((navigationItem) => (
+                    <DropdownMenuListItem
+                        key={navigationItem.link}
+                        navigationItem={navigationItem}
+                        onNavigate={onMenuToggleHandler}
+                        onExpand={() => {
+                            setHistoryMenuGroups([...(historyMenuGroups || []), currentMenuItems]);
```
sebaholesz: Same here. Maybe it could use a different syntax than spreading, but I will leave it to you.

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuListItem.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { ArrowIcon } from 'components/Basic/Icon/IconsSvg';
+
+type DropdownMenuListProps = {
+    navigationItem: MenuItem;
+    onExpand: () => void;
+    onNavigate: () => void;
+};
+
+export const DropdownMenuListItem: FC<DropdownMenuListProps> = ({ navigationItem, onExpand, onNavigate }) => {
+    const isWithChildren = !!navigationItem.children?.length;
```
sebaholesz, line 12: Again, you can resolve, as I do not see a reason in setting some strict rules, but do not tell me that `hasChildren` would not be nicer 😆

sebaholesz (changes requested):
Just small tips, nothing serious. Good job...

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenu.tsx
tvikito: Take a look. I think it was little bit improved.

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenu.tsx
tvikito: Here I don't see any other option, if I don't want to modify original value. Which I shouldn't since it is returned from `useState`.

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuListItem.tsx
tvikito: I wouldn't discuss this kind of topics anymore. I think we had already enough discussions about that. I will not convince you about my opinion, but yes, I see it as better naming :) We've agreed to not use only one form so I want to keep it.

sebaholesz (approved):

PR #3039, opened 2024-02-22
[project-base] fixed promo code mass generation

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Promo code mass generation re-used the same entities for limits and flags causing the flag was not set properly for all generated promo codes. This PR fixes that
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-fix-promo-code-mass.odin.shopsys.cloud
  - https://cz.mg-fix-promo-code-mass.odin.shopsys.cloud

### DIFF
--- a/project-base/app/src/Model/Order/PromoCode/PromoCodeData.php
+++ b/project-base/app/src/Model/Order/PromoCode/PromoCodeData.php
@@ -92,4 +92,15 @@ class PromoCodeData extends BasePromoCodeData
      * @var \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup[]
      */
     public array $limitedPricingGroups = [];
+
+    public function __clone()
+    {
+        foreach ($this->flags as $key => $flag) {
+            $this->flags[$key] = clone $flag;
+        }
+
+        foreach ($this->limits as $key => $limit) {
+            $this->limits[$key] = clone $limit;
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


TomasLudvik (approved):

project-base/app/src/Model/Order/PromoCode/PromoCodeData.php
```diff
@@ -92,4 +92,15 @@ class PromoCodeData extends BasePromoCodeData
      * @var \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup[]
      */
     public array $limitedPricingGroups = [];
+
+    public function __clone()
+    {
+        foreach ($this->flags as $key => $flag) {
+            $this->flags[$key] = clone $flag;
```
pk16011990, line 99: Are you sure cloning of Entities is ok? It can confuse developers, I think. The same goes for the limits below. Does not it work properly without cloning, only with references?

The second point is that another new array property needs to be added here. Do you have any idea how to prevent this? I have no idea 😅

grossmannmartin: Well, cloning entity itself is ok. It may be not very clear, I agree. 
Cloning was already a part of the mass-generating promo codes, but it was not working properly due to a shallow clone of the data object. 
For this purpose, there was a custom limit setting due to mass generation. 
So the problem you're describing was already there (= somebody forgot to consider a new field). 

It may work if the data object has only other data objects, but flags and limits are entities, so during the mass generation the entity is added to the first promo code, then updated to the second, etc.

A better way could be to rethink working with the mass generation, but as we considering reworking the whole promo code functionality, this seemed like a better cost-benefit solution.

PR #3047, opened 2024-02-28
[shopsys] Luigi's Box now searches through all the searchable entities

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| Reworked searching in Luigi's Box to promises, which leads to a single query on Luigi's Box API, which is cheaper and faster. Added articles, brands, and categories to the search, so now all searchable entities are searched via Luigi's Box.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tl-luigis-box-search-in-all-entities.odin.shopsys.cloud
  - https://cz.tl-luigis-box-search-in-all-entities.odin.shopsys.cloud


ecs.php
```diff
@@ -38,7 +39,10 @@
             FunctionLengthSniff::class => [
                 __DIR__ . '/utils/releaser/src/ReleaseWorker/Release/CreateAndPushGitTagsExceptProjectBaseReleaseWorker.php',
             ],
-        ]
+            ExtendedApiClassNamespaceSniffer::class => [
+                __DIR__ . '/packages/luigis-box/src/Model/Product/Connection/ProductConnectionFactory.php',
+            ],
```
grossmannmartin: Is it a temporary solution, or it should be skipped even in the final version? 
I would prefer to adjust the rule or comply with it

packages/luigis-box/src/Component/LuigisBox/Filter/ProductFilterToLuigisBoxFilterMapper.php
```diff
@@ -22,19 +22,17 @@ public function __construct(
...
-     * @param string $luigisBoxType
      * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData $productFilterData
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
      * @return array
      */
-    public function mapForSearch(string $luigisBoxType, ProductFilterData $productFilterData, Domain $domain): array
+    public function mapForSearch(ProductFilterData $productFilterData, Domain $domain): array
     {
         $luigisBoxFilter = [
             self::FILTER_AND => [],
```
grossmannmartin, line 30: the constants FILTER_AND and FILTER_OR have to be public?

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -14,8 +14,9 @@
 class LuigisBoxClient
 {
     public const string LUIGIS_BOX_TYPE_PRODUCT = 'item';
```
grossmannmartin: this constant is more confusing than helping for me, tbh... Even more because it's public
I understood the meaning only after examining the `mapAppTypeToLuigisBoxType` method.

I would prefer to inline the value there.

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -48,33 +49,30 @@ protected function getTrackerId(): string
 
     /**
      * @param string $query
-     * @param string $type
      * @param string $endpoint
      * @param int $page
-     * @param int $limit
+     * @param array $limitsByType
```
grossmannmartin: A more exact type would be really helpful... I believe it should be `array<string, int>`

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -0,0 +1,105 @@
...
+        $filter = [];
+        $orderingMode = null;
+
+        foreach ($luigisBoxBatchLoadData as $luigisBoxBatchLoadDataItem) {
+            if ($luigisBoxBatchLoadDataItem->getType() === LuigisBoxClient::MAIN_TYPE) {
+                $query = $luigisBoxBatchLoadDataItem->getQuery();
+                $endpoint = $luigisBoxBatchLoadDataItem->getEndpoint();
+                $page = $luigisBoxBatchLoadDataItem->getPage();
+                $filter = $luigisBoxBatchLoadDataItem->getFilter();
+                $orderingMode = $luigisBoxBatchLoadDataItem->getOrderingMode();
```
grossmannmartin: I am not sure how will it look in the final version, but now it seems like a search without the product will not work properly as the mandatory data will not be set.

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -48,33 +49,30 @@ protected function getTrackerId(): string
...
      * @param array $filter
      * @param string|null $orderingMode
-     * @return \Shopsys\LuigisBoxBundle\Component\LuigisBox\LuigisBoxResult
+     * @return \Shopsys\LuigisBoxBundle\Component\LuigisBox\LuigisBoxResult[]
      */
     public function getData(
         string $query,
-        string $type,
         string $endpoint,
         int $page,
```
grossmannmartin, line 67: this argument is once named page and once offset (getLuigisBoxApiUrl vs getData). Could it be unified to match the expected value?
I'm not sure whether to pass a page number or the offset

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+            $limitsByType[$luigisBoxBatchLoadDataItem->getType()] = $luigisBoxBatchLoadDataItem->getLimit();
+        }
+
+        return $this->promiseAdapter->all($this->mapDataByTypes(
+            $this->luigisBoxClient->getData($query, $endpoint, $page, $limitsByType, $filter, $orderingMode),
+            $limitsByType,
```
grossmannmartin: This seems confusing to me... limitsByType is widely used to determine the proper search. I'm not sure how the filter and orderingMode fit into that.
Is it exclusive to products? Can it be set differently for products and categories? 

I would rather have something like `getData($query, $isAutocomplete, $batchLoadDataByTypes)`

deciding the endpoint at the top level seems like a violation of encapsulation... the proper endpoint name should be the responsibility of the client class itself. The input data should be only if I want to fullsearch or autocomplete.

passing the array of configurations is more open to changes because I'm able to easily add for example independent pagination for categories and products.

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -0,0 +1,105 @@
...
+     * @param array $limitsByType
+     * @return array
+     */
+    protected function mapDataByTypes(array $luigisBoxResults, array $limitsByType): array
+    {
+        $mappedData = [];
+
+        foreach ($limitsByType as $type => $limit) {
+            $methodNameForDataMapping = 'map' . ucfirst($type) . 'Data';
+            $mappedDataOfCurrentType = $this->{$methodNameForDataMapping}($luigisBoxResults[$type], $limit);
```
grossmannmartin: I'm not sure about this approach... 
I don't like that the method usage is not clear for static analysis and ide... 

I would rather have the explicit mapping method

packages/luigis-box/src/Component/LuigisBox/Filter/ProductFilterToLuigisBoxFilterMapper.php
TomasLudvik: I have changed theirs visibility

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
TomasLudvik: I have removed the constant

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
TomasLudvik: I have added exception if product search is not present, I will catch it in future SearchProviders and convert it to user error

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
TomasLudvik: united

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
TomasLudvik: I have renamed used variables to be more clear, I have added validation for endpoints and also renamed the endpoint to action and we have discussed configuration and decided to keep it as it is.

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
TomasLudvik: I have changed it to if

ecs.php
TomasLudvik: I have reworked the factory to use the previous factory as a dependency instead of extending it.

ecs.php
```diff
@@ -9,6 +9,7 @@
 use Shopsys\CodingStandards\Sniffs\ConstantVisibilityRequiredSniff;
 use Shopsys\CodingStandards\Sniffs\ForceLateStaticBindingForProtectedConstantsSniff;
 use SlevomatCodingStandard\Sniffs\Functions\FunctionLengthSniff;
+use Sniffer\ExtendedApiClassNamespaceSniffer;
```
grossmannmartin: this use seems unused

packages/frontend-api/src/Model/Resolver/Brand/Search/BrandSearchQuery.php
```diff
@@ -0,0 +1,34 @@
...
+    ) {
+    }
+
+    /**
+     * @param \Overblog\GraphQLBundle\Definition\Argument $argument
+     * @return \GraphQL\Executor\Promise\Promise|\Shopsys\FrameworkBundle\Model\Product\Brand\Brand[]
+     */
+    public function brandSearchQuery(Argument $argument): Promise|array
+    {
+        $brandSearchResultsProvider = $this->brandSearchResultsProviderResolver->getSearchResultsProviderByDomainIdAndEntityName($this->domain->getId(), 'article');
```
grossmannmartin:
```suggestion
        $brandSearchResultsProvider = $this->brandSearchResultsProviderResolver->getSearchResultsProviderByDomainIdAndEntityName($this->domain->getId(), 'brand');
```

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -0,0 +1,202 @@
...
+        foreach ($luigisBoxBatchLoadData as $luigisBoxBatchLoadDataItem) {
+            if ($luigisBoxBatchLoadDataItem->getType() === LuigisBoxClient::MAIN_TYPE) {
+                $productSearchIsPresent = true;
+
+                break;
+            }
+        }
+
+        if (!$productSearchIsPresent) {
+            throw new ProductSearchIsMandatoryForAllLuigisBoxSearchesException();
```
grossmannmartin: When I call this query with luigisbox enabled, I get 500 error. I would expect a nice graphql error explaining what did I do wrong

```{
  categoriesSearch(search:"ele"){
  edges{
    node{
      name
      seoH1
    }
  }
}}
```

grossmannmartin (changes requested):

ecs.php
TomasLudvik: removed

packages/frontend-api/src/Model/Resolver/Brand/Search/BrandSearchQuery.php
TomasLudvik: fixed

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
TomasLudvik: I have added user error

grossmannmartin (approved):

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -18,6 +18,11 @@ class LuigisBoxClient
     public const string ACTION_SEARCH = 'search';
     public const string ACTION_AUTOCOMPLETE = 'autocomplete/v2';
 
+    public const string ARTICLE_TYPE_IN_LUIGIS_BOX = 'article';
+    public const string BRAND_TYPE_IN_LUIGIS_BOX = 'brand';
+    public const string CATEGORY_TYPE_IN_LUIGIS_BOX = 'category';
+    public const string PRODUCT_TYPE_IN_LUIGIS_BOX = 'item';
```
vitek-rostislav: I would prefer if the constants had the same prefix, something like "LUIGIS_BOX_TYPE_ARTICLE", "LUIGIS_BOX_TYPE_BRAND", etc., it simplifies the usage then IMHO. But I do not insist on that :wink:

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -192,15 +200,30 @@ protected function getIdFromIdentity(string $identity): int
...
+    {
+        if (in_array(self::MAIN_TYPE, $types, true)) {
+            return self::MAIN_TYPE;
+        }
+
+        return reset($types);
+    }
+
+    /**
+     * @param array $typesWithLimits
```
vitek-rostislav: could you be more specific in the annotation, please? is it `array<string, int>`?

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -182,21 +172,33 @@ protected function mapBranchData(LuigisBoxResult $luigisBoxResult): array
...
     /**
      * @param array $luigisBoxBatchLoadData
+     * @return \Shopsys\LuigisBoxBundle\Model\Batch\LuigisBoxBatchLoadData
      */
-    protected function validateProductSearchIsPresent(array $luigisBoxBatchLoadData): void
+    protected function getMainBatchLoadData(array $luigisBoxBatchLoadData): LuigisBoxBatchLoadData
     {
-        $productSearchIsPresent = false;
+        $mainBatchLoadData = null;
+        $firstBatchLoadData = null;
```
vitek-rostislav: the logic with main and first batch load data seems quite overwhelming to me. What about this?
```
protected function getMainBatchLoadData(array $luigisBoxBatchLoadData): LuigisBoxBatchLoadData
{
    $mainBatchLoadData = null;
    foreach ($luigisBoxBatchLoadData as $luigisBoxBatchLoadDataItem) {
        if ($luigisBoxBatchLoadDataItem->getType() === LuigisBoxClient::MAIN_TYPE) {
            $mainBatchLoadData = $luigisBoxBatchLoadDataItem;

            break;
        }
    }

    return $mainBatchLoadData ?? reset($luigisBoxBatchLoadData);
}
```

And if you wanted to be super fancy, you could handle the state when the input array is empty:
```
return $mainBatchLoadData ?? reset($luigisBoxBatchLoadData) ?? throw new WhateverExceptionNoBatchLoadData()
```

packages/luigis-box/src/Model/Article/ArticlesSearchResultsProvider.php
```diff
@@ -33,9 +35,11 @@ public function getArticlesSearchResults(
         Argument $argument,
     ): Promise|array {
         return $this->luigisBoxBatchLoader->load(
-            new LuigisBoxBatchLoadData(
+            $this->luigisBoxBatchLoadDataFactory->create(
                 'article',
```
vitek-rostislav: why do not you use constants from `LuigisBoxClient` here (and in other places where the `create` method is called)?

vitek-rostislav: + also in `LuigisBoxBatchLoader::mapDataByTypes()`?

vitek-rostislav (changes requested):

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
TomasLudvik: I have used first variant as it would not be called without any batch load data

packages/luigis-box/src/Model/Article/ArticlesSearchResultsProvider.php
TomasLudvik: Ok

vitek-rostislav (approved):

PR #3048, opened 2024-03-01
Added indexes for columns which were used for order by and where in entites TransferIssue and CronModuleRun

sspooky13:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Optimize administration homepage
|New feature| No
|Fixes issues| ... 

----
  - https://pt-admin-homepage-optimize.odin.shopsys.cloud
  - https://cz.pt-admin-homepage-optimize.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/Cron/CronModuleRun.php
+++ b/packages/framework/src/Component/Cron/CronModuleRun.php
@@ -8,7 +8,13 @@ use DateTime;
 use Doctrine\ORM\Mapping as ORM;
 
 /**
- * @ORM\Table(name="cron_module_runs")
+ * @ORM\Table(
+ *      name="cron_module_runs",
+ *      indexes={
+ *           @ORM\Index(columns={"started_at"}),
+ *           @ORM\Index(columns={"finished_at"}),
+ *       }
+ *  )
  * @ORM\Entity
  */
 class CronModuleRun
--- a/project-base/app/src/Model/Transfer/Issue/TransferIssue.php
+++ b/project-base/app/src/Model/Transfer/Issue/TransferIssue.php
@@ -13,7 +13,7 @@ use Gedmo\Mapping\Annotation as Gedmo;
  * @ORM\Table(
  *     name="transfer_issues",
  *     indexes={
- *          @ORM\Index(columns={"created_at"}),
+ *          @ORM\Index(columns={"created_at", "deleted_at", "transfer_id"}),
  *      }
  * )
  * @ORM\Entity
--- a/UPGRADE-15.0.md
+++ b/UPGRADE-15.0.md
@@ -264,6 +264,10 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 
 -   see #project-base-diff to update your project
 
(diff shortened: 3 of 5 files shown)


packages/framework/src/Component/Cron/CronModuleRun.php
```diff
@@ -8,7 +8,12 @@
 use Doctrine\ORM\Mapping as ORM;
 
 /**
- * @ORM\Table(name="cron_module_runs")
+ * @ORM\Table(
+ *      name="cron_module_runs",
+ *      indexes={
+ *           @ORM\Index(columns={"started_at", "finished_at"}),
```
vitek-rostislav: why is here a composite index for these two columns? Where do we use them together?

project-base/app/src/Model/Transfer/Issue/TransferIssue.php
```diff
@@ -13,7 +13,7 @@
  * @ORM\Table(
  *     name="transfer_issues",
  *     indexes={
- *          @ORM\Index(columns={"created_at"}),
+ *          @ORM\Index(columns={"created_at", "deleted_at"}),
```
vitek-rostislav: if the index is here primarily because of the following query, we should add `transfer_id` to the index as well :thinking: 

```
return $this->getQueryBuilder()
            ->select('ti, t')
            ->join(Transfer::class, 't', Join::WITH, 'ti.transfer = t')
            ->where('ti.deletedAt IS NULL')
            ->andWhere('ti.createdAt > :fromDateTime')
            ->setParameter('fromDateTime', $fromDateTime)
            ->orderBy('ti.createdAt', 'DESC')
            ->addOrderBy('ti.id', 'DESC');
```

vitek-rostislav (changes requested):

packages/framework/src/Component/Cron/CronModuleRun.php
sspooky13: I created composite index because I didn't deal with it but you have right, composite index in this case is useless, so I separated them. https://github.com/shopsys/shopsys/pull/3048/commits/8e1976e2455895447153e09dbdef8b13afc324b3

project-base/app/src/Model/Transfer/Issue/TransferIssue.php
sspooky13: index for `transfer_id` already exists, I don't think it's really necessary to create composite index for every usage of column, I agree it can be better but I don't think that will speed up extremely but if you want it I can add it to index

vitek-rostislav: yeah, add it there, please, you can send it to BV then :wink:

vitek-rostislav (approved):

project-base/app/src/Model/Transfer/Issue/TransferIssue.php
sspooky13: ok, i added `transfer_id` to index and migration and send it to BV, thanks 😊

pk16011990 (approved):

PR #3051, opened 2024-03-01
[shopsys] product recalculations scoping

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| In many occasions, to optimize the application, you need to restrict the scope of a product recalculation and elastic export - sometimes you do not need to recalculate visibility or selling denied, or you need to export just particular fields into Elasticsearch. For example, when launching a new domain, you need to export just the product URLs into Elasticsearch which takes significantly less time than the full recalculation and export. On the other hand, it might be hard to keep in mind all the dependencies (e.g. you need to know you need to recalculate visibility after you change `Product::$sellingFrom`, or you need to keep in mind that with a change of product URL, you always need to export `hreflang_links` into Elasticsearch as well, etc.). This PR provides the recalculation scoping feature and defines all the scope dependencies (see the table below which is an output of a new `shopsys:list:export-scopes` CLI command) which provides wide possibilities for optimizing the application.
|New feature| Yes 
|Fixes issues| ... 

image

----
  - https://rv-scopes-without-enums.odin.shopsys.cloud
  - https://cz.rv-scopes-without-enums.odin.shopsys.cloud

### DIFF
--- a/docs/asynchronous-processing/product-recalculations.md
+++ b/docs/asynchronous-processing/product-recalculations.md
@@ -115,11 +115,56 @@ That way we are sure the code works – the message is truly dispatched, thus th
     Calling `handleDispatchedRecalculationMessages()` method creates a snapshot in Elasticsearch before any changes are exported and restores it afterward.<br>
     Tests are not dependent on the changes made or the order of run.
 
+## Recalculation scope
+
+In many occasions, to optimize the application, you might need to restrict the scope of a product recalculation and elastic export.
+Sometimes you do not need to recalculate visibility or selling denied, or you need to export just particular fields into Elasticsearch.
+For example, when launching a new domain, you need to export just the product URLs into Elasticsearch which takes significantly less time than the full recalculation and export.
+On the other hand, it might be hard to keep in mind all the dependencies (e.g. you need to know you need to recalculate visibility after you change `Product::$sellingFrom`,
+or you need to keep in mind that with a change of product URL, you always need to export `hreflang_links` into Elasticsearch as well, etc.).
+
+For this purpose, recalculation scopes are defined as an associative array in the `Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeConfigFacade` class.
+The scopes are represented by instances of the `Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeRule` class and indexed by their names.
+Each scope rule defines which fields should be exported to Elasticsearch together (`$productExportFields` property) and whether some actions (recalculations) should be done before the export (`$productExportPreconditions` property).
+
+You can use `./bin/console shopsys:list:export-scopes` command to list and examine all the available scopes.
+
+The scope usage can be seen in action e.g. in `Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchAffectedProductsSubscriber` class:
+
+```php
+public function dispatchAffectedByBrand(BrandEvent $brandEvent): void
+{
+    $productIds = $this->affectedProductsFacade->getProductIdsWithBrand($brandEvent->getBrand());
+
+    $this->productRecalculationDispatcher->dispatchProductIds(
+        $productIds,
+        ProductRecalculationPriorityEnum::REGULAR,
+        [ProductExportScopeConfig::SCOPE_BRAND],
+    );
+}
+```
+
+Thanks to the usage of `SCOPE_BRAND` here, no visibility or selling denied recalculations are done after a brand is updated, only the brand-related fields (brand ID, name, and URL) are exported to Elasticsearch for the affected products.
+
+There are several methods that allow you to modify the scopes configuration further to suit your project needs.
+
+-   `addExportFieldsToExistingScopeRule()`
+-   `addNewExportScopeRule()`
+-   `overwriteExportScopeRule()`
+
+You can use the methods in the overridden `App\Model\Product\Elasticsearch\Scope\ProductExportScopeConfig::loadProductExportScopeRules()`.
+
+!!! note
+
+    The export fields restriction is ignored in situations when the product is not present in Elasticsearch (e.g. it was not exported yet) and visibility recalculation needs to be done.
+    In such cases, all the product fields are always exported to Elasticsearch to ensure no data are missing for the product.
+
 ## Invoke recalculations manually
 
 It's possible to invoke recalculations manually with the `./bin/console shopsys:dispatch:recalculations` command.
 
-This command accepts ids of products, that should be dispatched, or `--all` option to dispatch all products. You can also define the priority of the recalculations using `--priority` option.
+This command accepts ids of products, that should be dispatched, or `--all` option to dispatch all products.
+You can also define the priority and/or scopes of the recalculations using `--priority`, and/or `--scope` options.
 
 ```bash
 # dispatch products with ids 1, 2 and 3
@@ -130,4 +175,7 @@ This command accepts ids of products, that should be dispatched, or `--all` opti
 
 # dispatch product with id 22 into the high priority queue
 ./bin/console shopsys:dispatch:recalculations 22 --priority=high
+
+# dispatch all products with the "product_selling_denied_scope" scope
+./bin/console shopsys:dispatch:recalculations --all --scope=product_selling_denied_scope
 ```
--- a/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
+++ b/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
@@ -49,7 +49,7 @@ class DispatchRecalculationMessageCommand extends Command
             ->addOption(
                 'priority',
                 'p',
-                InputOption::VALUE_OPTIONAL,
+                InputOption::VALUE_REQUIRED,
                 sprintf('Define the message priority. Possible values are: %s', ProductRecalculationPriorityEnum::getPipeSeparatedValues()),
                 ProductRecalculationPriorityEnum::REGULAR->value,
             )
--- a/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
+++ b/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
@@ -52,6 +52,13 @@ class DispatchRecalculationMessageCommand extends Command
                 InputOption::VALUE_OPTIONAL,
                 sprintf('Define the message priority. Possible values are: %s', ProductRecalculationPriorityEnum::getPipeSeparatedValues()),
                 ProductRecalculationPriorityEnum::REGULAR->value,
+            )
+            ->addOption(
+                'scope',
+                's',
+                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
+                'Define the message scopes. Run "shopsys:list:export-scopes" command for listing available scopes',
+                [],
             );
     }
 
@@ -63,6 +70,7 @@ class DispatchRecalculationMessageCommand extends Command
         $symfonyStyle = new SymfonyStyle($input, $output);
 
         $productIds = $input->getArgument('productIds');
+        $scopes = $input->getOption('scope');
         $shouldRecalculateAll = $input->getOption('all');
         $priority = ProductRecalculationPriorityEnum::tryFrom($input->getOption('priority'));
 
@@ -85,20 +93,26 @@ class DispatchRecalculationMessageCommand extends Command
         }
 
         if ($shouldRecalculateAll) {
-            return $this->executeAll($symfonyStyle);
+            return $this->executeAll($symfonyStyle, $scopes);
         }
 
-        return $this->executeIds($productIds, $symfonyStyle, $priority);
+        return $this->executeIds($productIds, $symfonyStyle, $priority, $scopes);
     }
 
     /**
      * @param \Symfony\Component\Console\Style\SymfonyStyle $symfonyStyle
+     * @param string[] $scopes
      * @return int
      */
-    protected function executeAll(SymfonyStyle $symfonyStyle): int
+    protected function executeAll(SymfonyStyle $symfonyStyle, array $scopes): int
     {
-        $this->productRecalculationDispatcher->dispatchAllProducts();
-        $symfonyStyle->success('Dispatched all products to recalculate');
+        $this->productRecalculationDispatcher->dispatchAllProducts($scopes);
+        $message = 'Dispatched all products to recalculate';
+
+        if (count($scopes) > 0) {
+            $message .= sprintf(' with scopes: %s', implode(', ', $scopes));
+        }
+        $symfonyStyle->success($message);
 
         return Command::SUCCESS;
     }
@@ -107,12 +121,14 @@ class DispatchRecalculationMessageCommand extends Command
      * @param int[] $productIds
      * @param \Symfony\Component\Console\Style\SymfonyStyle $symfonyStyle
      * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationPriorityEnum $priority
+     * @param string[] $scopes
      * @return int
      */
     protected function executeIds(
         array $productIds,
         SymfonyStyle $symfonyStyle,
         ProductRecalculationPriorityEnumInterface $priority,
+        array $scopes,
     ): int {
         try {
             Assert::allNumeric($productIds, 'All product IDs must be numeric');
@@ -123,8 +139,12 @@ class DispatchRecalculationMessageCommand extends Command
             return Command::FAILURE;
         }
 
-        $dispatchedProductIds = $this->productRecalculationDispatcher->dispatchProductIds($productIds, $priority);
-        $symfonyStyle->success(['Dispatched message for IDs', implode(', ', $dispatchedProductIds), sprintf('Priority: %s', $priority->value)]);
+        $dispatchedProductIds = $this->productRecalculationDispatcher->dispatchProductIds($productIds, $priority, $scopes);
+        $symfonyStyle->success([
+            'Dispatched message for IDs', implode(', ', $dispatchedProductIds),
+            sprintf('Priority: %s', $priority->value),
+            sprintf('Scopes: %s', count($scopes) > 0 ? implode(', ', $scopes) : '-'),
+        ]);
 
         return Command::SUCCESS;
     }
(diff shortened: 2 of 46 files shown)


packages/framework/src/Command/DispatchRecalculationMessageCommand.php
```diff
@@ -52,6 +52,12 @@ protected function configure(): void
                 InputOption::VALUE_OPTIONAL,
                 sprintf('Define the message priority. Possible values are: %s', ProductRecalculationPriorityEnum::getPipeSeparatedValues()),
                 ProductRecalculationPriorityEnum::REGULAR->value,
+            )
+            ->addOption(
+                'scopes',
+                's',
+                InputOption::VALUE_OPTIONAL,
```
grossmannmartin: I believe it should be VALUE_REQUIRED as calling it like `shopsys:dispatch:recalculations --scopes` has no meaning (options themselves are optional by their nature). 
The same probably applies to the priority above


Edit: 
maybe `InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY` can be used, so it's then possible to use `./bin/console shopsys:dispatch:recalculations -s scope1 -s scope2 10`
to avoid arbitrary separation with the comma

see https://symfony.com/doc/current/console/input.html#using-command-options

packages/framework/src/Component/Elasticsearch/IndexFacade.php
```diff
@@ -93,42 +93,44 @@ public function export(AbstractIndex $index, IndexDefinition $indexDefinition, O
...
+                $lastProcessedId = array_key_last($currentBatchData);
+            } while ($currentBatchSize >= $index->getExportBatchSize());
+
+            $this->indexRepository->deleteNotPresent($indexDefinition, $exportedIds);
 
-        $this->sqlLoggerFacade->reenableLogging();
+            $progressBar->finish();
+            $output->writeln('');
+        } finally {
+            $this->sqlLoggerFacade->reenableLogging();
```
grossmannmartin, line 132: good idea 👍

packages/framework/src/Component/Reflection/ReflectionHelper.php
```diff
@@ -0,0 +1,20 @@
...
+
+class ReflectionHelper
+{
+    /**
+     * @param string $fqcn
+     * @return array
+     */
+    public static function getAllPublicClassConstants(string $fqcn): array
+    {
+        return array_values((new ReflectionClass($fqcn))->getConstants(ReflectionClassConstant::IS_PUBLIC));
```
grossmannmartin: At first, I was reluctant about using reflection in the runtime code. Then I did some measurements and I think it's not that much of a problem. 

But still, as this method is called for each processed product, by creating an array cache, we can improve performance. 

It's not necessary to have it as a service and clean the cache with resetinterface as the value of the cache can be changed only by changing the code

packages/framework/src/Model/Article/Elasticsearch/ArticleIndex.php
```diff
@@ -32,10 +33,18 @@ public function getTotalCount(int $domainId): int
...
-    public function getExportDataForBatch(int $domainId, int $lastProcessedId, int $batchSize): array
-    {
+    public function getExportDataForBatch(
+        int $domainId,
+        int $lastProcessedId,
+        int $batchSize,
+        array $fields = [],
+    ): array {
+        if ($fields !== []) {
+            throw new InvalidArgumentException('Scoping export by fields is not supported for articles.');
```
grossmannmartin: Maybe a separate exception would be nice?

packages/framework/src/Model/Article/Elasticsearch/ArticleIndex.php
```diff
@@ -32,10 +33,18 @@ public function getTotalCount(int $domainId): int
...
+     * @param string[] $fields
      * @return array
      */
-    public function getExportDataForBatch(int $domainId, int $lastProcessedId, int $batchSize): array
-    {
+    public function getExportDataForBatch(
+        int $domainId,
+        int $lastProcessedId,
+        int $batchSize,
+        array $fields = [],
```
grossmannmartin, line 43: I was thinking about a separation of the scoped and non-scoped exports. My thoughts wandered about two interfaces and implementing the right one to determine whether the index supports scoping or not... But I realize that in the end, we have one IndexFacade::exportIds method that should be probably also somehow separated... And things stopped to make sense... 

We can discuss it further if the idea seems interesting to you :)

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -57,10 +61,16 @@ public function __construct(
...
      * @param int $lastProcessedId
      * @param int $batchSize
+     * @param string[] $fields
      * @return array
      */
-    public function getProductsData(int $domainId, string $locale, int $lastProcessedId, int $batchSize): array
-    {
+    public function getProductsData(
+        int $domainId,
+        string $locale,
```
grossmannmartin, line 79: We should have pass DomainConfig when this was implemented... just a sigh, no need to do anything

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -115,58 +126,74 @@ public function getProductTotalCountForDomain(int $domainId): int
...
      * @param string $locale
+     * @param string[] $fields
      * @return array
      */
-    protected function extractResult(Product $product, int $domainId, string $locale): array
+    protected function extractResult(Product $product, int $domainId, string $locale, array $fields): array
+    {
+        if (count($fields) === 0) {
+            $fields = $this->productExportFieldProvider->getAll();
+        }
```
grossmannmartin: can fields be different for individual extractResult calls?

I'm considering whether this could be moved above foreach in `getProductsDataForIds` and `getPoductsData`?

packages/framework/src/Model/Product/Elasticsearch/Scope/ProductExportScopeConfig.php
```diff
@@ -0,0 +1,130 @@
...
+                    self::PRECONDITION_VISIBILITY_RECALCULATION,
+                    self::PRECONDITION_SELLING_DENIED_RECALCULATION,
+                ]),
+            ];
+        }
+
+        return $this->productExportScopeRules;
+    }
+
+    public function reset(): void
```
grossmannmartin: I'm not sure if we discussed it (and it's possible this was done on my request, in that case, I'm sorry 😇 )

Is it necessary to reset the scope rules? If I understand it correctly, the only way how to change the outcome of the getProductExportScopeRules method is by changing the source code, or?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -31,25 +33,34 @@ public function __construct(
...
      */
-    public function recalculate(array $productIds): void
+    public function recalculate(array $productIds, array $exportScopes): void
     {
         $idsToRecalculate = $this->productRecalculationRepository->getIdsToRecalculate($productIds);
 
-        $this->productVisibilityFacade->calculateProductVisibilityForIds($idsToRecalculate);
+        if ($this->productExportScopeConfigFacade->shouldRecalculateVisibility($exportScopes)) {
+            $this->productVisibilityFacade->calculateProductVisibilityForIds($idsToRecalculate);
+        }
```
grossmannmartin: what about extracting this if and the latter into separate methods? 

```
$this->calculateVisibilityIfNeeded($idsToRecalculate, $exportScopes);

$this->calculateSellingDeniedIfNeeded($idsToRecalculate, $exportScopes);
```

packages/framework/src/Model/Product/Recalculation/ProductRecalculationMessageHandler.php
```diff
@@ -49,10 +50,11 @@ protected function process(array $jobs): void
...
             $productIds[] = $message->productId;
             $acknowledgers[] = $ack;
+            $exportScopes = [...$exportScopes, ...$message->exportScopes];
         }
 
         try {
-            $this->productRecalculationFacade->recalculate($productIds);
+            $this->productRecalculationFacade->recalculate($productIds, $exportScopes);
             $this->ackAll($acknowledgers);
         } catch (Throwable $e) {
```
grossmannmartin: I'm afraid this will not work as expected. 

If we dispatch 
```
./bin/console shop:dis:reca -s product_name_scope 1
./bin/console shop:dis:reca -s product_description_scope 2
./bin/console shop:dis:reca -s product_variants_scope 3
```

then if the handler takes all three products in one batch, all three will export their name, descriptions, variants, and recalculate visibility and selling_denied

packages/framework/src/Resources/config/services.yaml
```diff
@@ -16,7 +16,7 @@ services:
         public: false
 
     Shopsys\FrameworkBundle\:
-        resource: '../../**/*{Calculation,Dispatcher,Facade,Factory,Formatter,Generator,Handler,InlineEdit,Listener,Loader,Mapper,Middleware,Parser,Provider,Recalculator,Registry,Repository,Resolver,Service,Scheduler,Subscriber,Transformer,DataFetcher}.php'
+        resource: '../../**/*{Calculation,Dispatcher,Facade,Factory,Formatter,Generator,Handler,InlineEdit,Listener,Loader,Mapper,Middleware,Parser,Provider,Recalculator,Registry,Repository,Resolver,Service,Scheduler,Subscriber,Transformer,DataFetcher,Config}.php'
```
grossmannmartin: I don't think we want to autoconfigure all Config classes as the service – for example FeedConfig, CronModuleConfig, ProductFilterConfig should not be services

grossmannmartin (changes requested):

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -45,17 +48,26 @@ public function recalculate(array $productIds, array $exportScopes): void
...
+
+        if ($shouldRecalculateSellingDenied) {
             $this->productSellingDeniedRecalculator->calculateSellingDeniedForProductIds($idsToRecalculate);
         }
 
         $fields = $this->productExportScopeConfigFacade->getExportFieldsByScopes($exportScopes);
 
         foreach ($this->domain->getAllIds() as $domainId) {
+            foreach ($idsToRecalculate as $productId) {
+                if ($shouldRecalculateVisibility && $this->productElasticsearchProvider->existsProduct($productId, $domainId) === false) {
```
grossmannmartin: can we optimize it to check all product ids at once?

project-base/app/config/services.yaml
```diff
@@ -28,7 +28,7 @@ services:
         resource: '../src/Twig/'
 
     App\:
-        resource: '../src/**/*{Calculation,Collector,Dispatcher,Facade,Factory,Generator,Handler,InlineEdit,Listener,Loader,Mapper,Middleware,Parser,Provider,Recalculator,Registry,Repository,Resolver,Service,Scheduler,Subscriber,Transformer,Validator,Transfer,Helper,Converter,DataFetcher}.php'
+        resource: '../src/**/*{Calculation,Collector,Config,Dispatcher,Facade,Factory,Generator,Handler,InlineEdit,Listener,Loader,Mapper,Middleware,Parser,Provider,Recalculator,Registry,Repository,Resolver,Service,Scheduler,Subscriber,Transformer,Validator,Transfer,Helper,Converter,DataFetcher}.php'
```
grossmannmartin: Just wanted to point this to be taken into account after considering the same comment in framework package
(about not autoconfigure the config classes)

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -399,4 +411,25 @@ protected function extractAccessoriesIds(Product $product): array
...
+    {
+        if ($this->variantCache === null) {
+            $pricingGroup = $this->pricingGroupSettingFacade->getDefaultPricingGroupByDomainId($domainId);
+            $this->variantCache = $this->productRepository->getAllSellableVariantsByMainVariant($mainVariant, $domainId, $pricingGroup);
+        }
+
+        return $this->variantCache;
+    }
+
+    #[Override]
```
grossmannmartin, line 434: I'm not sure whether Override should be added for methods implemented due to an interface

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
vitek-rostislav: makes sense, fixed in 941384c9d2, 554e1fd09b, and 6c51c58316

packages/framework/src/Component/Reflection/ReflectionHelper.php
vitek-rostislav: good point, added cache in b926e5c9b5

packages/framework/src/Model/Article/Elasticsearch/ArticleIndex.php
vitek-rostislav: created in f27d537f9d

packages/framework/src/Model/Article/Elasticsearch/ArticleIndex.php
vitek-rostislav: It seems nice but I am afraid it would require too much additional effort that would postpone this task.

Also, with the current state, you could probably implement scopes for articles easier on a project. (I am not sure anyone would need that so maybe this is not a valid argument).

I am open to discussing it and creating a separate task for this but it would probably not get the highest priority anytime soon :shrug:

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
vitek-rostislav: yes :smile:

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
vitek-rostislav: makes sense, refactored in 1d9ad3a625

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
vitek-rostislav: that is a tough question :smile: but I would say so - what if a method is removed from the interface? With the attribute, I will be warned that I am relying on something that does not work anymore.

Also, when I remove the interface implementation from the class, and forget about the method, I get also notified about that.

vitek-rostislav: but I fixed the indents - b8221ea92a :smile:

packages/framework/src/Model/Product/Elasticsearch/Scope/ProductExportScopeConfig.php
vitek-rostislav: I am also not sure if we discussed it. Nevermind, you are right, it is unnecessary, removed in 3ed6a59e11

project-base/app/config/services.yaml
vitek-rostislav: :+1: e33544480f

packages/framework/src/Resources/config/services.yaml
vitek-rostislav: :+1: 66a073e6df

project-base/app/config/services.yaml
vitek-rostislav: :+1: e33544480f

packages/framework/src/Model/Product/Recalculation/ProductRecalculationMessageHandler.php
vitek-rostislav: that is a good catch! reworked in 560ad89635

vitek-rostislav: and reworked even more in 62527a51fd

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
vitek-rostislav: reworked in 5334e75f14

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
vitek-rostislav: yeah, it looks nice but I need `$shouldRecalculateSellingDenied` later in the code so I would have to ask again.

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
grossmannmartin: make sense :) thanks for clarification

grossmannmartin (approved):

docs/asynchronous-processing/product-recalculations.md
```diff
@@ -115,11 +115,56 @@ That way we are sure the code works – the message is truly dispatched, thus th
...
+For example, when launching a new domain, you need to export just the product URLs into Elasticsearch which takes significantly less time than the full recalculation and export.
+On the other hand, it might be hard to keep in mind all the dependencies (e.g. you need to know you need to recalculate visibility after you change `Product::$sellingFrom`,
+or you need to keep in mind that with a change of product URL, you always need to export `hreflang_links` into Elasticsearch as well, etc.).
+
+For this purpose, recalculation scopes are defined as an associative array in the `Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeConfigFacade` class.
+The scopes are represented by instances of the `Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeRule` class and indexed by their names.
+Each scope rule defines which fields should be exported to Elasticsearch together (`$productExportFields` property) and whether some actions (recalculations) should be done before the export (`$productExportPreconditions` property).
+
+You can use `./bin/console shopsys:list:export-scopes` command to list and examine all the available scopes.
```
TomasLudvik, line 130: Consider adding this and other new commands to `check-console-commands` and proper part in the build of a fork in `docker-build.yaml`

TomasLudvik (approved):

docs/asynchronous-processing/product-recalculations.md
vitek-rostislav: good point, added in 273ab33

PR #3052, opened 2024-03-02
[project-base] [SSP-2146] cypress tests extra scenarios

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Before order refactoring, we wanted to implement a robust suite of tests, so we can be more sure while changing the code. This PR extends the test suite, refactors tests for better DX, and fixes bugs along the way.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-ssp-2146-cypress-tests-extra-scenarios.odin.shopsys.cloud
  - https://cz.sh-ssp-2146-cypress-tests-extra-scenarios.odin.shopsys.cloud


project-base/storefront/cypress/e2e/authentication/registration.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    it('should register as a B2C customer', () => {
+        cy.visit('/');
+
+        goToRegistrationPageFromHeader();
+        fillInRegstrationForm('commonCustomer', 'registraion-as-b2c@shopsys.com');
+
+        takeSnapshotAndCompare('b2c-registration-page');
+
+        cy.getByTID([TIDs.registration_submit_button]).click();
```
TomasLudvik: I am against all shortcuts as they can have multiple meanings. From the usage I guess it is "testID", is it essential to shortcut it?

project-base/storefront/cypress/e2e/cart/cartPage.cy.ts
```diff
@@ -1,63 +1,154 @@
...
+        checkAndHideSuccessToast('Promo code was added to the order.');
+
+        takeSnapshotAndCompare('apply-remove-and-apply-promocode_after-applying-first-promocode');
 
         continueToTransportAndPaymentSelection();
         checkUrl(url.order.transportAndPayment);
+        takeSnapshotAndCompare('transport-and-payment-page-with-applied-promo-code');
+
+        goBackToCartPage();
+        checkUrl(url.cart);
```
TomasLudvik: This style of "do something" and then "check something" for me would be better completely in one function, so it is easily reusable and it is always checked, that it is working correctly.

project-base/storefront/cypress/e2e/cart/cartPage.cy.ts
```diff
@@ -1,63 +1,154 @@
...
+        cy.getByTID([TIDs.blocks_promocode_promocodeinfo_code]).find('svg').click();
+        checkAndHideSuccessToast('Promo code was removed from the order.');
```
TomasLudvik: Just a nitpick! These parts of codes could be encapsulated in some nicely named methods like `removePromoCode`. Then this test would look like real user actions (in my opinion these tests should do same as user would do, so writing tests in same way seems logical to me):

```
...
continueToTransportAndPaymentSelection();
goBackToCartPage();
removePromoCode();
...
```

TomasLudvik (approved):

project-base/storefront/cypress/e2e/authentication/registration.cy.ts
sebaholesz: This is a result of quite extensive discussion among FE developers. Initially, we had data-testId, which we could pass to HTML, but could not use in the code (react does not understand kebab-case). Because of that, we were looking for a more suitable name, and agreed on `tid`. I understand that it might not be completely to your liking, but as I say, there was already a lot of discussions, we agreed on it, projects already applied those changes, and changing it again would be a pain in the ass, I guess.

project-base/storefront/cypress/e2e/cart/cartPage.cy.ts
sebaholesz: I agree, and that is exactly why it was refactored in a later commit.

project-base/storefront/cypress/e2e/cart/cartPage.cy.ts
sebaholesz: I agree, I try to follow this, but sometimes fail to find the balance. I will keep this in mind. If you have any specific ideas for this PR, let's discuss them.

PR #3073, opened 2024-03-18
[framework] remove enums from the packages

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| It is not possible to extend PHP enums (see https://www.php.net/manual/en/language.enumerations.object-differences.inheritance.php) so we should avoid their usage in the Shopsys packages that are meant to be extended
|New feature| No 
|Fixes issues| ... 

----
  - https://rv-goodbye-enums.odin.shopsys.cloud
  - https://cz.rv-goodbye-enums.odin.shopsys.cloud

### DIFF
--- a/UPGRADE-14.0.md
+++ b/UPGRADE-14.0.md
@@ -836,7 +836,7 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
         ```diff
             public function create(
                 ProductData $productData
-        +       ?ProductRecalculationPriorityEnumInterface $priority = null,
+        +       string $priority = ProductRecalculationPriorityEnum::REGULAR,
         -   )
         +   ): Product
         ```
@@ -846,7 +846,7 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
         -       $productId,
         +       int $productId,
                 ProductData $productData,
-        +       ?ProductRecalculationPriorityEnumInterface $priority = null,
+        +       string $priority = ProductRecalculationPriorityEnum::REGULAR,
         -   )
         +   ): Product
         ```
@@ -855,7 +855,7 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
             public function delete(
         -       $productId
         +       int $productId,
-        +       ?ProductRecalculationPriorityEnumInterface $priority = null,
+        +       string $priority = ProductRecalculationPriorityEnum::REGULAR,
         -   )
         +   ): void
         ```
--- a/packages/framework/src/Component/Reflection/ReflectionHelper.php
+++ b/packages/framework/src/Component/Reflection/ReflectionHelper.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Reflection;
+
+use ReflectionClass;
+use ReflectionClassConstant;
+
+class ReflectionHelper
+{
+    /**
+     * @var array<string, string[]>
+     */
+    protected static array $constantsIndexedByFqcn = [];
+
+    /**
+     * @param string $fqcn
+     * @return string[]
+     */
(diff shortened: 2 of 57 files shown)


packages/framework/src/Component/Reflection/ReflectionHelper.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Reflection;
+
+use ReflectionClass;
+use ReflectionClassConstant;
+
+class ReflectionHelper
```
TomasLudvik, line 10: I would like to see this as a base abstract class for enums, with method named like `slices` or `getSlices` so it feels more like an enum

project-base/app/schema.graphql
```diff
@@ -1627,8 +1627,8 @@ input ProductListInput {
 
 "One of possible types of the product list"
 enum ProductListTypeEnum {
-  COMPARISON
-  WISHLIST
+  comparison
+  wishlist
```
TomasLudvik: Does this change require any change on the storefront?

UPGRADE-14.0.md
```diff
@@ -836,7 +836,7 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
         ```diff
             public function create(
                 ProductData $productData
-        +       ?ProductRecalculationPriorityEnumInterface $priority = null,
+        +       string $priority = ProductRecalculationPriority::REGULAR,
```
TomasLudvik: There are a lot of things missing in the upgrade notes

TomasLudvik (changes requested):

UPGRADE-14.0.md
vitek-rostislav: actually all the changes are within the `14.0` version

project-base/app/schema.graphql
vitek-rostislav: good catch, it actually works both ways but I did not want to change it, fixed in 3f42c88107

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductList/ProductListTypesDataProvider.php
```diff
@@ -14,7 +15,7 @@ class ProductListTypesDataProvider
      */
     public static function getProductListTypes(): Iterator
     {
-        foreach (ProductListTypeEnum::cases() as $productListType) {
+        foreach (ReflectionHelper::getAllPublicClassConstants(ProductListTypeEnum::class) as $productListType) {
```
TomasLudvik, line 18: This should be ProductListTypeEnum::getAllCases()

TomasLudvik (approved):

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductList/ProductListTypesDataProvider.php
vitek-rostislav: sadly, I am not able to do that - the method `getAllCases` must be non-static but here we are in a data provider where I am not able to use non-static methods :sweat_smile:

packages/framework/src/Component/Enum/AbstractEnumCasesProvider.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Enum;
+
+use Shopsys\FrameworkBundle\Component\Reflection\ReflectionHelper;
+
+class AbstractEnumCasesProvider
```
TomasLudvik: I would rename this to something more generic e.g. `AbstractEnum`, as it can be used for more uses

vitek-rostislav: renamed :+1:

PR #3075, opened 2024-03-21
[project-base] replaced UUID pools by generation from entities data

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| UUID pools were the half solution and resolved in changes in UUIDs, need to add new UUIDs, etc. The new approach generates UUID by providing data from entities that should always result in the same UUID until data is changed.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tl-remove-uuid-pools.odin.shopsys.cloud
  - https://cz.tl-remove-uuid-pools.odin.shopsys.cloud


project-base/app/src/DataFixtures/Demo/OrderDataFixture.php
```diff
@@ -863,6 +802,9 @@ private function createOrder(
             null,
         );
 
+        $uniqueOrderHash .= $orderData->firstName . $orderData->lastName . $orderData->transport->getId() . $orderData->deliveryFirstName . $orderData->deliveryLastName;
+        $orderData->uuid = Uuid::uuid5(self::UUID_NAMESPACE, md5($uniqueOrderHash))->toString();
```
vitek-rostislav, line 806: this hash is heavily dependent on the data in the order so changing them would lead to a different UUID of the order, right? I am not sure whether this is a good idea, however, I am not able to come up with a better solution (I was thinking about the order number but that is generated later, right?)

vitek-rostislav: maybe a simple counter as in the article data fixture would do the trick? In such a case, the UUIDs are sensitive to the ordering of the data fixtures but that seems to me to be less probable scenario than changing the order data here :thinking:

vitek-rostislav (changes requested):
Neat solution with the UUID v5 :ok_hand: 
I have added some thoughts, please check them out.

Also, please check the failing cypress tests :pray:

vitek-rostislav (approved):

UPGRADE-15.0.md
```diff
@@ -276,6 +276,10 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   `create` method was renamed to `createIfAddressFilled` and return type was changed to `?DeliveryAddress`
 -   see #project-base-diff to update your project
 
+#### remove UUIDs pools and replace them by UUID generated by entity datas to avoid UUIDs changes in the future
```
grossmannmartin: can you please add a PR link?

PR #3083, opened 2024-03-27
add developer styleguide

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| add developer styleguide
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tv-ssp-2185-add-developer-styleguide.odin.shopsys.cloud
  - https://cz.tv-ssp-2185-add-developer-styleguide.odin.shopsys.cloud


project-base/storefront/components/Pages/Styleguide/StyleguideColors.tsx
```diff
@@ -0,0 +1,53 @@
...
+import React from 'react';
+import { twJoin } from 'tailwind-merge';
+
+class ColorItem {
+    constructor(
+        public name: string,
+        public className: string,
+        // eslint-disable-next-line no-empty-function
+    ) {}
+}
```
sebaholesz, line 11: Just a question, why to use classes instead of just regular objects?

sebaholesz:
I see that not all components are used and shown, such as uncontrolled text input, password and search inputs, etc. I am not sure how deep we want to go, just a question to consider

sebaholesz:
My question regarding this styleguide would be if we want to make it dynamic, or that's defeating the purpose (meaning that person viewing it can change some component params, such as disabled, or step for spinbox, etc)? Or if not that, do we want to show snippets of code responsible for that component to explain props? Maybe nobody would use that, its just for you to consider.

project-base/storefront/components/Pages/Styleguide/StyleguideColors.tsx
tvikito: Because it easier to create new values

sebaholesz (approved):

PR #3087, opened 2024-04-02
[frontend-api] main blog category URL is now available via settings query

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There was no easy way how to access the main blog category URL so on the storefront, we had to fetch all the blog categories and then get the link from the first one which was sub-optimal :slightly_smiling_face: 
|New feature| Yes 
|Fixes issues| ... 

----
  - https://rv-blog-url.odin.shopsys.cloud
  - https://cz.rv-blog-url.odin.shopsys.cloud


packages/frontend-api/src/Resources/config/graphql-types/SettingsDecorator.types.yaml
```diff
@@ -24,3 +24,7 @@ SettingsDecorator:
                 type: String!
                 resolve: "@=query('displayTimezoneQuery')"
                 description: "Timezone that is used for displaying time"
+            blogUrl:
```
TomasLudvik: Would not be better to name it as "mainBlogCategoryUrl" or "rootBlogCategoryUrl"?

TomasLudvik (changes requested):

packages/frontend-api/src/Resources/config/graphql-types/SettingsDecorator.types.yaml
vitek-rostislav: good point, fixed and push forced :innocent:

TomasLudvik (approved):

pk16011990 (approved):

PR #3089, opened 2024-04-03
[SSP-2201] SF defer parts of DOM

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR implements mechanism/logic which can be used for deferring (delaying) rendering and hydration of various parts of DOM. Using it has positive impact on TBT, and also other metrics. 
|New feature| Yes
|Fixes issues| ... 

----
  - https://sh-ssp-2201-sf-defer-parts-of-dom.odin.shopsys.cloud
  - https://cz.sh-ssp-2201-sf-defer-parts-of-dom.odin.shopsys.cloud


project-base/storefront/components/Blocks/Product/DeferredPromotedProducts.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const PromotedProductsProductsSliderPlaceholder = dynamic(() =>
+    import('./PromotedProductsProductsSliderPlaceholder').then(
+        (component) => component.PromotedProductsProductsSliderPlaceholder,
+    ),
+);
+
+export const DeferredPromotedProducts: FC = () => {
+    const [{ data: promotedProductsData, fetching }] = usePromotedProductsQuery();
+    const hadClientSideNavigation = useSessionStore((s) => s.hadClientSideNavigation);
```
tvikito: Cannot `hadClientSideNavigation` be directly inside `useDeferredRender`?

project-base/storefront/hooks/helpers/useDeferredRender.ts
```diff
@@ -0,0 +1,81 @@
...
+            return computeDefer(12);
+        case 'user_consent':
+            return computeDefer(13);
+        case 'gtm_head_script':
+            return computeDefer(14);
+        case 'gtm_events':
+            return computeDefer(16);
+        default:
+            return computeDefer(20);
+    }
```
tvikito: Maybe better idea would be have some map object instead of this long switch?

```
const deferMap = {
   loaders: 0,
   'blog_preview': 1,
   ....
}
```

and then get those values like 

`computeDefer(deferMap[place])`

project-base/storefront/components/Blocks/Banners/DeferredBannersSlider.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+    if (fetching) {
+        return <SkeletonModuleBanners />;
+    }
+
+    if (!sliderItemsData?.sliderItems.length) {
+        return null;
+    }
+
+    return shouldRender ? <BannersSlider sliderItems={sliderItemsData.sliderItems} /> : <BannersSliderPlaceholder />;
```
tvikito: I am not 100% sure how it will work but wouldn't be easier to have implemented `React.lazy()` with `Suspense`. I am not sure if you took a look on this feature, but with `fallback` it should do the same thing. https://react.dev/reference/react/lazy#suspense-for-code-splitting

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
```diff
@@ -1,19 +1,62 @@
...
+    { ssr: false },
+);
+const SkeletonPageStores = dynamic(
+    () => import('./SkeletonPageStores').then((component) => component.SkeletonPageStores),
+    { ssr: false },
+);
+const SkeletonPageWishlist = dynamic(
+    () => import('./SkeletonPageWishlist').then((component) => component.SkeletonPageWishlist),
+    { ssr: false },
+);
```
tvikito: Hmm, interesting idea for me. What about having every dynamic component exported directly in component file and rename this dynamic component with prefix "dynamic" like "DynamicSkeletonPageWishlist" or "SkeletonPageWishlistDynamic". Since we are (I guess even will be) using those component either as normal or dynamic, not combination, it can be even cleaner for those imports.

project-base/storefront/hooks/helpers/useDeferredRender.ts
tvikito: Or another idea

`const deferList = [loaders, 'blog_preview', ...]`

`computeDefer(deferList.indexOf(place))`

project-base/storefront/components/Blocks/Product/DeferredPromotedProducts.tsx
sebaholesz: Well, as you can imagine, that was my initial plan, but there were some bugs I was not able to explain. React did not like it. However, I tried it again after your feedback, and now it works... magic 🪄

project-base/storefront/hooks/helpers/useDeferredRender.ts
sebaholesz: Hmmm, this is quite interesting, but I must say that I like it. I rewrote it. I also removed the type, as we can use the array as a type directly

project-base/storefront/components/Blocks/Banners/DeferredBannersSlider.tsx
sebaholesz: I worked with this directly
image
https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading

In next.js, it should do the exact same thing. However, the entire problem is that, at least from what I have noticed. If I wrapped 10 things in 10 different suspense boundaries, they were all still hydrated at the same time, in the same task. So it did not help with task splitting and TBT.

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
sebaholesz: I am not sure I quite follow here. Let's talk it through.

sebaholesz: In the end, this file was not touched. Implementing dynamic imports caused more troubles than benefits.

project-base/storefront/components/Blocks/Product/Filter/DeferredFilterPanel.tsx
```diff
@@ -0,0 +1,15 @@
+import { FilterPanelProps } from './FilterPanel';
+import { SkeletonModuleFilterPanel } from 'components/Blocks/Skeleton/SkeletonModuleFilterPanel';
+import dynamic from 'next/dynamic';
+import { useDeferredRender } from 'utils/useDeferredRender';
+
+const FilterPanel = dynamic(() => import('./FilterPanel').then((component) => component.FilterPanel), {
+    ssr: false,
+    loading: () => <SkeletonModuleFilterPanel />,
```
tvikito, line 8: Hmm, I cannot wrap my head around this. What is this used for? Why some deferred components have this and some not? 🤔

project-base/storefront/utils/useDeferredRender.ts
```diff
@@ -0,0 +1,116 @@
...
+            const defer = getDeferByPageAndPlace(page, place);
+            const timer = setTimeout(() => {
+                setShouldRender(true);
+            }, defer);
+
+            return () => {
+                clearTimeout(timer);
+            };
+        }
+    }, []);
```
tvikito: This way you can remove eslint comment

```suggestion
    useEffect(() => {
        let timer: NodeJS.Timeout | undefined;
        
        if (isDeferredPage) {
            const defer = getDeferByPageAndPlace(page, place);
            timer = setTimeout(() => {
                setShouldRender(true);
            }, defer);
        }

        return () => {
            clearTimeout(timer);
        };
    }, []);
```

project-base/storefront/pages/products/[productSlug].tsx
```diff
@@ -1,58 +1,28 @@
...
-        >
-            {product?.__typename === 'RegularProduct' && <ProductDetailContent fetching={fetching} product={product} />}
-
-            {product?.__typename === 'MainVariant' && (
-                <ProductDetailMainVariantContent fetching={fetching} product={product} />
-            )}
-
-            <LastVisitedProducts currentProductCatnum={product?.catalogNumber} />
-        </CommonLayout>
+        <Suspense>
```
Comment (tvikito) [inline project-base/storefront/pages/products/[productSlug].tsx]: Hmm, so here Suspense works as we want? 🤔

project-base/storefront/store/slices/createPageLoadingStateSlice.ts
```diff
@@ -14,13 +14,15 @@ const CUSTOM_PAGE_TYPES = [
 export type PageType = FriendlyPagesTypesKey | (typeof CUSTOM_PAGE_TYPES)[number];
 
 export type PageLoadingStateSlice = {
+    hadClientSideNavigation: boolean;
```
tvikito, line 17: Why we do have past tense here? 🤔 I know I will be little annoying (with `is` again :D) but what about presence tense like `isClientSideNavigation`?

project-base/storefront/components/Layout/Header/HamburgerMenu/HamburgerMenu.tsx
```diff
@@ -4,15 +4,15 @@ import { MouseEventHandler } from 'react';
 import { twJoin } from 'tailwind-merge';
 
 type HamburgerMenuProps = {
-    onClick: MouseEventHandler<HTMLDivElement>;
+    onClick?: MouseEventHandler<HTMLDivElement>;
```
tvikito: I know another way to solve this placeholder would be creating another placeholder component which we already have a lot and we don't want another one, but this is not good in my eyes. Even though it probably won't be problem now we have component which require onClick callback in order to be working but we allow not not send this callback for the sake of the placeholder... 😏

But I think I have sort of solution. Let's require this callback but put some dummy function for the placeholder...?

tvikito (commented):
Oh my gosh. Interesting. Your work behind this feature is amazing! Really. Good job sir 🩵 BUT.

Now when I see all those changes, I can better understand @henzigo who mentioned he hopes that this will not be merged. I am not saying we shouldn't merge this (as you said we don't know about any other solution) but I just feel that the price we pay for this optimalization in "DX coins" seems to me huge. Now even those initially simple components seem to me soo heavy... Actually that's paradox because even though this makes metrics lighter whole app seems to me much heavier in the way that now we do have another level of big complexity😔

Few points to point out:

- components which we are using as both placeholder and actual render one don't anymore require props which are needed to have component working
- as already mentioned several times, maintenance of those placeholder components will be huge pain
- we have additional complexity for loading of the component itself, now I don't even count how many ways we do have (normal rendering, dynamic rendering, defer rendering, defer rendering with loading prop in dynamic function, ... I am not sure if there is more) which will make implementation heavy even before start work on some new component and definitely will partially kill motivation for refactoring
- not mentioning implementing design on new project - yes we are supposed to have universal design soon, but this will need to be adjusted in some ways for sure, this will make it also terrible to make those adjustments (for both placeholders and actual components)
- this is not our current worry but once we will decide to migrate to the new app router this is probably gonna make whole migration even more difficult, probably we will need to get rid of all these changes

Again my intention is not to say "let's not merge this", but I would like to connected people to be aware of those downsides for which we pay huge amount of those DX coins.

project-base/storefront/components/Blocks/BlogPreview/BlogPreviewPlaceholder.tsx
```diff
@@ -0,0 +1,73 @@
...
+import useTranslation from 'next-translate/useTranslation';
+
+type BlogPreviewPlaceholderProps = Pick<BlogPreviewProps, 'blogArticles' | 'blogUrl'>;
+
+export const BlogPreviewPlaceholder: FC<BlogPreviewPlaceholderProps> = ({ blogArticles, blogUrl }) => {
+    const { t } = useTranslation();
+
+    return (
+        <div>
+            <h2 className=" text-creamWhite">{t('Shopsys magazine')}</h2>
```
TomasGottvald: There is unneeded space before class name.

project-base/storefront/components/Blocks/Product/Filter/DeferredFilterPanel.tsx
```diff
@@ -0,0 +1,15 @@
...
+});
+
+export const DeferredFilterPanel: FC<FilterPanelProps> = (props) => {
```
TomasGottvald, line 11: DeferredXXX - i am thinking about switch naming - and let word 'deferred' to be last. Same as skeleton - i thought we are using this keyword at the end. For better orientation and search.

TomasGottvald: and for better upgrading too.

project-base/storefront/components/Layout/Footer/FooterPlaceholder.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const FooterPlaceholder: FC<FooterProps> = ({ simpleFooter, footerArticles, opening, phone }) => {
+    const { t } = useTranslation();
+    const { url } = useDomainConfig();
+    const [cookieConsentUrl, contactUrl] = getInternationalizedStaticUrls(['/cookie-consent', '/contact'], url);
+
+    return (
+        <>
+            {!simpleFooter && (
+                <>
+                    <a className=" text-white" href={'tel:' + phone}>
```
TomasGottvald: empty space before class name.

project-base/storefront/components/Layout/Footer/FooterPlaceholder.tsx
```diff
@@ -0,0 +1,57 @@
...
+                                            {item.name}
+                                        </ExtendedNextLink>
+                                    </li>
+                                ))}
+                            </ul>
+                        </div>
+                    ))}
+                </>
+            )}
+            {t('Copyright © 2021, Shopsys s.r.o. All rights reserved.')}
```
TomasGottvald, line 44: off topic - we still dont have dynamic year in footer txt?

project-base/storefront/store/slices/createPageLoadingStateSlice.ts
sebaholesz: The question I want an answer to is "had the client side navigation run already?". I need to know this in order to differentiate between when defer should be used and where it should not. This name makes sense to me. I can rename it to `didClientSideNavigationRun` or  `hasClientSideNavigationRun`, if that makes more sense to you.

project-base/storefront/components/Blocks/Product/Filter/DeferredFilterPanel.tsx
sebaholesz: The logic is that if you can provide a simple skeleton with no props, you can set `ssr:false` and use it as the loading component. If it was not used, the change of components would be visible and would cause layout shift. This way it is fixed. 

In some places this is not provided, because we cannot provide it (placeholder accepts props). In such situations, we import both components on the server (so `ssr:true`) and we only switch between them based on the `shouldRender` boolean. 

I will add description of all cases to the documentation and cookbooks.

project-base/storefront/utils/useDeferredRender.ts
sebaholesz: Thanks, changed.

project-base/storefront/pages/products/[productSlug].tsx
Comment (sebaholesz) [reply project-base/storefront/pages/products/[productSlug].tsx]: Short answer, yes. This has proved to be very helpful. Long answer where I give you reasons why it works would be really hard. Though I understand how Suspense works, here I can only give an intuitive answer.

Suspense works as a boundary (same as ErrorBoundary, but for `throw Promise` instead of `throw Error`). It takes the hydration and defers it to a later stage. However, our code works in a way that it has very little effect (we do not throw that many promises). Why it works here is (I suppose) because of those deferred components. Since there is nothing to be hydrated, the hydration task disappears completely when used. 

With that being said, it only works if I use both. If I only use Suspense + dynamic import, it has no effect. If I only use timeouts, it has some effect. When I use both, it has a superboosted effect.

project-base/storefront/components/Layout/Header/HamburgerMenu/HamburgerMenu.tsx
sebaholesz: I see your point. Added.

project-base/storefront/components/Blocks/BlogPreview/BlogPreviewPlaceholder.tsx
sebaholesz: Thanks, removed

project-base/storefront/components/Blocks/Product/Filter/DeferredFilterPanel.tsx
sebaholesz: Let's talk about this at the meeting. I have no strong opinion and personally do not care, but it would require us to do some serious amount of work

project-base/storefront/components/Layout/Footer/FooterPlaceholder.tsx
sebaholesz: Removed, thanks

project-base/storefront/components/Layout/Footer/FooterPlaceholder.tsx
sebaholesz: Nope. Not even the phone number and opening hours...

project-base/storefront/store/slices/createPageLoadingStateSlice.ts
tvikito: I looked at the code again to understand better but realized I am actually more confused. So maybe we can think about different naming than all of those. Because now I am asking myself why in the begging of the navigation I am setting value which is saying that navigation has ended...? 🤔

image

project-base/storefront/components/Layout/PageDeferWrapper.tsx
```diff
@@ -0,0 +1,14 @@
+import getConfig from 'next/config';
+import { Suspense } from 'react';
+
+const {
+    publicRuntimeConfig: { shouldUseDefer },
+} = getConfig();
+
+export const PageDeferWrapper: FC = ({ children }) => {
```
tvikito: We usually use word `Wrapper` because we already have component with the same name (otherwise every component which accept children is wrapper). But here it is not the case so I would rather use simple "PageDefer" name.

project-base/storefront/components/Layout/Header/HamburgerMenu/HamburgerMenu.tsx
```diff
@@ -4,7 +4,7 @@ import { MouseEventHandler } from 'react';
 import { twJoin } from 'tailwind-merge';
 
 type HamburgerMenuProps = {
-    onClick?: MouseEventHandler<HTMLDivElement>;
+    onClick: MouseEventHandler<HTMLDivElement>;
```
tvikito: Little improvement `onClick: MouseEventHandler<HTMLDivElement> | undefined`. This way you have to send onClick prop but you can send undefined value (which is little better than sending function which returns undefined, right?}.

project-base/storefront/store/slices/createPageLoadingStateSlice.ts
sebaholesz: The entire point of this variable is to divide the lifetime of the application into 2 parts:

1. before the first client side navigation has started
2. after the first client side navigation has started

This would make the natural name candidate to be `hasClientSideNavigationStarted` or `hadClientSideNavigationStarted`, which would theoretically make sense, but I just like it better without the `Started` suffix, which is unnecessary in my opinion. I do not know about any other option for the names. The main goal is to have a boolean pointer which tells the developer which of those two parts of application's lifetime he is currently in.

project-base/storefront/components/Layout/PageDeferWrapper.tsx
sebaholesz: No problem, renamed.

project-base/storefront/components/Layout/Header/HamburgerMenu/HamburgerMenu.tsx
sebaholesz: Good, refactored, thanks

PR #3095, opened 2024-04-09
product filters luigis box 15.0

tvikito:
| Q             | A
| ------------- | ---
|New feature| Yes
|Fixes issues| ... 

----
  - https://tv-product-filters-luigis-box-15-0.odin.shopsys.cloud
  - https://cz.tv-product-filters-luigis-box-15-0.odin.shopsys.cloud


project-base/storefront/utils/cart/useCurrentCart.ts
```diff
@@ -2,30 +2,35 @@ import { useCartQuery } from 'graphql/requests/cart/queries/CartQuery.generated'
...
     const isUserLoggedIn = useIsUserLoggedIn();
     const cartUuid = usePersistStore((store) => store.cartUuid);
     const packeteryPickupPoint = usePersistStore((store) => store.packeteryPickupPoint);
 
+    const [isCartHydrated, setIsCartHydrated] = useState(false);
     const isWithCart = isUserLoggedIn || !!cartUuid;
 
+    useEffect(() => {
+        setIsCartHydrated(true);
+    }, []);
```
sebaholesz, line 21: After working with this property, I am not sure what to make of it. Why does it have "cart" in the name? "is cart hydrated" sounds to me like cart is available, I can work with it. But that is not the case. Could we think of another name?

EDIT: I have actually removed this property from the `CurrentCartType` and now it's not used anywhere else in the app, but might be still worth renaming.

project-base/storefront/components/Pages/Search/SearchProductsContent.tsx
```diff
@@ -0,0 +1,89 @@
...
+    paginationScrollTargetRef,
+    searchProductsData,
+}) => {
+    const { t } = useTranslation();
+    const searchResultProducts = getMappedProducts(searchProductsData.edges);
+
+    useGtmPaginatedProductListViewEvent(searchResultProducts, GtmProductListNameType.search_results);
+
+    const isWithProductsShown = !!searchProductsData.totalCount;
+    const noProductsFound = parseInt(searchProductsData.productFilterOptions.maximalPrice) === 0;
```
sebaholesz: This condition is still wrong.

project-base/storefront/components/Pages/Search/utils.ts
```diff
@@ -134,21 +122,18 @@ export const useSearchProductsData = (
         );
     }, [currentSearchString, currentSort, JSON.stringify(currentFilter), currentPage, currentLoadMore]);
 
-    return [searchProductsData.products, searchProductsData.hasNextPage, fetching, loadMoreFetching];
+    return { searchProductsData: searchProductsData?.productsSearch, fetching, loadMoreFetching };
```
sebaholesz:
```suggestion
    return { searchProductsData: searchProductsData?.productsSearch, isFetching, isLoadMoreFetching };
```

project-base/storefront/components/Pages/Search/SearchPageContent.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    return (
+        <>
+            <Webline>
+                {(fetching || !isClient) && <SkeletonPageSearch />}
+
+                {!!searchData && !fetching && <SearchContent searchResults={searchData} />}
+
+                <SearchProducts />
+            </Webline>
```
sebaholesz, line 42: So we have a condition, that if we are fetching, search data, we show the skeleton, but displaying the products is completely independent. That means, that the products query can finish earlier and we might have a situation with a visible skeleton and also already fetched and visible products.

project-base/storefront/components/Pages/Search/SearchPageContent.tsx
```diff
@@ -0,0 +1,47 @@
...
+            <div className="mb-5 p-12 text-center">
+                <strong>{t('There are no results as you have searched with an empty query...')}</strong>
+            </div>
+        );
+    }
+
+    return (
+        <>
+            <Webline>
+                {(fetching || !isClient) && <SkeletonPageSearch />}
```
sebaholesz, line 37: The loading states here are somehow weird and they allow that for a bit, nothing is displayed, as visible in the video


https://github.com/shopsys/shopsys/assets/55181354/0b856f1b-8ede-4adf-bc37-da914ac9e5b5

project-base/storefront/components/Pages/Search/SearchPageContent.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+                {!!searchData && !fetching && <SearchContent searchResults={searchData} />}
```
sebaholesz, line 39: You should not allow a state where nothing is shown. Because of that, a ternary operator is usually safer.

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
```diff
@@ -41,7 +41,7 @@ export const SkeletonManager: FC<SkeletonManagerProps> = ({ isFetchingData, isPa
             return <SkeletonPageProductDetailMainVariant />;
         case 'category':
         case 'seo_category':
-            return <SkeletonPageProductsList />;
+            return <SkeletonModuleProductsList />;
         case 'brand':
         case 'flag':
             return <SkeletonPageProductsListSimple />;
```
sebaholesz: Now it is confusing, because below we have SkeletonPage, and here we have SkeletonModule. I would use unified naming.

sebaholesz (changes requested):

project-base/storefront/utils/store/useSetInitialStoreValues.ts
```diff
@@ -2,12 +2,14 @@ import { useRef } from 'react';
...
 import { usePersistStore } from 'store/usePersistStore';
+import { isClient } from 'utils/isClient';
 import { ServerSidePropsType } from 'utils/serverSide/initServerSideProps';
 import { v4 as uuidV4 } from 'uuid';
 
 export const useSetInitialStoreValues = ({ cookiesStore }: ServerSidePropsType) => {
     const isStoreSet = useRef(false);
 
+    const userId = usePersistStore((store) => store.userId);
     const updateUserId = usePersistStore((store) => store.updateUserId);
```
sebaholesz: I went through the code while implementing further LB changes, and I find it confusing that we use both `userId` and `userIdentifier`. Could we unify it so that it is the same as on the API (i.e. put `userIdentifier` everywhere)?

project-base/storefront/components/Blocks/Skeleton/SkeletonPageProductsList.tsx
```diff
@@ -1,57 +1,11 @@
...
-                        <div className="mb-7 grid w-full grid-cols-2 gap-3 lg:grid-cols-3 lg:gap-y-4 xl:grid-cols-4">
-                            {createEmptyArray(9).map((_, index) => (
-                                <SkeletonModuleProductListItem key={index} />
-                            ))}
-                        </div>
-                    </div>
-                </div>
-            </div>
-        </div>
+        <SkeletonModuleProductsList />
```
sebaholesz, line 9: Now the skeleton layout is broken. Could you please check it?


https://github.com/shopsys/shopsys/assets/55181354/130ba68a-e917-45e9-93e6-689cd95d76b6

sebaholesz (changes requested):

project-base/storefront/components/Pages/Search/utils.ts
tvikito: Refactored

project-base/storefront/utils/store/useSetInitialStoreValues.ts
tvikito: Hmm, I guess that's too late to refactor it right now since those changes are already merged. But I like it so I made another commit.

project-base/storefront/components/Pages/Search/SearchPageContent.tsx
tvikito: Oh, I've missed this. Refactored.

project-base/storefront/components/Pages/Search/SearchPageContent.tsx
tvikito: We have several places where we do the same conditions. We always will display some content - if we don't have searchData here we don't display the component. But in case of product search there is always at least this content 

image

project-base/storefront/components/Blocks/Skeleton/SkeletonManager.tsx
tvikito: Yup, already solved by refactoring.

project-base/storefront/components/Blocks/Skeleton/SkeletonPageProductsList.tsx
tvikito: Ah, thanks for catching it! Sure, there was still some mess within page and module for those contents. Refactored.

project-base/storefront/components/Pages/Search/SearchPageContent.tsx
tvikito: Hmm, that's right! I haven't realized this during refactoring. Refactored.

project-base/storefront/utils/cart/useCurrentCart.ts
tvikito: I am not sure what naming you would imagine...? We can think about different naming including "cart" I want to keep since this state is used for cart really. And "hydration" is used because it really means that it can be used internally within this hook otherwise you would need to do that basically in every place where you need to work with the cart and components rendered are different on server and client. 

And since you have removed exposal of this value outside I actually don't see any issue with the naming. So please add more ideas because I don't have any.

project-base/storefront/components/Pages/Search/SearchProductsContent.tsx
TomasLudvik: Can we please fix it also in version 14.0?

project-base/storefront/utils/cart/useCurrentCart.ts
sebaholesz: Yeah, since its only used here, I do not care that much

project-base/storefront/components/Pages/Search/SearchProductsContent.tsx
tvikito: No problem, after we will agree on this change I will merge it to 14.0 as well. https://github.com/shopsys/shopsys/pull/3095/commits/9d2f1976e2d5dfd0d85685a0cb6f856a94a9872b

sebaholesz: The new condition makes sense to me.

sebaholesz (approved):

PR #3098, opened 2024-04-11
[shopsys] add customer option for Verified by Customers Heureka

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| The customer now can decide in the order process whether he wants to receive Heureka satisfaction questionnaires within the Verified by Customers Heureka program. If so, the order is sent to Heureka after it is finished.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tv-ssp-1813-add-overeno-zakazniky-heureka.odin.shopsys.cloud
  - https://cz.tv-ssp-1813-add-overeno-zakazniky-heureka.odin.shopsys.cloud


project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationFormContent.tsx
```diff
@@ -60,6 +62,17 @@ export const ContactInformationFormContent: FC = () => {
                         onChange: (event) => updateContactInformation({ note: event.currentTarget.value }),
                     }}
                 />
+
+                <CheckboxControlled
```
grossmannmartin: This checkbox will be rendered every time. Even when the shop is in a language that does not support Heureka, or the Heureka itself is not activated. 

For me it's far from ideal, but I would like @pk16011990 to make a final decision.

project-base/app/tests/FrontendApiBundle/Functional/Order/graphql/CreateFullOrderMutation.graphql
```diff
@@ -43,7 +44,8 @@ mutation CreateFullOrder(
             deliveryStreet: $deliveryStreet
             deliveryCity: $deliveryCity
             deliveryCountry: $deliveryCountry
-            deliveryPostcode: $deliveryPostcode
+            deliveryPostcode: $deliveryPostcode,
```
grossmannmartin: we should make a final decision about commas in graphql queries... here is it really weird to have all fields without comma and one with 😄

grossmannmartin (changes requested):

project-base/app/tests/FrontendApiBundle/Functional/Order/graphql/CreateFullOrderMutation.graphql
vitek-rostislav: yeah, I have fixed it - [33f3ecc](https://github.com/shopsys/shopsys/pull/3098/commits/33f3eccde1e05aec2f07355c5e6262680a9c8fde)

Regarding the decision, I believe we need a check for that, I will create an issue :upside_down_face:

project-base/storefront/components/Pages/Order/ContactInformation/ContactInformationFormContent.tsx
vitek-rostislav: that is a good point. The condition was not hard to implement so I have added it in 252bf08138 :innocent:

vitek-rostislav:
I had to rebase the branch and retarget the PR to `14.0` as we want to have it in `v14.0` :slightly_smiling_face:

packages/frontend-api/src/Resources/config/graphql-types/SettingsDecorator.types.yaml
```diff
@@ -24,3 +24,7 @@ SettingsDecorator:
                 type: String!
                 resolve: "@=query('displayTimezoneQuery')"
                 description: "Timezone that is used for displaying time"
+            heurekaEnabled:
+                type: Boolean!
+                resolve: "@=query('heurekaEnabledQuery')"
+                description: "Returns true if the domain locale is supported and Heureka Shop Certification is enabled on the domain"
```
grossmannmartin: I would simplify the text only to `Returns true if Heureka is available for the current domain`

grossmannmartin (approved):

PR #3099, opened 2024-04-11
[Luigi's Box] recommended products

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| Luigi's Box comes with the ability to personalize recommended products for users. This PR added recommended products from Luigi's Box to the frontend API.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tl-luigis-box-recommendations.odin.shopsys.cloud
  - https://cz.tl-luigis-box-recommendations.odin.shopsys.cloud


packages/framework/src/Component/Enum/AbstractEnum.php
```diff
@@ -15,4 +15,14 @@ public function getAllCases(): array
     {
         return ReflectionHelper::getAllPublicClassConstants(static::class);
     }
+
+    /**
+     * @param string $case
+     */
+    public function validateCase(string $case): void
```
vitek-rostislav, line 22: love it!

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -342,14 +341,4 @@ protected function mapLimitsByTypeToLuigisBoxLimit(array $limitsByType): string
 
         return implode(',', $luigisBoxLimits);
     }
-
-    /**
-     * @param string $action
-     */
-    protected function validateActionIsValid(string $action): void
-    {
-        if (!in_array($action, [static::ACTION_SEARCH, static::ACTION_AUTOCOMPLETE], true)) {
```
vitek-rostislav, line 360: IMHO `LuigisBoxActionNotRecognizedException` still makes sense as it is more descriptive than the common `InvalidEnumCaseException`. My suggestion here is to keep the original exception, something like this:

```
protected function validateEndpointIsValid(string $endpoint): void
{
    try {
        $this->luigisBoxEndpointEnum->validateCase($endpoint);
    } catch (InvalidEnumCaseException $ex) {
        throw new LuigisBoxActionNotRecognizedException($endpoint);
    }
}
```

I am little confused with the action/endpoint names but I believe you get my point

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -222,21 +245,42 @@ protected function addAutocompleteSpecificParametersToUrl(
...
-        array $limitsByType,
-    ): string {
+    ): array {
+        $body = [];
+
+        if ($luigisBoxBatchLoadData instanceof LuigisBoxRecommendationBatchLoadData) {
+            $body = $this->addRecommendationSpecificParametersToBody($body, $luigisBoxBatchLoadData);
+        }
+
+        return [$body];
```
vitek-rostislav: if the condition above is not met, the method will return an array in an array, that is correct? `[[]]`? In `getData` method, there is a condition `if ($body !== [])` so I suppose you do not want to return `[[]]` here. What about this?

```
$body = [];

if ($luigisBoxBatchLoadData instanceof LuigisBoxRecommendationBatchLoadData) {
    $body = [$this->addRecommendationSpecificParametersToBody($body, $luigisBoxBatchLoadData)];
}

return $body;
```

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
```diff
@@ -222,21 +245,42 @@ protected function addAutocompleteSpecificParametersToUrl(
...
+    protected function addRecommendationSpecificParametersToBody(
+        array $body,
+        LuigisBoxBatchLoadData $luigisBoxBatchLoadData,
+    ): array {
         if ($luigisBoxBatchLoadData->getEndpoint() === LuigisBoxEndpointEnum::RECOMMENDATIONS) {
-            $url .= '&recommendation_type=' . $this->mapLimitsByTypeToLuigisBoxLimit($limitsByType);
+            $body['recommendation_type'] = $luigisBoxBatchLoadData->getType();
+            $body['user_id'] = $luigisBoxBatchLoadData->getUserIdentifier();
+            $body['size'] = $luigisBoxBatchLoadData->getLimit();
+            $body['hit_fields'] = ['url'];
```
vitek-rostislav, line 276: why only url? I see it is the same way in `getUrlWithBasicParameters` but I just do not understand it

packages/luigis-box/src/FrontendApi/Resolver/Recommendation/RecommendationQuery.php
```diff
@@ -0,0 +1,57 @@
...
+    public function __construct(
+        protected readonly ProductElasticsearchProvider $productElasticsearchProvider,
+        protected readonly LuigisBoxClient $luigisBoxClient,
+        protected readonly LuigisBoxBatchLoadDataFactory $luigisBoxBatchLoadDataFactory,
+    ) {
+    }
+
+    /**
+     * @param \Overblog\GraphQLBundle\Definition\Argument $argument
+     * @return \Shopsys\FrameworkBundle\Model\Product\Product[]
```
vitek-rostislav: I do not think `Product` instances are returned - the data are fetched from Elasticsearch, right?

packages/luigis-box/src/FrontendApi/Resolver/Recommendation/RecommendationQuery.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public function recommendationQuery(Argument $argument): array
+    {
+        $type = $argument['type'];
+        $limit = $argument['limit'];
+        $userIdentifier = $argument['userIdentifier'];
+        $productIds = [];
+
+        if (isset($argument['productUuids'])) {
+            $products = $this->productElasticsearchProvider->getSellableProductArrayByUuids($argument['productUuids']);
```
vitek-rostislav: Maybe we could fetch just the ids here? Let's discuss it please, I believe you can find inspiration in https://github.com/shopsys/shopsys/commit/4d2b53a578b2cdf1d44809b58fb42c9aac93f932 (which is sadly in `15.0` only)

packages/luigis-box/src/Resources/config/graphql-types/RecommendationType.types.yaml
```diff
@@ -0,0 +1,14 @@
+RecommendationType:
```
vitek-rostislav: I am afraid this needs to be a decorator, too, to enable the extension from the project, or am I missing something?

vitek-rostislav (changes requested):

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
TomasLudvik: I prefer [[]] as it is more easier for future extensions

packages/luigis-box/src/Component/LuigisBox/LuigisBoxClient.php
TomasLudvik: URL = ID

packages/luigis-box/src/FrontendApi/Resolver/Recommendation/RecommendationQuery.php
TomasLudvik: Changed

vitek-rostislav (approved):

project-base/storefront/components/Blocks/Product/AddToCartPopup.tsx
```diff
@@ -69,6 +76,16 @@ export const AddToCartPopup: FC<AddToCartPopupProps> = ({ onCloseCallback, added
                 </div>
             </div>
 
+            {isLuigisBoxActive && (
+                <Webline className="mb-6 px-0">
```
tvikito: Why do you need webline here? Simple div should be enough, or?

project-base/storefront/components/Blocks/Product/ProductsList/ProductListItem.tsx
```diff
@@ -26,6 +26,7 @@ type ProductItemProps = {
     isProductInWishlist: boolean;
     toggleProductInComparison: () => void;
     toggleProductInWishlist: () => void;
+    isMiniProduct?: boolean;
```
tvikito: What about naming like `isSimpleCard`? `mini` evoke me literally small product picture or something like that.

project-base/storefront/components/Blocks/Product/ProductsList/ProductsListContent.tsx
```diff
@@ -22,6 +22,7 @@ type ProductsListProps = {
     swipeHandlers?: SwipeableHandlers;
     className?: string;
     classNameProduct?: string;
+    isMiniSlider?: boolean;
```
tvikito: According to previous comment `isWithSimpleCards`...?

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
```diff
@@ -99,6 +102,15 @@ export const ProductDetailContent: FC<ProductDetailContentProps> = ({ product, f
                 />
 
                 <ProductDetailAvailabilityList ref={scrollTarget} storeAvailabilities={product.storeAvailabilities} />
+                {isLuigisBoxActive && (
+                    <Webline className="mb-6 px-0">
```
tvikito: Another case, I guess you can put simple div here.

project-base/storefront/pages/index.tsx
```diff
@@ -42,8 +52,20 @@ const HomePage: NextPage<ServerSidePropsType> = () => {
                 </Webline>
 
                 <Webline className="mb-6">
-                    <h2 className="mb-3">{t('Promoted products')}</h2>
-                    <PromotedProducts />
+                    {isLuigisBoxActive ? (
+                        <>
+                            <h2 className="mb-3">{t('Recommended for you')}</h2>
+                            <RecommendedProducts
+                                itemUuids={[]}
```
tvikito: Hmm, why not to set this empty array as default value in RecommendedProducts component? And don't send this prop if you don't have any data.

project-base/storefront/components/Blocks/Product/ProductsList/ProductListItem.tsx
sebaholesz: Yeah, no problem, renamed.

project-base/storefront/components/Blocks/Product/AddToCartPopup.tsx
sebaholesz: Yeah, you're right. Changed.

project-base/storefront/components/Blocks/Product/ProductsList/ProductsListContent.tsx
sebaholesz: No problem, renamed.

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
sebaholesz: I actually changed some of the HTML to be better suited for the place where its used. It should make more sense now.

project-base/storefront/pages/index.tsx
sebaholesz: Yeah, I can do that

project-base/storefront/components/Blocks/Product/AddToCartPopup.tsx
```diff
@@ -82,13 +82,16 @@ export const AddToCartPopup: FC<AddToCartPopupProps> = ({ key, addedCartItem: {
...
-                    <RecommendedProducts
-                        itemUuids={[product.uuid]}
-                        recommendationType={RecommendationTypeApi.BasketPopupApi}
-                    />
-                </div>
+                <RecommendedProducts
+                    itemUuids={[product.uuid]}
+                    recommendationType={RecommendationTypeApi.BasketPopupApi}
+                    render={(recommendedProductsContent) => (
+                        <div className="mb-6 px-0">
```
tvikito: px-0 does nothing

project-base/storefront/components/Blocks/Product/AddToCartPopup.tsx
```diff
@@ -82,13 +82,16 @@ export const AddToCartPopup: FC<AddToCartPopupProps> = ({ key, addedCartItem: {
...
-                    <div className="h2 mb-3">{t('Recommended for you')}</div>
-                    <RecommendedProducts
-                        itemUuids={[product.uuid]}
-                        recommendationType={RecommendationTypeApi.BasketPopupApi}
-                    />
-                </div>
+                <RecommendedProducts
+                    itemUuids={[product.uuid]}
+                    recommendationType={RecommendationTypeApi.BasketPopupApi}
+                    render={(recommendedProductsContent) => (
```
tvikito, line 88: I am not sure about this new concept (or at least new in the way that I haven't seen new place with this render implemented). What was the motivation behind going this way?

project-base/storefront/components/Blocks/Login/LoginPopup.tsx
tvikito, line 1: I understand that you took approach with removing only one component in Popup folder, but actually I like idea with having all popups in one place since they can be opened everywhere. What about rather move other popups into Popup folder?

project-base/storefront/store/slices/createPortalStateSlice.ts
tvikito, line 1: createPortalSlice should describe the file enough

project-base/storefront/store/slices/createPortalStateSlice.ts
```diff
@@ -0,0 +1,18 @@
...
+};
+
+export const createPortalStateSlice: StateCreator<PortalStateSlice> = (set) => ({
+    portalContent: null,
+
+    updatePortalContent: (updatedPortalContent) => {
+        set((s) => ({
+            ...s,
+            portalContent: updatedPortalContent,
+        }));
```
tvikito: image

project-base/storefront/store/useCookiesStore.ts
```diff
@@ -1,27 +1,47 @@
...
 export type CookiesStoreActions = {
     setCookiesStoreState: (value: Partial<CookiesStoreState>) => void;
 };
 
 export type CookiesStore = CookiesStoreState & CookiesStoreActions;
 
-export const defaultInitState: CookiesStoreState = {
-    lastVisitedProductsCatnums: undefined,
-};
+export const getDefaultInitState = (): CookiesStoreState => ({
```
tvikito: This function is used only in this file, no need to export this.

tvikito: Why this needs to be a function and not a simple object? It is because of the instance on the server, right? 🤔

project-base/storefront/store/useCookiesStore.ts
```diff
@@ -1,27 +1,47 @@
...
+export const createCookiesStoreOnServer = (cookieStoreStateFromContext: string | null) => {
+    let initState = getDefaultInitState();
+
+    if (cookieStoreStateFromContext) {
+        initState = JSON.parse(cookieStoreStateFromContext);
+    }
+
+    return createStore<CookiesStore>()((set) => ({
         ...initState,
         setCookiesStoreState: (value) => {
```
tvikito: Let's distinguish between those two `setCookiesStoreState` functions and make this `setCookiesStoreStateOnServer`

project-base/storefront/helpers/serverSide/initServerSideProps.ts
```diff
@@ -157,7 +158,9 @@ export const initServerSideProps = async <VariablesType extends Variables>({
             // JSON.parse(JSON.stringify()) fix of https://github.com/vercel/next.js/issues/11993
             urqlState: JSON.parse(JSON.stringify(currentSsrCache.extractData())),
             isMaintenance,
-            cookiesStore: getCookiesStore(context),
+            cookiesStore: serializeCookieStoreOnServer(
+                createCookiesStoreOnServer(getCookiesStoreStateFromContext(context)).getState(),
+            ),
```
tvikito: Hmm, I wonder what's the point of this `serializeCookieStoreOnServer ` function? So far I see that logic of this utility can be simplified to 

```suggestion
            cookiesStore: createCookiesStoreOnServer(getCookiesStoreStateFromContext(context)).getState()
```

project-base/storefront/store/useCookiesStore.ts
```diff
@@ -1,27 +1,47 @@
...
             set((state) => ({ ...state, ...value }));
         },
     }));
+};
```
tvikito: What about combine those two create stores into one function? 

```
export const createCookiesStore = ({
    cookieStoreFromServer,
    cookieStoreStateFromContext,
}: {
    cookieStoreFromServer?: CookiesStoreState;
    cookieStoreStateFromContext?: string | null;
}) => {
    let initState = cookieStoreFromServer || getDefaultInitState();

    if (cookieStoreStateFromContext) {
        initState = JSON.parse(cookieStoreStateFromContext);
    }

    return createStore<CookiesStore>()((set) => ({
        ...initState,
        setCookiesStoreState: (value) => {
            set((state) => ({ ...state, ...value }));
        },
    }));
};
```

project-base/storefront/helpers/cookies/cookiesStoreUtils.ts
```diff
@@ -22,14 +21,11 @@ export const useCookiesStoreSync = () => {
...
-    const isStoreWithValues = !!Object.values(store).filter((storeValue) => storeValue !== undefined).length;
-
-    if (isStoreWithValues) {
-        setCookie(COOKIES_STORE_NAME, store, { maxAge: THIRTY_DAYS_IN_SECONDS, ...options });
-    }
-};
+export const serializeCookieStoreOnServer = (cookieStore: CookiesStore): CookiesStoreState => ({
+    lastVisitedProductsCatnums: cookieStore.lastVisitedProductsCatnums,
+    userIdentifier: cookieStore.userIdentifier,
+});
```
tvikito: This functions seems redundant to me. If you agree on next comment about usage of this function let's remove this.

sebaholesz: Unfortunately, the state also contains the mutation function, which cannot be serialized, so if I just return that from the server as is, it fails.

project-base/storefront/helpers/serverSide/initServerSideProps.ts
sebaholesz: Unfortunately, the state also contains the mutation function, which cannot be serialized, so if I just return that from the server as is, it fails.

project-base/storefront/components/Blocks/Product/AddToCartPopup.tsx
sebaholesz: Removed

project-base/storefront/components/Blocks/Product/AddToCartPopup.tsx
sebaholesz: The idea is that we might want different text and wrapping component for each place. However, we want to only display the components (heading, wrapper) when there are actually products returned from the query. For that, we need to provide the wrapper from the outside, but control the displaying of it from the inside (depending on the query). This pattern solves it (same as for checkboxes, radiobuttons, etc.)

project-base/storefront/store/slices/createPortalStateSlice.ts
sebaholesz: No problem, renamed

project-base/storefront/store/slices/createPortalStateSlice.ts
sebaholesz: Nice, thanks, changed.

project-base/storefront/store/useCookiesStore.ts
sebaholesz: Thanks, removed the `export`

sebaholesz: Yes. If its an object, all customers have the same user identifier. This would behave differently if we had used edge runtime, but for node, there is only one instance of the server.

project-base/storefront/store/useCookiesStore.ts
sebaholesz: I actually renamed it to `setCookiesStoreStateOnClient` to make sure everybody knows it can only be used on the client and removed it from the server version, as it was not used. This also resolved the issue with serialization

project-base/storefront/helpers/serverSide/initServerSideProps.ts
sebaholesz: Actually, since I have removed the setter function, we do not need to serialize it manually and I was able to remove this function. So... resolved.

project-base/storefront/helpers/cookies/cookiesStoreUtils.ts
sebaholesz: Actually, since I have removed the setter function, we do not need to serialize it manually and I was able to remove this function. So... resolved.

project-base/storefront/store/useCookiesStore.ts
sebaholesz: With the recent changes, I do not see any benefit in it. Please, reconsider.

project-base/storefront/components/Blocks/Login/LoginPopup.tsx
sebaholesz: I have actually considered it myself. I have no problem with that. Done.

project-base/storefront/components/Blocks/Product/AddToCartPopup.tsx
tvikito: Hmm, actually I think for such a cases we use wrapper/content component. Don't we rather want to have it implemented the same way?

VaniaToper: Hi, I will try to write my opinion. I don't know if the concept is great in this case, I would have to read the whole code of pull request.
In generally on breno it's common practice to use this concept. You can find several places on the project, it brings a lot of flexibility for component, so I like it 👍

tvikito: Yes, we have already talk on Slack about this use case and I agree that's proper place to use this concept.

tvikito (approved):

PR #3110, opened 2024-04-16
[luigis-box] added parameter filter to Luigi's Box search

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| Introduced parametric filter on the search page when using Luigi's Box.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tl-luigis-box-parametric-filter.odin.shopsys.cloud
  - https://cz.tl-luigis-box-parametric-filter.odin.shopsys.cloud

### DIFF
--- a/docs/integration/luigis-box.md
+++ b/docs/integration/luigis-box.md
@@ -53,6 +53,11 @@ By default, these facets are used:
 -   `labels`
 -   `price_amount`
 
+### Parametric filter
+
+Parametric filter uses Luigi's Box AI for suggesting the best filter options for the user based on filtered products.
+You can read more about this in Luigi's Box [docs](https://docs.luigisbox.com/search/api.html#best-practices-use-dynamic-facets).
+
 ## Recommended products
 
 Luigi's Box uses data provided by searching and GTM to analyze user behavior and thanks to that provides personalized product recommendations.
--- a/packages/frontend-api/src/Model/Product/Filter/ProductFilterOptionsFactory.php
+++ b/packages/frontend-api/src/Model/Product/Filter/ProductFilterOptionsFactory.php
@@ -111,6 +111,24 @@ class ProductFilterOptionsFactory
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterConfig $productFilterConfig
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterCountData $productFilterCountData
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData $productFilterData
+     * @return \Shopsys\FrontendApiBundle\Model\Product\Filter\ProductFilterOptions
+     */
+    public function createFullProductFilterOptions(
+        ProductFilterConfig $productFilterConfig,
+        ProductFilterCountData $productFilterCountData,
+        ProductFilterData $productFilterData,
+    ): ProductFilterOptions {
+        $productFilterOptions = $this->createProductFilterOptions($productFilterConfig, $productFilterCountData, $productFilterData);
+        $this->fillBrands($productFilterOptions, $productFilterConfig, $productFilterCountData, $productFilterData);
+        $this->fillParameters($productFilterOptions, $productFilterConfig, $productFilterCountData, $productFilterData);
+
+        return $productFilterOptions;
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterConfig $productFilterConfig
      * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData $productFilterData
      * @param string $searchText
      * @return \Shopsys\FrontendApiBundle\Model\Product\Filter\ProductFilterOptions
--- a/packages/luigis-box/src/Component/LuigisBox/Filter/ProductFilterToLuigisBoxFilterMapper.php
+++ b/packages/luigis-box/src/Component/LuigisBox/Filter/ProductFilterToLuigisBoxFilterMapper.php
@@ -46,6 +46,7 @@ class ProductFilterToLuigisBoxFilterMapper
         $luigisBoxFilter = $this->mapAvailability($productFilterData, $luigisBoxFilter);
         $luigisBoxFilter = $this->mapFlags($productFilterData, $luigisBoxFilter, $domain->getLocale());
         $luigisBoxFilter = $this->mapBrands($productFilterData, $luigisBoxFilter);
+        $luigisBoxFilter = $this->mapParameters($productFilterData, $luigisBoxFilter);
 
         return $luigisBoxFilter;
     }
@@ -115,10 +116,8 @@ class ProductFilterToLuigisBoxFilterMapper
      */
     protected function mapFlags(ProductFilterData $productFilterData, array $luigisBoxFilter, string $locale): array
     {
-        if (count($productFilterData->flags) > 0) {
-            foreach ($productFilterData->flags as $flag) {
-                $luigisBoxFilter[self::FILTER_OR][] = 'labels:' . $flag->getName($locale);
-            }
+        foreach ($productFilterData->flags as $flag) {
+            $luigisBoxFilter[self::FILTER_OR][] = 'labels:' . $flag->getName($locale);
         }
 
         return $luigisBoxFilter;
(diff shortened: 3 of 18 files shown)


packages/frontend-api/src/Model/Product/Filter/ProductFilterOptionsFactory.php
```diff
@@ -109,6 +109,24 @@ public function createProductFilterOptions(
...
+     * @return \Shopsys\FrontendApiBundle\Model\Product\Filter\ProductFilterOptions
+     */
+    public function createFullProductFilterOptions(
+        ProductFilterConfig $productFilterConfig,
+        ProductFilterCountData $productFilterCountData,
+        ProductFilterData $productFilterData,
+    ): ProductFilterOptions {
+        $productFilterOptions = $this->createProductFilterOptions($productFilterConfig, $productFilterCountData, $productFilterData);
+        $this->fillBrands($productFilterOptions, $productFilterConfig, $productFilterCountData, $productFilterData);
+        self::fillParameters($productFilterOptions, $productFilterConfig, $productFilterCountData, $productFilterData);
```
vitek-rostislav:
```suggestion
        $this->fillParameters($productFilterOptions, $productFilterConfig, $productFilterCountData, $productFilterData);
```

`fillParameters` is not static :wink:

vitek-rostislav: Also, I see the `fillParameters` method is overridden in `project-base` where `$category` parameter is required :sob:

packages/frontend-api/src/Model/Product/Filter/ProductFilterOptionsFactory.php
```diff
@@ -109,6 +109,24 @@ public function createProductFilterOptions(
         return $productFilterOptions;
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterConfig $productFilterConfig
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterCountData $productFilterCountData
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData $productFilterData
+     * @return \Shopsys\FrontendApiBundle\Model\Product\Filter\ProductFilterOptions
+     */
+    public function createFullProductFilterOptions(
```
vitek-rostislav, line 118: are we ok with the fact the method is used from luigis-box package only? Let' discuss it please

packages/luigis-box/src/Component/LuigisBox/Filter/ProductFilterToLuigisBoxFilterMapper.php
```diff
@@ -139,4 +140,22 @@ protected function mapBrands(ProductFilterData $productFilterData, array $luigis
...
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterData $productFilterData
+     * @param array $luigisBoxFilter
+     * @return array
+     */
+    protected function mapParameters(ProductFilterData $productFilterData, array $luigisBoxFilter): array
+    {
+        if (count($productFilterData->parameters) > 0) {
```
vitek-rostislav: Is this condition necessary? You can just use `foreach ($productFilterData->parameters as $parameterFilterData) {` and if there are no parameters, it would not do anything. Or am I missing something?

docs/integration/luigis-box.md
```diff
@@ -53,6 +53,13 @@ By default, these facets are used:
 -   `labels`
 -   `price_amount`
 
+### Parametric filter
+
+Parametric filter uses Luigi's Box AI for suggesting the best filter options for the user based on filtered products.
+You can read more about this in Luigi's Box [docs](https://docs.luigisbox.com/search/api.html#best-practices-use-dynamic-facets).
+By default, is value `dynamic_facets_size` set to `5`.
+You can change this value by overwriting `LuigisBoxClient` class and changing the value of `COUNT_OF_DYNAMIC_PARAMETER_FILTERS` constant.
```
vitek-rostislav: I would add something like this:
```
You can also specify the particular filters you want to show on the search results page in `LuigisBoxClient::addSearchSpecificParametersToUrl()`
```

packages/luigis-box/src/Model/Product/Filter/LuigisBoxFacetsToProductFilterOptionsMapper.php
```diff
@@ -50,37 +60,109 @@ public function map(array $luigisBoxFacets, ProductFilterData $productFilterData
...
+     * @param array $brands
+     * @param \Shopsys\FrameworkBundle\Model\Product\Filter\ProductFilterCountData $productFilterCountData
+     * @return array|\Shopsys\FrameworkBundle\Model\Product\Brand\Brand[]
+     */
+    protected function mapBrands(
+        array $facetData,
+        array $brands,
+        ProductFilterCountData $productFilterCountData,
+    ): array {
+        if ($facetData['name'] === self::FACET_BRAND) {
```
vitek-rostislav: nitpick, I would appreciate an early return here:
```
 if ($facetData['name'] !== self::FACET_BRAND) {
    return $brands;
}

$flagCountsByName = $this->mapValuesToCountsByName($facetData['values']);
// ...
```
It is better readable this way IMHO

vitek-rostislav: the same applies to `mapFlags()` and `mapPriceRange()` functions

packages/luigis-box/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\LuigisBoxBundle\Model\Product\Parameter;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterValue;
+
+class ParameterRepository
```
vitek-rostislav: at the moment, we have 4 classes named `ParameterRepository` in the project, I do not like that, to be honest.

I do not think you will like my suggestion but could we rename this class to something else, please? :smile: 
What about these?
- `LuigisBoxParameterRepository`
- `ParameterValueRepository`
- `ParameterValueProvider`
- `ParameterProvider`

packages/luigis-box/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @param \Doctrine\ORM\EntityManagerInterface $em
+     */
+    public function __construct(
+        protected readonly EntityManagerInterface $em,
+    ) {
+    }
+
+    /**
+     * @param array $parameterValues
```
vitek-rostislav: this could be annotated as `string[]`, right?

packages/luigis-box/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -0,0 +1,40 @@
...
+        $parameterValues = $this->em->createQueryBuilder()
+            ->select('pv')
+            ->from(ParameterValue::class, 'pv')
+            ->where('pv.text IN(:parameterValues)')
+            ->andWhere('pv.locale = :locale')
+            ->setParameters([
+                'parameterValues' => $parameterValues,
+                'locale' => $locale,
+            ])
+            ->orderBy('pv.text')->getQuery()->getResult();
```
vitek-rostislav: I believe we should use `OrderByCollationHelper::createOrderByForLocale` here

vitek-rostislav (changes requested):

docs/integration/luigis-box.md
TomasLudvik: We have discussed this and I have updated it in a bit different way

packages/luigis-box/src/Model/Product/Filter/LuigisBoxFacetsToProductFilterOptionsMapper.php
TomasLudvik: Good catch!

packages/luigis-box/src/Model/Product/Parameter/ParameterRepository.php
TomasLudvik: I have renamed it to ParameterValueRepository

vitek-rostislav (approved):

PR #3118, opened 2024-04-19
[frontend-api] applied promo code is now taken into account in priceByTransportQuery and priceByPaymentQuery

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Transport and payment price is affected by total products price (including discounts) as there might be limit for free transport and payment. However, the promo code was omitted when the transport/payment price was calculated
|New feature| No 
|Fixes issues| ... 

----
  - https://rv-free-transport.odin.shopsys.cloud
  - https://cz.rv-free-transport.odin.shopsys.cloud

### DIFF
--- a/project-base/app/src/FrontendApi/Resolver/Price/PriceQuery.php
+++ b/project-base/app/src/FrontendApi/Resolver/Price/PriceQuery.php
@@ -1,0 +1,133 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\FrontendApi\Resolver\Price;
+
+use ArrayObject;
+use Shopsys\FrameworkBundle\Model\Payment\Payment;
+use Shopsys\FrameworkBundle\Model\Pricing\Price;
+use Shopsys\FrameworkBundle\Model\Transport\Transport;
+use Shopsys\FrontendApiBundle\Component\GqlContext\GqlContextHelper;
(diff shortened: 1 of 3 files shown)


project-base/app/src/FrontendApi/Resolver/Price/PriceQuery.php
```diff
@@ -0,0 +1,130 @@
...
+        );
+    }
+
+    /**
+     * @param \App\Model\Payment\Payment $payment
+     * @param string|null $cartUuid
+     * @param \ArrayObject|null $context
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\Price
+     */
+    public function priceByPaymentQuery(
```
grossmannmartin, line 80: I'm not sure about the paymentquery – the only difference against the original implementation is the @var annotation near cart variable.

grossmannmartin (changes requested):

project-base/app/src/FrontendApi/Resolver/Price/PriceQuery.php
vitek-rostislav: ah, thanks for pointing that out, it should have been like this: https://github.com/shopsys/shopsys/pull/3118/commits/e606c8e0d1246944c10117176e515f7b7bb8fb29

grossmannmartin (approved):

PR #3126, opened 2024-04-23
[frontend-api] addProductToListMutation: ensure new product list is created with non-conflicting uuid

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When a customer user is logged out due to the long inactivity, his product list uuids might be kept in the storage, and adding products to the list fails. Therefore, when an anonymous product list with a given uuid is not found, we still need to check whether a customer user's product list with the given uuid exists. If so, a new anonymous product list must be created with another uuid to avoid unique constraint violations. The same applies when the customer user's product list is not found by the given uuid -> we need to check for the existence of an anonymous list with the uuid before creating the new customer user's list
|New feature| No 
|Fixes issues| ... 

----
  - https://rv-product-list-fix.odin.shopsys.cloud
  - https://cz.rv-product-list-fix.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Model/Product/List/ProductListRepository.php
+++ b/packages/framework/src/Model/Product/List/ProductListRepository.php
@@ -119,4 +119,13 @@ class ProductListRepository
     {
         return $this->entityManager->getRepository(ProductList::class);
     }
+
+    /**
+     * @param string $uuid
+     * @return bool
+     */
+    public function existsProductListWithUuid(string $uuid): bool
+    {
+        return $this->getRepository()->count(['uuid' => $uuid]) > 0;
+    }
 }
--- a/project-base/app/tests/FrontendApiBundle/Functional/Product/ProductList/ProductListLoggedCustomerTest.php
+++ b/project-base/app/tests/FrontendApiBundle/Functional/Product/ProductList/ProductListLoggedCustomerTest.php
@@ -151,6 +151,28 @@ class ProductListLoggedCustomerTest extends GraphQlWithLoginTestCase
     }
 
     /**
+     * @dataProvider \Tests\FrontendApiBundle\Functional\Product\ProductList\ProductListTypesDataProvider::getProductListTypes
+     * @param string $productListType
+     */
+    public function testAddProductCreatesNewListWithNewUuidWhenUuidOfAnonymousListIsProvided(
+        string $productListType,
+    ): void {
+        $anonymousProductListUuid = $this->getAnonymousProductListUuid($productListType);
+        $productToAddId = 69;
+        $productToAdd = $this->getReference(ProductDataFixture::PRODUCT_PREFIX . $productToAddId);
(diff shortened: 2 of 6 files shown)


packages/framework/src/Model/Product/List/ProductListRepository.php
```diff
@@ -119,4 +119,13 @@ protected function getRepository(): EntityRepository
...
     }
+
+    /**
+     * @param string $uuid
+     * @return \Shopsys\FrameworkBundle\Model\Product\List\ProductList|null
+     */
+    public function findProductListByUuid(string $uuid): ?ProductList
+    {
+        return $this->getRepository()->findOneBy(['uuid' => $uuid]);
+    }
```
grossmannmartin: this method is used only for checking the presence of the product list, but we're not interested in the data, right? 
What about using `$this->getRepository()->count()` to avoid unnecessary entity hydration?

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductList/ProductListLoggedCustomerTest.php
```diff
@@ -150,6 +150,30 @@ public function testAddProductCreatesNewListWhenNewUuidIsProvided(string $produc
...
+    /**
+     * @dataProvider \Tests\FrontendApiBundle\Functional\Product\ProductList\ProductListTypesDataProvider::getProductListTypes
+     * @param string $productListType
+     */
+    public function testAddProductCreatesNewListWithNewUuidWhenUuidOfAnonymousListIsProvided(
+        string $productListType,
+    ): void {
+        $anonymousProductListUuid = $productListType === ProductListTypeEnum::COMPARISON
+            ? ProductListDataFixture::PRODUCT_LIST_COMPARISON_NOT_LOGGED_CUSTOMER_UUID
+            : ProductListDataFixture::PRODUCT_LIST_WISHLIST_NOT_LOGGED_CUSTOMER_UUID;
```
grossmannmartin: adding a new type of list would require changing this condition. What about rewriting it to `match`, so it will be easier to cover the new product list with the test?

grossmannmartin (changes requested):

packages/framework/src/Model/Product/List/ProductListRepository.php
vitek-rostislav: good idea! c81bd035ba

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductList/ProductListLoggedCustomerTest.php
vitek-rostislav: another great idea, thanks! 3267f6668f

grossmannmartin (approved):

PR #3128, opened 2024-04-24
[shopsys] display error messages in admin if legal conditions articles are not set

AndrejBlaho:
| Q             | A
| ------------- | ---
|Description, reason for the PR| display error messages in admin if legal conditions articles are not set
|New feature| Yes 
|Fixes issues| ... 

----
  - https://ab-empty-articles-warnings.odin.shopsys.cloud
  - https://cz.ab-empty-articles-warnings.odin.shopsys.cloud


packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -37,12 +37,21 @@ msgstr "-- Vyberte název převodu --"
 msgid "0"
 msgstr ""
 
+msgid "<a href=\"{{ url }}\">Cookies information article for domain {{ domainName}} is not set.</a>"
```
grossmannmartin:
```suggestion
msgid "<a href=\"{{ url }}\">Cookies information article for domain {{ domainName }} is not set.</a>"
```

you will need to change the source and also msgstrs and msgids in this and english PO file

packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -56,6 +60,8 @@ public function __construct(
         protected readonly BreadcrumbOverrider $breadcrumbOverrider,
         protected readonly DateTimeFormatterExtension $dateTimeFormatterExtension,
         protected readonly TransferIssueFacade $transferIssueFacade,
+        protected readonly LegalConditionsFacade $legalConditionsFacade,
+        protected readonly Domain $domain,
```
grossmannmartin: This is considered a BC break (see https://symfony.com/doc/current/contributing/code/bc.html#changing-classes). 
Can you please document the change in the upgrade notes and update the PR description (BC Break: yes)?

packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -160,6 +166,48 @@ protected function addWarningMessagesOnDashboard(): void
...
+            $domain = $this->domain->getDomainConfigById($notSetTermsAndConditions->getDomainId());
+
+            $this->addErrorFlashTwig(
+                t('<a href="{{ url }}">Terms and conditions article for domain {{ domainName}} is not set.</a>'),
+                [
+                    'url' => $this->generateUrl('admin_legalconditions_termsandconditions'),
+                    'domainName' => $domain->getName(),
+                ],
+            );
+        }
```
grossmannmartin: What about 

```
foreach ($this->domain->getAll() as $domainConfig) {
    if ($this->setting->getForDomain(Setting::TERMS_AND_CONDITIONS_ARTICLE_ID, $domainConfig->getId()) === null) {
        $this->addErrorFlashTwig(
            t('<a href="{{ url }}">Terms and conditions article for domain {{ domainName}} is not set.</a>'),
            [
                'url' => $this->generateUrl('admin_legalconditions_termsandconditions'),
                'domainName' => $domainConfig->getName(),
            ],
        );
    }
    
     // if for privacy policy
     
     // if for cookie article
}
```

That way you don't need to create a new method in LegalConditionsFacade (and you don't need to autowire it into this controller), SettingValueRepository may remain unchanged. 
As a bonus, you save a few DB queries as the setting values are loaded at once (and at this point are already loaded).

packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -160,6 +166,48 @@ protected function addWarningMessagesOnDashboard(): void
...
+
+        $notSetTermsAndConditionsArticles = $this->legalConditionsFacade->findNullSettingValues(Setting::TERMS_AND_CONDITIONS_ARTICLE_ID);
+
+        foreach ($notSetTermsAndConditionsArticles as $notSetTermsAndConditions) {
+            $domain = $this->domain->getDomainConfigById($notSetTermsAndConditions->getDomainId());
+
+            $this->addErrorFlashTwig(
+                t('<a href="{{ url }}">Terms and conditions article for domain {{ domainName}} is not set.</a>'),
+                [
+                    'url' => $this->generateUrl('admin_legalconditions_termsandconditions'),
```
grossmannmartin: If the article is not set on several domains, the link will be always the same because the domain can be switched only globally. 
I don't think that's a big problem, just wanted to point it out.

grossmannmartin (changes requested):

packages/framework/src/Resources/translations/messages.cs.po
AndrejBlaho: Fixed.

packages/framework/src/Controller/Admin/DefaultController.php
AndrejBlaho: Both done.

packages/framework/src/Controller/Admin/DefaultController.php
AndrejBlaho: Oh i see, that is much more elegant approach. Thank you, rewrote it as such.

packages/framework/src/Controller/Admin/DefaultController.php
AndrejBlaho: Noticed that. Should i do something with it or is it ok to keep it as it is?

grossmannmartin (approved):

packages/framework/src/Controller/Admin/DefaultController.php
grossmannmartin: Let's keep it as it is... I think it's ok :)

pk16011990 (approved):

PR #3132, opened 2024-04-25
redesign product page + some base elements

tvikito:
| Q             | A
| ------------- | ---
|Description, reason for the PR| redesign product page + some base elements
|New feature| No 
|Fixes issues| ... 

----
  - https://tv-ssp-2250-redesign-product-page.odin.shopsys.cloud
  - https://cz.tv-ssp-2250-redesign-product-page.odin.shopsys.cloud


project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailability.tsx
```diff
@@ -1,46 +1,44 @@
...
-            >
-                {product.availability.name}
-                <ArrowIcon className="ml-1 text-dark" />
-            </div>
-
+        <>
             {!!product.availableStoresCount && (
-                <span className="mr-1 text-sm">
-                    {t('This item is available immediately in {{ count }} stores', {
+                <span
```
VaniaToper: I think there should be a button instead of a span if it has an onClick

project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailability.tsx
```diff
@@ -1,46 +1,44 @@
...
+                    className={twJoin(
+                        'mr-1 text-sm flex cursor-pointer items-center no-underline hover:no-underline',
+                        product.availability.status === TypeAvailabilityStatusEnum.InStock &&
+                            'text-secondary hover:text-secondary',
```
VaniaToper: I like that there is a hover effect, but what about adding an active effect?
That's a general question for the whole PR

project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailability.tsx
```diff
@@ -1,46 +1,44 @@
...
-                    product.availability.status === TypeAvailabilityStatusEnum.OutOfStock && 'text-red hover:text-red',
-                )}
-                onClick={scrollOnClickHandler}
-            >
-                {product.availability.name}
-                <ArrowIcon className="ml-1 text-dark" />
-            </div>
-
+        <>
             {!!product.availableStoresCount && (
```
VaniaToper: I'd rather use return on the top instead of condition on the whole component
`if (!product.availableStoresCount) {
    return null
}`
....

project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailabilityList.tsx
```diff
@@ -3,55 +3,50 @@ import { ArrowRightIcon } from 'components/Basic/Icon/ArrowRightIcon';
...
+                                    {t('Store detail')}
+                                    <ArrowRightIcon className="ml-1" />
+                                </ExtendedNextLink>
+                            </li>
+                        ),
+                )}
+            </ul>
+        </div>
+    );
+};
```
VaniaToper: I think we will want to have some cleverer solution with some load more and map etc. But for the start is enought 👍

project-base/storefront/styles/globals.css
```diff
@@ -190,7 +200,7 @@ input[type='number'] {
 
 .Toastify__toast-theme--colored.Toastify__toast--success
     .Toastify__progress-bar-theme--colored.Toastify__progress-bar--success {
-    @apply !bg-greenLight;
+    @apply !bg-[#81f7da];
```
VaniaToper: I would rather define some green color and use it there

project-base/storefront/styles/globals.css
```diff
@@ -203,7 +213,7 @@ input[type='number'] {
 
 .Toastify__toast-theme--colored.Toastify__toast--error
     .Toastify__progress-bar-theme--colored.Toastify__progress-bar--error {
-    @apply !bg-redLight;
+    @apply !bg-[#f2a2a2];
```
VaniaToper: Same there with red color

VaniaToper (commented):
Good work 👍

project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailabilityList.tsx
tvikito: Yes, that's not part of this scope.

project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailability.tsx
tvikito: Yeah, good point! Refactored.

project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailability.tsx
tvikito: Yeah! Nice catch. Much better.

project-base/storefront/styles/globals.css
tvikito: This is not part of the scope which should be solved here, because there is no design for Toasts yet. But I solved it temporarily by some adjustments. Let's wait with other changes for new design then we can adjust.

project-base/storefront/styles/globals.css
tvikito: This is not part of the scope which should be solved here, because there is no design for Toasts yet. But I solved it temporarily by some adjustments. Let's wait with other changes for new design then we can adjust.

project-base/storefront/components/Pages/ProductDetail/ProductDetailAvailability.tsx
tvikito: It was agreed that we will leave it until we have design solution for this.

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
```diff
@@ -62,25 +60,41 @@ export const ProductDetailContent: FC<ProductDetailContentProps> = ({ product, f
...
                                 {formatPrice(product.price.priceWithVat)}
                             </div>
-                            <DeferredProductDetailAddToCart product={product} />
-                        </div>
 
-                        <ProductDetailAvailability product={product} scrollTarget={scrollTarget} />
+                            <ProductDetailAvailability product={product} />
 
-                        <DeferredComparisonAndWishlistButtons product={product} />
+                            <ProductDetailAddToCart product={product} />
```
sebaholesz: Here you want to use `DeferredProductDetailAddToCart`.

sebaholesz (changes requested):

sebaholesz (approved):

PR #3144, opened 2024-04-30
[framework] load products iteratively while generating image sitemaps

### LINKED ISSUE #2618: Image sitemap generating fail on memory
## Describe the bug

When generating the image sitemap, all offered products are loaded into memory at once. (See https://github.com/shopsys/shopsys/blob/master/packages/framework/src/Model/ImageSitemap/ImageSitemapFacade.php#L79). On our project (~20k products), this started to fail on `OutOfMemoryError`.
### To Reproduce
1. Load 20.000 products into your database
2. Run `php bin/console shopsys:cron --module="App\Model\ImageSitemap\ImageSitemapCronModule" --instance-name=export`

### Expected behavior
The sitemap should be generated regardless of the amount of products.

AndrejBlaho:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Image sitemap generation was failing on memory if there was large number of products loaded at once. Now products are loaded iteratively which should help with the issue. Rewriting cron into iterative one would solve the problem completely regardless of data size, however Presta bundle does not support iterative data load, so we decided to go with this compromise for now.
|New feature| No 
|Fixes issues| closes #2618 

----
  - https://ab-generate-image-sitemap-iteratively.odin.shopsys.cloud
  - https://cz.ab-generate-image-sitemap-iteratively.odin.shopsys.cloud

### DIFF
--- a/UPGRADE-15.0.md
+++ b/UPGRADE-15.0.md
@@ -826,6 +826,12 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 
 -   see #project-base-diff to update your project
 
+#### load products iteratively while generating image sitemaps ([#3144](https://github.com/shopsys/shopsys/pull/3144))
+
+-   `Shopsys\FrameworkBundle\Model\Product\ProductRepository` class was changed:
+    -   `getAllOfferedProducts()` method was removed, use `getAllOfferedProductsPaginated()` instead
+-   see #project-base-diff to update your project
+
 ### Storefront
 
 #### added query/mutation name to URL and headers ([#3041](https://github.com/shopsys/shopsys/pull/3041))
--- a/packages/framework/src/Model/ImageSitemap/ImageSitemapFacade.php
+++ b/packages/framework/src/Model/ImageSitemap/ImageSitemapFacade.php
@@ -15,6 +15,8 @@ use Shopsys\FrameworkBundle\Model\Product\ProductRepository;
 
 class ImageSitemapFacade
 {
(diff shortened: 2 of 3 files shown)


UPGRADE-15.0.md
```diff
@@ -810,6 +810,12 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     +       protected readonly ProductRecalculationDispatcher $productRecalculationDispatcher,
     ```
 
+#### load products iteratively while generating image sitemaps ([#3144](https://github.com/shopsys/shopsys/pull/3144))
+
+-   `Shopsys\FrameworkBundle\Model\Product\ProductRepository` class was changed:
+    -   `getAllOfferedProducts()` method was removed
```
vitek-rostislav:
```suggestion
    -   `getAllOfferedProducts()` method was removed, use `getAllOfferedProductsPaginated()` instead
```

vitek-rostislav (approved):

vitek-rostislav:
:heavy_check_mark: Tested locally with ~40k products in the DB, the problem is solved :ok_hand: In the **production** mode it works like a charm. The truth is, in dev mode, the generating fails on the memory exhaustion but the reason is the doctrine backtrace collecting that was enabled in https://github.com/shopsys/shopsys/pull/3055. 

So after this tweak, to be able to generate the sitemap **locally**, you need either switch to the production mode or set `profiling_collect_backtrace` to `false` in `packages/dev/doctrine.yaml`

PR #3145, opened 2024-04-30
[project-base] [SSP-2290] cookies store smarter init

sebaholesz:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Cookies store was initialized using very simple logic, as it was either created if there was no existing value, or the current value was kept. This did not handle scenarios where the store is partially correct, e.g. with some properties missing or some being there even though they are not used anymore. This PR implements smarter init logic so that these cases are handled.
|New feature| No
|Fixes issues| ... 

----
  - https://sh-ssp-2290-cookies-store-smarter-init.odin.shopsys.cloud
  - https://cz.sh-ssp-2290-cookies-store-smarter-init.odin.shopsys.cloud


project-base/storefront/utils/cookies/cookiesStore.ts
```diff
@@ -1,32 +1,82 @@
...
+        return newState;
+    }
+
+    return removeIncorrectCookiesStoreProperties(
+        Object.keys(newState),
+        addMissingCookiesStoreProperties(newState, JSON.parse(decodeURIComponent(cookiesStore))),
+    );
+};
+
+const addMissingCookiesStoreProperties = (
```
VaniaToper, line 45: I'm not sure if this function does what it's name says
It's actually only responsible for merging the previous state cookie store with the current one, so I don't see it getting the missing properties somewhere inside 🤔

VaniaToper (approved):

pk16011990 (approved):

PR #3157, opened 2024-05-08
[shopsys] a little spring cleanup

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There are a lot of features implemented in the project-base. Eventually, we want to get rid of the features in project-base, they should be moved to the appropriate packages. When analyzing the whole repository, I came across features/pieces of code, that I was able to move right away without additional effort so I did that, and the result is this PR.
|New feature| No 
|Fixes issues| ... 

----
  - https://rv-cleanup.odin.shopsys.cloud
  - https://cz.rv-cleanup.odin.shopsys.cloud


packages/frontend-api/src/Component/HttpFoundation/TransactionalMasterRequestListener.php
```diff
@@ -12,8 +12,8 @@
 
 class TransactionalMasterRequestListener extends BaseTransactionalMasterRequestListener
```
grossmannmartin: I'm not sure about this listener in frontend-api bundle. 
If anyone needs to extend the listener, it's really easy to forget about this and unintentionally add back the transactional behavior to graphql queries

project-base/app/templates/Admin/Content/OrderStatus/listGrid.html.twig
```diff
@@ -1,39 +0,0 @@
-{% extends '@ShopsysFramework/Admin/Grid/Grid.html.twig' %}
-
-{% block grid_action_cell_type_delete %}
-    {% if row.os.type == TYPE_NEW %}
-        <span title="{{ 'This status can\'t be deleted. This is automatically assigned status to all new orders.'|trans }}"
-            class="js-tooltip cursor-help in-icon in-icon--info" data-toggle="tooltip" data-placement="top"
-        >
-            {{ icon('info') }}
```
grossmannmartin, line 8: The parent uses `<i>` tag to render the icon. Shouldn't it be changed in the parent template?

grossmannmartin (changes requested):

packages/frontend-api/src/Component/HttpFoundation/TransactionalMasterRequestListener.php
vitek-rostislav: I was kind of struggling with this decision, you made a good point, I would love to move the logic to the framework itself, however, it means I have to require `webonyx/graphql-php` in the `framework` package - 8a274ebf71

project-base/app/templates/Admin/Content/OrderStatus/listGrid.html.twig
vitek-rostislav: I was looking into it and I found out that these notations are equivalent. However, I changed it to `{{ icon('info') }}` in the parent template in [cbee014](https://github.com/shopsys/shopsys/pull/3157/commits/cbee014426bd5aa977833fe8eba45a5b42edf197)

packages/frontend-api/src/Component/HttpFoundation/TransactionalMasterRequestListener.php
vitek-rostislav: refactored as discussed in b89a80c588

grossmannmartin (approved):

PR #3158, opened 2024-05-09
[administration] Prototyping Sonata

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| We are researching the possibility of replacing our current administration with some easy-to-generate administration to decrease the time consumption of new features. Sonata seems to be a viable solution, so we are prototyping it and finding out if it fits our needs.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tl-generated-admin.odin.shopsys.cloud
  - https://cz.tl-generated-admin.odin.shopsys.cloud

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -261,7 +261,7 @@ jobs:
             -   name: Check standards
                 run: |
                     docker compose exec -T php-fpm touch ./project-base/app/DEVELOPMENT
-                    docker compose exec -T php-fpm php phing -D production.confirm.action=y composer-dev standards
+                    docker compose exec -T php-fpm php phing -D production.confirm.action=y composer-dev frontend-api-generate-new-keys standards
             -   name: Check GraphQl schema
                 run: docker compose exec -T php-fpm project-base/app/check-schema.sh
     standards-storefront:
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,7 @@
         "psr-4": {
             "App\\": "project-base/app/src/",
             "Overblog\\GraphQLBundle\\__DEFINITIONS__\\": "project-base/app/var/overblogCompiledClasses",
+            "Shopsys\\Administration\\": "packages/administration/src/",
             "Shopsys\\CodingStandards\\": "packages/coding-standards/src/",
             "Shopsys\\FormTypesBundle\\": "packages/form-types-bundle/src/",
             "Shopsys\\FrontendApiBundle\\": "packages/frontend-api/src/",
@@ -39,6 +40,7 @@
     "autoload-dev": {
         "psr-4": {
             "Tests\\": "project-base/app/tests/",
+            "Tests\\Administration\\": "packages/administration/tests/",
             "Tests\\CodingStandards\\": "packages/coding-standards/tests/",
             "Tests\\FrameworkBundle\\": "packages/framework/tests/",
             "Tests\\FrontendApiBundle\\": "packages/frontend-api/tests/",
@@ -84,6 +86,7 @@
         "barryvdh/elfinder-flysystem-driver": "^0.4.3",
         "commerceguys/intl": "^1.0.0",
         "composer/composer": "^2.2.12",
+        "cweagans/composer-patches": "^1.7",
         "defuse/php-encryption": "^2.2.1",
         "doctrine/annotations": "^1.6.0",
         "doctrine/cache": "^1.7",
@@ -96,7 +99,7 @@
         "doctrine/doctrine-migrations-bundle": "^3.2.2",
         "doctrine/migrations": "^3.4.1",
         "doctrine/orm": "^2.11.2",
-        "doctrine/persistence": "^2.4",
+        "doctrine/persistence": "^3.3",
         "elasticsearch/elasticsearch": "^7.6.1",
         "enlightn/security-checker": "^1.3",
         "fakerphp/faker": "^1.19.0",
@@ -127,12 +130,12 @@
         "overblog/dataloader-bundle": "^0.6.0",
         "overblog/graphiql-bundle": "^0.3",
         "overblog/graphql-bundle": "1.3.2",
-        "phing/phing": "3.0.0-rc6",
+        "phing/phing": "^3.0.0",
         "phpdocumentor/reflection-docblock": "^5.3.0",
         "presta/sitemap-bundle": "^3.3",
         "prezent/doctrine-translatable": "^3.3.0",
         "prezent/doctrine-translatable-bundle": "^1.4",
-        "psr/log": "^1.0",
+        "psr/log": "^2.0",
         "ramsey/uuid": "^4.3.1",
         "roave/better-reflection": "^6.0",
         "scheb/2fa-bundle": "^5.7",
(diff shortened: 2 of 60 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -261,7 +261,7 @@ jobs:
             -   name: Check standards
                 run: |
                     docker compose exec -T php-fpm touch ./project-base/app/DEVELOPMENT
-                    docker compose exec -T php-fpm php phing -D production.confirm.action=y composer-dev standards
+                    docker compose exec -T php-fpm php phing -D production.confirm.action=y composer-dev frontend-api-generate-new-keys standards
```
grossmannmartin, line 264: Why are the frontend api keys suddenly needed for standards?

composer.json
```diff
@@ -257,9 +263,17 @@
         },
         "runtime": {
             "dotenv_path": "project-base/app/.env"
+        },
+        "enable-patching": true,
+        "patches": {
+            "composer-exit-on-patch-failure": true,
+            "twig/twig": {
+                "Hotfix for issue https://github.com/sonata-project/SonataAdminBundle/issues/8181": "https://github.com/TomasLudvik/Twig/commit/4e07748eb6d427688221e2c7b6dc8872c6d9065c.patch"
```
grossmannmartin: Shouldn't be this patch file part of the shopsys/shopsys repository?

project-base/app/translations/validators.cs.po
```diff
@@ -34,6 +34,9 @@ msgstr "Barva příznaku musí mít správný formát hexadecimálního kódu, n
 msgid "Flag name cannot be longer than {{ limit }} characters"
 msgstr "Název příznaku nemůže být delší než {{ limit }} znaků"
 
+msgid "Full name cannot be longer than {{ limit }} characters"
+msgstr ""
```
grossmannmartin: Can you fill in czech translations? Otherwise, I'm afraid we forget it.

project-base/app/config/packages/sonata_block.yaml
```diff
@@ -0,0 +1,2 @@
+sonata_block:
+    http_cache: false
```
grossmannmartin: can we unify all sonata configurations to `sonata_admin.yaml` file?
Now we have sonata_block definition here and in `sonata_admin.yaml`. 

Same applies also to `sonata_form.yaml`

project-base/app/config/routes/sonata_admin.yaml
```diff
@@ -0,0 +1,12 @@
+
+_sonata_admin:
+    prefix: /admin-new
```
grossmannmartin: in the future, we should take into account the `admin_url` parameter

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -154,6 +154,9 @@ msgstr "Administrátor <strong>{{ name }}</strong> je jediný a nemůže být sm
 msgid "Administrator activities history"
 msgstr "Historie aktivit administrátora"
 
+msgid "Administrator with ID: %s"
+msgstr ""
```
grossmannmartin: Can you fill in Czech translations? Otherwise, I'm afraid we forget it.

packages/framework/src/Model/Order/Order.php
```diff
@@ -1125,4 +1125,12 @@ public function getTrackingUrl(): ?string
             OrderMail::TRANSPORT_VARIABLE_TRACKING_NUMBER => $trackingNumber,
         ]);
     }
+
+    /**
+     * @return string
+     */
+    public function __toString(): string
+    {
+        return sprintf('Order with id: %s and number: %s', $this->id, $this->number);
```
grossmannmartin: the __toString method will be mandatory for all entities in administration?

packages/framework/src/Component/Domain/Domain.php
```diff
@@ -207,4 +207,18 @@ public function getDateTimeZone(): DateTimeZone
     {
         return $this->getCurrentDomainConfig()->getDateTimeZone();
     }
+
+    /**
+     * @return array<int, string>
+     */
+    public function getAllDomainsAsChoices(): array
```
grossmannmartin, line 214: When it's named `...asChoice`, shouldn't it be already in format required for choice? 
Now it's necessary to call array_flip

packages/framework/composer.json
```diff
@@ -75,7 +75,7 @@
         "presta/sitemap-bundle": "^3.3",
         "prezent/doctrine-translatable": "^3.3.0",
         "prezent/doctrine-translatable-bundle": "^1.4",
-        "psr/log": "^1.0",
+        "psr/log": "^2.0",
```
grossmannmartin, line 78: Should be this change also in packages/migration?

packages/administration/src/ShopsysAdministrationBundle.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+class ShopsysAdministrationBundle extends Bundle
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function build(ContainerBuilder $container): void
```
grossmannmartin: this method seems unnecessary

grossmannmartin (changes requested):

packages/framework/src/Model/Order/Order.php
henzigo: If you don't implement `__toString()` method, you will see Entity with some hash in administration as title or in success/error message

.github/workflows/docker-build.yaml
TomasLudvik: It seems, that Sonata is loading those keys and they were not present, so it caused standards to fail, see https://github.com/shopsys/shopsys/actions/runs/9080002098/job/24990013343

composer.json
TomasLudvik: It needs to be provided via URL (don't ask me why) and I did not find a way to load it from the local filesystem. It could be pushed to shopsys/shopsys and then linked, but maintaining fork and just changing those few line seemed to be much easier. 

Do you want me to push it to shopsys/shopsys and then link on it?

grossmannmartin: according to https://github.com/cweagans/composer-patches/tree/1.7.3#using-an-external-patch-file the local fila should be supported

TomasLudvik: Ok, I found a way, it is implemented now

packages/administration/composer.json
```diff
@@ -0,0 +1,51 @@
...
+            "dealerdirect/phpcodesniffer-composer-installer": true,
+            "cweagans/composer-patches": true
+        }
+    },
+    "extra": {
+        "enable-patching": true,
+        "patches": {
+            "composer-exit-on-patch-failure": true,
+            "twig/twig": {
+                "Hotfix for issue https://github.com/sonata-project/SonataAdminBundle/issues/8181": "packages/administration/src/Twig/sonata.patch"
```
grossmannmartin: the path in the context of the administration bundle will be different – `"src/Twig/sonata.patch"`

TomasLudvik: I have moved it to project-base

PR #3160, opened 2024-05-10
VACUUM FULL ANALYZE is now run once a week

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| It is a good practice to garbage-collect the Postgres database once in a while in production to avoid performance issues. We added a CRON module that performs `VACUUM FULL ANALYZE` on the database once a week so we don't have to think about it and run it manually anymore.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://rv-vacuum.odin.shopsys.cloud
  - https://cz.rv-vacuum.odin.shopsys.cloud
  - https://rv-vacuum.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/Doctrine/VacuumFullAnalyzeCronModule.php
+++ b/packages/framework/src/Component/Doctrine/VacuumFullAnalyzeCronModule.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Doctrine;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Exception;
+use Monolog\Logger;
+use Override;
+use Shopsys\FrameworkBundle\Component\Cron\CronControlFacade;
+use Shopsys\FrameworkBundle\Component\Maintenance\MaintenanceModeFacade;
+use Shopsys\Plugin\Cron\SimpleCronModuleInterface;
+use Symfony\Component\HttpClient\Exception\TimeoutException;
+
+class VacuumFullAnalyzeCronModule implements SimpleCronModuleInterface
+{
+    protected const int ONE_HOUR_IN_SECONDS = 3600;
+
+    protected const string VACUUM_CRON_INSTANCE_NAME = 'vacuum';
+
+    protected Logger $logger;
+
+    /**
+     * @param \Doctrine\ORM\EntityManagerInterface $entityManager
+     * @param \Shopsys\FrameworkBundle\Component\Cron\CronControlFacade $cronControlFacade
+     * @param \Shopsys\FrameworkBundle\Component\Maintenance\MaintenanceModeFacade $maintenanceModeFacade
+     */
+    public function __construct(
+        protected readonly EntityManagerInterface $entityManager,
+        protected readonly CronControlFacade $cronControlFacade,
+        protected readonly MaintenanceModeFacade $maintenanceModeFacade,
+    ) {
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function setLogger(Logger $logger): void
+    {
+        $this->logger = $logger;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function run(): void
+    {
+        if (!$this->cronControlFacade->lockCron()) {
+            $this->logger->error('Cron locking failed.');
+
+            return;
+        }
+
+        $this->logger->info('Cron is now locked.');
+
+        try {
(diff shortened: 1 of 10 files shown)


packages/framework/src/Component/Doctrine/VacuumFullAnalyzeCronModule.php
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    protected function unlockAllCronModules(): void
+    {
+        $this->lock->releaseLock(CronLockCommand::CRON_MUTEX_LOCK_NAME);
+        $this->logger->info('Cron lock was released.');
+    }
+
+    protected function waitUntilOtherCronModulesOff(): void
```
henzigo: Those functions are just duplicated in our base:

`waitUntilOtherCronModulesOff()` -> `CronWatchCommand`
`lockAllCronModules()` -> `CronLockCommand`
`maintenanceOn()` and `maintenanceOff()` -> `MaintenanceModeCommand`

packages/framework/src/Component/Doctrine/VacuumFullAnalyzeCronModule.php
```diff
@@ -0,0 +1,154 @@
...
+    public function run(): void
+    {
+        if (!$this->lockAllCronModules()) {
+            return;
+        }
+
+        try {
+            $this->waitUntilOtherCronModulesOff();
+            $this->maintenanceOn();
+            $this->makeVacuum();
```
henzigo, line 67: I don't like the idea of making VACUUM FULL from the application. Additionally, you should run VACUUM FULL as an edge case and not regularly. 

Postgres has AUTOVACUUM that should handle bloat in databases. If you have miswritten your application then you have to use VACUUM FULL. Running it periodically covers problems in your application.

Another view can be that ANALYSE rebuilds all statistics instead of updating them. So this means that this command can negatively impact the performance of some SQL commands.

I don't say we should entirely avoid using VACUUM FULL, but we should do that manually as an edge case.

There is an outdated but still relevant article about it: https://wiki.postgresql.org/wiki/VACUUM_FULL

vitek-rostislav: long time ago, we agreed we want this anyway :slightly_smiling_face:

project-base/app/config/cron.yaml
```diff
@@ -118,3 +118,9 @@ services:
     Shopsys\FrameworkBundle\Model\Watchdog\WatchdogCronModule:
         tags:
             - { name: shopsys.cron, hours: '*', minutes: '*/10', instanceName: watchdog, readableName: 'Sending watchdog emails' }
+
+    # Vacuum
+
+    Shopsys\FrameworkBundle\Component\Doctrine\VacuumFullAnalyzeCronModule:
```
grossmannmartin, line 124: I have a slightly bad feeling about the co-joined setting of time. Meaning the instance has 
`hours: '3', minutes: '15'`, and cron-vacuum is set to `15 3 * * 1`

1. I don't know how it behaves on these precise timings (both have the same time). 
2. It's really easy to forget about the other value when a change is needed.

packages/framework/src/Component/Doctrine/VacuumFullAnalyzeCronModule.php
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        $this->logger->info('Cron is now locked.');
+
+        try {
+            $this->logger->info('Waiting till all the cron instances are finished.');
+            $this->cronControlFacade->waitUntilCronInstancesAreFinished(
+                [self::VACUUM_CRON_INSTANCE_NAME],
+                self::ONE_HOUR_IN_SECONDS,
+            );
```
grossmannmartin, line 64: Are other instances automatically locked? Isn't possible that another cron will run during vacuum?

packages/framework/src/Component/Doctrine/VacuumFullAnalyzeCronModule.php
```diff
@@ -0,0 +1,86 @@
...
+    public function setLogger(Logger $logger): void
+    {
+        $this->logger = $logger;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function run(): void
```
grossmannmartin, line 49: I still don't think this is a good idea in general 🤷‍♂️

grossmannmartin (changes requested):

vitek-rostislav:
ok, we are closing this now as it is considered a bad practice

PR #3169, opened 2024-05-17
[framework] images are no longer processed by PHP to avoid quality decrease

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| When an image is processed by PHP, the quality might be decreased. This was unwanted and unnecessary as we already use CDN and imgProxy to get an image on the storefront.
|New feature| No 
|Fixes issues| ... 

----
  - https://rv-image-processing.odin.shopsys.cloud
  - https://cz.rv-image-processing.odin.shopsys.cloud


UPGRADE-15.0.md
```diff
@@ -832,6 +832,13 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
     -   `getAllOfferedProducts()` method was removed, use `getAllOfferedProductsPaginated()` instead
 -   see #project-base-diff to update your project
 
+#### images are no longer processed by PHP to avoid quality decrease ([#3169](https://github.com/shopsys/shopsys/pull/3169))
```
grossmannmartin: Shouldn't it be in the form of `do something`?  (like stop processing images by PHP to avoid decreasing quality)

packages/framework/src/Component/Image/Processing/ImageProcessor.php
```diff
@@ -71,30 +72,27 @@ public function createInterventionImage($filepath)
...
             throw new FileIsNotSupportedImageException($filepath);
         }
-        $newFilepath .= $extension;
-
-        $image = $this->createInterventionImage($filepath);
-        $data = $image->encode($extension)->getEncoded();
 
-        $this->filesystem->delete($filepath);
-        $this->filesystem->write($newFilepath, $data);
+        $newFilepath .= $extension;
```
grossmannmartin, line 86: This changes the application's behavior for GIF (and possibly other formats other than png/jpg). 
Previously, the GIF was transformed to the JPG format. 
Now in administration, it stays as a GIF (which means the image can be animated in administration) and is processed by imgproxy, resulting in a still image. 


https://github.com/shopsys/shopsys/assets/1177414/3eff8158-cf21-4229-b695-8c0bc83166f4

packages/framework/src/Component/Image/Processing/ImageThumbnailFactory.php
```diff
@@ -20,7 +22,7 @@ public function __construct(protected readonly ImageProcessor $imageProcessor)
      * @param string $filepath
      * @return \Intervention\Image\Image
      */
-    public function getImageThumbnail($filepath)
```
grossmannmartin, line 23: isn't possible to use imgproxy even in the administration and avoid creating a thumbnail?

grossmannmartin (changes requested):
can you please also check failed standards and tests?

UPGRADE-15.0.md
vitek-rostislav: I always forget about this :smile: yes, you are right, fixed, thanks! 10e41fa816

packages/framework/src/Component/Image/Processing/ImageThumbnailFactory.php
vitek-rostislav: yeah, I looked into it but it felt out of the scope of this PR, the thumbnails (and domain icons) are not processed via imgProxy as they are located in `content/admin/images/`. It would require the nginx configuration change and I was not sure whether it made sense now as we plan to change the administration completely :thinking: Let's discuss it further, please

packages/framework/src/Component/Image/Processing/ImageProcessor.php
vitek-rostislav: after discussions resolved - images are now always processed by imgProxy in admin - f1c5f05be8

packages/framework/src/Component/Image/Processing/ImageThumbnailFactory.php
vitek-rostislav: discussed and yes, it is possible! :slightly_smiling_face:

packages/framework/src/Component/Domain/DomainIconProcessor.php
```diff
@@ -33,21 +29,17 @@ public function __construct(
...
-            static::DOMAIN_ICON_CROP,
-        );
-        $resizedImage->encode(ImageProcessor::EXTENSION_PNG);
-
+    public function convertToDomainIconFormatAndSave(
+        int $domainId,
+        string $filepath,
+        string $domainImagesDirectory,
+    ): void {
         $targetFilePath = $domainImagesDirectory . '/' . $domainId . '.' . ImageProcessor::EXTENSION_PNG;
```
grossmannmartin, line 38: Does this mean when I upload JPG file, it will only be renamed to PNG? 
Is it ok?

edit: oh, now I see only PNG format is allowed in form type...  In that case, should we assume this will always be true (only PNG is allowed)? 

Nitpick: maybe the method should be named just `saveIcon`? it's not converting anymore, or?

packages/framework/src/Form/Admin/Domain/DomainFormType.php
```diff
@@ -26,14 +26,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'required' => false,
                 'file_constraints' => [
                     new Constraints\Image([
-                        'mimeTypes' => ['image/png', 'image/jpg', 'image/jpeg', 'image/gif'],
-                        'mimeTypesMessage' => 'Image can be only in JPG, GIF or PNG format',
+                        'mimeTypes' => ['image/png'],
+                        'mimeTypesMessage' => 'Image can be only in PNG format',
```
grossmannmartin, line 30: Are we ok with the behavior change?

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Domain/DomainFormType.php
vitek-rostislav: I discussed this with @pk16011990 and yes, we are ok with this change :slightly_smiling_face:

packages/framework/src/Component/Domain/DomainIconProcessor.php
vitek-rostislav: good point with the naming. Also, I have added check for the mime type into the method, just to be sure it works as expected regardless the form type constraints. 86ebaab209

grossmannmartin (approved):

pk16011990 (approved):

PR #3182, opened 2024-05-27
[project-base] fixed blogarticle demo data

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Demo data for blog article for search testing did not allow editation in admin. This is fixed now.
|New feature| No 
|Fixes issues| ... 

----
  - https://mg-fix-blogarticle.odin.shopsys.cloud
  - https://cz.mg-fix-blogarticle.odin.shopsys.cloud


project-base/app/translations/dataFixtures.cs.po
```diff
@@ -307,6 +307,9 @@ msgstr "<b>Zvolili jste platbu kreditní kartou. Prosím proveďte ji do dvou pr
...
 
+msgid "<div class=\"gjs-text-ckeditor\">Article text for search testing, the search phrase is &#34;Dina&#34;.</div>"
```
vitek-rostislav, line 310: I was surprised I saw this addition and no deletion, I was expecting the original string to be removed from here. I found out the same string is used in `ArticleDataFixture` where the "gjs-text-ckeditor" div is excluded from the message id. 

So I see in `BlogArticleDataFixture`, we include the divs in the message IDs, while in `ArticleDataFixture`, we exclude the divs from the translation message IDs. 

Does it bother me? I do not think so :smile:

vitek-rostislav (approved):

pk16011990 (approved):

PR #3187, opened 2024-05-28
[project-base] product data fixture refactoring

AndrejBlaho:
| Q             | A
| ------------- | ---
|Description, reason for the PR| There was some unnecessary repeating in product data fixtures. It is now simplified in places where it made sense for project base purposes. Also ProductDemoDataFactory was added as a foundation for creating demo data templates for certain groups of products that has a lot of common data which is not very applicable for project base demo data but could be in real projects.
|New feature| No 
|Fixes issues| ... 

----
  - https://ab-refactor-product-data-fixtures.odin.shopsys.cloud
  - https://cz.ab-refactor-product-data-fixtures.odin.shopsys.cloud

### DIFF
--- a/project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
+++ b/project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
@@ -33,7 +33,7 @@ class ProductDemoDataSetter
 
         foreach ($this->domain->getAllIds() as $domainId) {
             if ($vatReference !== null) {
-                $productVatsIndexedByDomainId[$domainId] = $this->persistentReferenceFacade->getReferenceForDomain($vatReference, Domain::FIRST_DOMAIN_ID, Vat::class);
+                $productVatsIndexedByDomainId[$domainId] = $this->persistentReferenceFacade->getReferenceForDomain($vatReference, $domainId, Vat::class);
             }
         }
         $productData->vatsIndexedByDomainId = $productVatsIndexedByDomainId;
--- a/project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
+++ b/project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\DataFixtures\Demo\DataSetter;
+
+use App\Model\Product\ProductData;
+use App\Model\Product\Unit\Unit;
+use DateTime;
+use Shopsys\FrameworkBundle\Component\DataFixture\PersistentReferenceFacade;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Model\Pricing\Vat\Vat;
+
+class ProductDemoDataSetter
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @param \Shopsys\FrameworkBundle\Component\DataFixture\PersistentReferenceFacade $persistentReferenceFacade
+     */
+    public function __construct(
+        private readonly Domain $domain,
+        private readonly PersistentReferenceFacade $persistentReferenceFacade,
+    ) {
+    }
+
+    /**
+     * @param \App\Model\Product\ProductData $productData
+     * @param string|null $vatReference
+     */
+    public function setVat(ProductData $productData, ?string $vatReference): void
+    {
+        $productVatsIndexedByDomainId = [];
+
+        foreach ($this->domain->getAllIds() as $domainId) {
+            if ($vatReference !== null) {
+                $productVatsIndexedByDomainId[$domainId] = $this->persistentReferenceFacade->getReferenceForDomain($vatReference, Domain::FIRST_DOMAIN_ID, Vat::class);
+            }
+        }
+        $productData->vatsIndexedByDomainId = $productVatsIndexedByDomainId;
+    }
+
+    /**
+     * @param \App\Model\Product\ProductData $productData
+     * @param string $unitReference
+     */
+    public function setUnit(ProductData $productData, string $unitReference): void
+    {
+        $productData->unit = $this->persistentReferenceFacade->getReference($unitReference, Unit::class);
+    }
+
+    /**
+     * @param \App\Model\Product\ProductData $productData
+     * @param string|null $date
+     */
+    public function setSellingTo(ProductData $productData, ?string $date): void
+    {
+        $productData->sellingTo = $date === null ? null : new DateTime($date);
+    }
+}
--- a/project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
+++ b/project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
@@ -4,22 +4,47 @@ declare(strict_types=1);
 
 namespace App\DataFixtures\Demo\DataSetter;
 
+use App\DataFixtures\Demo\CurrencyDataFixture;
+use App\DataFixtures\Demo\VatDataFixture;
+use App\Model\Category\Category;
+use App\Model\Product\Brand\Brand;
+use App\Model\Product\Flag\Flag;
+use App\Model\Product\Parameter\ParameterValueDataFactory;
 use App\Model\Product\ProductData;
 use App\Model\Product\Unit\Unit;
 use DateTime;
 use Shopsys\FrameworkBundle\Component\DataFixture\PersistentReferenceFacade;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Shopsys\FrameworkBundle\Model\Pricing\Currency\Currency;
+use Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroupFacade;
+use Shopsys\FrameworkBundle\Model\Pricing\PriceConverter;
 use Shopsys\FrameworkBundle\Model\Pricing\Vat\Vat;
+use Shopsys\FrameworkBundle\Model\Product\Parameter\ProductParameterValueDataFactory;
+use Shopsys\FrameworkBundle\Model\Stock\ProductStockDataFactory;
+use Shopsys\FrameworkBundle\Model\Stock\StockRepository;
 
 class ProductDemoDataSetter
 {
     /**
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
      * @param \Shopsys\FrameworkBundle\Component\DataFixture\PersistentReferenceFacade $persistentReferenceFacade
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroupFacade $pricingGroupFacade
+     * @param \Shopsys\FrameworkBundle\Model\Product\Parameter\ProductParameterValueDataFactory $productParameterValueDataFactory
+     * @param \App\Model\Product\Parameter\ParameterValueDataFactory $parameterValueDataFactory
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\PriceConverter $priceConverter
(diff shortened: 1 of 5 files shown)


project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
```diff
@@ -0,0 +1,80 @@
...
+        $productData->sellingDenied = false;
+
+        return $productData;
+    }
+
+    /**
+     * @param \App\Model\Product\ProductData $productData
+     * @param string|null $vatReference
+     */
+    private function setVat(ProductData $productData, ?string $vatReference): void
```
vitek-rostislav: the same method remained in `ProductDataFixture` itself. I can imagine it will be in this class only and will be public. I would then try to find a better name for this class. Or, maybe even better, there could be yet another class for that purpose - in this factory, there will be `createDefaultData()` method only, and then there can be something like `ProductDemoDataSetter` that would contain all these setter methods - `setVat`, `setBrand`, `setUnit`, etc. and the class would be used here, as well as in `ProductDataFixture` itself.

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $productVatsIndexedByDomainId = [];
+
+        foreach ($this->domain->getAllIds() as $domainId) {
+            if ($vatReference !== null) {
+                $productVatsIndexedByDomainId[$domainId] = $this->persistentReferenceFacade->getReferenceForDomain($vatReference, Domain::FIRST_DOMAIN_ID, Vat::class);
```
vitek-rostislav: After the refactoring is done, I would like to fix this line - there is a mistake - we assign the first domain Vat for all the domains. I am not sure whether this will break some tests or not, we will see :smile: :see_no_evil:

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
```diff
@@ -0,0 +1,80 @@
...
+use App\DataFixtures\Demo\VatDataFixture;
+use App\Model\Product\ProductData;
+use App\Model\Product\ProductDataFactory;
+use App\Model\Product\Unit\Unit;
+use DateTime;
+use Shopsys\FrameworkBundle\Component\DataFixture\PersistentReferenceFacade;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Model\Pricing\Vat\Vat;
+
+class ProductDemoDataFactory
```
vitek-rostislav, line 13: I think I would rather see this class in the framework package but let's discuss this in the team.

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
```diff
@@ -0,0 +1,80 @@
...
+     * @return \App\Model\Product\ProductData
+     */
+    public function createDefaultData(): ProductData
+    {
+        $productData = $this->productDataFactory->create();
+
+        $this->setVat($productData, VatDataFixture::VAT_HIGH);
+        $this->setUnit($productData, UnitDataFixture::UNIT_PIECES);
+        $this->setSellingTo($productData, null);
+
```
vitek-rostislav: I feel like there could be more data set in this method, I would try to spot the fields that are set multiple times with the same value, and use that value as the default one. Then, only products with different values would set their values appropriately. I see that means that some values will be set two times (in `createDefaultData` and then again with the unique value) but I believe it would not slow down the data fixtures loading (or at least not significantly).

To give an example - `sellingFrom` is set for the same date at least ten times so we could get rid of all those lines if the value is default:

Screenshot from 2024-05-31 20-21-50

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @return \App\Model\Product\ProductData
+     */
+    public function createDefaultData(): ProductData
+    {
+        $productData = $this->productDataFactory->create();
+
+        $this->setVat($productData, VatDataFixture::VAT_HIGH);
+        $this->setUnit($productData, UnitDataFixture::UNIT_PIECES);
+        $this->setSellingTo($productData, null);
```
vitek-rostislav: is it necessary to call this method here? the null values should be initialized already after creating the data object by `ProductDataFactory`. So I believe all the `set*` methods can have non-nullable their second argument - if you want to set a particular value to the field, call the setter, if you want to keep it null, do not do anything :shrug: In this manner, we can get rid of multiple lines in `ProductDataFixture` as well - e.g. `$this->setBrand($productData, null);` is called 75 times for vain, `$this->setFlags($productData, []);` is called 99 times in vain.

vitek-rostislav (changes requested):
Hi, the product data fixture class is still too large. I have added a few ideas but I am afraid we will still have the problem even after the changes are applied. The [concept](https://github.com/shopsys/shopsys/commit/25ef22d650a2546b89afa77bb6544fd15c088bb6#diff-5aca7968926a0973c2c6ba228a241f0249b5c2a896e57754483a1ebbc2481bed) proposed by @grossmannmartin was discussed and rejected?

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
AndrejBlaho: tried to rewrite it as you recommended

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
AndrejBlaho: fixed, lets see what will be reported from tests 😅

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
AndrejBlaho: not implementing this as we discussed

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
AndrejBlaho: Removed more lines where unnecessary empty value / array setting was happening, but i don't think there is much more we can get out of this approach, remaining data is just too varying in general.

Regarding sellingFrom i did not implemented that yet. There is at most 29 same values and trade-off between trimming down 29 lines and overriding value at 120+ other occurrences does not seem worth it to me, but i can do it if you think otherwise.

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
AndrejBlaho: you are right, removed all unnecessary ballast i saw

vitek-rostislav: I still see `$this->setSellingTo($productData, null);` in `createDefaultData` :frowning_face:

vitek-rostislav: moreover, there is no point in having the second nullable parameter in `setSellingTo` method - the data is null after the `ProductData` object is created in `ProductDataFactory`.

project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
```diff
@@ -0,0 +1,59 @@
...
+        private readonly Domain $domain,
+        private readonly PersistentReferenceFacade $persistentReferenceFacade,
+    ) {
+    }
+
+    /**
+     * @param \App\Model\Product\ProductData $productData
+     * @param string|null $vatReference
+     */
+    public function setVat(ProductData $productData, ?string $vatReference): void
```
vitek-rostislav: please, make `$vatReference` parameter non-nullable

project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
```diff
@@ -0,0 +1,59 @@
...
+    public function setUnit(ProductData $productData, string $unitReference): void
+    {
+        $productData->unit = $this->persistentReferenceFacade->getReference($unitReference, Unit::class);
+    }
+
+    /**
+     * @param \App\Model\Product\ProductData $productData
+     * @param string|null $date
+     */
+    public function setSellingTo(ProductData $productData, ?string $date): void
```
vitek-rostislav: please, make `$vatReference` parameter non-nullable

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
vitek-rostislav: please, move all the setters from `ProductDataFixture` :pray:

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
vitek-rostislav: you will need to rebase the branch first :slightly_smiling_face:

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
vitek-rostislav: ok, just move the `setSellingFrom` to `ProductDemoDataSetter` (with non-nullable `$date` parameter), please

AndrejBlaho: done, changed setSellingTo to non-nullable as well, but im not sure if we really want it
there might be scenarios where setting these values back to null after initial setting might be relevant, especially if one of our intentions is to create basis for creating product templates

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
AndrejBlaho: done

project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
AndrejBlaho: done

project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
AndrejBlaho: done

project-base/app/src/DataFixtures/Demo/DemoDataFactory/ProductDemoDataFactory.php
vitek-rostislav: yeah, that is a fair point but you can always set the value back to null directly on the data object, there is no need for a special method for that, you just use `$productData->sellingTo = null` :slightly_smiling_face:

vitek-rostislav (changes requested):
Hi, it looks nice now :slightly_smiling_face: 

I have two last points:
- `ProductDemoDataSetter::setBrand()` - please make the `$brandReference` parameter non-nullable
- you had the idea about removing the unnecessary empty lines from `ProductDataFixture` itself (and moving the domain loops), could you do that please? I believe it will also help the readability of the file, something like this:
image

vitek-rostislav (approved):

pk16011990 (approved):

PR #3194, opened 2024-06-03
[shopsys] split upgrade notes

AndrejBlaho:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Upgrade notes were a constant source of trivial merge conflicts. Also often resulting in non-functioning preview builds due to github policy of stopping builds of conflicting branches. Now upgrade notes are created in separate file for each pull request and merged when needed. Phing targets for creating and merging upgrade notes were added.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://ab-upgrade-notes-split.odin.shopsys.cloud
  - https://cz.ab-upgrade-notes-split.odin.shopsys.cloud


build.xml
```diff
@@ -527,6 +528,27 @@
...
+    </target>
+
+    <target name="upgrade-storefront" description="Creates  empty file named for storefront upgrade notes.">
+        <exec executable="php">
+            <arg value="-r"/>
+            <arg value="copy('upgrade-notes/_template.md', 'upgrade-notes/storefront_' . date('Ymd_His') . '.md');"/>
+        </exec>
+    </target>
+
+    <target name="upgrade-merge" description="Merges all upgrade notes into UPGRADE-{X.X}.md according to structure specified in upgrade-notes/_layout.md. Default value set in version property can be overridden by '-Dversion={X.X}'">
```
grossmannmartin: What do you think about asking for the version interactively? 

```
<target name="upgrade-merge">
    <input message="Enter version to merge upgrade notes for:" propertyName="version"/>
    <exec executable="sh">
             <arg value="-c"/>
             <arg value="upgrade-notes/upgrade_notes.sh ${version}"/>
         </exec>
```

that way we don't need to change the version in this file after release

upgrade-notes/backend_20240603_102758.md
```diff
@@ -0,0 +1,5 @@
+#### split upgrade notes ([#3194](https://github.com/shopsys/shopsys/pull/3194))
```
grossmannmartin: I don't think the upgrade notes for this PR are necessary as it affects only monorepo and nothing changes for the developers

upgrade-notes/backend_20240529_084428.md
```diff
@@ -0,0 +1,881 @@
...
+-   if your project was originally not developed from the Commerce Cloud version, or it was developed on a version lower than `v13.0.0`, these feature movements should not affect you during the upgrade
+-   otherwise, you might need to adjust your project to the changes:
+    -   if you had no custom changes in the moved features, you should be fine, you can safely remove the features from your project and use the ones from the packages (project-base diff in each section will help you with that)
+    -   if you had custom changes in the moved features, you will need to adjust your project to the changes
+        -   you should remove everything that was not modified in your project and keep just the custom changes using the recommended ways of the [framework extensibility](https://docs.shopsys.com/en/latest/extensibility/)
+    -   one way or another, you should pay a special attention to the database migrations that were added with the feature movement. If they suit your needs, you should keep them and remove the original migrations from your project, otherwise, you should skip the newly added migrations.
+
+<!-- Insert upgrade instructions in the following format:
```
grossmannmartin: I think this comment is no longer necessary, as we have the _template.md file

build.xml
```diff
@@ -527,6 +528,27 @@
...
+            <arg value="copy('upgrade-notes/_template.md', 'upgrade-notes/backend_' . date('Ymd_His') . '.md');"/>
+        </exec>
+    </target>
+
+    <target name="upgrade-storefront" description="Creates  empty file named for storefront upgrade notes.">
+        <exec executable="php">
+            <arg value="-r"/>
+            <arg value="copy('upgrade-notes/_template.md', 'upgrade-notes/storefront_' . date('Ymd_His') . '.md');"/>
+        </exec>
+    </target>
```
grossmannmartin: What about a single phing target that asks about the type? 

Something like 

```
<target name="upgrade-generate">
        <input message="Which one?"
               propertyName="upgradeType"
               validArgs="backend,storefront,both"
        />

        <if>
            <or>
                <equals arg1="${upgradeType}" arg2="backend"/>
                <equals arg1="${upgradeType}" arg2="both"/>
            </or>
            <then>
                ... generate backend ...
            </then>
        </if>

        <if>
            <or>
                <equals arg1="${upgradeType}" arg2="storefront"/>
                <equals arg1="${upgradeType}" arg2="both"/>
            </or>
            <then>
                ... generate storefront ...
            </then>
        </if>
    </target>
    ```

File: upgrade-notes/upgrade_notes.sh [hunk 5]
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -x
```
grossmannmartin: I would remove the `x` and add the `e` modifier
so script will stop on error and executed commands are not printed

upgrade-notes/_layout.md
```diff
@@ -0,0 +1,50 @@
+# UPGRADING FROM 14.x to 15.0
```
grossmannmartin, line 1: I think the current UPGRADE-15.0.md should be deleted

build.xml
grossmannmartin: There should be a note in the releaser to merge the files during release. 

I suggest add phing target call to  `utils/releaser/src/ReleaseWorker/ReleaseCandidate/UpdateUpgradeReleaseWorker.php`

something like `$this->processRunner->run('php phing upgrade-merge');`

grossmannmartin (changes requested):

build.xml
AndrejBlaho: That is neat. Changed it as such.

upgrade-notes/backend_20240603_102758.md
AndrejBlaho: Got it, now it is removed.

upgrade-notes/backend_20240529_084428.md
AndrejBlaho: I see, removed it as well.

build.xml
AndrejBlaho: That is very nice, rewrote it as such.

upgrade-notes/upgrade_notes.sh
AndrejBlaho: Got it.

upgrade-notes/_layout.md
AndrejBlaho: Deleted.

build.xml
AndrejBlaho: Added. Thank you for very inspiring CR.

grossmannmartin (approved):

upgrade-notes/upgrade_notes.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+if ($argc < 3) {
+    echo "Error: version or upgrade-notes path is not set. Pass the version and the path as arguments.\n";
+    exit(1);
+}
+
+$version = $argv[1];
+$upgradeNotesPath = $argv[2];
```
grossmannmartin: I would use `__DIR__` directly instead of specifying the dir in build.xml. 
I don't suppose we want to merge from another location

upgrade-notes/upgrade_notes.php
```diff
@@ -0,0 +1,70 @@
...
+
+// If upgrade notes file for this version exists extend it, otherwise use template
+if (file_exists($outputFile)) {
+    $upgradeNotesContent = file_get_contents($outputFile);
+} else {
+    $upgradeNotesContent = file_get_contents($layoutFile);
+}
+
+$backendNotes = concatenateFiles("backend_", $upgradeNotesPath);
+$upgradeNotesContent = str_replace("<backendNotes>", $backendNotes . "<backendNotes>", $upgradeNotesContent);
```
grossmannmartin: I suggest use comment (`<!-- backendNotes -->`) as a placeholder, so the placeholder will not be rendered in the markdown file. (same for the storefront notes). 

Also can you extract the string into the constant, so the string is not duplicated in the code?

grossmannmartin (changes requested):

upgrade-notes/upgrade_notes.php
AndrejBlaho: yes, that sounds about right

upgrade-notes/upgrade_notes.php
AndrejBlaho: That is definitely better option. Changed it.

grossmannmartin (approved):

build.xml
```diff
@@ -456,6 +457,42 @@
...
+        <input message="Which one?" propertyName="upgradeType" validArgs="backend,storefront,both"/>
+        <if>
+            <or>
+                <equals arg1="${upgradeType}" arg2="backend"/>
+                <equals arg1="${upgradeType}" arg2="both"/>
+            </or>
+            <then>
+                <exec executable="php">
+                    <arg value="-r"/>
+                    <arg value="copy('upgrade-notes/_template.md', 'upgrade-notes/backend_' . date('Ymd_His') . '.md');"/>
```
grossmannmartin: is it possible to use `path.upgrade-notes` property here and on the line 482?

upgrade-notes/upgrade_notes.php
```diff
@@ -0,0 +1,73 @@
...
+function concatenateFiles($prefix, $upgradeNotesPath) {
+    $content = "";
+    $files = glob("{$upgradeNotesPath}/{$prefix}*");
+    $lastIndex = count($files) - 1;
+
+    foreach ($files as $index => $file) {
+        if (is_file($file)) {
+            $content .= file_get_contents($file);
+            // Add an empty line after each file except the last one
+            if ($index != $lastIndex) {
```
pk16011990, line 25: When the last file is not a file (is_file condition above), new two lines are added to the last item. I'm not sure how `glob()` sorts items and maybe it is not a problem 🤷‍♂️

pk16011990 (approved):

upgrade-notes/upgrade_notes.php
AndrejBlaho: There are no aded lines if $file is not a file, no?

AndrejBlaho: Got it now. After our discussion i'm letting it go as it is.

PR #3211, opened 2024-06-10
[project-base][SSP-2246] adverts query optimisation

JanMolcik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| To improve performance, we should move towards filtering adverts on the BE rather than fetching all adverts and filtering them on the FE. As API is already prepared for this, FE was adjusted to use filter arguments for adverts query.
|New feature| No 
|Fixes issues| ... 

----
  - https://jm-ssp-2246-adverts-query-optimization.odin.shopsys.cloud
  - https://cz.jm-ssp-2246-adverts-query-optimization.odin.shopsys.cloud


project-base/storefront/graphql/requests/adverts/queries/AdvertsQuery.graphql
```diff
@@ -1,5 +1,5 @@
-query AdvertsQuery {
-    adverts {
+query AdvertsQuery($categoryUuid: Uuid, $positionName: String) @friendlyUrl {
```
sebaholesz: Not sure why `@friendlyUrl` was used here. I am not sure if this query can be considered a "friendly URL" one. What do you base it on?

project-base/storefront/components/Blocks/Adverts/Adverts.tsx
```diff
@@ -33,13 +33,17 @@ export const Adverts: FC<AdvertsProps> = ({
...
 }) => {
-    const [{ data: advertsData }] = useAdvertsQuery();
-
-    const filteredAdverts = advertsData?.adverts.filter((advert) => advert.positionName === positionName);
+    const [{ data: advertsData }] = useAdvertsQuery({
+        variables: {
+            categoryUuid: currentCategory?.uuid || null,
+            positionName
+        },
+    });
```
sebaholesz: Since you have added this, I think that the `shouldBeShown` function and the rest of the display logic can be reduced/removed. WDYT?

sebaholesz (changes requested):

project-base/storefront/graphql/requests/adverts/queries/AdvertsQuery.graphql
JanMolcik: Indeed, unnecessary `@friendlyUrl`. Removed.

project-base/storefront/components/Blocks/Adverts/Adverts.tsx
JanMolcik: Good point, removed unnecessary`shouldBeShown` filter and used advert uuid as react key in the `.map()` array rendering.

project-base/storefront/components/Blocks/Adverts/Adverts.tsx
```diff
@@ -36,58 +35,54 @@ export const Adverts: FC<AdvertsProps> = ({
...
+                            {advert.link ? (
+                                <ExtendedNextLink href={advert.link} target="_blank">
+                                    {ImageComponent}
+                                </ExtendedNextLink>
+                            ) : (
+                                ImageComponent
+                            )}
+                        </Fragment>
+                    );
+                }
```
sebaholesz: Just a minor suggestion. This could be extracted into a separate component (something like AdvertImage.tsx). I will leave this for you to decide. If you decide that it's an overkill, you can just resolve this thread and send it to tests.

sebaholesz (approved):

project-base/storefront/components/Blocks/Adverts/Adverts.tsx
```diff
@@ -33,57 +32,57 @@ export const Adverts: FC<AdvertsProps> = ({
...
 }) => {
-    const [{ data: advertsData }] = useAdvertsQuery();
-
-    const filteredAdverts = advertsData?.adverts.filter((advert) => advert.positionName === positionName);
+    const [{ data: advertsData }] = useAdvertsQuery({
+        variables: {
+            categoryUuid: currentCategory?.uuid || null,
+            positionName,
+        },
+    });
```
sebaholesz: I have just realized that after changing these params, SSR fetching will not work (can be seen by the fact that this query is fetched on the client). To fix this, you have to make sure that `{ query: AdvertsQueryDocument }` in `initServerSideProps.ts` is called with the same params as here.

sebaholesz (changes requested):

sebaholesz (approved):

packages/frontend-api/src/Model/Resolver/Advert/AdvertsQuery.php
```diff
@@ -44,7 +48,12 @@ public function advertsQuery(?string $positionName = null, ?string $categoryUuid
         $category = null;
 
         if ($categoryUuid !== null) {
-            $category = $this->categoryFacade->getByUuid($categoryUuid);
+            try {
+                $category = $this->categoryFacade->getByUuid($categoryUuid);
+            } catch (CategoryNotFoundException $exception) {
+                $seoMixCategory = $this->readyCategorySeoMixFacade->getByUuid($categoryUuid);
+                $category = $this->categoryFacade->getById($seoMixCategory->getCategory()->getId());
```
grossmannmartin: Isn't this sufficient?

```suggestion
                $category = $seoMixCategory->getCategory();
```

project-base/storefront/pages/wishlist.tsx
```diff
@@ -35,7 +36,21 @@ const WishlistPage: NextPage<ServerSidePropsType> = () => {
...
+            initServerSideProps({
+                context,
+                redisClient,
+                domainConfig,
+                t,
+                prefetchedQueries: [
+                    {
+                        query: AdvertsQueryDocument,
+                        variables: {
+                            positionName: 'productListSecondRow',
```
grossmannmartin: is the advert even rendered in the wishlist?

I have tried to create the advert for this position and in the wishlist is not rendered on the 15.0 nor on this branch


note: please check also other places like products by branch, products with flag

project-base/storefront/pages/wishlist.tsx
```diff
@@ -35,7 +36,21 @@ const WishlistPage: NextPage<ServerSidePropsType> = () => {
...
     ({ redisClient, domainConfig, t }) =>
         async (context) =>
-            initServerSideProps({ context, redisClient, domainConfig, t }),
+            initServerSideProps({
+                context,
+                redisClient,
+                domainConfig,
+                t,
+                prefetchedQueries: [
+                    {
```
grossmannmartin: When I refresh the wishlist page, I get the following error

```
Unhandled Runtime Error
Error: Hydration failed because the initial UI does not match what was rendered on the server.

Warning: Expected server HTML to contain a matching <div> in <div>.

See more info here: https://nextjs.org/docs/messages/react-hydration-error
```

project-base/storefront/pages/search.tsx
```diff
@@ -48,12 +54,20 @@ export const getServerSideProps = getServerSidePropsWrapper(({ redisClient, doma
...
+    return initServerSideProps<TypeSearchQueryVariables | TypeSearchProductsQueryVariables | TypeAdvertsQueryVariables>(
+        {
+            context,
+            redisClient,
+            domainConfig,
+            t,
+            prefetchedQueries: [
+                {
+                    query: AdvertsQueryDocument,
+                    variables: { positionName: 'productListSecondRow', categoryUuid: null },
```
grossmannmartin: is the advert even rendered on the search page?

I have tried to create the advert for this position and in the wishlist is not rendered on the 15.0 nor on this branch

project-base/storefront/pages/categories/[categorySlug].tsx
```diff
@@ -129,6 +134,20 @@ export const getServerSideProps = getServerSidePropsWrapper(
...
+                        query: AdvertsQueryDocument,
+                        variables: { positionName: 'productList', categoryUuid },
+                    },
+                    {
+                        query: AdvertsQueryDocument,
+                        variables: { positionName: 'productListMiddle', categoryUuid },
+                    },
+                    {
+                        query: AdvertsQueryDocument,
+                        variables: { positionName: 'productListSecondRow', categoryUuid },
```
Comment (grossmannmartin) [inline project-base/storefront/pages/categories/[categorySlug].tsx]: The behavior of the advert in category has changed. 

Previously (= on the 15.0 branch), when no category was set, the advert was not rendered anywhere. Now its rendered in the each category

project-base/storefront/pages/order/contact-information.tsx
```diff
@@ -32,7 +33,16 @@ export const getServerSideProps = getServerSidePropsWrapper(
                 redisClient,
                 domainConfig,
                 t,
-                prefetchedQueries: [{ query: CountriesQueryDocument }],
+                prefetchedQueries: [
+                    { query: CountriesQueryDocument },
+                    {
+                        query: AdvertsQueryDocument,
+                        variables: {
+                            positionName: 'cartPreview',
```
grossmannmartin: why is necessary to prefetch adverts with position cartPreview on the contact-information page?

grossmannmartin (changes requested):

project-base/storefront/pages/order/contact-information.tsx
JanMolcik: The need to prefetch this comes from `OrderSummary` component, which uses `<Adverts ... positionName="cartPreview" />`. This cascades down on these children components `ContactInformationPage -> ContactInformationWrapper -> OrderContentWrapper -> OrderSummary -> Adverts`. If I remove it I can see it is being fetched on the FE (in the networks tab).

project-base/storefront/pages/search.tsx
JanMolcik: Indeed it is not rendered. I went only by references and didn't realize that we display the adverts only if we receive a category in the props of `ProductsList` component.

project-base/storefront/pages/wishlist.tsx
JanMolcik: Indeed, it is not rendered. I went only by references and didn't realize that we display the adverts only if we receive a category in the props of the ProductsList component.

JanMolcik: The same applies also to products by branch and products by flag.

project-base/storefront/pages/wishlist.tsx
JanMolcik: I wasn't able to reproduce.

project-base/storefront/pages/categories/[categorySlug].tsx
Comment (JanMolcik) [reply project-base/storefront/pages/categories/[categorySlug].tsx]: Yes, I have found that the behaviour changed because of one specific check was previously done on the FE and aparently the BE does not check for that:

(content of the filter function `shouldBeShown` in `Adverts.tsx`)
```js
...
const isAdvertWithCategoryRestriction = ['productList', 'productListMiddle', 'productListSecondRow'].includes(
    advert.positionName,
);

if (isAdvertWithCategoryRestriction && !advert.categories.length) {
    return false;
}
...
```
So if the categories array for these 3 positionNames `productList`, `productListMiddle` and `productListSecondRow` is empty, the advert is not shown.

packages/frontend-api/src/Model/Resolver/Advert/AdvertsQuery.php
AndrejBlaho: fixed 😅

packages/framework/src/Model/Advert/AdvertRepository.php
```diff
@@ -17,8 +18,10 @@ class AdvertRepository
     /**
      * @param \Doctrine\ORM\EntityManagerInterface $em
      */
-    public function __construct(protected readonly EntityManagerInterface $em)
-    {
+    public function __construct(
+        protected readonly EntityManagerInterface $em,
+        protected readonly QueryBuilderExtender $queryBuilderExtender,
```
grossmannmartin: is the extender necessary?

project-base/storefront/pages/wishlist.tsx
grossmannmartin: OK, thanks for checking

grossmannmartin (approved):

packages/framework/src/Model/Advert/AdvertRepository.php
AndrejBlaho: now it should be removed

packages/frontend-api/src/Model/Resolver/Advert/AdvertsQuery.php
```diff
@@ -44,7 +48,12 @@ public function advertsQuery(?string $positionName = null, ?string $categoryUuid
         $category = null;
 
         if ($categoryUuid !== null) {
-            $category = $this->categoryFacade->getByUuid($categoryUuid);
+            try {
+                $category = $this->categoryFacade->getByUuid($categoryUuid);
+            } catch (CategoryNotFoundException $exception) {
+                $seoMixCategory = $this->readyCategorySeoMixFacade->getByUuid($categoryUuid);
```
pk16011990: Two points:
1. $categoryUuid should be renamed to `CategoryOrSeoCategoryUuid` (and in API?)
2. `ReadyCategorySeoMixFacade` is in project-base, It means packages/frontend-api depends on project-base

packages/frontend-api/src/Model/Resolver/Advert/AdvertsQuery.php
```diff
@@ -43,10 +43,6 @@ public function advertsQuery(?string $positionName = null, ?string $categoryUuid
 
         $category = null;
 
-        if ($categoryUuid !== null) {
```
grossmannmartin: this can stay... otherwise we drop some functionality from frontend-api package.

grossmannmartin (approved):

packages/frontend-api/src/Model/Resolver/Advert/AdvertsQuery.php
```diff
@@ -25,19 +25,19 @@ public function __construct(
...
     {
-        if ($positionName === AdvertPositionRegistry::POSITION_PRODUCT_LIST && $categoryUuid === null) {
+        if (array_key_exists(AdvertPositionRegistry::POSITION_PRODUCT_LIST, $positionNames) && $categoryUuid === null) {
             throw new AdvertPositionWithoutCategoryUserError('Cannot retrieve advert on product list page without setting category.');
         }
 
         $domainId = $this->domain->getId();
 
-        if ($positionName === null) {
+        if ($positionNames === null) {
```
AndrejBlaho: if ($positionNames === null || $positionNames === []) {

project-base/storefront/pages/wishlist.tsx
```diff
@@ -35,7 +35,12 @@ const WishlistPage: NextPage<ServerSidePropsType> = () => {
 export const getServerSideProps = getServerSidePropsWrapper(
     ({ redisClient, domainConfig, t }) =>
         async (context) =>
-            initServerSideProps({ context, redisClient, domainConfig, t }),
+            initServerSideProps({
+                context,
+                redisClient,
+                domainConfig,
+                t,
+            }),
```
sebaholesz: This seems like an unnecessary change. You can resolve this thread by yourself, but I just wanted to write it to let you know that such changes should be avoided, as they do not bring anything beneficial.

project-base/storefront/pages/search.tsx
```diff
@@ -48,12 +51,14 @@ export const getServerSideProps = getServerSidePropsWrapper(({ redisClient, doma
         return redirect;
     }
 
-    return initServerSideProps({
-        context,
-        redisClient,
-        domainConfig,
-        t,
-    });
+    return initServerSideProps<TypeSearchQueryVariables | TypeSearchProductsQueryVariables | TypeAdvertsQueryVariables>(
```
sebaholesz: These types are used when you add prefetched queries. They help you catch mistakes in used variables, etc. However, you do not use any prefetched queries here, so I would say they are unnecessary

project-base/storefront/pages/order/transport-and-payment.tsx
```diff
@@ -25,7 +26,21 @@ const TransportAndPaymentPage: FC<ServerSidePropsType> = () => {
 export const getServerSideProps = getServerSidePropsWrapper(
     ({ redisClient, domainConfig, t }) =>
         async (context) =>
-            initServerSideProps({ context, redisClient, domainConfig, t }),
+            initServerSideProps({
```
sebaholesz: It would be nice if you've added the variable type here

```suggestion
            initServerSideProps<TypeAdvertsQueryVariables>({
```

project-base/storefront/pages/order/contact-information.tsx
```diff
@@ -32,7 +33,16 @@ export const getServerSideProps = getServerSidePropsWrapper(
                 redisClient,
                 domainConfig,
                 t,
-                prefetchedQueries: [{ query: CountriesQueryDocument }],
+                prefetchedQueries: [
```
sebaholesz, line 39: It would be nice if you've added the variable type here

```
initServerSideProps<TypeAdvertsQueryVariables>({
```

project-base/storefront/pages/categories/[categorySlug].tsx
```diff
@@ -134,6 +139,15 @@ export const getServerSideProps = getServerSidePropsWrapper(
                 context,
                 client,
                 ssrExchange,
+                prefetchedQueries: [
```
Comment (sebaholesz) [inline project-base/storefront/pages/categories/[categorySlug].tsx:145]: It would be nice if you've added the variable type here

```
initServerSideProps<TypeAdvertsQueryVariables>({
```

project-base/storefront/components/Blocks/Adverts/Adverts.tsx
```diff
@@ -1,10 +1,7 @@
...
-import { Image } from 'components/Basic/Image/Image';
+import { AdvertImage } from './AdvertImage';
 import { Webline } from 'components/Layout/Webline/Webline';
-import { TypeAdvertsFragment } from 'graphql/requests/adverts/fragments/AdvertsFragment.generated';
 import { useAdvertsQuery } from 'graphql/requests/adverts/queries/AdvertsQuery.generated';
 import { TypeCategoryDetailFragment } from 'graphql/requests/categories/fragments/CategoryDetailFragment.generated';
-import { Fragment } from 'react';
 import { twJoin } from 'tailwind-merge';
 
 type PositionNameType =
```
sebaholesz, line 7: This is defined, but not used in many places (where it could be beneficial). For example where the queries are called. Would you say it would be helpful to use it there, or not really?

project-base/storefront/components/Blocks/Adverts/Adverts.tsx
```diff
@@ -33,57 +30,25 @@ export const Adverts: FC<AdvertsProps> = ({
...
 }) => {
-    const [{ data: advertsData }] = useAdvertsQuery();
-
-    const filteredAdverts = advertsData?.adverts.filter((advert) => advert.positionName === positionName);
+    const [{ data: advertsData }] = useAdvertsQuery({
+        variables: {
+            categoryUuid: currentCategory?.uuid || null,
+            positionNames: [positionName],
+        },
+    });
```
sebaholesz: I have noticed that you call this query with multiple positions (e.g. in `getServerSideProps` in `[...categorySlug].tsx`) and then call it in components with a single position. This way the SSR cache won't work. You have to call the query with the same variables both on the server and on the client in order to make it work.

https://github.com/user-attachments/assets/8e686e92-0375-4453-80dc-df9d6ecba039

sebaholesz (changes requested):

project-base/storefront/components/Blocks/Adverts/Adverts.tsx
JanMolcik: I think it would be beneficial if I needed to wrap two elements and pass a key inside the `map` function. I don't see the case here, thus not needed. Or am I missing something?

sebaholesz (approved):

PR #3228, opened 2024-06-17
[project-base][SSP-1551] added more OP metatags

JanMolcik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Added more OG metatags like site name, twitter OG tags and more. Some OG titles are prefixed with a value (brand detail, store detail). Some pages now use default images from their data for the OG image metatag.
|New feature| No 
|Fixes issues| ... 

----
  - https://jm-ssp-1551-add-open-graph-tags.odin.shopsys.cloud
  - https://cz.jm-ssp-1551-add-open-graph-tags.odin.shopsys.cloud


UPGRADE-15.0.md
```diff
@@ -1318,3 +1318,8 @@ On Chrome there was issue when clicking on next/previous button while slider is
 #### change free transport limit in demo data ([#3199](https://github.com/shopsys/shopsys/pull/3199))
 
 -   social links in footer were blacked-out for cypress snapshots as they caused issues, so do not forget to add the blackout to snapshots where these links are visible, and also regenerate your screenshots
+
+#### add more OG meta tags ([#3228](https://github.com/shopsys/shopsys/pull/3228))
```
grossmannmartin: I suggest avoiding abbreviations... open graph will be more understandable

project-base/storefront/components/Layout/CommonLayout.tsx
```diff
@@ -46,6 +51,8 @@ export const CommonLayout: FC<CommonLayoutProps> = ({
                 defaultDescription={description}
                 defaultHreflangLinks={hreflangLinks}
                 defaultTitle={title}
+                ogImageUrlDefault={ogImageUrlDefault ?? undefined}
```
sebaholesz: Why do you have `?? undefined` here? I thought the only option is to have a string or undefined 🤔 that means you could just pass `ogImageUrlDefault`, or not?

project-base/storefront/public/locales/cs/common.json
```diff
@@ -173,6 +173,7 @@
     "Max allowed limit reached.": "Maximální povolené množství dosaženo.",
     "Menu": "Menu",
     "Message": "Zpráva",
+    "metatagSiteName": "Shopsys Platform",
```
sebaholesz, line 177: Not sure if I completely agree with this approach. I would ask the PO if this is the approach we want to take, rather than loading it from some more dynamic source. I will leave it for you to decide.

sebaholesz (changes requested):

project-base/storefront/components/Layout/CommonLayout.tsx
JanMolcik: Correct. My mistake, I will remove it.

project-base/storefront/public/locales/cs/common.json
JanMolcik: As the site name should be generic (website/brand/compady), I do not see any source in the administration that could be used other than SEO headline, but that one is specifically for the home page headline/title. IMO we should keep the metatagSiteName in the translation file - it should not change often.

UPGRADE-15.0.md
JanMolcik: Ok makes sense, adjusted.

sebaholesz (approved):

pk16011990 (approved):

PR #3229, opened 2024-06-17
[shopsys] related order items

### LINKED ISSUE #2677: Unused potential of OrderItem::$relatedItem
## Describe the bug
We have `$relatedItem` property in `OrderItem` class (https://github.com/shopsys/shopsys/blob/13.0/project-base/app/src/Model/Order/Item/OrderItem.php#L50) but it is basically not used. It is just set when creating a new order item of type `coupon` (which is actually an order item of type `OrderItem::TYPE_PRODUCT` at the moment while it deserves its own type, but we all know that :slightly_smiling_face:). The [getters](https://github.com/shopsys/shopsys/blob/13.0/project-base/app/src/Model/Order/Item/OrderItem.php#L123) are then never used if I am not mistaken.

I have two suggestions for how to improve the current implementation:
1. `$relatedItem` should be rather implemented as one-to-many relation (there might be more related items eg. when selecting a collection of services for a product). Something like this is better for a wider scale of use-cases IMHO:
```
    /**
     * @var \App\Model\Order\Item\OrderItem|null
     * @ORM\ManyToOne(targetEntity="App\Model\Order\Item\OrderItem", inversedBy="relatedOrderItems")
     * @ORM\JoinColumn(nullable=true)
     */
    private ?OrderItem $parentOrderItem;

    /**
     * @var \Doctrine\Common\Collections\Collection<int, \App\Model\Order\Item\OrderItem>
     * @ORM\OneToMany(targetEntity="App\Model\Order\Item\OrderItem", mappedBy="parentOrderItem")
     */
    private Collection $relatedOrderItems;
```
2. We should really use the relation when we approach the related items (on the storefront, in the mail, etc.) - at the moment, we just iterate through the items and rely on the fact the related items are created right after their "parent" item - ie. we rely on the following: `parentItem->getId() + 1 === $relatedItem->getId()`. If we messed up the ids, the output would be broken. Instead, I suggest iterating just the `PRODUCT_TYPE` items and under them, iterating through their `relatedItems` to display the related coupons, services, or whatever the related items might be on the particular implementation.
3. BONUS - as mentioned above, there should be a separate type for the `OrderItem` that represents the discount/coupon. Sad enough, there are two other issues (https://github.com/shopsys/shopsys/issues/1390 and https://github.com/shopsys/shopsys/issues/1263) and one pull request (https://github.com/shopsys/shopsys/pull/1563) for this :frowning_face:

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Order items now may be related to each other. This is currently used to have the connection between products and applied discounts
|New feature| Yes 
|Fixes issues| closes #2677  

----
  - https://mg-related-order-items.odin.shopsys.cloud
  - https://cz.mg-related-order-items.odin.shopsys.cloud


packages/framework/src/Migrations/Version20240617204710.php
```diff
@@ -0,0 +1,28 @@
...
+
+class Version20240617204710 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        if (!$this->isAppMigrationNotInstalledRemoveIfExists('Version20200624103557')) {
+            $this->sql('ALTER TABLE order_items DROP related_order_item_id');
```
vitek-rostislav, line 48: just a thought on WIP :innocent: I think we should provide a migration for those who actually use the relation on their project, what do you say?

grossmannmartin: added, good idea :)

packages/framework/src/Model/Order/PlaceOrderFacade.php
```diff
@@ -117,15 +118,53 @@ public function createOrderOnly(OrderData $orderData): Order
      */
     protected function fillOrderItems(Order $order, OrderData $orderData): void
     {
+        // the collection is used to keep track of which items have already been created
+        $allOrderItems = new ArrayCollection($orderData->items);
```
vitek-rostislav: maybe a better name would be `$(all)OrderItemsToBeCreated`? or something like that? I got that idea after reading the line bellow where the newly created element is removed from the collection.

vitek-rostislav: another idea, is not it possible to keep track just by using the cache (`$alreadyCreatedOrderItems`)? I was having quite a hard time understanding the following code, especially this part can be confusing:
```
if (!$allOrderItems->contains($relatedOrderItemData)) {
    $relatedOrderItem = $alreadyCreatedOrderItems[$this->generateCacheKey($relatedOrderItemData)];
    $orderItem->addRelatedItem($relatedOrderItem);

    continue;
}
```

If we had just that one array, it could look like this:
```
$key = $this->generateCacheKey($relatedOrderItemData);
if (array_key_exists($key, $alreadyCreatedOrderItems, true)) {
    $orderItem->addRelatedItem($alreadyCreatedOrderItems[$key)]);

    continue;
}
```

I'll be glad to discuss it

vitek-rostislav (changes requested):
Hi, I have added a few notes, I am open to discussion, of course :slightly_smiling_face: 
Also, what about this comment from the [issue](https://github.com/shopsys/shopsys/issues/2677)?

> We should really use the relation when we approach the related items (on the storefront, in the mail, etc.) - at the moment, we just iterate through the items and rely on the fact the related items are created right after their "parent" item - ie. we rely on the following: parentItem->getId() + 1 === $relatedItem->getId(). If we messed up the ids, the output would be broken. Instead, I suggest iterating just the PRODUCT_TYPE items and under them, iterating through their relatedItems to display the related coupons, services, or whatever the related items might be on the particular implementation.

packages/framework/src/Model/Order/PlaceOrderFacade.php
grossmannmartin: good idea – changed, thank 👍

packages/framework/src/Model/Order/Order.php
```diff
@@ -776,6 +784,38 @@ public function getItems()
         return $this->items->getValues();
     }
 
+    /**
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\OrderItem[]
+     */
+    public function getItemsSortedWithRelatedItems(): array
+    {
+        $return = [];
```
vitek-rostislav: maybe rather `$itemsSortedWithRelatedItems`? :innocent:

packages/framework/src/Model/Order/PlaceOrderFacade.php
```diff
@@ -119,32 +117,30 @@ protected function fillOrderItems(Order $order, OrderData $orderData): void
     {
         $alreadyCreatedOrderItems = [];
 
-        foreach ($this->orderItemTypeEnum->getAllCasesSortedByPriority() as $orderItemType) {
```
vitek-rostislav, line 120: I like how this method is simplified now :ok_hand:

vitek-rostislav (approved):

packages/framework/src/Model/Order/Order.php
grossmannmartin: renamed

PR #3242, opened 2024-06-28
[shopsys] top products are now in elasticsearch

AndrejBlaho:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Top products data are now stored in and queried on frontend from elasticsearch which should lead to better performance.
|New feature| No 
|Fixes issues| ... 

----
  - https://ab-top-products-in-elastic.odin.shopsys.cloud
  - https://cz.ab-top-products-in-elastic.odin.shopsys.cloud


packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -62,6 +65,8 @@ public function __construct(
         protected readonly ProductExportFieldProvider $productExportFieldProvider,
         protected readonly PricingGroupSettingFacade $pricingGroupSettingFacade,
         protected readonly ProductRepository $productRepository,
+        protected readonly TopProductFacade $topProductFacade,
+        protected array $domainTopProductsIndexedByProductId = [],
```
TomasLudvik: This should be only class property not dependency as it is not provided via dependency injection

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -145,6 +150,8 @@ protected function extractResult(Product $product, int $domainId, string $locale
      */
     protected function getExportedFieldValue(int $domainId, Product $product, string $locale, string $field): mixed
     {
+        $topProduct = $this->domainTopProductsIndexedByProductId[$product->getId()] ?? null;
+
```
TomasLudvik: It would be better to have a method for this to check if top products are loaded and load them if not and then return the value.

project-base/app/src/FrontendApi/Resolver/Products/PromotedProductsQuery.php
```diff
@@ -4,48 +4,14 @@
```
TomasLudvik: This class can be removed now, right?

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -62,6 +65,8 @@ public function __construct(
         protected readonly ProductExportFieldProvider $productExportFieldProvider,
         protected readonly PricingGroupSettingFacade $pricingGroupSettingFacade,
         protected readonly ProductRepository $productRepository,
+        protected readonly TopProductFacade $topProductFacade,
```
TomasLudvik: This should be mentioned in upgrade notes

packages/frontend-api/src/Model/Resolver/Products/PromotedProductsQuery.php
```diff
@@ -15,22 +16,21 @@ class PromotedProductsQuery extends AbstractQuery
      * @param \Shopsys\FrameworkBundle\Model\Product\TopProduct\TopProductFacade $topProductFacade
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
      * @param \Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser $currentCustomerUser
+     * @param \Shopsys\FrameworkBundle\Model\Product\ProductElasticsearchProvider $productElasticsearchProvider
      */
     public function __construct(
         protected readonly TopProductFacade $topProductFacade,
         protected readonly Domain $domain,
         protected readonly CurrentCustomerUser $currentCustomerUser,
+        protected readonly ProductElasticsearchProvider $productElasticsearchProvider,
```
TomasLudvik: This should be mentioned in the upgrade notes

TomasLudvik (changes requested):

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
AndrejBlaho: what about $variantCache bellow? should that be changed as well?

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
AndrejBlaho: I rewrote it. Is this how you meant it?

project-base/app/src/FrontendApi/Resolver/Products/PromotedProductsQuery.php
AndrejBlaho: I remember brief discussion with @vitek-rostislav about this and conclusion was keep it as it is for some reason, but it makes sense to remove it as you suggest, so i deleted it for now. But i am not sure about the reasoning and most correct approach so please clarify.

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
AndrejBlaho: added

packages/frontend-api/src/Model/Resolver/Products/PromotedProductsQuery.php
AndrejBlaho: added as well

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
TomasLudvik: Yes, but you will have to mention it in UPGRADE notes.

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -25,11 +25,15 @@
 use Shopsys\FrameworkBundle\Model\Product\ProductRepository;
 use Shopsys\FrameworkBundle\Model\Product\ProductVisibility;
 use Shopsys\FrameworkBundle\Model\Product\ProductVisibilityFacade;
+use Shopsys\FrameworkBundle\Model\Product\TopProduct\TopProductFacade;
 use Shopsys\FrameworkBundle\Model\Seo\HreflangLinksFacade;
 use Symfony\Contracts\Service\ResetInterface;
 
 class ProductExportRepository implements ResetInterface
 {
+    protected array $domainTopProductsIndexedByProductId = [];
```
TomasLudvik:
```suggestion
    /**
      * @property array<int, \Shopsys\FrameworkBundle\Model\Product\TopProduct\TopProduct>|null
      */
    protected ?array $domainTopProductsIndexedByProductId = null;
```

TomasLudvik: I have wrote this suggestion by hand, you might need to correct it :)

project-base/app/src/FrontendApi/Resolver/Products/PromotedProductsQuery.php
vitek-rostislav: well, the general reason is that when moving logic from project-base to the packages, it is easier to keep the empty class in project-base - we do not have a reverse function in the annotation fixer so we would have to rewrite all the usages manually then. Moreover, at least about some of the classes we can assume they will be overwritten on the particular project sooner or later (entities, data objects, facades, ...) so already extended empty project-base classes can help the developers a little in the future.

TLDR; I do not think the empty classes in the project-base can do any harm, rather it is the opposite.

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
TomasLudvik: I would like you to use that method to also retrieve that product, so it could be named `getDomainTopProductById`-

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -417,6 +433,20 @@ protected function getResults(QueryBuilder $queryBuilder, array $fields, int $do
         return $results;
     }
 
+    /**
+     * @param int $domainId
+     */
+    protected function loadTopProductsIfNotLoaded(int $domainId): void
+    {
+        if (empty($this->domainTopProductsIndexedByProductId)) {
```
TomasLudvik: What if second domain ID is used?

TomasLudvik (changes requested):

project-base/app/src/FrontendApi/Resolver/Products/PromotedProductsQuery.php
TomasLudvik: My opinion is, that it is ok to have empty classes for the basic functionality, that is always changed (entity, factory,...) but I would remove it for other less common scenarios. Reasons:
- searching for class is duplicate and you need to check if the functionality is extended or that class is empty
- annotation fixer causes changes in files that do not need to be reviewed
- projects have empty classes that are actually not extended

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -417,6 +432,24 @@ protected function getResults(QueryBuilder $queryBuilder, array $fields, int $do
...
     }
 
+    /**
+     * @param int $productId
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Product\TopProduct\TopProduct|null
+     */
+    protected function getTopProductByDomainId(int $productId, int $domainId): ?TopProduct
+    {
+        if (!isset($this->domainTopProductsIndexedByDomainIdAndProductId[$domainId])) {
```
TomasLudvik:
```suggestion
        if (!array_key_exists($domainId, $this->domainTopProductsIndexedByDomainIdAndProductId)) {
```

TomasLudvik (approved):

TomasLudvik (approved):

pk16011990 (approved):

PR #3243, opened 2024-06-28
[project-base][SSP-2445] Hydration error in wish list & product comparison page.

JanMolcik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixing hydration error by always rendering loading skeleton on the server.
|New feature| No 
|Fixes issues| Wish list and comparison page hydration error 

----
  - https://jm-ssp-2445-wishlist-hydration-error.odin.shopsys.cloud
  - https://cz.jm-ssp-2445-wishlist-hydration-error.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/utils/productLists/useProductList.ts
+++ b/project-base/storefront/utils/productLists/useProductList.ts
@@ -7,6 +7,7 @@ import { useProductListQuery } from 'graphql/requests/productLists/queries/Produ
 import { TypeProductListTypeEnum } from 'graphql/types';
 import { useEffect } from 'react';
 import { usePersistStore } from 'store/usePersistStore';
+import { useSessionStore } from 'store/useSessionStore';
 import { useIsUserLoggedIn } from 'utils/auth/useIsUserLoggedIn';
 
 export const useProductList = (
@@ -20,6 +21,8 @@ export const useProductList = (
         removeProductError: () => void;
     },
 ) => {
+    const isProductListHydrated = useSessionStore((s) => s.isProductListHydrated);
+    const updatePageLoadingState = useSessionStore((s) => s.updatePageLoadingState);
     const productListUuids = usePersistStore((s) => s.productListUuids);
     const authLoading = usePersistStore((s) => s.authLoading);
     const updateProductListUuid = useUpdateProductListUuid(productListType);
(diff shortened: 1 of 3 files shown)


project-base/storefront/utils/productLists/useProductList.ts
```diff
@@ -30,7 +31,7 @@ export const useProductList = (
     const [, TypeRemoveProductFromListMutation] = useRemoveProductFromListMutation();
     const [, removeListMutation] = useRemoveProductListMutation();
 
-    const [{ data: productListData, fetching: isProductListFetching }] = useProductListQuery({
+    const [{ data: productListData, fetching }] = useProductListQuery({
```
sebaholesz: if you have to rename it, please rename it to `isProductListQueryFetching`, or something like that. We have renamed this just recently to more descriptive names and would definitely like to keep it.

project-base/storefront/utils/productLists/useProductList.ts
```diff
@@ -40,6 +41,10 @@ export const useProductList = (
         pause: (!productListUuid && !isUserLoggedIn) || authLoading !== null,
     });
 
+    useEffect(() => {
+        setIsProductListFetching(fetching);
+    }, [fetching]);
```
sebaholesz: I understand your motivation and thinking here. However, I must say I am not a big fan of this approach. This useEffect and state is a separate logic from the query hook and it actually does not fully match the `fetching` variable above. If it would, you would not be able to use it like this, because it would cause the same hydration problem. Because of that, I would be more in favour of using the same approach as we do in cart (with hydration). This way you postpone the fetching of product lists, but still use the correct fetching variable.

image

sebaholesz (changes requested):

project-base/storefront/utils/productLists/useProductList.ts
JanMolcik: Ok thank you for the recommendation. I will adjust it.

project-base/storefront/utils/productLists/useProductList.ts
JanMolcik: Sure

sebaholesz (approved):

pk16011990 (approved):

PR #3247, opened 2024-07-04
[shopsys] remove delete button from banner slider image when editing

AndrejBlaho:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Image is mandatory when creating new homepage slider banner, but it is not while editing. This resulted in scenario where admin could remove image and save banner. Which was fine for backend, but on frontend where image was expected this created error state. As a simple solution we removed delete button from images when editing banner slider and thus image cannot be removed, but only replaced by uploading new one.
|New feature| No 
|Fixes issues| ... 

----
  - https://ab-mandatory-banner-slider-image.odin.shopsys.cloud
  - https://cz.ab-mandatory-banner-slider-image.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Form/ImageUploadType.php
+++ b/packages/framework/src/Form/ImageUploadType.php
@@ -49,6 +49,7 @@ class ImageUploadType extends AbstractType
             'multiple' => null,
             'image_entity_class' => null,
             'extensions' => ImageProcessor::SUPPORTED_EXTENSIONS,
+            'hide_delete_button' => false,
         ]);
 
         $resolver->setNormalizer(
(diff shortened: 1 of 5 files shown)


packages/framework/src/Form/ImageUploadType.php
```diff
@@ -143,6 +145,10 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     ],
                 ]),
             );
+
+        if (!$options['hide_delete_button']) {
+            $builder->setAttribute('hide_delete_button', false);
+        }
```
TomasLudvik: This should be set on line 52

TomasLudvik (changes requested):

TomasLudvik (approved):

PR #3259, opened 2024-07-10
[project-base][SSP-2393] Fixed unknown error in navigation

JanMolcik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Fixing unknown error after navigating to nonexistent URL from navigation bar.
|New feature| No 
|Fixes issues| SSP-2393 

----
  - https://jm-ssp-2393-unknown-navigation-error.odin.shopsys.cloud
  - https://cz.jm-ssp-2393-unknown-navigation-error.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuListItem.tsx
+++ b/project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuListItem.tsx
@@ -16,7 +16,6 @@ export const DropdownMenuListItem: FC<DropdownMenuListProps> = ({ navigationItem
             <ExtendedNextLink
                 className="flex-1 font-bold text-dark no-underline uppercase"
                 href={navigationItem.link}
-                type="category"
                 onClick={onNavigate}
             >
                 {navigationItem.name}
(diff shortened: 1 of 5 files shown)


project-base/storefront/middleware.ts
```diff
@@ -29,11 +29,6 @@ export const middleware: NextMiddleware = async (request) => {
 
         const { search } = new URL(request.url);
         const queryParams = new URLSearchParams(search);
-        const slugTypeQueryParam = queryParams.get('slugType');
-
-        if (slugTypeQueryParam) {
-            return rewriteDynamicPages(slugTypeQueryParam as FriendlyPageTypesValue, request.url, search);
-        }
```
sebaholesz: Your change essentially makes it so that with every navigation, we inquire about the type of page. This is obviously much more robust (we do not need to specify page types in ExtendedNextLink), but it takes longer (every request, every navigation). Moreover, if we wanted to keep this adjustment, we would end up with a really large amount of redundant code, which is there precisely so that we don’t have to inquire about the page type every time.

sebaholesz (changes requested):

project-base/storefront/middleware.ts
JanMolcik: I understand that now, thank you for an explanation. I have removed the type from `ExtendedNextLink` in `NavigationItem` component. This change triggers the script for the type lookup in middleware (fetch request to `resolve-friendly-url`) only for Navigation links.

sebaholesz (approved):

grossmannmartin (approved):

PR #3262, opened 2024-07-11
[framework] slider parameter values now have mandatory parameter numeric value

TomasLudvik:
| Q             | A
| ------------- | ---
|Description, the reason for the PR| Slider values need to be decimal to easily use SQL for number comparison as it is not possible with the string values.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://tl-add-numeric-value-parameter.odin.shopsys.cloud
  - https://cz.tl-add-numeric-value-parameter.odin.shopsys.cloud


packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -427,4 +380,107 @@ function ($data) {
             ],
         );
     }
+
+    protected function checkEnabledMailTemplatesHavetheirBodyAndSubjectFilled(): void
```
vitek-rostislav: nitpick, there should be a capital "T" - checkEnabledMailTemplatesHave**T**heirBodyAndSubjectFilled

packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -427,4 +380,107 @@ function ($data) {
...
+            $this->addErrorFlashTwig(
+                t('<a href="{{ url }}">Default unit is not set.</a>'),
+                [
+                    'url' => $this->generateUrl('admin_unit_list'),
+                ],
+            );
+        }
+    }
+
+    protected function checkMandatoryArticlesExists(): void
```
vitek-rostislav: nitpick
```diff
- checkMandatoryArticlesExists
+ checkMandatoryArticlesExist
```

packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -427,4 +380,107 @@ function ($data) {
...
+
+        $firstTenProducts = $this->parameterFacade->getFirstTenProductsWithSliderParameterValuesWithoutTheirsNumericValueFilled($this->localization->getAdminLocale());
+
+        foreach ($firstTenProducts as $product) {
+            $this->addErrorFlashTwig(
+                sprintf('<a href="%s">%s</a>', $this->generateUrl('admin_product_edit', ['id' => $product->getId()]), $product->getName()),
+            );
+        }
+
+        if ($countOfProductsWithoutSliderNumericValueSet === 10) {
```
vitek-rostislav: it should be like this, right?
```
if ($countOfProductsWithoutSliderNumericValueSet >= 10) {
```

packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -427,4 +380,107 @@ function ($data) {
...
+            $this->addErrorFlashTwig(
```
vitek-rostislav: Just a note on this, I do not think error flash with this text is something extra useful, maybe it could be a link to the complete list or something like that, maybe we can add an advanced filter that filters the products without the numeric values set, however, this should be defined by a product owner :shrug: I know that the dashboard warning was your initiative and I appreciate that as is :slightly_smiling_face:

packages/framework/src/Form/Admin/Product/Parameter/ProductParameterValueFormType.php
```diff
@@ -28,10 +33,21 @@ public function __construct(private readonly ParameterFacade $parameterFacade)
...
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $parameters = $this->parameterFacade->getAll();
+        $parameterAttributes = [];
+
+        $i = 0;
+
+        foreach ($parameters as $parameter) {
+            $parameterAttributes[$i] = ['data-type' => $parameter->getParameterType()];
+            $i++;
```
vitek-rostislav: is this necessary? IMHO you can use the loop key here, or?
```
foreach ($parameters as $key => $parameter) {
```
Or you can even use the parameter ID for the index :thinking:

packages/framework/src/Migrations/Version20240709134704.php
```diff
@@ -0,0 +1,26 @@
...
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20240709134704 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE parameter_values ADD numeric_value NUMERIC(20, 6) DEFAULT NULL');
```
vitek-rostislav, line 17: do we not want at least some basic use-case migration of data that would fill-in this column where possible?

packages/luigis-box/src/Model/Product/Parameter/Value/ParameterValueRepository.php
```diff
@@ -36,4 +36,25 @@ public function getExistingParameterValuesByValuesAndLocale(array $parameterValu
             ])
             ->orderBy(OrderByCollationHelper::createOrderByForLocale('pv.text', $locale))->getQuery()->getResult();
     }
+
+    /**
+     * @param string[] $parameterValues,
+     * @param string $locale
+     * @return array
```
vitek-rostislav:
```
\Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterValue[]
```

packages/luigis-box/src/Model/Product/Parameter/Value/ParameterValueRepository.php
```diff
@@ -36,4 +36,25 @@ public function getExistingParameterValuesByValuesAndLocale(array $parameterValu
...
+     */
+    public function getSliderParameterValuesForMinAndMaxByLocale(
+        array $parameterValues,
+        string $locale,
+    ): array {
+        return $this->em->createQueryBuilder()
+            ->select('pv')
+            ->from(ParameterValue::class, 'pv')
+            ->where('pv.numericValue IN(:parameterValues)')
```
vitek-rostislav, line 52: should not here be `BETWEEN`? I mean, this way you will get two `ParameterValue` instances that match the minimal and maximal value from the input `$parameterValues` array, right? I thought you needed to get all the values within the input range :thinking: If I am right, maybe it would be worth adding also some check that the input array really contains exactly two items

build.xml
```diff
@@ -469,6 +469,10 @@
                     <arg value="-r"/>
                     <arg value="copy('${path.upgrade-notes}/_template.md', '${path.upgrade-notes}/backend_' . date('Ymd_His') . '.md');"/>
                 </exec>
+                <exec executable="php" escape="false" passthru="true">
+                    <arg value="-r"/>
+                    <arg value="echo 'File generated: ${path.upgrade-notes}/backend_' . date('Ymd_His') . '.md';"/>
```
vitek-rostislav: Love the idea, however, is not there a danger of changing the time between generating the file and outputting the name? I would rather save the file name in a property and then output it :thinking: This is what I got from chatbot (not tested though):
```
<!-- Set the timestamp in a property -->
                <exec executable="php" outputproperty="currentTimestamp">
                    <arg value="-r"/>
                    <arg value="echo date('Ymd_His');"/>
                </exec>
                
                <!-- Set the filename using the timestamp property -->
                <property name="upgradeNotesFilename" value="${path.upgrade-notes}/backend_${currentTimestamp}.md"/>
                
                <!-- Generate the file -->
                <exec executable="php">
                    <arg value="-r"/>
                    <arg value="copy('${path.upgrade-notes}/_template.md', '${upgradeNotesFilename}');"/>
                </exec>
                
                <!-- Output the filename -->
                <exec executable="php" escape="false" passthru="true">
                    <arg value="-r"/>
                    <arg value="echo 'File generated: ${upgradeNotesFilename}';"/>
                </exec>
```

packages/framework/src/Model/Product/Parameter/ParameterValue.php
```diff
@@ -34,6 +35,12 @@ class ParameterValue
      */
     protected $text;
 
+    /**
+     * @var string|null
+     * @ORM\Column(type="decimal", precision=20, scale=6, nullable=true)
+     */
+    protected $numericValue;
```
vitek-rostislav, line 42: based on the acceptance criteria, this value should now be used [here](https://github.com/shopsys/shopsys/blob/15.0/project-base/app/src/Model/Product/Elasticsearch/ProductExportRepository.php#L457), right?

I do not want to speak about the fact the linked code is in `project-base` :melting_face:

vitek-rostislav (changes requested):

packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -59,6 +63,8 @@ public function __construct(
         protected readonly DateTimeFormatterExtension $dateTimeFormatterExtension,
         protected readonly TransferIssueFacade $transferIssueFacade,
         protected readonly Domain $domain,
+        protected readonly ParameterFacade $parameterFacade,
+        protected readonly Localization $localization,
```
vitek-rostislav: this should be mentioned in the upgrade notes

vitek-rostislav (changes requested):

packages/framework/src/Migrations/Version20240709134704.php
TomasLudvik: I do not know the format of the data on projects, so I do not feel like we can even do some migration.

packages/luigis-box/src/Model/Product/Parameter/Value/ParameterValueRepository.php
TomasLudvik: Actually, for slider you need only two, because of ProductFilterOption::__construct

packages/framework/src/Controller/Admin/DefaultController.php
TomasLudvik: For me is this form "good enough" - you can fix those ten and come back for others.

packages/framework/src/Form/Admin/Product/Parameter/ProductParameterValueFormType.php
TomasLudvik: I found it depends on order, so I have kept it, it did not work with IDs

build.xml
TomasLudvik: I have reworked it

packages/luigis-box/src/Model/Product/Parameter/Value/ParameterValueRepository.php
vitek-rostislav: ok, I have to believe you :smile:

packages/framework/src/Migrations/Version20240709134704.php
vitek-rostislav: ok

vitek-rostislav (approved):

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -413,4 +413,47 @@ public function getParameterValueById(int $parameterValueId): ParameterValue
...
+    /**
+     * @param string $locale
+     * @return int
+     */
+    public function getCountOfProductsWithSliderParameterValuesWithoutTheirsNumericValueFilled(string $locale): int
+    {
+        return $this->getAllProductsWithSliderParameterValuesWithoutTheirsNumericValueFilledQueryBuilder($locale)
+            ->select('COUNT(DISTINCT product.id)')
+            ->getQuery()
+            ->getSingleScalarResult();
```
pk16011990: If no product is in the database, it throws `NoResultException`. It is a problem for a new empty eshop.

packages/framework/src/Controller/Admin/ParameterValueController.php
```diff
@@ -104,4 +108,49 @@ public function editAction(Request $request, int $id): Response
...
+
+        $form = $this->createForm(SliderParameterValuesUpdateFormType::class, $parameterValuesConversionDataIndexedByParameterValueId, [
+            'data' => $parameterValuesConversionDataIndexedByParameterValueId,
+        ]);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $this->parameterFacade->updateParameterValuesByConversion($parameter, $parameterValuesConversionDataIndexedByParameterValueId);
+            $this->addSuccessFlashTwig(
+                t('Parameter values associated with parameter - <strong><a href="{{ url }}">{{ parameter.name }}</a></strong> modified.'),
```
vitek-rostislav: the usage of `{{ parameter.name }}` seems quite unusual. Why do not you use just `{{ parameterName }}` variable?

packages/framework/src/Form/Admin/Product/Parameter/Value/SliderParameterValuesUpdateFormType.php
```diff
@@ -0,0 +1,49 @@
...
+
+class SliderParameterValuesUpdateFormType extends AbstractType
+{
+    /**
+     * @param \Symfony\Component\Form\FormBuilderInterface $builder
+     * @param array $options
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        //        d($options);
```
vitek-rostislav: remove this, please

packages/framework/src/Model/Product/Parameter/ParameterFacade.php
```diff
@@ -234,4 +238,53 @@ public function getParameterValueById(int $parameterValueId): ParameterValue
...
+    ): void {
+        foreach ($parameterValuesConversionDataIndexedByParameterValueId as $parameterValueId => $parameterValueConversionData) {
+            $parameterValue = $this->parameterRepository->getParameterValueById($parameterValueId);
+            $parameterValueData = $this->parameterValueDataFactory->createFromParameterValue($parameterValue);
+
+            $parameterValueData->uuid = null;
+            $parameterValueData->text = $parameterValueConversionData->newValueText;
+            $parameterValueData->numericValue = $parameterValueConversionData->newValueText;
+
+            $newParameterValue = $this->parameterValueFactory->create($parameterValueData);
```
vitek-rostislav, line 273: why do not we just update the existing values?

`updateParameterValueInProductsByConversion` would be unnecessary then, right?

packages/framework/src/Resources/views/Admin/Form/parameterValueConversionList.html.twig
```diff
@@ -0,0 +1,49 @@
+{% macro parameterValueRow(parameterValue, index) %}
+    <tr class="js-parameters-item js-form-group table-form__row" data-index="{{ index }}">
```
vitek-rostislav: the `js-parameters-item` css class is not necessary here, right?

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -2380,6 +2398,12 @@ msgstr "Zaplacená částka"
 msgid "Parameter"
 msgstr "Parametr"
 
+msgid "Parameter <strong><a href=\"{{ url }}\">{{ name }}</a></strong> created"
+msgstr "Parametr <strong><a href=\"{{ url }}\">{{ name }}</a></strong> byl vytvořen
```
vitek-rostislav: missing ending quote

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -2392,6 +2416,12 @@ msgstr "Hodnota parametru <strong><a href=\"{{ url }}\">{{ parameterValue.text }
 msgid "Parameter value of type color"
 msgstr "Hodnota parameteru typu barva"
 
+msgid "Parameter values"
+msgstr "Hodnoty parametru"
+
+msgid "Parameter values associated with parameter - <strong><a href=\"{{ url }}\">{{ parameter.name }}</a></strong> modified."
+msgstr "Hodnoty parametru spojené s parametrem - <strong><a href=\"{{ url }}\">{{ parameter.name }}</a></strong> byly upraveny.
```
vitek-rostislav: missing ending quote

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -321,7 +320,14 @@ protected function getParameterValuesIndexedByProductIdAndParameterName(
     public function getParametersByUuids(array $uuids): array
     {
         $parametersByUuid = [];
-        $parameters = $this->getParameterRepository()->findBy(['uuid' => $uuids]);
```
vitek-rostislav, line 323: btw, you can define the ordering as a second parameter of the `findBy` method :slightly_smiling_face:

vitek-rostislav (changes requested):

packages/framework/src/Model/Product/Parameter/ParameterFacade.php
TomasLudvik: Because they might be associated with other types of parameters

vitek-rostislav (approved):

Makefile
```diff
@@ -1,11 +1,11 @@
 generate-schema:
-	docker-compose exec php-fpm php phing frontend-api-generate-graphql-schema
+	docker compose exec php-fpm php phing frontend-api-generate-graphql-schema
```
vitek-rostislav, line 2: there are a few articles in https://docs.shopsys.com/en/14.0/installation/ where the minimal docker and docker compose versions are listed, could you update them, please?

project-base/app/src/DataFixtures/Demo/ParameterDataFixture.php
```diff
@@ -88,7 +88,6 @@ class ParameterDataFixture extends AbstractReferenceFixture implements Dependent
     public const string PARAM_WARRANTY_IN_YEARS = 'warranty_in_years';
     public const string PARAM_WATER_RESERVOIR_CAPACITY = 'water_reservoir_capacity';
     public const string PARAM_WEIGHT = 'weight';
-    public const string PARAM_WEIGHT_KG = 'weight_kg';
```
vitek-rostislav, line 91: nitpick regarding the commit message, "unified" is IMHO more appropriate word, see https://thecontentauthority.com/blog/unified-vs-united

vitek-rostislav (changes requested):

vitek-rostislav (approved):

PR #3263, opened 2024-07-11
[project-base][SSP-1581] refactored form validations

chlebektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Refactor and unification form validations.
|New feature| No 
|Fixes issues| ... 

----
  - https://tc-ssp-1581-refactor-form-validations.odin.shopsys.cloud
  - https://cz.tc-ssp-1581-refactor-form-validations.odin.shopsys.cloud


project-base/storefront/components/Forms/validationRules.ts
```diff
@@ -148,9 +148,9 @@ export const validateCompanyTaxNumber = (t: Translate): BaseSchema => {
         );
 };
 
-export const validateFirstPassword = (t: Translate): BaseSchema => {
+export const validatePassword = (t: Translate): BaseSchema => {
     return Yup.string()
-        .required(t('Please enter first password'))
+        .required(t('Please enter password'))
         .min(
```
sebaholesz, line 158: I know this code has been here since ages, but I am thinking if the "repeated" ("again") password should not only be validated by comparison with the "first" password. I think it should be its only validation rule, but this is a question for the PO

project-base/storefront/components/Forms/validationRules.ts
```diff
@@ -159,9 +159,9 @@ export const validateFirstPassword = (t: Translate): BaseSchema => {
         );
 };
 
-export const validateSecondPassword = (t: Translate): BaseSchema => {
+export const validatePasswordAgain = (t: Translate): BaseSchema => {
```
sebaholesz: Maybe more like 

```suggestion
export const validateRepeatedPassword = (t: Translate): BaseSchema => {
```

project-base/storefront/components/Forms/validationRules.ts
```diff
@@ -170,3 +170,26 @@ export const validateSecondPassword = (t: Translate): BaseSchema => {
...
+        .required(t('Please enter new password'))
+        .min(
+            VALIDATION_CONSTANTS.passwordMinLength,
+            t('Password must be at least {{ count }} characters long', {
+                count: VALIDATION_CONSTANTS.passwordMinLength,
+            }),
+        );
+};
+
+export const validateNewPasswordAgain = (t: Translate): BaseSchema => {
```
sebaholesz: Just a suggestion. Up to you...

```suggestion
export const validateRepeatedNewPassword = (t: Translate): BaseSchema => {
```

project-base/storefront/components/Layout/Footer/NewsletterForm/newsletterFormMeta.tsx
```diff
@@ -14,8 +15,8 @@ export const useNewsletterForm = (): [UseFormReturn<NewsletterFormType>, Newslet
     const { t } = useTranslation();
     const resolver = yupResolver(
         Yup.object().shape<Record<keyof { email: string; privacyPolicy: boolean }, any>>({
-            email: Yup.string().required(t('This field is required')).email(t('This value is not a valid email')),
-            privacyPolicy: Yup.bool().oneOf([true], t('You have to agree with our privacy policy')),
+            email: validateEmail(t),
+            privacyPolicy: Yup.boolean().isTrue(t('You have to agree with our privacy policy')),
```
sebaholesz: I can see this code in at least 3 places. Maybe it would be nice to extract it...

project-base/storefront/components/Pages/Customer/customerChangeProfileFormMeta.ts
```diff
@@ -29,16 +29,8 @@ export const useCustomerChangeProfileForm = (
...
-                then: validateFirstPassword(t),
-                otherwise: Yup.string(),
-            }),
-            passwordSecond: Yup.string().when('passwordFirst', {
-                is: (passwordFirst: string) => passwordFirst.length > 0,
-                then: validateSecondPassword(t),
-                otherwise: Yup.string(),
-            }),
+            passwordFirst: validatePassword(t),
+            passwordSecond: validatePasswordAgain(t),
```
sebaholesz: This change has broken the current logic of customer edit form. The logic before was that if the old password was not provided, it was assumed that the user is not trying to change his password and passwords were therefore not validated. Your change results in passwords being always validated.

sebaholesz: In this video (your state) you can see the unnecessary/unwanted validation

https://github.com/shopsys/shopsys/assets/55181354/3c8ce015-5084-4017-ae9b-002cd8bcdaba

sebaholesz: In this video (previous state), you can see that if I did not provide old password, new ones were not validated


https://github.com/shopsys/shopsys/assets/55181354/64cf03c1-0a54-496c-92b2-4cb8c055f94e

project-base/storefront/components/Forms/validationRules.ts
```diff
@@ -170,3 +170,26 @@ export const validateSecondPassword = (t: Translate): BaseSchema => {
...
+    return Yup.string()
+        .required(t('Please enter new password again'))
+        .min(
+            VALIDATION_CONSTANTS.passwordMinLength,
+            t('Password must be at least {{ count }} characters long', {
+                count: VALIDATION_CONSTANTS.passwordMinLength,
+            }),
+        )
+        .oneOf([Yup.ref('newPassword'), null], t('Passwords must match'));
```
sebaholesz: Seems like you've missed the `otherwise: Yup.string(),` case from the original code. Was that on purpose?

upgrade-notes/storefront_20240711_123953.md
```diff
@@ -0,0 +1,3 @@
+#### refactored form validations ([#3263](https://github.com/shopsys/shopsys/pull/3263))
+
+-   refactor and unification form validations.
```
sebaholesz: This should also tell developers what they should do in their code, such as:

- modify new validation rules (add their names) so that it matches your wanted validation logic
- use new validation rules in other forms you might have to fully utilize the new code
- etc.

sebaholesz (changes requested):

project-base/storefront/components/Pages/Customer/customerChangeProfileFormMeta.ts
```diff
@@ -29,14 +29,15 @@ export const useCustomerChangeProfileForm = (
...
             passwordFirst: Yup.string().when('passwordOld', {
                 is: (passwordOld: string) => passwordOld.length > 0,
-                then: validateFirstPassword(t),
+                then: validatePassword(t),
                 otherwise: Yup.string(),
             }),
             passwordSecond: Yup.string().when('passwordFirst', {
                 is: (passwordFirst: string) => passwordFirst.length > 0,
-                then: validateSecondPassword(t),
+                then: validateRepeatedPassword(t),
```
sebaholesz: There are quite some discrepancies in naming of all these password variables and validation rules. This is true on multiple levels:
- one name is given by the API (e.g. we have `newPassword`)
- another name is given by the form definition (e.g. we have `passwordFirst`)
- and sometimes another name is also given by the name of the validation rule (e.g. we have `passwordSecond` vs `validateRepeatedPassword`)

The best approach IMO would be to follow API naming as much as possible and as much as it makes sense (of course, API does not have any repeated passwords, but that should be easy to solve). You can find the name of all API argument names in the YAML definitions for a given mutation (e.g. `ChangePasswordInputDecorator`)

sebaholesz (changes requested):

project-base/storefront/components/Forms/validationRules.ts
```diff
@@ -163,37 +163,30 @@ export const validatePassword = (t: Translate): BaseSchema => {
         );
 };
 
-export const validateRepeatedPassword = (t: Translate): BaseSchema => {
+const passwordConfirmValidationSchema = (t: Translate, fieldName: string, requiredMessage: string) => {
```
sebaholesz: Please, rename `fieldName` to something more descriptive. I suppose this is the first password's field name, so name it as such. This could be `passwordFieldName` or similar.

project-base/storefront/components/Pages/Registration/registrationFormMeta.tsx
```diff
@@ -90,19 +90,19 @@ type RegistrationFormMetaType = {
...
             label: string;
             errorMessage: string | undefined;
         };
-        passwordSecond: {
-            name: 'passwordSecond';
+        passwordConfirm: {
+            name: 'passwordConfirm';
             label: string;
             errorMessage: string | undefined;
         };
```
sebaholesz: I would just remove these extra definitions
```ts
type RegistrationFormMetaType = {
    formName: string;
    messages: {
        error: string;
        successAndLogged: string;
    };
    fields: {
        [key in keyof RegistrationFormType]: {
            name: key;
            label: string | JSX.Element;
            errorMessage: string | undefined;
        };
    };
};
```

and then go to PasswordInputControlled.tsx:13 and change it to `label: ReactNode;`

the `TextInput` component which is used internally in `PasswordInputControlled` already accepts label as a ReactNode, so this is the only thing that has to be changed.

sebaholesz (changes requested):

sebaholesz (approved):

PR #3270, opened 2024-07-16
[project-base][SSP-2280] added delay to open navigation menu

chlebektomas:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Navigation menu is now more user-friendly with the addition of a delay.
|New feature| No 
|Fixes issues| ... 

----
  - https://tc-ssp-ssp-2280-navigation-menu-open-delay.odin.shopsys.cloud
  - https://cz.tc-ssp-ssp-2280-navigation-menu-open-delay.odin.shopsys.cloud


project-base/storefront/components/Layout/Header/Navigation/Navigation.tsx
```diff
@@ -7,7 +7,7 @@ export type NavigationProps = {
 
 export const Navigation: FC<NavigationProps> = ({ navigation }) => {
     return (
-        <ul className="relative hidden w-full lg:flex lg:gap-6 xl:gap-12">
+        <ul className="relative hidden w-full lg:flex">
```
sebaholesz, line 10: Not sure if this is what we want. Looking at the new design in Figma, I can see that the gap between menu items is set to 40px (tailwind size 10).

sebaholesz, line 10: It also means that the link can be now hovered even if the mouse cursor is actually far from the link. This is because the area between links is split between the two links (see video). I am not saying this is necessarily wrong, but I do not see this in ACs and I am not sure if we want it to behave this way.

https://github.com/user-attachments/assets/d3748d96-aac7-4288-85f9-ac73fa133897

sebaholesz (changes requested):

project-base/storefront/components/Layout/Header/Navigation/Navigation.tsx
chlebektomas: After consultation, the project owner confirmed that this is a valid behaviour.

project-base/storefront/components/Layout/Header/Navigation/NavigationItem.tsx
```diff
@@ -12,22 +13,30 @@ type NavigationItemProps = {
...
 
     return (
         <li className="group" onMouseEnter={() => setIsMenuOpened(true)} onMouseLeave={() => setIsMenuOpened(false)}>
             <ExtendedNextLink
                 href={navigationItem.link}
                 className={twJoin(
-                    'relative m-0 flex items-center px-2 py-4 text-sm font-bold uppercase text-white no-underline hover:text-secondary hover:no-underline group-hover:text-secondary group-hover:no-underline vl:text-base',
+                    'relative m-0 flex items-center px-6 xl:px-5 py-4 group-first-of-type:pl-0 text-sm font-bold uppercase text-white hover:text-white no-underline hover:no-underline vl:text-base transition-colors',
+                    (isMenuOpenedDelayed || !hasChildren) &&
+                        'hover:text-secondary group-hover:text-secondary group-hover:no-underline',
```
sebaholesz, line 25: After changing these styles, you also have to change `NavigationPlaceholder`, because otherwise their styles do not match, which causes a layout shift:

https://github.com/user-attachments/assets/50eb7776-f9a3-4f82-a06b-2187002ad9d1

sebaholesz (changes requested):

sebaholesz (approved):

PR #3277, opened 2024-07-23
[frontend-api] social network login types config

vitek-rostislav:
| Q             | A
| ------------- | ---
|Description, reason for the PR| We do not want Storefront to define the social login types when this is configured on the BE already. We want to pass the configuration via API so there is one source of truth only. Also, it is now possible to enable/disable particular social networks for particular domains separately.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://rv-socnet-config-api.odin.shopsys.cloud
  - https://cz.rv-socnet-config-api.odin.shopsys.cloud


TomasLudvik (approved):

project-base/storefront/components/Blocks/Login/LoginForm.tsx
```diff
@@ -80,6 +96,22 @@ export const LoginForm: FC<LoginFormProps> = ({
...
+                    cartUuid,
+                    shouldOverwriteCustomerUserCart,
+                    productListUuids,
+                )}
+            >
+                {getSocialNetworkIcon(socialNetwork)}
+            </ExtendedNextLink>
+        );
+    };
+
```
JanMolcik: Defining components inside the scope of other components is considered antipattern (the component gets recreated on each render instead of re-rendering it only when its properties change). It is best practice to define them outside of the scope and provide required data through props. You could even consider splitting them into their own files and making them reusable, but I don't know if this applies to this case.

project-base/storefront/components/Blocks/Login/LoginForm.tsx
```diff
@@ -60,6 +63,19 @@ export const LoginForm: FC<LoginFormProps> = ({
...
+                return <FacebookIcon className="w-7 text-[#1877f2]" />;
+            case TypeLoginTypeEnum.Google:
+                return <GoogleIcon className="w-6" />;
+            case TypeLoginTypeEnum.Seznam:
+                return <SeznamIcon className="w-6" />;
+            default:
+                return null;
+        }
+    };
+
```
JanMolcik: I would convert it to a stateless functional component instead of using it as a getter function. The intent of the component would be clearer and we could use the `<SocialNetworkIcon ... />` syntax.

```suggestion
    const SocialNetworkIcon = ({ socialNetwork }: { socialNetwork: TypeLoginTypeEnum }) => {
        switch (socialNetwork) {
            case TypeLoginTypeEnum.Facebook:
                return <FacebookIcon className="w-7 text-[#1877f2]" />;
            case TypeLoginTypeEnum.Google:
                return <GoogleIcon className="w-6" />;
            case TypeLoginTypeEnum.Seznam:
                return <SeznamIcon className="w-6" />;
            default:
                return null;
        }
    };

```

JanMolcik (changes requested):
Nice! Just suggested two improvements.

project-base/storefront/components/Blocks/Login/LoginForm.tsx
JanMolcik: And I would also move it outside of the scope of the `LoginForm` as suggested above.

project-base/storefront/components/Blocks/Login/LoginForm.tsx
```diff
@@ -115,38 +147,9 @@ export const LoginForm: FC<LoginFormProps> = ({
                             </FormButtonWrapper>
 
                             <div className="flex gap-2 lg:order-2 w-full justify-center">
```
chlebektomas: I would consider checking the length of the `settingsData?.settings?.socialNetworkLoginConfig` to prevent an empty div if there is no data.

project-base/storefront/components/Blocks/Login/LoginForm.tsx
vitek-rostislav: :ok_hand: fixed in 98beaf69d1

project-base/storefront/components/Blocks/Login/LoginForm.tsx
vitek-rostislav: :ok_hand: fixed in 98beaf69d1

project-base/storefront/components/Blocks/Login/LoginForm.tsx
vitek-rostislav: :ok_hand: fixed in 98beaf69d1

chlebektomas (approved):

JanMolcik (approved):

TomasLudvik (approved):

pk16011990 (approved):

PR #3285, opened 2024-07-25
[framework] address, name and some other user data is now nullable

AndrejBlaho:
| Q             | A
| ------------- | ---
|Description, reason for the PR| Due to social media login it is now possible that user is created without certain data like address and name that are mandatory throughout the application. These data are still mandatory in application and user is required to fill them ich chooses to edit profile or finish order, but they were changed to nullable in database layer so interim state with empty values is also valid
|New feature| No 
|Fixes issues| ... 

----
  - https://ab-nullable-user-data.odin.shopsys.cloud
  - https://cz.ab-nullable-user-data.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Model/Customer/User/CustomerUserUpdateDataFactory.php
+++ b/packages/framework/src/Model/Customer/User/CustomerUserUpdateDataFactory.php
@@ -106,6 +106,10 @@ class CustomerUserUpdateDataFactory implements CustomerUserUpdateDataFactoryInte
             $customerUser->getLastName(),
             $order->getLastName(),
         );
+        $customerUserUpdateData->customerUserData->telephone = Utils::ifNull(
+            $customerUser->getTelephone(),
+            $order->getTelephone(),
+        );
         $customerUserUpdateData->billingAddressData = $this->getAmendedBillingAddressDataByOrder(
             $order,
             $billingAddress,
--- a/packages/framework/tests/Unit/Model/Customer/TestCustomerProvider.php
+++ b/packages/framework/tests/Unit/Model/Customer/TestCustomerProvider.php
@@ -54,6 +54,7 @@ class TestCustomerProvider
         $customerUserData->firstName = 'Firstname';
         $customerUserData->lastName = 'Lastname';
         $customerUserData->email = 'no-reply@shopsys.com';
+        $customerUserData->telephone = 'telephone';
         $customerUserData->password = 'pa55w0rd';
         $customerUserData->domainId = Domain::FIRST_DOMAIN_ID;
         $customerUserData->customer = $customer;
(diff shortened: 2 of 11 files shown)


packages/framework/src/Model/Customer/User/CustomerUserUpdateDataFactory.php
```diff
@@ -132,7 +136,7 @@ protected function getAmendedBillingAddressDataByOrder(Order $order, BillingAddr
     {
         $billingAddressData = $this->billingAddressDataFactory->createFromBillingAddress($billingAddress);
 
-        if ($billingAddress->getStreet() === null) {
+        if (empty($billingAddress->getStreet())) {
```
vitek-rostislav: it is not allowed to use `empty`, see the failed standards :slightly_smiling_face: 
Is that change because of possible empty strings? Is it worth adding a migration that updates empty strings to null in the relevant columns? :thinking:

upgrade-notes/backend_20240725_061952.md
```diff
@@ -0,0 +1,3 @@
+#### address, name and some other user data is now nullable ([#3285](https://github.com/shopsys/shopsys/pull/3285))
+
```
vitek-rostislav, line 2: it would be nice to list here all the properties that are now nullable

vitek-rostislav (changes requested):

vitek-rostislav (changes requested):
Hi, check the failing tests and standards. please

packages/framework/tests/Unit/Model/Customer/TestCustomerProvider.php
```diff
@@ -47,6 +47,7 @@ public static function getTestCustomerUserData(bool $isCompany = true): Customer
         $customerUserData->firstName = 'Firstname';
         $customerUserData->lastName = 'Lastname';
         $customerUserData->email = 'no-reply@shopsys.com';
+        $customerUserData->telephone = 'telephone';
```
vitek-rostislav, line 57: Hmm, this indeed fixes the test but it reveals that we are missing a test for the social network login scenario:

1. a customer user does not have saved any nullable data (basically he has email only, the rest is null)
2. he creates an order where he fills in all the data
3. assert customer user data are amended with the order data

Could you please add such a test?

Btw, while testing this, I noticed the data update still does not work properly for the billing address data (I registered via google, created an order but in my profile, there are still no billing data). I did not try it with the delivery address data.

I believe that at least one of the problems is in `RegistrationDataFactory::createFromSocialNetworkProfile()` where empty strings are still used.

vitek-rostislav (dismissed):

grossmannmartin (approved):

PR #3287, opened 2024-07-25
[project-base][SSP-2513] Fixed skeleton loaders while navigating through main navigation

JanMolcik:
| Q             | A
| ------------- | ---
|Description, reason for the PR| To fix a bug in which the skeleton loaders would not show, I have added a check for an external link, for which we don't switch to the loading state, and for the rest (internal links), if no type provided, we show category skeleton loader by default.
|New feature| No 
|Fixes issues| [SSP-2513](https://shopsys.atlassian.net/browse/SSP-2513?atlOrigin=eyJpIjoiMmFlNjgyZmI2OTZjNGY0Yzk2ZWQyM2QzMDU0NDIyZDAiLCJwIjoiaiJ9) 

----
  - https://jm-ssp-2513-navigation-skeleton-bug.odin.shopsys.cloud
  - https://cz.jm-ssp-2513-navigation-skeleton-bug.odin.shopsys.cloud

### DIFF
--- a/project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
+++ b/project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
@@ -1,3 +1,4 @@
+import { useDomainConfig } from 'components/providers/DomainConfigProvider';
 // eslint-disable-next-line no-restricted-imports
 import NextLink, { LinkProps } from 'next/link';
 import { ComponentPropsWithoutRef, MouseEventHandler } from 'react';
@@ -9,12 +10,14 @@ import {
     FriendlyPagesTypesKey,
     FriendlyPagesTypesKeys,
 } from 'types/friendlyUrl';
+import { UrlObject } from 'url';
 import { SLUG_TYPE_QUERY_PARAMETER_NAME } from 'utils/queryParamNames';
 
 export type ExtendedNextLinkProps = Omit<ComponentPropsWithoutRef<'a'>, keyof LinkProps> &
     Omit<LinkProps, 'prefetch'> & {
         queryParams?: Record<string, string>;
         type?: PageType;
+        skeletonType?: PageType;
     };
 
 export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
@@ -24,9 +27,11 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
     as,
     onClick,
     type,
+    skeletonType,
     ...props
 }) => {
     const updatePageLoadingState = useSessionStore((s) => s.updatePageLoadingState);
+    const { url } = useDomainConfig();
 
     const isDynamic = type && FriendlyPagesTypesKeys.includes(type as any);
 
@@ -37,11 +42,11 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
         if (isWithoutOpeningInNewTab) {
             onClick?.(e);
 
-            if (type) {
-                updatePageLoadingState({ isPageLoading: true, redirectPageType: type });
-            } else {
-                updatePageLoadingState({ redirectPageType: undefined });
-            }
+            const isLinkExternal = isHrefExternal(href, url);
+            updatePageLoadingState({
+                isPageLoading: !!type || !isLinkExternal,
+                redirectPageType: type ?? skeletonType,
+            });
         }
     };
 
@@ -66,4 +71,18 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
             {children}
         </NextLink>
     );
+};
+
+const isHrefExternal = (href: string | UrlObject, baseUrl: string) => {
+    const currentHostname = new URL(baseUrl).hostname;
+
+    if (typeof href === 'object') {
+        return currentHostname !== href.hostname;
+    }
+
+    try {
+        return currentHostname !== new URL(href).hostname;
+    } catch (e) {
+        return false;
+    }
 };
--- a/project-base/storefront/components/Layout/Header/Navigation/DeferredNavigation.tsx
+++ b/project-base/storefront/components/Layout/Header/Navigation/DeferredNavigation.tsx
@@ -3,6 +3,8 @@ import { useNavigationQuery } from 'graphql/requests/navigation/queries/Navigati
 import dynamic from 'next/dynamic';
 import { useDeferredRender } from 'utils/useDeferredRender';
 
+const DEFAULT_SKELETON_TYPE = 'category';
+
 const NavigationPlaceholder = dynamic(() =>
     import('./NavigationPlaceholder').then((component) => component.NavigationPlaceholder),
 );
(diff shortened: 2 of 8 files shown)


project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -27,6 +29,7 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
     ...props
 }) => {
     const updatePageLoadingState = useSessionStore((s) => s.updatePageLoadingState);
+    const { url: baseUrl } = useDomainConfig();
```
sebaholesz: We try not to rename this parameter. Is it necessary?

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -67,3 +70,22 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
...
+        if (!linkHostname) {
+            return false;
+        }
+        return currentHostname !== linkHostname;
+    }
+
+    try {
+        return currentHostname !== new URL(href).hostname;
+    } catch (e) {
+        // URL is not valid -> relative pathname -> internal link
```
sebaholesz: I think the code is self-explanatory. We try to avoid comments, because there is nothing that ages quicker than comments in code.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -67,3 +70,22 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
...
+
+const isHrefExternal = (href: string | UrlObject, baseUrl: string) => {
+    const currentHostname = new URL(baseUrl).hostname;
+
+    if (typeof href === 'object') {
+        const linkHostname = href.hostname;
+        if (!linkHostname) {
+            return false;
+        }
+        return currentHostname !== linkHostname;
```
sebaholesz: Is the special condition necessary? If `linkHostname` is falsy, it will never match `currentHostname` as that should always be defined.

sebaholesz:
```suggestion
        return currentHostname !== href.hostname;
```

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -37,11 +40,11 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
...
 
-            if (type) {
-                updatePageLoadingState({ isPageLoading: true, redirectPageType: type });
-            } else {
-                updatePageLoadingState({ redirectPageType: undefined });
-            }
+            const isLinkExternal = isHrefExternal(href, baseUrl);
+            updatePageLoadingState({
+                isPageLoading: !!type || !isLinkExternal,
+                redirectPageType: isLinkExternal ? undefined : type ?? 'category',
```
sebaholesz: I would create a constant, something like 

```ts
const DEFAULT_PAGE_TYPE = 'category';
```

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
```diff
@@ -37,11 +40,11 @@ export const ExtendedNextLink: FC<ExtendedNextLinkProps> = ({
         if (isWithoutOpeningInNewTab) {
             onClick?.(e);
 
-            if (type) {
-                updatePageLoadingState({ isPageLoading: true, redirectPageType: type });
-            } else {
-                updatePageLoadingState({ redirectPageType: undefined });
-            }
+            const isLinkExternal = isHrefExternal(href, baseUrl);
```
sebaholesz: How does `isLinkExternal` relate to `isDynamic`. Are they correlated? Will they always be the opposite of each other?

sebaholesz (changes requested):

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
JanMolcik: **TL;DR** Relying just on the `type` or `isDynamic` doesn't cover all the navigation cases.

Examples explaining their difference:

1. As I have removed the type from the main navigation links - `isDynamic` is false for this case, the same as `isHrefExternal`, because we get the relative pathname, which results in an error and false from the function
2. Imagine someone in the administration putting an absolute URL of the category (eg. `https://ssfwcc.devel.shopsys.cloud/electronics`, which is a valid use case that Vašek Macíček shared with me) - in this case, we don't have a type but we can check that the hostnames match and for this case we can display skeletons.
3. If it is an external link I had an issue in the Arc browser that for the external link the page would open in a new small popup window on top of the current one and the page would be stuck in infinite skeleton loop (similar to the issue with Safari and opening link via mouse middle (wheel) button). In this case, `isDynamic` is false and the condition for `isPageLoading` will result in false too.

Number 3.) is the reason why we cannot simply put `true` to the `isPageLoading`. The `redirectPageType` could be simplified by removing the condition and provide the `DEFAULT_PAGE_TYPE` every time, but since the `isPageLoading` is false anyway it felt like it should be set to `undefined`.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
JanMolcik: Good idea, adjusted.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
JanMolcik: You are absolutely right.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
JanMolcik: OK I wasn't sure about the clarity of the code, but I also avoid using comments.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
JanMolcik: Because I thought just `url` might be confused as the`url` of the current link, but I guess that is `href`, so I will not rename it.

project-base/storefront/components/Basic/ExtendedNextLink/ExtendedNextLink.tsx
sebaholesz: Great explanation, thanks.

sebaholesz (approved):

PR #3293, opened 2024-07-30
[shopsys] datafixtures can be loaded only for some domains

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| If a project has to be run with many domains, it may be useful to load only some of them with demo data to speed up builds.
|New feature| Yes 
|Fixes issues| ... 

----
  - https://mg-datafixtures-only-domains.odin.shopsys.cloud
  - https://cz.mg-datafixtures-only-domains.odin.shopsys.cloud


project-base/app/src/DataFixtures/Demo/CartDataFixture.php
```diff
@@ -11,29 +11,34 @@
...
     ) {
     }
 
     /**
      * @param \Doctrine\Persistence\ObjectManager $manager
      */
-    public function load(ObjectManager $manager)
+    public function load(ObjectManager $manager): void
     {
+        $this->domain->switchDomainById($this->domainsForDataFixtureProvider->getFirstAllowedDomainConfig()->getId());
```
vitek-rostislav, line 40: why is this necessary, please? Would not it work the same without it? 
Or is it because of `CustomerUserDataFixture::CUSTOMER_PREFIX . '6'`?

project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
```diff
@@ -66,9 +66,9 @@ public function setUnit(ProductData $productData, string $unitReference): void
      */
     public function setFlags(ProductData $productData, array $flagReferences): void
     {
-        foreach ($this->domain->getAllIds() as $domainId) {
```
vitek-rostislav, line 69: do not we want to add `getAllAllowedDemoDataDomainIds` method to the provider?

vitek-rostislav (commented):
I like the fact it is a very universal solution that hypothetically allows enabling just the second domain for data fixtures, however, how would this (probably weird) setting affect tests? As the setting is committed and will be probably used to speed up CI builds, we should be careful with the connection between data fixtures and tests. It could also change the `domains-info.is-multidomain` phing property which is then used to determine which group of tests should be run. :grimacing:

project-base/app/src/DataFixtures/Demo/DataSetter/ProductDemoDataSetter.php
grossmannmartin: good idea, added and used in 6819fe5357

project-base/app/src/DataFixtures/Demo/CartDataFixture.php
grossmannmartin: The "problem" is with the called method `Shopsys\FrameworkBundle\Model\Cart\CartFacade::addProductToExistingCart`. This method uses `$this->domain->getId()`...

grossmannmartin:
> I like the fact it is a very universal solution that hypothetically allows enabling just the second domain for data fixtures, however, how would this (probably weird) setting affect tests? As the setting is committed and will be probably used to speed up CI builds, we should be careful with the connection between data fixtures and tests. It could also change the domains-info.is-multidomain phing property which is then used to determine which group of tests should be run. 😬

Yes, I agree this has to be used correctly, or the consequences may be unpleasant... we may force load of the first and second domain always, but is it worth it? If someone wants to do it, he does it anyway. The most common scenario is "test shop on the first domain", therefore I think the number of users, who disable loading data for the first domain is meager. And if, then they will know really soon that's not a way.

Maybe we could focus on the working shop with passing tests with the only 1st domain loaded (currently a few tests and datafixtures will fail).

project-base/app/src/DataFixtures/Demo/CartDataFixture.php
vitek-rostislav: ah, I see, thanks!

vitek-rostislav (approved):

vitek-rostislav (approved):

pk16011990 (approved):

PR #3304, opened 2024-08-02
[shopsys] performance improvement on product list/edit

grossmannmartin:
| Q                                                                                                              | A
|----------------------------------------------------------------------------------------------------------------| ---
| Description, the reason for the PR                                                                             | Minor performance improvements for product list and product edit page in administration. The performance impact is even more visible on e-shop with more domains.
| New feature                                                                                                    | No 
| [BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)                   | No 
| Fixes issues                                                                                                   | ... 
| Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)? | Yes

DB query count comparison on demo data with two domains

| Page         |  DB query count before | DB query count after |
|--------------|--------------|-------------|
| Product Grid | 138          | 107         |
| Product Hello kitty Edit | 196          | 138         |

----
  - https://mg-admin-performance.odin.shopsys.cloud
  - https://cz.mg-admin-performance.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Form/Admin/Product/Parameter/ProductParameterValueFormType.php
+++ b/packages/framework/src/Form/Admin/Product/Parameter/ProductParameterValueFormType.php
@@ -6,6 +6,7 @@ namespace Shopsys\FrameworkBundle\Form\Admin\Product\Parameter;
 
 use Shopsys\FrameworkBundle\Form\Locale\LocalizedType;
 use Shopsys\FrameworkBundle\Form\ValidationGroup;
+use Shopsys\FrameworkBundle\Model\Localization\Localization;
 use Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterFacade;
 use Shopsys\FrameworkBundle\Model\Product\Parameter\ProductParameterValuesLocalizedData;
 use Symfony\Component\Form\AbstractType;
@@ -21,9 +22,12 @@ class ProductParameterValueFormType extends AbstractType
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterFacade $parameterFacade
+     * @param \Shopsys\FrameworkBundle\Model\Localization\Localization $localization
      */
-    public function __construct(private readonly ParameterFacade $parameterFacade)
-    {
+    public function __construct(
+        private readonly ParameterFacade $parameterFacade,
+        private readonly Localization $localization,
+    ) {
     }
 
     /**
@@ -35,7 +39,7 @@ class ProductParameterValueFormType extends AbstractType
         $builder
             ->add('parameter', ChoiceType::class, [
                 'required' => true,
-                'choices' => $this->parameterFacade->getAll(),
+                'choices' => $this->parameterFacade->getAllWithTranslations($this->localization->getAdminLocale()),
                 'choice_label' => 'name',
                 'choice_value' => 'id',
                 'constraints' => [
--- a/packages/framework/src/Model/Product/Parameter/ParameterRepository.php
+++ b/packages/framework/src/Model/Product/Parameter/ParameterRepository.php
@@ -116,13 +116,34 @@ class ParameterRepository
      */
     public function getAll(): array
     {
-        return $this->em->createQueryBuilder()
-            ->select('p')
-            ->from(Parameter::class, 'p')
-            ->join('p.translations', 'pt')
+        return $this->getAllQueryBuilder()
             ->orderBy('p.orderingPriority', 'DESC')
             ->getQuery()
             ->execute();
+    }
+
+    /**
+     * @param string $locale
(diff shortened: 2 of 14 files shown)


packages/framework/src/Model/Product/ProductVisibilityRepository.php
```diff
@@ -107,6 +107,29 @@ public function findProductVisibilitiesByDomainIdAndProduct($domainId, Product $
...
+    public function getCountOfDomainsProductIsVisibleOn(
+        Product $product,
+        array $defaultPricingGroupIdsIndexedByDomainId,
+    ): int {
+        $queryBuilder = $this->em->createQueryBuilder()
+            ->select('count(pv.domainId)')
+            ->from(ProductVisibility::class, 'pv')
+            ->where('pv.product = :product')
+            ->andWhere('pv.pricingGroup IN (:pricingGroupIds)')
+            ->andWhere('pv.domainId IN (:domainIds)')
```
vitek-rostislav: I wanted to point out that this condition seems to be unnecessary as pricing group is already bound to a domain, however, that could open a discussion about `ProductVisibility::domainId` as a whole so I am not sure :smile:

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -125,6 +125,22 @@ public function getAll(): array
...
+    {
+        $this->em->createQueryBuilder()
+            ->select('p', 'pt')
+            ->from(Parameter::class, 'p')
+            ->join('p.translations', 'pt')
+            ->orderBy('p.id', 'asc')
+            ->getQuery()
+            ->execute();
+
+        return $this->getParameterRepository()->findBy([], ['id' => 'asc']);
```
vitek-rostislav: wow, this is an interesting approach. Executing the query loads the translations so they are then loaded in the following query? Do I understand it correctly? Is not it too magical? :smile: personally, I would prefer something like this - https://github.com/shopsys/shopsys/blob/15.0/packages/framework/src/Model/Payment/PaymentRepository.php#L154

grossmannmartin: 🙈  this is probably a mistake... I have changed it in bfd03e5c14

packages/framework/src/Model/Product/ProductVisibilityRepository.php
grossmannmartin: I have removed the domainId condition.
I kept the domainId as a keys in the array returned from `\Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroupSettingFacade::getAllDefaultPricingGroupsIdsIndexedByDomainId`
It seems more usable this way
d486350321

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -130,15 +127,22 @@ public function getAll(): array
...
     public function getAllWithTranslations(): array
     {
-        $this->em->createQueryBuilder()
-            ->select('p', 'pt')
-            ->from(Parameter::class, 'p')
+        return $this->getAllQueryBuilder()
+            ->addSelect('pt')
             ->join('p.translations', 'pt')
-            ->orderBy('p.id', 'asc')
+            ->orderBy('pt.name', 'asc')
```
vitek-rostislav: we should use `OrderByCollationHelper` here, right? :innocent:

vitek-rostislav (approved):
one more thing :slightly_smiling_face:

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
grossmannmartin: exactly


fixed in 77f3bfccb5

packages/framework/src/Form/Admin/Product/Parameter/ProductParameterValueFormType.php
```diff
@@ -21,9 +22,12 @@ class ProductParameterValueFormType extends AbstractType
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterFacade $parameterFacade
+     * @param \Shopsys\FrameworkBundle\Model\Localization\Localization $localization
      */
-    public function __construct(private readonly ParameterFacade $parameterFacade)
-    {
+    public function __construct(
+        private readonly ParameterFacade $parameterFacade,
+        private readonly Localization $localization,
```
pk16011990, line 29: Missing an upgrade notes?

grossmannmartin: the FormTypes are not meant to be extended via inheritance

PR #3308, opened 2024-08-05
[shopsys] GoPay configuration is based on domains instead of locale

grossmannmartin:

The GoPay payment methods configuration has been updated based on domains rather than locales.
This change enables different GoPay settings for each domain, addressing issues where multiple domains share the same currency but have different locales.
The GoPay library has been updated to its latest version, ensuring compatibility and incorporating the newest features and fixes from the GoPay API.
The configuration for GoPay methods has been consolidated into a single JSON format.
This streamlines the configuration process and facilitates easier integration and continuous integration (CI) setups.

#### Fixes issues
... 

- [ ] New feature 

## Summary by CodeRabbit

- **New Features**
	- Consolidated GoPay configuration into a single environment variable for improved management.
	- Introduced domain-specific handling for GoPay payment methods across multiple domains.
	- Added new validation logic for GoPay payment methods in forms.
	- Enhanced GraphQL API with a dedicated resolver for GoPay payment methods.

- **Bug Fixes**
	- Improved error handling and logging for payment processing related to GoPay.

- **Documentation**
	- Updated translation files for better clarity and context in user messages.

- **Chores**
	- Removed outdated individual GoPay environment variables from configuration files.

----
  - https://mg-gopay-by-domains.odin.shopsys.cloud
  - https://cz.mg-gopay-by-domains.odin.shopsys.cloud


packages/framework/src/Model/GoPay/GoPayClient.php
```diff
@@ -18,7 +18,7 @@
 
 class GoPayClient
 {
-    public const RESPONSE_STATUS_CODE_OK = 200;
+    public const int RESPONSE_STATUS_CODE_OK = 200;
```
malyMiso: I wonder if it is a good practice to set type for a constant that we use for comparing with data from 3rd party libraries. 
We won't be able to override it in projects. One day GoPay will decide to use string as response code and projects will have to create a new constant.  

I believe it is just an edge case, I just wonder what bets practice should be, when it comes to 3rd party libaries

+ I don't see it used anywhere else, so we can change visibility to `protected`

packages/framework/src/Model/GoPay/GoPayClient.php
```diff
@@ -98,7 +98,7 @@ public function getStatus(string $id): Response
 
         $response = $this->sendApiRequest($urlPath, GoPay::FORM, RequestMethods::GET);
 
-        if ((int)$response->statusCode !== (int)self::RESPONSE_STATUS_CODE_OK) {
+        if ((int)$response->statusCode !== self::RESPONSE_STATUS_CODE_OK) {
```
malyMiso, line 101: :question: 

1. As I see correctly `response.statusCode` should be the type of int. Is it type casting still needed ? 
2. should't it be `static::RESPONSE_STATUS_CODE_OK`

packages/framework/src/Model/GoPay/GoPayClientFactory.php
```diff
@@ -4,43 +4,102 @@
...
 
 class GoPayClientFactory
 {
+    protected const string PRODUCTION_URL = 'https://gate.gopay.cz/';
+    protected const string TEST_URL = 'https://gw.sandbox.gopay.com/';
+
+    /**
+     * @var array<int, \Shopsys\FrameworkBundle\Model\GoPay\GoPayClient>
+     */
+    protected array $clientByDomainId = [];
```
malyMiso: :thought_balloon: Nitpick

Maybe `clientsByDomainId` would be more accurate

packages/framework/src/Model/GoPay/GoPayClientFactory.php
```diff
@@ -4,43 +4,102 @@
...
+
     /**
      * @param array $config
      */
-    public function __construct(protected array $config)
+    public function __construct(
+        protected array $config,
```
malyMiso: 💭 Nitpick

Maybe `clients` would be more accurate

packages/framework/src/Model/GoPay/GoPayClientFactory.php
```diff
@@ -4,43 +4,102 @@
...
+                $clientConfig['clientId'],
+                $clientConfig['clientSecret'],
+                $clientConfig['isProductionMode'] ? static::PRODUCTION_URL : static::TEST_URL,
+                $domainConfig->getLocale(),
+            );
         }
 
-        $configByLocale = $this->config[$locale];
-        $this->config = array_merge($this->config, $configByLocale);
+        throw new GoPayNotEnabledOnDomainException();
```
malyMiso: Some exception message would help with faster debugging

e.g.
Gopay is not enabaled on domain `666`

packages/framework/src/Model/Payment/Service/PaymentServiceFacade.php
```diff
@@ -102,7 +103,7 @@ public function updatePaymentTransactionsByOrder(Order $order): void
                 if ($update) {
                     $this->paymentTransactionFacade->edit($paymentTransaction->getId(), $paymentTransactionData);
                 }
-            } catch (PaymentServiceFacadeNotRegisteredException|GoPayNotConfiguredException|PaymentTransactionHasNoAssignedPayment $exception) {
+            } catch (PaymentServiceFacadeNotRegisteredException|GoPayNotConfiguredException|GoPayNotEnabledOnDomainException|PaymentTransactionHasNoAssignedPayment $exception) {
                 $this->logger->error($exception->getMessage());
```
malyMiso: :bulb: 

It is not your code. I've noticed logging error without setting context. From my point of view, it is good practice to pass exception into context, it will pass stack trace to Sentry on exception

Source: https://github.com/getsentry/sentry-php/blob/master/src/Monolog/Handler.php#L63

project-base/app/.env.test
```diff
@@ -5,10 +5,15 @@ DATABASE_NAME=shopsys-test
...
-GOPAY_EN_CLIENTSECRET=test-data
-GOPAY_CS_GOID=test-data
-GOPAY_CS_CLIENTID=test-data
-GOPAY_CS_CLIENTSECRET=test-data
+GOPAY_CONFIG='[
+    {
+        "goid": "test-data",
+        "clientId": "test-data",
+        "clientSecret": "test-data",
+        "isProductionMode": false,
```
malyMiso, line 13: :question: 

What about setting one domain to production and second domain to testing mode ? Will be there such a scenario ?

packages/framework/src/Model/GoPay/GoPayClientFactory.php
malyMiso: :information_source: 
I saw few more constants with singular name, that are typed as array. I ignored it, because I was not sure, if it is on purpose

malyMiso (changes requested):

packages/framework/src/Model/GoPay/GoPayClient.php
grossmannmartin: I have changed the visibility to protected 👍 94f8535b4c

I think it's ok to type-hint these types of constants as we compare it in our code, where the response is typed to integer.

packages/framework/src/Model/GoPay/GoPayClient.php
grossmannmartin: 1. the type is sadly only annotated to int... in reality the status code may be string - for example in `\GoPay\Http\JsonBrowser::send()`
 
Snímek obrazovky 2024-08-13 v 13 59 21

2. now, when the `RESPONSE_STATUS_CODE_OK` is protected, the call was changed to `static::`

packages/framework/src/Model/GoPay/GoPayClientFactory.php
grossmannmartin: changed in 9657622c83

I think the others are fine, if you have any examples, please let me know

packages/framework/src/Model/GoPay/GoPayClientFactory.php
grossmannmartin: I'm not sure... the value is a JSON value with complete gopay config

packages/framework/src/Model/GoPay/GoPayClientFactory.php
grossmannmartin: added in 66a8e30786

packages/framework/src/Model/Payment/Service/PaymentServiceFacade.php
grossmannmartin: nice, thanks :)  added in 88eae2e5fc

project-base/app/.env.test
grossmannmartin: I believe so... for example when the first domain is running and the second will is still in development :) 

when the GoPay configuration data should be the same for the first and second domain it's a little cumbersome, but still doable

```
GOPAY_CONFIG='[
    {
        "goid": "<goid1>",
        "clientId": "<clientId1>",
        "clientSecret": "<clientSecret1>",
        "isProductionMode": true,
        "domains": [
            1
        ]
    },
    {
        "goid": "<goid1>",
        "clientId": "<clientId1>",
        "clientSecret": "<clientSecret1>",
        "isProductionMode": false,
        "domains": [
            2,3
        ]
    }
]'
```

malyMiso (approved):

packages/framework/src/Model/GoPay/PaymentMethod/GoPayPaymentMethodFacade.php
```diff
@@ -131,6 +131,7 @@ public function setFromGoPayRawData(
         $goPayPaymentMethodData->imageNormalUrl = $goPayMethodRawData['image']['normal'];
         $goPayPaymentMethodData->imageLargeUrl = $goPayMethodRawData['image']['large'];
         $goPayPaymentMethodData->paymentGroup = $goPayMethodRawData['group'];
+        $goPayPaymentMethodData->available = true;
```
vitek-rostislav, line 137: is this correct? I expected something like `$goPayMethodRawData['available'];`

vitek-rostislav, line 137: is this correct? I expected something like `$goPayMethodRawData['available'];`

project-base/app/src/DataFixtures/Demo/GoPayDataFixture.php
```diff
@@ -166,6 +166,7 @@ private function getGoPayPaymentMethodData(
         $goPayPaymentMethodData->imageLargeUrl = 'https://gate.gopay.cz/images/checkout/' . $data['image_large_url'] . '.png';
         $goPayPaymentMethodData->paymentGroup = $data['payment_group'];
         $goPayPaymentMethodData->domainId = $domainId;
+        $goPayPaymentMethodData->available = true;
```
vitek-rostislav: nitpick, it maybe we want an unavailable payment method in the data fixtures as well? I do not insist on that

vitek-rostislav (changes requested):
translations needs to be dumped :wink:

packages/framework/src/Model/GoPay/PaymentMethod/GoPayPaymentMethodFacade.php
grossmannmartin: The thing is, the GoPay API returns only the available payments, so anything returned is available

grossmannmartin: The thing is, the GoPay API returns only the available payments, so anything returned is available

project-base/app/src/DataFixtures/Demo/GoPayDataFixture.php
grossmannmartin: good idea, added

packages/framework/src/Model/GoPay/PaymentMethod/GoPayPaymentMethodFacade.php
vitek-rostislav: ah, ok, I see :+1:

vitek-rostislav (approved):

PR #3317, opened 2024-08-07
[framework] admin parameter form now validates parameter name uniqueness within a given locale

vitek-rostislav:
Having multiple parameters with the same name is quite confusing and it breaks Luigi's Box functionality as it requires the parameter names to be unique.

#### Fixes issues
... 

- [ ] New feature 

----
  - https://rv-parameter-unique-name.odin.shopsys.cloud
  - https://cz.rv-parameter-unique-name.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Form/Admin/Product/Parameter/ParameterFormType.php
+++ b/packages/framework/src/Form/Admin/Product/Parameter/ParameterFormType.php
@@ -4,9 +4,12 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Form\Admin\Product\Parameter;
 
+use Shopsys\FrameworkBundle\Component\Translation\Translator;
 use Shopsys\FrameworkBundle\Form\Locale\LocalizedType;
+use Shopsys\FrameworkBundle\Model\Localization\Localization;
 use Shopsys\FrameworkBundle\Model\Product\Parameter\Parameter;
 use Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterData;
+use Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterFacade;
 use Shopsys\FrameworkBundle\Model\Product\Unit\UnitFacade;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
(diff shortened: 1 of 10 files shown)


packages/framework/src/Resources/translations/validators.cs.po
```diff
@@ -145,6 +145,9 @@ msgstr "Hodnota parametru musí být číselná"
 msgid "Parameter value must be of type {{ type }}"
 msgstr "Hodnota parametru musí být typu {{ type }}"
 
+msgid "Parameter with this name (%locale%) already exists."
```
grossmannmartin: _nitpick_ 
what about 

Parameter with this name already exists for the locale "%locale%".

project-base/app/assets/js/admin/validation/parameterNameValidator.js
```diff
@@ -0,0 +1,14 @@
+import 'framework/common/components';
+import Register from 'framework/common/utils/Register';
+
+function parameterNameValidator ($container) {
```
grossmannmartin: shouldn't it be in the framework package?

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -548,4 +548,23 @@ public function getProductsByParameterValues(array $parameterValues): array
...
+        return $this->em->createQueryBuilder()
+            ->select('p')
+            ->from(Parameter::class, 'p')
+            ->join('p.translations', 'pt')
+            ->where('pt.name = :name')
+            ->andWhere('pt.locale = :locale')
+            ->setParameter('name', $name)
+            ->setParameter('locale', $locale)
+            ->getQuery()
+            ->getOneOrNullResult();
```
grossmannmartin: if existing data already contains the duplicities, this validator will result in the error 500, or?

packages/framework/src/Form/Admin/Product/Parameter/ParameterFormType.php
```diff
@@ -78,8 +86,39 @@ public function configureOptions(OptionsResolver $resolver): void
...
+
+        /** @var \Shopsys\FrameworkBundle\Model\Product\Parameter\Parameter|null $currentParameter */
+        $currentParameter = $formOptions['parameter'];
+
+        foreach ($parameterData->name as $locale => $name) {
+            if ($name === null) {
+                continue;
+            }
+
+            $parameter = $this->parameterFacade->findParameterByName($name, $locale);
```
grossmannmartin: _nitpick_ maybe we could load all parameter names at once before to avoid multiple same queries?

grossmannmartin (changes requested):

packages/framework/src/Resources/translations/validators.cs.po
vitek-rostislav: makes sense :+1: 0475b5c280

project-base/app/assets/js/admin/validation/parameterNameValidator.js
vitek-rostislav: sure it should, thank you, fixed in 338ebb0394

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
vitek-rostislav: good catch, fixed in f9a776139d

packages/framework/src/Form/Admin/Product/Parameter/ParameterFormType.php
vitek-rostislav: after the refactoring in f9a776139d it would be too complicated

grossmannmartin (approved):

grossmannmartin (approved):

PR #3349, opened 2024-08-15
[project-base] limiting image size options propagated for resizing in CDN to predefined set of options

AndrejBlaho:
This pull request introduces a limitation on the ImageResizer in shopsys/project-base/app/web/imageResizer.php to prevent potential abuse by restricting the range of allowable image dimensions. The changes ensure that only predefined dimensions are accepted, reducing the risk of CDN overload from generating numerous image sizes via malicious requests.

Previously, the ImageResizer allowed for arbitrary dimensions, which could be exploited by attackers to flood the CDN with a large number of image resize requests. This PR restricts the allowed dimensions to a predefined set (e.g., 30, 50, 200, 350, 500, 1000). When a request for an unsupported size is received, the nearest larger allowed size will be returned, or the maximum allowed size if the requested size exceeds it. This enhancement safeguards the system by minimizing unnecessary CDN load and ensuring better performance and security.

#### Fixes issues
... 

- [ ] New feature 

----
  - https://ab-limited-image-sizes-in-cdn.odin.shopsys.cloud
  - https://cz.ab-limited-image-sizes-in-cdn.odin.shopsys.cloud


project-base/app/web/imageResizer.php
```diff
@@ -21,11 +21,25 @@
...
-$width = $_GET['width'] ?? 0;
-$height = $_GET['height'] ?? 0;
+$width = (int)($_GET['width'] ?? 0);
+$height = (int)($_GET['height'] ?? 0);
 $gravity = 'no';
 $enlarge = 1;
 
+$allowedDimensions = [1000, 500, 350, 200, 50, 30];
+
+function getClosestAllowedDimension(int $requestedSize, array $allowedSizes): int {
```
grossmannmartin: this function poses two problems: 
1. for every number lower than 1000, it returns 1000
2. it's dependent on the order of elements in the array

Both problems may be solved by sorting the $allowedDimensions array first (from lowest number to highest).

grossmannmartin: can you put the function at the end of the file to the rest of the functions?

project-base/app/web/imageResizer.php
```diff
@@ -21,11 +21,25 @@
...
 
 $resize = $_GET['resize'] ?? 'fit';
-$width = $_GET['width'] ?? 0;
-$height = $_GET['height'] ?? 0;
+$width = (int)($_GET['width'] ?? 0);
+$height = (int)($_GET['height'] ?? 0);
 $gravity = 'no';
 $enlarge = 1;
 
+$allowedDimensions = [1000, 500, 350, 200, 50, 30];
```
grossmannmartin: This array could be higher as it's one of a few things that the project may modify.

Also. I would unify the variable names in function and in array declaration

project-base/app/web/imageResizer.php
```diff
@@ -21,11 +21,25 @@
...
+function getClosestAllowedDimension(int $requestedSize, array $allowedSizes): int {
+    foreach ($allowedSizes as $size) {
+        if ($requestedSize <= $size) {
+            return $size;
+        }
+    }
+    return $allowedSizes[0];
+}
+
+$width = getClosestAllowedDimension($width, $allowedDimensions);
```
grossmannmartin: I would merge these lines with the initialization, or move them closer together, so it's clear what is happening.

```php
$resize = $_GET['resize'] ?? 'fit';
$width = getClosestAllowedDimension((int)($_GET['width'] ?? 0), $allowedDimensions);
$height = getClosestAllowedDimension((int)($_GET['height'] ?? 0), $allowedDimensions);
$gravity = 'no';
$enlarge = 1;
```

or 

```php
$resize = $_GET['resize'] ?? 'fit';
$width = (int)($_GET['width'] ?? 0);
$width = getClosestAllowedDimension($width, $allowedDimensions);
$height = (int)($_GET['height'] ?? 0);
$height = getClosestAllowedDimension($height, $allowedDimensions);
$gravity = 'no';
$enlarge = 1;
```

grossmannmartin (changes requested):

project-base/app/web/imageResizer.php
AndrejBlaho: sure

AndrejBlaho: 1. you are right, now it should be ok
2. is it ok if we will leave it as it is or should i put there sorting function?

project-base/app/web/imageResizer.php
AndrejBlaho: done both

project-base/app/web/imageResizer.php
AndrejBlaho: done

project-base/app/web/imageResizer.php
```diff
@@ -82,3 +84,12 @@ function getExtension(string $url): string
...
 }
+
+function getClosestAllowedImageSize(int $requestedImageSize, array $allowedImageSizes): int {
+    foreach ($allowedImageSizes as $size) {
+        if ($requestedImageSize <= $size) {
+            return $size;
+        }
+    }
+    return $allowedImageSizes[0];
+}
```
grossmannmartin: please end file with line break

project-base/app/web/imageResizer.php
grossmannmartin: 2. I would add a sorting function - to prevent bad usage

grossmannmartin (approved):

PR #3365, opened 2024-08-22
[shopsys] administrator now can see list of last ten orders on customer edit page

TomasLudvik:
Better administration of customer orders.

- [x] New feature 

----
  - https://tl-customer-orders-admin.odin.shopsys.cloud
  - https://cz.tl-customer-orders-admin.odin.shopsys.cloud


packages/framework/src/Model/AdvancedSearchOrder/Filter/OrderCustomerIdFilter.php
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getAllowedOperators(): array
+    {
+        return [
+            self::OPERATOR_IS,
+            self::OPERATOR_NOT_SET,
```
malyMiso: :thinking: 

Translation of this operator may be a little misleading ? It says "not entered", will administrator know what does it mean or filter ? 
I believe "not registered customer" would fit better. I know it is general translation, but needed to point it out

packages/framework/src/Model/AdvancedSearchOrder/Filter/OrderCustomerIdFilter.php
```diff
@@ -0,0 +1,94 @@
...
+                $queryBuilder->andWhere('o.customer IS NULL');
+
+                continue;
+            }
+
+            try {
+                $customer = $this->customerFacade->getById((int)$ruleData->value);
+
+                $parameterName = 'customer_' . $index;
+                $queryBuilder->andWhere('o.customer = :' . $parameterName);
```
malyMiso: :question: 

When filtering orders by multiple customers (ID 1, ID 2) it will not find any order. Wouldn't be better to collect all IDs and use `IN` clause (find all orders with ID 1 or ID 2, or ...). I do not know what is preferred filter, can you please check it ?

malyMiso (changes requested):

packages/framework/assets/js/admin/components/AdvancedSearch.js
```diff
@@ -70,7 +70,7 @@ export default class AdvancedSearch {
     }
 
     static updateValueByOperator ($rulesContainer, $rule, operator) {
-        $rule.find('.js-advanced-search-rule-value').toggle(operator !== 'notSet');
+        $rule.find('.js-advanced-search-rule-value').toggle(operator !== 'notSet' && operator !== 'notRegistered');
```
malyMiso, line 73: Just a note, pls do not fix it. 

It would be nice to configure this on PHP level, so I don't need to override JS on project every time I add such an operator

malyMiso: The  operator `notSet` has many times same functionality, but translation can differ.

malyMiso (approved):

PR #3374, opened 2024-08-27
[framework] EntityLogEventListener now handles PersistentCollection directly instead of catching Exception

stanoMilan:
Problem: I enabled logging on the Product entity, then made a change through the administration, for example in CATNUM, but I didn't see any changes in the entity log.

Solution: The issue was that the method ChangeSetResolver::getResolvedChanges received a PersistentCollection object instead of an array, which is what this method expects when the collection is modified.

I'm not sure if it's the correct solution, but catching Throwable doesn't seem like the right approach to me.

### DIFF
--- a/packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
+++ b/packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
@@ -9,6 +9,7 @@ use Doctrine\ORM\Event\PostFlushEventArgs;
 use Doctrine\ORM\Event\PostPersistEventArgs;
 use Doctrine\ORM\Event\PostUpdateEventArgs;
 use Doctrine\ORM\Event\PreRemoveEventArgs;
+use Doctrine\ORM\PersistentCollection;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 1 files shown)


TomasLudvik (approved):

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
```diff
@@ -99,15 +96,11 @@ protected function log(string $action, object $entity): void
     {
         $loggableSetup = $this->loggableEntityConfigFactory->getLoggableSetupByEntity($entity);
 
-        try {
-            if (!$loggableSetup->isLoggable()) {
-                return;
-            }
-
-            $this->registerLog($entity, $loggableSetup, $action);
-        } catch (Throwable $exception) {
```
vitek-rostislav: I would rather keep this catch here. I do not want the application to crash when there is an error in entity logging - I prefer the way it was - it will silently add an entry into the log.

vitek-rostislav (changes requested):

vitek-rostislav (approved):

PR #3397, opened 2024-08-31
[shopsys] transport restrictions

vitek-rostislav:
Admin needs to set multiple price levels for transports, based on the total weight of products in cart so the pricing for the end user is reasonable. Also, the admin needs to exclude particular transports for particular products so the end user is not able to use inappropriate transport for his order (e.g. due to the dimensions of the products in cart).

#### Fixes issues
... 

- [x] New feature 

## Summary by CodeRabbit

- **New Features**
	- Introduced new transport management capabilities, allowing users to exclude specific transport options for products.
	- Added a new field for maximum weight in transport pricing, enhancing transport price management.
	- Implemented transport availability checks based on products in the cart, improving user experience during checkout.
	- Enhanced the transport pricing form with VAT considerations and weight limits for better pricing configurations.
	- Added new translation entries for improved localization support in Czech and English.
	- Introduced a structured form layout for managing transport prices, improving user interface interactivity.
	- Enhanced data integrity with unique indexing on transport prices based on weight limits and domain.
	- Enhanced transport price retrieval logic to accommodate multiple prices per domain, improving pricing flexibility.
	- Streamlined transport pricing calculations by integrating weight limits and VAT management.
	- Added a comprehensive set of Twig templates for managing transport prices with weight limits and VAT.
	- Implemented form validation for transport-related forms to ensure data integrity.

- **Bug Fixes**
	- Enhanced error handling for unavailable transport options, ensuring users receive clear feedback.

- **Tests**
	- Expanded test coverage for transport validation and cart modifications, ensuring robustness in handling transport options.
	- Updated tests to reflect new transport pricing structures and VAT handling.

----
  - https://rv-transport-prices.odin.shopsys.cloud
  - https://cz.rv-transport-prices.odin.shopsys.cloud


packages/framework/src/Migrations/Version20240831121309.php
```diff
@@ -0,0 +1,43 @@
...
+
+class Version20240831121309 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->sql('
+            CREATE TABLE products_excluded_transports (
```
grossmannmartin: nitpick:

usually we have the word `product` in singular by those type of tables

packages/framework/src/Model/Transport/TransportVisibilityCalculation.php
```diff
@@ -70,4 +73,21 @@ public function filterVisible(array $transports, array $visiblePaymentsOnDomain,
...
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Transport\Transport[] $transports
+     * @param \Shopsys\FrameworkBundle\Model\Cart\Cart $cart
+     * @return \Shopsys\FrameworkBundle\Model\Transport\Transport[]
+     */
+    public function filterTransportsByProductsInCart(array $transports, Cart $cart): array
+    {
+        $excludedTransports = array_merge(
```
grossmannmartin: the `$excludedTransports` array will contain a transport multiple times when the transport is present by multiple products. 
we could improve this I believe

packages/framework/src/Model/Transport/TransportVisibilityCalculation.php
```diff
@@ -70,4 +73,21 @@ public function filterVisible(array $transports, array $visiblePaymentsOnDomain,
...
+            ...array_map(static fn (Product $product) => $product->getExcludedTransports(), $cart->getProducts()),
```
grossmannmartin: This loading of excluded transports in a loop is n+1 problem-prone.
we may basically achieve the desired behavior with a simple select
`Select distinct transport_id from products_excluded_transports where product_id in (...)`

then we can just drop transports with ids from returned query

grossmannmartin (changes requested):

packages/framework/src/Migrations/Version20240831121309.php
vitek-rostislav: true, fixed in 7300d93091

vitek-rostislav: true, fixed in 7300d93091

vitek-rostislav: ...and in a7184e6dda

packages/framework/src/Model/Transport/TransportVisibilityCalculation.php
vitek-rostislav: so what about this? c03ca6f2de

I was kind of struggling with the right placement of the query as there is no entity `ProductExcludedTransport` hence was not convinced about creating the `ProductExcludedTransportRepository`  (and facade?) classes. Another idea was adding the method into `ProductRepository` (and facade? :smile: ), I could even use query builder but I would have to use  `SELECT FROM Product::class p JOIN p.excludedTransports ` which seemed overcomplicated so in the end, I just post the native query directly here where I needed it. I am open to discussion of course :slightly_smiling_face:

packages/framework/src/Form/Admin/Transport/Price/PriceWithWeightLimitType.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    #[Override]
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('price', MoneyType::class, [
+                'scale' => 6,
+            ])
+            ->add('maxWeight', IntegerType::class)
+            ->add('transportPriceId', HiddenType::class);
```
grossmannmartin: is it necessary to pass it as a hidden field? is it not possible to pass it as an option?

packages/framework/src/Form/Admin/Transport/TransportFormType.php
```diff
@@ -129,10 +133,19 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
-        $builderPricesGroup->add('pricesByDomains', PriceAndVatTableByDomainsType::class, [
-            'pricesIndexedByDomainId' => $this->transportFacade->getPricesIndexedByDomainId($transport),
-            'inherit_data' => true,
-            'render_form_row' => false,
+        $optionsByDomainId = [];
+
+        foreach ($this->domain->getAllIds() as $domainId) {
+            $optionsByDomainId[$domainId] = [
+                'domain_id' => $domainId,
+                'current_transport_prices_indexed_by_id' => $transport instanceof Transport ? $this->transportFacade->getPricesIndexedByTransportPriceId($transport) : [],
```
grossmannmartin: does this return a different results based on the domain id?
from the method annotation it does not seems so

packages/framework/src/Model/Transport/Transport.php
```diff
@@ -226,43 +225,19 @@ public function getPrices()
...
-        return false;
+        return array_column($this->prices->toArray(), null, 'id');
     }
 
     /**
-     * @param \Shopsys\FrameworkBundle\Model\Transport\TransportPrice $transportPrice
+     * @param \Doctrine\Common\Collections\Collection<int, \Shopsys\FrameworkBundle\Model\Transport\TransportPrice> $pricesCollection
      */
-    public function addPrice(TransportPrice $transportPrice): void
+    public function setPrices($pricesCollection)
```
grossmannmartin: should the setter on the entity accept the collection? 
Shouldn't it be isolated?

packages/framework/src/Model/Transport/TransportPriceRepository.php
```diff
@@ -0,0 +1,53 @@
...
+        $transportPrice = $this->em->createQueryBuilder()
+            ->select('tp')
+            ->from(TransportPrice::class, 'tp')
+            ->where('tp.transport = :transport')
+            ->andWhere('tp.domainId = :domainId')
+            ->andWhere('((tp.maxWeight >= :cartTotalWeight) OR (tp.maxWeight IS NULL))')
+            ->setParameter('transport', $transport)
+            ->setParameter('domainId', $domainId)
+            ->setParameter('cartTotalWeight', $cartTotalWeight)
+            ->orderBy('tp.maxWeight', 'ASC')
```
grossmannmartin, line 40: how are the nulls sorted?

packages/framework/src/Model/Transport/TransportRepository.php
```diff
@@ -170,7 +170,8 @@ public function getAllWithEagerLoadedDomainsAndTranslations(
 
         if ($totalWeight !== null) {
             $queryBuilder
-                ->andWhere('t.maxWeight IS NULL OR t.maxWeight >= :maxWeight')
+                ->join('t.prices', 'tp')
+                ->andWhere('tp.maxWeight IS NULL OR tp.maxWeight >= :maxWeight')
```
grossmannmartin, line 174: what if we have weights that meets both conditions?

packages/framework/src/Resources/translations/messages.en.po
```diff
@@ -3958,7 +3955,7 @@ msgstr ""
 msgid "Weight (g)"
 msgstr ""
 
-msgid "Weight limit"
+msgid "Weight limit (optional)"
```
grossmannmartin: Will it be clear enough what this field means? 
It's maximum weight to apply this price, right? Will it be understandable for admin?

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Transport/Price/PriceWithWeightLimitType.php
vitek-rostislav: Let's discuss it, I do not see any easy way how to do that - when defining the collection in the parent `TransportPricesWithWeightLimitsAndVatSelectType::buildForm()`, I do not have access to the data. Doing it this way was convenient as the transport price ID is a part of the form's data object.

packages/framework/src/Form/Admin/Transport/TransportFormType.php
vitek-rostislav: No no, it returns the same array of all transports regardless of their domain. Basically in each of the multidomain form entries, I am interested in "half" of the array. I tried rewriting it in 2217b3d07e so each `TransportPricesWithWeightLimitsAndVatSelectType` gets only the prices on the relevant domain, however, I do not think it is somehow better, it is just another additional logic that actually does not improve anything. Maybe a better approach would be  calling `$this->transportFacade->getPricesIndexedByTransportPriceId($transport)` before the domain ids are iterated to avoid multiple calls, and just pass the same array to the all domain entries, what do you think?

packages/framework/src/Model/Transport/TransportPriceRepository.php
vitek-rostislav: NULLs are the last which is convenient as it stands for "unlimited". If I wanted the Nulls to be first, I would have to use our custom `SortableNullsWalker` but this is not the case

packages/framework/src/Model/Transport/Transport.php
vitek-rostislav: what are the cons of this approach? let's discuss it, please, I have no problem with rewriting it to accept an array, however, it was quite convenient to do it this way here

packages/framework/src/Model/Transport/TransportRepository.php
vitek-rostislav: hm, interesting question, do you think the entry would be duplicated in the result? The `TransportPrice` is not included in the `SELECT` so IMHO it should be fine, or?

packages/framework/src/Resources/translations/messages.en.po
vitek-rostislav: what about this, better? :slightly_smiling_face: 3dbf32260b

packages/framework/src/Form/Admin/Transport/Price/PricesWithWeightLimitsType.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Form\Admin\Transport\Price;
+
+use Override;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+
+class PricesWithWeightLimitsType extends AbstractType
```
grossmannmartin: this class seems unused, is it on purpose?

packages/framework/src/Form/Admin/Transport/TransportFormType.php
grossmannmartin: Call it once before seems like a reasonable idea :)

packages/framework/src/Model/Transport/Transport.php
grossmannmartin: It exposes the collection that should be encapsulated inside the entity. Let's discuss it please

packages/framework/src/Model/Transport/TransportRepository.php
grossmannmartin: oh, I did not notice the tp was not selected. then it's ok :)

packages/framework/src/Resources/translations/messages.en.po
grossmannmartin: verynicebeautiful 👍

packages/framework/src/Model/Transport/PriceWithWeightLimitData.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Transport;
+
+class PriceWithWeightLimitData
+{
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\Money\Money|null
+     */
+    public $price;
```
grossmannmartin: I think this field should not be named price as it's just an amount with VAT.
Similarly maybe the whole class?

packages/framework/src/Model/Transport/TransportFacade.php
```diff
@@ -122,22 +121,22 @@ public function getVisibleByDomainId($domainId, $visiblePaymentsOnDomain)
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\Transport\Transport $transport
-     * @param \Shopsys\FrameworkBundle\Component\Money\Money[] $pricesIndexedByDomainId
+     * @param \Shopsys\FrameworkBundle\Model\Transport\TransportPricesWithVatData[] $pricesIndexedByDomainId
      */
     protected function updateTransportPrices(Transport $transport, array $pricesIndexedByDomainId): void
```
grossmannmartin: I was a little confused by the naming... I expected something different from `$pricesIndexedByDomainId`  than a type of `TransportPricesWithVatData[]` (I expected `array<int, Price>`

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Transport/Price/PricesWithWeightLimitsType.php
vitek-rostislav: yeah, it is a left-over from initial thoughts, removed in 1b3a9789e4

packages/framework/src/Form/Admin/Transport/TransportFormType.php
vitek-rostislav: fixed in 7b586f2500

packages/framework/src/Model/Transport/PriceWithWeightLimitData.php
vitek-rostislav: after discussion we kept it as is (the class was renamed to `PriceWithLimitData`)

packages/framework/src/Model/Transport/TransportFacade.php
vitek-rostislav: I refactored the naming after our discussion, however, I had to apply the fixups right away to save you from solving the conflicts next week :smile:

packages/framework/src/Model/Transport/Transport.php
vitek-rostislav: :+1: changed to array in 59eaa13d6b

grossmannmartin (approved):

packages/framework/src/Model/Transport/TransportRepository.php
```diff
@@ -170,12 +170,26 @@ public function getAllWithEagerLoadedDomainsAndTranslations(
...
 
         return $queryBuilder
             ->getQuery()
             ->getResult();
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Transport\Transport $transport
+     */
+    public function deleteAllPricesByTransport(Transport $transport): void
```
pk16011990, line 186: @vitek-rostislav is it necessary? TransportPrice has `onDelete="CASCADE"`

vitek-rostislav: this is for updating the collection (the simple approach "delete all" -> "recreate everything", see https://github.com/shopsys/shopsys/pull/3397/files#diff-671ebf9672321f890b717fc9aaea9739c8ce69b8daa85f85d7429aa52ec37c5cR127). I tried to implement the clever way of updating the collection with removing removed only, adding new only, and editing the edited ones, however, the code was too complicated and we agreed with  @grossmannmartin that it is not worth it

PR #3402, opened 2024-09-03
[framework] fixed wrong columns used in migration Version20240102112523

stanoMilan:
This migration (version 20240102112523) adds a new column total_product_price_without_vat to the orders table. It also includes an SQL script to update existing records in this table. The script calculates the value of total_product_price_without_vat based on the price_with_vat column from the order_items associated with each order.

### DIFF
--- a/UPGRADE-14.0.md
+++ b/UPGRADE-14.0.md
@@ -131,6 +131,10 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 -   method `\Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlFacade::removeFriendlyUrlsForAllDomains()` has been removed without a replacement
 -   see #project-base-diff to update your project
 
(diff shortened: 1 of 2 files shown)


TomasLudvik (approved):

UPGRADE-14.0.md
```diff
@@ -131,6 +131,10 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
 -   method `\Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlFacade::removeFriendlyUrlsForAllDomains()` has been removed without a replacement
 -   see #project-base-diff to update your project
 
+#### fixed wrong columns used in migration Version20240102112523 ([#3402](https://github.com/shopsys/shopsys/pull/3402))
```
vitek-rostislav, line 134: personally, I would rather amend the existing upgrade notes for https://github.com/shopsys/shopsys/pull/2952. I would add there the warning and a note "see also [#2952](2952) where the migration was fixed". However, I do not insist on that

vitek-rostislav (approved):

PR #3405, opened 2024-09-05
[framework] fixed redundant log for the Money type if the scale of compared object was different

stanoMilan:

Update:
The problem is caused by different scales of Money objects. By default, the scale is set to 6, but the default scale of the MoneyType form type is 2. Now ResolvedChanges::isOldValueSameAsNewValue takes into consideration MoneyType and compares it correctly.

Previous:
In our project, changes were being logged on the Product entity for the Money type, even when there was no actual change. This was likely due to rounding and the fact that it was being compared as a string. The system was checking values like 0.000000 and 0.00, which are essentially both 0. This might be an issue specific to our project, but is it correct that Money is compared as a string?

We have the following definition in the entity, for example:

```
/**
 * @var \Shopsys\FrameworkBundle\Component\Money\Money
 * @ORM\Column(type="money", precision=20, scale=6)
 */
private $discountInPercent;
```
And in the form type:

```
$builder->add('discountInPercent', MoneyType::class, [
    'scale' => 6,
    'required' => false,
    'invalid_message' => 'Please enter discount in correct format (positive number with decimal separator)',
    'constraints' => [
        new NotBlank(['message' => 'Please enter discount']),
        new MoneyRange([
            'min' => Money::zero(),
            'minMessage' => 'Discount must be greater or equal to {{ limit }}%',
            'max' => Money::create(100),
            'maxMessage' => 'Discount must be less or equal to {{ limit }}%',
        ]),
    ],
    'attr' => [
        'class' => 'form-line__item--pg',
    ],
    'label' => false,
]);
```
image

Perhaps the issue lies somewhere in our configuration.

### DIFF
--- a/UPGRADE-14.0.md
+++ b/UPGRADE-14.0.md
@@ -165,6 +165,10 @@ Follow the instructions in relevant sections, e.g. `shopsys/coding-standards` or
              - { name: doctrine.event_listener, event: postFlush, priority: 1 }
     ```
 
+#### fixed redundant log for the Money type if the scale of compared object was different ([#3405](https://github.com/shopsys/shopsys/pull/3405))
+
+-   class `\Shopsys\FrameworkBundle\Component\EntityLog\ChangeSet\ResolvedChanges` now implements `\JsonSerializable` interface to provide output for the `json_encode` function, update your implementation if you need to provide more than default data
+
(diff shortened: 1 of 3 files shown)


vitek-rostislav (approved):
nice catch :ok_hand:

PR #3417, opened 2024-09-09
[shopsys] immediate access token revocation

grossmannmartin:

When some customer user-related change is made – typically change password – the previously made login should be revoked. 
So when the user notices suspicious activity, he can change his password to protect his account. During this password change, the user's refresh tokens are removed, so it's not possible to re-authenticate with stored data.

But the access token remained valid until the expiration (max 5 minutes). This has opened the door for the attacker to do something bad. 

This PR introduces another layer of security – when some security-related change is made and the refresh tokens are removed, the time of this change is stored in the database. 
During the access token validity check, this time is compared to the `issued at` time of the access token. 
The token is considered invalid when some security-related change is made after issuing the access token.

#### Fixes issues
... 

- [ ] New feature 

## Summary by CodeRabbit

- **New Features**
	- Streamlined user identification process by integrating the security component for user retrieval.
	- Added a new column to track the last security change for customer users in the database.
	- Enhanced token validation logic to ensure tokens are valid based on issuance times and user security changes.
	- Introduced GraphQL mutations for user login and password change operations.

- **Bug Fixes**
	- Improved token invalidation process after user password changes.

- **Tests**
	- Expanded test coverage for login functionality and access token validity after password changes.

- **Documentation**
	- Updated upgrade notes to clarify immediate revocation of access tokens upon refresh token removal.

----
  - https://mg-immediate-pass-change.odin.shopsys.cloud
  - https://cz.mg-immediate-pass-change.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/EntityLog/Detection/DetectionFacade.php
+++ b/packages/framework/src/Component/EntityLog/Detection/DetectionFacade.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Component\EntityLog\Detection;
 
 use Shopsys\FrameworkBundle\Component\EntityLog\Enum\EntityLogSourceEnum;
-use Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser;
+use Shopsys\FrontendApiBundle\Model\User\FrontendApiUser;
 use Symfony\Component\Security\Core\Security;
 
 class DetectionFacade
@@ -16,11 +16,9 @@ class DetectionFacade
 
     /**
      * @param \Symfony\Component\Security\Core\Security $security
-     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser $currentCustomerUser
      */
     public function __construct(
         protected readonly Security $security,
-        protected readonly CurrentCustomerUser $currentCustomerUser,
     ) {
     }
 
@@ -29,10 +27,10 @@ class DetectionFacade
         $this->source = EntityLogSourceEnum::USER;
         $this->userIdentifier = 'notLoggedCustomer';
 
-        $currentCustomer = $this->currentCustomerUser->findCurrentCustomerUser();
+        $user = $this->security->getUser();
 
-        if ($currentCustomer) {
-            $this->userIdentifier = $currentCustomer->getUserIdentifier();
+        if ($user instanceof FrontendApiUser) {
+            $this->userIdentifier = $user->getUserIdentifier();
         }
     }
 
--- a/packages/framework/src/Model/Customer/User/CustomerUserFacade.php
+++ b/packages/framework/src/Model/Customer/User/CustomerUserFacade.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Model\Customer\User;
 
 use DateTime;
+use DateTimeInterface;
 use Doctrine\ORM\EntityManagerInterface;
 use Shopsys\FrameworkBundle\Component\String\HashGenerator;
 use Shopsys\FrameworkBundle\Model\Administrator\Administrator;
@@ -453,5 +454,15 @@ class CustomerUserFacade
     public function getAll(): array
     {
         return $this->customerUserRepository->getAll();
+    }
+
+    /**
+     * @param string $customerUserUuid
+     * @param \DateTimeInterface $referenceDateTime
+     * @return bool
+     */
+    public function isLastSecurityChangeOlderThan(string $customerUserUuid, DateTimeInterface $referenceDateTime): bool
+    {
+        return $this->customerUserRepository->isLastSecurityChangeOlderThan($customerUserUuid, $referenceDateTime);
     }
 }
--- a/packages/frontend-api/src/Model/Token/TokenAuthenticator.php
+++ b/packages/frontend-api/src/Model/Token/TokenAuthenticator.php
@@ -5,6 +5,9 @@ declare(strict_types=1);
 namespace Shopsys\FrontendApiBundle\Model\Token;
 
 use GraphQL\Error\FormattedError;
+use Lcobucci\JWT\Token\RegisteredClaims;
+use Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserFacade;
+use Shopsys\FrontendApiBundle\Model\Token\Exception\InvalidTokenUserMessageException;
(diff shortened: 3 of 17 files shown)


packages/framework/src/Component/EntityLog/Detection/DetectionFacade.php
```diff
@@ -16,11 +16,9 @@ class DetectionFacade
 
     /**
      * @param \Symfony\Component\Security\Core\Security $security
-     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser $currentCustomerUser
      */
     public function __construct(
         protected readonly Security $security,
-        protected readonly CurrentCustomerUser $currentCustomerUser,
```
vitek-rostislav, line 23: this change should be mentioned in the upgrade notes

packages/framework/src/Model/Customer/User/CustomerUserFacade.php
```diff
@@ -454,4 +455,14 @@ public function getAll(): array
     {
         return $this->customerUserRepository->getAll();
     }
+
+    /**
+     * @param string $uuid
+     * @param \DateTimeInterface $issuedAt
+     * @return bool
+     */
+    public function isIssuedAtStillValid(string $uuid, DateTimeInterface $issuedAt): bool
```
vitek-rostislav: I have two "problems" with this:
1) from CustomerUserFacade point of view, it is not clear what `issuedAt` stands for IMHO
2) and this will be a thin ice, it feels like a strictly frontend-api related functionality, should it be in the framework package?

vitek-rostislav: maybe it would make sense to rename it to something like `islLastSecurityChangeOlderThan` ... let's discus it :smile:

vitek-rostislav: ...or we could change it to getter (`getLastSecurityChangeDateTimeByUuid` and add `isIssuedAtStillValid` method to `TokenAutheticator` :thinking:

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
```diff
@@ -26,10 +29,12 @@ class TokenAuthenticator extends AbstractAuthenticator
     /**
      * @param \Shopsys\FrontendApiBundle\Model\Token\TokenFacade $tokenFacade
      * @param \Shopsys\FrontendApiBundle\Model\User\FrontendApiUserProvider $frontendApiUserProvider
+     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserFacade $customerUserFacade
      */
     public function __construct(
         protected readonly TokenFacade $tokenFacade,
         protected readonly FrontendApiUserProvider $frontendApiUserProvider,
+        protected readonly CustomerUserFacade $customerUserFacade,
```
vitek-rostislav, line 37: this should be mentioned in the upgrade notes

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginTest.php
```diff
@@ -16,12 +16,60 @@ class LoginTest extends GraphQlTestCase
      */
     protected TokenFacade $tokenFacade;
 
+    public function testAccessTokenIsNotValidAfterPasswordChange(): void
```
vitek-rostislav: Personally, I would expect this test in a new class (what about something like `LoginTokenInvalidationTest`), not in `LoginTest`, however, this is really marginal :shrug:

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginTest.php
```diff
@@ -16,12 +16,60 @@ class LoginTest extends GraphQlTestCase
...
+
+        $response = $this->getResponseContentForGql(
+            __DIR__ . '/graphql/ChangePasswordMutation.graphql',
+            [
+                'email' => $this->getDefaultCredentials()['email'],
+                'oldPassword' => $this->getDefaultCredentials()['password'],
+                'newPassword' => 'user124',
+            ],
+        );
+        $responseData = $this->getResponseDataForGraphQlType($response, 'ChangePassword');
```
vitek-rostislav: I would like to discuss the behavior from a UX point of view - is it ok to logout the current customer when he changes his own password? This might be out of the scope of this issue, however, I would expect the password change would logout the customer from all the devices except the current one :thinking: Let's talk about it, please

vitek-rostislav (changes requested):

packages/framework/src/Component/EntityLog/Detection/DetectionFacade.php
grossmannmartin: I thought about it, and then I wasn't sure about our latest decision for the upgrade notes 🤔

packages/framework/src/Model/Customer/User/CustomerUserFacade.php
grossmannmartin: Fair enough, let's discuss it, please... I think it should be in the framework package as the $lastSecurityChange is in the framework also

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginTest.php
grossmannmartin: This is a current behavior. 
But prior to this change, it behaves a little more weirdly, as the user is logged out after the access token expires.
We may discuss the possible improvements, but I would like to separate it to the new PR

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginTest.php
grossmannmartin: good idea, that allowed me to write the test more understandably (split into several private methods) 033186115b

packages/framework/src/Component/EntityLog/Detection/DetectionFacade.php
grossmannmartin: added 2b5a080058

packages/frontend-api/src/Model/Token/TokenAuthenticator.php
grossmannmartin: added 2b5a080058

packages/framework/src/Model/Customer/User/CustomerUserFacade.php
grossmannmartin: renamed as we discussed - 2f1226ed9f

vitek-rostislav (approved):

PR #3465, opened 2024-09-30
[shopsys] product inquiries

### LINKED ISSUE #2069: Cookbook: bad example for creating of new mail template
### What is happening
Cookbook in documentation about new mail template is bad. https://github.com/shopsys/shopsys/blob/master/docs/cookbook/adding-a-new-email-template.md contains example with `MailTemplateDataFixture` but it's not useful on running project... Usefull is wrong word, danger is better. Datafixture works perfectly on localhost during development, CI works too... but production fails, because new template is not in database. Solution is about removing changes in DataFixture (or rework it into editation) and create valid DB migration for all domains with `MultidomainMigrationTrait`

### Expected result
Cookbook about new email template is usable for running projects, not for local development.

grossmannmartin:

This PR adds an inquiry feature, so it's now possible to mark some products as `upon inquiry`. 
Such a product has its price hidden and cannot be bought.
The new mutation is introduced to send an inquiry for a product. Info about inquiring is sent by email to the customer and admin with editable email templates. 
Also, an admin panel was introduced to see sent inquiries.

#### Fixes issues
closes #2069  

- [ ] New feature 

## Release Notes

- **New Features**
	- Added functionality for managing inquiries in the admin interface, including listing and detail views.
	- Implemented a new monetary value representation called `HiddenMoney`, which masks actual amounts.

- **Database Changes**
	- New database tables for inquiries and updated product types to enhance data structure.

- **Enhancements**
	- Improved the product form to conditionally include a product type field.
	- Enhanced email notification system with customizable templates for inquiries.

----
  - https://mg-product-inquiry.odin.shopsys.cloud
  - https://cz.mg-product-inquiry.odin.shopsys.cloud


grossmannmartin:
@coderabbitai review

packages/framework/src/Model/Inquiry/Mail/InquiryMail.php
```diff
@@ -0,0 +1,184 @@
...
+    public const string VARIABLE_COMPANY_NUMBER = '{companyNumber}';
+    public const string VARIABLE_COMPANY_TAX_NUMBER = '{companyTaxNumber}';
+    public const string VARIABLE_NOTE = '{note}';
+
+    public const string VARIABLE_PRODUCT_NAME = '{productName}';
+    public const string VARIABLE_PRODUCT_CATALOG_NUMBER = '{productCatnum}';
+    public const string VARIABLE_PRODUCT_URL = '{productUrl}';
+    public const string VARIABLE_PRODUCT_IMAGE = '{productImageUrl}';
+
+    public const string VARIABLE_ADMIN_INQUIRY_DETAIL_URL = '{adminInquiryDetailUrl}';
```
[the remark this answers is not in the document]
grossmannmartin: @coderabbitai in PHP8.3 it is valid

packages/framework/src/Component/Money/HiddenMoney.php
```diff
@@ -0,0 +1,158 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Money;
+
+class HiddenMoney extends Money
```
vitek-rostislav, line 7: I would like to talk about this class. On the one hand, it is a neat solution and I can see the convenience, on the other hand, all the methods that need to be overridden seem to me a little overkill IMHO, moreover, the results of the arithmetic operations might be confusing here (I would probably rather throw exceptions on methods like `zero()`, `add()`, `isNegative()`, etc.

project-base/storefront/components/Pages/ProductDetail/ProductDetailAddToCart/ProductDetailAddToCart.tsx
```diff
@@ -50,33 +50,59 @@ export const ProductDetailAddToCart: FC<ProductDetailAddToCartProps> = ({ produc
...
-                            >
-                                {isAddingToCart ? <Loader className="w-[18px]" /> : <CartIcon className="w-[18px]" />}
-                                {t('Add to cart')}
-                            </Button>
-                        </div>
-                    </div>
-                </div>
-            )}
+            {(() => {
+                switch (true) {
```
vitek-rostislav: interesting construct, indeed :relaxed:

packages/frontend-api/src/Model/Mutation/Inquiry/CreateInquiryMutation.php
```diff
@@ -0,0 +1,77 @@
...
+     * @return bool
+     */
+    public function createInquiryMutation(Argument $argument): bool
+    {
+        try {
+            $inquiryData = $this->createInquiryDataFromArgument($argument);
+            $this->inquiryFacade->create($inquiryData);
+        } catch (ProductNotFoundException) {
+            throw new ProductNotFoundUserError(sprintf('Product with UUID "%s" not found', $argument['input']['productUuid']));
+        } catch (Exception $exception) {
```
vitek-rostislav: is this a new concept for our mutations? I mean I like it, just that we usually do not do that, right?

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Inquiry/Input/CreateInquiryInputDecorator.types.yaml
```diff
@@ -0,0 +1,69 @@
+CreateInquiryInputDecorator:
+    type: input-object
+    inherits:
+        - 'TelephoneInputObjectDecorator'
```
vitek-rostislav, line 5: nice, I wanted to propose using this and you are already using it :grin:  `NameInputObjectDecorator` can be used as well, btw.

One more thing, the decorators are defined in project-base, however, I moved them in https://github.com/shopsys/shopsys/commit/bdaf9c822b8f4d86e86bc35972f032a2b54ba829#diff-b28e90a71f949bb127244bd8dfe3882b3a1fe4ee430d645f657c134681d14a82R8

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Inquiry/Input/CreateInquiryInputDecorator.types.yaml
```diff
@@ -0,0 +1,69 @@
...
+                            message: "Please fill in numbers only"
+            companyTaxNumber:
+                type: "String"
+                description: "The customer’s company tax number"
+                validation:
+                    -   Length:
+                            max: 50
+                            maxMessage: "Tax number cannot be longer than {{ limit }} characters"
+                    -   Regex:
+                            pattern: '/^[0-9A-Z]*([0-9]+[A-Z]+|[A-Z]+[0-9]+)[0-9A-Z]*$/'
```
vitek-rostislav, line 48: we have the pattern already 3 times in our input type definitions :smile: it would be nice to get rid of the DRY violations in the API definitions in general, one day, one day, ...

I did not want to create another sticky note but here we go
IMG20241007144708 (1)

packages/framework/src/Model/Inquiry/InquiryRepository.php
```diff
@@ -24,4 +26,22 @@ protected function getInquiryRepository(): EntityRepository
...
+
+    /**
+     * @param string $locale
+     * @return \Doctrine\ORM\QueryBuilder
+     */
+    public function getInquiriesQueryBuilder(string $locale): QueryBuilder
+    {
+        return $this->getInquiryRepository()
+            ->createQueryBuilder('i')
+            ->addSelect('IDENTITY(i.product) as productId')
```
vitek-rostislav, line 54: does this work when the product is null?

packages/framework/src/Resources/views/Admin/Content/Inquiry/listGrid.html.twig
```diff
@@ -0,0 +1,19 @@
+{% extends '@ShopsysFramework/Admin/Grid/Grid.html.twig' %}
+
+{% block grid_value_cell_id_productName %}
+    {{ value }}
+    <a href="{{ url('admin_product_edit', { id: row.productId }) }}">{{ icon('forward-page') }}</a>
```
vitek-rostislav: product is nullable, right?

project-base/app/src/DataFixtures/Demo/InquiryDataFixture.php
```diff
@@ -0,0 +1,485 @@
...
+        $inquiryData = $this->inquiryDataFactory->create();
+        $inquiryData->firstName = 'Madison';
+        $inquiryData->lastName = 'Perry';
+        $inquiryData->email = 'madison.perry@example.com';
+        $inquiryData->telephone = '888444555';
+        $inquiryData->companyName = 'Halcyon Holdings Corporation';
+        $inquiryData->companyNumber = '99887755';
+        $inquiryData->companyTaxNumber = 'US99887755';
+        $inquiryData->note = 'Can you provide more info?';
+        $inquiryData->product = $inquiryProduct;
```
vitek-rostislav, line 477: to fulfill my holy trinity of product-nullability-related comments, what about setting the product to NULL in at least one of the demo inquiries (but I see this won't be easy with the current codebase, so :shrug: )

packages/framework/src/Model/Inquiry/InquiryGridFactory.php
```diff
@@ -47,6 +47,7 @@ public function createView(
         $grid->addColumn('company', 'company', t('Company (Company number)'), true);
         $grid->addColumn('createdAt', 'i.createdAt', t('Created'), true);
 
+        $grid->addActionColumn('file-all', t('Show detail '), 'admin_inquiry_detail', ['id' => 'i.id']);
```
vitek-rostislav: 'Show detail ' - trailing whitespace?

project-base/app/src/DataFixtures/Demo/InquiryDataFixture.php
```diff
@@ -394,7 +395,7 @@ public function load(ObjectManager $manager): void
         $inquiryData->createdAt = new DateTimeImmutable('2024-09-02 09:25:41');
         $this->inquiryFacade->create($inquiryData);
 
-        $inquiryData = $this->inquiryDataFactory->create();
+        $inquiryData = $this->inquiryDataFactory->create(Domain::FIRST_DOMAIN_ID);
```
vitek-rostislav, line 402: I believe we should use `Domain` service (or `DomainsForDataFixtureProvider` maybe) and create the inquiries for all the remaining domains

packages/framework/src/Model/Inquiry/Inquiry.php
```diff
@@ -146,6 +146,11 @@ public function getLastName()
         return $this->lastName;
     }
 
+    public function getFullName()
```
vitek-rostislav: this method could have return typehint (and should have annotation)

packages/framework/src/Model/Inquiry/Mail/InquiryMail.php
```diff
@@ -0,0 +1,184 @@
...
+    protected function getSubjectVariablesReplacements(Inquiry $inquiry): array
+    {
+        return [
+            self::VARIABLE_FULL_NAME => htmlspecialchars($inquiry->getFullName(), ENT_QUOTES),
+            self::VARIABLE_EMAIL => htmlspecialchars($inquiry->getEmail(), ENT_QUOTES),
+            self::VARIABLE_TELEPHONE => htmlspecialchars($inquiry->getTelephone(), ENT_QUOTES),
+            self::VARIABLE_COMPANY_NAME => $this->escapeOptionalString($inquiry->getCompanyName()),
+            self::VARIABLE_COMPANY_NUMBER => $this->escapeOptionalString($inquiry->getCompanyNumber()),
+            self::VARIABLE_COMPANY_TAX_NUMBER => $this->escapeOptionalString($inquiry->getCompanyTaxNumber()),
+            self::VARIABLE_PRODUCT_NAME => htmlspecialchars($inquiry->getProduct()?->getName() ?: '-', ENT_QUOTES),
```
vitek-rostislav: `escapeOptionalString` can be used here as well, or?

packages/framework/src/Model/Cart/Watcher/CartWatcher.php
```diff
@@ -60,6 +61,13 @@ public function getNotListableItems(Cart $cart, CurrentCustomerUser $currentCust
         foreach ($cart->getItems() as $item) {
             try {
                 $product = $item->getProduct();
+
+                if ($product->getProductType() === ProductTypeEnum::TYPE_INQUIRY) {
```
vitek-rostislav, line 65: neat :ok_hand:

vitek-rostislav (changes requested):
Hi, nice job there :ok_hand: In addition to my comments, I have these thoughts:
- it would be nice to extend `ProductVisibilityRepositoryTest` with a scenario that tests product inquiry is visible even if the price is missing
- it would be nice to add FE API test for getting product inquiry to assert its type, isInquiry, and price fields

packages/frontend-api/src/Model/Mutation/Inquiry/CreateInquiryMutation.php
grossmannmartin: right. When I'm looking at it now, I don't think this is the way. 

changed in 6b72050b65befa0220978968f53f9dfdff1ccf9c

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Inquiry/Input/CreateInquiryInputDecorator.types.yaml
grossmannmartin: nice, I used NameInputObjectDecorator, but I left them in project-base as you're already moving them - 191b0610cb

packages/framework/src/Model/Inquiry/InquiryRepository.php
grossmannmartin: this works, but the join below don't 😆 

fixed in 7753025b68

packages/framework/src/Resources/views/Admin/Content/Inquiry/listGrid.html.twig
grossmannmartin: Yes. Yes, it is. fixed in 7753025b68

project-base/app/src/DataFixtures/Demo/InquiryDataFixture.php
grossmannmartin: This is so true... I changed the logic a bit, so now it is possible to programmatically create an inquiry without the product – this may be useful in ERP connectors.
As a profit, I have set some inquiries to not have a product.  - ea2818aa53

packages/framework/src/Model/Inquiry/InquiryGridFactory.php
grossmannmartin: removed trailing whitespace - 90443b4ad3

project-base/app/src/DataFixtures/Demo/InquiryDataFixture.php
grossmannmartin: I don't find it necessary. 
This may be over-complicated as the inquiries from all domains are displayed in a single grid. 
When we iterate over all domains, we have to handle the duplicities or fake data on other domains. 
TBH, for me it is not worth the time. 
But I'm open to suggestions:)

packages/framework/src/Model/Inquiry/Inquiry.php
grossmannmartin: fixed in 54d9ead145

packages/framework/src/Model/Inquiry/Mail/InquiryMail.php
grossmannmartin: Yes. Yes, it should. 578eebd581

packages/framework/src/Component/Money/HiddenMoney.php
grossmannmartin: let's talk :)

project-base/app/tests/App/Functional/Model/Product/ProductVisibilityRepositoryTest.php
```diff
@@ -221,6 +222,17 @@ public function testIsVisibleOnAnyDomainWhenSellingInFuture(): void
         $this->assertFalse($productVisibility->isVisible());
     }
 
+    public function testIsVisibleWhenProductIsInquiryWithoutPrice(): void
```
vitek-rostislav, line 225: :heart_eyes:

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductInquiryTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace Tests\FrontendApiBundle\Functional\Product;
+
+use App\DataFixtures\Demo\ProductDataFixture;
+use App\Model\Product\Product;
+use Tests\FrontendApiBundle\Test\GraphQlTestCase;
+
+class ProductInquiryTest extends GraphQlTestCase
```
vitek-rostislav, line 11: :heart_eyes:

vitek-rostislav (approved):

JanMolcik (approved):

JanMolcik (approved):

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -310,6 +311,25 @@ protected function extractParameters(string $locale, Product $product): array
...
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @return string
+     */
+    protected function extractProductType(Product $product): string
+    {
+        if ($product->isMainVariant()) {
+            foreach ($product->getVariants() as $variant) {
```
vitek-rostislav: this should probably use `getVariantsForDefaultPricingGroup` method - otherwise, the only one basic variant product might be hidden and the price is then incorrectly displayed for the main variant, see:

image

packages/framework/src/Model/Product/Pricing/ProductPriceCalculation.php
```diff
@@ -61,6 +62,11 @@ protected function calculateMainVariantPrice(Product $mainVariant, $domainId, Pr
             $pricingGroup,
         );
 
+        $variants = array_filter(
```
vitek-rostislav, line 65: just a thought - should not the `productType !== ProductTypeEnum::TYPE_INQUIRY` condition be a part of `getAllSellableQueryBuilder`? I would expect so because technically, the inquiry products are not sellable :thinking: I am afraid of what this change would cause, on the other hand, I am also afraid of confusion among future developers that inquiry products pass the "is sellable" condition. Let's discuss this, please

packages/framework/src/Model/Product/Search/FilterQuery.php
```diff
@@ -75,6 +75,7 @@ public function applyOrdering(string $orderingModeId, PricingGroup $pricingGroup
         ];
 
         if ($orderingModeId === ProductListOrderingConfig::ORDER_BY_PRIORITY) {
+            $clone->sorting['product_type'] = 'asc';
```
vitek-rostislav: I am wondering, should not we somehow ensure the inquiry products are last even after adding a new product type (e.g. "voucher") on a project? We can discuss this as well :smile:

vitek-rostislav: I am thinking about adding a more explicit field into the index, something like "priority_by_product_type" that could be used here :thinking:

packages/product-feed-heureka-delivery/src/Model/FeedItem/HeurekaDeliveryDataRepository.php
```diff
@@ -33,6 +34,9 @@ public function getDataRows(
         int $maxResults,
     ): array {
         $queryBuilder = $this->productRepository->getAllSellableQueryBuilder($domainConfig->getId(), $pricingGroup);
+        $queryBuilder
+            ->andWhere('p.productType != :inquiryProductType')
```
vitek-rostislav, line 38: this relates to my question whether "inquiry" is or is not a "sellable" product? :slightly_smiling_face:

project-base/storefront/components/Blocks/Product/ProductsList/ProductListItem.tsx
```diff
@@ -118,7 +118,7 @@ export const ProductListItem = forwardRef<HTMLLIElement, ProductItemProps>(
                         />
                     )}
 
-                    {visibleItemsConfig.storeAvailability && (
+                    {visibleItemsConfig.storeAvailability && !product.isInquiryType && (
```
vitek-rostislav: instead of adding a condition on the storefront, shouldn't these fields (`availableStoresCount`, `storeAvailability`, ...) be null/empty/somehow undefined/0 in the API? Would that be difficult to implement? Do you want to talk about it? Do you want to hurt me now? :smile:

vitek-rostislav (changes requested):

project-base/storefront/components/Blocks/Product/ProductsList/ProductListItem.tsx
grossmannmartin: Just slightly 🤣
It was a deliberate choice. I can imagine few scenarios where the stock may be useful even for the inquiry products. This gives a freedom to implement these. 
For example we may wan't to distinguish products that are available, but have to be for example engraved or assembled and the products that have to be ordered from the supplier.
And yes it would probably be unnecessarily complex in the backend code.

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
grossmannmartin: changed, good catch 👍

packages/framework/src/Model/Product/Search/FilterQuery.php
grossmannmartin: added, good idea 👍

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -331,6 +333,22 @@ protected function extractProductType(Product $product, int $domainId): string
...
+     */
+    protected function extractPriorityByProductType(Product $product, int $domainId): int
+    {
+        $productType = $this->extractProductType($product, $domainId);
+
+        return match ($productType) {
+            ProductTypeEnum::TYPE_BASIC => 10,
+            ProductTypeEnum::TYPE_INQUIRY => 20,
+            default => 100,
+        };
```
vitek-rostislav: I would expect it in a reverse order - the higher priority the higher in the list is the product displayed, or? `ordering_priority` field works that way (the ordering is then `DESC`), I would unify that. However, I do not insist on that and I do not want to postpone the testing because of that, so I will leave this up to you, whether you want to change it :wink: One way or another, you need to resolve the conflicts before moving the task to tests

vitek-rostislav (approved):

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
grossmannmartin: changed, thanks

PR #3480, opened 2024-10-07
[shopsys] Convertim integration

TomasLudvik:
Convertim checkout (https://www.convertim.com) is part of Shopsys family and it was about time to finally implement it into our platform.

- [ ] New feature 


packages/frontend-api/src/Model/Resolver/Cart/CartItemResolverMap.php
```diff
@@ -0,0 +1,82 @@
...
+
+                if (count($discounts) === 0) {
+                    return [];
+                }
+
+                foreach ($discounts as $discount) {
+                    $discountsData[] = [
+                        'promoCode' => $discount->promoCode->getCode(),
+                        'totalDiscount' => new Price($discount->totalPriceWithoutVat, $discount->totalPriceWithVat),
+                        'unitDiscount' => new Price($discount->unitPriceWithoutVat, $discount->unitPriceWithVat),
```
vitek-rostislav: I believe you can use `getTotalPrice()` and `getUnitPrice()` methods on `$discount` object as it is an instance of `OrderItemData` class if I am not mistaken

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Cart/CartItemDecorator.types.yaml
```diff
@@ -13,3 +13,6 @@ CartItemDecorator:
             quantity:
                 type: "Int!"
                 description: "Quantity of items in the cart"
+            discount:
+                type: "[CartItemDiscount!]"
```
vitek-rostislav:
```suggestion
                type: "[CartItemDiscount!]!"
```

right?

vitek-rostislav:
```suggestion
                type: "[CartItemDiscount!]!"
```

right?

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Cart/CartItemDecorator.types.yaml
```diff
@@ -13,3 +13,6 @@ CartItemDecorator:
             quantity:
                 type: "Int!"
                 description: "Quantity of items in the cart"
+            discount:
```
vitek-rostislav: maybe, for better clarity, it should be named discount**s**?

packages/frontend-api/src/Model/Resolver/Cart/CartItemResolverMap.php
```diff
@@ -0,0 +1,82 @@
...
+                return $this->orderDataFactory->createFromCart($cartItem->getCart(), $this->domain->getCurrentDomainConfig());
+            },
+            $cartItem->getCart()->getCartIdentifier(),
+        );
+
+        $discountsData = [];
+
+        foreach ($orderData->items as $orderItem) {
+            if ($orderItem->type === OrderItemTypeEnum::TYPE_PRODUCT && $orderItem->product === $cartItem->getProduct()) {
+                $discounts = $orderItem->relatedOrderItemsData;
```
vitek-rostislav: I would name this rather `$relatedOrderItemsData` as at this point, we can not be sure there are discounts only, or? Later, in the foreach below, I would add something like 
```
if ($relatedOrderItemData->type !== OrderItemTypeEnum::TYPE_DISCOUNT`) {
    continue;
}
```

vitek-rostislav: maybe `$relatedOrderItemData->promoCode !== null` could be added into the condition as well

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Cart/CartItemDecorator.types.yaml
vitek-rostislav: could you also extend the API cart tests (or create new one(s)) that would cover this new field? :pray:

packages/framework/src/Model/Order/PromoCode/PromoCode.php
```diff
@@ -207,4 +208,16 @@ public function getPrefix()
...
     }
+
+    /**
+     * @return string
+     */
+    public function getDiscountTypeString(): string
+    {
+        return match ($this->discountType) {
+            self::DISCOUNT_TYPE_PERCENT => 'percent',
+            self::DISCOUNT_TYPE_NOMINAL => 'nominal',
```
vitek-rostislav: maybe we can change the field completely to string? I believe the migration would not be difficult to write and it would simplify the things :slightly_smiling_face:

packages/frontend-api/src/Model/Cart/CartWatcherFacade.php
```diff
@@ -143,10 +159,41 @@ protected function checkPromoCodeValidity(Cart $cart): void
...
                 $this->cartPromoCodeFacade->removePromoCode($cart, $promoCode);
                 $this->cartWithModificationsResult->addChangedPromoCode($promoCode->getCode());
             }
         }
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Cart\Cart $cart
+     */
+    protected function setPromoCodesWithAppliedDiscount(Cart $cart): void
```
vitek-rostislav, line 170: it does not feel right to be placed in the watcher, however, we already have it "wrong" and there is an issue for refactoring so I believe I can live with that. I wanted to propose you move the method into `CartWithModificationsResult` but you need all the services for that, so... 

However, another idea - is not there all the necessary information in the order data instance after using the order processor? :thinking: it feels like you are duplicating the calculation here

packages/framework/src/Model/Order/PromoCode/PromoCode.php
vitek-rostislav: btw, something like PromoCodeTyoeEnum could be created then, that would be beautiful :innocent:

packages/frontend-api/src/Model/Order/PromoCode/PromoCodeWithDiscount.php
```diff
@@ -0,0 +1,21 @@
...
+
+class PromoCodeWithDiscount
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\PromoCode\PromoCode $promoCode
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Price $discount
+     */
+    public function __construct(
+        public PromoCode $promoCode,
+        public Price $discount,
```
vitek-rostislav:
```suggestion
        public readonly PromoCode $promoCode,
        public readonly Price $discount,
```

packages/frontend-api/src/Model/Resolver/Order/PromoCode/PromoCodeResolverMap.php
```diff
@@ -0,0 +1,28 @@
...
+
+class PromoCodeResolverMap extends ResolverMap
+{
+    /**
+     * @return array
+     */
+    protected function map(): array
+    {
+        return [
+            'PromoCode' => [
```
vitek-rostislav, line 18: I believe, for better clarity, it would be better to name the API type "PromoCodeWithDiscount(+Decorator)" as well

project-base/app/tests/FrontendApiBundle/Functional/Cart/ApplyPromoCodeToCartTest.php
```diff
@@ -68,15 +68,17 @@ public function testApplyPromoCode(): void
                 promoCode: "' . $promoCode->getCode() . '"
             }) {
                 uuid
-                promoCode
+                promoCodes {
+                    code
```
vitek-rostislav, line 72: could you please add all the fields (code, type, discount) to the assertions in all the affected tests?

upgrade-notes/backend_20241007_110001.md
```diff
@@ -0,0 +1,5 @@
+#### enable Convertim usage on your project ([#{pullRequestId}](https://github.com/shopsys/shopsys/pull/{pullRequestId}))
+
+-   `frontend-api` `Cart` object now has property `promoCodes (array)` instead of current `promoCode (string)` and includes additional data about promo codes
+-   multiple constructor interfaces have been changed, review `PHP errors` and `standards` and fix possible issues
```
vitek-rostislav: I do not think we need to use this sentence, I hope we will put some general info at the beginning of the upgrade notes

vitek-rostislav (changes requested):

vitek-rostislav:
One more thing, in the acceptance criteria, they say it should be possible to enable/disable Convertim on particular domains and should be implemented as an optional project-base dependency, but I do not see that in this pull request. Is that going to be implemented elsewhere?

packages/frontend-api/src/Model/Cart/CartWatcherFacade.php
TomasLudvik: I agree with everything and that is why it is made this way.

> However, another idea - is not there all the necessary information in the order data instance after using the order processor? 🤔 it feels like you are duplicating the calculation here

Its one calculation and it felt ok to do it like this. You think it shoudl be reworked to running whole OrderData calculations?

packages/frontend-api/src/Model/Resolver/Order/PromoCode/PromoCodeResolverMap.php
TomasLudvik: I actually do not think so. Objects in API are not same objects as in backend. API users do not need to know, that our `PromoCode` entity does not include discount. They need `PromoCode` as final product :D  For example we call `CartWithModificationsResult` just a `Cart`.

project-base/app/tests/FrontendApiBundle/Functional/Cart/ApplyPromoCodeToCartTest.php
TomasLudvik: I have added it only to this one and created second one for nominal promo code. I do not feel, that including it in all tests brings any benefit.

TomasLudvik:
> One more thing, in the acceptance criteria, they say it should be possible to enable/disable Convertim on particular domains and should be implemented as an optional project-base dependency, but I do not see that in this pull request. Is that going to be implemented elsewhere?

This is currently only draft adding just frontend API functionality that does not need to be dependent on Convertim, so currently there is no switch needed.

packages/frontend-api/src/Model/Resolver/Order/PromoCode/PromoCodeResolverMap.php
vitek-rostislav: ok, I can live with that

packages/frontend-api/src/Model/Cart/CartWatcherFacade.php
vitek-rostislav: yes please, let's keep the unity here and use the processor, otherwise, changes of promo-code-calculation related logic would be error-prone. I discussed this with @grossmannmartin as well as I wanted the 3rd opinion a he also agrees with the processor usage here :pray:

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Cart/CartItemDecorator.types.yaml
```diff
@@ -14,5 +14,5 @@ CartItemDecorator:
                 type: "Int!"
                 description: "Quantity of items in the cart"
             discount:
-                type: "[CartItemDiscount!]"
+                type: "[CartItemDiscount!]!"
                 description: "Discount applied to the cart item"
```
vitek-rostislav: nitpick, 
```suggestion
                description: "Discounts applied to the cart item"
```

project-base/app/tests/FrontendApiBundle/Functional/Cart/ApplyPromoCodeToCartTest.php
```diff
@@ -78,7 +84,58 @@ public function testApplyPromoCode(): void
...
         self::assertEquals(CartDataFixture::CART_UUID, $data['uuid']);
-        self::assertEquals($promoCode->getCode(), $data['promoCodes'][0]['code']);
+        self::assertEquals(
+            [
+                'code' => $promoCode->getCode(),
+                'type' => 'percentage',
+                'discount' => [
+                    'priceWithVat' => '10.000000',
+                    'priceWithoutVat' => '10.000000',
+                    'vatAmount' => '0.000000',
```
vitek-rostislav: we should use `getFormattedMoneyAmountConvertedToDomainDefaultCurrency()` (or maybe better `getSerializedPriceConvertedToDomainDefaultCurrency()`) to assert the prices and keep the tests independent of domains setting changes on projects

project-base/app/tests/FrontendApiBundle/Functional/Cart/ApplyPromoCodeToCartTest.php
```diff
@@ -78,7 +84,58 @@ public function testApplyPromoCode(): void
...
+
+        self::assertEquals(CartDataFixture::CART_UUID, $data['uuid']);
+        self::assertEquals(
+            [
+                'code' => $promoCode->getCode(),
+                'type' => 'nominal',
+                'discount' => [
+                    'priceWithVat' => '100.000000',
+                    'priceWithoutVat' => '82.640000',
+                    'vatAmount' => '17.360000',
```
vitek-rostislav: ...same here :innocent:

project-base/app/tests/FrontendApiBundle/Functional/Cart/CartWithPromoCodeTest.php
```diff
@@ -66,6 +79,26 @@ public function testCartManipulationWithPromoCode(): void
...
+                    'totalDiscount' => [
+                        'priceWithVat' => '-14.000000',
+                        'priceWithoutVat' => '-11.570000',
+                        'vatAmount' => '-2.430000',
+                    ],
+                    'unitDiscount' => [
+                        'priceWithVat' => '-14.000000',
+                        'priceWithoutVat' => '-11.570000',
+                        'vatAmount' => '-2.430000',
+                    ],
```
vitek-rostislav: ...same here :innocent:

packages/framework/src/Migrations/Version20241009114907.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20241009114907 extends AbstractMigration
```
vitek-rostislav, line 10: :heart_eyes:

packages/framework/src/Model/Order/PromoCode/PromoCode.php
```diff
@@ -208,16 +205,4 @@ public function getPrefix()
...
-
-    /**
-     * @return string
-     */
-    public function getDiscountTypeString(): string
-    {
-        return match ($this->discountType) {
-            self::DISCOUNT_TYPE_PERCENT => 'percent',
-            self::DISCOUNT_TYPE_NOMINAL => 'nominal',
-            default => throw new InvalidPromoCodeDiscountTypeException($this->discountType),
```
vitek-rostislav: `InvalidPromoCodeDiscountTypeException` is now unused

packages/framework/src/Model/Order/PromoCode/PromoCodeTypeEnum.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Order\PromoCode;
+
+use Shopsys\FrameworkBundle\Component\Enum\AbstractEnum;
+
+class PromoCodeTypeEnum extends AbstractEnum
```
vitek-rostislav, line 9: I know you will probably hate me soon, however, we should also create `PromoCodeTypeEnum(+Decorator)` in FE API and use it in `PromoCodeDecorator.type` :upside_down_face:

packages/frontend-api/src/Model/Cart/CartWatcherFacade.php
TomasLudvik: We need to discuss this as Order does not take into account multiple PromoCodes so I am not able to do that right now @grossmannmartin @pk16011990 @vitek-rostislav

TomasLudvik: I have reworked this for single promo code

vitek-rostislav (approved):

PR #3481, opened 2024-10-07
[monorepo] removed dependency on symplify/monorepo-builder

grossmannmartin:

symplify/monorepo-builder was used exclusively to release a new version of Shopsys Platform.
Most of the code was already bent to our needs, so instead of upgrading to the new version (due to Symfony 6 compatibility), this PR completely removes the package and usage. 

This allows us to improve the release process more in our needs.

#### Fixes issues
... 

- [ ] New feature 

----
  - https://mg-upgrade-monorepo-builder.odin.shopsys.cloud
  - https://cz.mg-upgrade-monorepo-builder.odin.shopsys.cloud


utils/releaser/src/Command/ReleaseCommand.php
```diff
@@ -115,11 +115,27 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        $stage = $input->getOption(Option::STAGE);
+        $stage = $input->getOption(self::STAGE);
 
         return $stage !== null ? (string)$stage : $stage;
     }
 
+    /**
+     * @param string|null $stage
+     */
+    public function checkStage(?string $stage): void
```
vitek-rostislav: this can be private

utils/releaser/src/Command/SymfonyStyleFactory.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * @return \Symfony\Component\Console\Style\SymfonyStyle
+     */
+    public function getPreviouslyCreatedSymfonyStyle(): SymfonyStyle
+    {
+        if (!isset($this->symfonyStyle)) {
+            throw new ShouldNotHappenException('SymfonyStyle was not created yet.');
+        }
+
+        return $this->symfonyStyle;
```
vitek-rostislav, line 37: Isn't it overcomplicated? Is the fresh creation vs getting an already created instance important from outside of the factory? I would simplify it to one method that creates the instance if it is not in the cache, saves it into cache, and returns it, or? :shrug:

vitek-rostislav (approved):

utils/releaser/src/Command/ReleaseCommand.php
grossmannmartin: nice - 950c63c961

utils/releaser/src/Command/SymfonyStyleFactory.php
grossmannmartin: Well, the problem is that I need to get the instance in a place where I don't have the $input and $output accessible (e.g., in ReleaseWorkers). 
In order to unify it into a single method, I would have to make the arguments optional, which complicates things even further (imho).

vitek-rostislav: ah, ok, let's keep it as is :+1:

PR #3484, opened 2024-10-08
[shopsys] SSP-2765 - Parameter groups refactor + editable in Admin

AmpMVn:
Moving parameter groups to the framework, refactoring and the possibility of editing parameter groups in the administration, including sorting. Removed the deprecated visibility attribute, editing the frontend api and listing in the base project according to groups.

- [ ] New feature 

----
  - https://mvn-ssp-2765-product-groups.odin.shopsys.cloud
  - https://cz.mvn-ssp-2765-product-groups.odin.shopsys.cloud


packages/framework/src/Model/Product/Parameter/Parameter.php
```diff
@@ -48,11 +48,6 @@ class Parameter extends AbstractTranslatableEntity
      */
     protected $translations;
 
-    /**
-     * @ORM\Column(type="boolean")
-     */
-    protected $visible;
```
grossmannmartin, line 54: can you please remove the no longer necessary mapping removal?

in the class `App\Component\Doctrine\RemoveMappingsSubscriber`

line 
```
$this->removeColumnsFromEntityMappings(Parameter::class, ['visible'], $classMetadata);
```

packages/framework/src/Model/Product/Parameter/ParameterGroupDataFactoryInterface.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Parameter;
+
+interface ParameterGroupDataFactoryInterface
```
grossmannmartin: We have decided to drop the FactoryInterfaces from the whole codebase as they complicate things without a real added value. Can you please remove this interface?

packages/framework/src/Model/Product/Parameter/ParameterGroupFactoryInterface.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Parameter;
+
+interface ParameterGroupFactoryInterface
```
grossmannmartin: same about this factory interface as mentioned earlier

packages/framework/src/Model/Product/Parameter/ParameterGroupTranslation.php
```diff
@@ -15,7 +16,9 @@
 class ParameterGroupTranslation extends AbstractTranslation
 {
     /**
-     * @Prezent\Translatable(targetEntity="App\Model\Product\Parameter\ParameterGroup")
+     * @var \Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterGroup
+     * @Prezent\Translatable(targetEntity="Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterGroup")
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingNativeTypeHint
```
grossmannmartin: this suppress should not be necessary

packages/framework/src/Controller/Admin/ParameterGroupController.php
```diff
@@ -0,0 +1,189 @@
...
+        protected readonly ParameterGroupFacade $parameterGroupFacade,
+        protected readonly GridFactory $gridFactory,
+        protected readonly ParameterGroupDataFactoryInterface $parameterGroupDataFactory,
+    ) {
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     */
+    #[Route(path: '/product/parameter-group/list/')]
```
grossmannmartin, line 41: can you please add roles to allow limiting the privileges? 
You can find inspiration in this commit https://github.com/shopsys/shopsys/commit/7a24f93269a3156c5e6e958280fdddd28632897b

The goal of this is to be able to create an administrator (http://127.0.0.1:8000/admin/administrator/new/), select as a role `Parameter groups - full` or `Parameter groups - view`, and such administrator will be only able to manage the parameter groups.

We can discuss it further if you need.

packages/framework/src/Controller/Admin/ParameterGroupController.php
```diff
@@ -0,0 +1,189 @@
...
+
+        $this->addSuccessFlashTwig(
+            t('Parameter <strong>{{ name }}</strong> deleted'),
+            [
+                'name' => $fullName,
+            ],
+        );
+        //        } catch (ParameterNotFoundException $ex) {
+        //            $this->addErrorFlash(t('Selected parameter doesn\'t exist.'));
+        //        }
```
grossmannmartin: please remove commented code

packages/framework/src/Controller/Admin/ParameterGroupController.php
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \Shopsys\FrameworkBundle\Component\Grid\Grid
+     */
+    protected function getGrid(Request $request)
+    {
+        $queryBuilder = $this->parameterGroupFacade->getOrderedParameterGroupsQueryBuilder($request->getLocale());
+
+        $dataSource = new QueryBuilderDataSource($queryBuilder, 'pg.id');
+        //        dd($dataSource);
```
grossmannmartin: please remove commented code

packages/framework/src/Controller/Admin/ParameterGroupController.php
```diff
@@ -0,0 +1,189 @@
...
+        //        }
+
+        return $this->redirectToRoute('admin_parametergroup_list');
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \Shopsys\FrameworkBundle\Component\Grid\Grid
+     */
+    protected function getGrid(Request $request)
```
grossmannmartin: some methods in this class are type-hinted, while some are not. Can you please add types everywhere in this class?

In a new code, we should use types everywhere, except for entities and entity data objects.

packages/framework/src/Controller/Admin/ParameterGroupController.php
```diff
@@ -0,0 +1,189 @@
...
+    ) {
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     */
+    #[Route(path: '/product/parameter-group/list/')]
+    public function listAction(Request $request)
+    {
+        $grid = $this->getGrid($request);
```
grossmannmartin: I suggest not using `$request` to get locale, but using the `getLocale()` method of the `\Shopsys\FrameworkBundle\Component\Domain\Domain` class.

So the `getGrid()` method will accept no parameters, and the locale will be obtained inside it.

packages/framework/src/Model/Product/Parameter/ParameterGroupFacade.php
```diff
@@ -48,11 +52,82 @@ public function edit(int $parameterGroupId, ParameterGroupData $parameterGroupDa
         return $parameterGroup;
     }
 
+    /**
+     * @param int[][] $rowIds
+     */
+    public function saveOrdering(array $rowIds): void
+    {
+        foreach ($rowIds as $position => $rowId) {
+            $parameterGroup = $this->parameterRepository->getParameterGroupById($position);
```
grossmannmartin: shouldn't be here $rowId instead of $position?

packages/framework/src/Model/Product/Parameter/ParameterGroupFacade.php
```diff
@@ -48,11 +52,82 @@ public function edit(int $parameterGroupId, ParameterGroupData $parameterGroupDa
         return $parameterGroup;
     }
 
+    /**
+     * @param int[][] $rowIds
```
grossmannmartin: isn't it just `int[]`? maybe even better `array<int, int>`

packages/framework/src/Resources/views/Admin/Content/ParameterGroup/detail.html.twig
```diff
@@ -0,0 +1,17 @@
+{% extends '@ShopsysFramework/Admin/Layout/layoutWithPanel.html.twig' %}
+
+{% block title %}- {{ 'Parameter'|trans }}{% endblock %}
+{% block h1 %}{{ 'Parameter'|trans }}{% endblock %}
```
grossmannmartin: `Parameter group`?

packages/framework/src/Resources/views/Admin/Content/ParameterGroup/listGrid.html.twig
```diff
@@ -0,0 +1,18 @@
...
+    {{ parent() }}
+{% endblock %}
+
+{% block grid_no_data %}
+    {{ 'There is no parameter group yet.'|trans }}
+{% endblock %}
+
+{% block grid_action_cell_type_delete %}
+    {{ gridView.renderBlock('grid_action_cell', {actionColumn: actionColumn, row: row}) }}
+{% endblock %}
```
grossmannmartin: please configure your IDE to always save files with the ending line break

project-base/storefront/components/Pages/ProductDetail/ProductDetailTabs/ProductDetailTabs.tsx
```diff
@@ -40,30 +40,78 @@ export const ProductDetailTabs: FC<ProductDetailTabsProps> = ({ description, par
...
-                    <Table className="border-0 p-0 max-w-screen-lg mx-auto">
-                        {parameters.map((parameter) => (
-                            <Row
-                                key={parameter.uuid}
-                                className="bg-tableBackground odd:bg-tableBackgroundContrast border-none"
-                            >
-                                <Cell className="py-2 text-left text-sm font-bold uppercase leading-5">
-                                    {parameter.name}
-                                </Cell>
+                    {parameters
```
grossmannmartin: can you send this also to the frontend CR?

project-base/app/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -194,8 +195,10 @@ public function getProductParameterValuesDataByProducts(array $products, string
             ->leftJoin('pu.translations', 'put', Join::WITH, 'put.locale = :locale AND put.name IS NOT NULL')
             ->join('ppv.value', 'pv', Join::WITH, 'pv.locale = :locale')
             ->where('ppv.product IN (:products)')
-            ->orderBy('ordering_priority', 'DESC')
-            ->addOrderBy('parameter_name')
+            ->orderBy('parameter_id', 'ASC')
```
grossmannmartin: is the primary sorting by the parameter id correct?

grossmannmartin (changes requested):

project-base/app/src/DataFixtures/Demo/CategoryParameterDataFixture.php
```diff
@@ -5,13 +5,13 @@
 namespace App\DataFixtures\Demo;
 
 use App\Model\Category\Category;
-use App\Model\Product\Parameter\Parameter;
```
grossmannmartin, line 8: this PR should also have the upgrade notes for storefront and backend. 

use `./phing upgrade-generate` to generate templates and take a look at the other upgrade notes for inspiration

packages/framework/src/Controller/Admin/ParameterGroupController.php
AmpMVn: try/catch returned. I forgot to uncomment after testing

project-base/app/src/Model/Product/Parameter/ParameterRepository.php
AmpMVn: yes, it is useless here. I learned somewhere that in the case of continuing another order, I would preorder them by id. But maybe that's old school.

packages/framework/src/Controller/Admin/ParameterGroupController.php
```diff
@@ -21,22 +25,24 @@ class ParameterGroupController extends AdminBaseController
...
-     * @param \Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterGroupDataFactoryInterface $parameterGroupDataFactory
+     * @param \Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterGroupDataFactory $parameterGroupDataFactory
+     * @param \Shopsys\FrameworkBundle\Component\Domain\AdminDomainTabsFacade $adminDomainTabsFacade
      */
     public function __construct(
         protected readonly ParameterGroupFacade $parameterGroupFacade,
         protected readonly GridFactory $gridFactory,
-        protected readonly ParameterGroupDataFactoryInterface $parameterGroupDataFactory,
+        protected readonly ParameterGroupDataFactory $parameterGroupDataFactory,
+        protected readonly AdminDomainTabsFacade $adminDomainTabsFacade,
```
grossmannmartin: `AdminDomainTabsFacade` is commonly used with the domain tabs switcher - you can see it in action for example here http://127.0.0.1:8000/admin/article/list/
the idea is to show different data based on the selected domain. 

in this controller, the parameter groups are not dependent on the selected domain. 
I would use `\Shopsys\FrameworkBundle\Component\Domain\Domain::getLocale()`

packages/framework/src/Model/Product/Parameter/ParameterGroupFacade.php
```diff
@@ -53,11 +53,11 @@ public function edit(int $parameterGroupId, ParameterGroupData $parameterGroupDa
     }
 
     /**
-     * @param int[][] $rowIds
+     * @param array<int> $rowIds
      */
     public function saveOrdering(array $rowIds): void
     {
-        foreach ($rowIds as $position => $rowId) {
+        foreach ($rowIds as $position) {
```
grossmannmartin: should it be named `$position`?

project-base/app/config/packages/security.yaml
```diff
@@ -99,6 +100,7 @@ security:
             - ROLE_COUNTRY_VIEW
             - ROLE_STORE_VIEW
             - ROLE_PARAMETER_VALUE_VIEW
+            - ROLE_PARAMETER_GROUP_VIEW
             - ROLE_SEO_VIEW
             - ROLE_CATEGORY_SEO_VIEW
             - ROLE_FRIENDLY_URL_VIEW
```
grossmannmartin, line 106: please add also hierarchy 

```
ROLE_PARAMETER_GROUP_FULL: [ROLE_PARAMETER_GROUP_VIEW]
```

see the lines below where it is used for other roles

upgrade-notes/backend_20241009_135052.md
```diff
@@ -0,0 +1,4 @@
+#### Parameter groups refactor + editable in Admin ([#3484](https://github.com/shopsys/shopsys/pull/3484))
+
+-   removal of the deprecated visible attribute
+-   see #project-base-diff to update your project
```
grossmannmartin: according to upgrade notes. we're in the middle of thinking how the upgrade notes should be written... 

currently, I think we should: 
- heading should be in form `do this`
- mention critical changes in logic (this does not apply to this PR) 
-  mention movements of features from project-base to packages - see https://github.com/shopsys/shopsys/blob/bdc02ab7cb325942af13479d23bcd80e9c976598/upgrade-notes/backend_20241007_081817.md for inspiration
- mention changes in FE API

so, according to these claims, I suggest something like 

```
#### make parameter groups editable in admin ([#3484](https://github.com/shopsys/shopsys/pull/3484))

- 
- [features moved](#movement-of-features-from-project-base-to-packages) from project-base to the framework package:
    - Parameter groups
- field `visible` was removed from GraphQL field `Parameter`
```

or something like that

grossmannmartin (changes requested):

grossmannmartin (approved):

project-base/storefront/components/Pages/ProductDetail/ProductDetailTabs/ProductDetailTabs.tsx
```diff
@@ -40,30 +40,78 @@ export const ProductDetailTabs: FC<ProductDetailTabsProps> = ({ description, par
...
+                        .filter(parameter => parameter.group === null)
+                        .sort((a, b) => a.name.localeCompare(b.name))
+                        .length > 0 && (
```
JanMolcik: Please avoid complex logic inside the `return` statement. I would extract this above the `return` and give it a descriptive name such as `sortedIndividualParameters`. This would also help optimise the code as you are doing the same filtering and sorting below again to map each parameter to `Row`.

project-base/storefront/components/Pages/ProductDetail/ProductDetailTabs/ProductDetailTabs.tsx
```diff
@@ -40,30 +40,78 @@ export const ProductDetailTabs: FC<ProductDetailTabsProps> = ({ description, par
...
+                    {Object.entries(
+                        parameters.reduce((acc, parameter) => {
+                            if (parameter.group) {
+                                if (!acc[parameter.group]) {
+                                    acc[parameter.group] = [];
+                                }
+                                acc[parameter.group].push(parameter);
+                            }
+                            return acc;
+                        }, {})
```
JanMolcik: The same applies here. Please move complex logic to the body of the component instead. Also, did you setup linter in your IDE correctly? I can see bunch of type errors when checking your code locally 
image

project-base/storefront/components/Pages/ProductDetail/ProductDetailTabs/ProductDetailTabs.tsx
```diff
@@ -40,30 +40,78 @@ export const ProductDetailTabs: FC<ProductDetailTabsProps> = ({ description, par
...
+                    ).map(([groupName, groupParameters]) => (
+                        <div key={groupName}>
+                            <h2 className="text-lg font-bold max-w-screen-lg mx-auto my-4">{groupName}</h2>
+                            <Table className="border-0 p-0 max-w-screen-lg mx-auto">
+                                {groupParameters
+                                    .sort((a, b) => a.name.localeCompare(b.name))
```
JanMolcik: This sorting can also be moved into variable and named as `sortedGroupParameters` for example.

JanMolcik (changes requested):

JanMolcik (approved):

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -249,7 +249,9 @@ protected function getProductParameterValuesByProductQueryBuilder(Product $produ
         return $this->em->createQueryBuilder()
             ->select('ppv')
             ->from(ProductParameterValue::class, 'ppv')
+            ->join(ParameterGroup::class, 'pg')
```
grossmannmartin: I have two thoughts about this.
1. The `getProductParameterValuesByProductQueryBuilder` method is overridden in project-base without calling the parent method. This change probably doesn't change any behavior at all. – I suggest moving the method from project-base here and making changes here.
2. ParameterGroup does not have any relation with ProductParameterValue, so I'm not sure how this will behave.

grossmannmartin (changes requested):

packages/framework/src/Model/Product/Parameter/ParameterRepository.php
```diff
@@ -242,17 +242,36 @@ public function getParameterValueByValueTextAndLocale(string $valueText, string
...
+        Product $product,
+        ?string $locale = null,
+    ): QueryBuilder {
+        $queryBuild = $this->em->createQueryBuilder()
             ->select('ppv')
             ->from(ProductParameterValue::class, 'ppv')
-            ->join(ParameterGroup::class, 'pg')
+            ->join('ppv.parameter', 'p')
+            ->join('ppv.value', 'pv')
+            ->leftJoin(ParameterGroup::class, 'pg', Join::WITH, 'p.group = pg.id')
```
grossmannmartin: cannot we use this?
```
->leftJoin('p.group', 'pg')
```

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -1520,7 +1520,7 @@ msgid "Editing parameter"
 msgstr "Editace parametru"
 
 msgid "Editing parameter group"
-msgstr ""
+msgstr "Editace skupiny paramtrů"
```
grossmannmartin, line 1526: msgstr "Editace skupiny parametrů"

grossmannmartin (approved):

packages/framework/src/Model/Product/Parameter/ParameterGroupFacade.php
```diff
@@ -118,6 +129,9 @@ public function deleteById($parameterGroupId)
 
         $this->dispatchParameterGroupEvent($parameterGroup, ParameterGroupEvent::DELETE);
 
+        $affectedProductIds = $this->productFacade->getIdsByParameterGroupId($parameterGroupId);
+        $this->productRecalculationDispatcher->dispatchProductIds($affectedProductIds, ProductRecalculationPriorityEnum::REGULAR, [ProductExportScopeConfig::SCOPE_PARAMETERS]);
```
grossmannmartin: Can you move the whole exporting to the `DispatchAffectedProductsSubscriber`? 
So in the create/edit/delete you just dispatch the parameter group event (like it's on line 130), and the heavy lifting will do the subscriber.

grossmannmartin (approved):

grossmannmartin (approved):

PR #3497, opened 2024-10-11
[framework] SSP-2718 - extend CustomerFileNotFoundException for frontend 404 page

AmpMVn:
CustomerFileNotFoundException returned a backend 500 error when the image was not found. Now, if the image is not found, it returns the frontend 404 page in production environment. In the development environment returned Symfony exception.

- [ ] New feature 

----
  - https://mvn-ssp-2718-invalid-error-exception-customer-files.odin.shopsys.cloud
  - https://cz.mvn-ssp-2718-invalid-error-exception-customer-files.odin.shopsys.cloud


vitek-rostislav (changes requested):
Hi, I have a few things, it seems like it does not work this way, and in order to make it work, there will be some maintenance needed on the current code :innocent: :smile: we can arrange a call if needed :wink: 
1. so, to make it work in the production mode, the new exception headers need to be forwarded in `ErrorController::showAction()` to `createProductionResponse()` (so it can be returned in the `Response` object)
2. I noticed we have two `ErrorController` classes in the project, so please, remove the one from `project-base`, and move its logic into the one in the `framework` package (please, add an "features moved" entry to the upgrade notes)
3. you will need to modify `error_controller` setting in `project-base/app/config/packages/framework.yaml` accordingly then
4. [this might be extracted into a separate issue] the same problem described for the customer files is there also for regular files so I think we should solve it similarly there as well. That means, it would be nice to make that "not found exception with x-accel-redirect headers" throwing re-usable.

vitek-rostislav (changes requested):
Hi, thanks for the fixes, there are a few more things to fix now (see the failing builds):
- you need to dump translations (after moving the templates, the translation messages should be moved from project-base to the framework po files
- `project-base/app/templates/Admin/Layout/errorLayout.html.twig` should be moved to the framework package as well and the `@ShopsysFramework` notation should be used in the `extend` part of the moved error templates, this is probably why the smoke tests are failing -  there is a 500 error now on `admin_superadmin_errors` route

AmpMVn:
> * `project-base/app/templates/Admin/Layout/errorLayout.html.twig` should be moved to the framework package as well and the `@ShopsysFramework` notation should be used in the `extend` part of the moved error templates, this is probably why the smoke tests are failing -  there is a 500 error now on `admin_superadmin_errors` route

Strange. when I try it manually, I get the correct 403. but smoke tests say error.

PR #3508, opened 2024-10-16
[shopsys] frontend API testing on b2b domain

vitek-rostislav:
We needed a solution for testing other than the 1st domain to be able to test the B2B-domain-specific behavior. This PR solves this with two new test cases:
- `GraphQlB2bDomainTestCase`
- `GraphQlB2bDomainWithLoginTestCase`

There are also added tests for the B2B-domain-specific behavior, namely for the following FE API endpoints:
- `Register` mutation
- `ChangePersonalData` mutation
- `AddNewCustomerUser` mutation
- `EditCustomerUserPersonalData` mutation
- `RemoveCustomerUser` mutation
- `CustomerUsersQuery` query

#### Fixes issues
... 

- [x] New feature 

----
  - https://rv-tests-second-domain.odin.shopsys.cloud
  - https://cz.rv-tests-second-domain.odin.shopsys.cloud


packages/framework/build.xml
```diff
@@ -1470,6 +1470,34 @@
...
+                    <arg value="--testsuite"/>
+                    <arg value="FrontendApiFunctionalB2b"/>
+                    <arg value="--exclude-group"/>
+                    <arg value="multidomain"/>
+                    <arg value="--configuration"/>
+                    <arg value="${path.root}/phpunit.xml"/>
+                </exec>
+            </else>
+        </if>
+    </target>
```
grossmannmartin: I'm thinking, maybe we could write it in a way that reduces dry?

like 
```xml
<if>
    <istrue value="${domains-info.is-multidomain}"/>
    <then>
        <property name="excludeGroup" value="singledomain"/>
    </then>
    <else>
        <property name="excludeGroup" value="multidomain"/>
    </else>
</if>
<exec executable="${path.phpunit.executable}" logoutput="true" passthru="true" checkreturn="true">
    <arg value="--colors=always"/>
    <arg value="--testsuite"/>
    <arg value="FrontendApiFunctionalB2b"/>
    <arg value="--exclude-group"/>
    <arg value="${excludeGroup}"/>
    <arg value="--configuration"/>
    <arg value="${path.root}/phpunit.xml"/>
</exec>
```

what do you think?

project-base/app/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
```diff
@@ -133,4 +147,20 @@ public static function getRegisterQueryVariables(
...
         ];
     }
+
+    /**
+     * @param array $response
+     */
+    private function assertSuccessfulRegistration(array $response): void
+    {
+        $graphQlType = 'Register';
+        $this->assertResponseContainsArrayOfDataForGraphQlType($response, $graphQlType);
```
grossmannmartin: this assertion is not quite necessary. the same does the `getResponseDataForGraphQlType` internally

project-base/app/tests/FrontendApiBundle/Test/CommonGraphQlWithLoginTestCase.php
```diff
@@ -0,0 +1,75 @@
...
+     * @return string
+     */
+    private static function getLoginQuery(
+        ?string $customerUserEmail = null,
+        ?string $customerUserPassword = null,
+    ): string {
+        $customerUserEmail = $customerUserEmail ?? self::DEFAULT_USER_EMAIL;
+        $customerUserPassword = $customerUserPassword ?? self::DEFAULT_USER_PASSWORD;
+
+        return '
```
grossmannmartin: nitpick. we may use `project-base/app/tests/FrontendApiBundle/Functional/Login/graphql/LoginMutation.graphql`

project-base/app/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
```diff
@@ -43,6 +43,17 @@ public function testUniqueBillingAddressIsNotValidatedInB2cRegistration(): void
         $this->assertSuccessfulRegistration($response);
     }
 
+    public function testCompanyUserRegistration(): void
+    {
+        $registerQueryVariables = self::getRegisterQueryVariables();
+        $registerQueryVariables['companyCustomer'] = true;
+        $registerQueryVariables['companyNumber'] = '999';
```
grossmannmartin, line 50: this company number will pass the validation? I thought it has to be 8 digits

grossmannmartin (changes requested):

grossmannmartin (approved):

project-base/app/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
vitek-rostislav: seems like there is no such validation on the API side

https://github.com/shopsys/shopsys/blob/16.0/packages/frontend-api/src/Resources/config/graphql-types/ModelType/Customer/BillingAddress/Input/CompanyInputObjectDecorator.types.yaml#L23

project-base/app/tests/FrontendApiBundle/Test/CommonGraphQlWithLoginTestCase.php
vitek-rostislav: good point, fixed in 120747299f

project-base/app/tests/FrontendApiBundle/Functional/Customer/User/RegisterTest.php
vitek-rostislav: :+1: 94bd1a112a

packages/framework/build.xml
vitek-rostislav: nice, what about this? 0b5eba0984 I was thinking about removing the `domains-info.is-multidomain` completely as it was used just for this purpose, however, it feels quite general and removing could be an unnecessary BC break (however, YAGNI? :shrug: :smile: )

vitek-rostislav: nice, what about this? 0b5eba0984 I was thinking about removing the `domains-info.is-multidomain` completely as it was used just for this purpose, however, it feels quite general and removing could be an unnecessary BC break (however, YAGNI? :shrug: :smile: )

PR #3514, opened 2024-10-17
An attempt to speed up the build of the application, including tests

AmpMVn:
Tests and overall build takes a very long time. I am trying to speed up the system using parallel tests and using cache. Regenerate screenshots does not mark action as fail.

- [ ] New feature 

----
  - https://mvn-ssp-2821-github-actions-revision.odin.shopsys.cloud
  - https://cz.mvn-ssp-2821-github-actions-revision.odin.shopsys.cloud


project-base/storefront/cypress/e2e/dummy/dummyTest.cy.ts
```diff
@@ -0,0 +1,13 @@
+import { initializePersistStoreInLocalStorageToDefaultValues, takeSnapshotAndCompare } from '../../support';
+
+describe('Dummy Test for blank others group visit tests with screenshots', () => {
```
grossmannmartin, line 3: Why is this test necessary?

.github/workflows/docker-build.yaml
```diff
@@ -125,14 +125,22 @@ jobs:
...
         runs-on: ubuntu-22.04
         needs: variables
         permissions:
             contents: read
             packages: write
         steps:
+            -   name: Cache Docker layers
+                uses: actions/cache@v3
+                with:
+                    path: /tmp/.buildx-cache
```
grossmannmartin: how should this cache work? more specifically, in what case the cache will be used? 

- on consecutive runs it should not (otherwise, the changes in the code will not be taken into account)
- for another jobs in the workflow we use the already built image, so I'm not sure about the benefit of the cache


also this warning in the output seems like the caching does not work

Snímek obrazovky 2024-10-30 v 12 43 03

.github/workflows/docker-build.yaml
```diff
@@ -508,27 +589,31 @@ jobs:
                     mv docker-compose.cypress.yml docker-compose.yml
                     docker compose up -d --wait
                     docker compose exec php-fpm php phing -D production.confirm.action=y db-create frontend-api-generate-new-keys build-demo-dev-quick error-pages-generate elasticsearch-index-recreate elasticsearch-export
-            -   name: Run Cypress tests
+            -   name: Run regenerate Cypress tests for ${{ matrix.group }}
```
grossmannmartin, line 533: when some test fails, the others are skipped (see https://github.com/shopsys/shopsys/actions/runs/11591269071/job/32270835776?pr=3542) 
How will this behave when regenerating screenshots?

.github/workflows/docker-build.yaml
```diff
@@ -125,14 +125,22 @@ jobs:
...
         needs: variables
         permissions:
             contents: read
             packages: write
         steps:
+            -   name: Cache Docker layers
+                uses: actions/cache@v3
+                with:
+                    path: /tmp/.buildx-cache
+                    key: ${{ runner.os }}-docker-php-fpm-${{ hashFiles('project-base/app/docker/php-fpm/Dockerfile') }}-${{ hashFiles('composer.json') }}
```
grossmannmartin: the cache key doesn't seem right

Snímek obrazovky 2024-10-30 v 12 12 33

.github/workflows/docker-build.yaml
```diff
@@ -244,7 +314,8 @@ jobs:
                 github.event.pull_request.head.repo.full_name == 'shopsys/shopsys' ||
                 github.ref_protected == true
             ) &&
-            (needs.build-php-fpm-image.result == 'success' || needs.build-php-fpm-image.result == 'skipped')
+            (needs.build-php-fpm-image.result == 'success' || needs.build-php-fpm-image.result == 'skipped') &&
+            !contains(github.event.pull_request.labels.*.name, 'regenerate screenshots')
```
grossmannmartin: this was not done deliberately for this scenario:
- you push changes
- screenshots fail
- you add a label
- you rerun the already created workflow

because the labels are evaluated at the time when the workflow is created, not when it is run, the screenshots will not be regenerated.

grossmannmartin (changes requested):

project-base/storefront/cypress/e2e/dummy/dummyTest.cy.ts
AmpMVn: For tests that are not in any group.

.github/workflows/docker-build.yaml
```diff
@@ -353,6 +354,8 @@ jobs:
                 uses: actions/download-artifact@v4
                 with:
                     name: docker-compose
+            -   name: Pull Docker images
+                run: docker compose pull --parallel postgres webserver elasticsearch redis php-fpm
```
grossmannmartin: the same pull is run in the following step, is it ok?

grossmannmartin (approved):

PR #3532, opened 2024-10-23
[shopsys] Do not log out the user after changing the password

AmpMVn:
Every time the password was changed, all tokens were deleted and the user was logged out. I modified the method so that the excludedDeviceId parameter is optional, which is not deleted from the tokens and the user remains logged in after changing their password.

- [ ] New feature 

----
  - https://mvn-ssp-2714-customer-user-change-password.odin.shopsys.cloud
  - https://cz.mvn-ssp-2714-customer-user-change-password.odin.shopsys.cloud


packages/frontend-api/src/Model/Mutation/Customer/User/CustomerUserMutation.php
```diff
@@ -96,6 +96,8 @@ public function __construct(
      */
     public function changePasswordMutation(Argument $argument, InputValidator $validator): CustomerUser
     {
+        $user = $this->runCheckUserIsLogged();
+
         $this->runCheckUserIsLogged();
```
TomasLudvik: This is redundant now

TomasLudvik (changes requested):

TomasLudvik (approved):

pk16011990 (approved):

PR #3536, opened 2024-10-25
Convertim integration

TomasLudvik:
Convertim checkout (https://www.convertim.com/) is part of Shopsys family and it was about time to finally implement it into our platform.

#### Fixes issues
... 

- [x] New feature 

----
  - https://tl-convertim-backend.odin.shopsys.cloud
  - https://cz.tl-convertim-backend.odin.shopsys.cloud


packages/convertim/src/Model/Convertim/ConvertimConfigProvider.php
```diff
@@ -0,0 +1,69 @@
...
+namespace Shopsys\ConvertimBundle\Model\Convertim;
+
+use InvalidArgumentException;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class ConvertimConfigProvider
+{
+    public const string ENABLED = 'enabled';
+    public const string AUTHORIZATION_HEADER = 'authorizationHeader';
```
vitek-rostislav: both the constants can be protected

packages/convertim/src/Model/Convertim/ConvertimConfig.php
```diff
@@ -0,0 +1,34 @@
...
+
+class ConvertimConfig
+{
+    /**
+     * @param bool $isEnabled
+     * @param string $authorizationHeader
+     */
+    public function __construct(
+        protected readonly bool $isEnabled,
+        protected readonly string $authorizationHeader,
```
vitek-rostislav: nitpick, you can make these properties `public readonly` and get rid of the getters :slightly_smiling_face:

packages/convertim/src/Model/Convertim/ConvertimConfigProvider.php
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @var \Shopsys\ConvertimBundle\Model\Convertim\ConvertimConfig[] array
```
vitek-rostislav:
```suggestion
     * @var \Shopsys\ConvertimBundle\Model\Convertim\ConvertimConfig[]
```

vitek-rostislav:
```suggestion
     * @var \Shopsys\ConvertimBundle\Model\Convertim\ConvertimConfig[]
```

packages/convertim/manifest.json
```diff
@@ -0,0 +1,32 @@
...
+        "CONVERTIM_CONFIG": {
+            "1": {
+                "enabled": false,
+                "url": "",
+                "authorizationHeader": "",
+                "ab_test_type": "disabled",
+                "projectName": "shopsys",
+                "isDebug": true,
+                "clientId": "",
+                "clientSecret": ""
```
vitek-rostislav: I am a little bit confused with this - it seems like you can set all these variables in the config, but on the other hand, `ConvertimConfig` class actually works with `enabled` and `authorizationHeader` properties only. I would either remove the other settings from here, or extend the application so it works with all of them.

packages/convertim/src/Resources/config/services.yaml
```diff
@@ -11,3 +11,13 @@ services:
...
+
+    Shopsys\ConvertimBundle\Model\Convertim\ConvertimConfigProvider:
+        calls:
+            -   method: setConfigs
+                arguments:
+                    - '%convertim.config%'
+
+    Shopsys\ConvertimBundle\Model\Convertim\ConvertimLogger:
+        arguments:
+            - '@monolog.logger.cron'
```
vitek-rostislav: I would rather create a separate channel for convertim

vitek-rostislav: I would rather create a separate channel for convertim

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+        $lastOrder = count($lastOrders) > 0 ? reset($lastOrders) : null;
+
+        $billingAddress = $customerUser->getCustomer()->getBillingAddress();
+        $deliveryAddresses = $customerUser->getCustomer()->getDeliveryAddresses();
+
+        if (count($deliveryAddresses) > 0) {
+            /** @var \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddress $lastDeliveryAddress */
+            $lastDeliveryAddress = end($deliveryAddresses);
+
+            $deliveryAddress = $this->createDeliveryAddressFromDeliveryAddress($lastDeliveryAddress, $customerUser);
```
vitek-rostislav: this delivery address -> delivery address (same for the billing address) naming is quite confusing, I suggest using the term "convertimDeliveryAddress" for the variables and in the method names to distinguish them.

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        return new ConvertimBillingAddress(
+            (string)$billingAddress->getId(),
+            $customerUser->getFirstName(),
+            $customerUser->getLastName(),
+            $billingAddress->getStreet(),
+            $billingAddress->getCity(),
+            $billingAddress->getPostcode(),
+            $billingAddress->getCountry()->getCode(),
```
vitek-rostislav:
```suggestion
            $billingAddress->getCountry()?->getCode(),
```
country is nullable

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+        CustomerUser $customerUser,
+    ): ConvertimDeliveryAddress {
+        return new ConvertimDeliveryAddress(
+            (string)$lastDeliveryAddress->getId(),
+            $customerUser->getFirstName(),
+            $customerUser->getLastName(),
+            $lastDeliveryAddress->getStreet(),
+            $lastDeliveryAddress->getCity(),
+            $lastDeliveryAddress->getPostcode(),
+            $lastDeliveryAddress->getCountry()->getCode(),
```
vitek-rostislav:
```suggestion
            $lastDeliveryAddress->getCountry()?->getCode(),
```
country is nullable

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+            /** @var \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddress $lastDeliveryAddress */
+            $lastDeliveryAddress = end($deliveryAddresses);
+
+            $deliveryAddress = $this->createDeliveryAddressFromDeliveryAddress($lastDeliveryAddress, $customerUser);
+
+            if ($telephone === null) {
+                $telephone = $lastDeliveryAddress->getTelephone();
+            }
+        } else {
+            if ($billingAddress->getStreet() === null) {
```
vitek-rostislav: This is the second (and later in this file the third) occurrence in the code where we check the nullability of the billing address street to determine whether the billing address is filled. (The first one is in `CustomerUserUpdateDataFactory`.) I suggest creating something like `BillingAddress::isAddressFilled()` or `BillingAddress::isNotEmpty()` to avoid the confusion and duplication

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        return new CustomerDetail(
+            $customerUser->getEmail(),
+            $telephone ?: '',
+            $deliveryAddress,
+            $this->createBillingAddress($billingAddress, $customerUser),
+            $lastOrder?->getPayment()->getUuid(),
+            $lastOrder?->getTransport()->getUuid(),
+            $this->createLastSelectedPickupPoint($lastOrder),
```
vitek-rostislav: hypothetical scenario - in the past, the customer created two orders:
- first, an order with a pickup point (let's say Packetery)
- second, another order with another type of delivery (let's say Czech post)
 
Then, the user logged into Convertim where he saw the data from the last order (Czech post), nice.
Now, when he changes the delivery to Packetery in the current cart, isn't it suitable to fill in the last selected Packetery pickup place - the place that he selected in his second last order in the past?

My point is, should be the "last selected pickup point" send to Convertim from the last order only, or should we try to find the last pickup point in all his previous orders?

Could we discuss that, please? :smile:

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+            $lastDeliveryAddress = end($deliveryAddresses);
+
+            $deliveryAddress = $this->createDeliveryAddressFromDeliveryAddress($lastDeliveryAddress, $customerUser);
+
+            if ($telephone === null) {
+                $telephone = $lastDeliveryAddress->getTelephone();
+            }
+        } else {
+            if ($billingAddress->getStreet() === null) {
+                throw new CustomerDetailsNotFoundException(['customerUserUuid' => $userUuid]);
```
vitek-rostislav: maybe we can be more verbose here? The customer user exists but has an empty billing address so I would like to know have this information in the log. I suggest extending the context, what about this?

```
throw new CustomerDetailsNotFoundException([
    'customerUserUuid' => $userUuid,
    'additionalInfo' => 'Customer user billing address is empty',
]);
```

packages/convertim/src/Model/Order/OrderRepository.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\ConvertimBundle\Model\Order;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Shopsys\FrameworkBundle\Model\Order\Order;
+
+class OrderRepository
```
vitek-rostislav, line 11: I would rather add `findLastOrderByEmailAndDomainId` to the existing `OrderRepository` in the framework package to avoid having multiple order repositories, but I already know your arguments against this approach :smile: Let's discuss it, please :pray:

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -171,4 +177,109 @@ protected function createDeliveryAddressFromBillingAddress(
...
+            $order = $this->orderFacade->getByUuid($orderUuid);
+
+            $billingAddress = null;
+
+            if ($order->isDeliveryAddressSameAsBillingAddress()) {
+                $deliveryAddress = $this->createDeliveryAddressFromOrderBillingAddress($order);
+            } else {
+                $deliveryAddress = $this->createDeliveryAddressFromOrderDeliveryAddress($order);
+
+                if ($order->getStreet()) {
```
vitek-rostislav: this is unnecessary condition, order street is not nullable and should be always filled

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+    protected function createBillingAddress(
+        ?BillingAddress $billingAddress,
+        CustomerUser $customerUser,
+    ): ?ConvertimBillingAddress {
+        if ($billingAddress === null || $billingAddress->getStreet() === null) {
+            return null;
+        }
+
+        return new ConvertimBillingAddress(
+            (string)$billingAddress->getId(),
```
vitek-rostislav: why don't we use our uuid here?

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
```diff
@@ -0,0 +1,174 @@
...
+     * @param \Shopsys\FrameworkBundle\Model\Customer\DeliveryAddress $lastDeliveryAddress
+     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CustomerUser $customerUser
+     * @return \Convertim\Customer\DeliveryAddress
+     */
+    protected function createDeliveryAddressFromDeliveryAddress(
+        DeliveryAddress $lastDeliveryAddress,
+        CustomerUser $customerUser,
+    ): ConvertimDeliveryAddress {
+        return new ConvertimDeliveryAddress(
+            (string)$lastDeliveryAddress->getId(),
```
vitek-rostislav: why don't we use our uuid here?

packages/convertim/src/Controller/OrderController.php
```diff
@@ -48,4 +52,27 @@ public function getLastOrderDetail(Request $request): Response
...
+    #[Route('/check-order')]
+    public function checkOrder(Request $request): Response
+    {
+        if ($this->isProtectedRequest($request) === false) {
+            return $this->invalidAuthorizationResponse();
+        }
+
+        try {
+            $rawData = json_decode($request->getContent(), true, 512, JSON_THROW_ON_ERROR);
+            $convertimOrderDataFactory = new ConvertimOrderDataFactory();
```
vitek-rostislav, line 84: the factory is not service so it is not possible to autowire it? :slightly_frowning_face:

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
```diff
@@ -0,0 +1,118 @@
...
+use Shopsys\FrameworkBundle\Model\Cart\Cart;
+use Shopsys\FrameworkBundle\Model\Cart\CartFacade;
+use Shopsys\FrameworkBundle\Model\Cart\Payment\CartPaymentDataFactory;
+use Shopsys\FrameworkBundle\Model\Cart\Transport\CartTransportDataFactory;
+use Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserFacade;
+use Shopsys\FrameworkBundle\Model\Order\PromoCode\PromoCodeFacade;
+use Shopsys\FrameworkBundle\Model\Payment\PaymentFacade;
+use Shopsys\FrameworkBundle\Model\Product\ProductFacade;
+use Shopsys\FrameworkBundle\Model\Transport\TransportFacade;
+use Shopsys\FrontendApiBundle\Model\Cart\CartApiFacade;
```
vitek-rostislav, line 18: you need to add `shopsys/frontend-api` dependency to convertim's `composer.json` if you need to use its services  here

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
```diff
@@ -0,0 +1,118 @@
...
+     * @return \Shopsys\FrameworkBundle\Model\Cart\Cart
+     */
+    public function mapConvertimOrderDataToCart(ConvertimOrderData $convertimOrderData): Cart
+    {
+        $customerUser = $this->customerUserFacade->findCustomerUserByEmailAndDomain(
+            $convertimOrderData->getCustomerData()->getEmail(),
+            $this->domain->getId(),
+        );
+
+        $cart = $this->cartApiFacade->getCartCreateIfNotExists(null, null);
```
vitek-rostislav, line 52: are the two `null` values intended? Why don't you pass `$customerUser` there?

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
```diff
@@ -0,0 +1,118 @@
...
+        $this->applyConvertimTransportToCart($convertimOrderData, $cart);
+        $cart->setPaymentWatchedPrice(Money::create($convertimOrderData->getPaymentData()->getPriceWithVat()));
+        $cart->setTransportWatchedPrice(Money::create($convertimOrderData->getTransportData()->getPriceWithVat()));
+        $cart->assignCartToCustomerUser($customerUser);
+
+        return $cart;
+    }
+
+    /**
+     * @param array $convertimOrderPromoCodesData
```
vitek-rostislav:
```suggestion
     * @param \Convertim\Order\ConvertimOrderPromoCodesData[] $convertimOrderPromoCodesData
```

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * @param \Convertim\Order\ConvertimOrderData $convertimOrderData
+     * @param \Shopsys\FrameworkBundle\Model\Cart\Cart $cart
+     */
+    protected function applyConvertimTransportToCart(ConvertimOrderData $convertimOrderData, Cart $cart): void
+    {
+        $transport = $this->transportFacade->getByUuid($convertimOrderData->getTransportData()->getUuid());
+        $cartTransportData = $this->cartTransportDataFactory->create($cart, $transport->getUuid(), null);
```
vitek-rostislav: there is `ConvertimOrderTransportData::getExtra()` method available that returns `ConvertimOrderTransportExtraData` instance. The class provides information about the pickup place, should not we take this into account and possibly extract the pickup place identifier here?

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * @param \Convertim\Order\ConvertimOrderData $convertimOrderData
+     * @param \Shopsys\FrameworkBundle\Model\Cart\Cart $cart
+     */
+    protected function applyConvertimPaymentToCart(ConvertimOrderData $convertimOrderData, Cart $cart): void
+    {
+        $payment = $this->paymentFacade->getByUuid($convertimOrderData->getPaymentData()->getUuid());
+        $cartPaymentData = $this->cartPaymentDataFactory->create($cart, $payment->getUuid(), null);
```
vitek-rostislav, line 89: I am a little bit concerned about the `null` value for swift here, but it seems like we are not able to get it from convertim, right?

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
```diff
@@ -0,0 +1,118 @@
...
+            $product = $this->productFacade->getByUuid($orderItemData->getProductId());
+            $result = $this->cartFacade->addProductToExistingCart($product, (int)$orderItemData->getQuantity(), $cart, true);
+            $result->getCartItem()->setWatchedPrice(Money::create($orderItemData->getPriceWithVat()));
+        }
+
+        $this->applyConvertimPromoCodesToCart($convertimOrderData->getPromoCodes(), $cart);
+        $this->applyConvertimPaymentToCart($convertimOrderData, $cart);
+        $this->applyConvertimTransportToCart($convertimOrderData, $cart);
+        $cart->setPaymentWatchedPrice(Money::create($convertimOrderData->getPaymentData()->getPriceWithVat()));
+        $cart->setTransportWatchedPrice(Money::create($convertimOrderData->getTransportData()->getPriceWithVat()));
```
vitek-rostislav: it seems like these lines are unnecessary as the watched prices are set already in `applyConvertim(Payment|Transport)ToCart()` methods if I am not mistaken (`editCartPayment()` and `editCartTransport` are called there and the methods set the watched prices, right?)

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        throw new OrderItemNotFoundException(sprintf('Order item with uuid "%s" not found.', $uuid));
+    }
+
+    /**
+     * @param string|null $countryName
+     * @return \Shopsys\FrameworkBundle\Model\Country\Country
+     */
+    protected function getCountryByName(?string $countryName): Country
```
vitek-rostislav: this is quite complicated and still can end up with an error (when there is no country enabled on the domain) :thinking: Moreover, does Convertim really return the country name? Based on the examples in the docs, it seems like country ISO code, or?

I would rather add a new method to the country repository - `getByCode(string $countryCode)` and throw an exception when the country is not found

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
```diff
@@ -0,0 +1,190 @@
...
+        $transportOrderItemData->totalPriceWithVat = Money::create($convertimOrderTransportPaymentData->getPriceWithVat());
+        $transportOrderItemData->totalPriceWithoutVat = Money::create($convertimOrderTransportPaymentData->getPriceWithoutVat());
+        $transportOrderItemData->unitPriceWithVat = Money::create($convertimOrderTransportPaymentData->getPriceWithVat());
+        $transportOrderItemData->unitPriceWithoutVat = Money::create($convertimOrderTransportPaymentData->getPriceWithoutVat());
+
+        if ($convertimOrderTransportPaymentData->getVatRate() > 0) {
+            $transportOrderItemData->vatPercent = Money::create(100)->divide((Money::create($convertimOrderTransportPaymentData->getPriceWithoutVat())->divide($convertimOrderTransportPaymentData->getVatRate(), 6))->getAmount(), 6)->round(6)->getAmount();
+        } else {
+            $transportOrderItemData->vatPercent = '0';
+        }
```
vitek-rostislav: Why do we need to calculate it separately? Isn't the result of  `$convertimOrderTransportPaymentData->getVatRate()` return the number we need? 

From docs:
image

packages/convertim/src/Model/Order/OrderFacade.php
```diff
@@ -0,0 +1,46 @@
...
+        $orderData = $this->convertimOrderDataToOrderMapper->mapConvertimOrderDataToOrderData($convertimOrderData);
+
+        //        d($orderData);
+
+        $deliveryAddressUuid = $convertimOrderData->getCustomerData()->getConvertimCustomerDeliveryAddressData()->getUuid();
+
+
+        //        d($orderData->items);
+
+        $order = $this->placeOrderFacade->placeOrder($orderData, $deliveryAddressUuid);
```
vitek-rostislav: forgotten dumps :slightly_smiling_face: (or maybe they are here because this PR is still WIP. I am adding this note so we do not forget about it

packages/framework/src/Model/Order/OrderDataFactory.php
```diff
@@ -90,7 +90,7 @@ protected function fillFromOrder(OrderData $orderData, Order $order): void
         $orderData->companyName = $order->getCompanyName();
         $orderData->companyNumber = $order->getCompanyNumber();
         $orderData->companyTaxNumber = $order->getCompanyTaxNumber();
-        $orderData->isCompanyCustomer = $order->isCompanyCustomer();
+        $orderData->isCompanyCustomer = $order->getCompanyName() !== null && $order->getCompanyNumber() !== null;
```
vitek-rostislav: why? `Order::isCompanyCustomer()` is the same condition

packages/convertim/src/Controller/CustomerController.php
```diff
@@ -36,6 +36,10 @@ public function __construct(
     #[Route('/get-customer-details/{customerUuid}')]
     public function getCustomerDetail(Request $request): Response
     {
+        if (!$this->isConvertimEnabled()) {
+            return $this->convertimNotEnabledResponse();
+        }
+
         if ($this->isProtectedRequest($request) === false) {
             return $this->invalidAuthorizationResponse();
         }
```
vitek-rostislav: these two checks are now duplicated in each of the actions, what about creating an event listener that would perform the checks? Something like this: (code generated by chatgpt)

```
namespace App\EventListener;

use Symfony\Component\HttpKernel\Event\ControllerEvent;
use Symfony\Component\HttpKernel\KernelEvents;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use App\Controller\AbstractBaseController;

class ControllerCheckListener implements EventSubscriberInterface
{
    public static function getSubscribedEvents()
    {
        return [
            KernelEvents::CONTROLLER => 'onController',
        ];
    }

    public function onController(ControllerEvent $event)
    {
        $controller = $event->getController();

        // Only handle controller as an array (controller class with action method)
        if (is_array($controller) && $controller[0] instanceof AbstractBaseController) {
            $controller[0]->performCheck();
        }
    }
}
```

I am of course open to any other approaches to avoid the DRY violations

vitek-rostislav (changes requested):

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
```diff
@@ -0,0 +1,192 @@
...
+
+        throw new OrderItemNotFoundException(sprintf('Order item with uuid "%s" not found.', $uuid));
+    }
+
+    /**
+     * @param string|null $countryName
+     * @return \Shopsys\FrameworkBundle\Model\Country\Country
+     */
+    protected function getCountryByName(?string $countryName): Country
```
TomasLudvik: If there is no country, it will end with null (changed return typehint).

I have changed it to code. 

I do not want to throw exception as it would fail to save order, that is not the preferred way here.  So I would like to keep it as it is

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
TomasLudvik: I have tried to rename it to better way

packages/convertim/src/Model/Customer/CustomerDetailFactory.php
TomasLudvik: This should answer @pk16011990

packages/convertim/src/Model/Order/OrderRepository.php
TomasLudvik: After discussion we are leaving it as it is

packages/convertim/src/Controller/OrderController.php
TomasLudvik: Yes, the Convertim bundle is not Symfony bundle so it is not service.

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
TomasLudvik: It is only temporary for CartWatcher, it is not supposed to be persisted and is removed in few steps later.

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
TomasLudvik: I have reworked it as we discussed

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
TomasLudvik: I have solved this by getting Vat from entities

packages/convertim/src/Model/Order/ConvertimOrderDataToCartMapper.php
TomasLudvik: Convertim does not provide list of banks, but solves this directly in GoPay so it has no access to this information.

packages/convertim/src/Controller/CustomerController.php
TomasLudvik: We discussed it and we are keeping it as it is

packages/framework/src/Model/Customer/BillingAddress.php
```diff
@@ -230,4 +230,12 @@ public function getUuid()
     {
         return $this->uuid;
     }
+
+    /**
+     * @return bool
+     */
+    public function isBillingAddressFilled(): bool
+    {
+        return $this->street !== null && $this->city !== null && $this->postcode !== null && $this->country !== null;
```
vitek-rostislav, line 239: :green_heart:

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
```diff
@@ -175,9 +186,25 @@ protected function mapConvertimTransportDataToOrderItem(
...
+        $pickupPlaceIdentifier = $convertimOrderTransportData->getExtra()?->getPickUpPointCode();
+
+        if ($transport->isPersonalPickup()) {
+            try {
+                $store = $this->storeFacade->getByUuidAndDomainId(
+                    $pickupPlaceIdentifier,
+                    $this->domain->getId(),
+                );
+
+                $transportOrderItemData->name .= ' ' . $store->getName();
```
vitek-rostislav, line 232: Could you please somehow extract this logic so it is not duplicated with the code in `PersonalPickupPointMiddleware`? :pray:

vitek-rostislav (changes requested):

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
TomasLudvik: I have tried this and we have discussed it and it brought more problems than solved.

packages/convertim/src/Model/Product/ProductRepository.php
```diff
@@ -0,0 +1,59 @@
...
+            ->setParameter('uuids', $uuids)
+            ->setParameter('domainId', $this->domain->getId());
+
+        $products = $queryBuilder->getQuery()->getResult();
+
+        $productsByUuid = [];
+
+        foreach ($products as $product) {
+            $productsByUuid[$product->getUuid()] = $product;
+        }
```
vitek-rostislav: I believe you can use `->indexBy('p', 'p.uuid')` on the query builder so you would not have to do the indexing manually

project-base/storefront/components/Pages/Order/TransportAndPayment/transportAndPaymentUtils.tsx
```diff
@@ -360,3 +360,32 @@ export const useTransportAndPaymentPageNavigation = (validationMessages: Partial
...
+        }
+
+        if (daysUntilDelivery < 7) {
+            return t('Personal pickup in {{ count }} days', { count: daysUntilDelivery });
+        }
+
+        return t('Personal pickup in {{count}} weeks', {
+            count: Math.ceil(daysUntilDelivery / 7),
+        });
+    }
```
chlebektomas, line 390: just add new line please

chlebektomas (approved):

vitek-rostislav (approved):

project-base/storefront/components/Pages/Order/TransportAndPayment/transportAndPaymentUtils.tsx
JanMolcik: fixed

packages/convertim/src/Model/Order/ConvertimOrderDataToOrderDataMapper.php
```diff
@@ -301,6 +301,10 @@ protected function mapDiscounts(
         OrderItemData $orderItemData,
         OrderData $orderData,
     ): void {
+        if ($convertimOrderItemData->getDiscounts() === null) {
```
vitek-rostislav, line 316: I noticed the `$product` parameter is unused in the `mapDiscounts` method so it can be removed :wink:

packages/convertim/src/Model/Order/OrderFacade.php
```diff
@@ -29,9 +39,65 @@ public function __construct(
...
 
-        return $this->placeOrderFacade->placeOrder($orderData, $deliveryAddressUuid);
+        if ($order === null) {
+            $orderData = $this->convertimOrderDataToOrderMapper->mapConvertimOrderDataToOrderData($convertimOrderData);
+            $deliveryAddressUuid = $convertimOrderData->getCustomerData()->getConvertimCustomerDeliveryAddressData()->getUuid();
+
+            $order = $this->placeOrderFacade->placeOrder($orderData, $deliveryAddressUuid);
+        }
+
+        if ($order->getPayment()->isGoPay() && $convertimOrderData->getGoPayData() !== null) {
```
vitek-rostislav: nullability of `$convertimOrderData->getGoPayData()` is check inside `resolveGoPayStatus()` method so it might be removed from here?

vitek-rostislav (approved):

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Cart/CartItemDecorator.types.yaml
```diff
@@ -16,3 +16,6 @@ CartItemDecorator:
             discounts:
                 type: "[CartItemDiscount!]!"
                 description: "Discounts applied to the cart item"
+            vatPercent:
```
vitek-rostislav: how is this resolved please? I do not see the field in `CartItem` entity nor in `CartItemResolverMap`. Moreover, in `convertimUtils.tsx`, I see cart item vat is mapped as `vat: product.vat.percent,` so maybe `CartItemDecorator.vatPercent` field is unused and unnecessary in the end?

vitek-rostislav (changes requested):
As we discussed, I would like to avoid calling `VatFacade::get ById()` in `ProductArrayFieldMapper`. Besides that, I have just one thing :pray:

vitek-rostislav (approved):

vitek-rostislav (approved):

project-base/app/tests/FrontendApiBundle/Functional/Cart/AuthenticatedRemoveCartTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $newlyCreatedCart = $this->addTestingProductToNewCart($firstProductQuantity);
+        $secondProduct = $this->getReference(ProductDataFixture::PRODUCT_PREFIX . 72, Product::class);
+        $secondProductQuantity = 3;
+
+        $this->getResponseContentForGql(__DIR__ . '/../_graphql/mutation/AddToCartMutation.graphql', [
+            'cartUuid' => $newlyCreatedCart['uuid'],
+            'productUuid' => $secondProduct->getUuid(),
+            'quantity' => $secondProductQuantity,
+        ]);
+
```
vitek-rostislav, line 36: this is not necessary for the test, or?

project-base/app/tests/FrontendApiBundle/Functional/Cart/AuthenticatedRemoveCartTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $removeCartMutation = 'mutation {
+            RemoveCart(input: { cartUuid: null })
+        }';
```
vitek-rostislav: could you please move the mutation into a separate Gql file?

packages/frontend-api/src/Model/Mutation/Cart/CartMutation.php
```diff
@@ -135,4 +138,31 @@ public function addOrderItemsToCartMutation(
...
+            $cart = $this->cartApiFacade->findCart($customerUser, null);
+        }
+
+        if ($cart !== null) {
+            $this->cartFacade->deleteCart($cart);
+        }
+
+        return [
+            true,
+        ];
```
vitek-rostislav, line 166: Why do you return an array here? Is not it possible to return a boolean?

packages/frontend-api/src/Model/Mutation/Cart/CartMutation.php
```diff
@@ -135,4 +138,31 @@ public function addOrderItemsToCartMutation(
...
+    {
+        $validator->validate();
+        $input = $argument['input'];
+
+        if (array_key_exists('cartUuid', $input) && $input['cartUuid'] !== null) {
+            $cart = $this->cartApiFacade->findCart(null, $input['cartUuid']);
+        } else {
+            $customerUser = $this->currentCustomerUser->findCurrentCustomerUser();
+            $cart = $this->cartApiFacade->findCart($customerUser, null);
+        }
```
vitek-rostislav, line 158: How does it work when a logged customer calls this mutation with some `cartUuid`?

vitek-rostislav, line 158: How does it work when a logged customer calls this mutation with some `cartUuid`?

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Cart/CartInput/RemoveCartInputDecorator.types.yaml
```diff
@@ -0,0 +1,10 @@
+RemoveCartInputDecorator:
+    type: input-object
+    decorator: true
+    inherits:
+        - CartInputDecorator
+    config:
+        fields:
+            cartUuid:
+                type: "Uuid"
+                description: "Cart UUID to be removed"
```
vitek-rostislav: After clarifying the behavior,  I would like to extend this description - The API users should be aware of the correct usage - logged in users must not provide the uuid? Or they must pass `null` here? I do not like that, lets talk about it, please

project-base/app/tests/FrontendApiBundle/Functional/Cart/RemoveCartTest.php
```diff
@@ -0,0 +1,68 @@
...
+        $firstProductQuantity = 6;
+        $newlyCreatedCart = $this->addTestingProductToNewCart($firstProductQuantity);
+        $secondProduct = $this->getReference(ProductDataFixture::PRODUCT_PREFIX . 72, Product::class);
+        $secondProductQuantity = 3;
+
+        $this->getResponseContentForGql(__DIR__ . '/../_graphql/mutation/AddToCartMutation.graphql', [
+            'cartUuid' => $newlyCreatedCart['uuid'],
+            'productUuid' => $secondProduct->getUuid(),
+            'quantity' => $secondProductQuantity,
+        ]);
```
vitek-rostislav, line 35: Same as above, I do not think we need two products in the cart for purposes of this test

project-base/app/tests/FrontendApiBundle/Functional/Cart/RemoveCartTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $removeCartMutation = 'mutation {
+            RemoveCart(input: { cartUuid: "' . $newlyCreatedCart['uuid'] . '" })
```
vitek-rostislav: mutation in a Gql file, please

vitek-rostislav (changes requested):

vitek-rostislav (approved):

chlebektomas (approved):

chlebektomas (approved):

PR #3578, opened 2024-11-07
[Framework] Rename tv products for main and variants

AmpMVn:
In demodates, the main products were poorly distinguished from their variants. We renamed one TV to Sencor, the main variants were given the flag [M/H] and their variants were given the flag [V]

- [ ] New feature 

----
  - https://mvn-ssp-2282-change-main-secondary-variant-names.odin.shopsys.cloud
  - https://cz.mvn-ssp-2282-change-main-secondary-variant-names.odin.shopsys.cloud


project-base/app/tests/FrontendApiBundle/Functional/Product/Flag/FlagTest.php
```diff
@@ -118,9 +118,6 @@ public function testFlagByUuidFilteredByAnotherFlag(): void
             [
                 'name' => t('32" Philips 32PFL4308', [], Translator::DATA_FIXTURES_TRANSLATION_DOMAIN, $this->getFirstDomainLocale()),
             ],
-            [
-                'name' => t('32" Hyundai 32PFL4400', [], Translator::DATA_FIXTURES_TRANSLATION_DOMAIN, $this->getFirstDomainLocale()),
```
vitek-rostislav, line 122: this removal is quite surprising, there were no changes to the flags assignment, or? :thinking:

project-base/app/tests/FrontendApiBundle/Functional/Product/ProductVariantTest.php
```diff
@@ -56,7 +56,7 @@ public function testProductMainVariantResultData(): void
                     'variants' => [
                         // Variant 51,5" Hyundai 22HD44D is not sellable, so it's not present
```
vitek-rostislav: the variant is now renamed to `51,5" Sencor 22HD44D`, could you update the comment, please?

project-base/app/tests/FrontendApiBundle/Functional/Product/Flag/FlagTest.php
vitek-rostislav: Ah, I see, that was tricky :sweat_smile: So the product was renamed to `Television Hyundai [M]` so it is now alphabetically sorted "later" in the array but the query has a limit for the first 5 products so it "disappeared" from the array. 

Could you please tweak the test itself in a separate commit?
- I would modify the `$products` array so it is of a size of the given limit (5 items) - the rest is just confusing
- also, instead of using hardcoded flag IDs, I would use the persistent references (in the whole `FlagTest`, please)
image
- if we want to have it super tuned, we could also move the queries from strings into separate graphql files :smile: 

If you do not want to postpone this task with the refactorings, I am ok with that, just let me know, I will create a separate task for that :wink:

vitek-rostislav (changes requested):
I noticed there is still one variant under the "Television Hyundai [M]" left that was not renamed - it is missing the "[V]" postfix - please check the product with ID 74 (catnum 9176554)

image

vitek-rostislav:
Oh, and one more thing (actually, two more things), please, I noticed a weird translation for `msgid "36\" Hyundai 32PFL4308"` in data fixtures - for some reason, it is translated to Czech as `msgstr "36\" Hyundai 32PFL4360"` which is really weird and probably a mistake IMHO. It would be nice to unify the translation :pray: 

If I may be so bold, I would get rid of the "codes" in the variant names completely, I think they have no added value - when you look at the first set of variants (under "Television Philips [M]"), all of them have the same string in their name ("32PFL4308"). Basically the same applies to the remaining variant sets. However, if you do not want to mess with the names anymore, I can probably live with the current state as well :slightly_smiling_face:

vitek-rostislav (approved):

pk16011990 (approved):

PR #3589, opened 2024-11-11
Uploaded files can be open in a browser

malyMiso:
Uploaded files can be configured to being streamed to a browser instead of downloading. 

Deployment update: https://github.com/shopsys/deployment/pull/80

- [x] New feature 

----
  - https://mt-view-uploaded-file.odin.shopsys.cloud
  - https://cz.mt-view-uploaded-file.odin.shopsys.cloud
  - https://mt-view-uploaded-file.odin.shopsys.cloud/sk

----
  - https://mt-view-uploaded-file.odin.shopsys.cloud
  - https://cz.mt-view-uploaded-file.odin.shopsys.cloud
  - https://mt-view-uploaded-file.odin.shopsys.cloud/sk


vitek-rostislav:
I do not think we should close this, let's rather process it :slightly_smiling_face:

packages/framework/src/Controller/Front/UploadedFileController.php
```diff
@@ -21,23 +24,58 @@ public function __construct(
...
         }
     }
+
+    public function viewAction(int $uploadedFileId, string $uploadedFilename): StreamedResponse
+    {
+        try {
+            $uploadedFile = $this->getByIdSlugAndExtension($uploadedFileId, $uploadedFilename);
+
+            if (!$this->uploadedFileFacade->isUploadedFileViewableInBrowser($uploadedFile)) {
+                throw new NotFoundRedirectToStorefrontException(
```
grossmannmartin: can't we just fallback to download url?

grossmannmartin (approved):

packages/administration/translations/messages.en.po
```diff
@@ -2149,6 +2149,9 @@ msgstr ""
 msgid "Very Weak"
 msgstr ""
 
+msgid "View"
+msgstr "View"
```
grossmannmartin: English translations should not be filled if are the same

grossmannmartin (approved):

packages/framework/build.xml
```diff
@@ -1003,6 +1005,14 @@
         </if>
     </target>
 
+    <target name="recalculate-file-sizes" description="Populates file size for uploaded files and images that do not have it stored.">
+        <exec executable="${path.php.executable}" passthru="true" checkreturn="true">
+            <arg value="${path.bin-console}"/>
+            <arg value="shopsys:recalculate-file-sizes"/>
```
grossmannmartin: This command will be run after each deploy, but it's needed just once.

@pk16011990, please verify if it's ok by you. Otherwise, we could add some list of executed tasks (similar to database migration table)

packages/framework/src/Command/RecalculateFileSizesCommand.php
grossmannmartin: If we want to improve performance even further, we can bypass entity hydration completely and return only the IDs of files without filesize set.
That lower the memory usage and we won't need to clear entity manager

grossmannmartin (changes requested):

grossmannmartin (approved):

grossmannmartin (approved):

packages/framework/build.xml
pk16011990: Good point! @TomasLudvik, if it is not too complicated, pls do it.

packages/framework/src/Command/RunPostDeployTasksCommand.php
```diff
@@ -0,0 +1,36 @@
...
+use Override;
+use Shopsys\FrameworkBundle\Component\PostDeploy\Task\PostDeployTaskRunnerFacade;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+#[AsCommand(
+    name: 'shopsys:post-deploy:run-tasks',
```
grossmannmartin: We already have `deploy:` commands, where I think this would fit better.

packages/framework/src/Component/PostDeploy/Task/PostDeployTaskConfigLoader.php
```diff
@@ -0,0 +1,127 @@
...
+namespace Shopsys\FrameworkBundle\Component\PostDeploy\Task;
+
+use Psr\Container\ContainerInterface;
+use Shopsys\FrameworkBundle\Component\PostDeploy\Task\Exception\PostDeployTaskConfigException;
+use Symfony\Component\Config\Definition\Processor;
+use Symfony\Component\Filesystem\Exception\FileNotFoundException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Yaml\Yaml;
+
+class PostDeployTaskConfigLoader
```
grossmannmartin: I think we kinda re-invent the wheel here. 

What about putting post-deploy tasks under shopsys_framework semantic config?
I prototyped it in https://github.com/shopsys/shopsys/commit/83aa91f4443d9ed3a38262ed1a0bc0cd0835da31, and I like that it follows the already established Symfony's bundle configuration way and also we don't need to use custom loader at all.

packages/framework/src/DependencyInjection/Configuration.php
```diff
@@ -27,6 +31,48 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('tasks')
+                            ->normalizeKeys(false)
+                            ->useAttributeAsKey('name')
+                            ->arrayPrototype()
+                                ->children()
+                                    ->enumNode('run')
+                                        ->isRequired()
+                                        ->values((new PostDeployTaskRunEnum())->getAllCases())
```
grossmannmartin, line 44: Nitpick:
With `new` here, the enum may be a true enum as it's not extendable anyway. 

Or it should be `final` to make the intent clear.

I'm absolutely ok with a non-extendable task run list, as I don't think it will be necessary to modify this list.

grossmannmartin (approved):

PR #3616, opened 2024-11-25
[frontend-api] JwtConfiguration is now created on demand

TomasLudvik:
During run of simple tasks like checking standards there is no need to have frontend-api keys generated, but current situation required it so this PR changes such behavior and \Lcobucci\JWT\Configuration is not created until it is needed.

#### Fixes issues
... 

- [ ] New feature 

----
  - https://tl-jwt-configuration-fix.odin.shopsys.cloud
  - https://cz.tl-jwt-configuration-fix.odin.shopsys.cloud

### DIFF
--- a/packages/frontend-api/src/Model/Token/JwtConfigurationProvider.php
+++ b/packages/frontend-api/src/Model/Token/JwtConfigurationProvider.php
@@ -10,9 +10,11 @@ use Lcobucci\JWT\Signer\Key;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 
-class JwtConfigurationFactory
+class JwtConfigurationProvider
 {
-    protected const FRONTEND_API_KEYS_FILEPATH_PARAMETER = 'shopsys.frontend_api.keys_filepath';
+    protected const string FRONTEND_API_KEYS_FILEPATH_PARAMETER = 'shopsys.frontend_api.keys_filepath';
+
+    protected ?Configuration $configuration = null;
 
     /**
      * @param \Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface $parameterBag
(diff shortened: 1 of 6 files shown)


packages/frontend-api/src/Model/Token/JwtConfigurationFactory.php
```diff
@@ -29,15 +31,23 @@ public function __construct(
      */
     public function create(): Configuration
     {
+        if (isset($this->configuration)) {
```
vitek-rostislav: why do not you use `InMemoryCache` service for the caching?

packages/frontend-api/src/Model/Token/JwtConfigurationFactory.php
```diff
@@ -14,6 +14,8 @@ class JwtConfigurationFactory
 {
     protected const FRONTEND_API_KEYS_FILEPATH_PARAMETER = 'shopsys.frontend_api.keys_filepath';
 
+    protected Configuration $configuration;
```
vitek-rostislav: the class should not be defined as service anymore, right? So the following lines should be removed from the `services.yaml`:

```
    Lcobucci\JWT\Configuration:
        factory: [ '@Shopsys\FrontendApiBundle\Model\Token\JwtConfigurationFactory', create ]
```

vitek-rostislav: btw, the change deserves upgrade notes IMHO (something like "Lcobucci\JWT\Configuration is not a service anymore, use Shopsys\FrontendApiBundle\Model\Token\JwtConfigurationFactory::create() to get the configuration")

vitek-rostislav (changes requested):

packages/frontend-api/src/Model/Token/JwtConfigurationFactory.php
TomasLudvik: I do not want to use ResetInterface for this as it is not necessary

vitek-rostislav (approved):

vitek-rostislav (approved):

pk16011990 (approved):

PR #3628, opened 2024-11-29
added time-limited price lists with product special prices

grossmannmartin:

This PR adds a new section `Pricing -> Price lists`, where price lists can be managed.

Snímek obrazovky 2024-12-09 v 9 57 27

Each price list can have several products with their special price set. This price is valid only during specific time period.

Snímek obrazovky 2024-12-09 v 9 58 34

Product detail in administration now display all current and future price lists with the set special price for the current product.
The currently used price is highlighted.

Snímek obrazovky 2024-12-09 v 10 00 25

The currently used special price is respected on the storefront as a current selling price with the possibility to display the original price. 
The special price is used only if lower than the basic price. 

#### Fixes issues
... 

----
  - https://mg-price-lists.odin.shopsys.cloud
  - https://cz.mg-price-lists.odin.shopsys.cloud


packages/framework/src/Model/PriceList/PriceListDataFactory.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->fillNew($priceListData);
+
+        return $priceListData;
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\PriceList\PriceListData $priceListData
+     */
+    protected function fillNew(PriceListData $priceListData): void
+    {
```
vitek-rostislav, line 40: `$priceListData->domainId = $this->adminDomainTabsFacade->getSelectedDomainId()` could be here so the method is not empty :slightly_smiling_face:

packages/framework/src/Model/PriceList/PriceListRepository.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * @return \Doctrine\ORM\QueryBuilder
+     */
+    public function getPriceListGridQueryBuilder(): QueryBuilder
+    {
+        return $this->getPriceListRepository()
+            ->createQueryBuilder('pl')
+            ->addSelect('pl.name');
```
vitek-rostislav: why is this necessary, please?

packages/framework/src/Resources/views/Admin/Content/PriceList/detail.html.twig
```diff
@@ -0,0 +1,17 @@
+{% extends '@ShopsysFramework/Admin/Layout/layoutWithPanel.html.twig' %}
+
+{% block title %}- {{ 'Price list'|trans }} {{ priceList.name }}{% endblock %}
+{% block h1 %}{{ 'Price list'|trans }} {{ priceList.name }}{% endblock %}
```
vitek-rostislav, line 4: this seems unnecessary as both blocks are overridden in `new` and `edit` templates. However, I can imagine we have it like this in many more templates, right?

project-base/app/src/DataFixtures/Demo/PriceListDataFixture.php
```diff
@@ -31,21 +32,29 @@ public function load(ObjectManager $manager): void
...
             $priceListData->domainId = $domainConfig->getId();
+            $priceListData->validFrom = new DateTimeImmutable('2023-11-10 00:00:00');
+            $priceListData->validTo = new DateTimeImmutable('2023-11-10 23:59:59');
             $this->priceListFacade->create($priceListData);
 
             $priceListData = $this->priceListDataFactory->create();
             $priceListData->name = t('Items on sale', [], Translator::DATA_FIXTURES_TRANSLATION_DOMAIN, $domainConfig->getLocale());
             $priceListData->domainId = $domainConfig->getId();
+            $priceListData->validFrom = new DateTimeImmutable('2023-02-12 06:20:00');
+            $priceListData->validTo = new DateTimeImmutable('2084-05-10 08:30:00');
```
vitek-rostislav, line 81: I am looking forward to failing tests in 60 years :smile:

packages/framework/src/Controller/Admin/PriceListController.php
```diff
@@ -74,6 +74,7 @@ public function listAction(): Response
     public function newAction(Request $request): Response
     {
         $priceListData = $this->priceListDataFactory->create();
+        $priceListData->domainId = $request->get('domainId', Domain::FIRST_DOMAIN_ID);
```
vitek-rostislav: as mentioned above, why not `AdminDomainTabsFacade`?

packages/framework/assets/js/admin/components/ProductsPickerWithPriceWindow.js
```diff
@@ -0,0 +1,75 @@
+import '../../common/bootstrap/tooltip';
+import Translator from 'bazinga-translator';
+import Register from '../../common/utils/Register';
+
+export default class ProductsPickerWithPriceWindow {
```
vitek-rostislav: I just need to mention that that "withPrice" might be a little bit confusing - it looks like I can choose to display prices in the popup itself. But I do not have any better suggestion so :shrug: but we can at least talk about it

vitek-rostislav: I just need to mention  that "withPrice" might be a little bit confusing - it looks like I can choose to display prices in the popup itself. But I do not have any better suggestion so :shrug: but we can at least talk about it

vitek-rostislav: moreover, it would be awesome if we did not have to implement the javascript at all (eg. `markAddButtonAsAdded` is now three times in our codebase :frowning_face: But I know you tried, so let's just cry together for a while about that :smile:

vitek-rostislav: ...it will be fun rewriting to the new admin template, right?

packages/framework/assets/js/admin/components/ProductsWithPricePicker.js
```diff
@@ -0,0 +1,215 @@
...
+                _this.updateDiscount($item);
+            });
+        }
+
+        _this.updateDiscount($item);
+        _this.productItems.push($item);
+        $item.find('.js-products-with-price-picker-item-button-delete').click(() => {
+            _this.removeItem($item);
+        });
+        $item.find('.js-remove-form-table-line').click(() => {
```
vitek-rostislav: the `js-remove-form-table-line` does not exists anywhere in the codebase

packages/framework/assets/js/admin/components/ProductsWithPricePicker.js
```diff
@@ -0,0 +1,215 @@
...
+
+            const inputPrice = parseFloat($input.val().replace(',', '.'));
+            const basicDataPrice = $basicPrice.data('price');
+            const basicPrice = parseFloat(typeof basicDataPrice === 'string' ? basicDataPrice.replace(',', '.') : basicDataPrice);
+            const discount = Math.round((basicPrice - inputPrice) * 100) / 100;
+            const discountColor = basicPrice > inputPrice ? 'green' : 'red';
+            const discountPercentage = Math.floor(discount / (basicPrice === 0 ? 1 : basicPrice) * 100);
+
+            $discount.text(formatPrice(discount, $basicPrice.data('locale'), $basicPrice.data('currency')));
+            $discountPercentage.text(discountPercentage + '%');
```
vitek-rostislav: grammarnazi nitpick, there should be a space between the percentage and the `%` sign (at least according to the Czech grammar, and if you want to say "XY percent". However, if it is supposed to be an adjective, then everything is alright :smile:

packages/framework/src/Controller/Admin/PriceListController.php
vitek-rostislav: ah, now I probably understand - because of "all domains" tab, right?

packages/framework/src/Form/Admin/PriceList/PriceListProductPickerType.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Validator\Constraints;
+
+final class PriceListProductPickerType extends AbstractType
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Form\Transformers\ProductIdToProductTransformer $productsIdToProductsTransformer
+     */
+    public function __construct(
+        protected readonly ProductIdToProductTransformer $productsIdToProductsTransformer,
```
vitek-rostislav: nitpick, protected does not make sense in final class, or?

packages/framework/src/Model/PriceList/PriceList.php
```diff
@@ -77,6 +85,17 @@ protected function setData(PriceListData $priceListData): void
         $this->validFrom = $priceListData->validFrom;
         $this->validTo = $priceListData->validTo;
         $this->lastUpdate = new DateTimeImmutable();
+
+        // products with prices are not set here, they are set in the refreshProductWithPrices method
```
vitek-rostislav:
```suggestion
        // products with prices are not set here, they are set in the PriceListFacade::refreshProductWithPrices method
```

vitek-rostislav: offtopic, btw, is it just my feeling that we handle each collection in the codebase in its own unique way?

packages/framework/src/Model/PriceList/ProductWithPrice.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Shopsys\FrameworkBundle\Model\PriceList;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Table(name="products_with_prices")
+ * @ORM\Entity
+ */
+class ProductWithPrice
```
vitek-rostislav: at first sight, I would rather vote for something like `ProductWithSpecialPrice` however, I am not sure whether it is worth renaming. My point is, it might be confusing as it looks like regular product prices can be stored here...we have `product_manual_input_prices` and `products_with_prices` now, which might be hard to understand for newbies

packages/framework/src/Model/PriceList/ProductWithPrice.php
```diff
@@ -0,0 +1,76 @@
...
+    protected $priceList;
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\PriceList\ProductWithPriceData $productWithPriceData
+     */
+    public function __construct(
+        ProductWithPriceData $productWithPriceData,
+    ) {
+        $this->product = $productWithPriceData->product;
+        $this->priceAmount = $productWithPriceData->priceAmount;
```
vitek-rostislav: editing the price list recreates the whole collection, right? So there is no need for the "edit" method here

packages/framework/src/Resources/views/Admin/Form/productsWithPriceFormTheme.html.twig
```diff
@@ -0,0 +1,154 @@
+{% block price_list_products_picker_row -%}
```
vitek-rostislav: it is quite confusing (but I can imagine it is not unusual in our codebase) that this file has a different name from the type it belongs to (`PriceListProductsPickerType`)

packages/framework/src/Resources/views/Admin/Form/theme.html.twig
```diff
@@ -566,16 +566,18 @@
...
                     </a>
                 {% endif %}
             </span>
         </td>
         <td class="table-form__cell table-form__cell--actions">
-            <span class="cursor-pointer js-products-picker-item-button-delete">
-                <span class="svg svg-delete-thin in-icon"></span>
-                {{ form_widget(form, {attr: {class: 'js-products-picker-item-input'}}) }}
-            </span>
+            {%- if disabled == false %}
```
vitek-rostislav:
```suggestion
            {%- if disabled is same as false %}
```

packages/framework/src/Resources/views/Admin/Form/theme.html.twig
```diff
@@ -598,18 +600,20 @@
...
                 {% endfor %}
             </tbody>
         </table>
-        <div class="form-line form-line--no-top-border">
-            <div class="form-line__line form-line__line--no-space">
-                <a href="#" class="btn btn--plus js-products-picker-button-add">
-                    <i class="btn__icon">+</i>
-                    {{ label_button_add }}
-                </a>
+        {%- if disabled == false %}
```
vitek-rostislav:
```suggestion
        {%- if disabled is same as false %}
```

project-base/app/templates/Admin/Form/productsFormTheme.html.twig
```diff
@@ -1,61 +0,0 @@
-{% block products_widget %}
```
vitek-rostislav, line 1: just a note for my future self - check that movement of the `disabled` parameter in `productsWidgetItem` is mentioned in the upgrade notes :slightly_smiling_face:

packages/framework/src/Model/PriceList/PriceListFacade.php
```diff
@@ -62,26 +67,50 @@ public function create(PriceListData $priceListData): PriceList
...
+                [ProductExportScopeConfig::SCOPE_PRICE],
+            );
+
+            unset($originalProductIds[$productWithPrice->getProduct()->getId()]);
         }
 
+        $this->productRecalculationDispatcher->dispatchProductIds(
+            array_keys($originalProductIds),
+            ProductRecalculationPriorityEnum::HIGH,
+            [ProductExportScopeConfig::SCOPE_PRICE],
```
vitek-rostislav, line 110: just want to take a minute to appreciate all these mechanisms we have already implemented and now can be leveraged 

image

packages/framework/src/Model/PriceList/PriceListFacade.php
```diff
@@ -62,26 +67,50 @@ public function create(PriceListData $priceListData): PriceList
...
         $this->em->flush();
 
-        $this->refreshProductWithPrices($priceList, $priceListData);
+        $this->refreshProductWithPrices($priceList, $priceListData, $originalProductIds);
     }
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\PriceList\PriceList $priceList
      * @param \Shopsys\FrameworkBundle\Model\PriceList\PriceListData $priceListData
+     * @param array $originalProductIds
```
vitek-rostislav:
```suggestion
     * @param int[] $originalProductIds
```

packages/framework/src/Model/PriceList/PriceListFacade.php
```diff
@@ -62,26 +67,50 @@ public function create(PriceListData $priceListData): PriceList
...
+                $productWithPrice->getProduct()->getId(),
+                ProductRecalculationPriorityEnum::HIGH,
+                [ProductExportScopeConfig::SCOPE_PRICE],
+            );
+
+            unset($originalProductIds[$productWithPrice->getProduct()->getId()]);
         }
 
+        $this->productRecalculationDispatcher->dispatchProductIds(
+            array_keys($originalProductIds),
```
vitek-rostislav: This array flipping back and forth makes me think, why is removing an element by value so cumbersome? :smile: 
Maybe it is worth creating a function for that? Something like `ArrayHelper::removeElementByValue(array $array, mixed $value)`. This [stack overflow topic](https://stackoverflow.com/questions/7225070/php-array-delete-by-value-not-key) has some interesting ideas (your approach is one of them). We can discuss this. Personally, I find the most interesting this one:
```
foreach (array_keys($messages, 401, true) as $key) {
    unset($messages[$key]);
}
```

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceFacade.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param int $domainId
+     * @param int[] $variantIds
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\SpecialPrice\SpecialPrice[]
+     */
+    public function getCurrentAndFutureSpecialPrices(Product $product, int $domainId, array $variantIds = []): array
+    {
+        $specialPrices = $this->specialPriceRepository->getCurrentAndFutureSpecialPrices($product, $domainId, $variantIds);
+
+        return array_map(
+            function (array $specialPriceData) use ($domainId, $product) {
```
vitek-rostislav, line 65: Have you considered creating an object instead of using the array? `SpecialPriceData`? :slightly_smiling_face:

vitek-rostislav: maybe it could ease the interface of some methods (eg., at first sight, `SpecialPriceFactory::create()` and `createWithCalculations`?)

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceRepository.php
```diff
@@ -0,0 +1,72 @@
...
+            (:currentDate BETWEEN pl.validFrom AND pl.validTo)
+            OR (:currentDate < pl.validFrom)
+        ')
+            ->setParameter('productIds', [...$variantIds, $product->getId()])
+            ->setParameter('domainId', $domainId)
+            ->setParameter('currentDate', $currentDate)
+            ->orderBy('CASE
+                WHEN :currentDate BETWEEN pl.validFrom AND pl.validTo THEN 1
+                ELSE 2
+            END', 'ASC') // Current price lists (1) are prioritized over future ones (2)
```
vitek-rostislav, line 78: Some might disagree with me, however, I appreciate the comments here :smile:

project-base/app/src/Resources/definition/product/1.json
```diff
@@ -195,6 +195,36 @@
                     }
                 }
             },
+            "special_prices": {
```
vitek-rostislav, line 198: in the context of this commit, the structure defined here does not exactly match the structure actually exported in the `ProductExportRepository` class, however, I know it would be probably difficult to fix this in history so hopefully, it won't confuse anyone

project-base/app/src/Resources/definition/product/1.json
```diff
@@ -195,6 +195,36 @@
...
+                            "price_with_vat": {
+                                "type": "float"
+                            },
+                            "price_without_vat": {
+                                "type": "float"
+                            },
+                            "vat": {
+                                "type": "float"
+                            },
+                            "product_id": {
```
vitek-rostislav, line 227: is this because of the variants? 
// edit: Ah, now I found the usage in `getMinCurrentSellingPriceRuntimeField`...is not it possible to use `params['_source']['id]` there? Maybe it is a nitpick, it just feels weird extracting the product ID into a nested field and it only contains the same value as the "ID" field itself.

And a little off-topic - it seems like there are no special prices set for variants in data fixtures. Shouldn't they be there? It could open some more _opportunities_ for manual/automated testing, right? :grimacing:

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceFacade.php
```diff
@@ -18,6 +18,29 @@ public function __construct(
     ) {
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\SpecialPrice\SpecialPrice|null
+     */
+    public function getEffectiveSpecialPrice(Product $product, int $domainId): ?SpecialPrice
```
vitek-rostislav: Maybe "current" would be a more appropriate name here? Let's discuss it, please

vitek-rostislav: Ah, few lines later, I understand - it is not just the current but it also might be a future special price, right? So what about "relevant" instead of "effective"? And one more thing, according to our conventions, the methods here and in the repository should be named `find...`

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceRepository.php
```diff
@@ -20,6 +20,19 @@ public function __construct(
     ) {
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param int $domainId
+     * @return array{priceAmount:\Shopsys\FrameworkBundle\Component\Money\Money, validFrom: \DateTimeImmutable, validTo: \DateTimeImmutable, productListId: int}|null
```
vitek-rostislav: I would really vote for creating a data object. As much as I like these annotations, adding a new "property" now could be a little headache IMHO, or?

packages/frontend-api/src/Model/Price/PriceInfo.php
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Model\Price;
+
+use DateTimeInterface;
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Shopsys\FrameworkBundle\Model\Pricing\Price;
+
+class PriceInfo
```
vitek-rostislav, line 11: Just a note for a quick discussion as I know this is something we are generally probably ok with. However, for me, naming objects in the backend differently than the objects used in the API, makes the code navigation and understanding harder. E.g. `CartDecorator` type vs `CartWithModifications` object. There are probably many more examples of this "mismatch". `PriceInfo` is the latest one as in API, it is actually represented by `ProductPrice`, right?

packages/frontend-api/src/Model/Price/PriceInfoFactory.php
```diff
@@ -0,0 +1,99 @@
...
+
+        if ($specialPrice === null) {
+            $priceInfo->setSellingPrice($basicProductPrice);
+
+            return $priceInfo;
+        }
+
+        $priceInfo->nextPriceChange = $this->determineNextPriceChange($specialPrice);
+
+        if (!$specialPrice->isFuturePrice()) {
```
vitek-rostislav, line 37: I really like the concept of passing the current special price as the selling price :ok_hand: I can imagine the code on the storefront is then really straightforward

packages/frontend-api/src/Model/Price/PriceInfoFactory.php
```diff
@@ -0,0 +1,99 @@
...
+        if ($specialPrice->validFrom > $now) {
+            $futureDates[] = $specialPrice->validFrom;
+        }
+
+        if ($specialPrice->validTo > $now) {
+            $futureDates[] = $specialPrice->validTo;
+        }
+
+        if (count($futureDates) === 0) {
+            throw new LogicException('Special price was selected, but the validity is in the past. Check the implementation of the special price selection algorithm.');
```
vitek-rostislav, line 77: maybe a particular class::method as a hint should be mentioned here? "implementation of the special price selection algorithm" seems too general to me, I would probably not be sure where to go in case of such an error

packages/frontend-api/src/Model/Resolver/Price/ProductPriceResolverMap.php
```diff
@@ -16,7 +17,11 @@ protected function map(): array
     {
         return [
             'ProductPrice' => [
-                'isPriceFrom' => function (ProductPrice $productPrice) {
+                'isPriceFrom' => function (ProductPrice|PriceInfo $productPrice) {
```
vitek-rostislav: is this resolver necessary yet? It feels like `ProductPrice` API type is always and only resolved by `priceByProductQuery` which always returns `PriceInfo`, right? 
1) It feels like the original implementation was unnecessary as `ProductPrice.isPriceFrom` in API should be automatically mapped by `ProductPrice::isPriceFrom()` method
2) now, I would say `ProductPrice` object never comes as an input here anymore, so you could maybe just rename the API field to `priceFrom` or rename `PriceInfo::$priceFrom` to `PriceInfo::$isPriceFrom` (or add a getter?)

But maybe I am missing something :thinking:

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        $effectiveSpecialPrice = $this->getEffectiveSpecialPrice($specialPricesArray);
+
+        if ($effectiveSpecialPrice === null || $effectiveSpecialPrice->price->getPriceWithVat()->isLessThan($basicPrice->getPriceWithVat())) {
+            return $effectiveSpecialPrice;
+        }
+
+        return null;
+    }
```
vitek-rostislav: this is quite hard to read, could we invert it into this?
```
if ($effectiveSpecialPrice !== null && $effectiveSpecialPrice->price->getPriceWithVat()->isGreaterThanOrEqualTo($basicPrice->getPriceWithVat())) {
        return $effectiveSpecialPrice;
}

return null;
```

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * @param array $specialPricesArray
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\SpecialPrice\SpecialPrice|null
+     */
+    protected function getEffectiveSpecialPrice(array $specialPricesArray): ?SpecialPrice
```
vitek-rostislav: should be named `find...`

project-base/app/config/graphql/types/ModelType/Product/Product.types.yaml
```diff
@@ -38,10 +38,6 @@ Product:
                     type:
                         type: "String"
                         defaultValue: null
-            price:
```
vitek-rostislav, line 41: yay, few more lines removed from project-base :green_heart:

vitek-rostislav, line 41: yay, few more lines removed from project-base :green_heart:

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
```diff
@@ -0,0 +1,96 @@
...
+                if ($currentDateTime >= $specialPriceCandidate->validFrom && $currentDateTime <= $specialPriceCandidate->validTo) {
+                    if (array_key_exists($specialPriceCandidate->productId, $usedProductIds)) {
+                        continue;
+                    }
+
+                    if ($finalSpecialPrice === null || $finalSpecialPrice->price->getPriceWithVat()->isGreaterThan($specialPriceCandidate->price->getPriceWithVat())) {
+                        $finalSpecialPrice = $specialPriceCandidate;
+                    }
+
+                    $usedProductIds[$specialPriceCandidate->productId] = true;
```
vitek-rostislav, line 70: I would like to talk about the productId as once again, it feels like it is related to the variants, or I do not understand that at all :sweat_smile:

vitek-rostislav, line 70: I would like to talk about the productId as once again, it feels like it is related to the variants, or I do not understand that at all :sweat_smile:

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                if ($finalSpecialPrice === null && $currentDateTime < $specialPriceCandidate->validFrom) {
```
vitek-rostislav, line 73: could `$finalSpecialPrice` not being null mean we can add an early return in that case? Maybe it could ease the method (or maybe it could make it more complex? :smile: )

packages/framework/src/Form/Admin/Product/ProductFormType.php
```diff
@@ -604,6 +607,28 @@ private function createPricesGroup(FormBuilderInterface $builder, ?Product $prod
             ],
         );
 
+        if ($product !== null) {
+            $priceListOverviewOptionsByDomainId = [];
+
+            foreach ($this->domain->getAllIds() as $domainId) {
```
vitek-rostislav: shouldn't we use `getAdminEnabledDomainIds` here?

vitek-rostislav: ah, the rendering of just selected domains is ensured by MultidomainType, right? But using `getAdminEnabledDomainIds` could also have a positive effect on the page performance in particular scenarios, right? Or does it not work when all domains are not included in `options_by_domain_id`?

packages/framework/src/Resources/views/Admin/Form/Product/Price/priceListOverviewType.html.twig
```diff
@@ -0,0 +1,53 @@
...
+                                    {{ icon('forward-page') }}
+                                </a>
+                            </td>
+                            <td class="table-form__cell">{{ specialPrice.validFrom|formatDateTime }}</td>
+                            <td class="table-form__cell">{{ specialPrice.validTo|formatDateTime }}</td>
+                            <td class="table-form__cell">{{ specialPrice.price.priceWithVat|priceWithCurrencyByDomainId(domainId) }}</td>
+                        </tr>
+                        {% if not isHighlighted and specialPrice.isNowActive %}
+                            {% set isHighlighted = true %}
+                        {% endif %}
```
vitek-rostislav: maybe I am missing something but could not you just `{% set isHighlighted = specialPrice.isNowActive %}` as the first line in the foreach loop?

packages/framework/src/Model/Product/Search/FilterQuery.php
```diff
@@ -563,6 +661,10 @@ public function getQuery(): array
             ],
         ];
 
+        if ($this->runtimeFields !== []) {
```
vitek-rostislav, line 702: are these `if`s necessary? What happens what you pass empty array to `runtime_mappings`?

This would be easier then:
```
'body' => ['runtime_mappings' => $this->runtimeFields, 'from' => ..., ...]
```

packages/framework/src/Model/Product/Search/FilterQuery.php
```diff
@@ -146,6 +119,138 @@ public function applyOrdering(string $orderingModeId, PricingGroup $pricingGroup
         return $clone;
     }
 
+    /**
+     * @param int $pricingGroupId
+     * @return array
+     */
+    protected function getMinCurrentSellingPriceRuntimeField(int $pricingGroupId): array
```
vitek-rostislav, line 127: ChatGPT helped you to write it, right? Now it also helped me to understand it :smile: https://chatgpt.com/share/675b0330-11b4-8009-839d-69959f13ae89

packages/framework/src/Model/Product/Search/FilterQuery.php
```diff
@@ -146,6 +119,138 @@ public function applyOrdering(string $orderingModeId, PricingGroup $pricingGroup
...
+    public function getMaxCurrentSellingPriceRuntimeField(int $pricingGroupId): array
+    {
+        $scriptMaxValue = "
+            double finalPrice = 0;
+            DateTimeFormatter formatter = java.time.format.DateTimeFormatter.ofPattern('yyyy-MM-dd HH:mm:ss').withZone(java.time.ZoneOffset.UTC);
+
+            if (!params['_source']['prices'].isEmpty()) {
+                for (def price : params['_source']['prices']) {
+                    if (price['pricing_group_id'] === params['pricing_group_id']) {
+                        finalPrice = Math.max(finalPrice, price['filtering_minimal_price']);
```
vitek-rostislav, line 194: `filtering_minimal_price` is exported in `project-base` only, could you move it, please? :pray: 
https://github.com/shopsys/shopsys/blob/16.0/project-base/app/src/Model/Product/Elasticsearch/ProductExportRepository.php#L176

(the same applies to `filtering_maximal_price`)

docs/functional/time-limited-price-lists.md
```diff
@@ -0,0 +1,42 @@
...
+[TOC]
+
+## Overview
+
+Time-limited price lists allow administrators to set specific prices for products that are valid for a defined time range.
+This feature is ideal for managing promotions, sales campaigns, or other time-sensitive pricing strategies.
+
+## Managing the time-limited price lists
+
+Time-limited price lists can be managed in the `Price -> Price lists` section.
```
vitek-rostislav:
```suggestion
Time-limited price lists can be managed in the `Pricing -> Price lists` section.
```

project-base/app/tests/App/Functional/Model/Pricing/EffectiveSpecialPriceForProductTest.php
```diff
@@ -0,0 +1,153 @@
...
+        $priceAmount = '10';
+
+        $this->addProductToPriceList($helloKittyProduct, PriceListDataFixture::ACTIVE_SPECIAL_OFFERS_REFERENCE, $priceAmount);
+
+        $specialPrice = $this->specialPriceFacade->getEffectiveSpecialPrice($helloKittyProduct, self::TEST_DOMAIN_ID);
+
+        $this->assertNotNull($specialPrice, 'Special price should be set for product');
+        $this->assertTrue($specialPrice->isNowActive(), 'Special price should be active');
+        $this->assertFalse($specialPrice->isFuturePrice(), 'Special price should not be future price');
+        $this->assertMoney($priceAmount, $specialPrice->price->getPriceWithVat());
```
vitek-rostislav: nitpick, looks like this can be extracted into new private `assertSpecialPrice()` method :slightly_smiling_face:

vitek-rostislav: (but I see that not all the assert sequences are the same so it's up to you, maybe the common method does not make sense :thinking:

project-base/app/tests/App/Functional/Model/Pricing/EffectiveSpecialPriceForProductTest.php
```diff
@@ -0,0 +1,153 @@
...
+
+        $this->assertNull($specialPrice, 'No special price should be set for product');
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param string $priceListReferenceName
+     * @param string $productPriceAmount
+     */
+    public function addProductToPriceList(
```
vitek-rostislav: can be private :slightly_frowning_face:

vitek-rostislav: can be private :slightly_frowning_face:

project-base/app/tests/FrontendApiBundle/Functional/SpecialPrice/ProductPriceQueryTest.php
```diff
@@ -0,0 +1,482 @@
...
+use App\DataFixtures\Demo\ProductDataFixture;
+use DateTimeImmutable;
+use PHPUnit\Framework\Attributes\DataProvider;
+use Shopsys\FrameworkBundle\Component\Money\HiddenMoney;
+use Shopsys\FrameworkBundle\Model\Product\Product;
+use Shopsys\FrameworkBundle\Model\Product\ProductTypeEnum;
+use Shopsys\FrontendApiBundle\Model\Resolver\Price\ProductPriceQuery;
+use Tests\App\Test\TransactionFunctionalTestCase;
+
+final class ProductPriceQueryTest extends TransactionFunctionalTestCase
```
vitek-rostislav, line 16: quite an unusual way of testing the API :thinking: I am not saying there is anything wrong with that, I am just curious, is there any particular reason why you do not use `GraphQlTestCase`?

vitek-rostislav (changes requested):

packages/framework/src/Model/PriceList/PriceListDataFactory.php
grossmannmartin: I don't think we want to rely on the fact that this factory will be used only in the administration. If someone start to transfer the price lists, this may be unpleasant surprise

packages/framework/src/Model/PriceList/PriceListRepository.php
grossmannmartin: it's not, fixed in cb3e4699b7 and 5b8047770e

packages/framework/src/Resources/views/Admin/Content/PriceList/detail.html.twig
grossmannmartin: Exactly... it's the same in all templates.

project-base/app/src/DataFixtures/Demo/PriceListDataFixture.php
grossmannmartin: :D

packages/framework/src/Controller/Admin/PriceListController.php
grossmannmartin: Yes, yes it is because of the all domains

packages/framework/assets/js/admin/components/ProductsPickerWithPriceWindow.js
grossmannmartin: Let's talk about the names, please... I'm not happy with them, but I'm not sure about the right naming

grossmannmartin: I put all my dreams and hopes into the new admin template

packages/framework/assets/js/admin/components/ProductsWithPricePicker.js
grossmannmartin: nice catch e020e21059

packages/framework/assets/js/admin/components/ProductsWithPricePicker.js
grossmannmartin: Isn't it in Czech only? 

According to central brain of the humanity (chatgpt): 

The formatting of percentage values, including whether there should be a space between the number and the % symbol, depends on the style guide or convention you are following:
1. Without a space (common in English-speaking countries):
10%
This is the most widely accepted formatting in general usage.
2. With a space (used in some European style guides, e.g., SI/ISO standards):
10 %
This format is often used in technical and scientific contexts where consistency with international standards is preferred.

packages/framework/src/Form/Admin/PriceList/PriceListProductPickerType.php
grossmannmartin: no, it doesn't 👍 
512d79eab2

packages/framework/src/Model/PriceList/PriceList.php
grossmannmartin: 😆 I know... I think we handle each collection weirdly in a different way

packages/framework/src/Model/PriceList/ProductWithPrice.php
grossmannmartin: Let's discuss the naming please

packages/framework/src/Model/PriceList/ProductWithPrice.php
grossmannmartin: yes, I'm not sure whether we can use edit as the items may be reordered 🤔

packages/framework/src/Resources/views/Admin/Form/productsWithPriceFormTheme.html.twig
grossmannmartin: Yes, the name should be unified... good point ffb4e3ceb8

project-base/app/templates/Admin/Form/productsFormTheme.html.twig
grossmannmartin: it is 
```
- twig macro `productsWidgetItem` now allows to render the product picker as disabled
```

hmm... now I see I have it not like a movement, but as a feature 🤔 
not sure if it's a problem

packages/framework/src/Model/PriceList/PriceListFacade.php
grossmannmartin: Why indeed 😆 

Helper seems interesting, but if I want to be pedantic, I would need to flip the array and return array_keys each time I call the helper (and I call it inside the loop)

I can use ArrayCollection for easier removal.

Or I can collect the recalculated IDs and then diff the original and collected arrays. 

TBH, the code you've posted I had to look into the documentation to understand how and why it works :D

I think the single array_flip call and removing items by key is the most effective way (but I agree it's cumbersome)

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceFacade.php
grossmannmartin: 🤔 I'm not sure if I miss something... 
The getResult() returns an array. Then I would need to map the array to the data object. Wouldn't it be just another layer and the code I have will be just elsewhere?
Let's discuss it please

project-base/app/src/Resources/definition/product/1.json
grossmannmartin: I hope so... I wanted to do it properly, but it was already a three-version-together rebase, and it was really painful

project-base/app/src/Resources/definition/product/1.json
grossmannmartin: the product id is used for variants, more specifically for main variant. 

the variants can have special prices spread across multiple price lists, and I need a way to identify the prices belonging to the same product to determine the correct "price from" and prices for filter.

it should be covered in test `\Tests\FrontendApiBundle\Functional\SpecialPrice\ProductPriceQueryTest::testValidSpecialPriceIsReturnedForVariants()` where you can see how it should behave

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceFacade.php
grossmannmartin: Let's discuss the name, please :) but I agree about `find...`

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceRepository.php
grossmannmartin: it could be cumbersome... but I still need this annotation somewhere... probably in a factory for the object I want to create instead of 🤷

packages/frontend-api/src/Model/Price/PriceInfo.php
grossmannmartin: Yes. But the reason is that `ProductPrice` already exists in the framework package and serves a different purpose.
I did not want to add a confusing same name in frontend api bundle. 

Another option is to rename the type in api, but I thought it's an unnecessary breaking change (regenerate api, change all usages on storefront).

Let's discuss it please

packages/frontend-api/src/Model/Price/PriceInfoFactory.php
grossmannmartin: It may be either `specialPriceApiFactory->createSpecialPriceFromArray()`, or `specialPriceFacade->getEffectiveSpecialPrice()`. Depends on the source of product (elastic/database). And potentially more usages in the future.

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
grossmannmartin: let's talk about it ;)

packages/framework/src/Resources/views/Admin/Form/Product/Price/priceListOverviewType.html.twig
grossmannmartin: no, because multiple price lists may be active at the moment, but only one can be highlighted as used

packages/framework/src/Model/Product/Search/FilterQuery.php
grossmannmartin: partially yes 😄

project-base/app/tests/FrontendApiBundle/Functional/SpecialPrice/ProductPriceQueryTest.php
grossmannmartin: speed of execution and the necessity of a specific setup for each test scenario. It would require many updates of the real data or adjusting the datafixtures (add more variants, add more price lists, and find the combination where I can test all scenarios I was thinking of)

project-base/app/tests/App/Functional/Model/Pricing/EffectiveSpecialPriceForProductTest.php
grossmannmartin: 6c3dd3730f

project-base/app/tests/App/Functional/Model/Pricing/EffectiveSpecialPriceForProductTest.php
grossmannmartin: I don't think I want to add such a specific assertion.

packages/frontend-api/src/Model/Resolver/Price/ProductPriceResolverMap.php
grossmannmartin: that's true, thanks 👍  I renamed the property in the PriceInfo class and removed this resolver map. 
73e697ea97 and 2d7882fc61

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
grossmannmartin: I assume the returns should be inverted too.

Maybe this could work too 

```php
if ($effectiveSpecialPrice?->price->getPriceWithVat()->isGreaterThanOrEqualTo($basicPrice->getPriceWithVat())) {
    return null;
}

return $effectiveSpecialPrice
```

packages/framework/src/Model/PriceList/PriceListDataFactory.php
vitek-rostislav: that is true :+1:

packages/framework/assets/js/admin/components/ProductsWithPricePicker.js
vitek-rostislav: I do not know why I wasted your time with the original unnecessary comment, my apologies:smile:

packages/framework/src/Model/PriceList/PriceList.php
vitek-rostislav: what can we do...however, what about the suggestion above?

project-base/app/templates/Admin/Form/productsFormTheme.html.twig
vitek-rostislav: IMHO it should be in the movement section but I do not insist on that

packages/framework/src/Model/PriceList/PriceListFacade.php
vitek-rostislav: let's discuss this please yet

packages/frontend-api/src/Model/Price/PriceInfoFactory.php
vitek-rostislav: ok, fair enough :+1:

packages/framework/src/Resources/views/Admin/Form/Product/Price/priceListOverviewType.html.twig
vitek-rostislav: ah, ok, I see

packages/framework/src/Model/PriceList/PriceListFacade.php
grossmannmartin: changed to array_diff with collected dispatched product ids approach

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceFacade.php
grossmannmartin: renamed to findRelevant...

packages/framework/src/Model/Product/Search/FilterQuery.php
grossmannmartin: it's not working with empty runtime_mappings

packages/framework/src/Form/Admin/Product/ProductFormType.php
grossmannmartin: e45cf65756

packages/framework/src/Model/PriceList/PriceListFacade.php
grossmannmartin: 4b6eb9e487

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
grossmannmartin: 98e64a83a9

project-base/app/templates/Admin/Form/productsFormTheme.html.twig
grossmannmartin: 444b4e1883

packages/frontend-api/src/Model/Resolver/Price/SpecialPriceApiFactory.php
grossmannmartin: renamed to findRelevant... cc81a4cab5

packages/framework/src/Model/Pricing/SpecialPrice/SpecialPriceFacade.php
grossmannmartin: cc81a4cab5

packages/framework/src/Model/PriceList/ProductWithPrice.php
grossmannmartin: renamed to `PriceListProductPrice` along with the all usages. 
I had to create a separate commit... the rebase was too painful

b071ff85aa

packages/framework/src/Model/Product/Search/FilterQuery.php
vitek-rostislav: this is still in project-base :pray: (please, do not forget about the upgrade notes as well, then)

vitek-rostislav (approved):

project-base/storefront/public/locales/cs/common.json
```diff
@@ -110,6 +110,7 @@
     "Customized E-shop by": "E-shop na míru od",
     "Czechia": "Česko",
     "Date of order": "Datum objednání",
+    "days": "dnů",
```
JanMolcik: I am thinking about values `1, 2, 3, 4` days in Czech. We should use the pluralisation with `_one`, `_few` and `_many` similar to for example `articles count_*` to support `den`, `dny` and `dnů`.

project-base/storefront/public/locales/en/common.json
```diff
@@ -108,6 +108,7 @@
     "Customized E-shop by": "Customized E-shop by",
     "Czechia": "Czechia",
     "Date of order": "Date of order",
+    "days": "days",
```
JanMolcik: Same as comment above, but for english we would only need `_one` and `_many`.

project-base/storefront/utils/useCountdown.ts
```diff
@@ -0,0 +1,58 @@
...
+    useEffect(() => {
+        const currentTime = dayjs();
+        const endTimeDayjs = dayjs(endTime);
+
+        if (!endTimeDayjs.isValid()) {
+            return;
+        }
+
+        const diffTime = endTimeDayjs.unix() - currentTime.unix();
+        let duration = dayjs.duration(diffTime * 1000, 'milliseconds');
```
JanMolcik: You can possibly use 
```typescript
currentTime.difference(endTimeDayjs, 'milliseconds')
```
to simplify.

JanMolcik (changes requested):

project-base/storefront/public/locales/cs/common.json
chlebektomas: fixed

project-base/storefront/public/locales/en/common.json
chlebektomas: fixed

project-base/storefront/utils/useCountdown.ts
chlebektomas: fixed

JanMolcik (approved):

packages/framework/src/Form/Admin/PriceList/PriceListFormType.php
```diff
@@ -113,7 +113,7 @@ public function configureOptions(OptionsResolver $resolver): void
     public function checkDateValidity(PriceListData $priceListData, ExecutionContextInterface $context): void
     {
         if ($priceListData->validTo < $priceListData->validFrom) {
-            $context->buildViolation('"Valid to" must be greater than "Valid from"')
+            $context->buildViolation(t('"Valid to" must be greater than "Valid from"', [], 'validators'))
```
vitek-rostislav: nitpick, `Translator::VALIDATOR_TRANSLATION_DOMAIN` should be used. I am wondering whether there is a phpstan rule that would add at least a warning when we use a hardcoded string that is already defined in a constant in the codebase :thinking:

vitek-rostislav: https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExZHNuNXphdG1vdTVhOHc5N2l3bHQ1cWVtNXF2azkzd3FzczF2bHFjayZlcD12MV9naWZzX3NlYXJjaCZjdD1n/etSwiUkXNhTB6/giphy.gif

packages/framework/src/Resources/views/Admin/Form/priceListProductsPickerType.html.twig
```diff
@@ -78,9 +82,9 @@
             <span class="js-price-list-product-picker-item-product-name">
                 {% if product is not null %}
                     <a
-                        href="{{ url('admin_product_edit', {id: product.id}) }}"
-                        target="_blank"
-                        class="btn-link-style btn-link-style--iconic-l"
+                            href="{{ url('admin_product_edit', {id: product.id}) }}"
+                            target="_blank"
+                            class="btn-link-style btn-link-style--iconic-l"
```
vitek-rostislav: nitpick, the formatting is weird, there are too many spaces...

packages/framework/src/Model/PriceList/PriceListFacade.php
```diff
@@ -123,6 +123,17 @@ public function delete(int $priceListId): void
     {
         $priceList = $this->getById($priceListId);
 
+        $originalProductIds = array_map(
+            static fn (PriceListProductPrice $priceListProductPrice) => $priceListProductPrice->getProduct()->getId(),
+            $priceList->getPriceListProductPrices(),
+        );
```
vitek-rostislav: nitpick, the same array map is used in the edit function, I would say `PriceList::getProducts()` or `PriceList::getProductIds()` might be a handy function even for the future

packages/product-feed-luigis-box/src/Model/FeedItem/LuigisBoxProductFeedItemFactory.php
```diff
@@ -125,6 +126,20 @@ protected function getPrice(Product $product, DomainConfig $domainConfig): Price
...
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig $domainConfig
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\Price
+     */
+    protected function getBasicPrice(Product $product, DomainConfig $domainConfig): Price
+    {
+        return $this->productPriceCalculationForCustomerUser->calculateBasicPriceForCustomerUserAndDomainId(
+            $product,
+            $domainConfig->getId(),
+            null,
```
vitek-rostislav: it is not necessary to pass `null` here, it is a default value

vitek-rostislav: btw, the same applies to `getPrice()`  above where `calculatePriceForCustomerUserAndDomainId()` is called :slightly_smiling_face:

packages/product-feed-google/src/Model/FeedItem/GoogleFeedItemFactory.php
```diff
@@ -69,13 +74,31 @@ protected function getBrandName(Product $product): ?string
      */
     protected function getPrice(Product $product, DomainConfig $domainConfig): Price
     {
-        return $this->productPriceCalculationForCustomerUser->calculatePriceForCustomerUserAndDomainId(
+        return $this->productPriceCalculationForCustomerUser->calculateBasicPriceForCustomerUserAndDomainId(
             $product,
             $domainConfig->getId(),
             null,
```
vitek-rostislav: here as well, unnecessary null, right?

vitek-rostislav (approved):

packages/framework/src/Form/Admin/PriceList/PriceListFormType.php
grossmannmartin: Well, I'm not sure how the rule should behave. What if we have `SliderItemFormType::SCENARIO_CREATE = 'create';` and I create new button 
```
->add('create', SubmitType::class);
```

How should it detect whether it makes sense to warn about this value or not?

Maybe we could enforce the use of the constant in the `t` function... but it does not bother me that much 🙂 

Anyway, I have changed it

packages/framework/src/Resources/views/Admin/Form/priceListProductsPickerType.html.twig
grossmannmartin: better? 89d8434f25

packages/product-feed-luigis-box/src/Model/FeedItem/LuigisBoxProductFeedItemFactory.php
grossmannmartin: fixed df9800d95d

packages/product-feed-google/src/Model/FeedItem/GoogleFeedItemFactory.php
grossmannmartin: fixed 8a72f1146f

packages/framework/src/Model/PriceList/PriceListFacade.php
grossmannmartin: Nide! thanks d4c04d5843

docs/functional/time-limited-price-lists.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Managing the time-limited price lists
+
+Time-limited price lists can be managed in the `Pricing -> Price lists` section.
+
+Each price list has its own validity period, so it's possible to seamlessly create follow-up price promotions.
+If a product is assigned to the price list, the price is automatically used as a new selling price for the specified time period (no manual action is required).
+
+The special price for the product is used only if the entered price is actually lower than the basic price.
+That allows to create a promotion only for a specific customer group, or deal with the complex pricing strategy.
```
pk16011990:
```suggestion
That allows to create a promotion only for a specific customer group, or deal with a complex pricing strategy.
```

docs/functional/time-limited-price-lists.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Priority of the multiple active price lists
+
+Active price lists are prioritized by the last update date.
+Meaning that the most recently updated price list has the highest priority.
```
pk16011990:
```suggestion
This means that the most recently updated price list has the highest priority.
```

docs/functional/time-limited-price-lists.md
```diff
@@ -0,0 +1,42 @@
...
+The default sorting on the price list page is by the last update date, so the most recently updated price list is at the top.
+
+Currently used price for a current product is highlighted in the price list overview table on the product detail page in the administration.
+
+!!! warning
+
+    When new produdct is added or removed from a price list, the price list is effectively updated and the priority is changed.
```
pk16011990:
```suggestion
    When a new product is added or removed from a price list, the price list is effectively updated, and the priority is changed.
```

docs/functional/time-limited-price-lists.md
```diff
@@ -0,0 +1,42 @@
...
+    Keep that in mind to avoid unexpected behavior.
+
+### Behavior of product variants
+
+Only a specific variant of the product can be assigned to the price list due to the fact that the main variant is not sellable and do not have any price set.
```
pk16011990:
```suggestion
Only a specific variant of the product can be assigned to the price list due to the fact that the main variant is not sellable and does not have any price set.
```

docs/functional/time-limited-price-lists.md
```diff
@@ -0,0 +1,42 @@
...
+
+Each variant price is managed and determined separately.
+That means that each variant can be placed in a different price list and the price of a specific variant is determined by the price list containing the variant with the highest priority.
```
pk16011990:
```suggestion
That means that each variant can be placed on a different price list, and the price of a specific variant is determined by the price list containing the variant with the highest priority.
```

packages/framework/assets/js/admin/validation/form/validationPriceList.js
```diff
@@ -0,0 +1,16 @@
+import Register from '../../../common/utils/Register';
+
+export default function priceListValidator ($container) {
+    window.$('form[name="price_list_form"]').jsFormValidator({
```
pk16011990:
```suggestion
    $container.filterAllNodes('form[name="price_list_form"]').jsFormValidator({
```
To prevent multiple registrations

packages/framework/assets/js/admin/components/PriceListProductPicker.js
```diff
@@ -0,0 +1,212 @@
...
+        FormChangeInfo.showInfo();
+    }
+
+    static init ($container) {
+        $container.filterAllNodes('.js-price-list-product-picker').each(function () {
+            // eslint-disable-next-line no-new
+            new PriceListProductPicker($(this));
+        });
+
+        $('.js-price-list-product-picker-close').click(() => {
```
pk16011990:
```suggestion
        $container.filterAllNodes('.js-price-list-product-picker-close').click(() => {
```

packages/framework/src/Model/PriceList/PriceList.php
```diff
@@ -0,0 +1,168 @@
...
+    public function addPriceListProductPrice(PriceListProductPrice $priceListProductPrice): void
+    {
+        $this->priceListProductPrices->add($priceListProductPrice);
+        $priceListProductPrice->setPriceList($this);
+    }
+
+    /**
+     * @return int[]
+     */
+    public function getProductIds(): array
```
pk16011990, line 161: Wouldn't this be a problem with a large number of entities and memory? Price lists can have thousand(s) of products.

pk16011990 (changes requested):

packages/framework/src/Model/PriceList/PriceList.php
grossmannmartin: I think not, but let's test it on devel after merge

vitek-rostislav (approved):

pk16011990 (approved):

PR #3629, opened 2024-11-30
Create crud skeleton

henzigo:

This PR adds a basis for an easy definition of CRUD administration. It adds a basis for defining CRUD, including the new Datagrid component. 

Testing controllers are here https://github.com/shopsys/shopsys/tree/jg-crud-testing

#### Fixes issues
... 

----
  - https://jg-crud-admin.odin.shopsys.cloud
  - https://cz.jg-crud-admin.odin.shopsys.cloud


docs/administration/crud-controller/getting-started/creating-a-new-crud-controller.md
```diff
@@ -0,0 +1,31 @@
+# Creating a new Crud Controller
+
+This guide will show you how to create a new Crud Controller in your project.
+
+## 1. Create a new Crud Controller
+
+Create a new Crud Controller by extending the `Shopsys\AdministrationBundle\Controller\AbstractCrudController` class and must have defined attribute `#[CrudController]` with the entity class name.
```
vitek-rostislav:
```suggestion
Create a new Crud Controller by extending the `Shopsys\AdministrationBundle\Controller\AbstractCrudController` class and on your class, define attribute `#[CrudController]` with the entity class name.
```

docs/administration/crud-controller/index.md
```diff
@@ -0,0 +1,15 @@
...
+The Crud Controller is a powerful tool that allows you to easily design UI for creating, reading, updating, and deleting records in the database. It is a part of the [Administration](../index.md) module.
+
+It's designed to speed up the development of the administration interface and provide a consistent user experience. At the same time it's flexible enough to allow for customizations.
+
+Crud controller is built using known Shopsys components like [Grid](../grid.md) and Symfony components.
+
+## Getting started
+
+- [Creating a new Crud Controller](getting-started/creating-a-new-crud-controller.md)
+
```
vitek-rostislav: redundant empty line

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -0,0 +1,47 @@
...
+
+use Shopsys\AdministrationBundle\Component\Config\CrudConfig;
+use Shopsys\AdministrationBundle\Component\Config\CrudConfigData;
+use Shopsys\AdministrationBundle\Component\Config\PageType;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Response;
+
+class AbstractCrudController extends AbstractController
+{
+    private ?CrudConfigData $config = null;
```
vitek-rostislav, line 25: is  not there any case when I need to access the `$config` in my controller on a project?

vitek-rostislav: I mean, I can still access it using `getConfig()` method but it seems a little bit cumbersome, or?

docs/administration/crud-controller/reference/crud-controller.md
```diff
@@ -16,6 +16,21 @@ The `CrudConfig` class is used to configure the behavior of the Crud Controller.
...
+#### `setMenuSection(string $menuSection, ?string $submenuSection = null)`
+
+You can specify where the Crud Controller will be displayed in the administration menu.
+
+`$menuSection` is the name of the root-level menu item.
+`$submenuSection` can be used to specify a submenu item.
+
+Examples:
+- `$config->setMenuSection('products')` will create Crud controller item under `Products` section
+- `$config->setMenuSection('customers', 'promo_codes')` will create Crud controller under `Customers -> Promo Codes` section
```
vitek-rostislav, line 58: it is IMHO worth mentioning where to find the menu items' names - AFAIK in `SideMenuBuilder`, right? Too bad there is no command for listing all the items (or I was not able to find such a command)

vitek-rostislav: ... in the "end", those should be defined in `CrudConfig::setMenuTitle` calls, right?

packages/administration/src/Component/Config/CrudConfig.php
```diff
@@ -18,32 +18,78 @@ public function __construct()
...
         return $this;
     }
 
+    /**
+     * Sets the title of the menu item that will be used.
+     *
+     * @param string $menuTitle
+     * @return $this
+     */
+    public function setMenuTitle(string $menuTitle): self
```
vitek-rostislav, line 42: One thought - what if there are controllers that need to have multiple menu entries for multiple actions, is that possible?

docs/administration/crud-controller/reference/crud-controller.md
vitek-rostislav: ... I see there is TODO below that could cover that :slightly_smiling_face:

packages/administration/src/Component/Config/CrudConfigData.php
```diff
@@ -23,25 +25,43 @@ final class CrudConfigData
...
+        $this->entityName = $entityName;
+
+        $this->customPageTitles = [
+            PageType::CREATE->value => t('Creating new %entity_name%', ['%entity_name%' => $entityName]),
+            PageType::EDIT->value => t('Editing %entity_name%', ['%entity_name%' => $entityName]),
+            PageType::LIST->value => t('%entity_name% Overview', ['%entity_name%' => $entityName]),
+            PageType::DETAIL->value => t('Viewing %entity_name%', ['%entity_name%' => $entityName]),
+        ];
+
+        $this->menuTitle = t('%entity_name% Overview', ['%entity_name%' => $entityName]);
```
vitek-rostislav: I see a translation problem here - as `$entityName` is not translated, you will end up with titles like `Vytváření nového product` :thinking:

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -25,16 +34,38 @@ protected function configure(CrudConfig $config): CrudConfig
         return $config;
     }
 
+    /**
+     * @param \Shopsys\AdministrationBundle\Component\Config\Action\ActionsConfig $actions
+     * @return \Shopsys\AdministrationBundle\Component\Config\Action\ActionsConfig
+     */
+    protected function configureActions(ActionsConfig $actions): ActionsConfig
```
vitek-rostislav: I think this deserves a mention in the docs :slightly_smiling_face:

packages/administration/src/Resources/views/crud/new.html.twig
```diff
@@ -0,0 +1,23 @@
+{% extends '@ShopsysFramework/Admin/Layout/layoutWithPanel.html.twig' %}
```
vitek-rostislav: the templates for new end edit action are basically the same, we should create a common ancestor (`detail.html.twig`?) for them to avoid duplicates

packages/administration/src/Component/Config/Action/Builder/AbstractActionBuilder.php
```diff
@@ -0,0 +1,159 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Config\Action\Builder;
+
+use Shopsys\AdministrationBundle\Component\Config\Action\ActionData;
+use Shopsys\AdministrationBundle\Component\Config\Action\ActionRouteType;
+use Shopsys\AdministrationBundle\Component\Config\Action\ActionType;
+use Shopsys\AdministrationBundle\Component\Config\PageType;
+
+abstract class AbstractActionBuilder
```
vitek-rostislav: If I am not mistaken, this class is called `AbstractAction` in the final version. Is there any point in having the abstract class and then one final implementation - `Action` class? Maybe we could merge them?

Or is it to enable adding more action types on a project? :eyes: Let's discuss this, please

docs/administration/datagrid/index.md
```diff
@@ -0,0 +1,60 @@
...
+Datagrid uses some internal interfaces and classes to work properly and to provide decoupled architecture with the possibility to customize its behavior.
+
+### Adapter
+
+Adapters are used to fetch data from different sources. The adapter is responsible for fetching data and transforming it into a format that can be displayed in the Datagrid.
+
+There are currently two adapters available:
+
+- `ArrayAdapter` - Adapter that works with arrays
+- `OrmAdapter` - Adapter that works with Doctrine entities and it's used by Crud Controller
```
vitek-rostislav: link to the crud controller docs would be nice here

docs/administration/datagrid/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+You can also create your own adapter by implementing the `AdapterInterface`.
```
vitek-rostislav: FQCN would be nice here

docs/administration/datagrid/index.md
vitek-rostislav: + FQCN please

docs/administration/datagrid/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+Datagrid uses some internal interfaces and classes to work properly and to provide decoupled architecture with the possibility to customize its behavior.
+
+### Adapter
+
+Adapters are used to fetch data from different sources. The adapter is responsible for fetching data and transforming it into a format that can be displayed in the Datagrid.
+
+There are currently two adapters available:
+
+- `ArrayAdapter` - Adapter that works with arrays
```
vitek-rostislav: FQCN please

docs/administration/datagrid/index.md
```diff
@@ -0,0 +1,60 @@
...
+- `OrmAdapter` - Adapter that works with Doctrine entities and it's used by Crud Controller
+
+You can also create your own adapter by implementing the `AdapterInterface`.
+
+### Datagrid class
+
+The `Datagrid` class is the main class that is used to configure the Datagrid. It provides methods to configure columns, filters, actions, and other features.
```
vitek-rostislav: FQCN please

docs/administration/datagrid/index.md
```diff
@@ -0,0 +1,60 @@
...
+```php
+// For Array
+
+// Shopsys\AdministrationBundle\Component\Datagrid\Adapter\Array\ArrayAdapterFactory
+$adapter = $this->arrayAdapterFactory->create($array);
+
+
+// For ORM 
+
+// Shopsys\AdministrationBundle\Component\Datagrid\Adapter\Orm\OrmAdapterFactory
```
vitek-rostislav: it might be a little bit confusing whether you are referencing to `$adapter` or `$this->ormAdapterFactory`

docs/administration/datagrid/index.md
```diff
@@ -0,0 +1,60 @@
...
+$adapter = $this->ormAdapterFactory->create($entityClass);
+```
+
+2. Create Datagrid instance via Factory:
+
+```php
+
+// Shopsys\AdministrationBundle\Component\Datagrid\DatagridFactory
```
vitek-rostislav: it might be a little bit confusing whether you are referencing to `$datagrid` or `$this->datagridFactory`

docs/administration/index.md
```diff
@@ -2,8 +2,9 @@
 
 - [Crud Controller](./crud-controller/index.md)
 - [Administration Menu](./administration-menu.md)
-- [Grid](./grid.md)
-    - [Grid Rendering Customization](./grid-rendering-customization.md)
-    - [Grid Data Sources](./grid-data-sources.md)
+- [Datagrid](./datagrid/index.md)
+- [Grid (Legacy)](./legacy-grid/index.md)
```
vitek-rostislav: is it legacy indeed? I would rather say it is internal, or?

packages/administration/src/Component/Datagrid/Adapter/Orm/DatagridDataSource.php
```diff
@@ -0,0 +1,63 @@
...
+        ?string $orderSourceColumnName = null,
+        string $orderDirection = self::ORDER_ASC,
+    ): PaginationResult {
+        $queryBuilder = clone $this->queryBuilder;
+
+        if ($orderSourceColumnName !== null) {
+            $this->addQueryOrder($queryBuilder, $orderSourceColumnName, $orderDirection);
+        }
+
+        $queryPaginator = new QueryPaginator($queryBuilder, DatagridHydrator::HYDRATION_MODE);
```
vitek-rostislav, line 50: this (possibility to set the hint) should be taken into account now - https://github.com/shopsys/shopsys/pull/3641/commits/6ba3514642816d8933dcd7ff7938af9124c1668a

packages/administration/src/Component/Datagrid/Adapter/Orm/DatagridDataSource.php
```diff
@@ -0,0 +1,63 @@
...
+
+        $originalPaginationResult = $queryPaginator->getResult($page, $limit, $this->getTotalRowsCount());
+        $results = $originalPaginationResult->getResults();
+
+        foreach ($results as $key => $result) {
+            $results[$key] = call_user_func($this->manipulateRowCallback, $result, $results[$key], $results);
```
vitek-rostislav: this should be taken into account - https://github.com/shopsys/shopsys/pull/3641/commits/3aa026a0a94fb8d0fabebc39b5fc7fa48272506b - working with all entries in the callback

vitek-rostislav: ah, it already is, right? :slightly_smiling_face:

packages/administration/src/Component/Datagrid/Adapter/Orm/OrmAdapter.php
```diff
@@ -0,0 +1,77 @@
...
+     * @return \Shopsys\AdministrationBundle\Component\Datagrid\Adapter\Orm\ProxyQuery
+     */
+    private function createProxyQuery(string $entityClass): ProxyQuery
+    {
+        /** @var \Doctrine\ORM\EntityManager $entityManager */
+        $entityManager = $this->managerRegistry->getManagerForClass($entityClass);
+        $classMetadata = $entityManager->getClassMetadata($entityClass);
+
+        if (count($classMetadata->getIdentifierFieldNames()) !== 1) {
+            throw new RuntimeException('Crud controller does not support entities with composite primary keys.');
```
vitek-rostislav, line 76: is there a use case in our codebase now? I mean, what if we want to create a grid for such an entity?

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
```diff
@@ -0,0 +1,174 @@
...
+        private QueryBuilder $queryBuilder,
+        private readonly string $locale,
+    ) {
+        $this->rootAlias = current($this->queryBuilder->getRootAliases());
+        $this->queryBuilder->resetDQLPart('select');
+    }
+
+    public function __clone()
+    {
+        $this->queryBuilder = clone $this->queryBuilder;
```
vitek-rostislav, line 52: is not this confusing? when I call `clone` on the class, its query builder property is actually cloned instead?

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
```diff
@@ -0,0 +1,174 @@
...
+                if ($currentClassMetadata->hasField($part)) {
+                    $this->queryBuilder->addSelect("{$alias}.{$part}" . ' AS ' . $this->getAlias($path));
+
+                    continue;
+                }
+
+                try {
+                    // If entity does not have field, check if it has association to translations and try to select from translations
+                    $this->joinAssociation($currentClassMetadata, $path . '_tr', 'translations', $alias, $alias . '_tr');
+                    $this->queryBuilder->addSelect("{$alias}_tr.{$part}" . ' AS ' . $this->getAlias($path));
```
vitek-rostislav: do not we want the same for `*Domain` associations?

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
```diff
@@ -0,0 +1,174 @@
...
+
+                    continue;
+                } catch (InvalidArgumentException $exception) {
+                    // We don't need to throw exeption here
```
vitek-rostislav: ...becauseee..? :smile:

vitek-rostislav: ...it is handled a few lines below?

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
```diff
@@ -0,0 +1,174 @@
...
+                throw new InvalidArgumentException('Field "' . $part . '" not found in entity ' . $currentClassMetadata->getName());
+            }
+
+            $this->joinAssociation($currentClassMetadata, $path, $part, $alias, $joinAlias);
+
+            $currentClassMetadata = $this->entityManager->getClassMetadata(
+                $currentClassMetadata->getAssociationTargetClass($part),
+            );
+
+            $alias = $joinAlias;
```
vitek-rostislav, line 117: what is this good for in the last array iteration?

packages/administration/src/Component/Datagrid/DatagridFactory.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * @param \Shopsys\AdministrationBundle\Component\Datagrid\Adapter\AdapterInterface $adapter
+     * @return \Shopsys\AdministrationBundle\Component\Datagrid\Datagrid
+     */
+    public function create(AdapterInterface $adapter): Datagrid
+    {
+        $datagrid = new Datagrid($adapter, $this->gridFactory);
+
+        return $datagrid;
```
vitek-rostislav: nitpick, the return can be inlined, right?

packages/administration/src/Component/Doctrine/DatagridHydrator.php
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Shopsys\AdministrationBundle\Component\Doctrine;
+
+use Doctrine\ORM\Internal\Hydration\AbstractHydrator;
+
+final class DatagridHydrator extends AbstractHydrator
+{
+    public const HYDRATION_MODE = 'DatagridHydrator';
+
+    protected function hydrateAllData()
```
vitek-rostislav: missing return typehint

vitek-rostislav (changes requested):

docs/administration/crud-controller/reference/crud-controller.md
```diff
@@ -35,4 +65,27 @@ It's used if you want to create Crud Controller, but you don't want to be visibl
 
 This method defines actions that will be created for the given entity.
 
-For example, if you call `$config->setActions([PageType::LIST, PageType::EDIT])` the Crud Controller will create routes, buttons, etc. only for list and edit actions.
\ No newline at end of file
+For example, if you call `$config->setActions([PageType::LIST, PageType::EDIT])` the Crud Controller will create routes, buttons, etc. only for list and edit actions.
+
+#### `disable(bool $disabled)`
+
+Fully disable the Crud Controller. Crud controller will not be accessible by the user.
```
vitek-rostislav, line 72: Could you please add info on what this is useful for?

packages/administration/src/Component/Config/Action/ActionsConfig.php
```diff
@@ -4,122 +4,107 @@
...
-        $this->add(PageType::DETAIL, $backToListAction);
-        $this->add(PageType::CREATE, $backToListAction);
+        $this->add(ActionType::EDIT, $backToListAction);
+        $this->add(ActionType::DETAIL, $backToListAction);
+        $this->add(ActionType::CREATE, $backToListAction);
     }
 
     /**
-     * Add action to be displayed on specific pageType
+     * Add action to be displayed on specific Action
```
vitek-rostislav: on specific Action? Should it not be something like "controller" or "page" instead of action?

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
```diff
@@ -0,0 +1,94 @@
...
+    protected ?string $icon = null;
+
+    protected string $cssClass = '';
+
+    protected ?ActionRouteInterface $actionRoute = null;
+
+    /**
+     * @var callable|null
+     */
+    protected $displayIf = null;
```
vitek-rostislav: default value is not not necessary here

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
```diff
@@ -0,0 +1,94 @@
...
+     * @param string $label
+     * @return $this
+     */
+    abstract public static function create(string $name, string $label): self;
+
+    /**
+     * @param string $name
+     * @param string $label
+     */
+    protected function __construct(protected string $name, string $label)
```
vitek-rostislav: $name is not used

packages/administration/src/Component/Config/Action/Builder/Action.php
```diff
@@ -0,0 +1,85 @@
...
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Config\Action\Builder;
+
+use Shopsys\AdministrationBundle\Component\Config\Action\Builder\ActionRoute\CrudActionRouteData;
+use Shopsys\AdministrationBundle\Component\Config\Action\Builder\ActionRoute\RouteActionRouteData;
+use Shopsys\AdministrationBundle\Component\Config\Action\Builder\ActionRoute\UrlActionRouteData;
+use Shopsys\AdministrationBundle\Component\Config\ActionType;
+
+final class Action extends AbstractAction
```
vitek-rostislav: I am not sure whether it makes sense to create an abstract class here that has only one final implementation - do we expect project developers to create their own implementations?

packages/administration/src/Component/Config/Action/Builder/ActionRoute/RouteActionRouteData.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * @var array|callable(?object $entity): array
+     */
+    public $routeParameters;
+
+    /**
+     * @param string $routeName
+     * @param array|callable $routeParameters
+     * @param array|callable(?object $entity): array $routeParameters
```
vitek-rostislav: duplicated param

packages/administration/src/Component/Config/CrudConfig.php
```diff
@@ -46,15 +46,43 @@ public function setMenuTitle(string $menuTitle): self
...
+     * @param \Shopsys\AdministrationBundle\Component\Config\ActionType|\Shopsys\AdministrationBundle\Component\Config\ActionType[] $actions
+     * @return $this
+     */
+    public function disableAction(ActionType|array $actions): self
+    {
+        if (!is_array($actions)) {
+            $actions = [$actions];
+        }
+
+        Assert::allIsInstanceOf($actions, ActionType::class, 'The given action is not a valid action type');
```
vitek-rostislav, line 82: could be extracted into a separate method to avoid DRY

packages/administration/src/Component/Config/CrudConfigData.php
```diff
@@ -80,11 +87,47 @@ public function getMenuTitle(): string
...
-     * @return \Shopsys\AdministrationBundle\Component\Config\PageType[]
+     * @return \Shopsys\AdministrationBundle\Component\Config\ActionType[]
+     */
+    public function getActions(): array
+    {
+        if ($this->fullDisabled === true) {
+            return [];
+        }
+
+        return $this->enabledActions->toArray();
```
vitek-rostislav: getValues?

packages/administration/src/Resources/views/crud/inline/actions.html.twig
```diff
@@ -0,0 +1,13 @@
+{% macro render(actions) %}
+    <div class="wrap-bar__buttons">
+        {% for action in actions %}
+            <a href="{{ action.url }}" class="btn btn--plus wrap-bar__btn {{ action.cssClass }}">
```
vitek-rostislav: is `btn--plus` wanted every time? :thinking:

vitek-rostislav: (should be also dependent on the specified icon, right?

docs/administration/datagrid/configuration.md
```diff
@@ -0,0 +1,26 @@
+# Configuration
+
+
+
```
vitek-rostislav: too many empty lines here

packages/administration/src/DependencyInjection/Compiler/LoadControllersExtensionCompilerPass.php
```diff
@@ -0,0 +1,100 @@
...
+     * @param \ReflectionClass $reflectionClass
+     * @return \Shopsys\AdministrationBundle\Component\Attributes\CrudControllerExtension|null
+     */
+    private function getCrudControllerExtensionAttribute(ReflectionClass $reflectionClass): ?CrudControllerExtension
+    {
+        $attributes = $reflectionClass->getAttributes(CrudControllerExtension::class);
+
+        if (count($attributes) !== 0) {
+            /** @var \Shopsys\AdministrationBundle\Component\Attributes\CrudControllerExtension $attributeInstance */
+            $attributeInstance = $attributes[0]->newInstance();
```
vitek-rostislav, line 93: is not it dangerous to expect the first attribute will be `CrudControllerExtension`? There might be anything else in theory, or?

docs/administration/datagrid/fields.md
```diff
@@ -0,0 +1,83 @@
...
+## Configuration
+
+Fields are defined by the `Datagrid` class using the `add(string $name, array $options)`, `update(string $name, array $options)` and `remove(string $name)` methods. The `$name` parameter is the name of the field and the `$options` parameter is an array of options that define the behavior of the field.
+
+In case you are using `OrmAdapter`, you can use dot notation to access nested properties of the entity. Here are some examples of how to define field names:
+
+- `$datagrid->add('domainId')` - This will fetch the `domainId` property of the entity.
+- `$datagrid->add('currency.code')` - This will join the `currency` relation and fetch the `code` property of the `Currency` entity.
+- `$datagrid->add('status.name')` - This will join the `status` relation and look for the `name` property of the `Status` entity. In this case `Status` is a translatable entity and the `name` property is a translatable field. The datagrid will automatically fetch the correct translation based on the current locale.
+- `$datagrid->add('currency')` - This will fetch the `Currency` entity. In this case you need to define `template` or `transform` option to display the data in the field.
```
vitek-rostislav: or you can set it as not visible, right? Then you can access it in the template or something like this, or?

docs/administration/datagrid/configuration.md
```diff
@@ -41,6 +41,31 @@ $datagrid->update('name', [
 // Remove column from the datagrid
 $datagrid->remove('name');
 
+// Access to action configuration
```
vitek-rostislav: docs/administration/crud-controller/reference/actions.md and docs/administration/datagrid/actions.md are now empty, so I am posting a TODO here :slightly_smiling_face:

packages/administration/src/Component/Datagrid/DatagridActions.php
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * @phpstan-type ActionOptions array{
+ *     label?: string,
+ *     icon?: string,
+ *     routeName?: string,
+ *     additionalParameters?: array,
+ *     confirmMessage?: string,
+ * }
+ */
+final class DatagridActions
```
vitek-rostislav: maybe this should be named rather `DatagridRowActions` as in the future we might want to add actions for the grid as a whole (or is it already there? I am not sure)

packages/administration/src/Component/Datagrid/DatagridManager.php
```diff
@@ -27,4 +30,12 @@ public function createGrid(mixed $name, DataSourceInterface $query): Grid
     {
         return $this->gridFactory->create($name, $query);
     }
+
+    /**
+     * @return \Shopsys\AdministrationBundle\Component\Router\CrudRouteProvider
+     */
+    public function getCrudRouteProvider(): CrudRouteProvider
+    {
+        return $this->crudRouteProvider;
```
vitek-rostislav: I am not a big fan of having getters for services that can be autowired where needed :shrug:

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -114,9 +114,10 @@ public function createAction(): Response
     }
 
     /**
+     * @param int $id
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function deleteAction(): Response
+    public function deleteAction(int $id): Response
     {
         return $this->redirect($this->generateUrl('admin_default_dashboard'));
```
vitek-rostislav, line 135: implementing the delete action is not a part of this user story? :innocent:

packages/administration/src/Resources/views/crud/inline/actions.html.twig
```diff
@@ -1,10 +1,12 @@
...
         {% for action in actions %}
-            <a href="{{ action.url }}" class="btn btn--plus wrap-bar__btn {{ action.cssClass }}">
+            <a href="{{ action.url }}"
+               class="btn wrap-bar__btn {{ action.cssClass }}"
+               {% if action.openInNewTab %}target="_blank"{% endif %}
+            >
                 {% if action.icon is not null %}
-                    {# TODO: Replace by defined icon #}
-                    <i class="btn__icon">+</i>
+                    <i class="btn__icon svg svg-{{ action.icon }}"></i>
```
vitek-rostislav: needs to be updated now after https://github.com/shopsys/shopsys/pull/3765 is merged

project-base/app/src/Controller/Admin/TestController.php
```diff
@@ -0,0 +1,155 @@
...
+     */
+    protected function configureActions(ActionsConfig $actions): ActionsConfig
+    {
+        $actions->add(
+            ActionType::LIST,
+            Action::create('linkToDashboard', t('Link To dashboard'))
+                ->setCssClass('btn--primary')
+                ->setOpenInNewTab()
+                ->linkToRoute('admin_default_dashboard', fn () => [
+                    'id' => 1,
```
vitek-rostislav: this does not make sense here but I suppose it does not matter :slightly_smiling_face:

vitek-rostislav (changes requested):
Awesome job there, love it! :heart_eyes:

docs/administration/crud-controller/getting-started/creating-a-new-crud-controller.md
henzigo: fixed

docs/administration/crud-controller/index.md
henzigo: fixed

packages/administration/src/Controller/AbstractCrudController.php
henzigo: This is only used as a cache for that config. You can't access config property in your controller and we should also use getConfig() method instead of this private property inside that AbstractCrudController

docs/administration/crud-controller/reference/crud-controller.md
henzigo: We should refactor SideMenuBuilder in the future so that it correctly works with CRUD. For now, this is a different problem that I'm not capable of changing in this PR.

packages/administration/src/Component/Config/CrudConfig.php
henzigo: This is a very specific edge case that should be probably implemented differently. For now, every CrudController has one menu item.

packages/administration/src/Component/Config/CrudConfigData.php
henzigo: Yes exactly, it's designed to work without any additional configuration, but ofc you should configure some basic behavior like `setMenuTitle`

packages/administration/src/Controller/AbstractCrudController.php
henzigo: This is probably the last section that is missing in the docs. I will finish it after this first round of CR.

packages/administration/src/Resources/views/crud/new.html.twig
henzigo: I would like to leave it as it is for now.

docs/administration/datagrid/index.md
henzigo: FCQN updated. 

> link to the crud controller docs would be nice here

I don't think there should be a link to the crud controller as it's a standalone component. We can discuss it.

docs/administration/datagrid/index.md
henzigo: fixed

docs/administration/datagrid/index.md
henzigo: fixed

docs/administration/datagrid/index.md
henzigo: fixed

docs/administration/index.md
henzigo: I want to completely refactor this internal grid and remove it in near future as part of the new administration. That's why It's called legacy.

packages/administration/src/Component/Datagrid/Adapter/Orm/OrmAdapter.php
henzigo: We shouldn't as there are some issues with it when we want to use classMetadata with composite primary keys.

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
henzigo: Maybe when we will add support for DomainTabs.

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
henzigo: I completely refactored ProxyQuery :D

packages/administration/src/Component/Datagrid/DatagridFactory.php
henzigo: fixed

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
henzigo: completely refactored ProxyQuery :D

packages/administration/src/Component/Doctrine/DatagridHydrator.php
henzigo: outdated

project-base/app/src/Controller/Admin/TestController.php
henzigo: It does make sense as I wanted to check if passed parameters are working as expected :D

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
```diff
@@ -9,6 +9,14 @@
 use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
 use InvalidArgumentException;
+use RuntimeException;
+
+enum PartType
```
vitek-rostislav: This is quite unusual for me, I would rather see this in a separate file

packages/administration/src/Component/Config/CrudConfigData.php
vitek-rostislav: ok

packages/administration/src/Resources/views/crud/inline/actions.html.twig
henzigo: updated

packages/administration/src/Component/Datagrid/Adapter/Orm/DatagridDataSource.php
henzigo: I updated my code for including meta columns little bit. Look 8ac5bbc4

packages/administration/src/Component/Datagrid/Adapter/Orm/DatagridDataSource.php
henzigo: yes already used :)

docs/administration/crud-controller/reference/crud-controller.md
henzigo: I can add that as info that you can disable it when you for an example don't want to use this functionality in your project, but we probably don't need to write all possibilities for which this function is designed.

packages/administration/src/Component/Config/Action/Builder/Action.php
henzigo: It creates borders for programmers to edit/update/implement some custom behavior. For now, it seems useless, but with a new admin, we will probably want to create DropdownAction or something like that.

packages/administration/src/Component/Config/Action/Builder/ActionRoute/RouteActionRouteData.php
henzigo: fixed

packages/administration/src/Component/Config/CrudConfig.php
henzigo: I thought about it and in final I decided to not avoid this DRY as it will decrease readability. As the second argument is that this code will be probably modified in the future in favor of custom actions

packages/administration/src/Component/Config/CrudConfigData.php
henzigo: fixed

packages/administration/src/Resources/views/crud/inline/actions.html.twig
henzigo: `btn--plus` class was already removed. Anyway, I want to make some changes to rendering Actions so let's keep this conversation opened for now

docs/administration/datagrid/configuration.md
henzigo: fixed

packages/administration/src/DependencyInjection/Compiler/LoadControllersExtensionCompilerPass.php
henzigo: `$attributes` contains only `CrudControllerExtension` classes as it's filtered before those lines with:

```php
$attributes = $reflectionClass->getAttributes(CrudControllerExtension::class);
```

packages/administration/src/Component/Datagrid/DatagridManager.php
henzigo: I will refactor this

packages/administration/src/Controller/AbstractCrudController.php
henzigo: It is, but It's willingly ignored

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
henzigo: fixed

docs/administration/datagrid/fields.md
henzigo: updated

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
vitek-rostislav: instead of callable you can use Closure btw

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
henzigo: As we talked, QueryBuilder is mutable so without it any changes in cloned class would also affect original QueryBuilder

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
henzigo: I changed callable to Closure :) ... but now I have problem with standards :D

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
henzigo: It is now :)

packages/administration/src/Controller/AbstractCrudController.php
henzigo: Documented 🎉

docs/administration/datagrid/index.md
henzigo: Link added

docs/administration/datagrid/index.md
henzigo: Is it better now? :)

docs/administration/datagrid/index.md
henzigo: Also refactored

packages/administration/src/Component/Config/Action/ActionsConfig.php
henzigo: Is it better now?

docs/administration/datagrid/index.md
henzigo: Also refactored

docs/administration/index.md
henzigo: Edited to Internal :)

packages/administration/src/Component/Datagrid/Adapter/Orm/ProxyQuery.php
henzigo: For others. Without this line of code QueryBuilder would do this:

```php
$original = new ProxyQuery(...);
$copy = clone $original;

// Modify the cloned query
$copy->addSelect('someField');

// If we didn't clone QueryBuilder, this would also modify the original!
print_r($original->getQueryBuilder()->getDQLPart('select'));
```

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
```diff
@@ -32,6 +28,17 @@ abstract class AbstractAction
...
+    /**
+     * @return string
+     */
+    abstract protected function getTemplate(): string;
+
+    /**
+     * @param object|null $entity
+     * @return array<string, mixed>
+     */
+    abstract protected function getTemplateParameters(?object $entity): array;
```
vitek-rostislav: is $entity a proper name?

vitek-rostislav: or otherwise, should it be named `getTemplateParameters` when you need to pass the entity?

packages/administration/src/Component/Config/Action/Builder/Action.php
```diff
@@ -15,6 +15,8 @@ final class Action extends AbstractAction
 {
     private ?ActionRouteInterface $actionRoute = null;
 
+    private bool $openInNewTab = false;
```
vitek-rostislav: why is it moved from the abstract class here?

docs/administration/crud-controller/reference/actions.md
```diff
@@ -0,0 +1,140 @@
...
+The Crud Controller provides a set of default actions that are displayed based on the Crud Controller configuration and specific page.
+
+!!! note
+
+    Default actions are displayed only if corresponding conditions are met (e.g. action is enabled).
+
+
+### New action
+
+- **Application Name**   : `create`
```
vitek-rostislav: should not this be "Action name"?

docs/administration/crud-controller/reference/actions.md
```diff
@@ -0,0 +1,140 @@
...
+- **Action**             : Redirects to the `ActionType::CREATE` of the current Crud Controller.
+- **Displayed on**       : `ActionType::LIST` page
+
+### Back to list action
+
+- **Application Name**   : `backToList`
```
vitek-rostislav: should not this be "Action name"?

docs/administration/crud-controller/reference/actions.md
```diff
@@ -0,0 +1,140 @@
...
+```php
+// ...
+
+use App\Model\Article\Article;
+use Shopsys\AdministrationBundle\Component\Config\Action\ActionsConfig;
+use Shopsys\AdministrationBundle\Component\Config\Action\Builder\AbstractAction;
+use Shopsys\AdministrationBundle\Component\Config\Action\Builder\Action;
+use Shopsys\AdministrationBundle\Component\Config\ActionType;
+
+// ...
```
vitek-rostislav, line 62: in which class is this supposed to be?

docs/administration/crud-controller/reference/actions.md
```diff
@@ -0,0 +1,140 @@
...
+```php
+// ...
+
+use App\Controller\Admin\ArticleController;
+use App\Model\Article\Article;
+use Shopsys\AdministrationBundle\Component\Config\Action\ActionsConfig;
+use Shopsys\AdministrationBundle\Component\Config\Action\Builder\Action;
+use Shopsys\AdministrationBundle\Component\Config\ActionType;
+
+// ...
```
vitek-rostislav, line 105: in which class is this supposed to be?

docs/administration/index.md
```diff
@@ -3,7 +3,7 @@
 - [Crud Controller](./crud-controller/index.md)
 - [Administration Menu](./administration-menu.md)
 - [Datagrid](./datagrid/index.md)
-- [Grid (Legacy)](./legacy-grid/index.md)
+- [Grid (Internal)](./legacy-grid/index.md)
```
vitek-rostislav: the folder deserves renaming too :innocent:

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
```diff
@@ -81,27 +96,44 @@ public function setIcon(string $icon): self
...
+     * @param mixed $value
+     * @param bool $append If true, value will be appended to existing value. If false, existing value will be replaced.
      * @return $this
      */
-    public function displayIf(Closure $function): self
+    public function setAttribute(string $name, mixed $value, bool $append = false): self
     {
-        $this->displayIf = $function;
+        if (in_array($name, $this->forbiddenAttributes, true)) {
+            throw new InvalidArgumentException(sprintf('Attribute "%s" is forbidden to set.', $name));
```
vitek-rostislav: maybe it is worth mentioning something like "set this attribute using whateverFunction()"

vitek-rostislav (changes requested):

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
vitek-rostislav: I do not think so :smile:

henzigo: It's used [here](https://github.com/shopsys/shopsys/blob/cafff15df3311abd78fc98282bcaa6888074fcde/packages/administration/src/Component/Config/Action/ActionsConfig.php#L60-L62) :)

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
henzigo: Do you have any suggestions, on how to avoid passing that parameter? In the other hand, it still says what it does. Like you will get TemplateParameters

packages/administration/src/Component/Config/Action/Builder/Action.php
henzigo: Two reasons:

1. AbstractAction knows nothing about links so this setting is just weird. You can set it as open in a new tab, but you can't configure a link
2. I was wondering what would a DropdownAction look like, which would not have any link itself, but only a list of other Actions. And then it doesn't make sense either

docs/administration/crud-controller/reference/actions.md
henzigo: On top of this file is mentioned this:

Actions can be configured in `configureActions()` method of the Crud Controller.

```php
namespace App\Controller\Admin;

use App\Model\Article\Article;
use Shopsys\AdministrationBundle\Component\Attributes\CrudController;
use Shopsys\AdministrationBundle\Component\Config\Action\ActionsConfig;
use Shopsys\AdministrationBundle\Controller\AbstractCrudController;

#[CrudController(Article::class)]
class ArticleController extends AbstractCrudController
{
    // ...

    protected function configureActions(ActionsConfig $actions): ActionsConfig
    {
        
    }
}
```

So I think it's not necessary to specify everything in that sample :D

docs/administration/crud-controller/reference/actions.md
henzigo: Same as above

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
henzigo: I can update the message to something like ... is forbidden to set. Look if any method sets this attribute... 

But `$this->forbiddenAttributes` can be updated by any child class so your suggestion could introduce additional overhead over configuring these values.

docs/administration/crud-controller/reference/actions.md
henzigo: yes, fixed

docs/administration/crud-controller/reference/actions.md
henzigo: yes, fixed

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
henzigo: Eventually, I got the idea. `$this->forbiddenAttributes` now accepts key as name of attribute and value as an additional error message.

docs/administration/index.md
henzigo: Done

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
vitek-rostislav: ah, ok, I was confused by the constructor format - `$label` is not promoted but is initialized in the constructor body, while `$name` is a constructor-promoted property, ok, now I get it. It is a bit confusing, could it be unified, please?

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
vitek-rostislav: yeah, my bad, I probably misunderstood the method in the first place. However, now, when I see the usage, what about this?

image

packages/administration/src/Component/Config/Action/Builder/Action.php
vitek-rostislav: ok, thanks for the clarification

docs/administration/crud-controller/reference/actions.md
vitek-rostislav: you are right :+1:

packages/administration/src/Component/Config/Action/Builder/AbstractAction.php
henzigo: fixed

docs/administration/datagrid/configuration.md
henzigo: Both sections contains some text now :)

packages/administration/src/Component/Datagrid/DatagridActions.php
henzigo: I refactored Datagrid actions to DatagridRowActions and they are now defined by same AbstractAction class as CrudActions :)

packages/administration/src/Component/Datagrid/DatagridManager.php
henzigo: Ok after refactoring DatagridRowActions Datagrid needs only `Shopsys\FrameworkBundle\Component\Grid\GridFactory` that is now passed by the constructor :) So `DatagridManager` is no longer available

docs/administration/datagrid/row-actions.md
```diff
@@ -0,0 +1,88 @@
...
+    // or any other attribute (Some attributes are used internally and cannot be overridden)
+    ->setAttribute('data-attribute', 'value')
+    // You can conditionally display action only when you need it
+    ->displayIf(fn (array $row) => $row['domainId'] === 1)
+    // By default, the tooltip is displayed after hovering over the action. You can disable it if you want
+    ->showTooltip(false)
+    ;
+
+
+// Let's look about the methods that will help you with redirecting the user after clicking the
```
vitek-rostislav: this sentence is somehow broken, "Let's look about" does not sound correct, and there is no ending :sweat_smile:

docs/administration/datagrid/row-actions.md
```diff
@@ -0,0 +1,88 @@
...
+    ->showTooltip(false)
+    ;
+
+
+// Let's look about the methods that will help you with redirecting the user after clicking the 
+
+$rowAction
+    // Creates a link to a specific route. As the second argument you should specify parameters with an array or Closure
+    ->linkToRoute('admin_article_publish', fn (array $row) => [ 'id' => $row['id'] ])
+    // Use this in case you need to create an link to an external page
```
vitek-rostislav:
```suggestion
    // Use this in case you need to create a link to an external page
```

nitpick :slightly_smiling_face:

packages/administration/src/Component/Action/Action.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Action;
+
+final class Action extends AbstractAction
+{
+    use ActionRouteTrait;
```
vitek-rostislav: I would rather see all the trait logic directly in `AbstractAction`  :shrug: It seems like both the implementations need all the functionality defined in the trait, which is IMHO just complicating things...Or am I missing something?

packages/administration/src/Component/Datagrid/Datagrid.php
```diff
@@ -25,7 +26,7 @@ final class Datagrid
      */
     private ArrayCollection $fields;
 
-    private DatagridActions $actions;
+    private DatagridRowActions $rowActions;
```
vitek-rostislav: it feels like `DatagridRowActions` class is adding an unnecessary layer of complexity. IMHO, `Datagrid` should offer methods for manipulating the row actions (presented by his private `ArrayCollection` of `RowAction` instances (or probably `GridRowActionInterface` instances, right?)

vitek-rostislav: e.g. this usage is quite cumbersome:
```
foreach ($this->rowActions()->getRowActions() as $action) {
            $grid->addRowAction($action);
        }
```

vitek-rostislav: btw now it is not consistent with the `$fields` property

packages/framework/src/Component/Grid/GridView.php
```diff
@@ -149,23 +149,42 @@ public function renderCell(Column $column, ?array $row = null, ?FormView $formVi
...
+                return;
+            }
+
+            $renderData = $actionColumn->renderData();
+            echo $this->twig->render($renderData['template'], [...$renderData['parameters'], 'row' => $row]);
+
+            return;
+        }
+
+        throw new InvalidArgumentException(
```
vitek-rostislav: this is quite weird - you declare `GridRowColumnInterface` can be passed to `renderActionCell`, however, only `ActionColumn`|`GridRowActionInterface` are actually supported. Could we discuss it, please?

vitek-rostislav (changes requested):

docs/administration/datagrid/row-actions.md
henzigo: I deleted that message :)

docs/administration/datagrid/row-actions.md
henzigo: fixed

packages/administration/src/Component/Datagrid/Datagrid.php
henzigo: As we agreed, I changed `rowActions()` to `actions()` method

packages/administration/src/Component/Action/Action.php
henzigo: with a heavy heart adjusted

packages/framework/src/Component/Grid/GridView.php
henzigo: refactored :)

vitek-rostislav (changes requested):
So, hopefully last return of this task :smile: 
Please,
- rebase the branch,
- squash the fixups, 
- resolve the failing builds, 
- and add tests at least for the `ProxyQuery` class

docs/administration/crud-controller/getting-started/configure-list-page.md
```diff
@@ -1 +1,117 @@
-## Configure List Page
\ No newline at end of file
+## Configure List Page
```
grossmannmartin, line 1: the documentation is somehow weirdly structured. Link `Getting started` point to `Configure list page`

docs/administration/crud-controller/getting-started/creating-a-new-crud-controller.md
```diff
@@ -35,30 +35,11 @@ Now you can implement `configure()` method that allows you to customize some gen
 protected function configure(CrudConfig $config): CrudConfig
 {
     return $config
-        ->setTitle(PageType::LIST, t('My new Crud Controller'))
-        ->setMenuSection('customers')
-        ->disableAction(ActionType::DELETE)
+        ->setTitle(PageType::LIST, t('My new Crud Controller')) // Set the title of the list page
```
grossmannmartin: it's actually `ActionType` in the latest implementation... and it seems weird to use actionType here, but I can live with that

grossmannmartin: Is the main idea here that we would be able to call it multiple times to set each page? 
like
```
->setTitle(ActionType::LIST, t('List products'))
->setTitle(ActionType::EDIT, t(Edit product'))
```

docs/administration/crud-controller/getting-started/configure-list-page.md
```diff
@@ -1 +1,117 @@
...
+The first step is to define a datagrid to display the list of records. The list of methods for configuring a datagrid can be found in the [Datagrid Methods Configuration](../../datagrid/configuration.md#methods-configuration) reference.
+
+To start, we can define a simple datagrid with three columns: `id`, `number`, and `createdAt`.
+
+```php
+
+public function configureDatagrid(Datagrid $datagrid): Datagrid
+{
+    $datagrid
+        // This column is defined automatically and is used to identify the record. You can override it if you want to display a different identifier or you can hide it with second argument set to false.
```
grossmannmartin: the default value is already false, which makes the behavior quite strange now. 

datagrid without call `addIdentifier` do not display ID column
datagrid with `->addIdentifier('id')` **do not display ID column**
datagrid with `->addIdentifier('id', true)` *display ID column*

what about split the responsibility to `changeIdentifier(string $idColumnName)` and if I want to display column ID in grid I will use default `addColumn(...)`?

docs/administration/crud-controller/getting-started/configure-list-page.md
```diff
@@ -1 +1,117 @@
...
+    ;
+
+    return $datagrid;
+}
+```
+!!! note
+
+    Every column can be configured by passing an array with options. The list of available options can be found in the [Fields Configuration](../../datagrid/fields.md) reference.
+
+2. Modify Datagrid query
```
grossmannmartin: this should be heading probably. the same applies to `3. Configure list actions`

docs/administration/crud-controller/getting-started/configure-list-page.md
```diff
@@ -1 +1,117 @@
...
+ * @param \Doctrine\ORM\QueryBuilder $queryBuilder
+ */
+protected function configureQuery(QueryBuilder $queryBuilder): void
+{
+    $queryBuilder
+        ->addSelect('(CASE WHEN o.companyName IS NOT NULL
+                THEN o.companyName
+                ELSE CONCAT(o.lastName, \' \', o.firstName)
+            END) AS customerName')
+        ->andWhere('o.deleted = :deleted')
```
grossmannmartin, line 49: is it always `o`? Why `o`?

docs/administration/crud-controller/getting-started/configure-list-page.md
```diff
@@ -1 +1,117 @@
...
+    // ...
+    $datagrid
+        // Option `virtual` means that this column name will not be used in the query. Property is used to get the value from the query.
+        ->add('customerName', [
+            'label' => t('Customer Name'),
+            'virtual' => true,
+            'property' => 'customerName',
+        ])
+    ;
+
```
grossmannmartin, line 68: I see the `setPagination(bool $pagination)` method. Personally, I would rather have `enablePagination()` and `disablePagination()`

docs/administration/crud-controller/getting-started/configure-list-page.md
```diff
@@ -1 +1,117 @@
...
+    // This will add an action with the name `linkToDashboard` on the `ActionType::LIST` page
+    $actions->add(
+        ActionType::LIST,
+        Action::create('linkToDashboard', t('Link To dashboard'))
+            ->setAttribute('class', 'btn--primary', true)
+            ->setOpenInNewTab()
+            ->linkToRoute('admin_default_dashboard', fn () => [
+                'id' => 1,
+            ]),
+    );
```
grossmannmartin, line 102: linkToUrl accepts only closure. Shouldn't it also accept plain string? It seems obnoxious to be forced to wrap return string in anonymous function

docs/administration/datagrid/fields.md
```diff
@@ -0,0 +1,83 @@
...
+
+The following options are available for fields:
+
+- `label` - The label of the field. It is displayed in the header of the column.
+- `visible` - If set to `true`, the field is displayed in the datagrid. If set to `false`, the field is hidden, but fetching the data is still performed.
+- `sortable` - If set to `true`, the field can be sorted by clicking on the column header.
+- `virtual` - If set to `true`, the field is not fetched from the data source.
+- `help` - The help text that is displayed next to label in column header.
+- `template` - The template that is used to render the field. The template is a path to a Twig template file.
+- `transform` - A callback function that is used to transform the data before it is displayed in the field. The callback function receives the value of the field as the first parameter, row as the second argument and all rows as the third parameter.
```
grossmannmartin: If I use transform on a virtual field I get fatal error

code:
```
 protected function configureDatagrid(Datagrid $datagrid): Datagrid
    {
        $datagrid->add('name');

        $datagrid->add('oreo', [
            'label' => 'Oreo',
            'virtual' => true,
            'transform' => function () {
                return 'asd';
            },
        ]);

        $datagrid->setDefaultOrder('name', OrderingEnum::DESC);
        $datagrid->setPagination(false);

        return $datagrid;
    }
```

Snímek obrazovky 2025-03-05 v 10 15 30

docs/administration/datagrid/fields.md
```diff
@@ -0,0 +1,83 @@
...
+### Options
+
+The following options are available for fields:
+
+- `label` - The label of the field. It is displayed in the header of the column.
+- `visible` - If set to `true`, the field is displayed in the datagrid. If set to `false`, the field is hidden, but fetching the data is still performed.
+- `sortable` - If set to `true`, the field can be sorted by clicking on the column header.
+- `virtual` - If set to `true`, the field is not fetched from the data source.
+- `help` - The help text that is displayed next to label in column header.
+- `template` - The template that is used to render the field. The template is a path to a Twig template file.
```
grossmannmartin: it would be nice to document what variables are available in the template

docs/administration/datagrid/fields.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Configuration
+
+Fields are defined by the `Datagrid` class using the `add(string $name, array $options)`, `update(string $name, array $options)` and `remove(string $name)` methods. The `$name` parameter is the name of the field and the `$options` parameter is an array of options that define the behavior of the field.
+
+In case you are using `OrmAdapter`, you can use dot notation to access nested properties of the entity. Here are some examples of how to define field names:
+
+- `$datagrid->add('domainId')` - This will fetch the `domainId` property of the entity.
+- `$datagrid->add('currency.code')` - This will join the `currency` relation and fetch the `code` property of the `Currency` entity.
+- `$datagrid->add('status.name')` - This will join the `status` relation and look for the `name` property of the `Status` entity. In this case `Status` is a translatable entity and the `name` property is a translatable field. The datagrid will automatically fetch the correct translation based on the current locale.
```
grossmannmartin, line 13: It would be nice to automatically provide a fallback text if the translation is not entered. 
Something similar to the current product grid behavior

Snímek obrazovky 2025-03-05 v 10 26 02

now the empty field is rendered

docs/administration/datagrid/fields.md
```diff
@@ -0,0 +1,83 @@
...
+
+Fields are the building blocks of a datagrid. They are used to define the columns of the datagrid and to specify how the data should be displayed.
+
+## Configuration
+
+Fields are defined by the `Datagrid` class using the `add(string $name, array $options)`, `update(string $name, array $options)` and `remove(string $name)` methods. The `$name` parameter is the name of the field and the `$options` parameter is an array of options that define the behavior of the field.
+
+In case you are using `OrmAdapter`, you can use dot notation to access nested properties of the entity. Here are some examples of how to define field names:
+
+- `$datagrid->add('domainId')` - This will fetch the `domainId` property of the entity.
```
grossmannmartin, line 11: thinking... wouldn't be useful to provide a "syntax sugar" to quickly render text as the edit link? 
Now I have to write the custom template. 

something like

```
 $datagrid->add('name', [
            'with_detail_link' => true,
```

docs/administration/datagrid/fields.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Options
+
+The following options are available for fields:
+
+- `label` - The label of the field. It is displayed in the header of the column.
+- `visible` - If set to `true`, the field is displayed in the datagrid. If set to `false`, the field is hidden, but fetching the data is still performed.
+- `sortable` - If set to `true`, the field can be sorted by clicking on the column header.
+- `virtual` - If set to `true`, the field is not fetched from the data source.
+- `help` - The help text that is displayed next to label in column header.
```
grossmannmartin, line 24: help is not rendered if the field is not sortable or is virtual

docs/administration/datagrid/fields.md
grossmannmartin: transform behaves inconsistently for objects and scalar attributes

this code

```php
$datagrid->add('sellingFrom', [
    'label' => 'Selling from',
    'transform' => function (\DateTimeInterface $value) {
        return $value->format('Y-m-d');
    },
]);

$datagrid->add('sellingYear', [
    'label' => 'Selling Year',
    'virtual' => true,
    'template' => 'Admin/Crud/my.html.twig', // {{ row.sellingFrom|date('Y') }}
]);
```

works as expected (sellingFrom is a DateTime object) - outputs `2000-02-17 | 2000`

while this don't - outputs `name - transformed | NAME - TRANSFORMED`

```php
$datagrid->add('name', [
    'transform' => function (string $value) {
        return $value . ' - transformed';
    },
]);

$datagrid->add('oreo', [
    'label' => 'Oreo',
    'virtual' => true,
    'template' => 'Admin/Crud/my.html.twig', // {{ row.name|upper }}
]);
```

grossmannmartin (changes requested):

docs/administration/crud-controller/getting-started/creating-a-new-crud-controller.md
henzigo: yes exactly

docs/administration/crud-controller/getting-started/configure-list-page.md
henzigo: `o` like `object`

docs/administration/crud-controller/getting-started/configure-list-page.md
henzigo: I can add this option, but I wanted to have it as simple as possible :) 

```php
...
->linkToUrl(fn() => 'https://google.com')
```

It's not that bad :D

docs/administration/datagrid/fields.md
henzigo: Maybe we can discuss how this should work. Like You want to transform data that does not exists :D

henzigo: But I agree that there should be some check that will tell you something like `You cannot transform virtual field`

docs/administration/datagrid/fields.md
henzigo: Yes, but I don't know which ones are in the template :D

docs/administration/datagrid/fields.md
henzigo: You can do that with transform or custom template. We can add this functionality later

docs/administration/datagrid/fields.md
henzigo: We can add this functionality later

docs/administration/datagrid/fields.md
henzigo: I will look at it

docs/administration/crud-controller/getting-started/creating-a-new-crud-controller.md
henzigo: fixed

docs/administration/crud-controller/getting-started/configure-list-page.md
henzigo: fixed

docs/administration/crud-controller/getting-started/configure-list-page.md
henzigo: I'd like to have it as it is now. If its not problem I will not change it

docs/administration/crud-controller/getting-started/configure-list-page.md
henzigo: but you can now pass simply string

docs/administration/datagrid/fields.md
henzigo: Ok, I made it fully functional. In case you want to transforma  virtual field, in closure you will get null instead, but you can return what you want to show based on other fields, etc.

docs/administration/datagrid/fields.md
henzigo: Fixed. I misplaced it in template. Now it's working, but that code is placed twice due to weird CSS computing of arrows :D I can't wait for our new administration :D

docs/administration/datagrid/fields.md
henzigo: It works the same in both cases. The problem is that with our internal Grid, there is only one place where I can transform fields right after fetching data from DB. Because of this `row` already contains a transformed field. 

In your first case `row.sellingFrom` contains `2000-02-17` but `|date()` twig filter accepts string so it feels like you have the original value here, but it's not true. 

I updated docs a little bit so it will be clearer how it works now

docs/administration/datagrid/fields.md
henzigo: 👍 Updated docs for internal grid with that information and linked those pages :)

docs/administration/crud-controller/getting-started/configure-list-page.md
henzigo: You are right! Only one change in name of method `setIdentifier()`

docs/administration/crud-controller/getting-started/configure-list-page.md
henzigo: fixed

grossmannmartin (approved):
Awesome work ♥️. Really looking forward to use this

packages/framework/src/Resources/views/Admin/Grid/Grid.html.twig
```diff
@@ -102,25 +102,37 @@
...
+                    class="js-tooltip cursor-help in-icon in-icon--in-text in-icon--info"
+                    data-toggle="tooltip"
+                    data-placement="bottom"
+                    {% if grid.orderSourceColumnName == column.id %}
+                        style="margin-left: 20px;"
+                    {% endif %}
+                >
+                    {{ ux_icon('info') }}
+                </span>
+            {% endif %}
```
vitek-rostislav, line 123: this could be extracted into a macro to avoid the duplication :innocent:

docs/administration/internal-grid/grid-rendering-customization.md
```diff
@@ -3,9 +3,36 @@
...
+```
+<span class="my-awesome-column">
+    {{ value }}
+</span>
+```
+
+In template, you can use following variables:
+- `value` - the value of the column
+- `row` - the whole row of the grid. You can access to the row data using the column name as a key, e.g. `row['columnName']` or as a property, e.g. `row.columnName`
+- `column` - the column object itself
```
vitek-rostislav: it would be nice to add "(i.e. instance of XX class)"

docs/administration/internal-grid/grid-rendering-customization.md
```diff
@@ -3,9 +3,36 @@
...
+Then, set the template in your grid definition:
+```php
+$grid->addColumn('my_awesome_column', 'p.attribute', t('My awesome column'), true, [
+    'template' => 'Admin/Grid/my_awesome_column.html.twig',
+]);
+```
+
+## 2. Override the default template
+
+Ccreate a new template extending the original one, override appropriate blocks, and then set the template as a theme of your grid using `Grid::setTheme` method.
```
vitek-rostislav:
```suggestion
Create a new template extending the original one, override appropriate blocks, and then set the template as a theme of your grid using `Grid::setTheme` method.
```

vitek-rostislav (approved):
Hi, awesome, just a few last notes + please fix the standards (or remove the example, rebase it, and you can move the task to business validation :partying_face:

packages/framework/src/Resources/views/Admin/Grid/Grid.html.twig
henzigo: Yes, but they are not the same and after implementing the new administration it will be changed. I will not change it now :)

docs/administration/internal-grid/grid-rendering-customization.md
henzigo: fixed

docs/administration/internal-grid/grid-rendering-customization.md
henzigo: fixed

PR #3639, opened 2024-12-05
[framework] admin: sending test mail templates on click

vitek-rostislav:
When styling a mail template (as a developer, or even as an administrator), one appreciates the possibility of sending the final mail with all the variables with ease, without the necessity of actually doing the action that triggers the mail (creating an order, changing it's status, etc.). This PR adds a "Send test mail" button to the edit page of each mail template that allows the administrator to select an email address and an entity identifier to fill the variables from (if necessary).

image

#### Fixes issues
... 

----
  - https://rv-testing-mails.odin.shopsys.cloud
  - https://cz.rv-testing-mails.odin.shopsys.cloud


packages/framework/src/Controller/Admin/MailController.php
```diff
@@ -183,4 +194,56 @@ public function settingAction(Request $request): Response
             'form' => $form->createView(),
         ]);
     }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @param int $mailTemplateId
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    #[Route(path: '/mail/send/{mailTemplateId}', requirements: ['mailTemplateId' => '\d+'])]
```
grossmannmartin: Shouldn't the route only be accessible for XHR?

now I can open the route in the browser 
Snímek obrazovky 2024-12-10 v 21 14 06

packages/framework/src/Controller/Admin/MailController.php
```diff
@@ -183,4 +194,56 @@ public function settingAction(Request $request): Response
...
+        return $this->render('@ShopsysFramework/Admin/Content/Mail/mailTemplateSend.html.twig', [
+            'form' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @param \Symfony\Component\Form\FormInterface $form
+     * @return \Symfony\Component\HttpFoundation\JsonResponse
+     */
+    protected function getInvalidResponse(FormInterface $form): JsonResponse
```
grossmannmartin: maybe `createInvalidResponse` as the method always creates a new instance

packages/framework/src/Form/Admin/Mail/MailTemplateSendFormType.php
```diff
@@ -0,0 +1,80 @@
...
+                'data' => $currentAdministrator->getEmail(),
+                'constraints' => [
+                    new NotBlank(['message' => 'Please enter email address']),
+                    new Email(['message' => 'Please enter valid email']),
+                ],
+            ]);
+
+        if ($labelForEntityIdentifier !== null) {
+            $builder
+                ->add('prefillFromEntityIdentifier', IntegerType::class, [
```
grossmannmartin: I'm a little confused about the name. Shouldn't it be just `entityIdentifier`?  Or am I missing something?

packages/framework/src/Model/Mail/MailTemplateSender/MailTemplateSenderInterface.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Shopsys\FrameworkBundle\Model\Mail\MailTemplateSender;
+
+use Shopsys\FrameworkBundle\Model\Mail\MailTemplate;
+
+interface MailTemplateSenderInterface
+{
+    /**
+     * @return string|null
+     */
+    public function getFormLabelForEntityIdentifier(): ?string;
```
grossmannmartin, line 14: The null as a return value means "do not render any identifier field," but intuitively, I expected "render empty label"... 
Maybe we could talk about it, what do you think?

packages/framework/src/Resources/config/services.yaml
```diff
@@ -777,6 +777,10 @@ services:
             - '%env(MAILER_DSN)%'
             - '%env(bool:MAILER_FORCE_WHITELIST)%'
 
+    Shopsys\FrameworkBundle\Model\Mail\MailTemplateSender\MailTemplateSenderFacade:
+        arguments:
+            $mailTemplateSenders: !tagged shopsys.mail_template_sender
```
grossmannmartin: we should use `!tagged_iterator` for future compatibility
https://symfony.com/blog/new-in-symfony-7-2-deprecations#deprecated-the-tagged-tag

grossmannmartin (changes requested):

packages/framework/src/Controller/Admin/MailController.php
vitek-rostislav: I had in mind a scenario (for a developer) when you need to quickly test something multiple times, it could ease the task for you but I do not need that :slightly_smiling_face: 7baa7349eb

packages/framework/src/Controller/Admin/MailController.php
vitek-rostislav: :+1: fixed in 7baa7349eb

packages/framework/src/Form/Admin/Mail/MailTemplateSendFormType.php
vitek-rostislav: yeah, that was something from the beginning of the implementation but your suggestion is better - 7baa7349eb

packages/framework/src/Model/Mail/MailTemplateSender/MailTemplateSenderInterface.php
vitek-rostislav: yeah, let's discuss it, please

packages/framework/src/Resources/config/services.yaml
vitek-rostislav: nice :ok_hand: 7baa7349eb

grossmannmartin (approved):

PR #3413, opened 2024-09-06
Refactor stores and implement some features

henzigo:
#### Description, the reason for the PR

- Stores now includes `email`, `phone` and `directions`
- `contactInfo` was removed from stores
- On FE there is some changes in design and those properties are now visible on stores list and detail as well. 
- Filtering by city or postcode and ordering by distance was implemented
- `StoreFormType` was refactored into groups to improve UX

#### Fixes issues
... 

#### License Agreement for contributions

- [x] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://jg-stores-features.odin.shopsys.cloud
  - https://cz.jg-stores-features.odin.shopsys.cloud

PR #3641, opened 2024-12-05
admin improvements

RostislavKreisinger:

Performace improvements:
- Modified grid so that adding NULL FIRST / NULLS LAST is optional now, because it is much more time-consuming for huge collections.
- `QueryBuilderWithRowManipulatorDataSource` called callback for each item separately without the possibility of working with all items. Modified callback so that instead of calling ID = X for each item, ID IN (X, X, X) could be called only once. In Callback function it is possible to request the entire result from query as second parameter.
- Product grid: the query is modified to return only the data that will be displayed later. Instead of entity, the associated array is returned.
- new Trigram indexes were added on columns using LIKE conditions in Product, Order and Customer grids.

----
  - https://rk-ssp-2460-breno-admin-improvements.odin.shopsys.cloud
  - https://cz.rk-ssp-2460-breno-admin-improvements.odin.shopsys.cloud


packages/framework/src/Migrations/Version20241205144230.php
```diff
@@ -0,0 +1,43 @@
...
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20241205144230 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->sql('CREATE EXTENSION IF NOT EXISTS pg_trgm');
```
henzigo: I just want to add that for some databases the execute user may not have sufficient permissions for this command

vitek-rostislav: so instead of adding this into a migration, we should add it to https://github.com/shopsys/shopsys/blob/16.0/packages/framework/src/Command/CreateDatabaseCommand.php, right?

vitek-rostislav: plus add an upgrade note so that the command is run after/during deploy? :thinking:

henzigo: It should be in CreateDatabaseCommand but you will probably need to migrate existing database. The problem is that if it stays here, some operators won't get through this migration and it will crash because they won't have enough permissions. So I would probably rather advise when upgrading to manually install this extension.

packages/framework/src/Controller/Admin/ProductController.php
```diff
@@ -312,14 +321,22 @@ protected function getGrid(QueryBuilder $queryBuilder)
         $dataSource = new QueryBuilderWithRowManipulatorDataSource(
             $queryBuilder,
             'p.id',
-            function ($row) {
-                $product = $this->productFacade->getById($row['p']['id']);
-                $row['product'] = $product;
+            function ($product, $products) {
```
vitek-rostislav: it looks like a huge part of the product list grid creation is duplicated - here and in `ProductPickerController`. Could you please extract it into a new `ProductGridFactory`? The data source definition should be there definitely, as long as the grid creation with majority of the columns definitions :pray:

packages/framework/src/Resources/views/Admin/Content/ProductPicker/listGrid.html.twig
```diff
@@ -5,22 +5,22 @@
 {% endblock %}
 
 {% block grid_value_cell_id_name %}
-    {% if row.p.variantType == VARIANT_TYPE_MAIN %}
+    {% if row.variantType == VARIANT_TYPE_MAIN %}
```
vitek-rostislav, line 6: it would be awesome to get rid of DRY violations even here in these two templates (product list and product picker), maybe we could create a common ancestor for them?

packages/framework/src/Controller/Admin/ProductController.php
```diff
@@ -312,14 +321,22 @@ protected function getGrid(QueryBuilder $queryBuilder)
         $dataSource = new QueryBuilderWithRowManipulatorDataSource(
             $queryBuilder,
             'p.id',
-            function ($row) {
```
vitek-rostislav, line 315: could you please update the docs regarding this data source? :pray: https://github.com/shopsys/shopsys/blob/16.0/docs/administration/grid-data-sources.md#querybuilderwithrowmanipulatordatasource

packages/framework/src/Twig/ProductExtension.php
```diff
@@ -95,6 +96,19 @@ public function getProductListDisplayName(Product $product)
         return $product->getName();
     }
 
+    /**
+     * @param string|null $name
+     * @return string
+     */
+    public function getProductListDisplayNameByName(?string $name): string
```
vitek-rostislav, line 89: by adding this function, the original one (`getProductListDisplayName`) is not used anywhere anymore, could you please remove it and add the information to upgrade notes? :pray:

packages/framework/src/Model/Product/ProductVisibilityFacade.php
```diff
@@ -95,4 +101,76 @@ public function createAndRefreshProductVisibilitiesForPricingGroup(PricingGroup
...
+            foreach ($productVisibilities as $productId => $productVisibility) {
+                if ($productVisibilitiesIndexedByProductId[$productId]) {
+                    $productVisibilitiesIndexedByProductId[$productId] = $productVisibility;
+                }
+            }
+        }
+
+        return $productVisibilitiesIndexedByProductId;
+    }
+
```
vitek-rostislav: What about simplifying it like this? (Or something like this, it is just an idea and might be further refactored.

```
protected function getProductsVisibilityIndexedByProductId(array $productIds, bool $defaultVisibility): array
{
    $productVisibilitiesIndexedByProductId = array_fill_keys($productIds, $defaultVisibility);

        $productVisibilities = $this->getProductsVisibilitiesByDomainIndexedByProductId($productIds, $domainId);

        foreach ($productVisibilities as $productId => $productVisibility) {
            if ($defaultVisibility) {
                $productVisibilitiesIndexedByProductId[$productId] = $productVisibilitiesIndexedByProductId[$productId] && $productVisibility;
            } else {
                $productVisibilitiesIndexedByProductId[$productId] = $productVisibilitiesIndexedByProductId[$productId] || $productVisibility;
            }
        }
    }

    return $productVisibilitiesIndexedByProductId;

public function areProductsVisibleForDefaultPricingGroupOnSomeDomainIndexedByProductId(array $productIds): array
    {
    return $this->getProductsVisibilityIndexedByProductId($productIds, false);
}

public function areProductsVisibleForDefaultPricingGroupOnEachDomainIndexedByProductId(array $productIds): array
    {
    return $this->getProductsVisibilityIndexedByProductId($productIds, true);
}
```

vitek-rostislav: btw, it looks like there could be simple tests for the two new public methods :slightly_smiling_face:

packages/framework/src/Migrations/Version20241205144230.php
```diff
@@ -0,0 +1,43 @@
...
+
+        //product indexes
+        $this->sql('CREATE INDEX products_catnum_trgm_idx ON products USING gin (normalized(catnum) gin_trgm_ops);');
```
vitek-rostislav, line 18: this should also be mentioned in the upgrade notes - developers should create such indexes for all the fields they are normalizing and using in their custom admin search queries, right?

upgrade-notes/backend_20241205_151830.md
```diff
@@ -0,0 +1,7 @@
+#### Admin Grid performance improvements ([#2460](https://github.com/shopsys/shopsys/pull/{pullRequestId}))
+
+- Modified grid so that adding NULL FIRST / NULLS LAST is optional now, because it is much more time-consuming.
+- QueryBuilderWithRowManipulatorDataSource called callback for each item separately without the possibility of working with all items. Modified callback so that instead of calling ID = X for each item, ID IN (X, X, X) could be called only once. See example implementation in ProductControllet. In Callback function it is possible to request the entire result from query as second parameter.
+- Grid pulls out entire product (entity): Query is modified to return only the data that will be displayed later. Instead of entity, associated field is returned.
+- Search using LIKE cannot use B-tree index over columns with NORMALIZED function and LIKE condition %text%. Instead of standard B-tree index, new Trigram indexes were added on columns using LIKE conditions in Product, Order and Customer grids.
```
vitek-rostislav: I would rather see this as a description of the pull request on GitHub but I believe I can live with them here as well. Let's talk about it, please

upgrade-notes/backend_20241205_151830.md
```diff
@@ -0,0 +1,7 @@
+#### Admin Grid performance improvements ([#2460](https://github.com/shopsys/shopsys/pull/{pullRequestId}))
+
+- Modified grid so that adding NULL FIRST / NULLS LAST is optional now, because it is much more time-consuming.
+- QueryBuilderWithRowManipulatorDataSource called callback for each item separately without the possibility of working with all items. Modified callback so that instead of calling ID = X for each item, ID IN (X, X, X) could be called only once. See example implementation in ProductControllet. In Callback function it is possible to request the entire result from query as second parameter.
```
vitek-rostislav: ProductControllet -> ProductController

However, if you refactor the code, the example usage will be probably in the new ProductGridFactory :slightly_smiling_face:

packages/framework/src/Controller/Admin/OrderController.php
```diff
@@ -214,24 +242,24 @@ function ($row) {
...
-            'isAdvancedSearchFormSubmitted' => $this->advancedSearchOrderFacade->isAdvancedSearchOrderFormSubmitted(
-                $request,
-            ),
-        ]);
+        return $grid;
     }
 
     /**
      * @param array $row
+     * @param array $ids
```
vitek-rostislav:
```suggestion
     * @param int[] $ids
```

vitek-rostislav (changes requested):

docs/administration/grid-data-sources.md
```diff
@@ -40,14 +42,29 @@ $queryBuilder = $transportRepository->getQueryBuilderForAll()
 $dataSource = new QueryBuilderWithRowManipulatorDataSource(
     $queryBuilder,
     't.id',
-    function ($row) {
-        $transport = $transportRepository->findById($row['t']['id']);
+    function ($row, $results) {
+        if($this->transportsIndexedByLocalCache === null) {
```
vitek-rostislav:
```suggestion
        if ($this->transportsIndexedByIdLocalCache === null) {
```

docs/administration/grid-data-sources.md
```diff
@@ -40,14 +42,29 @@ $queryBuilder = $transportRepository->getQueryBuilderForAll()
...
+    SortableNullsWalker::class
 );
 ```
 
+As in the case of QueryBuilderDataSource, it is possible to change the setting of the $hint parameter, see QueryBuilderDataSource.
+
+### Tips for optimal usage QueryBuilderDataSource or QueryBuilderWithRowManipulatorDataSource
+
+- The query should return an array of scalar values instead of an entity if the table contains large collections of records.
+- Grid with callback calls the callback for each item separately. To avoid calling SQL queries for each item, use the second parameter of the callback function in which the entire result for the bulk query is stored in the local cache. See. OrderController.
```
vitek-rostislav:
```suggestion
- Grid with callback calls the callback for each item separately. To avoid calling SQL queries for each item, use the second parameter of the callback function in which the entire result for the bulk query is stored in the local cache. See e.g. `Shopsys\FrameworkBundle\Controller\Admin\OrderController::getOrdersGrid()`.
```

upgrade-notes/backend_20241205_151830.md
```diff
@@ -2,7 +2,8 @@
...
+- Twig filter getProductListDisplayName from ProductExtension is removed. Use getProductListDisplayNameByName instead.
 - Grid pulls out entire product (entity): Query is modified to return only the data that will be displayed later. Instead of entity, associated field is returned.
-- add to your postgreSQL extension pg_trgm: `CREATE EXTENSION IF NOT EXISTS pg_trgm WITH SCHEMA pg_catalog`. Search using LIKE cannot use B-tree index over columns with NORMALIZED function and LIKE condition %text%. Instead of standard B-tree index, new Trigram indexes were added on columns using LIKE conditions in Product, Order and Customer grids.
+- add to your postgreSQL extension pg_trgm: `CREATE EXTENSION IF NOT EXISTS pg_trgm WITH SCHEMA pg_catalog` manually for your production and devel environment (local and CI is covered by db-create target). Search using LIKE cannot use B-tree index over columns with NORMALIZED function and LIKE condition %text%. Instead of standard B-tree index, new Trigram indexes were added on columns using LIKE conditions in Product, Order and Customer grids.
+  - you need to install the extension before this migration Version20241205144230.
```
vitek-rostislav:
```suggestion
  - you need to install the extension before running the migration `Version20241205144230`. 
```

vitek-rostislav (changes requested):

packages/framework/src/Migrations/Version20241205144230.php
vitek-rostislav: I would add this to the upgrade notes (under the mention of the migration):

```
- check your code for usages of the `NORMALIZED(columnName) LIKE NORMALIZED(:text)` and create the indexes for corresponding columns the same way they are created in the migration `Version20241205144230`

vitek-rostislav:
- [x] check https://github.com/shopsys/shopsys/pull/3413#discussion_r1894167217 before merging

vitek-rostislav (approved):

PR #3648, opened 2024-12-11
Design system

chlebektomas:
New design system with CSS variables.

#### Fixes issues
... 

----
  - https://tc-ssp-2988-design-system.odin.shopsys.cloud
  - https://cz.tc-ssp-2988-design-system.odin.shopsys.cloud


project-base/storefront/components/Basic/Flag/Flag.tsx
JanMolcik, line 1: If we are using just `isLight()` from `tinycolor` package, do we actually need it? There is already simillar functionality to get contrast text based on bg color (`getYIQContrastTextColor()` function in `bannersUtils`). Could we somehow re-use it instead?

JanMolcik (changes requested):
Good job, I have got just one idea/proposal.

project-base/storefront/components/Basic/Flag/Flag.tsx
chlebektomas: fixed

project-base/storefront/components/Basic/Flag/Flag.tsx
```diff
@@ -1,5 +1,5 @@
+import { getYIQContrastTextColor } from 'components/Blocks/Banners/bannersUtils';
```
JanMolcik: Since it is now used by other components than just banner, I would move it somewhere to general utils.

JanMolcik (changes requested):

JanMolcik (approved):

PR #3658, opened 2024-12-12
[framework] improved rendering information about eshop required settings

grossmannmartin:

The information about the necessity of some settings is now separated from the default flash messages. 
The info about required settings is now rendered on every page on the storefront.
Missing warehouse and country is now reported.

+ Minor fixes and improvements 

#### Fixes issues
... 

----
  - https://mg-new-build.odin.shopsys.cloud
  - https://cz.mg-new-build.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Twig/RequiredSettingExtension.php
+++ b/packages/framework/src/Twig/RequiredSettingExtension.php
@@ -1,0 +1,224 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Twig;
+
+use Override;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\Setting\Setting;
+use Shopsys\FrameworkBundle\Model\Country\CountryFacade;
+use Shopsys\FrameworkBundle\Model\Mail\MailTemplateFacade;
+use Shopsys\FrameworkBundle\Model\Product\Parameter\ParameterFacade;
+use Shopsys\FrameworkBundle\Model\Product\Unit\Exception\UnitNotFoundException;
+use Shopsys\FrameworkBundle\Model\Product\Unit\UnitFacade;
+use Shopsys\FrameworkBundle\Model\Stock\StockFacade;
+use Symfony\Component\Routing\RouterInterface;
+use Twig\Environment;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+class RequiredSettingExtension extends AbstractExtension
+{
+    /**
+     * @var string[]
+     */
+    protected array $requiredSettingsMessages = [];
+
+    /**
+     * @param \Twig\Environment $twig
+     * @param \Symfony\Component\Routing\RouterInterface $router
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @param \Shopsys\FrameworkBundle\Component\Setting\Setting $setting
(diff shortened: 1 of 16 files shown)


packages/framework/src/Twig/RequiredSettingExtension.php
```diff
@@ -0,0 +1,235 @@
...
+            [
+                'requiredSettingsMessages' => $requiredSettingsMessages,
+            ],
+        );
+    }
+
+    /**
+     * @param string[] $warnings
+     */
+    protected function checkEnabledMailTemplatesHaveTheirBodyAndSubjectFilled(array &$warnings): void
```
vitek-rostislav: instead of using the references, I would rather make `requiredSettingsMessages` a class property, however, it is a matter of taste so I can live with this solution as well :slightly_smiling_face:

packages/framework/src/Twig/RequiredSettingExtension.php
```diff
@@ -0,0 +1,235 @@
...
+                    '%url%' => $this->router->generate('admin_unit_list'),
+                ],
+            );
+        }
+    }
+
+    /**
+     * @param string[] $warnings
+     */
+    protected function checkAtLeastOneWarehouseExists(array &$warnings): void
```
vitek-rostislav: in the code itself, I would rather use "stock" instead of warehouse, but I am not sure whether this rule is strictly followed even now, so :shrug:

vitek-rostislav (changes requested):
Hi, nice job :ok_hand: The failing build is a punishment for working late, I suppose :smile: Please, add upgrade notes for the changes, there are several protected methods removed from the controller, so it should be mentioned there :pray:

packages/framework/src/Twig/RequiredSettingExtension.php
grossmannmartin: What was I thinking? Of course the class property is a better solution 😇

fixed in c3bbb989a3

packages/framework/src/Twig/RequiredSettingExtension.php
grossmannmartin: fixed in c3bbb989a3

vitek-rostislav:
Reviewed and tested, works like a charm :ok_hand: However, I have some notes from the testing that might be interesting for @pk16011990 :wink: 

1. "Some required email templates are not fully set." message - the link leads to the list of mail templates (`admin/mail/template/`) but as an admin, I migh have no idea what to do next - which ones are the required? And what does it mean they are not set? Do I need to fill-in all of them? Do I need to open each one of them (and open their GrapesJS editor) to actually see which ones are not filled?
2. Messages regarding missing article settings (terms and conditions, privacy policy, etc.) - the links lead to the proper settings page, however, when there are no articles yet in my database, I have nothing to choose from. As an administrator, I might be lost about where to go next -do I need to create a new article or blog article? Where do I do that? I would add a link to the proper article creation page to these settings pages when no articles exist 
 
image

And one more thing regarding the articles setting - we have messages for each domain ("Privacy policy article for domain shopsys is not set.", "Privacy policy article for domain 2.shopsys is not set.", however, the link does not take the domain into account (it is not possible ATM, as the selected domain tab can not be passed via URL but that could be implemented) so an admin might set the article for the first domain and then uses the link from the message and is forwarded to the settings page with the article already selected. So he needs to realize he needs to switch the domain using the domain tab.

vitek-rostislav (approved):

grossmannmartin:
> Reviewed and tested, works like a charm 👌 However, I have some notes from the testing that might be interesting ...

Those are fair points, we could improve it in the future.

PR #3669, opened 2024-12-16
added deduplication for product messages in the queue, including scopes

janjavorek:

- export scopes are moved from message to Redis to be able to keep the product in queue only once, while being able to add new scope
- if more than 2000 ids are dispatched at once (usually through affected changes), only the message to dispatch messages is sent to improve performance
- If the product ID is already present in Redis, the message is not dispatched again
- If the product ID is not in Redis, a handler for the message exports all scopes

#### Fixes issues
... 

Fixing the problem when multiple consumers are processing product recalculation messages with the same product ID. Messages in the queue are filtered out before another message with the same product ID is dispatched.

----
  - https://jj-ssp-rabbitmq-deduplication2.odin.shopsys.cloud
  - https://cz.jj-ssp-rabbitmq-deduplication2.odin.shopsys.cloud

### DIFF
--- a/docs/asynchronous-processing/product-recalculations.md
+++ b/docs/asynchronous-processing/product-recalculations.md
@@ -38,7 +38,19 @@ When you need to recalculate visibility of all products, you should use the `Sho
 This method dispatches the special message "dispatch all" to the message broker, and this message is then handled by the async handler, which takes care of dispatching all product IDs to recalculation.
 That way, we may dispatch all products to recalculation during request without worrying about the size of the catalog – it's not necessary to load all products (nor their IDs) from the database, and the user interface is not blocked by this operation.
 
-## Dispatch recalculation message when indirect change is made
+## Product Deduplication and Batch Dispatching
+
+To optimize performance and avoid redundant processing, the recalculation dispatch mechanism incorporates product deduplication and batch dispatching strategies.
+Export scopes ([mentioned later in this article](#recalculation-scope)) are stored in Redis, allowing products to remain in the queue only once while still supporting the addition of new scopes dynamically.
+This ensures that if a product ID is already present in Redis, it will not be dispatched again, preventing unnecessary recalculations and improving efficiency.
+Additionally, if a product ID is not yet present in Redis, Shopsys Platform ensures that all scopes are exported, guaranteeing that all relevant data is properly processed.
+
+When a large number of product IDs, typically more than 2000, are dispatched at once due to i.e., [affected changes](#dispatch-recalculation-message-when-an-indirect-change-is-made), Shopsys Platform takes an optimized approach.
+Instead of dispatching all product IDs individually in a currently running request, a few messages are sent to trigger the dispatch, significantly enhancing performance in large-scale recalculations.
+
+These refinements effectively minimize redundant messages, making recalculations more efficient and well-suited for handling large product catalogs with frequent changes.
+
+## Dispatch recalculation message when an indirect change is made
 
 Sometimes it is necessary to trigger recalculation for product when some indirect change is made.
 For example, when a category is deleted, when the parameter translation of a product is changed, etc.
--- a/docs/asynchronous-processing/product-recalculations.md
+++ b/docs/asynchronous-processing/product-recalculations.md
@@ -52,7 +52,7 @@ These refinements effectively minimize redundant messages, making recalculations
 
 ## Dispatch recalculation message when an indirect change is made
 
-Sometimes it is necessary to trigger recalculation for product when some indirect change is made.
+Sometimes it is necessary to trigger recalculation for a product when some indirect change is made.
 For example, when a category is deleted, when the parameter translation of a product is changed, etc.
 
 For this situation, the `\Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchAffectedProductsSubscriber` subscriber is used.
@@ -88,7 +88,7 @@ You may disable this cron module if it doesn't suit your needs.
 
 ## Handle recalculations in tests
 
-In functional tests, you sometime need to trigger recalculations to be able to test some use-case.
+In functional tests, you sometimes need to trigger recalculations to be able to test some use-case.
 For example, when you need to test the change of a product visibility, you need to recalculate it after the change is made, so the data returned from GraphQL are accurate.
 
 You should use the `Tests\App\Test\WebTestCase::handleDispatchedRecalculationMessages()` method to process all dispatched recalculation messages.
@@ -129,10 +129,10 @@ That way we are sure the code works – the message is truly dispatched, thus th
 
 ## Recalculation scope
 
-In many occasions, to optimize the application, you might need to restrict the scope of a product recalculation and elastic export.
+On many occasions, to optimize the application, you might need to restrict the scope of a product recalculation and elastic export.
 Sometimes you do not need to recalculate visibility or selling denied, or you need to export just particular fields into Elasticsearch.
-For example, when launching a new domain, you need to export just the product URLs into Elasticsearch which takes significantly less time than the full recalculation and export.
-On the other hand, it might be hard to keep in mind all the dependencies (e.g. you need to know you need to recalculate visibility after you change `Product::$sellingFrom`,
+For example, when launching a new domain, you need to export just the product URLs into Elasticsearch, which takes significantly less time than the full recalculation and export.
+On the other hand, it might be hard to keep in mind all the dependencies (e.g., you need to know you need to recalculate visibility after you change `Product::$sellingFrom`,
 or you need to keep in mind that with a change of product URL, you always need to export `hreflang_links` into Elasticsearch as well, etc.).
 
 For this purpose, recalculation scopes are defined as an associative array in the `Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeConfigFacade` class.
@@ -141,7 +141,7 @@ Each scope rule defines which fields should be exported to Elasticsearch togethe
 
 You can use `./bin/console shopsys:list:export-scopes` command to list and examine all the available scopes.
 
-The scope usage can be seen in action e.g. in `Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchAffectedProductsSubscriber` class:
+The scope usage can be seen in action e.g., in `Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchAffectedProductsSubscriber` class:
 
 ```php
 public function dispatchAffectedByBrand(BrandEvent $brandEvent): void
@@ -158,7 +158,7 @@ public function dispatchAffectedByBrand(BrandEvent $brandEvent): void
 
 Thanks to the usage of `SCOPE_BRAND` here, no visibility or selling denied recalculations are done after a brand is updated, only the brand-related fields (brand ID, name, and URL) are exported to Elasticsearch for the affected products.
 
-There are several methods that allow you to modify the scopes configuration further to suit your project needs.
+There are several methods that allow you to modify the scope configuration further to suit your project needs.
 
 - `addExportFieldsToExistingScopeRule()`
 - `addNewExportScopeRule()`
--- a/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
+++ b/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
@@ -145,9 +145,10 @@ class DispatchRecalculationMessageCommand extends Command
             return Command::FAILURE;
         }
 
-        $dispatchedProductIds = $this->productRecalculationDispatcher->dispatchProductIds($productIds, $priority, $scopes);
+        $this->productRecalculationDispatcher->dispatchProductIds($productIds, $priority, $scopes);
+
         $symfonyStyle->success([
-            'Dispatched message for IDs', implode(', ', $dispatchedProductIds),
+            'Dispatched message for IDs', implode(', ', $productIds),
             sprintf('Priority: %s', $priority),
             sprintf('Scopes: %s', count($scopes) > 0 ? implode(', ', $scopes) : '-'),
         ]);
--- a/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
+++ b/packages/framework/src/Command/DispatchRecalculationMessageCommand.php
@@ -137,6 +137,8 @@ class DispatchRecalculationMessageCommand extends Command
         try {
             Assert::allNumeric($productIds, 'All product IDs must be numeric');
             Assert::notEmpty($productIds, 'You must specify at least one product ID');
+
+            $productIds = array_map('intval', $productIds);
         } catch (InvalidArgumentException $e) {
             $symfonyStyle->error($e->getMessage());
 
--- a/packages/framework/src/Model/Product/ProductFacade.php
+++ b/packages/framework/src/Model/Product/ProductFacade.php
@@ -269,9 +269,9 @@ class ProductFacade
     /**
      * @return iterable<array{id: int}>
      */
-    public function iterateAllProductIds(): iterable
+    public function iterateAllProductIdsExceptVariant(): iterable
     {
-        return $this->productRepository->iterateAllProductIds();
+        return $this->productRepository->iterateAllProductIdsExceptVariant();
     }
 
     /**
--- a/packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
+++ b/packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Nette\Utils\Json;
+use Psr\Log\LoggerInterface;
+use Shopsys\FrameworkBundle\Model\Product\ProductRepository;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+class DispatchProductIdsBatchMessageHandler implements MessageHandlerInterface
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\ProductRepository $productRepository
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDispatcherExecutor $productRecalculationDispatcherExecutor
+     * @param \Psr\Log\LoggerInterface $logger
+     */
+    public function __construct(
+        protected readonly ProductRepository $productRepository,
+        protected readonly ProductRecalculationDispatcherExecutor $productRecalculationDispatcherExecutor,
+        protected readonly LoggerInterface $logger,
+    ) {
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchProductIdsBatchMessage $message
+     */
+    public function __invoke(DispatchProductIdsBatchMessage $message): void
+    {
+        $dispatchedProductIds = $this->productRecalculationDispatcherExecutor->dispatchProductIds(
+            $message->productIds,
+            $message->productRecalculationPriorityEnum,
+            $message->exportScopes,
+        );
+
+        if ($dispatchedProductIds === []) {
+            $this->logger->info('No new products were dispatched as they are already in the queue.', [
+                'product_ids' => Json::encode($message->productIds),
+                'export_scopes' => Json::encode($message->exportScopes),
+            ]);
+
+            return;
+        }
+
+        $this->logger->info('Dispatched batch of products for recalculation.', [
+            'product_ids' => Json::encode($dispatchedProductIds),
+            'export_scopes' => Json::encode($message->exportScopes),
+        ]);
+    }
+}
--- a/packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
+++ b/packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
@@ -1,0 +1,170 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Nette\Utils\Json;
+use Redis;
+
+class ProductRecalculationDeduplicationFacade
+{
+    protected const int TTL = 10800;
+
+    /**
+     * @param \Redis $redisClient
+     * @param bool $isDeduplicationActive
+     */
+    public function __construct(
+        protected readonly Redis $redisClient,
+        protected readonly bool $isDeduplicationActive,
+    ) {
+    }
+
+    /**
+     * @param int[] $productIds
+     * @param string $priority
+     */
+    public function delete(
+        array $productIds,
+        string $priority,
+    ): void {
+        if (!$this->isDeduplicationActive) {
+            return;
+        }
+
+        $this->redisClient->hDel($this->getCacheKey($priority), ...array_map('strval', $productIds));
+    }
+
+    /**
+     * @param int[] $productIds
+     * @param string $priority
+     * @return array<int,string[]>
+     */
+    public function getScopesIndexedByProductId(
+        array $productIds,
+        string $priority,
+    ): array {
+        if (!$this->isDeduplicationActive) {
+            return array_fill_keys($productIds, []);
+        }
+
+        $exportScopesIndexedByProductIds = $this->doGetStoredExportScopes($productIds, $priority);
+
+        foreach ($exportScopesIndexedByProductIds as $productId => $exportScopesJson) {
+            if ($exportScopesJson === false) {
+                $exportScopesJson = '[]';
+            }
+
+            $exportScopes = Json::decode($exportScopesJson, true);
+            $exportScopesIndexedByProductIds[$productId] = $exportScopes;
+        }
+
+        return $exportScopesIndexedByProductIds;
+    }
+
+    /**
+     * @param int[] $productIdsWithMainVariants
+     * @param string[] $requestedExportScopes
+     * @param string $priority
+     * @return int[]
+     */
+    public function updateScopesAndReturnProductIdsToDispatch(
+        array $productIdsWithMainVariants,
+        array $requestedExportScopes,
+        string $priority,
+    ): array {
+        if (!$this->isDeduplicationActive) {
+            return $productIdsWithMainVariants;
+        }
+
+        $exportScopesIndexedByProductIds = $this->doGetStoredExportScopes($productIdsWithMainVariants, $priority);
+        $productIdsToDispatch = [];
+
+        foreach ($exportScopesIndexedByProductIds as $productId => $storedScopesJson) {
+            if ($storedScopesJson === false) {
+                $productIdsToDispatch[] = $productId;
+            }
+
+            $updatedScopesJson = $this->updateScopesByRequestedScopes($requestedExportScopes, $storedScopesJson);
+            $exportScopesIndexedByProductIds[$productId] = $updatedScopesJson;
+        }
+
+        $this->redisClient->hMset($this->getCacheKey($priority), $exportScopesIndexedByProductIds);
+        $this->redisClient->rawCommand(
+            'HEXPIRE',
+            $this->redisClient->getOption(Redis::OPT_PREFIX) . $this->getCacheKey($priority),
+            self::TTL,
+            'FIELDS',
+            count($exportScopesIndexedByProductIds),
+            ...array_keys($exportScopesIndexedByProductIds),
+        );
+
+        return $productIdsToDispatch;
+    }
+
+    /**
+     * @param string[] $requestedExportScopes
+     * @param false|string $storedScopesJson
+     * @return string
+     */
+    protected function updateScopesByRequestedScopes(
+        array $requestedExportScopes,
+        false|string $storedScopesJson,
+    ): string {
+        // no scopes stored; store requested scopes
+        if ($storedScopesJson === false) {
+            return Json::encode($requestedExportScopes);
+        }
+
+        $storedScopes = Json::decode($storedScopesJson, true);
+
+        // all scopes are already stored; do not modify stored scopes
+        if ($storedScopes === []) {
+            return $storedScopesJson;
+        }
+
+        // requested all scopes; store all scopes
+        if ($requestedExportScopes === []) {
+            return '[]';
+        }
+
+        // all requested scopes already stored; do not modify stored scopes
+        if (array_diff($requestedExportScopes, $storedScopes) === []) {
+            return $storedScopesJson;
+        }
+
+        // update stored scopes with newly requested scopes
+        $newScopes = array_unique(array_merge($storedScopes, $requestedExportScopes));
+
+        return Json::encode($newScopes);
+    }
+
+    /**
+     * @param int[] $productIds
+     * @param string $priority
+     * @return array<int, string|false>
+     */
+    protected function doGetStoredExportScopes(array $productIds, string $priority): array
+    {
+        $exportScopesIndexedByProductIds = $this->redisClient->hMget(
+            $this->getCacheKey($priority),
+            array_map('strval', $productIds),
+        );
+
+        if ($exportScopesIndexedByProductIds === false) {
+            $exportScopesIndexedByProductIds = array_fill_keys($productIds, false);
+        }
+
+        return $exportScopesIndexedByProductIds;
+    }
+
+    /**
+     * @param string $priority
+     * @return string
+     */
+    protected function getCacheKey(string $priority): string
+    {
+        return 'product_recalculation_ids_' . $priority;
+    }
+}
--- a/packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
+++ b/packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
@@ -11,11 +11,15 @@ use Shopsys\FrameworkBundle\Model\Product\Product;
 
 class ProductRecalculationDispatcher extends AbstractMessageDispatcher
 {
+    protected const int DISPATCH_BATCH_SIZE = 2000;
+
     /**
      * @param \Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeConfig $productExportScopeConfig
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDispatcherExecutor $productRecalculationDispatcherExecutor
      */
     public function __construct(
         protected readonly ProductExportScopeConfig $productExportScopeConfig,
+        protected readonly ProductRecalculationDispatcherExecutor $productRecalculationDispatcherExecutor,
     ) {
     }
 
@@ -23,14 +27,13 @@ class ProductRecalculationDispatcher extends AbstractMessageDispatcher
      * @param \Shopsys\FrameworkBundle\Model\Product\Product[] $products
      * @param string $productRecalculationPriorityEnum
      * @param string[] $exportScopes
-     * @return int[]
      */
     public function dispatchProducts(
         array $products,
         string $productRecalculationPriorityEnum = ProductRecalculationPriorityEnum::REGULAR,
         array $exportScopes = [],
-    ): array {
-        return $this->dispatchProductIds(
+    ): void {
+        $this->dispatchProductIds(
             array_map(static fn (Product $product) => $product->getId(), $products),
             $productRecalculationPriorityEnum,
             $exportScopes,
@@ -41,26 +44,38 @@ class ProductRecalculationDispatcher extends AbstractMessageDispatcher
      * @param int[] $productIds
      * @param string $productRecalculationPriorityEnum
      * @param string[] $exportScopes
-     * @return int[]
      */
     public function dispatchProductIds(
         array $productIds,
         string $productRecalculationPriorityEnum = ProductRecalculationPriorityEnum::REGULAR,
         array $exportScopes = [],
-    ): array {
+    ): void {
         $this->verifyExportScopes($exportScopes);
         $productIds = array_unique($productIds);
 
-        foreach ($productIds as $productId) {
-            $message = match ($productRecalculationPriorityEnum) {
-                ProductRecalculationPriorityEnum::HIGH => new ProductRecalculationPriorityHighMessage((int)$productId, $exportScopes),
-                ProductRecalculationPriorityEnum::REGULAR => new ProductRecalculationPriorityRegularMessage((int)$productId, $exportScopes),
-                default => throw new UnknownProductRecalculationPriorityException($productRecalculationPriorityEnum),
-            };
-            $this->messageBus->dispatch($message);
+        $productIdsCount = count($productIds);
+
+        if ($productIdsCount <= static::DISPATCH_BATCH_SIZE) {
+            $this->productRecalculationDispatcherExecutor->dispatchProductIds(
+                $productIds,
+                $productRecalculationPriorityEnum,
+                $exportScopes,
+            );
+
+            return;
         }
 
-        return $productIds;
+        for ($i = 0; $i < $productIdsCount; $i += static::DISPATCH_BATCH_SIZE) {
+            $productIdsBatch = array_slice($productIds, $i, static::DISPATCH_BATCH_SIZE);
+
+            $this->messageBus->dispatch(
+                new DispatchProductIdsBatchMessage(
+                    $productIdsBatch,
+                    $exportScopes,
+                    $productRecalculationPriorityEnum,
+                ),
+            );
+        }
     }
 
     /**
--- a/packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcherExecutor.php
+++ b/packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcherExecutor.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Symfony\Component\Messenger\MessageBusInterface;
+
+/**
+ * @internal This class is not intended to be used directly from your code. Use ProductRecalculationDispatcher
+ */
+class ProductRecalculationDispatcherExecutor
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationRepository $productRecalculationRepository
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDeduplicationFacade $productRecalculationDeduplicationFacade
+     * @param \Symfony\Component\Messenger\MessageBusInterface $messageBus
+     */
+    public function __construct(
+        protected readonly ProductRecalculationRepository $productRecalculationRepository,
+        protected readonly ProductRecalculationDeduplicationFacade $productRecalculationDeduplicationFacade,
+        protected readonly MessageBusInterface $messageBus,
+    ) {
+    }
+
+    /**
+     * @param int[] $productIds
+     * @param string $productRecalculationPriorityEnum
+     * @param string[] $exportScopes
+     * @return int[]
+     */
+    public function dispatchProductIds(
+        array $productIds,
+        string $productRecalculationPriorityEnum = ProductRecalculationPriorityEnum::REGULAR,
+        array $exportScopes = [],
+    ): array {
+        $productIdsWithMainVariants = $this->productRecalculationRepository->replaceVariantIdsWithMainVariantIds($productIds);
+
+        $productIdsToDispatch = $this->productRecalculationDeduplicationFacade->updateScopesAndReturnProductIdsToDispatch(
+            $productIdsWithMainVariants,
+            $exportScopes,
+            $productRecalculationPriorityEnum,
+        );
+
+        if ($productIdsToDispatch === []) {
+            return [];
+        }
+
+        foreach ($productIdsToDispatch as $productId) {
+            $message = match ($productRecalculationPriorityEnum) {
+                ProductRecalculationPriorityEnum::HIGH => new ProductRecalculationPriorityHighMessage($productId),
+                ProductRecalculationPriorityEnum::REGULAR => new ProductRecalculationPriorityRegularMessage($productId),
+                default => throw new UnknownProductRecalculationPriorityException($productRecalculationPriorityEnum),
+            };
+
+            $this->messageBus->dispatch($message);
+        }
+
+        return $productIdsToDispatch;
+    }
+}
--- a/packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
+++ b/packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
 
+use Nette\Utils\Json;
+use Psr\Log\LoggerInterface;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexDefinitionLoader;
 use Shopsys\FrameworkBundle\Component\Elasticsearch\IndexFacade;
@@ -29,6 +31,7 @@ class ProductRecalculationFacade
      * @param \Shopsys\FrameworkBundle\Model\Product\ProductSellingDeniedRecalculator $productSellingDeniedRecalculator
      * @param \Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeConfigFacade $productExportScopeConfigFacade
      * @param \Shopsys\FrameworkBundle\Model\Product\ProductElasticsearchProvider $productElasticsearchProvider
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDeduplicationFacade $productRecalculationDeduplicationFacade
      */
     public function __construct(
         protected readonly IndexFacade $indexFacade,
@@ -40,17 +43,33 @@ class ProductRecalculationFacade
         protected readonly ProductSellingDeniedRecalculator $productSellingDeniedRecalculator,
         protected readonly ProductExportScopeConfigFacade $productExportScopeConfigFacade,
         protected readonly ProductElasticsearchProvider $productElasticsearchProvider,
+        protected readonly ProductRecalculationDeduplicationFacade $productRecalculationDeduplicationFacade,
     ) {
     }
 
     /**
-     * @param array<int, string[]> $exportScopesIndexedByProductId
+     * @param int[] $productIds
+     * @param string $priority
+     * @param \Psr\Log\LoggerInterface|null $logger
      */
-    public function recalculate(array $exportScopesIndexedByProductId): void
+    public function recalculate(array $productIds, string $priority, ?LoggerInterface $logger = null): void
     {
+        $exportScopesIndexedByProductId = $this->productRecalculationDeduplicationFacade->getScopesIndexedByProductId($productIds, $priority);
+
         foreach ($this->groupProductIdsWithSameScopes($exportScopesIndexedByProductId) as $productIdsWithScopes) {
-            $idsToRecalculate = $this->productRecalculationRepository->getIdsToRecalculate($productIdsWithScopes[self::KEY_PRODUCT_IDS]);
+            $idsToRecalculate = $this->productRecalculationRepository->getIdsToRecalculateByMainVariantIds(
+                $productIdsWithScopes[self::KEY_PRODUCT_IDS],
+            );
+
             $this->recalculateWithScope($idsToRecalculate, $productIdsWithScopes[self::KEY_SCOPES]);
+
+            $logger?->info('Products recalculated', [
+                'product_ids' => Json::encode($idsToRecalculate),
+                'export_scopes' => Json::encode($productIdsWithScopes[self::KEY_SCOPES]),
+                'priority' => $priority,
+            ]);
+
+            $this->productRecalculationDeduplicationFacade->delete($idsToRecalculate, $priority);
         }
     }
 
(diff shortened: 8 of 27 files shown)


project-base/app/config/packages/snc_redis.yaml
```diff
@@ -58,3 +58,11 @@ snc_redis:
                 prefix: '%env(REDIS_PREFIX)%:fe:'
                 connection_timeout: 2
                 read_write_timeout: 5
+        product_recalculation_queue:
```
henzigo: Why did you decide to use Redis instead of some RabbitMQ plugins for deduplications? Are there any benefits/downsides of this?

packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Shopsys\FrameworkBundle\Model\Product\ProductRepository;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+class DispatchProductIdsBatchMessageHandler implements MessageHandlerInterface
```
henzigo, line 12: What is the purpose of this class? I don't understand the benefit of this. It only iterates over products and then dispatches all products again

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -50,17 +56,27 @@ public function dispatchProductIds(
...
-        foreach ($productIds as $productId) {
-            $message = match ($productRecalculationPriorityEnum) {
-                ProductRecalculationPriorityEnum::HIGH => new ProductRecalculationPriorityHighMessage((int)$productId, $exportScopes),
-                ProductRecalculationPriorityEnum::REGULAR => new ProductRecalculationPriorityRegularMessage((int)$productId, $exportScopes),
-                default => throw new UnknownProductRecalculationPriorityException($productRecalculationPriorityEnum),
-            };
-            $this->messageBus->dispatch($message);
+        for ($i = 0; $i < $productIdsCount; $i += self::DISPATCH_BATCH_SIZE) {
+            $productIdsBatch = array_slice($productIds, $i, self::DISPATCH_BATCH_SIZE);
+            $idsToRecalculate = $this->productRecalculationRepository->getIdsToRecalculate($productIdsBatch);
```
henzigo: You shouldn't call repository directly and you will slow down dispatching messages with this call. Moreover, you are calling this method already in Handler so this could probably create some problems with recalculating same product variants.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
```diff
@@ -0,0 +1,135 @@
...
+    protected function buildValueByProductIdAndScopes(
+        int $productId,
+        array $scopes = [],
+    ): string {
+        if (count($scopes) > 0 && $this->contains($productId)) {
+            $cacheId = $this->getCacheId($productId);
+            $scopesString = $this->redisClient->get($cacheId);
+
+            if ($scopesString !== false && $scopesString !== '') {
+                $scopes = array_unique(array_merge($scopes, explode(',', $scopesString)));
```
henzigo: What about using JSON instead of exploding and imploding strings?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Redis;
+use Shopsys\FrameworkBundle\Component\Redis\RedisFacade;
+
+class ProductRecalculationCacheFacade
```
henzigo: What about using Hashes instead of a simple get/set? It will be grouped under one key with all options you need with benefit of executing batch operations. I'm talking about `HMGET`, `HMSET` and `HMDEL`.

Look: https://redis.io/docs/latest/commands/hmset/ and https://redis.io/docs/latest/commands/hmget/

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
```diff
@@ -0,0 +1,135 @@
...
+        protected readonly Redis $redisClient,
+        protected readonly RedisFacade $redisFacade,
+    ) {
+    }
+
+    /**
+     * @param int $productId
+     * @return bool
+     */
+    public function contains(
```
henzigo: This method is useless. You can simply call get(); if it returns false, then the key does not exist. Moreover, you will save one command for each product

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
```diff
@@ -0,0 +1,135 @@
...
+        int $productId,
+    ): string {
+        return (string)$productId;
+    }
+
+    /**
+     * @param int $productId
+     * @param string[] $scopes
+     */
+    public function save(
```
henzigo: This should be probably named better. Something like `saveOrUpdateScopes` because I was confused that you are replacing scopes instead of updating them.

henzigo (commented):
I know this PR is not ready for CR yet, but I was curious and there's result 😃

packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
janjavorek: The main purpose is optimization, if you're changing for example unit name or parameter then all affected products are dispatched at that time and if you have 50k and more affected products then you'll be waiting for a looong time to save the change.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: Yes, we already discussed this issue, and the main difference is that from now on, you're dispatching only main variants and regular products. Variants are not dispatched at all and are retrieved during the recalculation process afterward.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
janjavorek: Ok, I changed only the content and switched `exists` with `get`. I think is more readable if name of the method is `contains`

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
janjavorek: Ok, changed

project-base/app/config/packages/snc_redis.yaml
janjavorek: Redis gives you more control over messages at the application level, and I believe it is more suitable for our purposes. It allows you to handle ELS scopes and deduplicate messages at that level.

packages/framework/src/Model/Product/Recalculation/DispatchAffectedProductsSubscriber.php
```diff
@@ -119,10 +119,15 @@ public function dispatchAffectedByUnit(UnitEvent $unitEvent): void
...
+        $this->productRecalculationDispatcher->dispatchProductIdsByBatches(
             $productIds,
             ProductRecalculationPriorityEnum::REGULAR,
-            [ProductExportScopeConfig::SCOPE_UNIT],
+            [ProductExportScopeConfig::SCOPE_UNIT, ProductExportScopeConfig::SCOPE_FLAGS],
+        );
+        $this->productRecalculationDispatcher->dispatchProductIdsByBatches(
+            [82, 69],
+            ProductRecalculationPriorityEnum::REGULAR,
+            [ProductExportScopeConfig::SCOPE_CATEGORIES, ProductExportScopeConfig::SCOPE_FLAGS],
```
grossmannmartin: this seems more like a forgotten debug statement, or?

project-base/app/config/packages/messenger.yaml
```diff
@@ -13,6 +13,7 @@ framework:
             Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchAllProductsMessage: product_recalculation_priority_regular
             Shopsys\FrameworkBundle\Model\Order\Messenger\PlacedOrderMessage: placed_order_transport
             Symfony\Component\Mailer\Messenger\SendEmailMessage: send_email_transport
+            Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchProductIdsBatchMessage: product_recalculation_priority_regular
```
grossmannmartin: Shouldn't be the dispatch itself placed with a high priority? Then IDs will be deduplicated immediately

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
grossmannmartin: interesting idea, @janjavorek did you consider it?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -11,11 +11,17 @@
...
+
     /**
      * @param \Shopsys\FrameworkBundle\Model\Product\Elasticsearch\Scope\ProductExportScopeConfig $productExportScopeConfig
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationCacheFacade $productRecalculationCacheFacade
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationRepository $productRecalculationRepository
      */
     public function __construct(
         protected readonly ProductExportScopeConfig $productExportScopeConfig,
+        protected readonly ProductRecalculationCacheFacade $productRecalculationCacheFacade,
+        protected readonly ProductRecalculationRepository $productRecalculationRepository,
```
grossmannmartin: I would rather use the repository only in the related facade (in some scenarios in different facades), but I don't think it should be used here

packages/framework/src/Component/Redis/RedisClientFacade.php
```diff
@@ -20,10 +20,11 @@ public function __construct(
     /**
      * @param string $key
      * @param mixed $data
+     * @param mixed $options
      */
-    public function save(string $key, mixed $data): void
+    public function save(string $key, mixed $data, mixed $options = null): void
```
grossmannmartin: this change seems unnecessary, or?

packages/framework/src/Component/Redis/RedisClientFacade.php
```diff
@@ -52,4 +53,13 @@ public function delete(string $key): void
     {
         $this->redisClient->del($key);
     }
+
+    /**
+     * @param string $key
+     * @return mixed
+     */
+    public function get(string $key): mixed
```
grossmannmartin: this method seems unused

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
```diff
@@ -0,0 +1,136 @@
...
+                $result[$productId] = array_unique(array_merge($cachedScopes, $exportScopes));
+            } else {
+                $result[$productId] = [];
+            }
+        }
+
+        return $result;
+    }
+
+    public function clear(): void
```
grossmannmartin: This method is unused, and the implementation is confusing. ProductRecalculationCacheFacade accepts a scoped redis client (`@snc_redis.product_recalculation_queue`), but the clean method will clear all redis clients, which is unexpected.

I suggest to remove the clear method and the dependency on the RedisFacade

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -96,4 +113,75 @@ protected function verifyExportScopes(array $exportScopes): void
...
+    protected function filterNotCachedProductIds(
+        array $productIds,
+        array $scopes,
+        string $productRecalculationPriorityEnum,
+    ): array {
+        $filteredProductIds = [];
+        $productIds = array_unique($productIds);
+
+        foreach ($productIds as $productId) {
+            if ($productRecalculationPriorityEnum === ProductRecalculationPriorityEnum::HIGH) {
```
grossmannmartin: does this mean that the high priority will not be deduplicated?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -96,4 +113,75 @@ protected function verifyExportScopes(array $exportScopes): void
...
+                $filteredProductIds[] = $productId;
+
+                continue;
+            }
+
+            $exists = $this->productRecalculationCacheFacade->contains($productId);
```
grossmannmartin: this may be called for multiple ids at once in redis (so we don't have to query redis 2000 times). 

also `contains` is called even in the saveOrUpdateScopes, which worsens performance

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -49,18 +55,29 @@ public function dispatchProductIds(
...
-        foreach ($productIds as $productId) {
-            $message = match ($productRecalculationPriorityEnum) {
-                ProductRecalculationPriorityEnum::HIGH => new ProductRecalculationPriorityHighMessage((int)$productId, $exportScopes),
-                ProductRecalculationPriorityEnum::REGULAR => new ProductRecalculationPriorityRegularMessage((int)$productId, $exportScopes),
-                default => throw new UnknownProductRecalculationPriorityException($productRecalculationPriorityEnum),
-            };
-            $this->messageBus->dispatch($message);
+        for ($i = 0; $i < $productIdsCount; $i += self::DISPATCH_BATCH_SIZE) {
+            $productIdsBatch = array_slice($productIds, $i, self::DISPATCH_BATCH_SIZE);
+            $replaceVariantProductIds = $this->productRecalculationRepository->replaceVariantIdsWithMainVariantIds($productIdsBatch);
```
grossmannmartin: During dispatchAllProductsMessageHandler processing, this method is called, even though the product IDs does not contain any variant ID (the `iterateAllProductIdsExceptVariant()` method is used)

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -96,4 +113,75 @@ protected function verifyExportScopes(array $exportScopes): void
...
+            ),
+        );
+    }
+
+    /**
+     * @param array $productIds
+     * @param string $productRecalculationPriorityEnum
+     * @param string[] $exportScopes
+     */
+    public function dispatchProductIdsByBatches(
```
grossmannmartin: Doesn't this bring confusion to the developer when to use this and when a dispatchProductIds? 
What about unifying it and making it a simpler API to work with?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
```diff
@@ -0,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Redis;
+use Shopsys\FrameworkBundle\Component\Redis\RedisFacade;
+
+class ProductRecalculationCacheFacade
```
grossmannmartin: I'm not sure about the use of word cache here... it's not used as a traditional cache. I would vote for `Deduplication,` so the purpose is clear, or something more describing the purpose.

grossmannmartin (changes requested):

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
janjavorek: Ok. It's better. I changed it and the deduplication is using Hashes now.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
janjavorek: Ok, I changed the name to ProductRecalculationDeduplicationFacade

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: Ok, dispatchProductIds is now protected and you're supposed to use method dispatchProductIdsByBatches or dispatchSingleProductId

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: It should be fixed by implementing Hashes

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: yes

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * @param int[] $productIds
+     */
+    public function delete(
+        array $productIds,
+    ): void {
+        foreach ($productIds as $productId) {
+            $this->redisClient->hDel($this->getCacheKey(), (string)$productId);
+        }
```
henzigo:
```suggestion
       $this->redisClient->hDel($this->getCacheKey(), ...$productIds);
```

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: I could not use the ProductRecalculationFacade directly due to memory issues, so I left it as it is. We can have a call if you need further explanation.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationCacheFacade.php
janjavorek: Ok, I changed it. It's better

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: Yes, but I think the problem is not that significant, and I cannot think of an easy solution for it. You need to call the replacement function in all other cases anyway; only the DispatchAllProductsMessageHandler calls are unnecessary.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
janjavorek: I couldn't do it this way because hDel accepts only strings as fields, and I would have needed to convert the entire array of integers to strings.

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
```diff
@@ -143,7 +143,12 @@ protected function executeIds(
             return Command::FAILURE;
         }
 
-        $dispatchedProductIds = $this->productRecalculationDispatcher->dispatchProductIds($productIds, $priority, $scopes);
+        $dispatchedProductIds = [];
+
+        foreach ($productIds as $productId) {
+            $this->productRecalculationDispatcher->dispatchSingleProductId($productId, $priority, $scopes);
+            $dispatchedProductIds[] = $productId;
+        }
```
grossmannmartin: the previous implementation returned truly dispatched IDs, for example if main_variant was provided, it returned main variant id and appropriate variant ids. 

What will happen when I dispatch variants from the same main variant? Do I get several identical messages?
Is there a reason why not to use batch dispatch?

All in all, the dispatchedProductIds just creates a copy of `$productIds` array, so in its current form, I think it's redundant.

packages/framework/src/Model/Product/ProductFacade.php
```diff
@@ -163,7 +163,7 @@ public function edit(
 
         if ($product->isMainVariant()) {
             $removedVariantIds = $product->refreshVariants($productData->variants);
-            $this->productRecalculationDispatcher->dispatchProductIds($removedVariantIds, $priority);
+            $this->productRecalculationDispatcher->dispatchProductIdsByBatches($removedVariantIds, $priority);
```
grossmannmartin: I would keep the name `dispatchProductIds`. The fact that it does it "by batches" is an inner implementation that should not be a concern of a caller.

packages/framework/src/Model/Product/Recalculation/DispatchAffectedProductsSubscriber.php
```diff
@@ -119,10 +119,10 @@ public function dispatchAffectedByUnit(UnitEvent $unitEvent): void
     {
         $productIds = $this->affectedProductsFacade->getProductIdsWithUnit($unitEvent->getUnit());
 
-        $this->productRecalculationDispatcher->dispatchProductIds(
+        $this->productRecalculationDispatcher->dispatchProductIdsByBatches(
             $productIds,
             ProductRecalculationPriorityEnum::REGULAR,
-            [ProductExportScopeConfig::SCOPE_UNIT],
+            [ProductExportScopeConfig::SCOPE_UNIT, ProductExportScopeConfig::SCOPE_FLAGS],
```
grossmannmartin: when a change in a unit is made, flag scope should be also dispatched? 
It's unexpected

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,114 @@
...
+    public function __construct(
+        protected readonly Redis $redisClient,
+    ) {
+    }
+
+    /**
+     * @param int[] $productIds
+     * @return array
+     */
+    public function getNotCachedKeys(array $productIds): array
```
grossmannmartin: Are those NOT cached keys? it seems like not cached keys are skipped (`!array_key_exist`)

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
```diff
@@ -143,7 +143,12 @@ protected function executeIds(
             return Command::FAILURE;
         }
 
-        $dispatchedProductIds = $this->productRecalculationDispatcher->dispatchProductIds($productIds, $priority, $scopes);
+        $dispatchedProductIds = [];
+
+        foreach ($productIds as $productId) {
+            $this->productRecalculationDispatcher->dispatchSingleProductId($productId, $priority, $scopes);
```
grossmannmartin: `./bin/console shop:dis:rec 153 152` command returns following error

```
In ProductRecalculationDispatcher.php line 88:

  Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDispatcher::dispatchSingleProductId(): Argument #1 ($productId) must be of type int, string given, cal
  led in /var/www/html/packages/framework/src/Command/DispatchRecalculationMessageCommand.php on line 149
```

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $result = [];
+        $productIdsCached = $this->redisClient->hMget($this->getCacheKey(), array_map('strval', $productIds));
```
grossmannmartin: I'm not sure about this implementation. hmget will return the provided keys with false value even if it doesn't exist in redis... 
try 
```
d($this->redisClient->hMget($this->getCacheKey(), ['5242', '845', '9996']));
```

will return 
```
^ array:3 [
  5242 => false
  845 => false
  9996 => false
]
```

and the next code takes a false assumption about non-existing keys

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * @param int[] $productIds
+     * @throws \RedisException
+     * @return array
+     */
+    protected function getScopesByProductIds(array $productIds): array
+    {
+        $scopes = $this->redisClient->hMget($this->getCacheKey(), array_map('strval', $productIds));
+
+        return array_map(static fn (string|bool $scope) => $scope === false ? [] : json_decode($scope, true), $scopes);
```
grossmannmartin: try dispatch not-existing product id 
```
./bin/console shop:dis:rec 153 56663
```

you will get 
```
In ProductRecalculationDeduplicationFacade.php line 80:

  array_map(): Argument #2 ($array) must be of type array, false given
```

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * @param string[][] $exportScopesIndexedByProductId
+     * @return array<int,string[]>
+     */
+    public function getScopesIndexedByProductId(array $exportScopesIndexedByProductId): array
+    {
+        $result = [];
+
+        foreach ($exportScopesIndexedByProductId as $productId => $exportScopes) {
+            $cachedScopeByProductId = $this->redisClient->hGet($this->getCacheKey(), $productId);
```
grossmannmartin: can we use `hmget` outside foreach?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
grossmannmartin: I think creating the `replaceVariantIdsWithMainVariantIds` method in ProductRecalculationFacade should not cause memory issues, or?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -96,4 +113,63 @@ protected function verifyExportScopes(array $exportScopes): void
...
+        array $scopes,
+        string $productRecalculationPriorityEnum,
+    ): array {
+        $productIds = array_unique($productIds);
+
+        $filteredProductIds = array_filter($productIds, static fn (int $productId) => $productRecalculationPriorityEnum === ProductRecalculationPriorityEnum::HIGH);
+
+
+        $notCachedProductIds = $this->productRecalculationDeduplicationFacade->getNotCachedKeys($productIds);
+        $this->productRecalculationDeduplicationFacade->saveOrUpdateScopes($productIds, $scopes);
```
grossmannmartin: I would not expect this side-effect from a method named `filter...`

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
```diff
@@ -21,57 +22,96 @@ public function __construct(
      * @param int[] $productIds
      * @return int[]
      */
-    public function getIdsToRecalculate(array $productIds): array
+    protected function dropVariantIdsFromProductIds(array $productIds): array
```
grossmannmartin: this method seems unused

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
```diff
@@ -21,57 +22,96 @@ public function __construct(
...
+            $variantIds = array_filter($variantIdsIndexedByMainVariantIds[$mainVariantId], static fn (int $variantId): bool => $variantId !== $mainVariantId && $variantsCount > 1);
+
+            $structuredResults = [...$structuredResults, ...$variantIds];
+
+            if (!in_array($mainVariantId, $variantIds, true)) {
+                $structuredResults[] = $mainVariantId;
+            }
+        }
+
+        return $structuredResults;
```
grossmannmartin: this seems more like a `Facade` code

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
```diff
@@ -21,57 +22,96 @@ public function __construct(
...
+        return $groupedResults;
     }
 
     /**
-     * @param int[] $productIds
+     * @param int[] $mainVariantsIds
      * @return int[]
      */
-    protected function dropVariantIdsFromProductIds(array $productIds): array
+    public function getIdsToRecalculate(array $mainVariantsIds): array
```
grossmannmartin: Maybe I don't understand the code enough (and I truly don't understand it enough 😆 )
but wouldn't be this sql enough to get proper ids?

```
SELECT id FROM product WHERE id IN (/*main_variant_ids*/) OR main_variant_id IN (/*main_variant_ids*/)
```

grossmannmartin (changes requested):

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: There seems to be an issue with ELS. Is using the Repository instead of the Facade a significant enough problem that I should take the time to resolve it?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
janjavorek: ok, method was moved to the ProductRecalculationFacade

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
janjavorek: removed

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: I moved it outside of the method

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
janjavorek: This was discussed during the call, and the resolution is that we left it as it is. The Repository is used.

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
janjavorek: dispatchedProductIds is now used, and the method also follows the batch processing principle, also discussed during the call.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
janjavorek: Yes, the logic was fixed, and the method now correctly returns notCachedKeys

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
janjavorek: Yes, I fixed the code, and it should now work correctly with the response from hMget

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
janjavorek: Yes, I moved it outside the foreach

packages/framework/src/Model/Product/Recalculation/ProductRecalculationRepository.php
janjavorek: I'm not sure, but this method is also supposed to work not only with main variants but also with variants and regular products, and it should return the appropriate IDs regardless of the input. I think this query might not be sufficient.

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
```diff
@@ -143,9 +143,10 @@ protected function executeIds(
             return Command::FAILURE;
         }
 
-        $dispatchedProductIds = $this->productRecalculationDispatcher->dispatchProductIds($productIds, $priority, $scopes);
```
grossmannmartin, line 146: If I dispatch a message with a non-existing scope, the command behaves like nothing is wrong, but then just silently fails and moves the message to the failed queue. 
It would be nice to check scopes before and even check if the code warns programmer if he tries to export a non-existing scope programmatically (e.g., with `dispatchProductIds` method)

packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
```diff
@@ -0,0 +1,35 @@
...
+    ) {
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Recalculation\DispatchProductIdsBatchMessage $message
+     */
+    public function __invoke(DispatchProductIdsBatchMessage $message): void
+    {
+        foreach ($message->productIds as $productId) {
+            $this->productRecalculationDispatcher->dispatchSingleProductId(
```
grossmannmartin: It would be nice to know that the message was consumed. Now it seems like it does nothing.

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @param int[] $productIds
+     * @param string[] $scopes
+     */
+    public function saveOrUpdateScopes(
+        array $productIds,
+        array $scopes = [],
+    ): void {
+        $cacheKey = $this->getCacheKey();
+        $existingScopes = $this->getScopesByProductIds($productIds);
```
grossmannmartin: As we spoke, it seems like this method doesn't return the product IDs properly, causing the following code to malfunction (nothing is stored in Redis, and no deduplication takes place).

grossmannmartin (changes requested):

packages/framework/src/Command/DispatchRecalculationMessageCommand.php
janjavorek: Yes, scopes are now verified before dispatching products batch

packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
janjavorek: Ok I added info message about dispatched batch of product ids

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
janjavorek: fixed

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+
+        //        foreach ($cachedScopes as $productId => $scopes) {
+        //            if ($scopes === false) {
+        //                continue;
+        //            }
+        //
+        //            $scopesByProductIds[(int)$productId] = json_decode($scopes, true);
+        //        }
```
grossmannmartin: please remove the commented code

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public function delete(
+        array $productIds,
+    ): void {
+        $this->redisClient->hDel($this->getCacheKey(), ...array_map('strval', $productIds));
+    }
+
+    /**
+     * @param int[] $productIds
+     * @throws \RedisException
```
grossmannmartin: please remove `@throws` annotation as we don't write them

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,163 @@
...
+            } else {
+                $result[$productId] = [];
+            }
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param array $productIds
```
grossmannmartin:
```suggestion
     * @param int[] $productIds
```

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,163 @@
...
+        $resultValuesIndexedByProductIds = $this->redisClient->hMget($this->getCacheKey(), array_map('strval', $productIds));
+
+        $cachedScopesIndexedByProductIds = [];
+
+        foreach ($resultValuesIndexedByProductIds as $productId => $values) {
+            if ($values === false) {
+                continue;
+            }
+
+            $cachedScopesIndexedByProductIds[(int)$productId] = $values;
```
grossmannmartin: I have noticed retyping to int at several places, are those necessary?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,163 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Redis;
+
+class ProductRecalculationDeduplicationFacade
```
grossmannmartin, line 10: It would be really awesome to have some tests for this class and maybe for other parts of the PR 😇

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -41,26 +47,58 @@ public function dispatchProducts(
...
         $this->verifyExportScopes($exportScopes);
+
         $productIds = array_unique($productIds);
+        $dispatchedIds = [];
+        $productIdsCount = count($productIds);
+
+        for ($i = 0; $i < $productIdsCount; $i += self::DISPATCH_BATCH_SIZE) {
+            $productIdsBatch = array_slice($productIds, $i, self::DISPATCH_BATCH_SIZE);
+            $replaceVariantProductIds = array_unique($this->productRecalculationRepository->replaceVariantIdsWithMainVariantIds($productIdsBatch));
+            $filteredProductIds = $this->filterNotCachedProductIds($replaceVariantProductIds, $productRecalculationPriorityEnum);
```
grossmannmartin: can't we directly use `ProductRecalculationDeduplicationFacade::getNotCachedKeys()`? 

unique on array is called twice in vain, and if we start deduplicating the high queue too, we can simplify the code a bit

packages/framework/src/Model/Product/Recalculation/ProductRecalculationFacade.php
```diff
@@ -128,4 +150,27 @@ protected function groupProductIdsWithSameScopes(array $exportScopesIndexedByPro
 
         return $result;
     }
+
+    /**
+     * @param int[] $mainVariantsIds
+     * @return int[]
+     */
+    public function getIdsToRecalculate(array $mainVariantsIds): array
```
grossmannmartin: ~if I dispatch `./bin/console shop:dis:rec 82 69` (both main variants), nothing is exported to elasticsearch.~

⚠️  edit: I have noticed I have all products in the hashmap. This happens when you run `./phing db-demo`, so after clean install, you're not able to update any products

------

Also, this method seems really complex.

This method should return proper ids to recalculate - so it should return all variants, main variants and regular products?
if that's the case, I don't think why selecting by `id` or `main_variant_id` should not work 🤔  let's discuss it please

grossmannmartin (dismissed):

project-base/app/phpstan.neon
```diff
@@ -30,9 +30,6 @@ parameters:
             # phpstan-symfony extension cannot work right with symfony test container at the moment https://github.com/phpstan/phpstan-symfony/issues/27
             message: '#^Service "[^"]+" is private.$#'
             path: %currentWorkingDirectory%/tests/*
-        -
-            message: '#^Service "test.service_container" is not registered in the container.$#'
-            path: %currentWorkingDirectory%/tests/*
         -
             message: '#^Service "snc_redis.test" is not registered in the container.$#'
             path: %currentWorkingDirectory%/tests/App/Functional/Component/Redis/Redis*FacadeTest.php
```
vitek-rostislav: it looks like the entry below (`snc_redis.test`) should have been removed, not this one (`test.service_container`), right?

packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
```diff
@@ -0,0 +1,51 @@
...
+            $message->exportScopes,
+        );
+
+        if ($dispatchedProductIds === []) {
+            $this->logger->info('No new products were dispatched as they are already in the queue.', [
+                'product_ids' => Json::encode($message->productIds),
+                'export_scopes' => Json::encode($message->exportScopes),
+            ]);
+
+            return;
```
vitek-rostislav, line 43: I am wondering, this message might sometimes be a little bit misleading in the log because (if I understand it correctly) at least export scopes might have been updated here, right? Maybe (and just maybe) we could add that information here. But I am not sure whether it is worth the additional complexity. Let's discuss it please

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
```diff
@@ -0,0 +1,151 @@
...
+        foreach ($exportScopesIndexedByProductIds as $productId => $storedScopesJson) {
+            if ($storedScopesJson === false) {
+                $productIdsToDispatch[] = $productId;
+            }
+
+            $updatedScopesJson = $this->updateScopesByRequestedScopes($requestedExportScopes, $storedScopesJson);
+            $exportScopesIndexedByProductIds[$productId] = $updatedScopesJson;
+        }
+
+        if ($this->isMessengerActive) {
```
vitek-rostislav: I am wondering, does it make any sense to perform anything above when messenger is not active?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
```diff
@@ -41,26 +44,38 @@ public function dispatchProducts(
...
+            $productIdsBatch = array_slice($productIds, $i, static::DISPATCH_BATCH_SIZE);
+
+            $this->messageBus->dispatch(
+                new DispatchProductIdsBatchMessage(
+                    $productIdsBatch,
+                    $exportScopes,
+                    $productRecalculationPriorityEnum,
+                ),
+            );
+        }
```
vitek-rostislav, line 78: Would not this do the same trick in a more elegant way?

```
foreach (array_chunk($productIds, static::DISPATCH_BATCH_SIZE) as $productIdsBatch) {
    $this->messageBus->dispatch(
        new DispatchProductIdsBatchMessage(
            $productIdsBatch,
            $exportScopes,
            $productRecalculationPriorityEnum,
        ),
    );
}
```

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcherExecutor.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\Recalculation;
+
+use Symfony\Component\Messenger\MessageBusInterface;
+
+/**
+ * @internal This class is not intended to be used directly from your code
```
vitek-rostislav: maybe we could add a hint of what should be used "instead" - `ProductRecalculationDispatcher` itself, right?

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcherExecutor.php
```diff
@@ -0,0 +1,64 @@
...
+        $productIdsWithMainVariants = $this->productRecalculationRepository->replaceVariantIdsWithMainVariantIds($productIds);
+
+        $productIdsToDispatch = $this->productRecalculationDeduplicationFacade->updateScopesAndReturnProductIdsToDispatch(
+            $productIdsWithMainVariants,
+            $exportScopes,
+        );
+
+        if ($productRecalculationPriorityEnum === ProductRecalculationPriorityEnum::HIGH) {
+            $productIdsToDispatch = array_merge($productIdsToDispatch, $productIds);
+        }
```
vitek-rostislav: I do not understand this, could you explain it, please?

project-base/app/config/packages/snc_redis.yaml
```diff
@@ -50,6 +50,14 @@ snc_redis:
                 prefix: '%env(REDIS_PREFIX)%:%build-version%:cache:main_friendly_url_slugs:'
                 connection_timeout: 2
                 read_write_timeout: 5
+        product_recalculation_queue:
```
vitek-rostislav, line 53: is this queue persisted during deployment? In other words, is it cleaned within `clean-redis(-old)` call? :eyes: 

When I call `shopsys:redis:clean-cache`, it iterates through all the clients, including `:queue:recalculation:`, is that intentional?

project-base/app/tests/App/Functional/Model/Product/Recalculation/ProductRecalculationDeduplicationFacadeTest.php
```diff
@@ -0,0 +1,195 @@
...
+
+namespace Tests\App\Functional\Model\Product\Recalculation;
+
+use Nette\Utils\Json;
+use PHPUnit\Framework\Attributes\DataProvider;
+use Redis;
+use Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDeduplicationFacade;
+use Tests\App\Test\TransactionFunctionalTestCase;
+
+class ProductRecalculationDeduplicationFacadeTest extends TransactionFunctionalTestCase
```
vitek-rostislav, line 14: love it :heart_eyes:

upgrade-notes/backend_20250206_212223.md
```diff
@@ -0,0 +1,13 @@
+#### add product recalculation queue deduplication ([#3669](https://github.com/shopsys/shopsys/pull/3669))
+
+- ensure you have Redis on version 7.4 (see #3673)
```
vitek-rostislav:
```suggestion
- ensure you have Redis on version 7.4 (see [#3673](https://github.com/shopsys/shopsys/pull/3673))
```

upgrade-notes/backend_20250206_212223.md
```diff
@@ -0,0 +1,13 @@
...
+- method use `\Shopsys\FrameworkBundle\Model\Product\ProductFacade::iterateAllProductIds()` was removed, use `iterateAllProductIdsExceptVariant()`instead
+- method use `\Shopsys\FrameworkBundle\Model\Product\ProductRepository::iterateAllProductIds()` was removed, use `iterateAllProductIdsExceptVariant()`instead
+- `ProductRecalculationMessage` messages no longer contain `exportScopes`. Scopes are now stored in Redis.
```
vitek-rostislav: maybe it is safer to mention the particular classes(s) - `AbstractProductRecalculationMessage` and its descendants - `ProductRecalculationPriorityHighMessage` and `ProductRecalculationPriorityRegularMessage`

upgrade-notes/backend_20250206_212223.md
```diff
@@ -0,0 +1,13 @@
...
+- `Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationRepository`
+    - method `getIdsToRecalculate()` was removed
+    - method `getMainVariantIdsOfVariants()` was removed
+    - method `dropVariantIdsFromProductIds()` was removed
+    - method `getVariantIds()` was removed
+    - use `replaceVariantIdsWithMainVariantIds()` or `getIdsToRecalculateByMainVariantIds()` instead
+- see #project-base-diff to update your project
```
vitek-rostislav, line 15: maybe it is worth mentioning the docs as well? `docs/asynchronous-processing/product-recalculations.md`

docs/asynchronous-processing/product-recalculations.md
```diff
@@ -38,9 +38,21 @@ When you need to recalculate visibility of all products, you should use the `Sho
...
-## Dispatch recalculation message when indirect change is made
+## Product Deduplication and Batch Dispatching
 
-Sometimes it is necessary to trigger recalculation for product when some indirect change is made.
+To optimize performance and avoid redundant processing, the recalculation dispatch mechanism incorporates product deduplication and batch dispatching strategies.
+Export scopes ([mentioned later in this article](#recalculation-scope)) are stored in Redis, allowing products to remain in the queue only once while still supporting the addition of new scopes dynamically.
+This ensures that if a product ID is already present in Redis, it will not be dispatched again, preventing unnecessary recalculations and improving efficiency.
+Additionally, if a product ID is not yet present in Redis, Shopsys Platform ensures that all scopes are exported, guaranteeing that all relevant data is properly processed.
+
+When a large number of product IDs, typically more than 2000, are dispatched at once due to i.e., affected changes, Shopsys Platform takes an optimized approach.
```
vitek-rostislav: I would add something like `see the [next section](#dispatch-recalculation-message-when-an-indirect-change-is-made)` after the mention about "affected changes"

vitek-rostislav (changes requested):
:rocket:

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
grossmannmartin: array_chunk is slower 😇  

I had the same twitch and did some measurements. For large arrays, this is faster way

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcherExecutor.php
grossmannmartin: Sure... basically high priority queue is always dispatched. 

Imagine following scenario: 
- you dispatch a regular message `1`
- there are many messages, so `1` taking forever to handle
- you dispatch a high message `1`
- due to the item already present in Redis, the high message will be thrown away. 

This solves this issue with the most naive way - high is always dispatched. 
But it has some drawbacks... if a product was requested to recalculate with scope, the scope will be handled in high queue, and when the regular item is processed, it will be recalculated with all scope, because the redis entry is no longer.

I'm more than open to suggestions

project-base/app/config/packages/snc_redis.yaml
grossmannmartin: most definitely not... I will look at it

project-base/app/phpstan.neon
grossmannmartin: Of course... fixed in e20b7bbd3f

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDeduplicationFacade.php
grossmannmartin: make sense: aa092efe9e and 908fe4a0f1

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcherExecutor.php
grossmannmartin: good idea - 1d386b9b2b

upgrade-notes/backend_20250206_212223.md
grossmannmartin: 6a8f07b139

upgrade-notes/backend_20250206_212223.md
grossmannmartin: good idea - 0f39d5d369 (I have also fixed docs building error due to non-existing navigation item)

docs/asynchronous-processing/product-recalculations.md
grossmannmartin: yup -  ca7f9a0fc9

project-base/app/config/packages/snc_redis.yaml
grossmannmartin: great catch! fixed in b0874a0888

packages/framework/src/Model/Product/Recalculation/DispatchProductIdsBatchMessageHandler.php
grossmannmartin: I'm not sure, to be honest... I would need to pass back between two layers the calculated scopes to log it here... 
It's not worth from my point of view, but I'm open to discussion

vitek-rostislav: yeah, you have just confirmed my feelings about it, I am closing this thread

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcher.php
vitek-rostislav: ok, thanks for the info :+1:

packages/framework/src/Model/Product/Recalculation/ProductRecalculationDispatcherExecutor.php
vitek-rostislav: let's keep this open and wait until Monday :pray:

grossmannmartin: what about this? dfd53eced0

vitek-rostislav (approved):

vitek-rostislav:
Tested, works like a charm:

- :heavy_check_mark: product dispatched multiple times is actually processed only once thanks to the deduplication
- :heavy_check_mark: when the same product is dispatched with a different scope, the scopes are updated in the Redis queue, and then the product is processed only once with the union of the scopes
- :heavy_check_mark: priority high/regular are each duplicated separately - when a product is dispatched to the regular queue and then to the high priority queue, it is processed twice then - this is a wanted behavior
- :heavy_check_mark: batching of dispatched products - when dispatching a higher number of products at once, multiple messages with chunks of the products are dispatched with deduplication in mind (even if there were multiple product ids in the originally dispatched array, each product is correctly processed only once)

One note from testing:

- I would add information about priority to the log in `ProductRecalculationFacade::recalculate()`, what do you think?

pk16011990 (approved):

PR #3679, opened 2024-12-17
Search in administration

RostislavKreisinger:
in the header in the administration it is now possible to search by the name of the sections that the given administrator has allowed according to his rights

----
  - https://rk-ssp-2904-admin-search.odin.shopsys.cloud
  - https://cz.rk-ssp-2904-admin-search.odin.shopsys.cloud


packages/framework/src/Controller/Admin/DefaultController.php
```diff
@@ -476,4 +482,83 @@ protected function checkAllSliderNumericValuesAreSet(): void
             );
         }
     }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    #[Route(path: '/dashboard/search/')]
+    public function searchAction(Request $request): Response
```
grossmannmartin: _nitpick_ 
I would prefer a separate controller for this action. It seems independent of the `DefaultController`

project-base/app/assets/js/admin/search/search.js
```diff
@@ -0,0 +1,111 @@
...
+        //         const $row = $element.closest('tr');
+        //         console.log($row);
+        //         if ($row) $row.toggleClass('focused');
+        //     });
+        //
+        //     $element.on('blur', () => {
+        //         const $row = $element.closest('tr');
+        //         if ($row) $row.toggleClass('focused');
+        //     });
+        // });
```
grossmannmartin: this commented code is necessary?

packages/framework/src/Resources/views/Admin/Layout/layout.html.twig
```diff
@@ -28,6 +28,20 @@
                             <img class="web__header__logo__svg" src="{{ asset('public/admin/images/logo.svg') }}" alt="Shopsys Platform">
                         </a>
                     </div>
+                    <div class="web__header__search js-header-search">
```
grossmannmartin: I'm thinking whether it would not be better to put the search input above the menu instead of in the header.

Like this, so the purpose is clearer.
Snímek obrazovky 2024-12-18 v 16 51 02

project-base/app/assets/js/admin/search/search.js
```diff
@@ -0,0 +1,111 @@
+import Ajax from '@shopsys/framework/js/common/utils/Ajax';
+import Register from 'framework/common/utils/Register';
+
+export default class Search {
```
grossmannmartin: this javascript should be placed in the `packages/framework/assets/js/admin/search` folder

project-base/app/assets/styles/admin/todo.less
```diff
@@ -95,3 +95,83 @@
 input[type="time"] {
     cursor: text;
 }
+
+.web__header__search {
```
grossmannmartin: those styles should be placed in `packages/framework/assets/styles/admin/todo.less` file

grossmannmartin (changes requested):
This is awesome 💚

project-base/app/assets/js/admin/search/search.js
RostislavKreisinger: forgotten code, thx

packages/framework/src/Resources/views/Admin/Layout/layout.html.twig
RostislavKreisinger: i dont thing so. From real use is header better because is always visible, from product detail form is left menu hide. Ask tom mandok how he uses search.

packages/framework/src/Controller/Admin/SearchAdminController.php
```diff
@@ -0,0 +1,102 @@
...
+    protected function containsLabelSearchString(string $label, string $searchString): bool
+    {
+        return str_contains($this->convertStingWithDiacritics($label), $this->convertStingWithDiacritics($searchString));
+    }
+
+    /**
+     * @param string $string
+     * @return string
+     */
+    protected function convertStingWithDiacritics(string $string): string
```
grossmannmartin:
```suggestion
    protected function convertStringWithDiacritics(string $string): string
```

grossmannmartin (approved):

PR #3682, opened 2024-12-17
[framework] trans function now support named arguments

### LINKED ISSUE #2574: extract translation misbehaves when locale is provided along with parameters
## Describe the bug
It is possible to pass locale and translation domain to the `t` function. 

The method header is like 
```php
/**
 * @param string $id
 * @param array $parameters
 * @param string|null $domain
 * @param string|null $locale
 * @return string
 */
function t($id, array $parameters = [], $domain = null, $locale = null)
```

Following calls works fine
```php
t('my string');
t('my %adjective% string', ['%adjective%' => 'awesome']);
t('my string', locale: 'cs');
t('my string, locale: $locale);
```

However following throws an exception on `php phing translations-dump` command

```php
t('my %adjective% string', ['%adjective%' => 'awesome'], locale: $locale);
// Can only extract the message ID or message domain from a scalar, concatenated string or "\Shopsys\FrameworkBundle\Component\Translation\Translator" class constant, but got "PhpParser\Node\Expr\Variable".

// third parameter is not actually a translation domain, but named argument (fourth)
```

```php
t('my %adjective% string', ['%adjective%' => 'awesome'], null, $locale);
// Can only extract the message ID or message domain from a scalar, concatenated string or "\Shopsys\FrameworkBundle\Component\Translation\Translator" class constant, but got "PhpParser\Node\Expr\ConstFetch".

// null is not accepted even though it's the default value. Maybe domain should not be nullable and Translator::DEFAULT_DOMAIN should be the default value
```

Maybe it would be nice to check following (I suspect it will not extract into proper translation domain file)
```php
t('my string', domain: 'another-translation-domain')
```

### Expected behavior
No error is raised 🙂

henzigo:
#### Fixes issues
closes #2574 

----
  - https://jg-trans-named-arguments.odin.shopsys.cloud
  - https://cz.jg-trans-named-arguments.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/Translation/functions.php
+++ b/packages/framework/src/Component/Translation/functions.php
@@ -7,11 +7,11 @@ use Shopsys\FrameworkBundle\Component\Translation\Translator;
 /**
  * @param string $id
  * @param array $parameters
- * @param string|null $translationDomain
+ * @param string|null $domain
  * @param string|null $locale
  * @return string
  */
-function t(string $id, array $parameters = [], ?string $translationDomain = null, ?string $locale = null): string
+function t(string $id, array $parameters = [], ?string $domain = null, ?string $locale = null): string
 {
-    return Translator::staticTrans($id, $parameters, $translationDomain, $locale);
+    return Translator::staticTrans($id, $parameters, $domain, $locale);
 }
--- a/packages/framework/src/Component/Translation/functions.php
+++ b/packages/framework/src/Component/Translation/functions.php
@@ -7,7 +7,7 @@ use Shopsys\FrameworkBundle\Component\Translation\Translator;
 /**
  * @param string $id
  * @param array $parameters
- * @param string|null $domain
+ * @param string|null $domain Translation domain (default is "messages")
  * @param string|null $locale
  * @return string
  */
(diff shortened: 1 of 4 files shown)


packages/framework/src/Component/Translation/functions.php
```diff
@@ -7,11 +7,11 @@
 /**
  * @param string $id
  * @param array $parameters
- * @param string|null $translationDomain
+ * @param string|null $domain
```
TomasLudvik: Please revert this change, as this naming was intentional to prevent confusion with e-shop domain.

TomasLudvik (changes requested):

packages/framework/src/Component/Translation/functions.php
henzigo: However, this will be inconsistent with other trans functions used in PHP or backend JS.

henzigo: I have updated phpdoc with description of what you can pass to this parameter. I think, that we should be consistent with parameters naming across these functions.

vitek-rostislav: I vote for `$translationDomain` so it is obvious here at least at the most exposed interface (we could rename all the "domain" usages but this one seems crucial. The comment could be a nice solution but it is not visible every time:
image

Screenshot from 2024-12-20 11-22-58

image

TomasLudvik: I have asked @vitek-rostislav for 3rd opinion, so we can have decision here and he agreed with me, so please change it as requested.

henzigo: I'm closing this since you backed down after your call with me

vitek-rostislav (approved):
Tested, works like a charm :ok_hand:

PR #3687, opened 2024-12-20
NoDomainSelectedException is no longer throwed during maintenance page + additional Sentry context

TomasLudvik:

Additional context is sent to Sentry to help with debugging.

!!! Branch is deployed on alpha to test if context is sent to Sentry correctly.

#### Fixes issues
... 

----
  - https://tl-sentry.odin.shopsys.cloud
  - https://cz.tl-sentry.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/AddSentryContextSubscriber.php
+++ b/packages/framework/src/Component/AddSentryContextSubscriber.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component;
+
+use App\Environment;
+use RedisException;
+use Sentry\State\Scope;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\Domain\Exception\NoDomainSelectedException;
+use Shopsys\FrameworkBundle\Component\Localization\DisplayTimeZoneProviderInterface;
+use Shopsys\FrameworkBundle\Component\Maintenance\MaintenanceModeFacade;
+use Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser;
+use Shopsys\FrameworkBundle\ShopsysFrameworkBundle;
+use Symfony\Component\Console\ConsoleEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Filesystem\Exception\FileNotFoundException;
+use Symfony\Component\HttpKernel\KernelEvents;
+use function Sentry\configureScope;
+
+class AddSentryContextSubscriber implements EventSubscriberInterface
+{
+    /**
+     * @param \Shopsys\FrameworkBundle\Component\Maintenance\MaintenanceModeFacade $maintenanceModeFacade
+     * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @param \Shopsys\FrameworkBundle\Component\Localization\DisplayTimeZoneProviderInterface $displayTimeZoneProvider
+     * @param \Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser $currentCustomerUser
+     */
+    public function __construct(
+        protected readonly MaintenanceModeFacade $maintenanceModeFacade,
+        protected readonly Domain $domain,
+        protected readonly DisplayTimeZoneProviderInterface $displayTimeZoneProvider,
+        protected readonly CurrentCustomerUser $currentCustomerUser,
+    ) {
+    }
+
+    public function setSentryContext(): void
+    {
+        $context = [
+            'environment' => class_exists('App\Environment') ? Environment::getEnvironment() : '',
+            'version' => ShopsysFrameworkBundle::VERSION,
+            'systemTimeZone' => date_default_timezone_get(),
+            'userId' => $this->currentCustomerUser->findCurrentCustomerUser()?->getId(),
+        ];
+
+        try {
+            $context['maintenance-page'] = $this->maintenanceModeFacade->isEnabled();
+        } catch (RedisException) {
+        }
+
+        try {
+            $context['currentDomainId'] = $this->domain->getId();
+            $context['currentDomainName'] = $this->domain->getName();
+            $context['currentDomainLocale'] = $this->domain->getLocale();
+            $context['displayTimeZone'] = $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($this->domain->getId())->getName();
+        } catch (NoDomainSelectedException | FileNotFoundException) {
+        }
+
+        configureScope(function (Scope $scope) use ($context): void {
+            $scope->setContext('Shopsys', $context);
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            KernelEvents::REQUEST => 'setSentryContext',
+            ConsoleEvents::COMMAND => 'setSentryContext',
+        ];
+    }
+}
--- a/packages/framework/src/Component/Collector/ShopsysFrameworkDataCollector.php
+++ b/packages/framework/src/Component/Collector/ShopsysFrameworkDataCollector.php
@@ -6,6 +6,7 @@ namespace Shopsys\FrameworkBundle\Component\Collector;
 
 use PharIo\Version\Version;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\Domain\Exception\NoDomainSelectedException;
 use Shopsys\FrameworkBundle\Component\Localization\DisplayTimeZoneProviderInterface;
 use Shopsys\FrameworkBundle\ShopsysFrameworkBundle;
 use Symfony\Component\HttpFoundation\Request;
@@ -34,12 +35,20 @@ class ShopsysFrameworkDataCollector extends DataCollector
             'version' => ShopsysFrameworkBundle::VERSION,
             'docsVersion' => $this->resolveDocsVersion(ShopsysFrameworkBundle::VERSION),
             'domains' => $this->domain->getAll(),
-            'currentDomainId' => $this->domain->getId(),
-            'currentDomainName' => $this->domain->getName(),
-            'currentDomainLocale' => $this->domain->getLocale(),
             'systemTimeZone' => date_default_timezone_get(),
-            'displayTimeZone' => $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($this->domain->getId())->getName(),
         ];
+
+        try {
+            $this->data['currentDomainId'] = $this->domain->getId();
+            $this->data['currentDomainName'] = $this->domain->getName();
+            $this->data['currentDomainLocale'] = $this->domain->getLocale();
+            $this->data['displayTimeZone'] = $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($this->domain->getId())->getName();
+        } catch (NoDomainSelectedException) {
+            $this->data['currentDomainId'] = 0;
+            $this->data['currentDomainName'] = '-';
+            $this->data['currentDomainLocale'] = '-';
+            $this->data['displayTimeZone'] = '-';
+        }
     }
 
     /**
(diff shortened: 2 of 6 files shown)


project-base/app/src/Component/Domain/DomainAwareSecurityHeadersSetter.php
```diff
@@ -29,7 +30,11 @@ public function onKernelResponse(ResponseEvent $event): void
             return;
         }
 
-        if (!$this->domain->isHttps()) {
+        try {
+            if (!$this->domain->isHttps()) {
+                return;
+            }
+        } catch (NoDomainSelectedException) {
```
grossmannmartin: I'm not sure if this solution will just postpone the problem into the future.
It's really easy to implement a feature that will use the domain during maintenance page.
Isn't possible to resolve domain before the maintenance page subscriber?

packages/framework/src/Component/AddSentryContextSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+    public function setSentryContext(): void
+    {
+        $context = [
+            'environment' => class_exists('App\Environment') ? Environment::getEnvironment() : '',
+            'version' => ShopsysFrameworkBundle::VERSION,
+            'systemTimeZone' => date_default_timezone_get(),
+        ];
+
+        try {
+            $this->redisFacade->pingAllClients();
```
henzigo: Why is this call needed? this is a heavy operation and it can be performance problem :)

project-base/app/src/Component/Domain/DomainAwareSecurityHeadersSetter.php
vitek-rostislav: I had similar worries, we can brainstorm it @TomasLudvik

packages/framework/src/Component/AddSentryContextSubscriber.php
vitek-rostislav: is it heavy indeed? I thought the ping is quite a lightweight operation...Do you have a better suggestion how to check whether the redis is available? The reason is that this subscriber is called on console command, and one command is called during the image build as a composer post install script (`shopsys:domains-urls:configure`). In that case, Redis is not available hence you are not able to check whether the maintenance mode is enabled :grimacing:

packages/framework/src/Component/AddSentryContextSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+        protected readonly Domain $domain,
+        protected readonly DisplayTimeZoneProviderInterface $displayTimeZoneProvider,
+        protected readonly RedisFacade $redisFacade,
+    ) {
+    }
+
+    public function setSentryContext(): void
+    {
+        $context = [
+            'environment' => class_exists('App\Environment') ? Environment::getEnvironment() : '',
```
vitek-rostislav, line 41: is this ever going to be anything else from `PRODUCTION`? In the other environment, there should not be sentry reporting at all, or? https://github.com/shopsys/shopsys/blob/16.0/project-base/app/config/packages/sentry.yaml#L1

packages/framework/src/Component/AddSentryContextSubscriber.php
vitek-rostislav: ah, maybe we can remove this line as `$this->redisClientFacade->contains` should also throw `RedisException` if redis is not available :thinking: :slightly_smiling_face:

vitek-rostislav (changes requested):

packages/framework/src/Component/AddSentryContextSubscriber.php
henzigo: Yes, it's a heavy operation. It opens ~10 (It depends on how many it was already connected during request) connections to Redis and every connection is time consuming

packages/framework/src/Component/AddSentryContextSubscriber.php
henzigo: What if you will turn it on ? :D Like I would love to be able to easily enable Sentry on localhost :D

packages/framework/src/Component/AddSentryContextSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+            'version' => ShopsysFrameworkBundle::VERSION,
+            'systemTimeZone' => date_default_timezone_get(),
+        ];
```
henzigo, line 45: What do you think about passing the customer uuid if he is logged in?

packages/framework/src/Component/AddSentryContextSubscriber.php
TomasLudvik: I would like to keep it as it is bcs you can have different environments if some1 changes them

project-base/app/src/Component/Domain/DomainAwareSecurityHeadersSetter.php
TomasLudvik: Reworked

packages/framework/src/Component/AddSentryContextSubscriber.php
TomasLudvik: Removed

packages/framework/src/Component/AddSentryContextSubscriber.php
TomasLudvik: I have added userId :-)

upgrade-notes/backend_20241217_144851.md
```diff
@@ -0,0 +1,3 @@
+#### prevent NoDomainSelectedException from being thrown during maintenance page ([#{pullRequestId}](https://github.com/shopsys/shopsys/pull/{pullRequestId}))
```
grossmannmartin: can you fix the upgrade notes?

project-base/storefront/next-env.d.ts
```diff
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/basic-features/typescript for more information.
```
grossmannmartin: this change is necessary?

project-base/app/src/Component/Domain/DomainAwareSecurityHeadersSetter.php
```diff
@@ -7,6 +7,7 @@
 use App\Component\Setting\Setting;
 use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\Domain\DomainAwareSecurityHeadersSetter as BaseDomainAwareSecurityHeadersSetter;
+use Shopsys\FrameworkBundle\Component\Domain\Exception\NoDomainSelectedException;
```
grossmannmartin: is this used?

packages/framework/src/Component/Domain/DomainSubscriber.php
```diff
@@ -39,7 +39,7 @@ public static function getSubscribedEvents(): array
     {
         return [
             // Setting domain by request must be done before loading other services (eg.: routing, localization...)
-            KernelEvents::REQUEST => [['onKernelRequest', 100]],
+            KernelEvents::REQUEST => [['onKernelRequest', 10000000]],
```
grossmannmartin: _nitpick_ 
I find it hard to determine the priority with such many zeros.

what about setting some reasonable values for this and maintenanceModeSubscriber? 10000 and 20000 for instance

packages/framework/src/Component/Collector/ShopsysFrameworkDataCollector.php
```diff
@@ -34,12 +35,17 @@ public function collect(Request $request, Response $response, ?Throwable $except
...
-            'displayTimeZone' => $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($this->domain->getId())->getName(),
         ];
+
+        try {
+            $this->data['currentDomainId'] = $this->domain->getId();
+            $this->data['currentDomainName'] = $this->domain->getName();
+            $this->data['currentDomainLocale'] = $this->domain->getLocale();
+            $this->data['displayTimeZone'] = $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($this->domain->getId())->getName();
+        } catch (NoDomainSelectedException) {
+            return;
```
grossmannmartin: How does it behave when no domain is selected? 
getter methods are strictly typed, so my guess is it will break just a moment later. 

What about setting some default values ('-' for strings, 0 for int maybe)

packages/framework/src/Component/AddSentryContextSubscriber.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $context = [
+            'environment' => class_exists('App\Environment') ? Environment::getEnvironment() : '',
+            'version' => ShopsysFrameworkBundle::VERSION,
+            'systemTimeZone' => date_default_timezone_get(),
+            'userId' => $this->currentCustomerUser->findCurrentCustomerUser()?->getId(),
+        ];
+
+        try {
+            $context['maintenance-page'] = $this->redisClientFacade->contains(MaintenanceModeSubscriber::MAINTENANCE_KEY);
```
grossmannmartin: this adds several requests to the Redis to check if the maintenance page is enabled. 
It would be nice to have a provider that caches the first response and keep the maintenance state in memory

grossmannmartin (changes requested):

project-base/storefront/next-env.d.ts
TomasLudvik: I do not know why, but this is showing up all the time, I will revert it

packages/framework/src/Component/AddSentryContextSubscriber.php
```diff
@@ -0,0 +1,75 @@
...
+use Shopsys\FrameworkBundle\Component\Redis\RedisClientFacade;
+use Shopsys\FrameworkBundle\Model\Customer\User\CurrentCustomerUser;
+use Shopsys\FrameworkBundle\ShopsysFrameworkBundle;
+use Symfony\Component\Console\ConsoleEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Filesystem\Exception\FileNotFoundException;
+use Symfony\Component\HttpKernel\KernelEvents;
+use function Sentry\configureScope;
+
+class AddSentryContextSubscriber implements EventSubscriberInterface
```
henzigo, line 22: Just one more thing... What if you change this subscriber to [Monolog Processor](https://symfony.com/doc/current/logging/processors.html)? Probably Sentry will (with proper configuration) register this processor and you will have that informations in your logs as well as in Sentry

What do you think?

packages/framework/src/Component/AddSentryContextSubscriber.php
TomasLudvik: I have introduced MaintenanceModeFacade for this

packages/framework/src/Component/AddSentryContextSubscriber.php
TomasLudvik: It is interesting idea. I would like to push current changes further so it solves the main problem and we can implement it in the other PR.

packages/framework/src/Component/Maintenance/MaintenanceModeFacade.php
```diff
@@ -0,0 +1,44 @@
...
+    public function disable(): void
+    {
+        $this->redisClientFacade->save(MaintenanceModeSubscriber::MAINTENANCE_KEY, true);
+        $this->isMaintenanceModeEnabled = true;
+    }
+
+    public function enable(): void
+    {
+        $this->redisClientFacade->delete(MaintenanceModeSubscriber::MAINTENANCE_KEY);
+        $this->isMaintenanceModeEnabled = false;
```
grossmannmartin: function `enable` sets `isEnabled` to false - it seems counterintuitive

grossmannmartin (changes requested):

packages/framework/src/Component/Maintenance/MaintenanceModeFacade.php
TomasLudvik: 🤦

packages/framework/src/Component/Maintenance/MaintenanceModeSubscriber.php
```diff
@@ -13,16 +12,14 @@
 
 class MaintenanceModeSubscriber implements EventSubscriberInterface
 {
-    public const MAINTENANCE_KEY = 'maintenance';
-
-    protected ?bool $isMaintenanceMode = null;
+    public const string MAINTENANCE_KEY = 'maintenance';
```
henzigo: This constant should be in MaintenanceModeFacade :)

grossmannmartin (approved):

vitek-rostislav (approved):

Issue #1625, opened 2020-01-27
Missing factories for classes


vitek-rostislav:
### What is happening
There are some classes that are instantiated directly in the code which makes it harder to extend. So far I have encountered these:
- `ProductFilterData`
- `Price`
- `ProductPrice`

I believe there are more of them that deserve their own factories.

### Expected result
The framework is examined for missing factories and the suitable factories are added

pk16011990:
Class `Price` is finale for me. Extended class `price` points rather to a bad design in my opinion. 

Do you have an example of good reason for extending of `Price`?

vitek-rostislav:
Hi @pk16011990, sorry for the late response. Actually, I have no example for extending `Price` :smile: I think I just wanted to show some examples of services that are created directly using `new` keyword. As I do not have any example for extending `Price`, I can remove It from the list. However, at least the other two definitely deserve their factories.

As expected result states, I want the framework to be examined for missing factories because I have found at least two so far.

grossmannmartin:
Hi, thanks for the issue 👍 
I would like to fix this in the upcoming major version, but sadly I can't make any promises.

vitek-rostislav:
just bumped into another class that is missing factory - `ProductParameterValuesLocalizedData` - now I am forced to overwrite whole `ProductParameterValueToProductParameterValuesLocalizedTransformer` as the data object is created directly there

vitek-rostislav:
factory for `ProductFilterData` was added in https://github.com/shopsys/shopsys/pull/2983 :tada:

PR #3724, opened 2025-01-08
extensibility enhancements (factories and final classes)

vitek-rostislav:
There were some classes created directly which made their extension from project-base unnecessarily complicated. On the other hand, there are some classes that are not intended to be extended, so they are declared as final now.

New factories created for:

- `ProductParameterValuesLocalizedData`
- `CartWithModificationsResult`

Classes declared final:

- `Price` + introduced new `PriceInterface`
- `ProductPrice` + introduced new `ProductPriceInterface`
- `OrderTotalPrice` + introduced new `OrderTotalPriceInterface`
- `QuantifiedItemPrice` + introduced new `QuantifiedItemPriceInterface`

Classes instantiated directly where adding a factory would not have any added value IMHO. The reason is that those classes accept a few properties in the constructor and there is no point in creating an empty instance and setting the properties afterwards. Hence, a factory would not ease their extension. Moreover, those objects are mostly created only once in the codebase :pray: 

- `QuantifiedProduct`
- `AddProductResult`
- `AddToCartResult`
- all the kinds of grid data sources (e.g. `QueryBuilderDataSource`) - we might make them public as well, however, we are currently working on a [new CRUD approach](https://github.com/shopsys/shopsys/pull/3629) so there is no point in changing these at the moment.

As a bonus, `ProductSellingPrice` was removed completely and its usages were replaced by `ProductPrice` that is now enriched by a `$pringGroup` property.

#### Fixes issues
#1625 

----
  - https://rv-factories.odin.shopsys.cloud
  - https://cz.rv-factories.odin.shopsys.cloud


packages/framework/src/Model/Product/Pricing/ProductPriceCalculation.php
```diff
@@ -79,18 +83,19 @@ protected function calculateMainVariantPrice(Product $mainVariant, $domainId, Pr
             $variantPrices[] = $this->calculatePrice($variant, $domainId, $pricingGroup);
         }
 
+        /** @var \Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPrice $minVariantPrice */
         $minVariantPrice = $this->getMinimumPriceByPriceWithoutVat($variantPrices);
         $from = $this->arePricesDifferent($variantPrices);
 
-        return new ProductPrice($minVariantPrice, $from);
+        return new ProductPrice($minVariantPrice->getPrice(), $from);
```
TomasLudvik: This makes me very unsure about benefit of these changes. We have made those classes final, so they cannot be changed but in the same time we introduced interface to enable the change. I think this two things are in contradiction to each other.

If you do not agree, then at least ProductPrice should accept PriceInterface instead of Price and then you do not need this.

packages/framework/src/Model/Product/Pricing/ProductPriceCalculationForCustomerUser.php
```diff
@@ -30,7 +30,7 @@ public function __construct(
      * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
      * @return \Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPrice
      */
-    public function calculatePriceForCurrentUser(Product $product)
+    public function calculatePriceForCurrentUser(Product $product): ProductPrice
```
TomasLudvik: Here you use final class return typehint so this method cannot be changed. Should ProductPrice implement ProductPriceInterface?

packages/framework/src/Model/Product/Pricing/ProductSellingPrice.php
```diff
@@ -5,31 +5,32 @@
...
     /**
      * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup $pricingGroup
-     * @param \Shopsys\FrameworkBundle\Model\Pricing\Price $sellingPrice
+     * @param \Shopsys\FrameworkBundle\Model\Product\Pricing\ProductPrice $productPrice
      */
-    public function __construct(protected readonly PricingGroup $pricingGroup, protected readonly Price $sellingPrice)
-    {
+    public function __construct(
+        private readonly PricingGroup $pricingGroup,
+        private readonly ProductPrice $productPrice,
```
TomasLudvik: This makes me wonder if we neeed ProductPrice and ProductSellingPrice, would not be easier to have just one object for both of these use cases?

packages/framework/src/Form/Admin/Product/Price/PricesWithCalculatedSellingPricesType.php
```diff
@@ -33,7 +33,7 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
 
         foreach ($this->pricingGroupFacade->getByDomainId($options['domain_id']) as $pricingGroup) {
             $builder->add((string)$pricingGroup->getId(), MoneyWithCalculatedPriceType::class, [
-                'selling_price' => $sellingPrices !== null ? $sellingPrices[$pricingGroup->getId()]?->getSellingPrice() : null,
+                'selling_price' => $sellingPrices !== null ? $sellingPrices[$pricingGroup->getId()]?->getProductPrice() : null,
```
TomasLudvik: Should be selling_price be renamend to product_price?

TomasLudvik (changes requested):

packages/framework/src/Model/Product/Pricing/ProductSellingPrice.php
vitek-rostislav: that is a great idea, resolved in 3649fd9b8f

packages/framework/src/Form/Admin/Product/Price/PricesWithCalculatedSellingPricesType.php
vitek-rostislav: resolved in 3649fd9b8f

packages/framework/src/Model/Product/Pricing/ProductPriceCalculationForCustomerUser.php
vitek-rostislav: it is now :ok_hand:

packages/framework/src/Model/Product/Pricing/ProductPriceCalculation.php
vitek-rostislav: all final classes now have their interfaces and the interfaces are used in the code instead of them

packages/framework/src/Model/Product/Pricing/ProductPrice.php
```diff
@@ -4,27 +4,27 @@
...
      * @param \Shopsys\FrameworkBundle\Model\Pricing\Price $price
      * @param bool $priceFrom
      */
-    public function __construct(Price $price, $priceFrom)
-    {
-        $this->priceFrom = $priceFrom;
-
-        parent::__construct($price->getPriceWithoutVat(), $price->getPriceWithVat());
+    public function __construct(
+        private readonly Price $price,
```
TomasLudvik: Here should be also used interface, right?

packages/framework/src/Form/Admin/Product/Price/PricesWithCalculatedPricesType.php
```diff
@@ -13,7 +13,7 @@
 use Symfony\Component\Form\FormView;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class PricesWithCalculatedSellingPricesType extends AbstractType
+class PricesWithCalculatedPricesType extends AbstractType
```
TomasLudvik: What are Prices with CalculatedPrices? I thought we are no longer calculating prices right?

packages/framework/src/Model/Product/Pricing/ProductPrice.php
```diff
@@ -4,27 +4,27 @@
 
 namespace Shopsys\FrameworkBundle\Model\Product\Pricing;
 
+use Override;
+use Shopsys\FrameworkBundle\Component\Money\Money;
 use Shopsys\FrameworkBundle\Model\Pricing\Price;
+use Shopsys\FrameworkBundle\Model\Pricing\PriceInterface;
 
-class ProductPrice extends Price
+final class ProductPrice implements PriceInterface
```
TomasLudvik: You should also implement `ProductPriceInterface` right?

TomasLudvik (changes requested):

packages/framework/src/Model/Product/Pricing/ProductPrice.php
vitek-rostislav: yes, it was added in the later commits - 9d69acaa8ebc6c987d0a87fc744295443205a209

packages/framework/src/Form/Admin/Product/Price/PricesWithCalculatedPricesType.php
vitek-rostislav: it is that form type used in the product admin form to render the prices (the prices are calculated from the input price for each pricing group):
image

packages/framework/src/Model/Product/Pricing/ProductPrice.php
vitek-rostislav: yes, it was added in the later commits - https://github.com/shopsys/shopsys/commit/9d69acaa8ebc6c987d0a87fc744295443205a209

TomasLudvik (approved):

packages/framework/src/Form/Admin/Product/Price/PricesWithCalculatedPricesType.php
vitek-rostislav: [8b3ee40](https://github.com/shopsys/shopsys/pull/3724/commits/8b3ee40db51330e4d3c1495dc7534a82686db1d5)

PR #3732, opened 2025-01-13
interfaces cleanup

vitek-rostislav:
There were a lot of interfaces that were unnecessary, so we are simplifying the codebase by removing them

#### Fixes issues
... 

----
  - https://rv-remove-interfaces.odin.shopsys.cloud
  - https://cz.rv-remove-interfaces.odin.shopsys.cloud


packages/framework/src/Component/Grid/Exception/OrderingNotSupportedException.php
```diff
@@ -6,7 +6,7 @@
 
 use Exception;
```
grossmannmartin, line 8: what about those interfaces? they were left intentionaly?

Snímek obrazovky 2025-01-13 v 16 12 31

grossmannmartin (approved):

packages/framework/src/Component/Grid/Exception/OrderingNotSupportedException.php
vitek-rostislav: - `MailException` was forgotten, removed in cd48270491
-  `FormException` removed in ab00a6b96a
- `PerformanceException` removed in 52baacd569

Please let's discuss once more the following - they are used in catch blocks, and some of them might be replaced by regular common exception classes (but some of them probably not - e.g. there are PromoCode exceptions that extend NotFoundException)...

- `FileUploadException` - removed in c60874fed2
- `ImageConfigException` - removed in b9eda6f7a9
- `MoneyException` - removed in 97ea283f9d
- `UploadedFileConfigException` - removed in 9f2a00c3dc
- `PromoCodeException` - removed in 71eb9aac16
- `VariantException` - removed in fe9af7d5f6

packages/framework/src/Component/UploadedFile/Config/Exception/InvalidUploadedFileConfigException.php
```diff
@@ -4,8 +4,8 @@
 
 namespace Shopsys\FrameworkBundle\Component\UploadedFile\Config\Exception;
 
-use Throwable;
+use Exception;
 
-interface UploadedFileConfigException extends Throwable
+abstract class InvalidUploadedFileConfigException extends Exception
```
grossmannmartin: I'm not sure if we need to make any exceptions abstract.
Reason for it - if I need to create a new exception and place it properly into the exception hierarchy, I'm now put into thinking whether to make the parent exception abstract or not. 
For example, I have EntitiyIdentifierException and now I need to create `EntityIdentifierIsCompoundException extends EntityIdentifierException`. Should I rewrite everything and make `EntitiyIdentifierException` abstract?

Also, Exception is a class and now I'm extending a valid class into abstract class feels weird in a context of excetptions.

vitek-rostislav: ok, fair point, makes sense, I will remove the abstract declarations :+1:

vitek-rostislav: removed, rebased, force pushed

pk16011990 (approved):

pk16011990 (approved):

PR #3784, opened 2025-02-08
manual complaint creation (without order)

vitek-rostislav:
It is now possible to create complaints "manually", i.e. without choosing an existing order item to complain about.

Instead, a customer can manually input the order (or any document) number, complaint item name, and (optionally) a catalog number. Additionally, if the catalog number is found in the database, the corresponding product is linked to the complaint.

#### Fixes issues
... 

----
  - https://rv-complaints-without-order.odin.shopsys.cloud
  - https://cz.rv-complaints-without-order.odin.shopsys.cloud


packages/frontend-api/src/Model/Complaint/ComplaintApiFacade.php
```diff
@@ -99,7 +99,12 @@ public function createFromComplaintInputArgument(Argument $argument): Complaint
     {
         $input = $argument['input'];
 
-        $order = $this->orderApiFacade->getByUuid($input['orderUuid']);
+        $orderUuid = $input['orderUuid'];
+        $order = null;
+
+        if ($orderUuid !== null) {
+            $order = $this->orderApiFacade->getByUuid($orderUuid);
+        }
```
TomasLudvik, line 107: Here should be some validation, if the order is owned by the customer

TomasLudvik (changes requested):

TomasLudvik (approved):

project-base/storefront/components/Blocks/Popup/CreateComplaintPopup.tsx
```diff
@@ -42,7 +42,12 @@ export const CreateComplaintPopup: FC<CreateComplaintPopupProps> = ({ orderUuid,
     const user = useCurrentCustomerData();
 
     const defaultDeliveryAddressChecked = user?.deliveryAddresses[0]?.uuid || '';
-    const [formProviderMethods] = useComplaintForm(defaultDeliveryAddressChecked, user?.email || '');
+    const creationWithoutOrder = orderUuid === null;
```
JanMolcik: Boolean values should start with `is` for example `isCreationWithoutOrder` or `isWithoutOrder`.

project-base/storefront/components/Pages/Customer/ComplaintDetail/ComplaintDetailComplaintItem.tsx
```diff
@@ -39,38 +36,40 @@ export const ComplaintDetailComplaintItem: FC<ComplaintDetailComplaintItemProps>
...
-                                    query: { orderNumber: complaintOrder?.number },
-                                }}
-                            >
-                                {complaintOrder?.number}
-                            </ExtendedNextLink>
-                        ) : (
-                            complaintOrder?.number
-                        )}
-                    </span>
+                    {complaintOrder !== undefined && (
```
JanMolcik: Here I believe you can simply use just `{complaintOrder && ...`. Note that sometimes it might happen, when the value is `0` that it is not removed and the `0` is rendered instead. In that case, you can use "double-bang" `{!!complaintOrder && ...` that converts any falsy or truthy values to pure `false` or `true` booleans. :-)

project-base/storefront/components/Pages/Customer/Complaints/complaintFormMeta.ts
```diff
@@ -21,6 +21,7 @@ import * as Yup from 'yup';
 export const useComplaintForm = (
     defaultDeliveryAddressChecked: string,
     defaultEmail: string,
+    creationWithoutOrder: boolean,
```
JanMolcik: Same as above, please prepend `is` when working with booleans.

JanMolcik (changes requested):

project-base/storefront/components/Pages/Customer/ComplaintDetail/ComplaintDetailComplaintItem.tsx
vitek-rostislav: ah. sure. Here, the value can be null/undefined/api type, so never "0" value so fixed to the simple `{complaintOrder && ...`

JanMolcik (approved):

JanMolcik (approved):

TomasLudvik (approved):

JanMolcik (approved):

PR #3795, opened 2025-02-13
added smoth scroll to error fields in complaint form popup

chlebektomas:
Smooth scrolling was added to the error field in the complaint form popup.

----
  - https://tc-ssp-3174-form-on-error-scroll.odin.shopsys.cloud
  - https://cz.tc-ssp-3174-form-on-error-scroll.odin.shopsys.cloud


JanMolcik (changes requested):
I would like to fix this feature generally for any form and any error. I have explored that a bit before but the solutions is not trivial as we are using controlled form elements that do not contain ref on the global hook-form level.

JanMolcik (approved):

PR #3801, opened 2025-02-14
removed no longer necessary command for migrating images to new proxy structure

TomasLudvik:
In version 14.0 we have changed image folder structure and introduced migration command to easily switch to this new structure, after three versions we are removing this command as it was one time action and is no longer necessary.

#### Fixes issues
... 

----
  - https://tl-remove-image-migrate-for-proxy.odin.shopsys.cloud
  - https://cz.tl-remove-image-migrate-for-proxy.odin.shopsys.cloud


packages/framework/src/Migrations/Version20250214133657.php
```diff
@@ -7,16 +7,14 @@
...
 {
     /**
      * @param \Doctrine\DBAL\Schema\Schema $schema
      */
     public function up(Schema $schema): void
     {
-        $this->sql('INSERT INTO setting_values (name, domain_id, value, type) VALUES
-            (\'imageStructureMigratedForProxy\', 0, \'false\', \'boolean\')
-        ');
+        $this->sql('DELETE FROM setting_values WHERE name = \'imageStructureMigratedForProxy\'');
```
vitek-rostislav, line 17: I would also remove the entry for `Version20231020173331` from the migrations database table (otherwise, there will be a warning about a missing migration)

```
DELETE FROM "migrations"
WHERE "version" = 'Shopsys\FrameworkBundle\Migrations\Version20231020173331';
```

upgrade-notes/backend_20250214_123809.md
```diff
@@ -0,0 +1,6 @@
+#### remove no longer necessary image migration for proxy ([#3801](https://github.com/shopsys/shopsys/pull/3801))
+
+- if you are upgrading from Shopsys Platform 14.0 or newer, you do not need to do any other steps
+- if you are upgrading from Shopsys Platform 13.0 or older, you need to check this PR and do the migration of images folder manually
+    - note: in version 14.0 we have changed image folder structure and introduced migration command to easily switch to this new structure, after three versions we are removing this command as it was one time action and is no longer necessary
```
vitek-rostislav, line 9: these upgrade notes will be for the `17.0` version, there is no way anyone upgrading from any other version than `16.0` so I am not sure if these notes make any sense :thinking: 

The only thing I would put here would be the removal of `MigrateImagesCommand`, `IMAGE_STRUCTURE_MIGRATED_FOR_PROXY` setting, and `migrate-images-for-proxy` phing target as they are no longer necessary and I would add an instruction to remove their `Version20231020173331` entry from the migrations lock file :slightly_smiling_face:

vitek-rostislav (changes requested):
Sorry for being a pain in the ass :innocent:

upgrade-notes/backend_20250214_123809.md
TomasLudvik: I have added those mentions and I have also keeped the text as everyone can continue without need to do anything, because they have been upgrading from version >= 14.0, but you never know as there might be someone dedicated to upgrade theirs project all the way from version 7 :D I like to keep it safe.

vitek-rostislav (approved):

PR #3813, opened 2025-02-19
Administration overhaul

### LINKED ISSUE #406: Information in validation/readme.txt should be moved into docs folder and transferred into md file
### What is happening
In `packages/framework/src/Resources/scripts/common/validation/readme.txt` (and duplicated in `packages/framework/src/Resources/scripts/admin/validation/readme.txt`) there are some useful information about the js validation but the file is lost among source code files.

Content of the files:
```
Client-side validation requires:
* subform must have HTML container with attribute id="{form_name}_{subform_name}" eg.: customer_deliveryAddressData
* collection of inputs or subforms must have HTML container with attribute id="{form_name}_{collection_name}" eg.: product_parameters
* when dynamically adding new collection item, this code must be added:
	Shopsys.validation.addNewItemToCollection('#{form_name}_{collection_name}', index);
* when dynamically removing new collection item, this code must be added:
  Shopsys.validation.removeItemFromCollection('#{form_name}_{collection_name}', index);
```

### Expected result
The information about js validation are placed in md file in docs.

### LINKED ISSUE #413: remove option macro from forms, use separate form types instead
### What is happening

At the moment, all form types have a custom option named `macro` (see FormRenderingConfigurationExtension). This feature is completely undocumented and behaves very unexpectedly - it is supported in a few specific form types and it completely switches the rendering (using a Twig macro, which is the reason behind the option's name). Also, when not set to a specific supported values, it will prevent the form input from rendering without any warning or exception.

See `theme.html.twig` for the usages:
https://github.com/shopsys/shopsys/blob/3317ff259c2230024463ca4cd7dff09eddf6374a/packages/framework/src/Resources/views/Admin/Form/theme.html.twig#L93-L101
https://github.com/shopsys/shopsys/blob/3317ff259c2230024463ca4cd7dff09eddf6374a/packages/framework/src/Resources/views/Admin/Form/theme.html.twig#L209-L217
https://github.com/shopsys/shopsys/blob/3317ff259c2230024463ca4cd7dff09eddf6374a/packages/framework/src/Resources/views/Admin/Form/theme.html.twig#L287-L292

### Expected result

The `macro` option is removed. These inputs are implemented as individual FormType implementations, similarly to `PriceTableType`, `ProductsType`, etc.

New form types are documented.

### LINKED ISSUE #2541: Select2: disabled elements are hidden
## Describe the bug
If you wanna disable some elements in choice type but you wanna still show it, you can't. Actually, if you set element as disabled it will set as hidden

### To Reproduce
1. Go to product detail in administration
2. Open file `ProductFormType` in your IDE
3. into brand element add these lines
```diff
            ->add('brand', ChoiceType::class, [
                'required' => false,
                'choices' => $this->brandFacade->getAll(),
                'choice_label' => 'name',
                'choice_value' => 'id',
+               'choice_attr' => function (Brand $brand) {
+                   return [
+                       'disabled' => 'disabled',
+                   ];
+               },
                'placeholder' => t('-- Choose brand --'),
                'label' => t('Brand'),
            ])
```
5. Refresh page with product detail in administration
6. Open brand choice list and see nothing

### Expected behavior
Disabled elements are disabled, not hidden

### Environment (please complete the following information)
 - OS: Windows 11 with docker
 - Browser: Brave (chromium)
 - Version: actual master (in unnamed project doesn't work too, he is in version 9.1.2)
 - Device: Desktop computer

### LINKED ISSUE #2623: Possibility to add an icon to any form type
In administration, there are a few form types that allow adding an icon, icon title, or symbol after input. This is ensured by `appendix_block` in `theme.html.twig`. I would like to have the possibility in all the form types (e.g. it would be very handy in `DisplayOnlyType`.) I have to implement it myself on my project now - extending `displayOnlyFields.html.twig` template, add `{{- block('appendix_block') -}}` there, and then define the icon in my form type extension by adding
```
'attr' => [
                    'readonly' => 'readonly',
                    'class' => '',
                    'icon' => true,
                    'iconTitle' => 'whatever',
                ]
```  
(I have to re-define the default values of the first two attributes - readonly and class - to make it work).

grossmannmartin:

  This pull request introduces a complete redesign and modernization of the Shopsys Platform administration theme. The legacy custom-built theme has been replaced with the modern, responsive [Tabler UI](https://tabler.io/) framework.

  **Key motivations:**
  - **Modernization & Maintainability**: Move from legacy frontend stack (custom CSS, LESS, older JS) to modern best practices (SCSS, modern JS, documented framework)
  - **Improved UX**: Cleaner, more intuitive, fully responsive interface across all devices
  - **Faster Development**: Building on well-documented Tabler framework enables quicker feature development
  - **Significant Cleanup**: Removed over 12,000 lines of legacy code, reducing codebase complexity

  **Major technical changes:**
  - **UI Framework**: Complete migration to Tabler UI with systematic template updates across all admin sections
  - **Assets**: LESS → SCSS conversion, updated webpack config, modernized dependencies
  - **JavaScript**: Replaced legacy jQuery plugins with modern libraries (tom-select, coloris), removed obsolete code
  - **Forms**: New form rendering system aligned with Tabler styling
  - **Components**: Modern modals, tooltips, responsive grids, improved accessibility

  **⚠️ Breaking Changes**: This is a major release with significant BC breaks for any project with admin customizations. Custom admin themes, CSS overrides, and JavaScript customizations will require complete rewrite to target the new Tabler-based DOM structure and CSS classes.


<!-- If you have introduced any BC breaks (https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/), please add UPGRADE notes using `php phing upgrade-generate` -->

<!-- If you have introduced a new feature, please update docs -->

#### Fixes issues

fixes #413, fixes #2623, closes #406, closes #2541, closes #2542

<!-- Replace -->
----
  - https://mg-admin-theme.odin.shopsys.cloud
  - https://cz.mg-admin-theme.odin.shopsys.cloud
  - https://mg-admin-theme.odin.shopsys.cloud/sk
<!-- Replace -->


packages/administration/templates/form/theme.html.twig
```diff
@@ -0,0 +1,68 @@
+{% extends 'bootstrap_5_horizontal_layout.html.twig' %}
+
+{%- block textarea_widget -%}
+    {% set attr = attr|merge({'data-bs-toggle': 'autosize'}) %}
+    {{- parent() -}}
+{%- endblock -%}
+
+{%- block form_widget_simple -%}
+    {% if symbolAfterInput is defined or symbolBeforeInput is defined or rawSymbolAfterInput is defined or rawSymbolBeforeInput is defined %}
```
vitek-rostislav, line 9: Those symbols before/after inputs deserve mention in docs IMHO

vitek-rostislav: btw it looks like https://github.com/shopsys/shopsys/issues/2623 is solved by your PR, right? Hooray :)

packages/administration/templates/form/theme.html.twig
```diff
@@ -0,0 +1,73 @@
...
+{%- block form_label -%}
+    {% if label is not same as(false) %}
+        {{- parent() -}}
+    {% endif %}
+{%- endblock -%}
+
+{%- block form_group_class -%}
+    {# if label is not rendered, render form group with full width #}
+    {%- if label is same as(false) -%}
+        col
```
vitek-rostislav, line 57: is this something that should stay here or is it a forgotten placeholder?

packages/administration/templates/bundles/TwigBundle/Exception/error500.html.twig
```diff
@@ -1,26 +1,19 @@
...
-</html>
+{% extends '@ShopsysAdministration/layout/layout_error.html.twig' %}
+
+{% block error_content %}
+    <div class="empty-header">500</div>
+
+    <p class="empty-title">{{ 'Oops! Error occurred.'|trans }}</p>
+
+    <p class="empty-subtitle text-secondary">
+        {{ 'Uh-oh, uh-oh. And specially <strong>"%status_code% %status_text%"</strong>, as if it was on purpose!'|transHtml({ '%status_code%': status_code, '%status_text%': status_text }) }}
```
vitek-rostislav, line 9: Love the fact we have these messages back :)

packages/administration/templates/layout/layout_with_panel.html.twig
```diff
@@ -0,0 +1,87 @@
...
+                                    <div class="dropdown-menu">
+                                        <div class="btn-list m-3 justify-content-end">
+                                            {{ btnContent|raw }}
+                                        </div>
+                                    </div>
+                                </div>
+
+                                <div class="btn-list d-none d-md-flex">
+                                    {{ btnContent|raw }}
+                                </div>
```
vitek-rostislav, line 75: btnContent is here rendered twice because of the responsive, right?

packages/administration/templates/form/type/ColorPickerType.html.twig
```diff
@@ -0,0 +1,4 @@
+{% block color_picker_widget %}
+    {% set attr = attr|merge({'data-coloris': ''}) %}
```
vitek-rostislav: just a idea - is it necessary to define the data attribute via the custom widget template? Would not it be easier to define that directly in ColorPickerType? (`$resolver->setDefaults(['attr' => ['data-coloris' => '']]);` in `configureOptions`). However, when I type it, I am not sure which approach is better. Using the template would be probably more common approach, right? Let's discuss it, please

docs/introduction/using-form-types.md
```diff
@@ -74,10 +74,6 @@ there is `DisplayOnlyType` which does not map property onto `entity` and let you
 
 Displays custom URL based on routing system.
 
-### [LocalizedFullWidthType]({{github.link}}/packages/framework/src/Form/LocalizedFullWidthType.php)
-
-Displays localized field in vertical order of full-width label and inputs.
-
```
vitek-rostislav, line 80: this is removed because everything is now full width by default?

packages/framework/src/Form/DatePickerType.php
```diff
@@ -31,6 +31,9 @@ public function configureOptions(OptionsResolver $resolver): void
             'widget' => 'single_text',
             'format' => static::FORMAT_PHP,
             'html5' => false,
+            'attr' => [
+                'data-js-datepicker' => 'true',
+            ],
```
vitek-rostislav, line 36: ah, in addition to my previous commit, we have inconsistency here :smile:

packages/administration/templates/form/theme.html.twig
vitek-rostislav: ah, I see it should return a css class, so "col" is ok, right?

packages/framework/assets/js/admin/components/SortableValues.js
```diff
@@ -46,17 +46,33 @@ export default class SortableValues {
...
     }
 
     enableOptionOfItem($item) {
         const $container = $item.closest('.js-sortable-values-container');
         const $input = $item.find('.js-sortable-values-item-input');
         const $option = $container.find(`.js-sortable-values-input [value="${$input.val()}"]`);
 
-        $option.prop('disabled', false);
+        const $select = $option.closest('.js-sortable-values-input');
+        const tomSelectInstance = $select[0].tomselect;
```
vitek-rostislav, line 62: disableOption and enableOptionOfItem share basically the same code (finding the tomSelect instance and updating it with disabled value), maybe it is worth extracting into a function?

vitek-rostislav: but it is rather nitpick

packages/framework/src/Form/Admin/Store/StoreFormType.php
```diff
@@ -164,6 +164,7 @@ private function createUserInformationGroup(FormBuilderInterface $builder): Form
             ])
             ->add('openingHours', CollectionType::class, [
                 'label' => t('Opening hours'),
+                'help' => t('Enter in the store local time'),
                 'entry_type' => OpeningHoursRangeCollectionFormType::class,
```
vitek-rostislav, line 168: love the fact we can use this :partying_face:

packages/framework/src/Controller/Admin/AdministratorController.php
```diff
@@ -78,16 +78,16 @@ public function listAction(): Response
...
-        $grid->setActionColumnClassAttribute('table-col table-col-10');
         $grid->addEditActionColumn('admin_administrator_edit', ['id' => 'a.id']);
         $grid->addDeleteActionColumn('admin_administrator_delete', ['id' => 'a.id'])
             ?->setConfirmMessage(t('Do you really want to remove this administrator?'));
 
-        $grid->setTheme('@ShopsysFramework/Admin/Content/Administrator/listGrid.html.twig');
+        $grid->setTheme('@ShopsysAdministration/content/administrator/listGrid.html.twig');
 
-        return $this->render('@ShopsysFramework/Admin/Content/Administrator/list.html.twig', [
+        return $this->render('@ShopsysAdministration/content/administrator/list.html.twig', [
```
vitek-rostislav, line 98: I suppose for now, we can live with the fact that the framework controller should actually not know about the administration package, right?

packages/administration/templates/content/advert/list.html.twig
```diff
@@ -1,17 +1,19 @@
-{% extends '@ShopsysFramework/Admin/Layout/layoutWithPanel.html.twig' %}
+{% extends '@ShopsysAdministration/layout/layout_with_panel.html.twig' %}
 
 {% block title %}- {{ 'Advertising system'|trans }}{% endblock %}
+
+{% block pre_title %}{{ 'Overview'|trans }}{% endblock %}
 {% block h1 %}{{ 'Advertising system'|trans }}{% endblock %}
```
vitek-rostislav, line 6: I am wondering, it is also rather a nitpick, however, maybe we could get rid of 43 occurrences of `{% block pre_title %}{{ 'Overview'|trans }}{% endblock %}` in the templates? Maybe we could create a common ancestor for the list templates? It could also remove the simplify the main content rendering that duplicates `gridView.render()` calls in a lot of the templates.  Maybe this idea conflicts with CRUD controllers, but we can at least talk about it.

packages/framework/src/Controller/Admin/ArticleController.php
```diff
@@ -232,9 +232,10 @@ public function saveOrderingAction(Request $request): JsonResponse
 
     /**
      * @param string $articlePlacement
+     * @param string $articlePlacementTitle
      * @return \Shopsys\FrameworkBundle\Component\Grid\Grid
      */
-    protected function getGrid(string $articlePlacement): Grid
+    protected function getGrid(string $articlePlacement, string $articlePlacementTitle = ''): Grid
```
vitek-rostislav: the default value is not necessary, or is it because of BC?

packages/administration/templates/content/complaint/advancedSearch/advancedSearch.html.twig
```diff
@@ -0,0 +1,25 @@
+{{ form_start(advancedSearchForm) }}
+    <div id="js-advanced-search-rules-box">
```
vitek-rostislav, line 2: long time ago, I had an idea of refactoring the advanced search templates to avoid DRY, I am putting the note here so that we can discuss it, or so that I do not forget to create an issue :)

packages/administration/templates/content/complaint/listGrid.html.twig
```diff
@@ -9,12 +9,12 @@
...
     {% else %}
         {{ row.cmp.manualDocumentNumber }}
-        {{ info_icon('Complaint was created manually without link to a particular order'|trans, tooltipPlacement: 'right') }}
+        {{ info_icon('Complaint was created manually without link to a particular order'|trans) }}
     {% endif %}
 {% endblock %}
 
 {% block grid_value_cell_id_domain_id %}
-    {{ domainIcon(value, 'small') }}
+    {{ domain_icon(value) }}
```
vitek-rostislav: this could be placed in the main grid theme, right?

packages/administration/templates/content/categorySeo/readyCombination.html.twig
```diff
@@ -0,0 +1,47 @@
...
+{% block pre_title %}{{ 'Extended SEO category'|trans }}{% endblock %}
+{% block h1 %}{{ 'Edit the combination with SEO values'|trans }}{% endblock %}
+
+{% block main_content %}
+    <div class="card mb-3">
+        <div class="card-body">
+            <h3 class="card-title">{{ 'Filter summary' }}</h3>
+            <div class="datagrid">
+                {{ _self.infoLine('Domain'|trans, selectedCategorySeoMixCombinationDomainConfig.name) }}
+
```
vitek-rostislav, line 16: no need to import self as _self?

packages/administration/templates/form/type/MultidomainType.html.twig
```diff
@@ -0,0 +1,38 @@
+{% block multidomain_row %}
+    {%- set row_class = (row_attr.class|default('') ~ ' mb-3')|trim -%}
+    <div {% with {attr: row_attr|merge({class: row_class})} %}{{ block('attributes') }}{% endwith %}>
+        {{ form_label(form) }}
+
+        {{ form_help(form) }}
+
+        {% if layout is same as constant('\\Shopsys\\FrameworkBundle\\Form\\Locale\\LocalizedType::LAYOUT_INLINE') %}
+            <fieldset>
```
vitek-rostislav, line 9: wring constant IMHO, should be the one from MultidomainType class, right?

vitek-rostislav: ah, ok, it is rewritten later

packages/administration/templates/content/blog/article/list.html.twig
```diff
@@ -0,0 +1,28 @@
...
+    <a href="{{ url('admin_blogarticle_new') }}" class="btn btn-primary">
+        {{ ux_icon('plus') }}
+        {{ 'Create new blog article'|trans }}
+    </a>
+{% endblock %}
+
+{% block main_content %}
+    {{ form_start(quickSearchForm) }}
+    {% include '@ShopsysAdministration/partial/_quick_search_form.html.twig' with {
+        form_input: quickSearchForm.text,
```
vitek-rostislav, line 18: I love the idea of creating th quick search form template to avoid DRY. I still see a few "direct" usages of the quick search form that do not use the template, is that intended or could it be yet rewritten as well? The "few" is probably just the following two places - `packages/administration/templates/content/complaint/quickSearchFormContent.html.twig` and `packages/administration/templates/content/product/quickSearchFormContent.html.twig`

packages/framework/src/Form/Admin/Pricing/Currency/CurrencySettingsFormType.php
```diff
@@ -53,8 +53,8 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                     ],
                 ],
             ])
-            ->add('actionBar', ActionBarType::class, [
-                'save_label' => t('Save changes'),
+            ->add('save', SubmitType::class, [
+                'label' => t('Save default currencies'),
```
vitek-rostislav, line 57: nice catch to change this :clap:

packages/framework/src/Model/Product/Flag/FlagGridFactory.php
```diff
@@ -40,7 +40,21 @@ public function create(?string $editRole): Grid
...
+            'f.id',
+            function ($row) {
+                $color = strtr($row['f']['rgbColor'], ['#' => '']);
+                $r = hexdec(substr($color, 0, 2));
+                $g = hexdec(substr($color, 2, 2));
+                $b = hexdec(substr($color, 4, 2));
+
+                $brightness = ($r * 0.299 + $g * 0.587 + $b * 0.114);
+                $row['textColor'] = $brightness > 128 ? '#25283d' : '#ffffff';
+
```
vitek-rostislav: setting the text color is a brilliant idea. However, where do the numbers come from? Is there an existing formula for the brightness calculation from rgb? Could we maybe extract it into a function so it is somehow better understandable? Maybe a comment above the function - something like "to ensure text is well readable with the flag background color..." + the link to the formula (if exists) would be great. Btw, how is this solved on the storefront? :smile:

vitek-rostislav: ah, I see `getYIQContrastTextColor` in `colors.ts` :grin:

packages/administration/templates/content/flag/deleteForbidden.html.twig
```diff
@@ -0,0 +1,23 @@
...
+
+{% if hasPromoCodeDependency %}
+    <p>{{ 'Promo code'|trans }}</p>
+{% endif %}
+
+{% if hasSeoMixDependency %}
+    <p>{{ 'Extended SEO categories'|trans }}</p>
+{% endif %}
+
+{# buttons should be consistent withass the result of Shopsys.window JS function in Resources/scripts/admin/window.js #}
```
vitek-rostislav: "withass" is a typo, right? :smile: Moreover, `Resources/scripts/admin/window.js` does not exist. Moremoreover, maybe the comment is not neccessary at all? Or should be added to the common delete confirm templates as well? Lets at least talk about it, please

packages/administration/templates/content/flag/listGrid.html.twig
```diff
@@ -15,7 +15,7 @@
 
 {% block grid_value_cell_id_rgbColor %}
     {% if value is not empty %}
-        <span class="in-pill" style="background-color: {{ value }};">{{ value }}</span>
+        <span class="badge badge-pill" style="background-color: {{ value }}; color: {{ row.textColor}}; width: 7em">{{ value }}</span>
     {% endif %}
```
vitek-rostislav, line 19: one weird thing - I am not able to mark and copy the rgb text in the grid

packages/administration/templates/content/default/cronDetail.html.twig
```diff
@@ -0,0 +1,20 @@
+{% extends '@ShopsysAdministration/layout/layout_with_panel.html.twig' %}
+
+{% block title %}- {{ 'Cron detail'|trans }} - {{ cronName }}{% endblock %}
+{% block h1 %}{{ 'Cron detail'|trans }} - {{ cronName }}{% endblock %}
```
vitek-rostislav: really a nitpick, but after I saw a lot of usages of the pre-title in this PR, I would say this would look better:
```
{% block pre_title %}{{ 'Cron detail'|trans }}{% endblock %}
{% block h1 %}{{ cronName }}{% endblock %}
```

packages/administration/templates/content/payment/listGrid.html.twig
```diff
@@ -0,0 +1,17 @@
...
+
+{% block grid_value_cell_id_name %}
+    <a href="{{ url('admin_payment_edit', { id: row.p.id }) }}">{{ value|nameWithFallbackOnEmpty }}</a>
+{% endblock %}
+
+{% block grid_value_cell_id_price %}
+    <div class="badges-list">
+        <span class="badge">
+            {{ value|priceWithCurrencyByDomainId(row.domainId) }}
+        </span>
```
vitek-rostislav: also here, it is not possible to highlight and copy the text (price) in the grid now

vitek-rostislav: looks like it is a general "tabler badge" problem. However, the bootstrap badges themselves do not seem to have such an issue

packages/administration/templates/content/mail/detail.html.twig
```diff
@@ -0,0 +1,19 @@
+{% extends '@ShopsysAdministration/layout/layout_with_panel.html.twig' %}
+{% import '@ShopsysAdministration/content/mail/variablesList.html.twig' as applicableVariablesMacro %}
+
+{% block title %}- {{ 'Email template'|trans }}{% endblock %}
+
```
vitek-rostislav, line 5: it seems like all our "detail" templates define title and h1 blocks in vain, as they are always extended in both "new" and "edit" templates. Just saying :smile: Probably it is the same on 17.0 already :shrug:

vitek-rostislav: When I take the idea further, we could probably get rid of most of the entity-specific detail templates and replace them with one entityDetail.html.twig that would look like this:
```
{% extends '@ShopsysAdministration/layout/layout_with_panel.html.twig' %}
{% block main_content %}
    {{ form(form) }}
{% endblock %}
```

But once again, hey, this will be probably resolved by CRUD when it is completed so why bother, right?

packages/framework/src/Form/Admin/Superadmin/MailWhitelistFormType.php
```diff
@@ -46,7 +46,7 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
         $builder
             ->add('mailWhitelistEnabled', YesNoType::class, [
                 'label' => t('Enable whitelist'),
-                'required' => true,
+                'required' => false,
             ])
```
vitek-rostislav, line 49: from my point of view, YesNo should always be required, as you have to select one of the values, there is no option to select none of them (I understand it as submitting the form with NULL value for the given field - which is not allowed here.) However, it is probably more of a philosophical question here, lets discuss it please. I also see that required true renders asterisks for the radio labels as well which is weird from UX point of view, so :shrug:

packages/administration/templates/content/notificationBar/edit.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends '@ShopsysFramework/Admin/Content/NotificationBar/detail.html.twig' %}
+{% extends '@ShopsysAdministration/content/notificationBar/detail.html.twig' %}
 
 {% block title %}- {{ 'Editing notification bar'|trans }}{% endblock %}
 {% block h1 %}{{ 'Editing notification bar'|trans }}{% endblock %}
```
vitek-rostislav, line 4: no pre_title here? :innocent: Moreover, one thing I realized regarding the "pre_title" - in context of setting title and h1, should not the "pre_title" block be rather called "pre_h1". Moremoreover (sorry for the repeated joke), I see h1 block is actually rendered as h2 tag in the layout template, uff :smile:

packages/administration/templates/content/uploadedFile/list.html.twig
```diff
@@ -0,0 +1,27 @@
...
+            form_input: quickSearchForm.text,
+            form_submit: quickSearchForm.submit,
+            info_message: 'Search in uploaded files by filename, extension or translated name. It is possible to use wildcards * and ?.'|trans,
+            placeholder: 'Search by filename, extension or translated name...'|trans
+        } %}
+    {{ form_end(quickSearchForm) }}
+
+    {{ gridView.render() }}
+    {{ gridView.renderHiddenInputs('page') }}
+{% endblock %}
```
vitek-rostislav, line 27: I was wondering what is this good for and I do not think we use it properly :smile: lets talk about it, please

packages/administration/templates/content/order/listGrid.html.twig
```diff
@@ -18,7 +22,7 @@
 {% endblock %}
 
 {% block grid_value_cell_id_domain_id %}
-    {{ domainIcon(value, 'small') }}
+    {{ domain_icon(value) }}
 {% endblock %}
```
vitek-rostislav: this could be also placed in the basic grid theme, however, it is not that frequent as I expected (3 duplicate usages at the moment). And another thing that will be solved by CRUD...

packages/framework/assets/js/admin/components/OrderTransportAndPaymentPrefiller.js
```diff
@@ -24,24 +24,25 @@ export default class OrderTransportAndPaymentPrefiller {
 
     onOrderTransportChange() {
         const selectedTransportId = this.$transportSelect.val();
-        $('#order_form_orderItems_orderTransport_priceWithVat').val(
+
+        $('#order_form_orderItems_orderTransport_unitPriceWithVat').val(
             this.transportPricesWithVatByTransportId[selectedTransportId].amount,
```
vitek-rostislav, line 29: oh yes, so this was the cause of an issue I noticed and noted a while ago and forgot to create an actual issue for that :smile:

packages/framework/src/Controller/Admin/ParameterGroupController.php
```diff
@@ -171,15 +171,14 @@ protected function getGrid(): Grid
         $grid->addColumn('name', 'pgt.name', t('Name'));
         $grid->setDefaultOrder('pg.position');
 
-        $grid->setActionColumnClassAttribute('table-col table-col-10');
         $grid->addEditActionColumn('admin_parametergroup_edit', ['id' => 'pg.id']);
```
vitek-rostislav, line 182: I realized there is not any usage of `setActionColumnClassAttribute` method anymore. Maybe we could remove it, however, I do not mind keeping it for BC :shrug:

packages/administration/templates/content/priceList/listGrid.html.twig
```diff
@@ -10,17 +10,17 @@
...
-        '1': { 'class': 'warning', 'label': 'Future'|trans },
-        '-1': { 'class': 'error', 'label': 'Expired'|trans }
+        '0': { 'class': 'green', 'label': 'Active'|trans },
+        '1': { 'class': 'purple', 'label': 'Future'|trans },
+        '-1': { 'class': 'red', 'label': 'Expired'|trans }
     } %}
 
-    <span class="in-message--{{ statusMap[value].class }} in-message--small">
+    <span class="status status-{{ statusMap[value].class }}">
         {{ statusMap[value].label }}
```
vitek-rostislav, line 20: IMHO this renders better than the "badge badge-pill" combination that is used elsewhere (e.g. at the flags list). I would consider using "status status-XY" everywhere, but now when I type it, I realize there is probably the colors range limitation, right? :/

packages/framework/assets/js/admin/components/PriceListProductPicker.js
```diff
@@ -1,9 +1,9 @@
-import 'magnific-popup';
 import Translator from 'bazinga-translator';
 import { formatPrice } from '../../common/utils/priceFormatter';
 import Register from '../../common/utils/Register';
 import Window from '../utils/Window';
 import FormChangeInfo from './FormChangeInfo';
+import ModalWindow from '@shopsys/administration/src/js/utils/modalWindow';
```
vitek-rostislav, line 7: this is also a reverted dependency, are we ok with that? I mean, I think it is just weird, it will work within the whole ecosystem, right?

packages/administration/templates/content/promoCode/listGrid.html.twig
```diff
@@ -1,15 +1,13 @@
...
 {% block grid_value_cell_id_active %}
-    {% if row['pc']['isActive'] is same as(true) %}
-        <span style="color: green">{{ 'Yes'|trans }}</span>
-    {% else %}
-        <span style="color: red">{{ 'No'|trans }}</span>
-    {% endif %}
+    {% set value = row['pc']['isActive'] %}
+
+    {{ block('grid_value_cell_type_boolean') }}
 {% endblock %}
```
vitek-rostislav: it should work the same out of the box when you remove `grid_value_cell_id_active` blcok completely

packages/framework/src/Form/Admin/Seo/SeoRobotsSettingFormType.php
```diff
@@ -20,12 +21,13 @@ final class SeoRobotsSettingFormType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options): void
     {
         $builder
+            ->add('messageIsForced', MessageType::class, [
+                'message_level' => MessageType::MESSAGE_LEVEL_WARNING,
```
vitek-rostislav, line 25: even though it is not important, the name might be confusing (it is probably copied from `MailWhitelistFormType`)

vitek-rostislav: However, I love the idea of defining the messages in form and keeping the templates clean then. Also any future change of styling of the messages would be much easier now :clap:

packages/framework/src/Resources/views/Admin/Content/Stock/list.html.twig
```diff
@@ -1,26 +0,0 @@
...
-
-{% block main_content %}
-
-    <div class="js-multiple-grids-rows-unified">
-        {{ gridView.render() }}
-
-        {% embed '@ShopsysFramework/Admin/Inline/FixedBar/fixedBar.html.twig' %}
-            {% block fixed_bar_content %}
-                <button class="js-drag-and-drop-grid-submit-all btn btn--disabled" type="button">{{ 'Save changes in order'|trans }}</button>
-            {% endblock %}
```
vitek-rostislav: I noticed the button is now rendered automatically for drag and drop grids. I am a little bit worrried about the UX - it is not in the fixed bar anymore, hence users do not see the "you have made changes..." message when they change the ordering. I suppose I can live with that, however, I would at least reconsider the "Save changes in order" button states - it is light grey when disabled and regular grey (btn-secondary) when enabled - the change is not well noticeable IMHO. Lets talk about it, please

packages/administration/templates/content/store/listGrid.html.twig
```diff
@@ -0,0 +1,22 @@
+{% extends '@ShopsysAdministration/datagrid/grid.html.twig' %}
+
+{% block grid_value_cell_id_name %}
+    <a href="{{ url('admin_store_edit', { id: row.s.id }) }}">{{ value }}</a> {% if row.s.isDefault %}*{% endif %}
```
vitek-rostislav: this could be unified with stocks - there is a badge pill used instead if the asterisk to mark the default one :)

packages/administration/templates/content/personalData/index.html.twig
```diff
@@ -0,0 +1,40 @@
...
+{% block main_content %}
+    {{ render(controller('Shopsys\\FrameworkBundle\\Controller\\Admin\\DomainController::domainTabsAction')) }}
+
+    {{ form_start(form) }}
+    <div class="card mb-3">
+        <div class="card-header">
+            <h3 class="card-title">{{ 'Personal information overview'|trans }}</h3>
+            <div class="card-actions">
+                <a href="{{ findUrlByDomainId('front_personal_data', {}, domainId) }}" target="_blank">
+                    {{ findUrlByDomainId('front_personal_data', {}, domainId) }}
```
vitek-rostislav, line 18: could we put some label here? Something like "The storefront link: http://..."

packages/administration/templates/form/content/ProductParameterValue.html.twig
```diff
@@ -0,0 +1,73 @@
...
+        </div>
+    </div>
+{% endmacro %}
+
+{% block product_parameter_value_row %}
+    {% import _self as self %}
+    <div class="js-form-group">
+        {% if form_errors(form) is not empty %}
+            <p class="card-subtitle">{{ form_errors(form) }}</p>
+        {% endif %}
```
vitek-rostislav, line 38: using the same styling for both errors and help is intended here?

packages/administration/templates/partial/select_admin_domains.html.twig
```diff
@@ -0,0 +1,28 @@
+<div class="dropdown-menu overflow-y-auto overflow-x-hidden js-domains-select" style="max-height: 300px; min-width: calc(100% + 5rem);">
+    {{ form_start(form, {'action': path('admin_domains_filter'), 'method': 'POST'}) }}
+    {% for child in form.domains %}
+        <label class="dropdown-item" onclick="event.stopPropagation()">
```
vitek-rostislav: could not we deal with the `onClick` handler within `SelectAdminDomains.js`?

packages/administration/templates/partial/select_admin_domains.html.twig
```diff
@@ -0,0 +1,28 @@
...
+                   {% if child.vars.checked %}checked{% endif %}
+            >
+            {{ child.vars.label }}
+        </label>
+        {% do child.setRendered %}
+    {% endfor %}
+
+    <div class="dropdown-divider"></div>
+
+    <label class="dropdown-item" onclick="event.stopPropagation()">
```
vitek-rostislav: could not we deal with the `onClick` handler within `SelectAdminDomains.js`?

packages/administration/templates/partial/required_settings.html.twig
```diff
@@ -0,0 +1,11 @@
+{# @var requiredSettingsMessages string[] #}
+<div class="alert alert-warning d-flex" role="alert">
```
vitek-rostislav, line 2: is this annotation good for anything? I mean, I can understand it for objects where you need to access a property/method,  but for an array of strings?

packages/administration/templates/partial/quick_domain_filter.html.twig
```diff
@@ -0,0 +1,35 @@
...
+                       }) }}">
+                        {{ 'All domains'|trans }}
+                    </a>
+                {% endif %}
+
+                {% for domainConfig in domainConfigs %}
+                    {% if domainConfig.id == selectedDomainId %}
+                        <span class="btn active">
+                    {{ domain_icon(domainConfig.id) }}
+                            {{ domainConfig.name }}
```
vitek-rostislav, line 22: here is a weird spacing, no biggie, but :shrug:

packages/framework/src/Twig/IconExtension.php
```diff
@@ -47,18 +47,15 @@ public function getFunctions(): array
...
     protected function getInfoIcon(
         string $message,
-        string $tooltipPlacement = 'top',
         string $additionalClasses = '',
     ): string {
-        return $this->twigEnvironment->render('@ShopsysFramework/Admin/Inline/Icon/info_icon.html.twig', [
+        return $this->twigEnvironment->render('@ShopsysAdministration/partial/info_icon.html.twig', [
             'message' => $message,
-            'tooltipPlacement' => $tooltipPlacement,
             'additionalClasses' => $additionalClasses,
```
vitek-rostislav, line 59: the placement is now handled automatically, right?

packages/administration/src/Twig/DomainIconExtension.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Shopsys\AdministrationBundle\Twig;
+
+use Override;
+use Shopsys\AdministrationBundle\Twig\Runtime\DomainIconRuntime;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+final class DomainIconExtension extends AbstractExtension
+{
```
vitek-rostislav, line 13: are you sure you want this final? :smile:

packages/administration/config/services.yaml
```diff
@@ -32,3 +32,8 @@ services:
     Shopsys\AdministrationBundle\Controller\:
         resource: '../src/Controller/'
         tags: [ 'controller.service_arguments' ]
+
+    Shopsys\AdministrationBundle\Twig\Runtime\DomainIconRuntime:
+        arguments:
+            - '%kernel.project_dir%'
+        tags: ['twig.runtime']
```
vitek-rostislav, line 54: the tags should be added automatically, but you need that here because of defining the arguments here?

packages/administration/src/Twig/DateTimeAgoExtension.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function getFilters(): array
+    {
+        return [
+            new TwigFilter('date_time_ago', [DateTimeAgoRuntime::class, 'dateTimeAgo']),
+        ];
```
vitek-rostislav, line 22: I just wonder, why is the runtime better than defining the functionality directly here in the extension?

vitek-rostislav: ah, performance, right? https://symfony.com/doc/current/templates.html#templates-twig-extension

packages/framework/assets/js/admin/components/OrderItems.js
```diff
@@ -99,9 +96,7 @@ export default class OrderItems {
 
     removeItem($item) {
         const $collection = $item.closest('#js-order-items');
-        const index = $item.data('index');
 
-        removeItemFromCollection('#js-order-items', index);
         $item.remove();
```
vitek-rostislav, line 102: Should we add `FormChangeInfo.showInfo();` here as well?

packages/framework/assets/js/admin/components/Parameters.js
```diff
@@ -12,8 +9,6 @@ export default class Parameters {
             const $collection = $(this).closest('.js-parameters');
 
             const $item = $(this).closest('.js-parameters-item');
-            const index = $item.data('index');
-            removeItemFromCollection('.js-parameters', index);
             $item.remove();
```
vitek-rostislav, line 12: `FormChangeInfo.showInfo();` instead?

packages/framework/assets/js/admin/components/TransportPriceWithWeightLimitCollection.js
```diff
@@ -14,8 +10,6 @@ export default class TransportPriceWithWeightLimitCollection {
 
         $transportPricesCollection.on('click', '.js-transport-prices-item-remove', function (event) {
             const $item = $(this).closest('.js-transport-prices-item');
-            const index = $item.data('index');
-            removeItemFromCollection('.js-transport-prices', index);
             $item.remove();
```
vitek-rostislav, line 14: `FormChangeInfo.showInfo();` instead?

packages/framework/assets/js/admin/components/TransportPriceWithWeightLimitCollection.js
```diff
@@ -35,8 +29,6 @@ export default class TransportPriceWithWeightLimitCollection {
             $collection.append($item);
             new Register().registerNewContent($item);
 
-            addNewItemToCollection('.js-transport-prices', index);
-
```
vitek-rostislav, line 39: `FormChangeInfo.showInfo();` instead?

packages/framework/assets/js/admin/validation/form/validationAdvert.js
```diff
@@ -1,47 +0,0 @@
...
-import { VALIDATION_GROUP_DEFAULT } from './validation';
-
-export default function validationAdvert() {
-    const $advertForm = $('form[name="advert_form"]');
-
-    const getCheckedType = () => $advertForm.find('input[name="advert_form[settings][type]"]:checked').val();
-
-    const initAdvertForm = () => {
-        $advertForm.find('.js-advert-type-content').hide().filter(`[data-type=${getCheckedType()}]`).show();
-    };
```
vitek-rostislav, line 13: oh yes, that is that clever idea to put a hiding behavior into a script named "validationAdvert" :smile:

packages/framework/assets/js/admin/validation/readme.txt
```diff
@@ -1,7 +0,0 @@
-Client-side validation requires:
-* subform must have HTML container with attribute id="{form_name}_{subform_name}" eg.: customer_deliveryAddressData
```
vitek-rostislav, line 2: Another issue completed :smile: https://github.com/shopsys/shopsys/issues/406

docs/administration/javascript-utilities.md
```diff
@@ -0,0 +1,199 @@
...
+
+| Attribute                   | Description                               | Default          |
+| --------------------------- | ----------------------------------------- | ---------------- |
+| `data-confirm-window`       | Required: enables confirmation behavior   | -                |
+| `data-confirm-message`      | Confirmation message content              | `''`             |
+| `data-confirm-style`        | Modal style (danger, warning, info, etc.) | `'danger'`       |
+| `data-confirm-continue-url` | URL to navigate to when confirmed         | Element's `href` |
+
+#### Configuration Options
+
```
vitek-rostislav: as the configuration options are usable within the programmatic usage, I would move this section above (below "Programmatic Usage" example)

packages/administration/templates/form/type/BlogCategoriesType.html.twig
```diff
@@ -0,0 +1,33 @@
+{% block blog_categories_widget %}
+    <div class="js-category-tree-form"
```
vitek-rostislav, line 2: just a thought that could be extracted (and maybe already is) - are not the form type + templates same as the ones for categories? Could ot we unify them?

packages/framework/src/Resources/views/Admin/Form/theme.html.twig
```diff
@@ -1,1112 +0,0 @@
...
-
-{% block form_row %}
-    {% if render_form_row %}
-        <div class="form-line{% if multidomain is defined and multidomain %} form-line--no-horizontal-gap{% endif %}{{ attr.disabledField is defined ? ' form-input-disabled form-line--disabled' }}">
-        {{ form_errors(form) }}
-        {{ form_label(form, label) }}
-    {% endif %}
-    {{ form_widget(form) }}
-    {% if attr.disabledField is defined %}
-        <span class="form-disabled {{ attr.disabledFieldClass|default('') }} js-tooltip" data-toggle="tooltip" data-placement="{{ attr.disabledFieldPlacement|default('right') }}"
```
vitek-rostislav, line 73: I am wondering whether the form fields disabling still works...Probably so, however, I would like to talk about it. The only direct usage of `disabledFormFields`  is in `packages/administration/templates/content/category/list.html.twig`now, is that correct? (at least, I would expect the same in the blog categories list template :))

packages/framework/assets/js/admin/jQuery/registerTooltip.js
```diff
@@ -0,0 +1,21 @@
+import { Tooltip } from '@tabler/core';
+import $ from 'jquery';
+
+export default function registerTooltip() {
```
vitek-rostislav, line 4: should not this be in the administration package? :)

packages/framework/assets/js/admin/registerAdmin.js
```diff
@@ -1,15 +1,14 @@
...
 import './search/search';
 
-import './validation/customization';
-import './validation/form';
-
 import Register from '../common/utils/Register';
+import registerTooltip from './jQuery/registerTooltip';
 
 export default function registerAdmin(afterRegistrationCallback = null) {
+    registerTooltip();
```
vitek-rostislav: at the tme I am review the pull request, product picker is not working. I would like to ensure tooltip works for dynamically added content, however, I am not able to find any relevant use case besides the products picker :smile: So let's take this comment as a reminder

packages/framework/assets/js/admin/components/Product.js
```diff
@@ -7,21 +7,20 @@ export default class Product {
...
 
     static initializeSideNavigation($container) {
-        const $productDetailNavigation = $container.find('.js-product-detail-navigation');
-        const $webContent = $('.web__content');
+        const $productDetailNavigation = $container.find('.toc .nav');
+        const $topOffset = $('.page-header').offset().top;
 
-        $('.form-group__title, .form-full__title').each(function () {
+        $('#product_form h3').each(function () {
             const $title = $(this);
```
vitek-rostislav, line 17: another feature idea/request - we could highlight the currently "active" section in the TOC (we can create a separate issue for that but I was told it should be easy :grin:)

packages/framework/assets/js/admin/components/Advert.js
```diff
@@ -0,0 +1,33 @@
+import Register from '../../common/utils/Register';
+
+export default class Advert {
+    constructor($container) {
+        this.$positionNameSelect = $container.find('#advert_form_settings_positionName');
+        this.$categoryTreeRow = $container.find('[data-js-advert-categories]');
```
vitek-rostislav, line 6: I know we have discussed this in a team meeting, however, I have another (not-so-mind-blowing-idea) idea - what if we set a js prefixed class for the affected inputs? So We do not have to rely on the form structure, and rather use that class as an identifier of the desired element?

packages/framework/src/Form/Admin/Complaint/ComplaintFormType.php
```diff
@@ -169,8 +169,8 @@ private function createBasicInformationGroup(
                         'groups' => [static::VALIDATION_GROUP_TYPE_MONEY_RETURN],
                     ]),
                 ],
-                'attr' => [
-                    'class' => 'js-complaint-bank-account-number',
+                'row_attr' => [
+                    'data-js-complaint-bank-account-number' => null,
```
vitek-rostislav, line 173: why is this better than using just the js-prefixed class?

vitek-rostislav (changes requested):
Hi, awesome job there, wow :clap: I have a added some comments, please check them out. I have also noticed a few functional bugs, I will put them below in a separate comment

vitek-rostislav:
From testing during the review:
- [x] There is no cursor-pointer for radio or checkbox inputs
- [x] Products picker is not working (I am not able to add products, the icons in the list are not interactive - no tooltips for selling denied, no popover for visibilities)
- [x] Product picker has a broken styling (see e.g. "Add product" in Order edit form)
- [x] When you click between label + tomselect  (see e.g. "Default VAT rate" setting at `admin/vat/list/`) - it opens the select, but another click at the same position does not close it
- [x] I see an empty results pop-up when nothing is found within the search, the styling seems a little bit broken in such a case, see the screen below. Btw, the position seems a little bit misplaced, even if there are some results, I would expect it below the input, not covering it

Broken (?) empty search:
image

Broken product picker:
image

vitek-rostislav:
btw, feel free to rebase the branch before adding any fixups (if you feel like solving the conflicts now :smile: )

packages/administration/templates/form/theme.html.twig
grossmannmartin: exactly :)

packages/administration/templates/layout/layout_with_panel.html.twig
grossmannmartin: yes, I didn't find a better way

docs/introduction/using-form-types.md
grossmannmartin: yes

packages/framework/src/Controller/Admin/AdministratorController.php
grossmannmartin: It's kind of an in-between state now... I realize it looks weird, but I would like to move controllers and other administration-related code in the future

packages/administration/templates/form/type/ColorPickerType.html.twig
grossmannmartin: d7217935e6 rewritten. I like it more in type itself

packages/framework/src/Form/DatePickerType.php
grossmannmartin: inconsitenty eliminated

packages/framework/assets/js/admin/components/SortableValues.js
grossmannmartin: extracted into method in  701ec89f0a

packages/administration/templates/content/advert/list.html.twig
grossmannmartin: I think we would assume many things in that case - we had to assume that the grid is always passed as the `gridView` variable for instance (for example parameterGroup/list.html.twig uses different name). 

Then it would bring the confusion about what template should I extend. If I have list that should have different pre_title from `Overview`, should I extend the basic list template (basically in vain in that case)?

I would rather push crud controllers and unify it on that level

packages/framework/src/Controller/Admin/ArticleController.php
grossmannmartin: It's probably just a forgotten attempt of something. Removed in 34d4da31ad

packages/administration/templates/content/complaint/advancedSearch/advancedSearch.html.twig
grossmannmartin: I had the dream... But the reality showed that it's not so simple. I would like to do it separately.

packages/administration/templates/content/complaint/listGrid.html.twig
grossmannmartin: true - 7b7c67761a

packages/administration/templates/content/categorySeo/readyCombination.html.twig
grossmannmartin: It seems like it work directly with _self  🤷‍♂️ 

I found this article when they mention that macros are now automatically imported https://symfony.com/blog/simpler-macros-in-twig-templates

packages/administration/templates/content/blog/article/list.html.twig
grossmannmartin: Good catch! I have unified them in 03fd6027a0, bf344085dd, 00c41aa878

packages/framework/src/Model/Product/Flag/FlagGridFactory.php
grossmannmartin: I have extracted the manipulator function into a separate named function. Hopefully it's better - d82fe78d91

packages/administration/templates/content/flag/deleteForbidden.html.twig
grossmannmartin: comment is removed. it's not consistent with the javascript as JS renders the buttons in footer, which is not possible in ajaxconfirm - a506ff8791

packages/administration/templates/content/flag/listGrid.html.twig
grossmannmartin: It seems to be tabler badges behavior.

packages/administration/templates/content/default/cronDetail.html.twig
grossmannmartin: Yes, yes it would. e455207f0f

packages/administration/templates/content/mail/detail.html.twig
grossmannmartin: Yes, I have similar concernes as with list templates.

packages/administration/templates/content/payment/listGrid.html.twig
grossmannmartin: It's probably a feature 😄

packages/framework/src/Form/Admin/Superadmin/MailWhitelistFormType.php
grossmannmartin: I see... I don't know whether it should be always required. Probably semantically here it should be required... but the rendering is quite weird as you mentioned. 
Do you think I should try to fix the rendering instead?

packages/administration/templates/content/notificationBar/edit.html.twig
grossmannmartin: You are absolutely right 😄 

I tried to balance between right and compatible. I am open to suggestions, though.

packages/administration/templates/content/uploadedFile/list.html.twig
grossmannmartin: Let's talk please.... I have absolutely no idea what this should do and why we don't use it properly 😇

packages/administration/templates/content/order/listGrid.html.twig
grossmannmartin: already done in one of previous commit

packages/framework/src/Controller/Admin/ParameterGroupController.php
grossmannmartin: I would keep it for now. Maybe it will be handy. This seems exactly like one of the `we have framework` methods

packages/administration/templates/content/priceList/listGrid.html.twig
grossmannmartin: I have switched from badges to statuses in flags and prices in transport and payment list in 0a7e8c87c9 and 94026ca5db

Other usages of badge seems like a fit to me

packages/framework/assets/js/admin/components/PriceListProductPicker.js
grossmannmartin: Yes it's "just" weird. It will work

packages/administration/templates/content/promoCode/listGrid.html.twig
grossmannmartin: cool - 5fd1934c6f

packages/framework/src/Form/Admin/Seo/SeoRobotsSettingFormType.php
grossmannmartin: changed - e7a22e7c94

packages/framework/src/Resources/views/Admin/Content/Stock/list.html.twig
grossmannmartin: Let's talk please :)

packages/administration/templates/content/store/listGrid.html.twig
grossmannmartin: good catch 09b80a734a

packages/administration/templates/content/personalData/index.html.twig
grossmannmartin: 21669e05bc

packages/administration/templates/form/content/ProductParameterValue.html.twig
grossmannmartin: yes, they are actually rendered differently in the theme itself

Snímek obrazovky 2025-09-17 v 16 29 06

packages/administration/templates/partial/select_admin_domains.html.twig
grossmannmartin: we could - e0aea12e61

packages/administration/templates/partial/select_admin_domains.html.twig
grossmannmartin: e0aea12e61

packages/administration/templates/partial/required_settings.html.twig
grossmannmartin: It's more for the reference of what is passed to the template

packages/administration/templates/partial/quick_domain_filter.html.twig
grossmannmartin: ugh.... phpstorm format span tags very... oddly. fixed in fcb0f84cf0

packages/framework/src/Twig/IconExtension.php
grossmannmartin: right

packages/administration/config/services.yaml
grossmannmartin: exactly

packages/administration/src/Twig/DomainIconExtension.php
grossmannmartin: sure. If someone needs a new function, they should create new extension. 
If they want to modify the behavior ... well... they can change the template and if the logic is bad, they should send PR 🤓

packages/administration/src/Twig/DateTimeAgoExtension.php
grossmannmartin: Yes, performance

packages/framework/assets/js/admin/components/OrderItems.js
grossmannmartin: Oh, nice! 290eb03960

packages/framework/assets/js/admin/validation/form/validationAdvert.js
grossmannmartin: Yes 😄

packages/framework/assets/js/admin/components/Parameters.js
grossmannmartin: 1553145516

packages/framework/assets/js/admin/components/TransportPriceWithWeightLimitCollection.js
grossmannmartin: a873127fde

packages/administration/templates/form/type/BlogCategoriesType.html.twig
grossmannmartin: They are similar. Please let's try to unify them separately

docs/administration/javascript-utilities.md
grossmannmartin: f187c31244

packages/framework/src/Resources/views/Admin/Form/theme.html.twig
grossmannmartin: let's talk. I'm not sure what exactly do you mean

packages/framework/assets/js/admin/jQuery/registerTooltip.js
grossmannmartin: moved - ecc453dbd7

packages/framework/assets/js/admin/registerAdmin.js
grossmannmartin: product picker now works (probably fixed by refactoring during scrollspy implementation)

packages/framework/assets/js/admin/components/Product.js
grossmannmartin: implemented scrollspy in 5667bbb037

packages/framework/src/Form/Admin/Complaint/ComplaintFormType.php
grossmannmartin: two main points:
- it's not mixed with classes for styling
- classes are not merged, but overwritten, so I would be forced to copy tabler classes here

data attributes seems cleaner solution to me

packages/framework/assets/js/admin/components/Advert.js
grossmannmartin: we should / could... What about making it systematically (as a separate issue) instead of ad-hoc fix of one occurence?

grossmannmartin:
@vitek-rostislav 

Thank you so much for the comprehensive review. 

about points from testing: 

> There is no cursor-pointer for radio or checkbox inputs

This seems to be tabler default

> Products picker is not working (I am not able to add products, the icons in the list are not interactive - no tooltips for selling denied, no popover for visibilities)

Should be fixed

> Product picker has a broken styling (see e.g. "Add product" in Order edit form)

Should be fixed

> When you click between label + tomselect (see e.g. "Default VAT rate" setting at admin/vat/list/) - it opens the select, but another click at the same position does not close it

Yes. The behavior is the same for clicking on the label It's probably because label take all the space to the end of the line 

> I see an empty results pop-up when nothing is found within the search, the styling seems a little bit broken in such a case, see the screen below. Btw, the position seems a little bit misplaced, even if there are some results, I would expect it below the input, not covering it

that's weird... where did this come from? fixed

packages/administration/templates/content/complaint/advancedSearch/advancedSearch.html.twig
vitek-rostislav: and maybe after CRUD :smile:

packages/administration/templates/content/categorySeo/readyCombination.html.twig
vitek-rostislav: 6 years ago, ok :D

packages/administration/templates/content/flag/listGrid.html.twig
vitek-rostislav: yes :(

packages/framework/src/Form/Admin/Superadmin/MailWhitelistFormType.php
vitek-rostislav: yes please, try it if it will be simple, otherwise keep it as is :pray:

packages/administration/templates/content/notificationBar/edit.html.twig
vitek-rostislav: maybe we will address it in the far future

packages/framework/assets/js/admin/components/TransportPriceWithWeightLimitCollection.js
vitek-rostislav: yet here

packages/administration/templates/form/type/BlogCategoriesType.html.twig
vitek-rostislav: TODO create an issue

packages/framework/assets/js/admin/registerAdmin.js
vitek-rostislav: the added products now do not render image properly, see https://prnt.sc/uawwS8oSKfcB

packages/framework/assets/js/admin/components/Product.js
```diff
@@ -7,24 +7,66 @@ export default class Product {
...
             });
         });
+
+        Product.initIntersectionObserver(sections);
+    }
+
+    static initIntersectionObserver(sections) {
+        if (sections.length === 0) return;
+
+        const HEADER_HEIGHT = 100;
```
vitek-rostislav: nitpick, this could be a class constant as it is also a few lines above

packages/framework/assets/js/admin/components/Advert.js
vitek-rostislav: yes, TODO create an issue

packages/framework/src/Resources/views/Admin/Form/theme.html.twig
grossmannmartin: it just works

packages/administration/templates/form/theme.html.twig
grossmannmartin: added documentation in 9749940187

packages/framework/assets/js/admin/components/TransportPriceWithWeightLimitCollection.js
grossmannmartin: 🫣 I missed that one - fixed 008ee7adbe

packages/framework/src/Form/Admin/Superadmin/MailWhitelistFormType.php
grossmannmartin: I have reworked it - now the yesNoType has constraint and is required on the form type level. This simplifies usage significantly. 

I did not found any valid use-case for making yesnotype nullable. 

Labels individual radios are rendered as not-required (to remove the asterisk), but it's ok because constraint is forced on the parent form type level.  

d80920b56e

grossmannmartin: and one more fix in ff71921d7e :innocent:

packages/framework/assets/js/admin/registerAdmin.js
grossmannmartin: I have fixed the rendering of the add button in chrome on the page with order detail in product picker -  57f2c95a11

Also I have added product image to be shown immediately after adding a product instead of after save -  94f88c76cf

(also I have updated an issue to unify product pickers - internal ID SSP-2459)

packages/administration/templates/content/uploadedFile/list.html.twig
grossmannmartin: I have created a new issue for this - internal ID SSP-3584

packages/administration/templates/form/type/BlogCategoriesType.html.twig
grossmannmartin: created an issue - internal ID SSP-3585

packages/framework/assets/js/admin/components/Advert.js
grossmannmartin: created an issue - internal id SSP-3586

packages/framework/assets/js/admin/components/Product.js
grossmannmartin: fixed in 5fc8900a7a

grossmannmartin:
@vitek-rostislav I have tried to somehow fix the problem about select does not close after second click on label. I failed miserably... Let's discuss what to do with it, please

packages/framework/src/Resources/views/Admin/Content/Stock/list.html.twig
grossmannmartin: I have improved the form change info. Now the info is displayed even when grid do not use fixed bar. 
Also, few bugs were fixed:
- form change info was not displayed in production environment
- beforeunload do not allow change the message anymore
- fix beforeunload in safari

vitek-rostislav (approved):

packages/administration/assets/src/js/passwordWithStrengthInput.js
```diff
@@ -0,0 +1,112 @@
...
+
+        const classes = {};
+
+        classes.lower = /[a-z]/.test(value) ? 'text-correct' : 'text-incorrect';
+        classes.upper = /[A-Z]/.test(value) ? 'text-correct' : 'text-incorrect';
+        classes.number = /\d/.test(value) ? 'text-correct' : 'text-incorrect';
+        classes.length = value.length >= 10 ? 'text-correct' : 'text-incorrect';
+
+        for (const key in classes) {
+            console.log(`[data-js-password-help-text-requirement="${key}"]`);
```
vitek-rostislav: Forgotten console log?

packages/administration/assets/src/js/utils/passwordStrengthCalculator.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        const pool = lower + upper + digit + symbol + control + other;
+        const entropy = chars * Math.log2(pool) + (length - chars) * Math.log2(chars);
+
+        if (entropy >= 120) return this.STRENGTH_VERY_STRONG;
+        if (entropy >= 100) return this.STRENGTH_STRONG;
+        if (entropy >= 80) return this.STRENGTH_MEDIUM;
+        if (entropy >= 60) return this.STRENGTH_WEAK;
+        return this.STRENGTH_VERY_WEAK;
```
vitek-rostislav: This is quite unusual formatting, however, if it passes our standards, then :shrug:

packages/framework/src/Resources/translations/validators.cs.po
```diff
@@ -4,6 +4,9 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Language: cs\n"
 
+msgid ""
```
vitek-rostislav, line 7: I am wondering how to avoid this. It is because of the empty strings in the constraint messages, right? I see it is not possible to set the message to `null`, as then the default "This value is not valid" message is used. However, the messages are automatically extracted thanks to our implementation, so maybe we could tweak our `ConstraintMessageExtractor` to skip the empty strings :thinking:

packages/administration/assets/src/js/utils/passwordStrengthCalculator.js
grossmannmartin: We could write it like 

```
switch (true) {
    case entropy >= 120: return this.STRENGTH_VERY_STRONG;
    case entropy >= 100: return this.STRENGTH_STRONG;
    case entropy >= 80:  return this.STRENGTH_MEDIUM;
    case entropy >= 60:  return this.STRENGTH_WEAK;
    default:             return this.STRENGTH_VERY_WEAK;
}
```

Is it better? Or do you have any other idea?

packages/administration/assets/src/js/passwordWithStrengthInput.js
grossmannmartin: yes, removed :) thanks

packages/administration/assets/src/js/utils/passwordStrengthCalculator.js
vitek-rostislav: Yeah, I like the switch better when I see it (I have learned not to dislike the `switch(true)` approach some time ago :smile: ). However, I did not have the switch in mind when writing the comment, I was referring just to the inline returns and the missing empty line before the last one. I would just expect something like this:

```
if (entropy >= 120) {
    return this.STRENGTH_VERY_STRONG;
}

if (...) {
// ...

return this.STRENGTH_VERY_WEAK;
```

grossmannmartin: I used the switch instead. Full ifs seem too long for me

packages/framework/src/Resources/translations/validators.cs.po
grossmannmartin: Oh, that's where it comes from... the empty string.
Null even causes an error in one of the constraints. That's a good idea to skip empty strings (in another PR, please 😄).

vitek-rostislav: sticky note it is :smile:

packages/administration/templates/content/closedDay/edit.html.twig
```diff
@@ -0,0 +1,6 @@
+{% extends '@ShopsysAdministration/content/closedDay/detail.html.twig' %}
+
+{% block title %}- {{ 'Editing holiday / internal day - {{ name }}'|trans({ '{{ name }}': closedDay.name }) }}{% endblock %}
```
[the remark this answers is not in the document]
grossmannmartin: no

packages/administration/templates/content/stock/edit.html.twig
```diff
@@ -0,0 +1,14 @@
+{% extends '@ShopsysAdministration/content/stock/detail.html.twig' %}
+
+{% block title %}- {{ 'Editing warehouse'|trans }} - {{ stock.name }}{% endblock %}
+
+
```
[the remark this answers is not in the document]
grossmannmartin: This is definitely the worst imperfection in this PR. denied

packages/administration/src/Twig/Runtime/DateTimeAgoRuntime.php
```diff
@@ -0,0 +1,63 @@
...
+            );
+        }
+
+        $datePart = match ($dateStr) {
+            $today => 'today',
+            $yesterday => 'yesterday',
+            default => throw new LogicException('Date should be either today or yesterday at this point'),
+        };
+
+        return $datePart . ' ' . t('at') . ' ' . $this->dateTimeFormatter->format(
```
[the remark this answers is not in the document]
grossmannmartin: I don't think so, Tim

packages/administration/templates/content/superadmin/urlsListGrid.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends '@ShopsysFramework/Admin/Layout/layoutWithPanel.html.twig' %}
+{% extends '@ShopsysAdministration/layout/layout_with_panel.html.twig' %}
```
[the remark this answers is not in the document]
grossmannmartin: Because it's named *grid, it does not mean it contains grid-specific blocks

PR #3881, opened 2025-03-25
admin locale now can be set to any locale

vitek-rostislav:
It is no more required to use one of the SF domain locales for admin localization. When admin locale is set to a value that is not persisted within entity translations (i.e. the  locales used on the domains), translatable attributes in admin are then displayed in the first domain locale.

#### Fixes issues
... 

----
  - https://rv-admin-locale.odin.shopsys.cloud
  - https://cz.rv-admin-locale.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Command/AdminLocalesInfoCommand.php
+++ b/packages/framework/src/Command/AdminLocalesInfoCommand.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Command;
 
 use Override;
-use Shopsys\FrameworkBundle\Model\Localization\Localization;
+use Shopsys\FrameworkBundle\Model\Administrator\AdministratorLocalizationFacade;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
@@ -19,10 +19,10 @@ use Symfony\Component\Console\Style\SymfonyStyle;
 class AdminLocalesInfoCommand extends Command
 {
     /**
-     * @param \Shopsys\FrameworkBundle\Model\Localization\Localization $localization
+     * @param \Shopsys\FrameworkBundle\Model\Administrator\AdministratorLocalizationFacade $administratorLocalizationFacade
      */
     public function __construct(
-        protected readonly Localization $localization,
+        protected readonly AdministratorLocalizationFacade $administratorLocalizationFacade,
     ) {
         parent::__construct();
     }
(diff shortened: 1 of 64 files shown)


packages/framework/src/Command/AdminLocalesInfoCommand.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Administrator\AdministratorLocalizationFacade $administratorLocalizationFacade
+     * @param string|null $name
+     */
+    public function __construct(
+        protected readonly AdministratorLocalizationFacade $administratorLocalizationFacade,
+        ?string $name = null,
+    ) {
+        parent::__construct($name);
+    }
```
grossmannmartin, line 28: as we hate the name set in the AsCommand attribute, we may omit the name here and simplify the constructor to just

```php
public function __construct(
        protected readonly AdministratorLocalizationFacade $administratorLocalizationFacade,
    ) {
        parent::__construct();
    }
```

packages/framework/src/Component/CKEditor/CKEditorRendererDecorator.php
```diff
@@ -5,17 +5,17 @@
...
 {
     /**
      * @param \FOS\CKEditorBundle\Renderer\CKEditorRendererInterface $baseCkEditorRenderer
-     * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+     * @param \Shopsys\FrameworkBundle\Model\Localization\Localization $localization
      */
     public function __construct(
         protected readonly CKEditorRendererInterface $baseCkEditorRenderer,
-        protected readonly RequestStack $requestStack,
+        protected readonly Localization $localization,
```
grossmannmartin: Localization seems unused

packages/framework/src/Component/EntityLog/Model/EntityLogFacade.php
```diff
@@ -119,7 +119,7 @@ public function getEntityReadableIdentifierByEntityAndLoggableSetup(
         }
 
         if ($loggableSetup->isLocalized()) {
-            return call_user_func([$entity, $functionName], $this->localization->getDefaultAdminLocale());
+            return call_user_func([$entity, $functionName], $this->administratorLocalizationFacade->getLocaleForEntityLog());
```
grossmannmartin: I'm not sure if this specific method should be placed in administratorLocalizationFacade. 
Maybe it would be better to have this method directly in EntityLogFacade?

grossmannmartin (changes requested):

packages/framework/src/Component/EntityLog/Model/EntityLogFacade.php
vitek-rostislav: that is a good idea :ok_hand:

grossmannmartin:
Nice work here 🦾 

I have some points from testing, can you please check them out?

### 1  

If you set the admin locale to a domain that is not set as the domain locale, the build fails during the Version20241106123834 migration execution. 

### 2

If you have a shop with only one domain, and you set allowed admin locales to `cs, en`, the locale switcher is not displayed. 
(This will probably be a problem from the past, but now it's obvious as we're separating the locales)

grossmannmartin (approved):

PR #3889, opened 2025-03-28
admin product detail: display main category full path

vitek-rostislav:

Displaying just the main category name could be confusing (there might be multiple categories with the same name) so displaying the full path is clearer for the administrators. Moreover, moving the information inside the categories widget makes the product detail page more compact.

Before:
Screenshot from 2025-03-28 14-15-02

After:
Screenshot from 2025-03-28 14-15-34

Moreover, the new style is better for one-domain projects, where the original approach was at least weird, see the screens below:

Before:
image

After:
image

#### Fixes issues
... 

----
  - https://rv-main-category.odin.shopsys.cloud
  - https://cz.rv-main-category.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Form/CategoriesType.php
+++ b/packages/framework/src/Form/CategoriesType.php
@@ -5,7 +5,12 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Form;
 
 use Override;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Form\Transformers\CategoriesTypeTransformer;
+use Shopsys\FrameworkBundle\Model\Category\CategoryFacade;
+use Shopsys\FrameworkBundle\Model\Category\Exception\CategoryNotFoundException;
+use Shopsys\FrameworkBundle\Model\Localization\Localization;
+use Shopsys\FrameworkBundle\Model\Product\Product;
 use Symfony\Component\Form\AbstractType;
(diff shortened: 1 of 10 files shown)


packages/framework/src/Form/CategoriesType.php
```diff
@@ -84,4 +94,24 @@ public function getParent(): ?string
...
+    {
+        if ($options['product'] === null) {
+            return null;
+        }
+
+        try {
+            $domainConfig = $this->domain->getDomainConfigById($options['domain_id']);
+            $categoriesInPath = $this->categoryFacade->getCategoryNamesInPathFromRootToProductMainCategoryOnDomain($options['product'], $domainConfig);
+
+            return implode(' > ', $categoriesInPath);
```
grossmannmartin, line 119: it's a little strange that the main category path is translated to the domain language, while categories are in the admin locale. 

Snímek obrazovky 2025-03-28 v 18 33 52

grossmannmartin (approved):

packages/framework/src/Form/CategoriesType.php
vitek-rostislav: yes! Thanks for pointing this out :pray: I will keep  this open until https://github.com/shopsys/shopsys/pull/3881 is merged and then try to do something about it :relaxed:

vitek-rostislav: this should do the trick - 8b2fb12b2d

pk16011990 (approved):

PR #3917, opened 2025-04-08
added icons for displaying visibility of categories

TomasLudvik:
Screenshot at Apr 08 11-37-44

----
  - https://tl-category-visibility.odin.shopsys.cloud
  - https://cz.tl-category-visibility.odin.shopsys.cloud


vitek-rostislav (approved):
I like it, I have just one thought - that "..on all domains" text will be displayed also on 1-domain projects, right? We could either modify the code to include that scenario or it can be adjusted on such a project by the translation in messages.po file :slightly_smiling_face: So I believe it is good as is :slightly_smiling_face:

vitek-rostislav:
ok, one thing from quick test - the icons are not displayed at all when I select a particular domain using the domain tabs

vitek-rostislav (changes requested):
ok, one thing from a quick test - the icons are not displayed at all when I select a particular domain using the domain tabs

pk16011990 (approved):

PR #3945, opened 2025-04-24
UploadedFileDataFixture uses first allowed domain locale

malyMiso:
UploadedFileDataFixture uses first allowed domain locale. Otherwise it is dependent on current domain config, which is not available in data fixtures, until CartDataFixture run. Current domain is set in CartDataFixture, therefore fixtures that run after  CartDataFixture can access current domain config :see_no_evil: . 

#### Fixes issues
... 

----
  - https://mt-demo-files.odin.shopsys.cloud
  - https://cz.mt-demo-files.odin.shopsys.cloud


grossmannmartin:
it's already solved since 22197792d38d6db9e4d19a1326da49793c779aaa

PR #3949, opened 2025-04-25
split customer translations to separate domain

grossmannmartin:

Now it's easier to translate only the part of the application the customer is interested in (admin/storefront).
New translation domains `customerMessages` and `customerValidators` are used for translations that are presented to the customer, while the previously existing ones are used in administration/backend. 

#### Fixes issues
... 

----
  - https://mg-split-translations.odin.shopsys.cloud
  - https://cz.mg-split-translations.odin.shopsys.cloud


packages/framework/src/Model/Complaint/ComplaintResolutionEnum.php
```diff
@@ -24,14 +25,26 @@ public function getAllIndexedByTranslations(): array
         ];
     }
 
+    /**
+     * @return array<string, string>
+     */
+    public function getAllIndexedByTranslationsForCustomer(): array
+    {
+        return [
+            t('Fix', domain: Translator::CUSTOMER_TRANSLATION_DOMAIN) => static::FIX,
```
vitek-rostislav, line 34: I am wondering, have you considered the possibility of adding multiple translation domains at once in one `t` function call)? It does not make sense for the translation itself as the translator would not know which translation to use then, right? I am just not happy about the current necessity for the multiple definitions...However, maybe it is just in this file that obvious. This is just the beginning of the review, I will see how I feel about it during the process now :smile:

packages/framework/src/Resources/views/Mail/Order/addresses.html.twig
```diff
@@ -1,3 +1,4 @@
+{% trans_default_domain 'customerMessages' %}
 <table border="0" cellpadding="0" cellspacing="0" role="presentation" width="100%" style="table-layout: fixed; mso-table-lspace: 0pt; mso-table-rspace: 0pt;">
```
vitek-rostislav, line 2: nice, I did not know this

packages/framework/src/Component/Translation/FrontendApiDetectorTrait.php
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * @return string
+     */
+    protected function getTranslationDomain(): string
+    {
+        if ($this->isFrontendApiBundleConstraintClass) {
+            return Translator::CUSTOMER_VALIDATOR_TRANSLATION_DOMAIN;
+        }
```
vitek-rostislav: maybe we could come up with more specific name for the trait? From the current name, it is not obvious that CUSTOMER_VALIDATOR_TRANSLATION_DOMAIN will be used if in FE API...I am afraid it might be misued. What about `FrontendApiValidationTranslationDomainDetectorTrait`? Or something like that

project-base/app/patch/overblog-validator.patch
```diff
@@ -0,0 +1,11 @@
+--- ./vendor/overblog/graphql-bundle/src/Validator/InputValidator.php
++++ ./vendor/overblog/graphql-bundle/src/Validator/InputValidator.php
+@@ -118,7 +118,7 @@
+             // @phpstan-ignore-next-line (only for Symfony 4.4)
+             $builder
+                 ->setTranslator($this->defaultTranslator)
+-                ->setTranslationDomain('validators');
++                ->setTranslationDomain('customerValidators');
+         }
```
vitek-rostislav, line 9: We are not able to use the constant here, or?

project-base/app/translations/customerValidators.cs.po
```diff
@@ -0,0 +1,33 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: cs\n"
+
+msgid "An array of values is not supported as an input for \"slider\" type parameter. Use \"minimalValue\" and \"maximalValue\" instead."
+msgstr "Pole hodnot (\"values\") není podporováno jako vstup pro parametry typu \"slider\". Namísto toho použijte \"minimalValue\" and \"maximalValue\"."
```
vitek-rostislav: off topic, I would rather use parameters here to avoid mis-translation of "slider", "minimalValue", and "maximalValue"

project-base/app/translations/customerValidators.cs.po
```diff
@@ -0,0 +1,33 @@
...
+
+msgid "Email cannot be longer than {{ limit }} characters"
+msgstr "E-mail nesmí být delší než {{ limit }} znaků"
+
+msgid "Minimal and maximal value are not supported as an input for other than \"slider\" type parameter. Use \"values\" instead."
+msgstr "Minimální a maximalní hodnota nejsou podporovány jako vstup pro parametry jiného typu než \"slider\". Namísto toho použijte \"values\"."
```
vitek-rostislav: the same off topic here

composer.json
```diff
@@ -275,6 +277,11 @@
         },
         "runtime": {
             "dotenv_path": "project-base/app/.env"
+        },
+        "patches": {
+            "overblog/graphql-bundle": {
+                "Change default validator domain": "project-base/app/patch/overblog-validator.patch"
```
vitek-rostislav, line 282: should not be the patch in fe api package?

vitek-rostislav (changes requested):
Nice job there! I have added just a few notes :)

project-base/app/patch/overblog-validator.patch
grossmannmartin: As it's a code from vendor, we would have to import the class, or use FQCN. I'm not sure if it's better

composer.json
grossmannmartin: Just the patch file, or the patch file and composer.json entry?

Patch file itself may make sense. The composer entry - we still have to update composer file in project base and allow patching from vendor, so this seems more clear to me

packages/framework/src/Model/Complaint/ComplaintResolutionEnum.php
grossmannmartin: I know, this code does not look OK at the first glance. 
The ugly truth is we have multiple such duplications, but this one is pretty obvious. 
I don't think we can, more specifically "should" avoid the duplication. 

But the truth is, the statuses should be rather the database value and then it would make much more sense

packages/framework/src/Component/Translation/FrontendApiDetectorTrait.php
grossmannmartin: renamed in 9df02a2d34

project-base/app/translations/customerValidators.cs.po
grossmannmartin: what about 9b7b9c4eef?

project-base/app/translations/customerValidators.cs.po
grossmannmartin: what about 9b7b9c4eef?

composer.json
grossmannmartin: let's talk about it

vitek-rostislav (approved):

PR #3963, opened 2025-05-05
social login tweaks and fixes

vitek-rostislav:
- Removed unused dependencies
- added definition for front-login SF page route to avoid the path hardcoding in the controller - the page is also defined in the SF routes (see `project-base/storefront/config/routes.ts`) and hence it might be translated which should be then taken into account in the corresponding  `project-base/app/config/shopsys-routing/routing_front_XX.yaml` config files

#### Fixes issues
... 

----
  - https://rv-social-login-route.odin.shopsys.cloud
  - https://cz.rv-social-login-route.odin.shopsys.cloud

### DIFF
--- a/packages/frontend-api/src/Controller/SocialNetworkController.php
+++ b/packages/frontend-api/src/Controller/SocialNetworkController.php
@@ -114,14 +114,17 @@ class SocialNetworkController extends AbstractController
         $refererUrl = $refererUrl ?? $homepageUrl;
         $request->getSession()->remove(self::REFERER_URL);
         $refererUrl = str_replace($this->domain->getUrl(), '', $refererUrl);
-        $url = '/social-login?redirect=' . $refererUrl;
-        $url .= '&showCartMergeInfo=' . ($showCartMergeInfo ? 'true' : 'false');
-        $url .= '&isRegistration=' . ($isRegistration ? 'true' : 'false');
+        $parameters = [
+            'redirect' => $refererUrl,
+            'showCartMergeInfo' => $showCartMergeInfo ? 'true' : 'false',
+            'isRegistration' => $isRegistration ? 'true' : 'false',
+        ];
 
         if ($addExceptionMessage) {
-            $url .= '&exceptionType=socialNetworkLoginException&socialNetwork=' . $type;
+            $parameters['exceptionType'] = 'socialNetworkLoginException';
+            $parameters['socialNetwork'] = $type;
         }
 
-        return $url;
+        return $this->generateUrl('front_social_network_login_page', $parameters, UrlGeneratorInterface::RELATIVE_PATH);
     }
 }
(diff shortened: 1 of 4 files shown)


project-base/app/config/shopsys-routing/routing_front_sk.yaml
```diff
@@ -41,3 +41,6 @@ front_stores:
 
 front_complaint_detail:
     path: /zakaznik/detail-reklamace
+
```
grossmannmartin: please add upgrade notes :)

grossmannmartin (approved):

packages/frontend-api/src/Controller/SocialNetworkController.php
```diff
@@ -126,7 +114,7 @@ protected function getRefererUrl(
         $refererUrl = $refererUrl ?? $homepageUrl;
         $request->getSession()->remove(self::REFERER_URL);
         $refererUrl = str_replace($this->domain->getUrl(), '', $refererUrl);
-        $url = '/social-login?redirect=' . $refererUrl;
+        $url = sprintf('%s?redirect=%s', $this->generateUrl('front_social_network_login_page', referenceType: UrlGeneratorInterface::RELATIVE_PATH), $refererUrl);
```
pk16011990: Why don't we use generateUrl parameters for the URL instead of the sprintf? And it will be escaped correctly.

project-base/app/config/shopsys-routing/routing_front_cs.yaml
```diff
@@ -41,3 +41,6 @@ front_stores:
 
 front_complaint_detail:
     path: /zakaznik/detail-reklamace
+
+front_social_network_login_page:
+    path: /social-login
```
pk16011990: It looks like a forgotten translation. Why isn't it translated?

pk16011990 (changes requested):

project-base/app/config/shopsys-routing/routing_front_cs.yaml
vitek-rostislav: for some reason, all the SF routes need to be put in the `routes.ts`, even though they are not translated - see https://github.com/shopsys/shopsys/blob/17.0/project-base/storefront/config/routes.ts. This route does not need to be translated (it is just a SF "landing page" after social login is processed on the BE side) but it can be :shrug: I hope the duplication of the definitions will be solved in SSP-3404 after the app router is finished

packages/frontend-api/src/Controller/SocialNetworkController.php
vitek-rostislav: that is a good idea :pray: 14c125f5c1

project-base/app/config/shopsys-routing/routing_front_cs.yaml
vitek-rostislav: moved the route definition to `shopsys_front.yaml`

grossmannmartin (approved):

pk16011990 (approved):

PR #3966, opened 2025-05-06
little environment cleanup

vitek-rostislav:
Phing targets `build-new` and `build-demo` now switch the application to the production environment to avoid confusion when running locally. 

Moreover, Symfony `debug-bundle` and `web-profiler-bundle` packages were moved into the `dev` dependencies as they are not necessary for production.

#### Fixes issues
... 

----
  - https://rv-fix-bundles.odin.shopsys.cloud
  - https://cz.rv-fix-bundles.odin.shopsys.cloud


project-base/app/config/bundles.php
```diff
@@ -46,6 +46,5 @@
     Shopsys\ConvertimBundle\ShopsysConvertimBundle::class => ['all' => true],
     Becklyn\OrderedFormBundle\BecklynOrderedFormBundle::class => ['all' => true],
     Symfony\UX\Icons\UXIconsBundle::class => ['all' => true],
-    Symfony\Bundle\MakerBundle\MakerBundle::class => ['dev' => true],
```
grossmannmartin: ~I don't know about the upgrade notes 🙁~
~If someone follows the upgrade notes, then the maker PR tells them to add this line and then it stays there, or?~

I was thinking about it, and I think this change is not necessary. The main confusion was on my end by running commands without the preferred order.

Although we may improve the platform with the following changes: 
- add switch env type to production into the `build-new` and `build-demo` targets 
    - when you run these targets on development env, the composer-prod is invoked, but the app stays in the dev environment 
- move `symfony/web-profiler-bundle` and `symfony/debug-bundle` to the require-dev section

vitek-rostislav: reworked so this PR is now completely changed :smile:

grossmannmartin (approved):

pk16011990 (approved):

PR #3968, opened 2025-05-08
navigation item route

vitek-rostislav:
FE API now provides information about the friendly URL route name of the navigation item if available, which is then used on SF to display the proper skeleton loader for the navigation item link.

Refactored friendly URL routes definition from yaml to PHP file so the routes can be defined within new `FriendlyUrlRouteEnum` that is then used in the FE API as well.

#### Fixes issues
... 

----
  - https://rv-navigation-link-type.odin.shopsys.cloud
  - https://cz.rv-navigation-link-type.odin.shopsys.cloud


packages/framework/src/Model/Navigation/NavigationItemFacade.php
```diff
@@ -119,15 +127,18 @@ protected function fixUrlInNavigationItemData(NavigationItemData $navigationItem
...
-        if (strpos($navigationItemData->url, 'http') === 0) {
+        $domainUrl = $this->domain->getDomainConfigById($navigationItemData->domainId)->getUrl();
+        $navigationItemData->url = str_replace($domainUrl, '', $navigationItemData->url);
+
+        if (str_starts_with($navigationItemData->url, 'http')) {
             return;
         }
 
-        if (strpos($navigationItemData->url, 'www') === 0) {
+        if (str_starts_with($navigationItemData->url, 'www')) {
```
grossmannmartin: about http and www check - what if I create and use a category named `WWW Tools and libraries`?

project-base/storefront/types/friendlyUrl.ts
```diff
@@ -20,17 +21,17 @@ export type FriendlyUrlPageType =
...
     stores: 'front_stores',
-    store: 'front_stores_detail',
-    flag: 'front_flag_detail',
-    seo_category: 'front_category_seo',
+    store: TypeFriendlyUrlRouteEnum.FrontStoresDetail,
+    flag: TypeFriendlyUrlRouteEnum.FrontFlagDetail,
+    seo_category: TypeFriendlyUrlRouteEnum.FrontCategorySeo,
     complaintNew: 'front_customer_complaint_new',
     complaintList: 'front_customer_complaint_list',
     orderList: 'front_customer_order_list',
```
grossmannmartin, line 37: Does it make sense to add those routes to the enum, too?

grossmannmartin (changes requested):

project-base/storefront/types/friendlyUrl.ts
JanMolcik: I agree, why don't we put them all to the `TypeFriendlyUrlRouteEnum` enum.

JanMolcik (approved):

project-base/storefront/types/friendlyUrl.ts
vitek-rostislav: Yeah, that would be nice, but actually, those are two types of routes:
1) friendly URL (covered by the new enum as they are already completely defined on the BE) - e.g. `/hello-kitty` for a product detail, `tv-audio` for a category, etc.
2) localized URL (not covered by the new enum at the moment) - `/new-complaint`/`nova-reklamace`, `/detail-objednavky`/`order-detail`, etc. -  many of such routes are defined on SF only while some of them are defined on both sides

I would love to unify it, ideally having just one place with their definitions (BE?) and an enum that covers all the routes (or more probably, two enums). I have created a user story for that (see SSP-3404) because with the app router, this section is being rewritten and it makes sense to postpone further refactorings regarding the routes.

packages/framework/src/Model/Navigation/NavigationItemFacade.php
vitek-rostislav: changed the check to `www.` :slightly_smiling_face:

grossmannmartin (approved):

JanMolcik (approved):

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuContent.tsx
```diff
@@ -99,7 +104,7 @@ export const MobileMenuContent: FC<MobileMenuContentProps> = ({ navigationItems,
             />
 
             <MenuItems
-                className="translate-x-[120%]"
+                className={twJoin('absolute top-16')}
```
JanMolcik: No need for the `twJoin` here.

JanMolcik (approved):

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuContent.tsx
chlebektomas: fixed

PR #3970, opened 2025-05-08
DISABLE_FORM_FIELDS_FROM_TRANSFER now does not affect superadmin users

vitek-rostislav:
Now, we are able to allow form fields disabling on all environments while keeping the fields always enabled for superadmin. Hence, testers are able to ensure the desired fields are indeed disabled for regular administrators and still test the features even without access to an external system when logged as superadmin.

#### Fixes issues
... 

----
  - https://rv-disable-form-fields.odin.shopsys.cloud
  - https://cz.rv-disable-form-fields.odin.shopsys.cloud

### DIFF
--- a/docs/administration/disabling-form-fields.md
+++ b/docs/administration/disabling-form-fields.md
@@ -1,13 +1,55 @@
 # Disabling form fields
 
-During project implementation, there is usually necessary to have imported some fields from other systems.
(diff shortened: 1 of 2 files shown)


grossmannmartin (approved):

vitek-rostislav:
We will solve this more systematically using optional fields lock

PR #3985, opened 2025-05-16
administrator without filtered domains will see new domains automatically when added

TomasLudvik:

----
  - https://tl-fix-admin-without-new-domains.odin.shopsys.cloud
  - https://cz.tl-fix-admin-without-new-domains.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Controller/Admin/SelectAdminDomainsController.php
+++ b/packages/framework/src/Controller/Admin/SelectAdminDomainsController.php
@@ -58,7 +58,7 @@ class SelectAdminDomainsController extends AdminBaseController
             if (count($data) === 0) {
                 $this->addErrorFlash(t('Please select at least one domain.'));
             } else {
-                if (array_values($data) === $this->domain->getAllIds()) {
+                if (count($data) === count($this->domain->getAllIds())) {
                     $data = [];
                 }
 
--- a/packages/framework/src/Controller/Admin/SelectAdminDomainsController.php
+++ b/packages/framework/src/Controller/Admin/SelectAdminDomainsController.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Controller\Admin;
 
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Form\Admin\Administrator\AdminDomainsFormType;
 use Shopsys\FrameworkBundle\Model\Administrator\AdministratorDataFactory;
 use Shopsys\FrameworkBundle\Model\Administrator\AdministratorFacade;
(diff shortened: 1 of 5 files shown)


packages/framework/src/Model/Administrator/AdministratorDataFactory.php
```diff
@@ -33,7 +33,6 @@ public function create(): AdministratorData
     {
         $administratorData = $this->createInstance();
 
-        $administratorData->displayOnlyDomainIds = $this->domain->getAllIds();
```
grossmannmartin, line 36: domain dependency is no longer necessary

packages/framework/src/Model/Administrator/Administrator.php
```diff
@@ -203,7 +203,7 @@ public function __construct(AdministratorData $administratorData)
         $this->roles = new ArrayCollection();
         $this->transferIssuesLastSeenDateTime = $administratorData->transferIssuesLastSeenDateTime;
         $this->uuid = Uuid::uuid4()->toString();
-        $this->displayOnlyDomainIds = [];
+        $this->displayOnlyDomainIds = null;
```
grossmannmartin: this is null by default

packages/framework/src/Model/Administrator/AdministratorData.php
```diff
@@ -59,6 +59,6 @@ public function __construct()
         $this->roles[] = Roles::ROLE_ADMIN;
         $this->transferIssuesLastSeenDateTime = new DateTime('1970-01-01 00:00:00');
         $this->roleGroup = null;
-        $this->displayOnlyDomainIds = [1];
+        $this->displayOnlyDomainIds = null;
```
grossmannmartin: this is null by default

packages/framework/src/Migrations/Version20250516104642.php
```diff
@@ -0,0 +1,22 @@
...
+class Version20250516104642 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE administrators ALTER display_only_domain_ids DROP NOT NULL');
+        $this->sql('UPDATE administrators SET display_only_domain_ids = null');
```
grossmannmartin, line 20: I'm not sure about this - this will reset the domain filter for every administrator. Is it necessary?

packages/framework/src/Controller/Admin/SelectAdminDomainsController.php
```diff
@@ -57,6 +58,10 @@ public function filterDomainsAction(Request $request): Response
             if (count($data) === 0) {
                 $this->addErrorFlash(t('Please select at least one domain.'));
             } else {
+                if (array_values($data) === $this->domain->getAllIds()) {
```
grossmannmartin: Can we be sure the order will always be the same? 
Isn't comparing the count of items sufficient?

grossmannmartin (changes requested):

packages/framework/src/Migrations/Version20250516104642.php
TomasLudvik: Yes because there is old migration, that sets it to some value. This means that current administrators with set filter will have to set it once again. I do not think its a big issue.

PR #3998, opened 2025-05-21
ensure that GrapesJS content is always editable

TomasLudvik:
GrapesJS was not working until now, when there was only whitespace / empty html / null in body

----
  - https://tl-fix-empty-grapesjs-content.odin.shopsys.cloud
  - https://cz.tl-fix-empty-grapesjs-content.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Model/Mail/MailTemplateDataFactory.php
+++ b/packages/framework/src/Model/Mail/MailTemplateDataFactory.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Model\Mail;
 
+use Shopsys\FrameworkBundle\Component\Domain\AdminDomainTabsFacade;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
+use Shopsys\FrameworkBundle\Component\GrapesJs\EnsureCorrectGrapesJsFormatHelper;
 use Shopsys\FrameworkBundle\Component\UploadedFile\UploadedFileDataFactory;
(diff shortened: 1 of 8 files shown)


packages/framework/src/Model/Mail/MailTemplateDataFactory.php
```diff
@@ -88,4 +93,33 @@ public function createFromOrderStatuses(array $orderStatuses, array $mailTemplat
...
+     * @param int $domainId
+     * @return string
+     */
+    protected function ensureBodyIsInCorrectGrapeJsFormat(
+        ?string $body,
+        int $domainId = Domain::FIRST_DOMAIN_ID,
+    ): string {
+        $locale = $this->domain->getDomainConfigById($domainId)->getLocale();
+
+        if ($body === null || ctype_space($body) || strip_tags($body) === '') {
```
grossmannmartin: Certain content (like `<div>    </div>`, for example) can still cause trouble.

what about 
```suggestion
        if ($body === null || trim(strip_tags($body)) === '') {
```

grossmannmartin (approved):

packages/framework/src/Model/Mail/MailTemplateDataFactory.php
TomasLudvik: I have reworked it a bit more. I have removed box-content and added sizing for editable div in Grapes, so it is always visible. I have made those changes directly in history, as there is not much code and I find it easier to review again.

grossmannmartin (approved):

pk16011990 (approved):

PR #4003, opened 2025-05-23
ensure that /resolve-friendly-url is accessible only on internal networks

TomasLudvik:
This URL is for backend resolving of URLs of Storefront only

----
  - https://tl-secure-resolve-friendly-url.odin.shopsys.cloud
  - https://cz.tl-secure-resolve-friendly-url.odin.shopsys.cloud

### DIFF
--- a/docker/nginx/nginx.conf
+++ b/docker/nginx/nginx.conf
@@ -42,6 +42,12 @@ server {
     }
 
     location ~ ^/resolve-friendly-url {
+        allow 10.0.0.0/8;
+        allow 127.0.0.0/8;
+        allow 172.16.0.0/12;
+        allow 192.168.0.0/16;
+        deny all;
+
         fastcgi_pass php-upstream;
         include fastcgi_params;
         fastcgi_param SCRIPT_FILENAME $realpath_root/resolveFriendlyUrl.php;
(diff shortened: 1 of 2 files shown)


docker/nginx/nginx.conf
```diff
@@ -42,6 +42,12 @@ server {
     }
 
     location ~ ^/resolve-friendly-url {
+        allow 10.0.0.0/8;
+        allow 127.0.0.0/8;
+        allow 172.0.0.0/8;
+        allow 192.0.0.0/8;
```
henzigo: Correct CIDR for those ranges are:

```
172.16.0.0/12
192.168.0.0/16
```

Furthermore, loopback IP (`127.0.0.0/8`) is useless, since requests never come from the nginx container itself

henzigo (changes requested):

docker/nginx/nginx.conf
TomasLudvik: I had my storefront under 172.17.x.x address, thats why I used 172.0.0.0/8, also I used 127... for people that are running it in native / hybrid mode.

henzigo (approved):

pk16011990 (approved):

TomasLudvik:
Closing this as we decided, that only production is the place when we need to limit this and that is done in https://github.com/shopsys/deployment/pull/52

PR #4008, opened 2025-05-27
DatabaseSearchingHelper::getFullTextLikeSearchString argument is not nullable anymore

vitek-rostislav:
Inside, the method calls `getLikeSearchString` that also has not-nullable argument :shrug: 

Furthermore, removed some unused methods from `CategoryRepository` and `CategoryFacade` classes.

#### Fixes issues
closes #3473 

----
  - https://rv-db-search.odin.shopsys.cloud
  - https://cz.rv-db-search.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/String/DatabaseSearchingHelper.php
+++ b/packages/framework/src/Component/String/DatabaseSearchingHelper.php
@@ -24,10 +24,10 @@ class DatabaseSearchingHelper
     }
 
     /**
-     * @param string|null $string
+     * @param string $string
      * @return string
      */
-    public function getFullTextLikeSearchString(?string $string): string
+    public function getFullTextLikeSearchString(string $string): string
     {
         return '%' . self::getLikeSearchString($string) . '%';
     }
--- a/packages/framework/src/Model/Category/CategoryFacade.php
+++ b/packages/framework/src/Model/Category/CategoryFacade.php
@@ -307,21 +307,6 @@ class CategoryFacade
     }
 
     /**
-     * @param int $domainId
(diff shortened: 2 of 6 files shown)


packages/framework/src/Component/String/DatabaseSearchingHelper.php
```diff
@@ -24,10 +24,10 @@ public static function getLikeSearchString(string $string): string
     }
 
     /**
-     * @param string|null $string
+     * @param string $string
      * @return string
      */
-    public function getFullTextLikeSearchString(?string $string): string
+    public function getFullTextLikeSearchString(string $string): string
```
grossmannmartin, line 30: I think the problem is much deeper. for example if you take a look at `Shopsys\FrontendApiBundle\Model\Store\StoreRepository::getBasicFilteredQueryBuilder()` 
where we call 
`->setParameter('searchText', $this->databaseSearchingHelper->getFullTextLikeSearchString($storesFilterOptions->getSearchText()));` but the getSearchText may return null... 

What about flipping the logic and making `getLikeSearchString` accept null and internally convert to an empty string if necessary?

grossmannmartin (changes requested):

packages/framework/src/Component/String/DatabaseSearchingHelper.php
vitek-rostislav: actually, all the usages are "secured" by `if $string !== null` conditions, so :shrug:

grossmannmartin (approved):

PR #4017, opened 2025-06-01
[CRUD] Delete action

### LINKED ISSUE #2379: annotation fixer should take into account parent class
## Describe the bug
Newly created class with the same name (and appropriate namespace) as some from framework forces annotation-fixer to add annotations from supposedly parent class.

### To Reproduce

1. Create new class `\App\Model\Product\ProductFacade` without extending the Framework's ProductFacade
2. Introduce some changes that should add new annotations to this class if it was extended 
3. Run `./phing annotations-fix`
4. See added annotations that does not make sense in this context

henzigo:

Implementation of delete action for CRUD controller. Updated TestController for testing can be [found here](https://github.com/shopsys/shopsys/commit/7c1bf35e83027ab11c2b7ca1153e8fc6b62a1345)

#### Fixes issues
closes #2379

----
  - https://jg-crud-delete.odin.shopsys.cloud
  - https://cz.jg-crud-delete.odin.shopsys.cloud
  - https://jg-crud-delete.odin.shopsys.cloud/sk

----
  - https://jg-crud-delete.odin.shopsys.cloud
  - https://cz.jg-crud-delete.odin.shopsys.cloud
  - https://jg-crud-delete.odin.shopsys.cloud/sk


docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
```diff
@@ -0,0 +1,115 @@
...
+    // Return null to proceed with deletion
+    return null;
+}
+```
+
+The `beforeDelete` method:
+- Receives the entity object that's about to be deleted
+- Can return a `RedirectResponse` to cancel the deletion and redirect the user
+- Should return `null` to proceed with the deletion
+
```
vitek-rostislav: here should be some "Read more" headline?

docs/administration/crud-controller/reference/crud-providers.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### 3. Use Translations
+
+```php
+public function __toString(): string
+{
+    // Use the t() function for translatable strings
+    return t('Category "%name%" in %locale%', [
+        '%name%' => $this->getName(),
+        '%locale%' => $this->getLocale(),
```
vitek-rostislav: this might be misleading - usually, there is no locale in the entity class

docs/administration/crud-controller/reference/crud-providers.md
```diff
@@ -0,0 +1,251 @@
+# CRUD Providers
```
vitek-rostislav: it seems like the `reference/crud-providers.md` and `getting-starting/adding-create-edit-and-delete-actions.md` pages are kind of duplicating themselves. Is not that redundant? What is the purpose of splitting the information into two pages?

docs/administration/crud-controller/reference/crud-controller.md
```diff
@@ -21,20 +25,95 @@ Crud controller generates pretty URLs for each action. The URL is generated base
...
-| `create`   | `admin_crud_{controller_name}_create` | `/admin/{controller-name}/create`     |
-| `edit`     | `admin_crud_{controller_name}_edit`   | `/admin/{controller-name}/edit`       |
-| `delete`   | `admin_crud_{controller_name}_delete` | `/admin/{controller-name}/delete`     |
+| ActionType | Route name                            | Route path                                   |
+|------------|---------------------------------------|----------------------------------------------|
+| `list`     | `admin_crud_{controller_name}_list`   | `/admin/{controller-name}/`                  |
+| `detail`   | `admin_crud_{controller_name}_detail` | `/admin/{controller-name}/{entityId}/detail` |
+| `create`   | `admin_crud_{controller_name}_create` | `/admin/{controller-name}/create`            |
+| `edit`     | `admin_crud_{controller_name}_edit`   | `/admin/{controller-name}/{entityId}/edit`   |
+| `delete`   | `admin_crud_{controller_name}_delete` | `/admin/{controller-name}/{entityId}/delete` |
```
vitek-rostislav: should not it be inverted? I mean, I would expect the entityID to be the last part of the path

docs/administration/crud-controller/reference/crud-controller.md
```diff
@@ -21,20 +25,95 @@ Crud controller generates pretty URLs for each action. The URL is generated base
...
+```php
+protected function configureQuery(QueryBuilder $queryBuilder): void
+{
+    $queryBuilder
+        ->andWhere('o.deleted = :deleted')
+        ->setParameter('deleted', false);
+}
+```
+
+#### `beforeDelete(object $entity): ?RedirectResponse`
```
vitek-rostislav: Generally, I agree with the idea and understand the need for something like this, however, I am not sure whether I like the approach (or at least, the examples). I feel like all the checks performed in the `beforeDelete` methods should actually be done in facades which throw exceptions in such cases. So IMHO we should be able to react to those exceptions rather than duplicating the logic in `beforeDelete` methods. Let's talk about it, please

vitek-rostislav: Maybe, we do not need such a method at all, we can _just move_ the current `try-catches` from the controllers to the new `CrudProvider` classes?

packages/administration/src/Component/Crud/CrudProviderInterface.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Crud;
+
+/**
+ * Interface for CRUD controllers that need custom business logic
```
vitek-rostislav: I am not sure about the comment, I would remove it, or update it to be more accurate

packages/administration/src/Component/Crud/CrudProviderInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+interface CrudProviderInterface
+{
+    /**
+     * Get entity by ID
```
vitek-rostislav: this seems unnecessary, or?

packages/administration/src/Component/Crud/CrudProviderInterface.php
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * @param int $id
+     * @return object
+     */
+    public function getById(int $id): object;
+
+    /**
+     * Delete an entity by its object
```
vitek-rostislav: this seems unnecessary, or?

packages/administration/src/Component/Registry/CrudControllerDefinitionRegistry.php
```diff
@@ -19,22 +19,27 @@ final class CrudControllerDefinitionRegistry
 
     /**
      * @param \Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver $entityNameResolver
-     * @param array<int, array{class: class-string<\Shopsys\AdministrationBundle\Controller\AbstractCrudController>, entityClass: string}> $crudControllers
+     * @param array<int, array{class: class-string<\Shopsys\AdministrationBundle\Controller\AbstractCrudController>, entityClass: class-string, controlClass?: class-string|null}> $crudControllers
```
vitek-rostislav: it should be `providerClass`, right?

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -128,11 +150,40 @@ public function createAction(): Response
...
+        }
+
+        $crudProvider->delete($entity);
+
+        $this->addSuccessFlashTwig(
+            t('<strong>{{ objectName }}</strong> was deleted successfully.'),
+            [
+                'objectName' => ObjectNameHelper::getObjectName($entity),
+            ],
+        );
```
vitek-rostislav: I would like to discuss the flash messages, on one hand, I like that there is no duplication of the success message necessary, on the other hand, I should handle the failure flash message in my `CrudProviderInterface` implementation, right? It might be confusing - which messages should be handled by myself and which are handled automatically

packages/administration/src/Component/Registry/CrudProviderRegistry.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Registry;
+
+use InvalidArgumentException;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+final class CrudProviderRegistry
```
vitek-rostislav: why do we use the registry with container + has/set methods, instead of using iterator for all the services implementing CrudProviderInterface? You would not have to make them public in `RegisterCrudProvidersCompilerPass` then. Moreover, in the current implementation, the `shopsys.admin.crud_provider` is unused

packages/administration/src/Controller/AbstractCrudController.php
vitek-rostislav: maybe we can add something like `CrudProviderInterface::getDeleteSuccessMessage(): ?string` and use the generic message only if that returns null? :thinking:

packages/framework/src/Component/Router/Security/RouteCsrfProtector.php
```diff
@@ -7,18 +7,24 @@
...
-class RouteCsrfProtector implements EventSubscriberInterface
+class RouteCsrfProtector implements EventSubscriberInterface, ResettableInterface
 {
     public const CSRF_TOKEN_REQUEST_PARAMETER = 'routeCsrfToken';
     public const CSRF_TOKEN_ID_PREFIX = 'route_';
 
+    /**
+     * @var array<string, array<string, bool>>
+     */
+    protected array $checkedRoutes = [];
```
vitek-rostislav: you could/should use `InMemoryCache` here, or?

vitek-rostislav (changes requested):

docs/administration/crud-controller/reference/crud-providers.md
henzigo: 👍

docs/administration/crud-controller/reference/crud-controller.md
henzigo: Removed

packages/administration/src/Component/Crud/CrudProviderInterface.php
henzigo: removed

packages/administration/src/Component/Crud/CrudProviderInterface.php
henzigo: removed

packages/administration/src/Component/Crud/CrudProviderInterface.php
henzigo: removed

packages/administration/src/Component/Registry/CrudControllerDefinitionRegistry.php
henzigo: you are right. Fixed

packages/framework/src/Component/Router/Security/RouteCsrfProtector.php
henzigo: You are right. Implemented

packages/administration/src/Component/Registry/CrudProviderRegistry.php
henzigo: Dont ask ... I dont know :D ... changed

docs/administration/crud-controller/reference/crud-controller.md
henzigo: Based on AI, probably not. This should be the correct way to build URLs, especially when composing child controllers or custom actions. Like:

The playlist has videos:

Currently: 
- /playlist/{id}/edit
- /playlist/{id}/video/{id}/edit

VS:
- /playlist/edit/{id}
- /playlist/{id}/video/edit/{id}

The same pattern is defined as best practice for REST API

docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
henzigo: this page is going to be edited a lot so I'd probably leave it like this for now

packages/administration/src/Controller/AbstractCrudController.php
henzigo: I couldn't find any benefit in your suggestion. If you have a specific example where this would help, please post it.

henzigo: However, I have encountered one shortcoming and that is error handling. Now there is a try catch, where at the moment of error a general message is printed and a message with a bunch of information is logged for better debug

docs/administration/crud-controller/reference/crud-providers.md
henzigo: The getting started section should contain only simple and must be steps to get it working. On the other hand, a reference should contain additional information and some edge cases, and like everything possible with it.

docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
vitek-rostislav: well there is a list of links without any headline, it feels a little bit broken :shrug:

docs/administration/crud-controller/reference/crud-providers.md
vitek-rostislav: yeah, and still, both pages describe very similar things, could you take a look at it, please? We can talk about it if you want

docs/administration/crud-controller/reference/crud-controller.md
vitek-rostislav: I do not like the fact we will be changing almost all the paths in the admin. This should be discussed with PO. From my point of view, this is an unnecessary change that will piss off a lot of people :smile:

packages/administration/src/Controller/AbstractCrudController.php
vitek-rostislav: ok, now it is unified, all the flash messages are handled generically by `AbstractCrudController` hence the developer does not need to bother with them. However, I still think there might be a scenario where a developer will need to somehow overwrite the generic messages though. If not for the deletion, surely they will need that for a creation, e.g. now we have this message after the admin is created:
```
$this->addSuccessFlashTwig(
                t('Administrator <strong><a href="{{ url }}">{{ name }}</a></strong> created. A link to set a password has been sent to his email.'),
```

So basically, I think providing a way how to define custom success/failure messages if necessary would be handy (and easy to implement :wink: :wink: :smile: )

henzigo: What about just adding another flash message instead of modifying the default one?

docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
```diff
@@ -0,0 +1,93 @@
...
+use Shopsys\AdministrationBundle\Component\Attributes\CrudController;
+use Shopsys\AdministrationBundle\Controller\AbstractCrudController;
+
+#[CrudController(entityClass: Order::class)]
+class OrderController extends AbstractCrudController
+{
+    public function configure(CrudConfig $config): void
+    {
+        $config
+            ->registerHandler(OrderDeleteHandler::class)
```
vitek-rostislav, line 60: I have one doubt that I need to share - I am afraid that the new system will be in the end hard to understand, not so much for creating new stuff, but for changing the behavior of the framework-defined stuff. Maybe it will become clearer to me later, however, at the moment, I am really afraid of overcomplicating things. Do you think it will be clear to project developers when to use the "classic" inheritance approach, and when to follow other patterns? For example here - if we had CRUD OrderController in the framework, and there would be registered the OrderDeleteHandler this way. On a project, if I needed to modify the behavior of OrderDeleteHandler, how would I do that? Or should we believe this won't be a use case? I can imagine the handlers to work basically just as bridges for facades which can be still extended the old-fashion way. But if this is the case, let's rather mark the handlers final (:smile:) in the framework? Maybe the comment is too confusing, let's discuss it, please

packages/administration/src/Component/Config/CrudConfig.php
```diff
@@ -211,6 +221,47 @@ public function setCustomRoleSection(string $roleSection): self
...
+            if (count($actionTypes) === 0) {
+                throw new InvalidArgumentException(sprintf(
+                    'Handler class "%s" does not correspond to any CRUD action.',
+                    $handlerClass,
+                ));
+            }
+
+            foreach ($actionTypes as $actionType) {
+                if ($this->handlerClasses[$actionType->value] !== null) {
+                    throw new RuntimeException(sprintf(
```
vitek-rostislav, line 239: follow up to my previous comment - so if I understand it correctly, I am not able define my own handler that could replace the one defined in the framework?

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,202 @@
...
+- **One handler per action** - Each action can have only one handler registered and cannot be changed once set
+- **No mixing handlers** - If you register a handler implementing `CrudHandlerInterface`, you cannot register any other handlers (it claims all actions)
+- **Automatic service registration** - Handlers are automatically registered as services - no manual service configuration needed
+
+---
+
+## Entity Naming
+
+When working with CRUD handlers, it's important to ensure your entities have proper string representations. Entity names are displayed throughout the CRUD interface in:
+
```
vitek-rostislav, line 143: I am wondering, is there anything that forces me to do that? I will probably see a descriptive error sooner or later when I do not define that for my entity, right?

vitek-rostislav: ah, ok, I see it below, the object hash is used if not provided :+1:

packages/administration/src/Component/Config/CrudConfig.php
```diff
@@ -211,6 +221,47 @@ public function setCustomRoleSection(string $roleSection): self
         return $this;
     }
 
+    /**
+     * Register handler class or classes for CRUD actions.
+     *
+     * @param array<class-string<\Shopsys\AdministrationBundle\Component\Crud\Handler\HandlerInterface>> $handler
```
vitek-rostislav: should not here be `|string` as well?

packages/administration/src/Component/Config/CrudConfigData.php
```diff
@@ -29,7 +32,16 @@ public function __construct(
         private ?string $routePrefix,
         private ?string $customRoleConstant,
         private ?string $customRoleSection,
+        private array $handlerClasses,
     ) {
+        foreach ($this->enabledActions as $action) {
+            if (array_key_exists($action->value, $this->handlerClasses) && $this->handlerClasses[$action->value] === null) {
```
vitek-rostislav, line 30: is there a use case for keeping both properties? I am not able to enable action without the handler, so should not be the handler registration sufficient?

packages/framework/src/Component/HttpFoundation/Event/MasterTransactionRollbackEvent.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\HttpFoundation\Event;
+
+final class MasterTransactionRollbackEvent
```
vitek-rostislav: should not it extend `Event` class? :)

packages/administration/src/Component/Crud/Definition.php
```diff
@@ -24,15 +29,28 @@ public function __construct(
...
-     * @return \Shopsys\AdministrationBundle\Controller\AbstractCrudControllerExtension[]
+     * @template T of object
+     * @param class-string<T>|null $interface
+     * @return ($interface is null
+     *     ? array<\Shopsys\AdministrationBundle\Controller\AbstractCrudControllerExtension>
+     *     : array<T&\Shopsys\AdministrationBundle\Controller\AbstractCrudControllerExtension&\Shopsys\AdministrationBundle\Component\Crud\Extension\CrudHookableExtensionInterface>
+     * )
      */
-    public function getExtensions(): array
+    public function getExtensions(?string $interface = null): array
```
vitek-rostislav: maybe this should be called `$hookableInterface` and annotated with `CrudHookableExtensionInterface` class-string?

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -137,11 +203,21 @@ final protected function getConfiguredActions(ActionType $actionType): array
...
-        }
+        $this->executeExtensions(fn (AbstractCrudControllerExtension $extension) => $extension->configureActions($actionsConfig));
 
         return $actionsConfig->getActions($actionType);
     }
+
+    /**
+     * @param \Closure $callback
+     * @param string|null $hookableInterface
+     */
```
vitek-rostislav, line 177: can `CrudHookableExtensionInterface` class-string be annotated here as well?

packages/framework/src/Component/FlashMessage/FlashMessageTrait.php
```diff
@@ -110,6 +110,16 @@ public function isFlashMessageBagEmpty(): bool
             && !$flashBag->has(FlashMessage::KEY_WARNING);
     }
 
+    /**
+     * @return bool
+     */
+    public function hasErrorMessages(): bool
+    {
+        $flashBag = $this->getSession()->getFlashBag();
```
vitek-rostislav:
```suggestion
return $this->getSession()->getFlashBag()->has(FlashMessage::KEY_ERROR);
```

packages/administration/src/Component/Action/RowAction.php
```diff
@@ -89,6 +114,18 @@ public function validate(mixed $data): bool
             throw new InvalidArgumentException('Route must be set for row action. Use one of the "linkTo*" methods.');
         }
 
+        // Evaluate disabled message callback and modify action state if disabled
+        if ($this->disabledMessageCallback !== null) {
```
vitek-rostislav: IMHO the logic does not fit into the `validate()` method :thinking: I am wondering, where to put it. I thought about `create` but that would not work because it is static, right? We could rename the method to something like finalizeConfiguration() or something like that, but more I think about it, the it feels like a code smell. See https://blog.ploeh.dk/2011/05/24/DesignSmellTemporalCoupling/ Let's talk about it, please

packages/framework/src/Component/ClassExtension/ClassExtensionRegistry.php
```diff
@@ -75,9 +76,17 @@ protected function getOtherClassesExtensionMap(): array
...
                 $projectClassFqcn = str_replace($package['namespace'], $package['app_namespace'], $packageClassFqcn);
 
-                if (class_exists($projectClassFqcn)) {
-                    $otherClassesMap[$packageClassFqcn] = $projectClassFqcn;
+                if (class_exists($projectClassFqcn) === false) {
+                    continue;
+                }
+
+                $projectReflectionClass = ReflectionObject::createFromName($projectClassFqcn);
+
```
vitek-rostislav: it feels like we do not need the reflection here, we can leverage `is_subclass_of` or, if you wish to be strict about the direct parent -> child relation, `get_parent_class`. Or am I missing something?

vitek-rostislav (changes requested):

docs/administration/crud-controller/reference/handlers.md
henzigo: Yes :)

packages/administration/src/Component/Action/RowAction.php
henzigo: I refactored it so now its more encapsulated. What do you think?

packages/administration/src/Component/Config/CrudConfig.php
henzigo: Fixed

packages/administration/src/Component/Config/CrudConfig.php
henzigo: You still can by using services.yaml. But I would not recommend it.

```yaml
Shopsys\FrameworkBundle\Model\Order\OrderCrudHandler:
        class: App\Model\Order\OrderCrudHandler
        tags: [ 'shopsys.admin.crud_handler' ]
```

We should keep handlers as simple as possible, and instead of overriding them, we should decorate those actions with Hooks

packages/administration/src/Component/Config/CrudConfigData.php
henzigo: Yes, as you may want to disable some action even though there is already a registered handler.

packages/administration/src/Component/Crud/Definition.php
henzigo: fixed

packages/administration/src/Controller/AbstractCrudController.php
henzigo: fixed

packages/framework/src/Component/FlashMessage/FlashMessageTrait.php
henzigo: fixed

packages/framework/src/Component/HttpFoundation/Event/MasterTransactionRollbackEvent.php
henzigo: fixed

packages/framework/src/Component/ClassExtension/ClassExtensionRegistry.php
henzigo: ok refactored

docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
henzigo: > if I needed to modify the behavior of OrderDeleteHandler, how would I do that? Or should we believe this won't be a use case?

You should keep the handler actions as simple as possible and use decoration/events to build other features around it.

> I can imagine the handlers to work basically just as bridges for facades which can be still extended the old-fashion way.

Yes, for now, it's the simplest way we can easily implement it.

> Do you think it will be clear to project developers when to use the "classic" inheritance approach, and when to follow other patterns?

We should force (a little) developers to write better code. This solution will force them to think about how to write the feature a little better. But you can still bypass it, as I wrote in the next comment, or you can extend facades, etc.

packages/administration/src/Component/Config/CrudConfigData.php
vitek-rostislav: yes, but we can still allow actions to be disabled while not forcing users to define both actions and handlers for them :slightly_smiling_face:

docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
vitek-rostislav: Ok, my concern about the system being complicated at the moment is still here. If the developers need to change the given behavior, they might be confused about how to achieve that. I do not think we can solve it effectively now as there are too many "ifs". Maybe this particular part of code will not be a problem at all.

packages/administration/src/Component/Config/CrudConfig.php
vitek-rostislav: Yes, that is really not the way you should do it. So, when I put it another way:
1) Why don't we autowire the handler into the controller and use `->registerHandler($this->orderCrudHandler)` instead?
2) Why do we need to restrict the reconfiguration of the handlers? I mean, at the moment, we are able to reconfigure everything in `CrudControllerExtension`, except for the handler setting :shrug:

packages/administration/src/Component/Config/CrudConfigData.php
vitek-rostislav: Maybe we can reopen the topic once all the handlers are implemented

packages/administration/src/Component/Action/RowAction.php
vitek-rostislav: I think it is better now, however, the problem of temporal coupling is still there. I am wondering, what about this concept:
- create a configuration object that would just hold the raw config data (renderTooltip, isDisabled, disabledMessageCallback, ...)
- add a factory that would create the `RowAction` instance from the configuration and the actual row data

But maybe this would be overcomplicated.

I am still worried about the necessity of the function calls. If you create an action and do not call `prepareAction`, you can not be sure it is in a valid state. Btw, I tried asking Claude:
```
Code Smell Analysis

  The Problem: State Mutation in prepareAction()

  In RowAction::prepareAction() (lines 111-130), there's a problematic pattern:

  if ($disabledMessage !== null) {
      $this->isDisabled = true;
      $this->label = $disabledMessage;      // ← mutates state
      $this->confirmMessage = null;          // ← mutates state
      $this->renderTooltip = true;           // ← mutates state
  }

  This method has side effects - it modifies object state ($this->label, $this->isDisabled, etc.) within what is
  conceptually a "preparation" or "validation" step.

  Why It Works (Mostly)

  The second commit (0f264fd) introduced cloning in AbstractAction::build():

  public function build(mixed $data): array|null
  {
      $action = clone $this;  // ← Clone protects original from mutation

      if ($action->prepareAction($data) === false) {
          return null;
      }
      // ... uses cloned $action
  }

  This protects the original action from being permanently mutated between rows.

  The Remaining Code Smells

  1. Method name is misleading: prepareAction() suggests preparation/validation, but it actually mutates state. A boolean
   return with side effects violates the principle of least surprise.
  2. Mixed responsibility: The method does two things:
    - Decides whether to display (return bool)
    - Transforms state for disabled actions (mutation)
  3. Shallow clone risk: If RowAction ever contains nested objects or closures with mutable state, clone won't deep-copy
  them.
  4. Hidden coupling: The behavior depends on understanding that build() clones first. Someone calling prepareAction()
  directly (e.g., in tests or subclasses) would get permanent mutation.

  Potential Solutions

  Option A: Split into two methods
  protected function shouldDisplay(mixed $data): bool { ... }
  protected function configureForRow(mixed $data): void { ... }  // explicit mutation

  Option B: Return a DTO/value object instead of mutating
  protected function prepareAction(mixed $data): ?RowActionState
  {
      // Return immutable state object, or null to skip rendering
  }

  Option C: Keep current approach but rename
  protected function buildState(mixed $data): bool  // Name implies mutation is expected

  My Recommendation

  The current code works correctly due to the cloning, but it's fragile and confusing. The cleanest fix would be Option B
   - have prepareAction() return the computed state (disabled status, modified label, etc.) as an immutable value object
  rather than mutating $this. This makes the data flow explicit and eliminates the need for cloning.

```

TLDR, Claude suggests (and I agree) that we should have two separate objects - one for the initial configuration, another one providing the prepared object after `prepareAction` is called. What do you think?

packages/administration/assets/src/styles/custom.scss
```diff
@@ -15,3 +15,9 @@
     grid-template-columns: repeat(auto-fill, 17.5rem);
     column-gap: 1rem;
 }
+
+.link-disabled {
```
grossmannmartin, line 80: this mimics tabler's class `disabled`. I would not create duplicate classes to prevent confusion

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,264 @@
...
+### Interface Hierarchy
+
+Handlers use a hierarchical interface structure where each interface extends the previous one:
+
+```
+HandlerInterface (base)
+    └── ReadHandlerInterface (adds getById)
+        └── DeleteHandlerInterface (adds delete)
+            └── CrudHandlerInterface (marker for full CRUD)
```
grossmannmartin, line 21: This structure seems too complex. 
I don't understand the meaning of the Interface I should not implement (HandlerInterface). 
What about removing Read and move getbyid to handlerinterface

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,264 @@
...
+            ->registerHandler(OrderCrudHandler::class)
+        ;
+    }
+}
+```
+
+**Registration Rules:**
+
+- **Automatic action enabling** - Registering a handler automatically enables its corresponding actions in the CRUD controller (e.g., `DeleteHandlerInterface` enables Delete action)
+- **One handler per action** - Each action can have only one handler registered and cannot be changed once set
```
grossmannmartin: I don't understand both conditions... Why shouldn't I be able to register multiple handlers (composition) and why I should not be able to de/re-register handler? 

If I need to change a behavior, I can override the underlying Facade class and that's, in my opinion, more error-prone than letting user change the handler

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,264 @@
...
+- **No mixing handlers** - If you register a handler implementing `CrudHandlerInterface`, you cannot register any other handlers (it claims all actions)
```
grossmannmartin, line 73: Same as above. I find registering multiple handlers more understandable than hook.

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,264 @@
...
+- **Automatic service registration** - Handlers are automatically registered as services - no manual service configuration needed
```
grossmannmartin, line 74: Handler is a service, but we don't use it as a service. 
The coupling is hidden inside the implementation instead of the clear declaration of dependencies.
I would rather request handlers in the constructor

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,264 @@
...
+   ↓
+6. Redirect to list page
+
+--- On Exception ---
+
+1. Exception caught at any step above
+   ↓
+2. All on<Action>Error() hooks executed
+   ↓
+3. MasterTransactionRollbackEvent dispatched (database rollback)
```
grossmannmartin: What about other non-rollbackable steps? I mean when I dispatch message to the queue, send email, ... 
How to work with that?

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,264 @@
...
+When working with CRUD handlers, it's important to ensure your entities have proper string representations. Entity names are displayed throughout the CRUD interface in:
+
+- Flash messages (e.g., "Order #123 was deleted successfully")
+- Page titles and breadcrumbs
+- Error messages
+- Form labels
+
+### Implementing `__toString()` Method
+
+To provide user-friendly entity names, implement the `__toString()` method in your entity class:
```
grossmannmartin: I don't think implementing the __toString() method is the way to go. 
It mixes up the domain and the presentation layer.

packages/administration/src/Component/Action/RowAction.php
```diff
@@ -77,19 +102,31 @@ public static function create(string $name, string $label, ?string $icon = null)
...
             throw new InvalidArgumentException('Route must be set for row action. Use one of the "linkTo*" methods.');
         }
 
-        return parent::validate($data);
+        // Evaluate disabled message callback and modify action state if disabled
+        if ($this->disabledMessageCallback !== null) {
+            $disabledMessage = call_user_func($this->disabledMessageCallback, $data);
+
+            if ($disabledMessage !== null) {
+                $this->isDisabled = true;
```
grossmannmartin: why do we even need to modify the internal state? 

what about simple 
```php

$action->addCallback(function ($row, $action) {
    if ($row['number'] === '1765199941') {
        $action->disableWithMessage('Ah ah ah! You didn't say the magic word!');
    }
}
```

then we just adjust the rendering.

packages/administration/src/Component/Config/CrudConfigData.php
grossmannmartin: > Yes, as you may want to disable some action even though there is already a registered handler.

That could be also solved by allowing de-registering the handler. 
What is the point of registered delete handler with disabled delete action?

packages/administration/src/Component/Doctrine/Util/ObjectNameHelper.php
```diff
@@ -0,0 +1,25 @@
...
+     */
+    public static function getObjectName(object $object): string
+    {
+        if ($object instanceof Stringable) {
+            return (string)$object;
+        }
+
+        $className = get_class($object);
+
+        return sprintf('%s@%s', $className, spl_object_hash($object));
```
grossmannmartin: I don't think this is a good user-facing name. I would rather keep just a short class name without any identifier.

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -121,11 +139,59 @@ public function createAction(): Response
...
-     * @return \Symfony\Component\HttpFoundation\Response
+     * @return \Symfony\Component\HttpFoundation\RedirectResponse
      */
-    public function deleteAction(int $id): Response
+    #[CsrfProtection]
+    public function deleteAction(int $id): RedirectResponse
     {
-        return $this->redirect($this->generateUrl('admin_default_dashboard'));
+        /** @var \Shopsys\AdministrationBundle\Component\Crud\Handler\DeleteHandlerInterface $handler */
+        $handler = $this->definition->getHandlerForAction(ActionType::DELETE);
```
grossmannmartin, line 114:
```php
getDeleteActionHandler(): DeleteHandlerInterface {}
```

enables a typecheck instead of relying on the @var annotation

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -121,11 +139,59 @@ public function createAction(): Response
...
+        $entity = $handler->getById($id);
+        $objectName = ObjectNameHelper::getObjectName($entity);
```
grossmannmartin: basic handler interface can force the `getReadableName(): string` method. 
That way we separate the crud from the other logic and keep things along

packages/administration/src/Controller/AbstractCrudControllerExtension.php
```diff
@@ -8,8 +8,9 @@
 use Shopsys\AdministrationBundle\Component\Config\ActionsConfig;
 use Shopsys\AdministrationBundle\Component\Config\CrudConfig;
 use Shopsys\AdministrationBundle\Component\Datagrid\Datagrid;
+use Shopsys\FrameworkBundle\Controller\Admin\AdminBaseController;
 
-abstract class AbstractCrudControllerExtension
+abstract class AbstractCrudControllerExtension extends AdminBaseController
```
grossmannmartin, line 13: I don't understand why the Extension should extend adminbasecontroller. 
This seems unnecessary and error-prone

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -121,11 +139,59 @@ public function createAction(): Response
...
+
+        try {
+            $this->executeExtensions(fn (CrudDeleteHookExtensionInterface $extension) => $extension->beforeDelete($entity), CrudDeleteHookExtensionInterface::class);
+            $handler->delete($entity);
+            $this->executeExtensions(fn (CrudDeleteHookExtensionInterface $extension) => $extension->afterDelete($entity), CrudDeleteHookExtensionInterface::class);
+
+            if ($this->isFlashMessageBagEmpty()) {
```
grossmannmartin, line 122: A completely unrelated flash message may be added during the request. I don't think this is a good detection mechanism. 

We may have a separate flash store for crud and display only relevant here

packages/framework/src/Component/Grid/GridView.php
```diff
@@ -171,12 +171,13 @@ public function renderActionCell(ActionColumn|GridRowActionInterface $actionColu
             return;
         }
 
-        if ($actionColumn->validate($row) === false) {
+        $actionData = $actionColumn->build($row);
```
grossmannmartin, line 149: isn't build called twice? here and in the actions template?

packages/framework/src/Component/HttpFoundation/TransactionalMasterRequestListener.php
```diff
@@ -57,6 +57,11 @@ public function onKernelResponse(ResponseEvent $event): void
      * @param \Symfony\Component\HttpKernel\Event\ExceptionEvent $event
      */
     public function onKernelException(ExceptionEvent $event): void
+    {
+        $this->rollback();
+    }
+
+    public function rollback(): void
```
grossmannmartin: I don't think this should be public

packages/administration/assets/src/styles/custom.scss
henzigo: `disabled` also disable pointer events so tooltip on hover wont work

docs/administration/crud-controller/reference/handlers.md
henzigo: It doesn't make sense as for create action, you don't need to implement getById method

docs/administration/crud-controller/reference/handlers.md
henzigo: That's why there is `onDeleteError` method.

packages/administration/src/Component/Action/RowAction.php
henzigo: I like it.

packages/framework/src/Component/Grid/GridView.php
henzigo: No, that template is not used by Grid

packages/framework/src/Component/HttpFoundation/TransactionalMasterRequestListener.php
henzigo: This method is called by event, so it must be public

packages/administration/src/Component/Action/RowAction.php
henzigo: refactored

packages/administration/src/Component/Action/RowAction.php
henzigo: refactored a little bit

packages/administration/src/Component/Config/CrudConfig.php
henzigo: Added unregistering handlers

docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
henzigo: Added unregistering handlers

packages/administration/src/Controller/AbstractCrudController.php
henzigo: It looks tricky. Do not change it for now

packages/administration/src/Controller/AbstractCrudControllerExtension.php
henzigo: Extension is like any other controller. You can add flash message or custom action.

packages/administration/src/Component/Config/CrudConfig.php
```diff
@@ -211,6 +221,80 @@ public function setCustomRoleSection(string $roleSection): self
...
+                $this->handlerClasses[$actionType->value] = $handlerClass;
+                $this->enableAction($actionType);
+            }
+        }
+
+        return $this;
+    }
+
+    /**
+     * Unregister handler class or classes for CRUD actions.
```
vitek-rostislav: unnecessary and partly misleading comment (that "or classes" part, right?). Could you remove it, please?

packages/administration/src/Component/Config/CrudConfig.php
```diff
@@ -211,6 +221,80 @@ public function setCustomRoleSection(string $roleSection): self
...
+     * @param class-string<\Shopsys\AdministrationBundle\Component\Crud\Handler\HandlerInterface> $handler
+     * @return $this
+     */
+    public function unregisterHandler(string $handler): self
+    {
+        Assert::classExists($handler);
+        Assert::implementsInterface($handler, HandlerInterface::class);
+
+        $isHandlerRegistered = false;
+
```
vitek-rostislav, line 265: `$wasHandlerRegistered` would be more appropriate name IMHO, or even better `$handlerExists`, what do you think?

packages/administration/src/Component/Config/CrudConfig.php
```diff
@@ -211,6 +221,80 @@ public function setCustomRoleSection(string $roleSection): self
...
+    }
+
+    /**
+     * Unregister handler class or classes for CRUD actions.
+     *
+     * @param class-string<\Shopsys\AdministrationBundle\Component\Crud\Handler\HandlerInterface> $handler
+     * @return $this
+     */
+    public function unregisterHandler(string $handler): self
+    {
```
vitek-rostislav, line 260: it would be nice to add some "How to re-register a handler" section to docs :)

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -0,0 +1,264 @@
...
+            ->registerHandler(OrderCrudHandler::class)
+        ;
+    }
+}
+```
+
+**Registration Rules:**
+
+- **Automatic action enabling** - Registering a handler automatically enables its corresponding actions in the CRUD controller (e.g., `DeleteHandlerInterface` enables Delete action)
+- **One handler per action** - Each action can have only one handler registered
```
vitek-rostislav: ...maybe here, something like "if you want to define your own handler instead of an exisitng one, you need to unregister the original one using...etc"

packages/administration/src/Component/Action/RowAction.php
```diff
@@ -60,6 +70,40 @@ public function showTooltip(bool $render = true): self
...
+     *
+     * The callback receives row data and the action instance, allowing dynamic modification
+     * of the action based on the current row. Multiple callbacks can be added and will be
+     * executed in order.
+     *
+     * @param \Closure(mixed, self): void $callback Function receives row data and action instance
+     * @return self
+     */
+    public function addCallback(Closure $callback): self
+    {
```
vitek-rostislav, line 74: maybe `addBuildCallback()`, and `$buildCallbacks` for the property would be more clear names

docs/administration/crud-controller/getting-started/creating-a-new-crud-controller.md
```diff
@@ -14,32 +14,41 @@ declare(strict_types=1);
...
 ```php
 
 public function configure(CrudConfig $config): void
 {
     $config
-        ->setTitle(ActionType::LIST, t('My new Crud Controller')) // Set the title of the list page
+        ->setTitle(ActionType::LIST, t('Orders')) // Set the title of the list page
         ->setMenuSection('customers') // Set the menu section where the controller will be placed
```
vitek-rostislav, line 43: If I am not mistaken, there are constants for the menu entries, so we could use one here (`SideMenuBuilder::ROOT_CUSTOMER`)

vitek-rostislav: or at lease, add a note to the comment

docs/administration/crud-controller/reference/crud-controller.md
```diff
@@ -19,24 +23,84 @@ Crud controller generates pretty URLs for each action. The URL is generated base
...
+### `configure(CrudConfig $config): void`
 
+Configure general behavior of the controller. Customizable options are available [here](#crud-config).
+
+```php
+protected function configure(CrudConfig $config): void
+{
+    $config
+        ->setTitle(ActionType::LIST, t('Products'))
+        ->setMenuSection('products');
```
vitek-rostislav, line 50: same here, they can leverage a constant

packages/administration/src/Component/Crud/Handler/CrudHandlerInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Crud\Handler;
+
+/**
+ * Marker interface indicating full CRUD support
```
vitek-rostislav, line 8: maybe we can mark it as internal?

vitek-rostislav (changes requested):

packages/administration/src/Component/Crud/Handler/CrudHandlerInterface.php
henzigo: fixed

docs/administration/crud-controller/reference/crud-controller.md
henzigo: will not change in this PR

docs/administration/crud-controller/getting-started/creating-a-new-crud-controller.md
henzigo: will not change in this PR

packages/administration/src/Component/Action/RowAction.php
henzigo: Will not change. PRGs don't need to understand internal behaviour

docs/administration/crud-controller/reference/handlers.md
henzigo: edited

packages/administration/src/Component/Config/CrudConfig.php
henzigo: I don't want to add this as the recommended way.

packages/administration/src/Component/Config/CrudConfig.php
henzigo: fixed

packages/administration/src/Component/Config/CrudConfig.php
henzigo: fixed

docs/administration/crud-controller/reference/handlers.md
```diff
@@ -69,7 +69,7 @@ class OrderController extends AbstractCrudController
 **Registration Rules:**
 
 - **Automatic action enabling** - Registering a handler automatically enables its corresponding actions in the CRUD controller (e.g., `DeleteHandlerInterface` enables Delete action)
-- **One handler per action** - Each action can have only one handler registered
+- **One handler per action** - Each action can have only one handler registered. if you want to define your own handler instead of an existing one, you need to unregister the original one first.
```
vitek-rostislav:
```suggestion
- **One handler per action** - Each action can have only one handler registered. If you want to define your own handler instead of an existing one, you need to unregister the original one first.
```

vitek-rostislav (approved):

docs/administration/crud-controller/reference/handlers.md
vitek-rostislav: you still have that lowercase "if" 😉

vitek-rostislav (commented):
Topics that still need further discussion (I know we have discussed those, but it has been a long time since).

- [x] handlers re-registration is not easy, not documented, and claimed to be unwanted. So, at the moment, there is that "hidden" `unregisterHandler` way? Fair to say, there is a note in the docs about this (in `handlers.md`) but I feel it could be a little bit more descriptive (at least mention the proper method?)
- [x] I am not sure about that `__toString` for entities, it is quite controversial, and I do not remember we agreed to keep it as is
- [x] I agree with @grossmannmartin that we should improve the translationsto avoid "smazán(a)" constructs. I can imagine creating a separate task for that

packages/framework/src/Component/Utils/Presentable.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Utils;
```
vitek-rostislav, line 5: Why is the interface not in the `administration` package?

vitek-rostislav: (I can live with the current state, I am just curious)

vitek-rostislav (approved):

packages/framework/src/Component/Utils/Presentable.php
henzigo: It is intended to be a universal contract that can be used throughout the entire application

packages/administration/translations/messages.cs.po
```diff
@@ -148,6 +151,9 @@ msgstr "Množství"
 msgid "An authentication exception occurred."
 msgstr "Došlo k výjimce během přihlášení."
 
+msgid "An error occurred while deleting <strong>{{ objectName }}</strong>."
+msgstr "Při odstraňování položky „<strong>{{ objectName }}</strong>\" došlo k chybě."
```
grossmannmartin: Here and in sk translation, we should adjust the apostrophes.

grossmannmartin (approved):

PR #4030, opened 2025-06-09
mailer does not try to send message to non-existing email

malyMiso:
This PR solves problem with active mail whitelist. When mail whitelist is active and mail is not supposed to be sent, it sends mail to non-existing mail "no-reply@domain.tld". This is a problem, because mail cannot be delivered, therefore admin mail is spammed by Mail Delivery System with "Undelivered Mail Returned to Sender" for every mail address that meets whitelist criteria 

In https://github.com/symfony/symfony/pull/48409 was added feature to reject mail
... 

#### Fixes issues
... 

----
  - https://mt-mailer-reject.odin.shopsys.cloud
  - https://cz.mt-mailer-reject.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Model/Mail/EventListener/EnvelopeListener.php
+++ b/packages/framework/src/Model/Mail/EventListener/EnvelopeListener.php
@@ -9,7 +9,6 @@ use Shopsys\FrameworkBundle\Model\Mail\Email;
 use Shopsys\FrameworkBundle\Model\Mail\MailerSettingProvider;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Mailer\Event\MessageEvent;
-use Symfony\Component\Mime\Address;
 use Symfony\Component\Mime\Header\MailboxListHeader;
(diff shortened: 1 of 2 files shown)


TomasLudvik (approved):

packages/framework/src/Model/Mail/EventListener/EnvelopeListener.php
```diff
@@ -48,8 +47,7 @@ public function onMessage(MessageEvent $event): void
         }
 
         if ($allowedRecipients === []) {
-            // set a non-existing address because recipient list cannot be empty
-            $allowedRecipients = [new Address('no-reply@domain.tld')];
+            $event->reject();
```
vitek-rostislav, line 50: I think you need to add `return;` here, because `$event->getEnvelope()->setRecipients($allowedRecipients);` still fails (recipients can not be empty), see the failing tests please

vitek-rostislav (changes requested):

pk16011990 (approved):

PR #4034, opened 2025-06-11
make all form classes final as they are not extended directly but via extension

TomasLudvik:
Forms should not be extended directly but via extension

----
  - https://tl-make-forms-final-again.odin.shopsys.cloud
  - https://cz.tl-make-forms-final-again.odin.shopsys.cloud

### DIFF
--- a/docs/contributing/code-quality-principles.md
+++ b/docs/contributing/code-quality-principles.md
@@ -122,7 +122,7 @@ Also, it would be difficult to change the shared code just for one use-case.
 Modules don't depend on details of each other, modules depend on interfaces.
 
 ```diff
-class DatePickerType extends AbstractType
+final class DatePickerType extends AbstractType
 {
 -   public function __construct(protected readonly DisplayTimeZoneProvider $displayTimeZoneProvider)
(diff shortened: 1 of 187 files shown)


henzigo:
What do you think about creating a fixer for this? Look https://github.com/shopsys/shopsys/commit/b6dee0b3876da2d55ea8ae3166c397362eebbdb7

Another thing I dislike is forcing programmers to make all `FormTypes` final. I can imagine a large number of situations where `extends` would be useful. For example, I have an `AddressFormType` and I would like to create a `BillingAddressFormType` from it, which would have extra fields defined with company data. In this case, I would be unnecessarily copying the entire original FormType instead of just the extends.

So I like this approach for SSP programmers, but we should give freedom to those who build on this platform.

vitek-rostislav (changes requested):
I agree with @henzigo here, please add the fixer that would require `final` for all forms in packages, I would exclude forms in project-base from that though. 

Regarding that forms composition mentioned by Honza, I believe we can just skip the rule for such a form when the scenario occurs :shrug:

vitek-rostislav (approved):

PR #4047, opened 2025-06-17
replace jquery-ui with modern components

### LINKED ISSUE #152: Dates on the frontend are not localized

grossmannmartin:

This PR partially removes dependency on jQuery UI by replacing components with modern, lighter alternatives:

#### Datepicker

Replaced jquery-ui/datepicker with https://flatpickr.js.org/ - a lightweight, powerful datetime picker.

Thanks to this, we're able to render the datepicker in a proper locale if the administration locale is changed.
Also, usability on mobile has improved greatly.

#### Slider

Replaced jquery-ui/slider with https://refreshless.com/nouislider/ - a lightweight JavaScript range slider.

#### Sortable

Replaced jquery-ui/sortable with https://sortablejs.github.io/Sortable/ - a modern drag-and-drop library.
This is the future-compatible change, as the jquery-ui/slider does not work properly with elements placed in a flex grid, and we plan to use this in the reimagined admin theme.

All functionality has been preserved, and the user experience remains the same or improved.

Additional improvements

- Updated ESLint to version 8.57.1
- Updated codebase to follow new eslint rules.

#### Fixes issues
closes #152  

----
  - https://mg-remove-jquery-ui.odin.shopsys.cloud
  - https://cz.mg-remove-jquery-ui.odin.shopsys.cloud


packages/framework/assets/js/admin/components/GridInlineEdit.js
```diff
@@ -152,7 +148,6 @@ export default class GridInlineEdit {
                 (new Register()).registerNewContent($formRow);
                 $grid.find('.js-inline-edit-rows').prepend($formRow);
                 $formRow.find('input[type=text]:first').focus();
-                $grid.find('.js-drag-and-drop-grid-rows').sortable('disable');
```
vitek-rostislav, line 155: This was here to enable drag and drop for new items in the inline edit grid, right? AFAIK, we have no such grid at the moment. However, if we had, would it work?

vitek-rostislav (approved):

packages/framework/assets/js/admin/components/GridInlineEdit.js
grossmannmartin: Yes, I tested this manually, and the drag and drop with inline edit behaves the same :)

pk16011990 (approved):

PR #4054, opened 2025-06-24
OrderItem::hasProduct replaced with OrderItem::isTypeProductAndHasProduct

TomasLudvik:
Has method should be used for checking if value is set. Throwing error in such case does not make sense.

----
  - https://tl-fix-order-item.odin.shopsys.cloud
  - https://cz.tl-fix-order-item.odin.shopsys.cloud

### DIFF
--- a/packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
+++ b/packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
@@ -106,12 +106,11 @@ class OrderItemTest extends TestCase
         $orderItem->getProduct();
     }
 
-    public function testProductCannotHaveProduct(): void
+    public function testTransportDoesNotHaveProduct(): void
     {
         $orderItem = $this->createOrderTransport();
 
-        $this->expectException(WrongItemTypeException::class);
-        $orderItem->hasProduct();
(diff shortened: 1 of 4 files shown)


packages/framework/tests/Unit/Model/Order/Item/OrderItemTest.php
```diff
@@ -110,8 +110,7 @@ public function testProductCannotHaveProduct(): void
     {
         $orderItem = $this->createOrderTransport();
 
-        $this->expectException(WrongItemTypeException::class);
-        $orderItem->hasProduct();
+        $this->assertFalse($orderItem->isTypeProductAndHasProduct());
```
vitek-rostislav, line 113: It seems that this test method had a weird name. Could you please rename it to something like `testTransportCannotHaveProduct()`? :pray:

upgrade-notes/backend_20250723_143958.md
```diff
@@ -0,0 +1,3 @@
+#### OrderItem::hasProduct should not throw an exception ([#4054](https://github.com/shopsys/shopsys/pull/4054))
```
vitek-rostislav: Maybe the pull request (and its description) itself should be updated now, right?

vitek-rostislav (approved):

pk16011990 (approved):

PR #3084, opened 2024-03-28
[shopsys] refactored order creation

grossmannmartin:
| Q             | A
| ------------- | ---
|Description, reason for the PR| This PR completely changes how the order is created. Previously used OrderPreviewCalculation was replaced with more versatile middlewares. See documentation for more info
|New feature| Yes 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| Yes 
|Fixes issues| ... 
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

----
:globe_with_meridians: Live Preview:
  - https://mg-order-processor.odin.shopsys.cloud
  - https://cz.mg-order-processor.odin.shopsys.cloud

PR #4057, opened 2025-06-25
added ability to easily add new item to OrderData that will be persisted after OrderFacade::edit

TomasLudvik:
When developer wants to add new item via OrderFacade::edit, there was no easy way to do it. This PR finishes basic implementation from #3084 so it is easily available.

----
  - https://tl-fix-add-new-item.odin.shopsys.cloud
  - https://cz.tl-fix-add-new-item.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Model/Order/OrderData.php
+++ b/packages/framework/src/Model/Order/OrderData.php
@@ -313,7 +313,15 @@ class OrderData
      */
     public function addItem(OrderItemData $item): void
     {
-        $this->items[] = $item;
+        $count = 0;
+
+        foreach ($this->items as $key => $value) {
+            if (str_starts_with((string)$key, self::NEW_ITEM_PREFIX)) {
+                $count++;
+            }
+        }
+
+        $this->items[self::NEW_ITEM_PREFIX . $count] = $item;
     }
 
     /**
--- a/project-base/app/tests/App/Functional/Model/Order/OrderFacadeEditTest.php
+++ b/project-base/app/tests/App/Functional/Model/Order/OrderFacadeEditTest.php
@@ -14,7 +14,6 @@ use Override;
 use Shopsys\FrameworkBundle\Component\Money\Money;
 use Shopsys\FrameworkBundle\Model\Order\Item\Exception\OrderItemNotFoundException;
 use Shopsys\FrameworkBundle\Model\Order\Item\OrderItemTypeEnum;
-use Shopsys\FrameworkBundle\Model\Order\OrderData;
 use Shopsys\FrameworkBundle\Model\Pricing\PricingSetting;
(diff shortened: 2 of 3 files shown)


project-base/app/tests/App/Functional/Model/Order/OrderFacadeTest.php
```diff
@@ -171,14 +170,13 @@ public function testEdit(): void
...
+        $orderData->addItem($orderItemData1, true);
+        $orderData->addItem($orderItemData2, true);
 
-        $orderData->items = $orderItemsData;
         $this->orderFacade->edit($order->getId(), $orderData);
 
         $orderFromDb = $this->orderRepository->getById($order->getId());
 
-        $this->assertCount(5, $orderFromDb->getItems());
+        $this->assertCount(6, $orderFromDb->getItems());
```
grossmannmartin: This test also checks the removal of the order item (see lines 154,155), which now contains unused orderItemsData manipulation. 

This number should remain the same, and one of the items should be removed form order during this test

packages/framework/src/Model/Order/OrderData.php
```diff
@@ -310,10 +310,23 @@ public function addTotalPrice(PriceInterface $priceToAdd, string $type): void
 
     /**
      * @param \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData $item
+     * @param bool $new
      */
-    public function addItem(OrderItemData $item): void
+    public function addItem(OrderItemData $item, bool $new = false): void
```
grossmannmartin: I'm not sure about this. It brings much confusion. 
The addItem() method is called in AddXXX middlewares, and there is always created a new item, but the method is called without setting $new=true.

Also, how should it behave when I call consecutively add with the same product, once with add, once without?

```
$orderData->addItem($orderItemData1, true);
$orderData->addItem($orderItemData1, false);
```

Now, the item is silently ignored. Which tells me that for some cases I must add $new=true (calling productFacade::edit()), while for the others I don't (middlewares). 

I would rather check the failing scenario during productfacade::edit and try to fix it there.

grossmannmartin (changes requested):

project-base/app/tests/App/Functional/Model/Order/OrderFacadeTest.php
TomasLudvik: I have removed changes in this test

packages/framework/src/Model/Order/OrderData.php
TomasLudvik: I have reworked it

grossmannmartin (approved):

pk16011990 (approved):

PR #4113, opened 2025-07-22
enable domain configuration with path fragment

vitek-rostislav:
##### Backend
- In domain configuration, it is now allowed to set a path suffix in the domain URL (e.g., locale)
- added 3rd domain to default demo setup with URL `127.0.0.1:8000/sk` to ensure the domain with locale in path can be properly tested
- Administration is now available from the first domain only. Moreover, when the first domain has a locale suffix, administration is available at the base path + admin URL. E.g., when the first domain is `127.0.0.1:8000/cs`, administration will be available on `127.0.0.1:8000/admin` and access to admin from all domains (e.g., `127.0.0.1/cs/admin`  will be redirected there
- Within the `AdminContext`, there is no "current domain config" from the domains configuration available anymore, hence in administration we can no longer use `$this->domain->getId()` and similar methods
- sitemaps gather items from all domains with the same host
- When sending an email for an administrator (e.g., admin reset password), the mail template translation is searched based on the given administrator's preferred locale

##### Storefront 
- The domain recognition is adjusted to properly work with the defined path fragments
- robots.txt gathers all the information from domains with the same host at the root level (e.g. the robots information for `127.0.0.1/sk` domain is available at `127.0.0.1/robots.txt`)
- all the storages are now identified with domain ID to avoid conflicts in local storage, cookies, and persist store across domains with the same base URL

#### Disclaimer
Even though the ~13.000 added lines might seem horrifying here, it is needless to say that almost 10.000 LOC is taken by adding the 3rd domain with SK translations :slightly_smiling_face: 

### TODOs
- [x] add SF upgrade notes
- [x] release a new version of the Kubernetes deployment that supports the new behavior - see https://github.com/shopsys/deployment/pull/57
- [x] prepare cloud deployment (dev, alpha, prod) for the 3rd domain
- [x] update https://github.com/shopsys/deployment/blob/main/kubernetes/configmap/nginx.yaml
- [x] add BE upgrade notes :grimacing: 
- [x] check sitemaps once again
- [x] finish "SK" translations in po files
- [x] fix CI build so it starts all 3 domains, see `docker-build.yaml`, etc.
- [x] generating friendly URL must not use another domain postfix (e.g. When you create a product with name "sk", the resulting URL must not be example.com/sk if that is actually another domain's URL)
- [x] if I have an e-shop with only two domains, for example https://example.com/fr and https://example.com/pl, then after accessing only https://example.com/ (such a domain does not exist), there will be a redirection to the domain with the locale preferred in the browser and if we do not have such a locale, then to the first locale in the e-shop configuration
- [x] add more tests that cover the "explicit locale in path" behavior (e.g., for updated `Domain::switchDomainByRequest()`)
- [x] fix SF tests
- [x] adjust sitemaps - domains with locales in the path should be grouped under the same base URL
- [x] adjust `robots.txt` - same as sitemaps
- [x] fix storefront referer redirect after login/logout issue - when I logout at `127.0.0.2:8000/zakaznik/reklamace`, I am redirected to `127.0.0.2:8000/cs/prihlaseni?r=zakaznik/reklamace` - there is default locale `cs` in the URL that should not be there.
- [x] verify social login still works for domains with locale in path :cold_face: 
- [x] check proper html lang setting 	with `default` locale in context (could be resolved via `_document.tsx`)
- [x] check alternative SEO domains settings - there should be an automation that simulates the feature for the domains with the same base URL
- [x] fix codeception acceptance tests
- [x] fix hreflang links for homepage - at the moment, all the links lead to the current domain (see `SeoPageResolverMap`)
- [x] check gopay usage on 3rd domain

#### Fixes issues
... 

----
  - https://rv-language-domains.odin.shopsys.cloud
  - https://cz.rv-language-domains.odin.shopsys.cloud
  - https://rv-language-domains.odin.shopsys.cloud/sk


packages/framework/src/Component/Domain/AdminDomainSubscriber.php
```diff
@@ -0,0 +1,85 @@
...
+use Override;
+use Shopsys\FrameworkBundle\Component\Context\AdminContext;
+use Shopsys\FrameworkBundle\Component\Context\ContextResolverInterface;
+use Shopsys\FrameworkBundle\Model\Administration\AdminUrlProvider;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+class AdminDomainSubscriber implements EventSubscriberInterface
```
henzigo, line 18: This class should be in AdministrationBundle

vitek-rostislav: true, I will move it

TomasLudvik:
@coderabbitai review

packages/framework/src/Component/Domain/AdminDomainSubscriber.php
```diff
@@ -0,0 +1,85 @@
...
+        $request = $event->getRequest();
+        $path = $request->getPathInfo();
+
+        if (!$event->isMainRequest() || !$this->contextResolver->isCurrentContext(AdminContext::class)) {
+            return;
+        }
+
+        $currentUrl = $request->getUri();
+
+        // Get the first domain configuration
```
TomasLudvik: There are a lot of not necessary comments in this file that are probably AI-generated. These should be removed.

packages/framework/src/Component/Domain/Domain.php
```diff
@@ -156,33 +155,57 @@ public function getDomainConfigById($domainId)
...
-                $this->currentDomainConfig = $domainConfig;
+            if ($domainConfig->getBaseUrl() !== $requestBaseUrl) {
+                continue;
+            }
+
+            $postfix = '/' . trim($domainConfig->getPostfix() ?? '', '/');
+
+            // Ensure it matches full path segments only
+            if ($postfix !== '/' && !str_starts_with($requestPath, $postfix . '/') && $requestPath !== $postfix) {
+                continue;
```
TomasLudvik: I would refactor this to method with some good name like isPathOfDomainWithPostfix

packages/framework/src/Component/Domain/Domain.php
```diff
@@ -156,33 +155,57 @@ public function getDomainConfigById($domainId)
...
      */
-    public function switchDomainByRequest(Request $request)
+    public function switchDomainByRequest(Request $request): void
     {
-        $url = $request->getSchemeAndHttpHost();
+        $requestBaseUrl = $request->getSchemeAndHttpHost();
+        $requestPath = '/' . trim($request->getPathInfo(), '/');
+
+        $matchingConfig = null;
+        $longestMatchLength = -1;
```
TomasLudvik, line 174: Something like $longestPostfixLength would be more descriptive

packages/framework/src/Component/Domain/Domain.php
```diff
@@ -156,33 +155,57 @@ public function getDomainConfigById($domainId)
...
+            // Ensure it matches full path segments only
+            if ($postfix !== '/' && !str_starts_with($requestPath, $postfix . '/') && $requestPath !== $postfix) {
+                continue;
+            }
+
+            $length = strlen($postfix);
 
-                return;
+            if ($length <= $longestMatchLength) {
+                continue;
```
TomasLudvik: I have wondered why this is necessary until I realized that you want to match the whole longest postfix that you can find. This should be solved via `!str_starts_with($requestPath, $postfix . '/')` that you are using in previous condition, because you use `/` at the end of the string, right? Or do we enable use of `/` in the postfix?

packages/framework/src/Component/Router/NormalizeAdminUrlTrailingSlashSubscriber.php
```diff
@@ -59,7 +59,7 @@ public function onKernelException(ExceptionEvent $event): void
     protected function redirectToExistingPath(string $newPath, ExceptionEvent $event): void
     {
         try {
-            $this->currentDomainRouter->match($newPath);
+            $this->administrationRouter->match($newPath);
```
TomasLudvik, line 69: Is it correct to use administrationRouter here?

packages/framework/src/Model/Inquiry/Mail/InquiryMail.php
```diff
@@ -68,6 +71,8 @@ public function createMessageForAdmin(MailTemplate $template, Inquiry $inquiry):
             $this->getBodyVariablesReplacementsForAdmin($inquiry),
             $this->getSubjectVariablesReplacements($inquiry),
         );
+
+        return $messageData;
     }
```
TomasLudvik, line 74: This change seems to be unnecessary

packages/framework/tests/Unit/Component/Router/DomainRouterTest.php
```diff
@@ -21,9 +25,35 @@ public function testGetRouter()
...
-        $domainRouter = new DomainRouter($context, $basicRouterMock, $localizedRouterMock, $friendlyUrlRouterMock);
         $this->expectException(NotSupportedException::class);
         $domainRouter->setContext($context);
     }
+
+    /**
+     * @return \Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig
+     */
+    protected function getDomainConfig(): DomainConfig
+    {
```
TomasLudvik: There is this exact domain config created in multiple places - would not be better to introduce some Helper class that will provide this everywhere?

project-base/app/tests/FrontendApiBundle/Test/GraphQlB2bDomainTestCase.php
```diff
@@ -21,4 +23,22 @@ protected function setUp(): void
...
+
+    /**
+     * @param string $routeName
+     * @param array<string, mixed> $parameters
+     * @param int $pathType
+     * @return string
+     */
+    #[Override]
+    protected function getLocalizedPathOnFirstDomainByRouteName(
+        string $routeName,
```
TomasLudvik: This is only B2B specific?

project-base/app/tests/FrontendApiBundle/Test/GraphQlB2bDomainWithLoginTestCase.php
```diff
@@ -74,4 +76,22 @@ protected function getCustomerUserByDefaultCredentials(): CustomerUser
...
+
+    /**
+     * @param string $routeName
+     * @param array<string, mixed> $parameters
+     * @param int $pathType
+     * @return string
+     */
+    #[Override]
+    protected function getLocalizedPathOnFirstDomainByRouteName(
+        string $routeName,
```
TomasLudvik: Oh this is second place - should it be moved to GraphqlTestCase?

project-base/storefront/next.config.js
```diff
@@ -79,6 +82,22 @@ const nextConfig = {
...
+            {
+                publicGraphqlEndpoint: process.env.PUBLIC_GRAPHQL_ENDPOINT_HOSTNAME_2, // TODO
+                url: process.env.DOMAIN_HOSTNAME_3,
+                defaultLocale: 'sk',
+                currencyCode: 'EUR',
+                fallbackTimezone: 'Europe/Prague',
+                domainId: 3,
+                mapSetting: {
+                    latitude: 48.669,
+                    longitude: 19.699,
```
TomasLudvik, line 94: These are Slovak coordinates, should we switch those for domain one and two?

packages/framework/src/Model/Administrator/Mail/ResetPasswordMailFacade.php
```diff
@@ -30,11 +30,13 @@ public function __construct(
     /**
      * @param \Shopsys\FrameworkBundle\Model\Administrator\Administrator $administrator
      */
-    public function sendMail(Administrator $administrator)
+    public function sendMail(Administrator $administrator): void
     {
+        $domainId = $this->domain->getFirstDomainIdMatchingAdminSelectedLocale($administrator);
+
```
TomasLudvik, line 36: Just a hunch - this will need upgrade notes as I think many projects did not translate their admin emails until now.

packages/framework/src/Model/Administrator/Mail/TwoFactorAuthenticationMailFacade.php
```diff
@@ -34,9 +36,13 @@ public function __construct(
     #[Override]
     public function sendAuthCode(TwoFactorInterface $administrator): void
     {
+        if (!$administrator instanceof Administrator) {
+            throw new InvalidArgumentException('Expected instance of Administrator.');
+        }
```
TomasLudvik, line 41: Can you make this message a bit more descriptive?

translate_po.py
```diff
@@ -0,0 +1,284 @@
+#!/usr/bin/env python3
+"""
```
TomasLudvik: These py scripts should be removed, right?

docs/introduction/how-to-set-up-domains-and-locales.md
```diff
@@ -70,6 +70,38 @@ Set the url address for the domain in `config/domains_urls.yaml`.
 
     When you add a domain with the new url address on the MacOS platform, you need to enable this url address also in the network interface, see [Installation Using Docker for MacOS](../installation/installation-using-docker-macos.md#21-enable-second-domain-optional)
 
+##### Domain URLs with Locale Postfixes
+
```
TomasLudvik, line 74: This is not only possible usage - do we want to extend this somehow?

docs/introduction/how-to-set-up-domains-and-locales.md
```diff
@@ -70,6 +70,38 @@ Set the url address for the domain in `config/domains_urls.yaml`.
...
+    - id: 1
+      url: http://127.0.0.1:8000
+    - id: 2
+      url: http://127.0.0.1:8000/cs
+    - id: 3
+      url: http://127.0.0.1:8000/sk
+```
+
+**Domain Resolution Logic:**
+
```
TomasLudvik: Do you feel this is important in the docs? I feel like not as people do not care, it if just works :)

docs/introduction/how-to-set-up-domains-and-locales.md
```diff
@@ -175,7 +207,49 @@ php phing domains-db-functions-create
...
+**Behavior with Domain Postfixes:**
+
+- Admin is always accessible from the base URL of the first domain (without postfix)
+- Even if the first domain has a postfix (e.g., `http://127.0.0.1:8000/en`), admin access is at `http://127.0.0.1:8000/admin` (base URL)
+- Access attempts from other domains are redirected to the first domain's base URL
+- Admin path and query parameters are preserved during redirects
+
+**Technical Implementation:**
+
```
TomasLudvik: Is it necessary to mention this?

TomasLudvik (changes requested):
I would like you to review all comments added and restrict them to needed only.

translate_po.py
vitek-rostislav: Yes, I have already removed them in one of the later commits

packages/framework/src/Component/Domain/AdminDomainSubscriber.php
vitek-rostislav: Yes, now it is better, right? 207dc74788

packages/framework/src/Component/Domain/Domain.php
vitek-rostislav: Well the thing is, I am not able to guarantee the domain's configuration. When there is "root" domain defined after a locale-specific domain, the root domain would be processed later in the loop and would override the resolved config. So you for the following configuration:
```
1: http://127.0.0.1:8000/cs
2: http://127.0.0.1:8000
```
and for request `/cs`, you would match the first domain, but then, you would also match the second domain.

I have slightly refactored it in 41e0e0ded5 and added more tests to cover the described behavior. We can talk about it if you want :wink:

packages/framework/src/Component/Domain/Domain.php
vitek-rostislav: refactored in 41e0e0ded5

packages/framework/src/Component/Domain/Domain.php
vitek-rostislav: refactored in 41e0e0ded5

packages/framework/src/Component/Router/NormalizeAdminUrlTrailingSlashSubscriber.php
vitek-rostislav: Yes, the subscriber is used for routes in administration only, so when you encounter 404 in admin, it tries the same path with or without a trailing slash, and if it matches an admin route, it performs the redirect. I have added the context resolver and renamed the class to make it more obvious that it is used within admin only. What do you think? 9cfc283224

packages/framework/src/Model/Inquiry/Mail/InquiryMail.php
vitek-rostislav: yes, fixed in 3c656d5884

project-base/storefront/next.config.js
vitek-rostislav: I do not know, honestly. I believe we can address it in a separate issue if needed. At the current 17.0 with two domains, we have the first domain in English, however, it has stores in Czechia. The second domain is in Czech with stores in Slovakia :shrug: I have just added another domain that is in Slovak, so it made sense to have the Slovak stores there. We can brainstorm better data fixtures for the first domain, however, I believe it is not worth it. Let's talk about it

packages/framework/src/Model/Administrator/Mail/ResetPasswordMailFacade.php
vitek-rostislav: noted :+1:

packages/framework/tests/Unit/Component/Router/DomainRouterTest.php
vitek-rostislav: well exactly the same domain config is created here and in `DomainTest::createDomainConfigFirst()`. TBH, I do not think creating a helper would have any added value. Or are there any other places that I am missing?

project-base/app/tests/FrontendApiBundle/Test/GraphQlB2bDomainTestCase.php
vitek-rostislav: It was just a weird remnant from the past times, I do not know why it was there :facepalm: removed in deddaa8a2d

project-base/app/tests/FrontendApiBundle/Test/GraphQlB2bDomainWithLoginTestCase.php
vitek-rostislav: It was just a weird remnant from the past times, I do not know why it was there :facepalm: removed in deddaa8a2d

packages/framework/src/Model/Administrator/Mail/TwoFactorAuthenticationMailFacade.php
vitek-rostislav: f27dfb4c8e, better?

docs/introduction/how-to-set-up-domains-and-locales.md
vitek-rostislav: TBH, I do not know what should be added. Do you have any suggestions, please?

docs/introduction/how-to-set-up-domains-and-locales.md
vitek-rostislav: Yes, it felt important to me (this was actually me telling Claude Code to include this information in the docs :smile:

docs/introduction/how-to-set-up-domains-and-locales.md
vitek-rostislav: true, removed in 5654d961f5

packages/framework/tests/Unit/Component/Router/DomainRouterTest.php
vitek-rostislav: I will take a look at it

docs/introduction/how-to-set-up-domains-and-locales.md
vitek-rostislav: I will make it more general

docs/introduction/how-to-set-up-domains-and-locales.md
vitek-rostislav: improved in 0466ce9233

packages/framework/src/Component/Domain/Domain.php
vitek-rostislav: ...and refactored more in 97d01f8b02 :smile:

packages/framework/src/Component/Domain/Domain.php
vitek-rostislav: ...and refactored more in 97d01f8b02 :smile:

docs/introduction/how-to-set-up-domains-and-locales.md
vitek-rostislav: tweaked in 95ebe9fd2f. However, I need to discuss the storefront part, as that has the implementation more tied to the locales themselves. I will address that after the BE review is finished :slightly_smiling_face:

packages/framework/tests/Unit/Component/Router/DomainRouterTest.php
vitek-rostislav: ok, so here it is, easypick :smile: 4b6f8af2bb

docs/introduction/how-to-set-up-domains-and-locales.md
vitek-rostislav: 95ebe9fd2f024d2682fe06dd6735e378ca3a863b

packages/framework/src/Component/Domain/AdminDomainSubscriber.php
TomasLudvik: Screenshot at Sep 11 13-50-20

:sweat_smile:

TomasLudvik (approved):

TomasLudvik (approved):

packages/framework/src/Component/Domain/AdminDomainSubscriber.php
vitek-rostislav: e34e79534d

project-base/storefront/components/providers/PersistStoreProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+            return null;
+        }
+
+        return parsedData.state;
+    } catch (error) {
+        logException(error);
+
+        return null;
+    }
+};
```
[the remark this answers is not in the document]
vitek-rostislav: I do not want that programatically

upgrade-notes/backend_20250915_140343.md
```diff
@@ -0,0 +1,22 @@
+#### enable domain configuration with path fragment ([#4113](https://github.com/shopsys/shopsys/pull/4113))
+
+- `Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig` constructor now has new required `$baseUrl` and optional `$postfix` parameters
+    - `DomainConfig::getUrl()` method returns the whole URL including the postfix, if you need just the host part, use `DomainConfig::getBaseUrl()` method instead
+- `Shopsys\FrameworkBundle\Component\Domain\Domain::MAIN_ADMIN_DOMAIN_ID` constant has been removed as the administration domain is no longer tied to a specific domain ID
```
TomasLudvik, line 5: Well we expect first domain as admin domain - right?

upgrade-notes/backend_20250915_140343.md
```diff
@@ -0,0 +1,22 @@
...
+    - the meaning of the parameter has changed, it now represents additional path prefixes instead of the original route names for the administration context
+    - the configuration parameter name has changed from `admin_context_route_prefixes` to `admin_context_additional_path_prefixes`
+- `Shopsys\FrameworkBundle\Component\Router\NormalizeUrlTrailingSlashSubscriber` has been renamed to `NormalizeAdminUrlTrailingSlashSubscriber`
+    - the subscriber now uses `Shopsys\FrameworkBundle\Component\Router\AdministrationRouter $administrationRouter` instead of the original `CurrentDomainRouter $currentDomainRouter`
+- `Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\NormalizeUrlTrailingSlashListener` class has been removed as redundant
+    - the functionality is covered by `NormalizeAdminUrlTrailingSlashSubscriber`
+- `Shopsys\FrameworkBundle\Model\Sitemap\SitemapListener::addUrlForHomepage()` method no longer accepts `string $section` parameter
+    - the section name is now resolved using `SitemapFacade::getSectionNameForDomainConfig()` method
+- see #project-base-diff to update your project
```
TomasLudvik, line 23: There should be also storefront upgrade notes right?

TomasLudvik, line 23: Should we also mention need to update deployment bundle?

TomasLudvik (changes requested):

upgrade-notes/backend_20250915_140343.md
vitek-rostislav: Well, not exactly, it is the first domain minus the suffix :shrug: Btw the only usage of `MAIN_ADMIN_DOMAIN_ID` constant was in `InquiryMail` to generate a link to the `admin_inquiry_detail` route, which is now replaced with the administration router. 

But if you do not like the sentence here in the upgrade notes, I can change it to 
```
- Shopsys\FrameworkBundle\Component\Domain\Domain::MAIN_ADMIN_DOMAIN_ID constant has been removed.
```

upgrade-notes/backend_20250915_140343.md
vitek-rostislav: Yes, definitely, I will add them after the SF review is finished :innocent:

vitek-rostislav: Good catch, I will upgrade the bundle once it is ready (in progress here - https://github.com/shopsys/deployment/pull/57) and explicitly mention that in the upgrade notes to upgrade it on a project

project-base/storefront/utils/domain/getHostAndDomainFromRequest.ts
```diff
@@ -0,0 +1,168 @@
...
+        .filter((entry) => {
+            const domainUrlObj = new URL(entry.domainUrl);
+            return domainUrlObj.host === requestUrlObj.host;
+        })
+        .sort((a, b) => {
+            const aPath = new URL(a.domainUrl).pathname;
+            const bPath = new URL(b.domainUrl).pathname;
+            // Sort by path length descending (longest first)
+            return bPath.length - aPath.length;
+        });
```
JanMolcik: We should pre-parse domain configuration at the startup instead of creating URL objects from domainUrl repeatedly

Problem:
```ts
// Current inefficient approach
const matchingDomains = domainEntries
    .filter((entry) => {
        const domainUrlObj = new URL(entry.domainUrl); // ❌ URL parsing in filter
        return domainUrlObj.host === requestUrlObj.host;
    })
    .sort((a, b) => {
        const aPath = new URL(a.domainUrl).pathname; // ❌ URL parsing in sort
        const bPath = new URL(b.domainUrl).pathname; // ❌ URL parsing in sort
        return bPath.length - aPath.length;
    });
```

Instead, above in the `const domainEntries = ...` we could pre-parse it right away:
```ts
const domainEntries = Object.keys(STATIC_REWRITE_PATHS).map((domainUrl, index) => {
    const normalizedDomainUrl = removeTrailingSlash(domainUrl);
    const locale = getExplicitPathDomainLocaleOrDefault(normalizedDomainUrl);
    const domainUrlObj = new URL(normalizedDomainUrl);

    return {
        domainUrl: normalizedDomainUrl,
        locale,
        domainId: index + 1,
        host: domainUrlObj.host,
        pathname: domainUrlObj.pathname,
    };
});
```

upgrade-notes/backend_20250915_140343.md
JanMolcik, line 1: Missing storefront upgrade notes.

project-base/storefront/components/providers/PersistStoreProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    return migratedPersistedState as PersistStore;
+                },
+            },
+        ),
+    );
+};
+
+export const PersistStoreProvider: FC<PersistStoreProviderProps> = ({ children }) => {
+    const domainConfig = useDomainConfig();
+    const storeRef = useRef<StoreApi<PersistStore>>(createPersistStore(domainConfig.domainId));
```
JanMolcik: As discussed, if you switch to different domain using client-side navigation, the original persist store will still be used instead the newly created one (using old `domainId`). Replacing `useRef(...)` with
```ts
useMemo(() => createPersistStore(domainConfig.domainId), [domainConfig.domainId]);
``` 
should fix it.

project-base/storefront/utils/domain/getHostAndDomainFromRequest.ts
JanMolcik, line 1: This `getHostAndDomainFromRequest` became such a big beast :-D how about splitting it into separate parts for saving mental capacity & better debugging? I have this suggestion:

🔧 Proposed Solution: Function Decomposition

Step 1: Extract Request Parsing

```ts
// utils/domain/requestParser.ts
type ParsedRequest = {
    host: string;
    protocol: string;
    pathname: string;
    headers: Headers;
};

const parseRequest = (request: NextRequest): ParsedRequest => {
    const requestHeaders = new Headers(request.headers);
    const requestHost = requestHeaders.get('host');

    if (!requestHost) {
        throw new Error('Host was not found in the request header.');
    }

    const protocol = requestHeaders.get('x-forwarded-proto') || 'http';
    const originalUrl = new URL(request.url);

    return {
        host: requestHost,
        protocol,
        pathname: originalUrl.pathname,
        headers: requestHeaders,
    };
};
```

Step 2: Extract Domain Matching

```ts
// utils/domain/domainMatcher.ts
type DomainMatch = {
    domainUrl: string;
    domainId: number;
    locale: string;
    isExactMatch: boolean;
};

const findDomainByPath = (
    parsedRequest: ParsedRequest,
    domainEntries: DomainEntry[]
): DomainMatch | null => {
    const requestUrlObj = new URL(`${parsedRequest.protocol}://${parsedRequest.host}`);

    // Filter domains by matching host
    const matchingDomains = domainEntries
        .filter(entry => {
            const domainUrlObj = new URL(entry.domainUrl);
            return domainUrlObj.host === requestUrlObj.host;
        })
        .sort((a, b) => {
            // Sort by path length (longest first for specificity)
            const aPath = new URL(a.domainUrl).pathname;
            const bPath = new URL(b.domainUrl).pathname;
            return bPath.length - aPath.length;
        });

    // Find exact path match
    for (const domainEntry of matchingDomains) {
        const domainUrlObj = new URL(domainEntry.domainUrl);
        const domainPath = domainUrlObj.pathname;

        if (isPathMatch(parsedRequest.pathname, domainPath)) {
            return {
                domainUrl: domainEntry.domainUrl,
                domainId: domainEntry.domainId,
                locale: domainEntry.locale,
                isExactMatch: true,
            };
        }
    }

    return null;
};

const isPathMatch = (requestPath: string, domainPath: string): boolean => {
    if (domainPath === '/') return true;

    return (
        requestPath === domainPath ||
        requestPath.startsWith(domainPath + '/') ||
        requestPath + '/' === domainPath + '/'
    );
};
```

Step 3: Extract Browser Locale Detection

```ts
// utils/domain/localeDetector.ts
const findDomainByBrowserLocale = (
    parsedRequest: ParsedRequest,
    domainsWithSameHost: DomainEntry[]
): DomainMatch => {
    const acceptLanguage = parsedRequest.headers.get('accept-language') || '';
    const browserLocales = parseBrowserLocales(acceptLanguage);

    // Try to match browser's preferred locale
    let targetDomain = domainsWithSameHost[0]; // Default fallback

    for (const browserLocale of browserLocales) {
        const matchingDomain = domainsWithSameHost.find(domain => {
            // Skip default locale for explicit matching
            if (domain.locale === DEFAULT_LOCALE) return false;
            return domain.locale.toLowerCase() === browserLocale.toLowerCase();
        });

        if (matchingDomain) {
            targetDomain = matchingDomain;
            break;
        }
    }

    return {
        domainUrl: targetDomain.domainUrl,
        domainId: targetDomain.domainId,
        locale: targetDomain.locale,
        isExactMatch: false,
    };
};

const parseBrowserLocales = (acceptLanguage: string): string[] => {
    if (!acceptLanguage) return [];

    const locales = acceptLanguage
        .split(',')
        .map(lang => {
            const [locale, qValue] = lang.trim().split(';q=');
            const quality = qValue ? parseFloat(qValue) : 1.0;
            const languageCode = locale.split('-')[0].toLowerCase();
            return { locale: languageCode, quality };
        })
        .sort((a, b) => b.quality - a.quality)
        .map(item => item.locale);

    return Array.from(new Set(locales));
};
```

Step 4: Refactored Main Function

```ts
// utils/domain/getHostAndDomainFromRequest.ts (now clean!)
export const getHostAndDomainFromRequest = (
    request: NextRequest
): { host: string; domainId: number; currentLocale: string | undefined; redirect?: boolean }
=> {
    // Step 1: Parse request (5 lines)
    const parsedRequest = parseRequest(request);
    const domainEntries = getDomainEntries(); // Pre-processed config

    // Step 2: Try exact domain match (3 lines)
    const exactMatch = findDomainByPath(parsedRequest, domainEntries);
    if (exactMatch?.isExactMatch) {
        return {
            host: exactMatch.domainUrl + '/',
            domainId: exactMatch.domainId,
            currentLocale: exactMatch.locale,
        };
    }

    // Step 3: Fallback to browser locale (5 lines)
    const domainsWithSameHost = getDomainsForHost(parsedRequest.host, domainEntries);
    if (domainsWithSameHost.length > 0) {
        const localeMatch = findDomainByBrowserLocale(parsedRequest, domainsWithSameHost);
        return {
            host: localeMatch.domainUrl + '/',
            domainId: localeMatch.domainId,
            currentLocale: localeMatch.locale,
            redirect: true,
        };
    }

    // Step 4: Ultimate fallback (3 lines)
    const ultimateFallback = domainEntries[0];
    return {
        host: ultimateFallback.domainUrl + '/',
        domainId: ultimateFallback.domainId,
        currentLocale: ultimateFallback.locale,
        redirect: true,
    };
};
```

Benefits of This Refactoring

1. Easier Testing

```ts
// Before: 1 huge test file with 120+ test cases
// After: Focused test files

describe('parseRequest', () => {
    // 5 test cases for request parsing
});

describe('findDomainByPath', () => {
    // 8 test cases for path matching  
});

describe('findDomainByBrowserLocale', () => {
    // 6 test cases for locale detection
});

describe('getHostAndDomainFromRequest', () => {
    // 4 integration test cases (much simpler!)
});
```

2. Easier Debugging

```ts
// Bug report: "Slovak domain not working"
// Before: Debug 120 lines
// After: 
// 1. Test parseRequest() → ✅ Works
// 2. Test findDomainByPath() → ❌ Bug found in 15 lines!
// 3. Fix in isolated function → ✅ Fixed
```

3. Easier Maintenance

```ts
// New requirement: "Add Czech language support"
// Before: Modify 120-line function (risky)
// After: Modify findDomainByBrowserLocale() (safe)

// Performance optimization needed?  
// Before: Optimize entire 120-line function
// After: Optimize specific helper (getDomainEntries pre-processing)
```

project-base/storefront/vitest/utils/domain/getHostAndDomainFromRequest.test.ts
JanMolcik, line 1: Possibly add separate tests for each util function after `getHostAndDomainFromRequest` splitting.

JanMolcik (changes requested):

project-base/storefront/components/providers/PersistStoreProvider.tsx
JanMolcik: I have pushed 1 `[DROP]` commit with the setup for testing this issue.

vitek-rostislav: fixed in e431bd3fdf

upgrade-notes/backend_20250915_140343.md
vitek-rostislav: I know :innocent: I will add them, I just wanted to first have the review of the code :smile:

project-base/storefront/components/providers/PersistStoreProvider.tsx
vitek-rostislav: However, there is still one problem - when I do the client domain switch, I see the red notification about the number of products in the wishlist that is left over from the domain I switch from, even though there are no products in the wishlist on the switched domain. The header is then properly updated (the number disappears) once I click on it and navigate to the wishlist section, where I see there are no products (which is correct). 
image

vitek-rostislav: I would like to discuss it further. I was able to fix that problem for the product lists; however, I do not like the fact that we do not have a universal solution on the URQL level. This is a message from Claude:
```
 Exactly right! You've identified the core issue. Let me explain what's
  happening:

  1. productListUuid - This comes from the persist store (line 39:
  productListUuids[productListType]), which is now properly recreated per domain
  thanks to our earlier fix with useMemo. So this is correctly null or undefined
  for the new domain.
  2. productListData - This comes from the URQL query result (line 46:
  useProductListQuery). When you switch domains:
    - The query gets paused because !productListUuid (line 53)
    - But URQL keeps the previous query result in its React state
    - Even though the query is paused, productListData still contains the cached
  data from the previous domain's query
  3. Why does this happen?
    - URQL's useQuery hook maintains its result state across re-renders
    - When a query is paused, URQL doesn't clear the previous result - it just
  stops fetching
    - This is by design in URQL - it assumes you might want to keep showing the
  last known data while paused

  So the sequence is:
  1. Domain 1: Query runs, fetches wishlist with 3 items → productListData has 3
  items
  2. Switch to Domain 2: productListUuid becomes null, query pauses
  3. But productListData still contains the 3 items from Domain 1
  4. Without our fix, the UI would show those 3 items even though they're from
  the wrong domain

  That's why returning undefined when !productListUuid is the correct fix - it
  explicitly clears the data when we know we're on a domain without a wishlist
  UUID.

```

c5cbafead3

vitek-rostislav:
```
The problem is clear now - UrqlWrapper is above DomainConfigProvider in the
  component tree, so it can't use the context. And during client-side navigation,
   pageProps.domainConfig might not be updating properly.

  Let me think about this differently. Perhaps we should just go with the simpler
   solution - keep the fix in useProductList that returns undefined when there's
  no UUID. This is actually a good defensive programming practice anyway.

  But if you want a more general solution, we could consider:

  1. Moving the URQL provider inside the DomainConfigProvider (but this would be
  a bigger refactor)
  2. Creating a separate component that sits between DomainConfigProvider and
  children that monitors domain changes
  3. Using a different cache strategy in URQL
```

project-base/storefront/utils/domain/getHostAndDomainFromRequest.ts
vitek-rostislav: 8f4410ba37 - awesome, the result is much better now IMHO :ok_hand: It is not exactly the same as your proposal, Claude and I have modified it a bit, but the idea is the same :slightly_smiling_face: 

I am keeping the original tests for `getHostAndDomainFromRequest` untouched, as they nicely describe and verify the behaviour as a whole. I will add suitable tests for the sub-utils in another commit

upgrade-notes/backend_20250915_140343.md
vitek-rostislav: In fact, I would like to discuss that with you. I am not sure how to approach the upgrade notes for this task, as projects will either have the functionality already custom-implemented or will never need it, so I would appreciate a brainstorming regarding this with you.

project-base/storefront/utils/domain/getHostAndDomainFromRequest.ts
vitek-rostislav: nice catch, refactored within 8f4410ba37

project-base/storefront/vitest/utils/domain/getHostAndDomainFromRequest.test.ts
vitek-rostislav: here they are :slightly_smiling_face: c03e6a53f3

JanMolcik (approved):

pk16011990 (approved):

TomasLudvik (approved):

PR #4124, opened 2025-07-31
refactor administrator login as user

vitek-rostislav:

After https://github.com/shopsys/shopsys/pull/4113 is finished, administration will always be available on the 1st domain only. (Moreover, when the 1st domain has a language postfix, e.g., `example.com/cs`, the administration will always be available on the base URL without the postfix - `example.com/admin`). From now on, the cross-domain administrator login (SSO) is no longer supported.

The current feature of impersonating an administrator as a user must have been reworked to allow administrators to impersonate users that belong to a domain that is different from the admin domain. Now, instead of setting the storefront access and refresh token cookies from the administration domain, a one-time exchange token is generated and used for the login via the new `LoginViaExchangeToken` mutation.

#### Fixes issues
... 

## Summary by CodeRabbit

* **New Features**
  * Introduced a secure "login as customer user" feature for administrators using one-time exchange tokens.
  * Added a new GraphQL mutation and supporting API endpoints for logging in via exchange tokens.
  * Implemented frontend support for automatic login via exchange token in the storefront homepage.

* **Improvements**
  * Centralized and simplified the rendering of "Log in as user" links in the admin interface.
  * Updated the login flow to use exchange tokens instead of SSO or multidomain tokens.

* **Bug Fixes**
  * Removed legacy single sign-on (SSO) and multidomain login functionality from both backend and frontend.

* **Documentation**
  * Updated schema documentation and upgrade notes to reflect the new login process and mutation.

* **Chores**
  * Added and updated tests to cover the new exchange token login flow.
  * Removed obsolete tests and configuration related to multidomain login and SSO.

----
  - https://rv-refactor-login-admin-as-user.odin.shopsys.cloud
  - https://cz.rv-refactor-login-admin-as-user.odin.shopsys.cloud


vitek-rostislav:
@coderabbitai review

packages/frontend-api/src/Model/Security/LoginAsUserFacade.php
```diff
@@ -50,27 +45,32 @@ public function __construct(
...
-        $user = $this->customerUserRepository->getCustomerUserById($customerUserId);
-        $administrator = $this->administratorFrontSecurityFacade->getCurrentAdministrator();
+        $user = $exchangeTokenEntity->getCustomerUser();
+        $administrator = $exchangeTokenEntity->getAdministrator();
 
         $this->customerUserLoginTypeFacade->updateCustomerUserLoginTypes(
             $this->customerUserLoginTypeDataFactory->create($user, LoginTypeEnum::ADMIN),
         );
 
+        $this->loginAsUserExchangeTokenFacade->delete($exchangeTokenEntity);
```
grossmannmartin, line 72: Do we have/need any mechanism to delete old unused exchange tokens?

packages/frontend-api/src/Controller/Admin/CustomerUserController.php
```diff
@@ -33,17 +42,29 @@ public function __construct(
     public function loginAsCustomerUserAction(int $customerUserId): Response
     {
         try {
-            return $this->render('@ShopsysFrontendApi/Admin/Content/Login/loginAsCustomerUser.html.twig', [
-                'tokens' => $this->loginAsUserFacade->loginAdministratorAsCustomerUserAndGetAccessAndRefreshToken($customerUserId),
-                'url' => $this->generateUrl('front_homepage', [], UrlGeneratorInterface::ABSOLUTE_URL),
-                'showCartMergeInfo' => 'false',
-            ]);
+            $customerUser = $this->customerUserFacade->getCustomerUserById($customerUserId);
```
grossmannmartin, line 44: can we protect this action with CSRF token? 
It seems like an unpleasant security risk now.

packages/frontend-api/src/Model/Security/LoginAsUserExchangeTokenDataFactory.php
```diff
@@ -0,0 +1,39 @@
...
+     * @param \Shopsys\FrameworkBundle\Model\Administrator\Administrator $administrator
+     * @return \Shopsys\FrontendApiBundle\Model\Security\LoginAsUserExchangeTokenData
+     */
+    public function create(
+        CustomerUser $customerUser,
+        Administrator $administrator,
+    ): LoginAsUserExchangeTokenData {
+        $exchangeTokenData = $this->createInstance();
+
+        $exchangeTokenData->token = bin2hex(random_bytes(32));
```
grossmannmartin: I'm thinking - wouldn't be more secure to store just hash into a DB and verify it similarly as the password is? 

It would require a slightly different approach to validation of the exchange token (identifier + token?). 
And as the token has a short lifespan, maybe it's not such a problem...

packages/frontend-api/src/Migrations/Version20250730125606.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $this->sql('
+            CREATE TABLE login_as_user_exchange_tokens (
+                token VARCHAR(64) NOT NULL,
```
grossmannmartin, line 21: token should be unique. I'm thinking about problem with code / several page submit / ... 
what do you think?

packages/frontend-api/src/Model/Security/LoginAsUserExchangeTokenRepository.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param string $token
+     * @return \Shopsys\FrontendApiBundle\Model\Security\LoginAsUserExchangeToken|null
+     */
+    public function findValidByToken(string $token): ?LoginAsUserExchangeToken
+    {
+        return $this->entityManager->createQueryBuilder()
+            ->select('e')
+            ->from(LoginAsUserExchangeToken::class, 'e')
+            ->where('e.token = :token')
+            ->andWhere('e.expiresAt > :now')
```
grossmannmartin: I don't expect millions of rows in the `login_as_user_exchange_tokens` table, but still... Did you consider adding some index? 

But if you add an unique index (as mentioned earlier), this index will be used to search... so probably one stone, two birds.

packages/frontend-api/src/Resources/config/graphql-types/Mutation/MutationDecorator.types.yaml
```diff
@@ -30,6 +30,13 @@ MutationDecorator:
                     input:
                         type: RefreshTokenInput!
                 resolve: "@=mutation('refreshTokensMutation', args)"
+            LoginViaExchangeToken:
```
grossmannmartin, line 33: maybe (just maybe) we could introduce some form of rate limiting to this query? 
So it may be called only once per minute or so... (reducing from ~10⁷⁷ years to check all possibilities to ~6×10⁷⁸ years 😄 )

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginViaExchangeTokenTest.php
```diff
@@ -0,0 +1,143 @@
...
+
+    public function testLoginViaExchangeTokenMutationWithValidToken(): void
+    {
+        $graphQlType = 'LoginViaExchangeToken';
+
+        $customerUser = $this->getReference(CustomerUserDataFixture::CUSTOMER_PREFIX . '1', CustomerUser::class);
+        $administrator = $this->getReference(AdministratorDataFixture::SUPERADMINISTRATOR, Administrator::class);
+
+        $exchangeTokenData = $this->loginAsUserExchangeTokenDataFactory->create($customerUser, $administrator);
+        $exchangeToken = $this->loginAsUserExchangeTokenFacade->create($exchangeTokenData);
```
grossmannmartin: I'm thinking - do we need a data object and an entity in this scenario? Wouldn't just the entity suffice? ( I mean in this scenario = in this whole PR)

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginViaExchangeTokenTest.php
```diff
@@ -0,0 +1,143 @@
...
+
+        try {
+            $this->tokenFacade->getTokenByString($responseData['refreshToken']);
+        } catch (Throwable) {
+            $this->fail('Refresh token is not valid');
+        }
+
+        // Verify the exchange token is no longer valid (one-time use)
+        $expiredToken = $this->loginAsUserExchangeTokenFacade->findValidByToken($exchangeToken->getToken());
+        $this->assertNull($expiredToken, 'Exchange token should be deleted after use');
```
grossmannmartin: If we decide to move the token storage elsewhere, this test will still pass. 
Shouldn't we try to login with the same exchange token again and check the response?

edit: Oh, I see this scenario tested in one of the following test scenarios... Couldn't be those two tests unified into one?

packages/framework/src/Resources/views/Admin/Content/Customer/User/edit.html.twig
```diff
@@ -1,15 +1,8 @@
...
-    {% set loginAsUserUrl = getLoginAsUserUrl(customerUser) %}
-    {% if loginAsUserUrl is not null %}
-        <a href="{{ loginAsUserUrl }}" target="_blank" class="btn-link-style btn-link-style--iconic-l">
-            {{ ux_icon('forward-page') }}
-            <span>
-                {{ 'Log in as user'|trans }}
-            </span>
-        </a>
-    {% endif %}
+    {{ macro.loginAsUserLink(customerUser) }}
```
grossmannmartin: I'm not sure about such generic name for imported file... I would use `loginAsUserLinkMacro.loginAsUserLink(customerUser)`. Even though it's longer. 
If we add another macro, we won't struggle with the general naming

project-base/storefront/components/Pages/HomePage/ExchangeTokenHandler.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useEffect, useRef } from 'react';
+import { useLoginViaExchangeToken } from 'utils/auth/useLoginViaExchangeToken';
+import { showErrorMessage } from 'utils/toasts/showErrorMessage';
+
+export const ExchangeTokenHandler: FC = () => {
+    const router = useRouter();
+    const loginViaExchangeToken = useLoginViaExchangeToken();
+    const processedTokens = useRef(new Set<string>());
+
+    useEffect(() => {
```
grossmannmartin, line 11: I would like to hear an opinion of one of FE devs. 
I have expected use of middleware to handle the login, but maybe it's not possible 🤷‍♂️

And with the upcoming approuter we should probably avoid use of useEffect.

project-base/storefront/components/Pages/HomePage/ExchangeTokenHandler.tsx
```diff
@@ -0,0 +1,37 @@
...
+        const { exchangeToken } = router.query;
```
grossmannmartin, line 12: IF (and that's a big if) we would be able to use POST and do not pass the token in url, that would be really nice touch.

grossmannmartin (changes requested):

packages/frontend-api/src/Controller/Admin/CustomerUserController.php
vitek-rostislav: good catch, ba441047ff

packages/frontend-api/src/Migrations/Version20250730125606.php
vitek-rostislav: Well, the token is the primary key at the moment, so it actually has the unique constraint

packages/framework/src/Resources/views/Admin/Content/Customer/User/edit.html.twig
vitek-rostislav: sure - c803c75d3c

packages/frontend-api/src/Resources/config/graphql-types/Mutation/MutationDecorator.types.yaml
vitek-rostislav: ah, good idea, we even had that already for the regular login :ok_hand: b7ca3d70a3

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginViaExchangeTokenTest.php
vitek-rostislav: I believe following the common pattern does not hurt here. Even if we do not consider this to be a common entity, I still think wrapping the input arguments into a DTO is a good practice :shrug:

vitek-rostislav: However, if you feel like it would be really better without the data object, I will simplify it, I have no strong feeling one way or another here

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginViaExchangeTokenTest.php
vitek-rostislav: Yes, I have moved the null check into that second test. I think it is better now, right? cdea6bce1f
Another option is to remove that assertion completely, as from an API point of view, I basically do not care :shrug:

project-base/storefront/components/Pages/HomePage/ExchangeTokenHandler.tsx
vitek-rostislav: yes, I am definitelly going to send this to SF dev review, so we will see :slightly_smiling_face:

vitek-rostislav: maybe the review will become re-implementation actually :smile:

project-base/storefront/components/Pages/HomePage/ExchangeTokenHandler.tsx
vitek-rostislav: let's talk about it, please, I am afraid I am not able to do that easily in an acceptable way :smile:

packages/frontend-api/src/Model/Security/LoginAsUserFacade.php
vitek-rostislav: we have now :slightly_smiling_face: eca0037245

packages/frontend-api/src/Model/Security/LoginAsUserExchangeTokenDataFactory.php
vitek-rostislav: let's talk about it, please

packages/frontend-api/src/Controller/Admin/CustomerUserController.php
vitek-rostislav: So you are telling me that adding the annotation is not enough? Thank go we have smoke tests. d86758ea7c :smile:

vitek-rostislav: ...and once again and better :smile: 393e8b4b76

packages/frontend-api/src/Model/Security/LoginAsUserExchangeTokenDataFactory.php
vitek-rostislav: reworked to store just the token hash, not the unecrypted value cff0debc8b

project-base/app/tests/FrontendApiBundle/Functional/Login/LoginViaExchangeTokenTest.php
vitek-rostislav: in the end, I have removed the data object + factory when reworking the feature with token encryption in cff0debc8b

project-base/storefront/components/Pages/HomePage/ExchangeTokenHandler.tsx
vitek-rostislav: discussed, closing :smile:

packages/frontend-api/src/Model/Security/LoginAsUserExchangeTokenRepository.php
```diff
@@ -6,32 +6,44 @@
...
+
+        $passwordHasher = $this->passwordHasherFactory->getPasswordHasher(Administrator::class);
+
+        foreach ($validTokens as $tokenEntity) {
+            if ($passwordHasher->verify($tokenEntity->getToken(), $unencryptedToken)) {
+                return $tokenEntity;
+            }
+        }
+
+        return null;
```
vitek-rostislav: @grossmannmartin this is something that Claude did (well, as most of the code :smile: ) and at first, I wanted to rewrite it to pass the encrypted token into that method and just keep that original simple find. However, that would require encrypting the input token in `LoginAsUserExchangeTokenFacade::findValidByToken()`. The solution with `passwordHasher->verify` here seems more elegant in the end, even more when we do not expect having a lot of unexpired tokens in the table anytime. However, I am open to dicsussion here, I could totally live with encrypting the input token and searching by the encrypted value :shrug:

packages/frontend-api/src/Controller/Admin/CustomerUserController.php
grossmannmartin: 😄

project-base/storefront/components/Pages/HomePage/ExchangeTokenHandler.tsx
grossmannmartin: //cc @chlebektomas @JanMolcik

packages/frontend-api/src/Model/Security/LoginAsUserExchangeTokenRepository.php
grossmannmartin: I have the same thoughts when I read the code first time.   Let's keep it this way.

grossmannmartin (approved):

JanMolcik (approved):

pk16011990 (approved):

PR #4127, opened 2025-08-02
customer social login registration: normalize empty strings to null

vitek-rostislav:
When a user creates a registration using a social profile (facebook, google, ...), we can not be sure about the format of incoming data. As we prefer `null` values over empty strings in the application, we should rather normalize the data this way to avoid unexpected behavior. E.g., the customer user data (such as name and surname) are updated from his first order if they are `null` after registration - this breaks when they are empty strings, actually.

#### Fixes issues
... 

----
  - https://rv-social-registration-data.odin.shopsys.cloud
  - https://cz.rv-social-registration-data.odin.shopsys.cloud

### DIFF
--- a/packages/framework/tests/Unit/Component/UtilsTest.php
+++ b/packages/framework/tests/Unit/Component/UtilsTest.php
@@ -9,14 +9,14 @@ use Shopsys\FrameworkBundle\Component\Utils\Utils;
 
 class UtilsTest extends TestCase
 {
-    public function testIfNull()
+    public function testIfNull(): void
     {
         $this->assertTrue(Utils::ifNull(null, true));
         $this->assertFalse(Utils::ifNull(false, true));
         $this->assertTrue(Utils::ifNull(true, false));
     }
 
-    public function testSetArrayDefaultValueExists()
+    public function testSetArrayDefaultValueExists(): void
     {
         $array = [
             'key' => 'value',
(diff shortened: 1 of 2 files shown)


packages/framework/src/Component/Utils/Utils.php
```diff
@@ -16,6 +16,19 @@ public static function ifNull(mixed $testVariable, mixed $default): mixed
         return $testVariable ?? $default;
     }
 
+    /**
+     * @param string|null $string
+     * @return string|null
+     */
+    public static function emptyStringToNull(?string $string): ?string
```
grossmannmartin: I believe the same feature is already provided by `\Shopsys\FrameworkBundle\Component\String\TransformStringHelper::getTrimmedStringOrNullOnEmpty()`.

when even the returned string is trimmed (which I think should be)

grossmannmartin (changes requested):

packages/framework/src/Component/Utils/Utils.php
vitek-rostislav: Ah, good catch, I had that feeling as well, but then I did not find the method in `Utils` class, and forgot about `TransformStringHelper` :facepalm: fixed in 818f556d29

grossmannmartin (approved):

pk16011990 (approved):

PR #4134, opened 2025-08-07
AdminContext now includes elfinder routes

vitek-rostislav:
Without these routes, the wysiwyg was not properly translated to the selected locale of the current administrator.

Moreover, introduced `shopsys.admin_context_route_prefixes` parameter for easy future extension

#### Fixes issues
... 

----
  - https://rv-fix-elfinder-localization.odin.shopsys.cloud
  - https://cz.rv-fix-elfinder-localization.odin.shopsys.cloud

### DIFF
--- a/packages/framework/src/Component/Context/AdminContext.php
+++ b/packages/framework/src/Component/Context/AdminContext.php
@@ -5,15 +5,18 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Component\Context;
 
 use Override;
+use Shopsys\FrameworkBundle\Component\Utils\Utils;
 use Symfony\Component\HttpFoundation\RequestStack;
 
 final class AdminContext extends AbstractContext
 {
     /**
      * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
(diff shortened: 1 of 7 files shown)


packages/framework/src/Resources/config/parameters_common.yaml
```diff
@@ -33,3 +33,9 @@ parameters:
         - 'admin_login'
         - 'admin_login_check'
         - 'admin_logout'
+
+    shopsys.admin_context_route_prefixes:
```
grossmannmartin: I would be slightly happier if we used configuration instead of parameter (see \Shopsys\FrameworkBundle\DependencyInjection\Configuration), but I can live with this solution

grossmannmartin (approved):

packages/framework/src/Resources/config/parameters_common.yaml
vitek-rostislav: What about this? :slightly_smiling_face: https://github.com/shopsys/shopsys/pull/4134/commits/40ed8b1b064ac3e47bd2da14101610fbfa5e605b

pk16011990 (approved):

PR #4135, opened 2025-08-07
Introduced factories for all implementations of DataSourceInterface + QueryBuilderDataSources now order textual columns with correct collation

### LINKED ISSUE #2386: DataGrid ordering don't allow sorting by locale
## Describe the bug
DataGrid column only enable/disable sorting, but then there is no info about column type, so it's not possible to sort with any collation. 

### To Reproduce

1. Switch administration to czech
2. Create new product named `Čistící prostředek`
4. Show list of products in admin and sort by name
5. See the new product sorted at the end instead of after `C`, where it should be

### Expected behavior
All already existing grids with sorting should sort properly by locale.  
It should be possible and easy to sort columns in newly created grids by locale.

TomasLudvik:
Introduced factories for all implementations of DataSourceInterface. This leads to easier extensibility of DataSources + it is easier to use DI.

QueryBuilderDataSources now order textual columns with correct collation so data in grids are ordered correctly for current administrator language.

#### Fixes issues
closes #2386

----
  - https://tl-grids-ordering-by-administrator-language.odin.shopsys.cloud
  - https://cz.tl-grids-ordering-by-administrator-language.odin.shopsys.cloud

### DIFF
--- a/docs/administration/internal-grid/grid-data-sources.md
+++ b/docs/administration/internal-grid/grid-data-sources.md
@@ -2,23 +2,31 @@
 
 As a data source, the [Grid Component](index.md) requires an implementation of [`DataSourceInterface`]({{github.link}}/packages/framework/src/Component/Grid/DataSourceInterface.php).
 
-You can find 4 implementations of the interface in Shopsys Platform.
+## Use Factories for Creating DataSources
 
-## [`QueryBuilderDataSource`]({{github.link}}/packages/framework/src/Component/Grid/QueryBuilderDataSource.php)
+- [`QueryBuilderDataSourceFactory`]({{github.link}}/packages/framework/src/Component/Grid/QueryBuilderDataSourceFactory.php)
+- [`QueryBuilderWithRowManipulatorDataSourceFactory`]({{github.link}}/packages/framework/src/Component/Grid/QueryBuilderWithRowManipulatorDataSourceFactory.php)
+- [`ArrayDataSourceFactory`]({{github.link}}/packages/framework/src/Component/Grid/ArrayDataSourceFactory.php)
+- [`ArrayWithPaginationDataSourceFactory`]({{github.link}}/packages/framework/src/Component/Grid/ArrayWithPaginationDataSourceFactory.php)
+- [`MoneyConvertingDataSourceDecoratorFactory`]({{github.link}}/packages/framework/src/Component/Grid/MoneyConvertingDataSourceDecoratorFactory.php)
 
 The most commonly used data source is created from Doctrine Query Builder.
 
 ### Example of usage
 
 ```php
-/** @var Doctrine\ORM\EntityManagerInterface $entityManager */
+public function __construct(
+    private readonly \Doctrine\ORM\EntityManagerInterface $em,
+    private readonly \Shopsys\FrameworkBundle\Component\Grid\QueryBuilderDataSourceFactory $queryBuilderDataSourceFactory,
+) {
+}
 
-$queryBuilder = $entityManager->createQueryBuilder();
+$queryBuilder = $this->em->createQueryBuilder();
 
 $queryBuilder->select('p')
     ->from(Product::class, 'p');
 
-$dataSource = new QueryBuilderDataSource($queryBuilder, 'p.id', SortableNullsWalker::class);
+$dataSource = $this->queryBuilderDataSourceFactory->create($queryBuilder, 'p.id');
 ```
 
 QueryBuilderDataSource contains a third default parameter $hint which is pre-set to SortableNullsWalker::class as a way to sort null values to the beginning in the case of ASC sorting, or to the end in the case of DESC sorting.
@@ -33,20 +41,25 @@ e.g., you can add some calculated price into the data set.
 ### Example of usage
 
 ```php
-/** @var Shopsys\FrameworkBundle\Model\Transport\TransportRepository $transportRepository */
-/** @var Shopsys\FrameworkBundle\Model\Localization\Localization $localization */
+public function __construct(
+    private readonly \Doctrine\ORM\EntityManagerInterface $em,
+    private readonly \Shopsys\FrameworkBundle\Component\Grid\QueryBuilderWithRowManipulatorDataSourceFactory $queryBuilderWithRowManipulatorDataSourceFactory,
+    private readonly \Shopsys\FrameworkBundle\Model\Localization\Localization $localization,
+    private readonly \Shopsys\FrameworkBundle\Model\Transport\TransportRepository $transportRepository,
+) {
+}
 
-$queryBuilder = $transportRepository->getQueryBuilderForAll()
+$queryBuilder = $this->transportRepository->getQueryBuilderForAll()
     ->addSelect('tt')
     ->join('t.translations', 'tt', Join::WITH, 'tt.locale = :locale')
-    ->setParameter('locale', $localization->getCurrentLocaleForTranslatableEntities());
+    ->setParameter('locale', $this->localization->getCurrentLocaleForTranslatableEntities());
 
-$dataSource = new QueryBuilderWithRowManipulatorDataSource(
+$dataSource = $this->queryBuilderWithRowManipulatorDataSourceFactory->create(
     $queryBuilder,
     't.id',
     function ($row, $results) {
         if ($this->transportsIndexedByIdLocalCache === null) {
-            $this->transportsIndexedByIdLocalCache = $transportRepository->findAllByIdsIndexedById(array_map(fn ($result) => $result['t']['id'], $results));
+            $this->transportsIndexedByIdLocalCache = $this->transportRepository->findAllByIdsIndexedById(array_map(fn ($result) => $result['t']['id'], $results));
         }
 
         $transport = $this->transportsIndexedByIdLocalCache[$row['t']['id']];
@@ -74,10 +87,14 @@ Data source that is created from an array. It is suitable when you need to displ
 ### Example of usage
 
 ```php
-/** @var Shopsys\FrameworkBundle\Component\Domain\Domain $domain */
+public function __construct(
+    private readonly \Shopsys\FrameworkBundle\Component\Domain\Domain $domain,
+    private readonly \Shopsys\FrameworkBundle\Component\Grid\ArrayDataSourceFactory $arrayDataSourceFactory,
+) {
+}
 
 $domainData = [];
-foreach ($domain->getAll() as $domainConfig) {
+foreach ($this->domain->getAll() as $domainConfig) {
     $domainData[] = [
         'id' => $domainConfig->getId(),
         'name' => $domainConfig->getName(),
@@ -86,7 +103,7 @@ foreach ($domain->getAll() as $domainConfig) {
     ];
 }
 
-$dataSource = new ArrayDataSource($domainData, 'id');
+$dataSource = $this->arrayDataSourceFactory->create($domainData, 'id');
 ```
 
 ## [`MoneyConvertingDataSourceDecorator`]({{github.link}}/packages/framework/src/Component/Grid/MoneyConvertingDataSourceDecorator.php)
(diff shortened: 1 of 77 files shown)


henzigo:
Hm, what about using `und-x-icu` as the default collation for the database that will correctly order characters for almost every language instead of manually setting specific collation for every query that needs sorting by text? 

Or better, we can create custom und collation. This should be the same as default `und-x-icu`, but this is explicitly defined. `und-x-icu` can differ by os/versions:

```sql
CREATE COLLATION "und-x-icu-l3"
  (provider = icu,
   locale = 'und-u-kn-true-ks-level3',
   deterministic = true);
  ```
  
  That has some advantages:
  - kn=true (natural numeric sort) -> „1, 2, 10“ instead of „1, 10, 2“
  - ks=level3 (https://www.postgresql.org/docs/current/collation.html#ICU-COLLATION-COMPARISON-LEVELS)  -> explicitly distinguishes between diacritics and case

So with this, we don't need to explicitly add `COLLATE` for every SQL query that we want to sort.

Only one negative with this is that existing databases must be manually dumped and restored to the newly created database with this default collation.

We can have a call about it.

**Comment by grossmannmartin** [comment]
@henzigo interesting idea. I did not know about it. 

I tried to verify the idea, but it seems like postgres is struggling with the default collation being some ICU collation (or maybe I am doing something wrong).

I tried this

```
CREATE DATABASE razeni3
  LC_COLLATE 'und-x-icu'
  LC_CTYPE   'und-x-icu'
  TEMPLATE   template0;

CREATE TABLE t (name varchar(255));

insert into t (name) values ('Ala'),('ala'),('Caj'),('Čaj'),('Hrad'),('Chata'),('chleba'),('Řeka'),('Zena'),('Žena'),
  ('Snek'),('Šnek'),('dama'),('dzban'),('džem'),('Zebra'),('Źrebak'),('Żaba'),('Łódź'),
  ('kor'),('kór'),('kör'),('kőr'),('úr'),('űr'),('csiga'),('gyár'),('nyár'),('tyúk'),('zsír'),
  ('dzseki'),('dzsungel');
```

And the simple ordered query (`select * from t order by name`) still returns the data in the wrong sort order. 

It seems we would be forced to create all text fields with the appropriate collation, or still use collation at the query level. 

I'm afraid that someone one day forgets to add collation to one text column and all hell breaks loose.

So if we stick with the query level collation, then I don't find much difference with using specific collation or und_x. 
As a bonus, locale-specific collation sorts properly digraphs (ch in czech, dz in slovak, dzs, ly ny in hungarian, ...)

Maybe I did something wrong, in that case I would be happy if you correct me :)

**Comment by henzigo** [comment]
Try this:
```sql
CREATE DATABASE sort
  TEMPLATE template0
  ENCODING 'UTF8'
  LOCALE_PROVIDER = icu
  ICU_LOCALE = 'und-u-kn-true-ks-level3'; 
```

TomasLudvik:
I have tried it and result is:

und-x-icu | cs-x-icu
-- | --
ala | ala
Ala | Ala
Caj | Caj
Čaj | csiga
Chata | Čaj
chleba | dama
csiga | dzban
dama | dzseki
dzban | dzsungel
džem | džem
dzseki | gyár
dzsungel | Hrad
gyár | Chata
Hrad | chleba
kor | kor
kór | kór
kör | kör
kőr | kőr
Łódź | Łódź
nyár | nyár
Řeka | Řeka
Snek | Snek
Šnek | Šnek
tyúk | tyúk
úr | úr
űr | űr
Żaba | Żaba
Zebra | Zebra
Zena | Zena
Žena | Źrebak
Źrebak | zsír
zsír | Žena

docs/administration/internal-grid/grid-data-sources.md
```diff
@@ -33,20 +41,25 @@ e.g., you can add some calculated price into the data set.
...
 
-$queryBuilder = $transportRepository->getQueryBuilderForAll()
+$queryBuilder = $this->transportRepository->getQueryBuilderForAll()
     ->addSelect('tt')
     ->join('t.translations', 'tt', Join::WITH, 'tt.locale = :locale')
-    ->setParameter('locale', $localization->getCurrentLocaleForTranslatableEntities());
+    ->setParameter('locale', $this->localization->getCurrentLocaleForTranslatableEntities());
 
-$dataSource = new QueryBuilderWithRowManipulatorDataSource(
+$dataSource = $this->queryBuilderWithRowManipulatorDataSource->create(
```
grossmannmartin: Shouldn't this be `$this->queryBuilderWithRowManipulatorDataSourceFactory` and the class in the constructor   `QueryBuilderWithRowManipulatorDataSourceFactory` instead of `QueryBuilderDataSourceFactory`?

docs/cookbook/create-basic-grid.md
```diff
@@ -71,17 +71,28 @@ And because we want to get data from the database, we use a data source created
...
++    * @var \Shopsys\FrameworkBundle\Component\Grid\QueryBuilderDataSourceFactory
++    */
++   protected $queryBuilderDataSourceFactory;
+
 -   public function __construct(GridFactory $gridFactory)
-+   public function __construct(GridFactory $gridFactory, EntityManagerInterface $entityManager)
++   public function __construct(
++       GridFactory $gridFactory,
++       EntityManagerInterface $entityManager,
++       QueryBuilderDataSourceFactory $queryBuilderDataSourceFactory
```
grossmannmartin: nitpick: Constructor property promotion would help with readability

packages/framework/src/Component/Grid/QueryBuilderDataSource.php
```diff
@@ -18,12 +17,12 @@ class QueryBuilderDataSource implements DataSourceInterface
...
      * @param \Doctrine\ORM\QueryBuilder $queryBuilder
      * @param string $rowIdSourceColumnName
-     * @param string|null $hint
+     * @param array $hints
      */
     public function __construct(
         protected readonly QueryBuilder $queryBuilder,
         protected readonly string $rowIdSourceColumnName,
-        protected readonly ?string $hint = SortableNullsWalker::class,
+        protected array $hints = [],
```
grossmannmartin: Doesn't this change the default value? 

If someone uses the query builder data source, will it still have the sortablenullswalker?
Maybe when we now have factories, we could make this option mandatory (remove the default value)?

packages/framework/src/Component/Grid/QueryBuilderDataSourceFactory.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function create(
+        QueryBuilder $queryBuilder,
+        string $rowIdSourceColumnName,
+        ?array $hints = null,
+    ): QueryBuilderDataSource {
+        return new QueryBuilderDataSource(
+            $queryBuilder,
+            $rowIdSourceColumnName,
+            $hints ?? $this->hintsHelper->getDefaultHints(),
```
grossmannmartin, line 33: Oh, I see now the default hints are set here... I am for removing the default value from QueryBuilderDataSource constructor

packages/framework/src/Component/Grid/QueryBuilderWithRowManipulatorDataSource.php
```diff
@@ -16,15 +15,15 @@ class QueryBuilderWithRowManipulatorDataSource extends QueryBuilderDataSource
...
      * @param \Closure $manipulateRowCallback
-     * @param string|null $hint
+     * @param array $hints
      */
     public function __construct(
         QueryBuilder $queryBuilder,
         string $rowIdSourceColumnName,
         protected readonly Closure $manipulateRowCallback,
-        ?string $hint = SortableNullsWalker::class,
+        protected array $hints = [],
```
grossmannmartin: same here

packages/framework/src/Component/Paginator/QueryPaginator.php
```diff
@@ -19,12 +18,12 @@ class QueryPaginator implements PaginatorInterface
     /**
      * @param \Doctrine\ORM\QueryBuilder $queryBuilder
      * @param string|null $hydrationMode
-     * @param string|null $hint
+     * @param array $hints
```
grossmannmartin: What about a more exact type annotation? I think it should be `array<string, mixed>` (probably everywhere where hints are used)

upgrade-notes/backend_20250813_093009.md
```diff
@@ -0,0 +1,14 @@
...
+
+- added new factories for all DataSource types to ensure consistent instantiation
+    - `\Shopsys\FrameworkBundle\Component\Grid\QueryBuilderDataSourceFactory`
+    - `\Shopsys\FrameworkBundle\Component\Grid\QueryBuilderWithRowManipulatorDataSourceFactory`
+    - `\Shopsys\FrameworkBundle\Component\Grid\ArrayDataSourceFactory`
+    - `\Shopsys\FrameworkBundle\Component\Grid\ArrayWithPaginationDataSourceFactory`
+    - `\Shopsys\FrameworkBundle\Component\Grid\MoneyConvertingDataSourceDecoratorFactory`
+- added proper collation support for textual columns ordering by administrator language
+- `\Shopsys\FrameworkBundle\Component\Doctrine\CollationOrderByWalker`
+    - new Doctrine walker that applies collation to ORDER BY clauses for textual fields (ASCII_STRING, STRING, TEXT, JSON types)
```
grossmannmartin: Json type is no longer an option according to the code

grossmannmartin (changes requested):

docs/administration/internal-grid/grid-data-sources.md
TomasLudvik: You are absolutely right!

docs/cookbook/create-basic-grid.md
```diff
@@ -71,17 +71,18 @@ And because we want to get data from the database, we use a data source created
...
-    {
+-   {
++   public function __construct(
++       protected readonly GridFactory $gridFactory,
++       protected readonly EntityManagerInterface $entityManager,
++       protected readonly QueryBuilderDataSourceFactory $queryBuilderDataSourceFactory
++   ) {
         $this->gridFactory = $gridFactory;
 +       $this->entityManager = $entityManager;
++       $this->queryBuilderDataSourceFactory = $queryBuilderDataSourceFactory;
```
grossmannmartin: these initialization shouldn't be here due to use of constructor property promotion

grossmannmartin (approved):
please check build

grossmannmartin (approved):

packages/framework/src/Model/Customer/User/CustomerUserRepository.php
```diff
@@ -115,55 +117,61 @@ public function getCustomerUserListQueryBuilderByQuickSearchData(
...
+                (' . $ordersCountSubquery->getDQL() . ') as ordersCount,
+                (' . $ordersSumSubquery->getDQL() . ') as ordersSumPrice,
+                (' . $lastOrderSubquery->getDQL() . ') as lastOrderAt')
+            ->from(Customer::class, 'c')
+            ->join(CustomerUser::class, 'cu', Join::WITH, 'cu.customer = c.id AND cu.domainId = :selectedDomainId AND NOT EXISTS (
+                SELECT 1 FROM ' . CustomerUser::class . ' cu2 
+                WHERE cu2.customer = c.id AND cu2.domainId = :selectedDomainId AND cu2.id < cu.id
+            )')
+            ->join('cu.pricingGroup', 'pg')
+            ->join(BillingAddress::class, 'ba', Join::WITH, 'ba.customer = c.id AND NOT EXISTS (
```
grossmannmartin: I think this query still needs some love...

- The original query uses a left join; this modified version will exclude any customers without a billing address (and pricing group? Not sure about that). 

- Original query and modified version return different data. 
 The problem is with the aggregated numbers of orders, their value, and the date of the last order. Original query aggregates all customer users, while the modified version gets only one 

Original:

Image

Modified:

Image

- Original query is much more performative. Three correlated subqueries = 90 separate queries for 30 results. Additionally, the query will become significantly slower with more data in the database. (You can check the cost in query plan in explain query)

grossmannmartin (changes requested):

grossmannmartin (approved):

pk16011990 (approved):

pk16011990 (approved):

PR #4150, opened 2025-08-20
codeception acceptance tests now use pg_restore to dump DB from sql file

vitek-rostislav:
After the new Postgres client version (`17.6.`) was released, the SQL created by `pg_dump` now contains `\restrict` and `\unrestrict` meta commands that are not valid SQL for PDO, so the database restoring stopped working in our Codeception acceptance tests.

- using `populator` in `acceptance.suite.yml` is the recommended approach, see https://codeception.com/docs/modules/Db.html#Populator
- removed `DatabaseHelper` - Db module is now configured directly using env variables
- for monorepo, added var/cache folder to simplify differences when running from project/monorepo

As a bonus, the acceptance tests run should now be a little bit (~15s) faster.

#### Fixes issues
... 

----
  - https://rv-fix-codeception-postgres.odin.shopsys.cloud
  - https://cz.rv-fix-codeception-postgres.odin.shopsys.cloud


packages/framework/src/Component/Doctrine/DatabaseConnectionCredentialsProvider.php
```diff
@@ -8,14 +8,12 @@ class DatabaseConnectionCredentialsProvider
...
     /**
      * @param string $databaseHost
-     * @param string $databasePort
      * @param string $databaseName
      * @param string $databaseUsername
      * @param string $databasePassword
      */
     public function __construct(
         protected readonly string $databaseHost,
-        protected readonly string $databasePort,
```
grossmannmartin: I would rather do the opposite - add port to the DatabaseDumpCommand. 
Now when someone tries to use a different port (which is allowed by the configuration of the doctrine), dump will fails

packages/framework/src/Component/Doctrine/DatabaseConnectionCredentialsProvider.php
```diff
@@ -8,14 +8,12 @@ class DatabaseConnectionCredentialsProvider
...
         protected readonly string $databaseName,
         protected readonly string $databaseUsername,
```
grossmannmartin: Maybe I would consider removing this class entirely and passing the necessary credentials directly into the database dump command (which should be the only place that uses this class).

But I am completely ok with keeping this

project-base/app/tests/App/Acceptance/acceptance.suite.yml
```diff
@@ -8,12 +8,13 @@ actor: AcceptanceTester
...
-            user: ~
-            password: ~
-            dump: ../var/cache/test-db-dump.sql
+            dsn: 'pgsql:host=%DATABASE_HOST%;port=%DATABASE_PORT%;dbname=%DATABASE_NAME%'
+            user: '%DATABASE_USER%'
+            password: '%DATABASE_PASSWORD%'
+            dump: var/cache/test-db-dump.sql
             populate: true
             cleanup: true
+            populator: 'PGPASSWORD=$password pg_restore -U $user -h %DATABASE_HOST% -d %DATABASE_NAME% $dump'
```
grossmannmartin: Why do we use different sources for user and host/name (`$user` and `%DATABASE_NAME%`)? 
Can't we use `%DATABASE_USER%` instead of `$user`?

grossmannmartin (changes requested):

project-base/app/tests/App/Acceptance/acceptance.suite.yml
vitek-rostislav: yes, we can, the thing is, I do not have the `$` variables in the config for host and name. But I can unify it so that I use just the env variables in the populator, it makes sense :+1:

packages/framework/src/Component/Doctrine/DatabaseConnectionCredentialsProvider.php
vitek-rostislav: For BC's sake, I am keeping the provider (So that the command custructor is not changed :slightly_smiling_face: )

packages/framework/src/Component/Doctrine/DatabaseConnectionCredentialsProvider.php
vitek-rostislav: great point :ok_hand: aea91b436b

project-base/app/tests/App/Acceptance/acceptance.suite.yml
vitek-rostislav: In the end, I unified it to use just the variables (because of `$dump`). I am able to use just the variables in the populator as the docs say "Variable names are being taken from config and DSN which has a keyword=value format, so you should expect to have a variable named as the keyword with the full value inside it." - so even `host` and `dbname` can be used here because of their usage in dsn. So my final version is here - 29c99718e8

grossmannmartin (approved):

pk16011990 (approved):

PR #4163, opened 2025-08-29
order payment method is now changed if user selected different payment method in GoPay payment gate

TomasLudvik:
User can change payment method in GoPay e.g. when he selected in cart that he want to pay using credit card, but in GoPay he decides to use bank transfer. In such case we (until now) did not change selected payment in the order.

----
  - https://tl-fix-gopay-changed-payment-method.odin.shopsys.cloud
  - https://cz.tl-fix-gopay-changed-payment-method.odin.shopsys.cloud


packages/framework/src/Model/Payment/PaymentFacade.php
```diff
@@ -350,4 +350,22 @@ public function isPaymentVisibleAndEnabledOnCurrentDomain(Payment $payment): boo
...
         }
     }
+
+    /**
+     * @param string $externalPaymentMethod
+     * @param \Shopsys\FrameworkBundle\Model\Transport\Transport $transport
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Payment\Payment|null
+     */
+    public function getPaymentByExternalMethodTransportAndDomainId(
```
grossmannmartin: Shouldn't it be `findPayment...` when it can return null?

packages/framework/src/Model/GoPay/OrderGoPayStatusUpdateCronModule.php
```diff
@@ -61,7 +64,15 @@ public function run(): void
...
             try {
+                $previousTransactionMethod = $order->getLastTransaction()?->getExternalPaymentMethod();
+
                 $this->paymentServiceFacade->updatePaymentTransactionsByOrder($order);
+
+                $lastTransactionMethod = $order->getLastTransaction()?->getExternalPaymentMethod();
+
+                if ($lastTransactionMethod !== $previousTransactionMethod) {
+                    $this->orderFacade->updatePaymentByLastPaymentTransaction($order);
+                }
```
grossmannmartin: This check is duplicated in two places. What about changing the `updatePaymentByLastPaymentTransaction` method so it can return whether the last transaction has changed the payment?

packages/framework/src/Model/GoPay/OrderGoPayStatusUpdateCronModule.php
```diff
@@ -61,7 +64,15 @@ public function run(): void
             $oldIsOrderPaid = $order->isPaid();
 
             try {
+                $previousTransactionMethod = $order->getLastTransaction()?->getExternalPaymentMethod();
```
grossmannmartin: shouldn't it be lastGoPayTransaction?

grossmannmartin (changes requested):

packages/framework/src/Model/GoPay/OrderGoPayStatusUpdateCronModule.php
TomasLudvik: This code is no longer used

grossmannmartin (approved):

pk16011990 (approved):

PR #4173, opened 2025-09-04
prevent exceeding available stock for products with negative stock disabled

grossmannmartin:
#### Description, the reason for the PR

Added a new **"Allow negative stock"** checkbox option on the product edit page that controls whether customers can order more items than are currently in stock.

  **Before this change:**
  All products could be ordered in unlimited quantities regardless of available stock levels.

  **After this change:**
  - Products with **"Allow negative stock" enabled** (default): customers can order unlimited quantities as before
  - Products with **"Allow negative stock" disabled** (NEW): customers can only order up to the available stock quantity

  **Automatic quantity adjustment:**
  When a product has negative stock disabled and the requested quantity exceeds available stock, the system automatically reduces the cart quantity to match available stock. Items are completely removed from the cart if no stock is available.

  These quantity adjustments happen automatically during cart operations, and customers are informed about the changes.

#### Fixes issues
... 

#### License Agreement for contributions

- [X] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://mg-exceed-available-stock.odin.shopsys.cloud
  - https://cz.mg-exceed-available-stock.odin.shopsys.cloud

PR #4193, opened 2025-09-16
Product gift

RostislavKreisinger:
The administrator can assign a gift (a regular product) to selected products. The customer will then be able to add this gift to their order for a special price (typically 1 CZK, or even 0 CZK).

### Main Functionality

- In the administration (section Pricing) a new section “Gifts with Product” will be available.
- Gift = regular product, only specially linked.
- Each gift plan contains:
- name (mandatory),
- gift (mandatory – the product that will be added as a gift),
- list of products to which this gift will be added,
- validity period (from / to; both can be null = unlimited),
- gift price (multi-domain, can also be 0 CZK).
- 

### Administrator

- Sees and manages the list of gift plans.
- Can set the price of the gift globally (multi-domain).
- In the administration there is a new non-deletable label “Gift with Product”, which also appears on the FE.

### Customer (FE)

- On the product detail page, sees the available gift (only if it is in stock or backorders are allowed).
- In the cart, sees the added gifts (they can be removed via the cross, after refresh they are added back).
- The gift is clickable (link to product detail).
- The gift behaves like a standard order item → shown in order summary, emails, and order detail.

... 

- [ok] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
  - https://rk-ssp-3561-product-gift.odin.shopsys.cloud
  - https://cz.rk-ssp-3561-product-gift.odin.shopsys.cloud
  - https://rk-ssp-3561-product-gift.odin.shopsys.cloud/sk


packages/framework/src/Model/Cart/Item/CartItem.php
```diff
@@ -191,4 +200,36 @@ public function getUuid()
...
+     */
+    public function getType()
+    {
+        return $this->type;
+    }
+
+    /**
+     * @param string $type
+     */
+    public function setType($type): void
```
vitek-rostislav: it can be protected, it is never used anywhere (or you can get rid of the setter completely and just set the property in the constructor)

packages/framework/src/Model/Cart/Item/CartItemFactory.php
```diff
@@ -23,16 +23,18 @@ public function __construct(protected readonly EntityNameResolver $entityNameRes
...
         Product $product,
         int $quantity,
         ?Money $watchedPrice,
+        string $type,
     ): CartItem {
         $entityClassName = $this->entityNameResolver->resolve(CartItem::class);
 
-        return new $entityClassName($cart, $product, $quantity, $watchedPrice);
+        return new $entityClassName($cart, $product, $quantity, $watchedPrice, $type);
     }
```
vitek-rostislav, line 43: We could add a validation of the type here (see e.g. `ComplaintStatusFactory`)

packages/framework/src/Model/Cart/Cart.php
```diff
@@ -226,6 +226,18 @@ public function getQuantifiedProducts()
         return $quantifiedProducts;
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Cart\Item\CartItem $cartItem
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\QuantifiedProduct
+     */
+    protected function createQuantifiedProduct(CartItem $cartItem): QuantifiedProduct
+    {
```
vitek-rostislav, line 234: thumbs up for extracting the creation into a separate method :)

packages/framework/src/Model/Cart/Cart.php
```diff
@@ -497,4 +509,20 @@ public function assignCartToCustomerUser(?CustomerUser $customerUser): void
...
+        return array_filter($this->getItems(), fn (CartItem $cartItem) => $cartItem->isProductGift());
+    }
+
+    /**
+     * @return \Shopsys\FrameworkBundle\Model\Cart\Item\CartItem[]
+     */
+    public function getProductCartItems(): array
+    {
+        return array_filter($this->getItems(), fn (CartItem $cartItem) => $cartItem->isProduct());
+    }
```
vitek-rostislav, line 527: IMHO we should make getItems() protected now, so that a programmer is always forced to choose which items they want. Or rename it to something like `getAllItems()`. When you look at the current usages of `Cart::getItems()`, I do not think they are correct now, and should be replaced by one of these new methods. It is used e.g., in `CartFacade::addProductToExistingCart()` - and I think it creates a buggy scenario when you have a gift in a cart and then try to add the same "gift" product "regularly", it will actually increase the quantity of the gift, right?

vitek-rostislav: I tried to test the described expected bug, and it even makes it impossible to add the gift product regularly when it is already in the cart as a gift

packages/framework/src/Model/Order/Item/QuantifiedProduct.php
```diff
@@ -8,15 +8,23 @@
...
 {
+    public const string CART_ITEM_TYPE_KEY = 'cartItemType';
+
     protected int $quantity;
 
+    /**
+     * @var array<string,mixed>
+     */
+    protected array $additionalData;
+
```
vitek-rostislav, line 19: I like this approach :+1:

packages/framework/src/Model/Order/Processing/OrderProcessorMiddleware/AddProductsMiddleware.php
```diff
@@ -39,21 +41,39 @@ public function handle(
...
         return $orderProcessingStack->processNext($orderProcessingData);
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\OrderData $orderData
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\QuantifiedProduct $quantifiedProduct
+     * @param \Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessingData $orderProcessingData
+     */
+    protected function addProductOrderItemData(
+        OrderData $orderData,
```
vitek-rostislav, line 58: a lot of developers will thank you for the extractions :)

packages/framework/src/Model/Order/Processing/OrderProcessorMiddleware/AddProductGiftsMiddleware.php
```diff
@@ -0,0 +1,97 @@
...
+        $orderData->addTotalPrice($quantifiedItemPrice->getTotalPrice(), OrderItemTypeEnum::TYPE_PRODUCT_GIFT);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\QuantifiedItemPriceInterface $quantifiedItemPrice
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\QuantifiedProduct $quantifiedProduct
+     * @param string $locale
+     * @return \Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData
+     */
+    protected function createProductGiftItemData(
```
vitek-rostislav: the method is duplicated with `AddProductsMiddleware::createProductItemData()`, PHPStorm says :smile: And it is right, could extract it somewhere to avoid DRY vlolation?

packages/framework/src/Controller/Admin/GiftPlanController.php
```diff
@@ -0,0 +1,205 @@
...
+    #[CanView(methods: [HttpMethod::GET])]
+    public function editAction(Request $request, int $id): Response
+    {
+        $giftPlan = $this->giftPlanFacade->getById($id);
+
+        $giftPlanData = $this->giftPlanDataFactory->createFromGiftPlan($giftPlan);
+
+        $form = $this->createForm(GiftPlanFormType::class, $giftPlanData, [
+            'scenario' => GiftPlanFormType::SCENARIO_EDIT,
+            'giftPlan' => $giftPlan,
```
vitek-rostislav, line 78: I know we used this approach in the past, and it is probably still used somewhere, however, you can get rid of this option completely. Instead, you can use `$options['giftPlan'] !== null` in the form :)

packages/framework/src/Controller/Admin/GiftPlanController.php
```diff
@@ -0,0 +1,205 @@
...
+            ->from(GiftPlan::class, 'gp')
+            ->where('gp.domainId = :selectedDomainId')
+            ->setParameter('selectedDomainId', $this->adminDomainTabsFacade->getSelectedDomainId());
+        $dataSource = new QueryBuilderDataSource(
+            $queryBuilder,
+            'gp.id',
+        );
+
+        $grid = $this->gridFactory->create('giftPlanList', $dataSource, AdminRoleConstant::ROLE_GIFT_PLAN);
+        $grid->enablePaging();
```
vitek-rostislav, line 142: I would create a separate `GiftPlanGridFactory` for that (as we sometimes do)

packages/framework/src/Form/Admin/GiftPlan/GiftPlanFormType.php
```diff
@@ -0,0 +1,122 @@
...
+    #[Override]
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builderSettingsGroup = $builder->create('settings', GroupType::class, [
+            'label' => t('Settings'),
+        ]);
+
+        $builderSettingsGroup
+            ->add('domain', DisplayOnlyType::class, [
+                'data' => $this->adminDomainTabsFacade->getSelectedDomainConfig()->getName(),
```
vitek-rostislav: more UX friendly would be probably enabling the admin to change the domain when creating the plan, but I believe we can live with this :shrug:

packages/framework/src/Form/Admin/GiftPlan/GiftPlanFormType.php
```diff
@@ -0,0 +1,122 @@
...
+                    'label' => t('ID'),
+                ]);
+        }
+
+        $builderSettingsGroup
+            ->add('name', TextType::class, [
+                'required' => true,
+                'constraints' => [
+                    new Constraints\NotBlank(['message' => 'Please enter name of gift plan']),
+                ],
```
vitek-rostislav, line 76: please add a length contraint

packages/framework/src/Model/Product/GiftPlan/GiftPlan.php
```diff
@@ -0,0 +1,179 @@
...
+     * @ORM\ManyToOne(targetEntity="Shopsys\FrameworkBundle\Model\Product\Product")
+     * @ORM\JoinColumn(name="gift_product_id", referencedColumnName="id", nullable=false, onDelete="CASCADE")
+     */
+    protected $giftProduct;
+
+    /**
+     * @var \DateTime|null
+     * @ORM\Column(type="datetime", nullable=true)
+     */
+    protected $validFrom = null;
```
vitek-rostislav: the null default value is not necessary

packages/framework/src/Model/Product/GiftPlan/GiftPlan.php
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * @var \DateTime|null
+     * @ORM\Column(type="datetime", nullable=true)
+     */
+    protected $validTo = null;
```
vitek-rostislav: the null default value is not necessary

packages/framework/src/Model/Product/GiftPlan/Exception/GiftPlanNotFoundException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\GiftPlan\Exception;
+
+use Exception;
+
+class GiftPlanNotFoundException extends Exception
+{
```
vitek-rostislav, line 10: This should extend HttpNotFoundException to get a proper 404 instead of 500 error :)

packages/framework/src/Model/Product/GiftPlan/GiftPlanDataFactory.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\GiftPlan;
+
+class GiftPlanDataFactory
+{
+    public function __construct()
```
vitek-rostislav: unnecessary constructor

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -2116,6 +2128,30 @@ msgstr "Vygenerováno"
 msgid "Generated batches"
 msgstr "Vygenerované dávky"
 
+msgid "Gift 1+1"
+msgstr "Dárek 1+1"
```
vitek-rostislav: This string is in the po files only

project-base/app/src/DataFixtures/Demo/CartDataFixture.php
```diff
@@ -54,6 +54,7 @@ public function getDependencies(): array
         return [
             ProductDataFixture::class,
             CustomerUserDataFixture::class,
+            GiftPlanDataFixture::class,
         ];
```
vitek-rostislav, line 58: this should not be necessary now, right? But I suppose you have added the dependency to avoid troubles in the future if anyone tries to add gifts to cart cata fixture?

project-base/app/src/DataFixtures/Demo/GiftPlanDataFixture.php
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * @param \Doctrine\Persistence\ObjectManager $manager
+     */
+    #[Override]
+    public function load(ObjectManager $manager)
+    {
+        foreach ($this->domain->getAllIds() as $domainId) {
+            $this->giftPlanSettingFacade->setGiftPriceWithVat(Money::create(10), $domainId);
+
```
vitek-rostislav, line 69: I would use `PriceConverter` (see e.g., `TransportDataFixture`) - that way, you would have more believeable data fixtures on all domains + moreover, tests should be more robust then

project-base/app/src/DataFixtures/Demo/GiftPlanDataFixture.php
```diff
@@ -0,0 +1,127 @@
...
+            $giftPlanData->domainId = $domainId;
+            $giftPlanData->validFrom = null;
+            $giftPlanData->validTo = new DateTime('now + 30 days');
+            $giftPlanData->giftProduct = $this->getReference(ProductDataFixture::PRODUCT_PREFIX . '72');
+            $giftPlanData->mainProducts = [$this->getReference(ProductDataFixture::PRODUCT_PREFIX . '14')];
+            $giftPlan = $this->createGiftPlan($giftPlanData);
+
+            if ($domainId === Domain::FIRST_DOMAIN_ID) {
+                $this->addReference(self::GIFT_PLAN_VALID, $giftPlan);
+            }
```
vitek-rostislav: I would say we could create references for all domains using `addReferenceForDomain` but I can live with this

project-base/app/src/DataFixtures/Demo/UploadedFileDataFixture.php
```diff
@@ -53,7 +54,7 @@ public function load(ObjectManager $manager): void
         $this->addUploadedFile(
             $this->getReference(ProductDataFixture::PRODUCT_PREFIX . 2),
             'example-file.pdf',
-            [$this->domain->getLocale() => 'Example file'],
+            [LocaleHelper::LOCALE_EN => 'Example file'],
```
vitek-rostislav: why this change? And if, why not using `$this->createUploadedFileTranslatedNames('Example file'),` as above?

packages/framework/src/Migrations/Version20250911060839.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        foreach ($this->getAllDomainConfigs() as $domainConfig) {
+            $sql = 'INSERT INTO setting_values (name, domain_id, value, type) VALUES
+                (\'giftInputPrice\', ' . $domainConfig->getId() . ', \'0\', \'string\')
+            ';
```
vitek-rostislav, line 25: I do not think it is an issue, however, we have now different default value here in the migration, and in `GiftPlanDataFixture`. Maybe it is intentional (I can imagine we rather want to test non-zero price, while the default value on a production should rather be zero)

packages/framework/src/Model/Cart/Watcher/CartWatcher.php
```diff
@@ -46,6 +49,16 @@ public function getModifiedPriceItemsAndUpdatePrices(Cart $cart)
             $cartItem->setWatchedPrice($price->getPriceWithVat());
         }
 
+        $giftPrice = $this->giftPlanSettingFacade->getInputGiftPrice($this->domain->getId());
+
+        foreach ($cart->getProductGiftCartItems() as $cartItem) {
+            if ($cartItem->getWatchedPrice()->equals($giftPrice)) {
```
vitek-rostislav: null pointer exception may occur here, btw, I recommend https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended- :wink:

vitek-rostislav: however, yes, a similar issue is in the current code a few lines above :shrug:

packages/framework/src/Model/Product/GiftPlan/GiftCartFacade.php
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            $newCartItem = $this->cartItemFactory->create($cart, $giftProduct, $giftQuantity, $giftPrice, CartItemTypeEnum::TYPE_PRODUCT_GIFT);
+            $cart->addItem($newCartItem);
+            $cart->setModifiedNow();
+            $this->em->persist($newCartItem);
+            $this->em->flush();
+        }
+
+        $this->em->flush();
```
vitek-rostislav: do we need to flush here? Seems unnecessary (what if there is no $giftCartItemSetup in the list?)

packages/framework/src/Model/Product/GiftPlan/GiftPlanRepository.php
```diff
@@ -0,0 +1,153 @@
...
+
+        return $giftPlan;
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $mainProduct
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Product\GiftPlan\GiftPlan[]
+     */
+    public function findActiveGiftPlansByMainProductAndDomainId(Product $mainProduct, int $domainId): array
```
vitek-rostislav, line 106: this is a query that will be called often within cart manipulations, right? I believe it is worth optimizing by adding proper indexes (at least composite of domain_id, valid_from, valid_to?). I would consult that with AI :smile:

packages/framework/src/Model/Product/GiftPlan/GiftCartFacade.php
```diff
@@ -0,0 +1,158 @@
...
+        return $giftsSetup;
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroup $pricingGroup
+     * @param int $domainId
+     * @return bool
+     */
+    protected function isGiftProductSellable(Product $product, PricingGroup $pricingGroup, int $domainId): bool
```
vitek-rostislav: this functionality is duplicated inside `GiftFlagSynchronizerFacade::shouldHaveGiftFlag`, we should have just one place where is this "isGiftProductSellable" logic implemented. Moreover, you will need to synchronize with https://github.com/shopsys/shopsys/pull/4173 - after that one is merged, the following scenario will have to be added to the "sellable gift resolution" - when a gift product has not allowed negative stock quantities and has stock quantity <= 0, it is not sellable as well

packages/framework/src/Model/Product/GiftPlan/RecalculateGiftFlagsCronModule.php
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\GiftPlan;
+
+use Monolog\Logger;
+use Override;
+use Shopsys\Plugin\Cron\SimpleCronModuleInterface;
+
+class RecalculateGiftFlagsCronModule implements SimpleCronModuleInterface
```
vitek-rostislav, line 11: I realized two things about this cron module:
1) it might not work as intended - products need to be recalculated before this is run (imagine gift's selling allowed expires at midnight) - you can not be sure all products are reculculated at the time you run this CRON (the recalculation is triggerred at midnight and is processed in rabbitMQ, so it might be still running)
2) you do not need that at all, as the automatic flag assignment is done within `ProductRecalculationFacade::recalculate()` which is triggered for all products at midnight (as mentioned above). So that one should do the job you need

packages/framework/src/Model/Product/GiftPlan/GiftCartFacade.php
```diff
@@ -0,0 +1,158 @@
...
+     * @return \Shopsys\FrameworkBundle\Model\Product\GiftPlan\GiftCartItemSetup[]
+     */
+    protected function getGiftCartItemSetupListByCart(Cart $cart, int $domainId): array
+    {
+        /** @var \Shopsys\FrameworkBundle\Model\Product\GiftPlan\GiftCartItemSetup[] $giftsSetup */
+        $giftsSetup = [];
+        $pricingGroup = $this->currentCustomerUser->getPricingGroup();
+
+        foreach ($cart->getProductCartItems() as $item) {
+            $giftPlans = $this->giftPlanFacade->findActiveGiftPlansByMainProductAndDomainId($item->getProduct(), $domainId);
```
vitek-rostislav: just an idea, I am wondering, if we could optimize that by creating a single query that would accept an array of products and return an array of sellable gifts indexed by the input products ids. I would be probably very complicated query, however, at the moment, it feels like there will be quite a lot of queries anytime we manipulate with cart items, lets talk about it, please

packages/framework/src/Model/Product/GiftPlan/GiftCartFacade.php
```diff
@@ -0,0 +1,158 @@
...
+                continue;
+            }
+
+            $giftQuantity = $giftCartItemSetup->getQuantity();
+
+            if ($giftQuantity === $item->getQuantity()) {
+                continue;
+            }
+
+            $item->changeQuantity($giftQuantity);
```
vitek-rostislav, line 170: this is the place where a gift quantity is changed when I change its main item quantity? I would not expect it in `getInvalidProductGiftCartItems` method :)

packages/framework/src/Model/Order/Processing/OrderInput.php
```diff
@@ -49,7 +50,23 @@ public function __construct(
      */
     public function addProduct(Product $product, int $quantity): void
     {
-        $this->products[] = new QuantifiedProduct($product, $quantity);
+        $quantifiedProduct = new QuantifiedProduct($product, $quantity);
+        $quantifiedProduct->setAdditionalData(QuantifiedProduct::CART_ITEM_TYPE_KEY, CartItemTypeEnum::TYPE_PRODUCT);
+
+        $this->products[] = $quantifiedProduct;
```
vitek-rostislav: I would rather call `addQuantifiedProduct()` here for clarity and consistency

packages/framework/tests/Unit/Model/Order/Processing/OrderProcessorMiddleware/AddProductGiftsMiddlewareTest.php
```diff
@@ -0,0 +1,99 @@
...
+            Money::create(2)->getAmount(),
+            $actualOrderData->getItemsByType(CartItemTypeEnum::TYPE_PRODUCT_GIFT)[0]->totalPriceWithVat->getAmount(),
+        );
+
+        $this->assertSame(Money::create(244)->getAmount(), $actualOrderData->totalPrice->getPriceWithVat()->getAmount());
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Item\QuantifiedItemPrice[] $quantifiedItemPrices
+     * @return \Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessorMiddleware\AddProductGiftsMiddleware
```
vitek-rostislav, line 72: nitpick, as we only have one product and one gift in this test, we do not need to pass `$quantifiedItemPrices` array, just one price, the mock setting could be then simplified to something like 
```php
$quantifiedProductPriceCalculation->method('calculateGiftPrice')->willReturn$quantifiedItemPrices);
```

(the same for `createAddProductsMiddleware` below)

packages/framework/src/Model/Product/GiftPlan/GiftPlanFacade.php
```diff
@@ -0,0 +1,89 @@
...
+        return $this->giftPlanRepository->findActiveGiftPlansByMainProductAndDomainId($mainProduct, $domainId);
+    }
+
+    /**
+     * @param int[] $mainProductIds
+     * @param int $domainId
+     * @return int[][]
+     */
+    public function findActiveGiftProductIdsByMainProductIds(array $mainProductIds, int $domainId): array
+    {
```
vitek-rostislav, line 95: ah, here it is, I think we could leverage this function in the cart as well, or?

packages/framework/src/Model/Product/GiftPlan/GiftPlanRepository.php
```diff
@@ -0,0 +1,125 @@
...
+    public function findActiveGiftProductIdsByMainProductIds(array $mainProductIds, int $domainId): array
+    {
+        if ($mainProductIds === []) {
+            return [];
+        }
+
+        $now = new DateTime();
+
+        // QueryBuilder podobný jako u findActiveGiftPlansByMainProductAndDomainId,
+        // ale vrací páry (mainProductId, giftProductId)
```
vitek-rostislav: english please :) moreover, it is probably unnecessary

packages/framework/src/Model/Product/GiftPlan/GiftPlanRepository.php
```diff
@@ -0,0 +1,125 @@
...
+            ->setParameter('mainProductIds', $mainProductIds)
+            ->setParameter('domainId', $domainId)
+            ->setParameter('now', $now)
+            ->orderBy('mp.id', 'ASC')
+            ->addOrderBy('gift.id', 'ASC');
+
+        $rows = $qb->getQuery()->getArrayResult();
+
+        // Připravíme výsledek s prázdnými poli pro všechny požadované hlavní produkty
+        $giftIdsIndexedByMainProductId = array_fill_keys(array_map('intval', $mainProductIds), []);
```
vitek-rostislav, line 153: unnecessary comment

packages/framework/src/Model/Product/GiftPlan/GiftPlanRepository.php
```diff
@@ -0,0 +1,125 @@
...
+
+        foreach ($rows as $row) {
+            $mainId = (int)$row['mainProductId'];
+            $giftId = (int)$row['giftProductId'];
+            $giftIdsIndexedByMainProductId[$mainId][] = $giftId;
+        }
+
+        // Odstranění případných duplicit (ochrana pro jistotu) a stabilní seřazení
+        foreach ($giftIdsIndexedByMainProductId as $mainId => $giftIds) {
```
vitek-rostislav, line 161: unnecessary comment

packages/framework/src/Model/Product/Search/FilterQueryFactory.php
```diff
@@ -338,4 +338,19 @@ protected function filterByCategory(FilterQuery $filterQuery, Category $category
...
         return $this->categoryAutomatedFilterFacade->applyFiltersByCategory($filterQuery, $category);
     }
+
+    /**
+     * @param int[] $productIds
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Product\Search\FilterQuery
+     */
+    public function createProductFilterQuery(array $productIds, int $domainId): FilterQuery
+    {
```
vitek-rostislav: Nitpick, maybe rather `createFilteredByProductIds`?

packages/frontend-api/src/Model/Product/BatchLoad/ProductElasticsearchBatchProvider.php
```diff
@@ -178,4 +178,20 @@ protected function getSellableByProductsIdsInCategoryFilterQuery(
...
         return $this->categoryAutomatedFilterFacade->applyFiltersByCategory($filterQuery, $category);
     }
+
+    /**
+     * @param int[][] $productsIds
+     * @param int $domainId
+     * @return array
+     */
+    public function getBatchedByProductsIds(array $productsIds, int $domainId): array
+    {
```
vitek-rostislav: now I noticed, we already have `getBatchedSellableByProductIds()` method here, could not we use that one? It uses existing `FilterQueryFactory::createSellableProductsByProductIdsFilter()` which looks very similar to your new `createProductFilterQuery()` method

packages/frontend-api/src/Model/Product/BatchLoad/ProductsBatchLoader.php
```diff
@@ -141,4 +147,45 @@ protected function sortByOriginalArray(array $arrayForSorting, array $originalAr
...
         return $sortedItems;
     }
+
+    /**
+     * @param int[] $productIds
+     * @return \GraphQL\Executor\Promise\Promise
+     */
+    public function loadProductGiftsByMainProductIds(array $productIds): Promise
+    {
+        $domainId = $this->domain->getId();
```
vitek-rostislav, line 157: This whole first part was quite hard to understand for me. Now I know, it basically handles the scenario, when none of the input products have any gift assigned. So after a few iterations with Claude, I suggest this:
```php
  $domainId = $this->domain->getId();
  $giftProductIdsIndexedByMainProductId = $this->giftPlanFacade->findActiveGiftProductIdsByMainProductIds($productIds, $domainId);

  // Check if no gifts exist at all
  if (array_filter($giftProductIdsIndexedByMainProductId) === []) {
      return $this->promiseAdapter->all(
          array_map(fn() => [], $productIds)
      );
  }
```

What do you think? For even better readability, we could wrap that `array_filter($giftProductIdsIndexedByMainProductId) === []` into a variable/method with descriptive name

packages/frontend-api/tests/Unit/Model/Product/ProductsBatchLoaderTest.php
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    public function testNoGiftsReturnsEmptyArraysInOrder(): void
+    {
+        $loader = $this->createLoader();
+
+        $productIds = [10, 20, 30];
+
+        // Facade vrátí mapu bez dárků pro všechny vstupní produkty
+        $this->giftPlanFacade->method('findActiveGiftProductIdsByMainProductIds')
```
vitek-rostislav, line 51: unneccesary (and Czech) comment

packages/frontend-api/tests/Unit/Model/Product/ProductsBatchLoaderTest.php
```diff
@@ -0,0 +1,214 @@
...
+            ->with($productIds, 1)
+            ->willReturn([
+                10 => [],
+                20 => [],
+                30 => [],
+            ]);
+
+        // Provider vůbec nemá být volán, protože nejsou žádná gift IDs
+        $this->productElasticsearchBatchProvider->expects($this->never())
```
vitek-rostislav, line 60: unneccesary (and Czech) comment

packages/frontend-api/tests/Unit/Model/Product/ProductsBatchLoaderTest.php
```diff
@@ -0,0 +1,214 @@
...
+            ->method('getBatchedByProductsIds');
+
+        $promise = $loader->loadProductGiftsByMainProductIds($productIds);
+
+        $result = $this->resolvePromise($promise);
+
+        // Výsledek musí mít stejné indexy (0..n-1) a prázdné seznamy
+        $this->assertSame([[], [], []], $result);
```
vitek-rostislav, line 68: unneccesary (and Czech) comment

packages/frontend-api/tests/Unit/Model/Product/ProductsBatchLoaderTest.php
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    public function testSharedGiftIdMappedToMultipleProductsWithoutDuplicatesAndMaintainsOrder(): void
+    {
+        $loader = $this->createLoader();
+
+        // Vstupní pořadí je důležité
+        $productIds = [50, 40, 60];
```
vitek-rostislav, line 76: unneccesary (and Czech) comment

packages/frontend-api/tests/Unit/Model/Product/ProductsBatchLoaderTest.php
```diff
@@ -0,0 +1,214 @@
...
+
+        // 100 je sdílený dárek pro 50 i 40, 110 jen pro 60
+        $this->giftPlanFacade->method('findActiveGiftProductIdsByMainProductIds')
```
vitek-rostislav, line 79: English please :)

packages/frontend-api/tests/Unit/Model/Product/ProductsBatchLoaderTest.php
```diff
@@ -0,0 +1,214 @@
...
+            ->with($productIds, 1)
+            ->willReturn([
+                50 => [100],
+                40 => [100],
+                60 => [110],
+            ]);
+
+        // Provider vrátí entity pro gift IDs, indexované dle konstanty
+        $this->productElasticsearchBatchProvider->expects($this->once())
```
vitek-rostislav, line 88: etc, I will not comment all the comments, just double check them, please, remove unnecessary, and translate the important ones

project-base/storefront/graphql/requests/cart/fragments/CartItemFragment.graphql
```diff
@@ -38,5 +38,8 @@ fragment CartItemFragment on CartItem {
             name
         }
         vatPercent
+        gifts {
+            ...ProductGiftFragment
```
vitek-rostislav: this is something I would like to discuss. In the cart you do not need to fetch the gifts of the items at all as you just get the gifts within the items themselves, right? However, you have added this into the fragment, because of the "add to cart" popup window, right? IMHO it adds an unnecessary load for the cart itself, as from now on, we will try to load all the gifts everywhere we access the cart items (probably at the cart page at least.). Let's discuss it, please

vitek-rostislav: check the data in the cart query response - you have gift items within the items, plus gifts under the product item - https://prnt.sc/KcGnO99ejVWc

project-base/storefront/components/Blocks/OrderItemGiftCard/OrderItemGiftCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+    availability: TypeAvailability;
+    mainImage?: TypeImageFragment | null;
+    fullName: string;
+    categoryName: string;
+    quantity: number;
+    unit: string | null;
+    price: TypeProductPriceFragment;
+};
+
+const OrderItemGiftCardComp: FC<OrderItemGiftCardProps> = ({
```
vitek-rostislav, line 22: that `Comp` suffix stands for "component"? I would remove it :)

project-base/storefront/graphql/requests/cart/fragments/CartItemFragment.graphql
vitek-rostislav: I even have an idea, how to rewrite it so you do not have to fetch `cartitem.product.gifts` - what about adding "added gift" into the "add product result"?

packages/framework/src/Model/Order/Item/OrderItem.php
```diff
@@ -409,6 +409,20 @@ public function setProduct($product): void
         $this->product = $product;
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product|null $productGift
+     */
+    public function setProductGift($productGift): void
+    {
```
vitek-rostislav, line 426: not related to this particular code, however, I need to put this somewhere. We should make sure that the gift product is set as a related order item for the main product order item, what do you think?

packages/framework/src/Migrations/Version20250915101119.php
```diff
@@ -0,0 +1,33 @@
...
+    use MultidomainMigrationTrait;
+
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $this->sql('INSERT INTO flags (uuid, visible, rgb_color) VALUES (\'3b709b97-604d-462d-b045-885f126e78d2\', true, \'#000000\')');
+        $flagId = $this->connection->lastInsertId();
```
vitek-rostislav, line 27: I am not sure whether I have already mentioned or not - it should not be possible to delete this flag - it is not resolved yet

packages/framework/src/Migrations/Version20250915101119.php
```diff
@@ -0,0 +1,33 @@
...
+
+        foreach ($this->getAllLocales() as $locale) {
+            $this->sql('INSERT INTO flag_translations (name, translatable_id, locale) VALUES (:name, :flagId, :locale)', [
+                'name' => 'Product gift',
+                'flagId' => $flagId,
```
vitek-rostislav, line 32: you cna use translator here, see e.g. `Version20180603135341`

packages/framework/src/Model/Product/GiftPlan/GiftFlagSynchronizerFacade.php
```diff
@@ -0,0 +1,175 @@
...
+
+        return false;
+    }
+
+    /**
+     * @return \Shopsys\FrameworkBundle\Model\Product\Flag\Flag
+     */
+    protected function getGiftFlag(): Flag
+    {
+        return $this->flagFacade->getByUuid(self::FLAG_GIFT_UUID);
```
vitek-rostislav: maybe we could add this into InMemoryCache? Now, when you recalculate 10000 products, it will be called 10000 times if I am not mistaken

packages/framework/src/Model/Product/GiftPlan/GiftFlagSynchronizerFacade.php
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        if (!$changed) {
+            return;
+        }
+
+        $this->em->flush();
+
+        $this->productRecalculationDispatcher->dispatchSingleProductId(
+            $mainProduct->getId(),
```
vitek-rostislav: lets dicsuss this, I believe we do not need to dispatch the product here, as the recalculation itself should already happen within recalculations in the consumer only

packages/framework/src/Model/Product/GiftPlan/GiftPlanEventSubscriber.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Model\Product\GiftPlan;
+
+use Override;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class GiftPlanEventSubscriber implements EventSubscriberInterface
```
vitek-rostislav, line 10: IMHO we should/could leverage existing `DispatchAffectedProductsSubscriber` and implement the logic there similarly to other entities changes that trigger the products recalculations

packages/framework/src/Model/Product/GiftPlan/GiftPlanRepository.php
```diff
@@ -0,0 +1,153 @@
...
+            $giftIdsIndexedByMainProductId[$mainId] = array_values(array_unique($giftIds));
+        }
+
+        return $giftIdsIndexedByMainProductId;
+    }
+
+    /**
+     * Vrátí všechny gift plány bez časového filtru.
```
vitek-rostislav: unnecessary czech comment

packages/framework/src/Model/Product/GiftPlan/GiftPlanRepository.php
```diff
@@ -0,0 +1,153 @@
...
+    public function findAll(): array
+    {
+        return $this->getRepository()->findAll();
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $giftProduct
+     * @return \Shopsys\FrameworkBundle\Model\Product\GiftPlan\GiftPlan[]
+     */
+    public function findByGiftProduct(Product $giftProduct): array
```
vitek-rostislav: unused method

packages/framework/src/Model/Product/Product.php
```diff
@@ -717,7 +717,7 @@ protected function setDomains(ProductData $productData)
      * @param int $domainId
      * @return \Shopsys\FrameworkBundle\Model\Product\ProductDomain
      */
-    protected function getProductDomain(int $domainId)
+    public function getProductDomain(int $domainId)
```
vitek-rostislav: maybe, we could keep this protected, and just add the setter for the flags into `Product` class itself. I do not insist on that

packages/frontend-api/src/Model/Resolver/Price/ProductPriceQuery.php
```diff
@@ -67,6 +76,35 @@ public function priceByProductQuery(Product|array $data): PriceInfo
...
+     */
+    public function giftPriceByProductQuery(Product|array $data): PriceInfo
+    {
+        $domainId = $this->domain->getId();
+
+        if ($data instanceof Product) {
+            $vat = $data->getVatForDomain($domainId);
+        } else {
+            $vatPercent = $data['vat_percent'];
+            $vatData = $this->vatDataFactory->create();
```
vitek-rostislav, line 91: this is quite an interesting solution, however, would not it be "better" to fetch the VAT object from DB by the rate?

vitek-rostislav (changes requested):

packages/framework/src/Model/Cart/Item/CartItemFactory.php
RostislavKreisinger: for me its new unknown feature, thx

packages/framework/src/Model/Cart/Cart.php
RostislavKreisinger: Renaming a function and making it protected is, in my opinion, an unnecessarily crude intervention in the interface. However, I took the note about where it is used into account and incorporated it.

packages/framework/src/Model/Order/Processing/OrderProcessorMiddleware/AddProductGiftsMiddleware.php
RostislavKreisinger: you are right, moved to OrderitemsDatafactory

project-base/storefront/graphql/requests/cart/fragments/CartItemFragment.graphql
RostislavKreisinger: I will create a separate fragment for addProductResult so that I will not needlessly query the gifts in the cart in CartItem.

packages/framework/src/Model/Product/GiftPlan/RecalculateGiftFlagsCronModule.php
RostislavKreisinger: removed

project-base/storefront/components/Blocks/OrderItemGiftCard/OrderItemGiftCard.tsx
RostislavKreisinger: suffix Comp is same as another component.

project-base/storefront/graphql/requests/cart/fragments/CartItemFragment.graphql
RostislavKreisinger: done

packages/framework/src/Model/Order/Item/OrderItem.php
RostislavKreisinger: it is not related item, is separete item

packages/framework/src/Model/Product/GiftPlan/GiftPlanEventSubscriber.php
RostislavKreisinger: i am for separate services because is little bit different usecase.

packages/frontend-api/src/Model/Resolver/Price/ProductPriceQuery.php
RostislavKreisinger: i dont want do some system changing solution, this is here sufficient solution.

packages/framework/src/Model/Product/GiftPlan/GiftFlagSynchronizerFacade.php
RostislavKreisinger: changed

packages/framework/src/Model/Product/GiftPlan/GiftCartFacade.php
```diff
@@ -0,0 +1,157 @@
...
+
+        return $productVisibility->isVisible() && !$product->getCalculatedSellingDenied();
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Cart\Cart $cart
+     * @param int $domainId
+     * @return array
+     */
+    public function getRefreshProductGiftCartItemsAndGetInvalidItems(Cart $cart, int $domainId): array
```
vitek-rostislav: you can remove the first "get" from the method name -> `refreshProductGiftCartItemsAndGetInvalidItems`

packages/framework/src/Model/Product/GiftPlan/GiftFlagSynchronizerFacade.php
```diff
@@ -0,0 +1,179 @@
...
+
+        foreach ($giftPlans as $giftPlan) {
+            $this->dispatchMainProductForGriftFlagRecalculation($giftPlan->getMainProducts());
+        }
+    }
+
+    /**
+     * @param int $mainProductId
+     */
+    public function recalculateGiftFlagForMainProductIdAndDomainId(int $mainProductId): void
```
vitek-rostislav: there should not be "...andDomainId" in the function name anymore :)

packages/framework/src/Model/Product/GiftPlan/GiftFlagSynchronizerFacade.php
```diff
@@ -0,0 +1,179 @@
...
+use Shopsys\FrameworkBundle\Model\Product\Flag\FlagFacade;
+use Shopsys\FrameworkBundle\Model\Product\Product;
+use Shopsys\FrameworkBundle\Model\Product\ProductFacade;
+use Shopsys\FrameworkBundle\Model\Product\ProductVisibilityFacade;
+use Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationDispatcher;
+use Shopsys\FrameworkBundle\Model\Product\Recalculation\ProductRecalculationPriorityEnum;
+
+class GiftFlagSynchronizerFacade
+{
+    public const FLAG_GIFT_UUID = '3b709b97-604d-462d-b045-885f126e78d2';
```
vitek-rostislav: public const string :)

vitek-rostislav (approved):

project-base/storefront/components/Blocks/Popup/AddToCartPopup.tsx
```diff
@@ -85,6 +86,11 @@ export const AddToCartPopup: FC<AddToCartPopupProps> = ({ key, addedCartItem: {
                     </div>
                 </div>
             </div>
+            <div>
+                {product.gifts.length > 0 && <p className="h3 mb-3">{t('Gifts')}</p>}
+                {product.gifts.length > 0 &&
+                    product.gifts.map((gift, index) => <ProductGift key={index} gift={gift} />)}
```
JanMolcik: I would rather wrap it under a single condition 

```tsx
{product.gifts.length > 0 && (
    <>
        <p className="h3 mb-3">{t('Gifts')}</p>
        {product.gifts.map((gift, index) => (
            <ProductGift key={index} gift={gift} />
        ))}
    </>
)}
```

project-base/storefront/components/Blocks/Product/ProductGift.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const ProductGift: FC<ProductGiftProps> = ({ gift, variant = 'default' }) => {
+    const { t } = useTranslation();
+    const [firstImage] = gift.images;
+    const mainImage = gift.images.length ? firstImage : undefined;
+
+    return (
+        <div
+            className={twJoin(
+                'bg-backgroundMore relative flex items-center rounded-xl',
```
JanMolcik: Typo - use `bg-background-more` instead.

project-base/storefront/components/Pages/Cart/CartList/CartListItem.tsx
```diff
@@ -102,26 +105,37 @@ export const CartListItem: FC<CartListItemProps> = ({
...
+                            size="large"
+                            step={1}
+                            onChangeValueCallback={setSpinboxValue}
+                        />
+                    ) : (
+                        <div className="min-w-[100px] text-center">{quantity}</div>
+                    )}
 
-                    {isPriceVisible(product.price.priceWithVat) && (
+                    {isPriceVisible(product.price.priceWithVat) && isProduct && (
```
JanMolcik: I would move the condition for `isProduct` and `isProductGift` to the front as it is easier to follow right away (the same as it is in `CartInHeaderListItem`).

project-base/storefront/components/Pages/ProductDetail/ProductDetailContent.tsx
```diff
@@ -66,6 +67,15 @@ export const ProductDetailContent: FC<ProductDetailContentProps> = ({ product, i
...
 
                         <div className="bg-background-more flex flex-col gap-4 rounded-xl p-3 sm:p-6">
+                            {product.gifts.length > 0 && <p className="h3 mb-3">{t('Gifts')}</p>}
+                            {product.gifts.length > 0 && (
+                                <div>
+                                    {product.gifts.map((gift, index) => (
+                                        <ProductGift key={index} gift={gift} />
+                                    ))}
+                                </div>
+                            )}
```
JanMolcik: Same as in https://github.com/shopsys/shopsys/pull/4193#discussion_r2374792850

JanMolcik (approved):

PR #4194, opened 2025-09-17
Promotion X + Y for free

TomasLudvik:
New feeature where administrator can set on product promotion e.g. 3 + 2 free. This means That if customer ads 3 products to cart, he is then eligible to add another 2 for free.

Products with such setting have automatic flag with the name of promotion. 

----
  - https://tl-deal-x-y-free-codex.odin.shopsys.cloud
  - https://cz.tl-deal-x-y-free-codex.odin.shopsys.cloud
  - https://tl-deal-x-y-free-codex.odin.shopsys.cloud/sk


packages/framework/src/Form/Admin/Product/ProductFormType.php
```diff
@@ -602,6 +610,52 @@ private function createPricesGroup(FormBuilderInterface $builder, ?Product $prod
...
+    private function createPromotionGroup(FormBuilderInterface $builder, ?Product $product): FormBuilderInterface
+    {
+        $promotionGroup = $builder->create('promotionGroup', GroupType::class, [
+            'label' => t('Promotion X + Y free'),
+        ]);
+
+        if ($this->isProductMainVariant($product)) {
+            $promotionGroup->add('promotionX', DisplayOnlyType::class, [
+                'mapped' => false,
+                'required' => false,
```
grossmannmartin: DisplayOnlyType is automatically not required and not mapped

packages/framework/src/Form/Admin/Product/ProductFormType.php
```diff
@@ -181,6 +183,12 @@ public function configureOptions(OptionsResolver $resolver): void
                 'data_class' => ProductData::class,
                 'attr' => ['novalidate' => 'novalidate'],
                 'csrf_token_id' => self::CSRF_TOKEN_ID,
+                'constraints' => [
+                    new RequiredPair([
```
grossmannmartin: I would rather use the `when` condition instead of class level constraint. 
This one is hard to find, and the UX is not good ( what are promotionX and promotionY?)

Snímek obrazovky 2025-09-18 v 8 32 03


e.g. this as a constraint for both fields (just change the field id)
```
new Constraints\When(
    expression: 'this.getParent().get("promotionY").getData() !== null',
    constraints: [
        new Constraints\NotBlank(),
    ],
),
```

packages/framework/src/Migrations/Version20250910120000.php
```diff
@@ -0,0 +1,25 @@
...
+class Version20250910120000 extends AbstractMigration
+{
+    /**
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE products ADD COLUMN promotion_x INT DEFAULT NULL');
+        $this->sql('ALTER TABLE products ADD COLUMN promotion_y INT DEFAULT NULL');
```
grossmannmartin: Wouldn't be more future-proof to create a separate entity for this promotion? If there is a request for time/domain/... limitation, the structure will be already ready.

And as a bonus we could create a foreign key to flags and don't rely on comparing x and y of two entities

packages/framework/src/Model/Order/Item/OrderItemData.php
```diff
@@ -119,6 +119,24 @@ public function getTotalPrice(): PriceInterface
         return new Price($this->totalPriceWithoutVat, $this->totalPriceWithVat);
     }
 
+    /**
+     * @return \Shopsys\FrameworkBundle\Model\Pricing\PriceInterface
+     */
+    public function getTotalPriceIncludingAppliedDiscounts(): PriceInterface
```
grossmannmartin: this method breaks the purpose of middleware processing. 
It creates another place where some counting logic is applied.  
On change, this has to be updated too and it creates hidden dependency on the order of middlewares. 

This should be handled exclusively by middlewares

packages/framework/src/Model/Order/Processing/OrderProcessorMiddleware/ApplyPromotionMiddleware.php
```diff
@@ -0,0 +1,121 @@
...
+use Shopsys\FrameworkBundle\Component\Translation\Translator;
+use Shopsys\FrameworkBundle\Model\Order\Item\OrderItemData;
+use Shopsys\FrameworkBundle\Model\Order\Item\OrderItemDataFactory;
+use Shopsys\FrameworkBundle\Model\Order\Item\OrderItemTypeEnum;
+use Shopsys\FrameworkBundle\Model\Order\OrderData;
+use Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessingData;
+use Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessingStack;
+use Shopsys\FrameworkBundle\Model\Product\Product;
+
+class ApplyPromotionMiddleware implements OrderProcessorMiddlewareInterface
```
grossmannmartin: I would use more specific name (what kind of promotion?)

packages/framework/src/Model/Order/PlaceOrderFacade.php
```diff
@@ -56,7 +56,9 @@ public function placeOrder(
         ?string $deliveryAddressUuid = null,
     ): Order {
         foreach ($orderData->getItemsByType(OrderItemTypeEnum::TYPE_DISCOUNT) as $discount) {
-            $discount->promoCode->decreaseRemainingUses();
+            if ($discount->promoCode !== null) {
```
grossmannmartin: Wouldn't be better to have a separate order item type for this promotion? 
It will be more versatile and unified with gifts implemented in #4193

packages/framework/src/Model/Product/Elasticsearch/Scope/ProductExportFieldProvider.php
```diff
@@ -56,6 +56,8 @@ class ProductExportFieldProvider
 
     public const string SELLING_FROM = 'selling_from';
     public const string PRODUCT_VIDEOS = 'product_videos';
+    public const string PROMOTION_X = 'promotion_x';
+    public const string PROMOTION_Y = 'promotion_y';
```
grossmannmartin: ~Do we actually need this in elasticsearch and graphql?~
~I don't see any usage, and I can't think of the use case for this.~

~The actual info we need is already provided from the flag / cart item / order item.~


Edit: oh, now I see the usage... Then I would like to change the name. Promotion X and Promotion Y do not tell much. What about `promotionBuyQuantity` and `promotionFreeQuantity` or maybe `promotion { buyQuantity, freeQuantity }`

packages/frontend-api/src/Model/Resolver/Products/DataMapper/ProductEntityFieldMapper.php
```diff
@@ -357,4 +357,22 @@ public function getVatPercent(Product $product): string
...
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Product $product
+     * @return int|null
+     */
+    public function getPromotionY(Product $product): ?int
+    {
+        return $product->getPromotionY();
+    }
```
grossmannmartin, line 386: These two methods are not necessary as they're named the same as a method in the Product entity.  

(but if we rename the field, we would need the method)

project-base/storefront/components/Blocks/OrderItemProductCard/OrderItemProductPrice.tsx
```diff
@@ -37,9 +43,11 @@ export const OrderItemProductPrice: FC<OrderItemProductPriceProps> = ({ productP
                     {formatPrice(mapPriceForCalculations(productPrice.basicPrice.priceWithVat) * quantity)}
                 </div>
 
-                {isSpecialPrice && (
+                {(isSpecialPrice || (freeQuantity !== null && freeQuantity > 0)) && (
                     <div className="font-secondary text-price-discounted font-bold whitespace-nowrap">
-                        {formatPrice(mapPriceForCalculations(productPrice.priceWithVat) * quantity)}
+                        {formatPrice(
+                            mapPriceForCalculations(productPrice.priceWithVat) * (quantity - (freeQuantity || 0)),
+                        )}
```
grossmannmartin, line 52: The price with the applied promotion is not clear. See the difference between special price and promotion. I think they should be unified

Snímek obrazovky 2025-09-18 v 11 02 06

project-base/storefront/graphql/requests/products/fragments/ListedProductFragment.graphql
```diff
@@ -30,6 +30,8 @@ fragment ListedProductFragment on Product {
         __typename
         name
     }
+    promotionX
+    promotionY
```
grossmannmartin: I don't think we need those values in lists. We should request only the fields that are used on the storefront

project-base/storefront/public/locales/cs/common.json
```diff
@@ -54,6 +58,10 @@
     "Brands": "Značky",
     "Breadcrumb navigation": "Navigace z kolečka",
     "But at other addresses we have a lot for you...": "Ale na jiných adresách toho pro Vás máme spousty...",
+    "Buy {{ count }} pieces_one": "Nakupte {{ count }} kus",
+    "Buy {{ count }} pieces_few": "Nakupte {{ count }} kusy",
+    "Buy {{ count }} pieces_many": "Nakupte {{ count }} kusů",
+    "Buy {{ count }} pieces_other": "Nakupte {{ count }} kusů",
```
grossmannmartin, line 59: Wouldn’t it be better to merge these two strings into one sentence? It might be easier for translators to handle it as a whole rather than as two separate fragments.

grossmannmartin (changes requested):

project-base/storefront/public/locales/cs/common.json
TomasLudvik: Well there is problem with the plural settings as it can be set only once per translation

packages/framework/src/Model/AdvancedSearch/Filter/ProductHasPromotionXyFilter.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    #[Override]
+    public function getValueFormOptions(): array
+    {
+        return [
+            'expanded' => false,
+            'multiple' => false,
+            'choices' => [
+                t('Yes') => 1,
+                t('No') => 0,
```
grossmannmartin: I must confess, the advanced filter is kinda confusing to me. 
I would expect similar behavior as `Excluded from sale` - just simple is/not (has/not?)

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
```diff
@@ -0,0 +1,140 @@
...
+            $promotionXyData->freeQuantity = $y;
+            $promotionXy = $this->productPromotionXyFactory->create($promotionXyData);
+            $this->em->persist($promotionXy);
+            $this->em->flush();
+        }
+
+        $flagData = $this->flagDataFactory->create();
+        $flagData->visible = true;
+        $flagData->rgbColor = '#efd6ff';
+        $flagData->promotionXy = $promotionXy;
```
grossmannmartin, line 124: this feels arbitrary 🤔 - I mean the color.

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
```diff
@@ -0,0 +1,140 @@
...
+    public function findOrCreatePromotionFlag(int $x, int $y): Flag
+    {
+        $flag = $this->productPromotionXyRepository->findFlagByQuantities($x, $y);
+
+        if ($flag !== null) {
+            return $flag;
+        }
+
+        $promotionXy = $this->productPromotionXyRepository->findPromotionXyByQuantities($x, $y);
+
```
grossmannmartin, line 111: Not quite sure where the proper place to comment is, but:

When the promotion in the product detail is filled, two separate rows in the table product_promotion_xy are created. One is then used for the flag, the other for the product. 
When I change the value of X and Y on the product detail, a new row is always created. The previous row will remain as an orphan in the database.

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
```diff
@@ -0,0 +1,140 @@
...
+        if ($promotionBuyQuantity === null || $promotionFreeQuantity === null) {
+            return;
+        }
+
+        $flag = $this->findOrCreatePromotionFlag($promotionBuyQuantity, $promotionFreeQuantity);
+
+        foreach ($this->domain->getAllIds() as $domainId) {
+            $productData->flagsByDomainId[$domainId] = $productData->flagsByDomainId[$domainId] ?? [];
+
```
grossmannmartin, line 79: `$productData->flagsByDomainId[$domainId]` has to be an array already due to the first foreach.

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
```diff
@@ -0,0 +1,140 @@
...
+            if (!in_array($flag, $productData->flagsByDomainId[$domainId], true)) {
+                $productData->flagsByDomainId[$domainId][] = $flag;
```
grossmannmartin: flag will never be in the array because promotion flags are removed in the first foreach, or?

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
```diff
@@ -0,0 +1,140 @@
...
+            $locale,
+        );
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\Flag\Flag $flag
+     * @return bool
+     */
+    protected function isPromotionFlag(Flag $flag): bool
+    {
```
grossmannmartin, line 162: This method seems redundant to me. If I already have the Flag object, I can simply call hasPromotionXy()

packages/framework/src/Model/Product/Product.php
```diff
@@ -378,6 +385,36 @@ public function getEan()
...
+        if ($this->promotionXy === null) {
+            return 0;
+        }
+
+        $buyQuantity = $this->promotionXy->getBuyQuantity();
+        $freeQuantity = $this->promotionXy->getFreeQuantity();
+
+        $totalPromotionsSize = $buyQuantity + $freeQuantity;
+
+        $numberOfAppliedFullPromotions = intdiv($quantity, $totalPromotionsSize);
```
grossmannmartin, line 398: I can set promotion as "Buy 0 and get 0 for free". This then break adding product to the cart. 
I'm not sure if it's just because of this, but this fails due to the division by zero

grossmannmartin (changes requested):

packages/framework/src/Model/Product/ProductDataFactory.php
```diff
@@ -76,6 +78,7 @@ public function create(): ProductData
     {
         $productData = $this->createInstance();
         $this->fillNew($productData);
+        $productData->promotionXyData = $this->productPromotionXyDataFactory->create();
```
grossmannmartin: does not this belong into a fillNew() method?

packages/framework/src/Model/Product/ProductDataFactory.php
```diff
@@ -183,6 +186,9 @@ protected function fillFromProduct(ProductData $productData, Product $product):
         $productData->sellingTo = $product->getSellingTo();
         $productData->sellingDenied = $product->isSellingDenied();
         $productData->unit = $product->getUnit();
+        $productData->promotionXyData = $product->getPromotionXy() === null ?
+            $this->productPromotionXyDataFactory->create() :
+            $this->productPromotionXyDataFactory->createFromEntity($product->getPromotionXy());
```
grossmannmartin: when product has no promotion, why set the data to empty promotion? shouldn't be this fact represented by null value?

packages/framework/src/Model/Product/ProductFacade.php
```diff
@@ -194,6 +206,42 @@ public function edit(
...
+        if ($promotionBuyQuantity === null || $promotionFreeQuantity === null) {
+            if ($currentPromotion !== null) {
+                $product->setPromotionXy(null);
+                $this->em->remove($currentPromotion);
+            }
+
+            return;
+        }
+
+        $productPromotionXyData = $this->productPromotionXyDataFactory->create();
```
grossmannmartin, line 247: This is probably the reason why promotion is always created as mentioned earlier

packages/framework/src/Model/Product/ProductPromotionXy.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->setData($productPromotionXyData);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Product\ProductPromotionXyData $productPromotionXyData
+     * @return \Shopsys\FrameworkBundle\Model\Product\ProductPromotionXy
+     */
+    public static function create(ProductPromotionXyData $productPromotionXyData)
+    {
+        return new static($productPromotionXyData);
```
grossmannmartin: I'm not quite sure about this method. I think we should use only constructor and factory methods to create objects. Not the static create method

packages/framework/src/Model/Product/ProductPromotionXyData.php
```diff
@@ -0,0 +1,24 @@
...
+    public $buyQuantity;
+
+    /**
+     * @var int|null
+     */
+    public $freeQuantity;
+
+    public function __construct()
+    {
+        $this->buyQuantity = null;
```
grossmannmartin: null is inherently set as a default value if the property does not has an explicit type

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -541,6 +541,9 @@ msgstr "Hromadná úprava byla provedena"
 msgid "Bulk promo code generation"
 msgstr "Hromadné generování kupónu"
 
+msgid "Buy quantity"
+msgstr ""
```
grossmannmartin: missing translation

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -2071,6 +2074,9 @@ msgstr "Kategorie čtvrtého sloupce"
 msgid "Free from order value"
 msgstr "Zdarma od hodnoty objednávky"
 
+msgid "Free quantity"
+msgstr ""
```
grossmannmartin: missing translation

grossmannmartin (changes requested):

project-base/app/config/packages/shopsys_framework.yaml
```diff
@@ -8,6 +8,7 @@ shopsys_framework:
         processing_middlewares:
             - 'Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessorMiddleware\SetCustomerUserMiddleware'
             - 'Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessorMiddleware\AddProductsMiddleware'
+            - 'Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessorMiddleware\ApplyPromotionXyMiddleware'
```
grossmannmartin, line 9: and add the upgrade notes for storefront and backend, please 😇

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
TomasLudvik: I have reworked it

packages/framework/src/Model/Product/Product.php
TomasLudvik: I have added validation to form type

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
TomasLudvik: I have moved it to the constant

packages/framework/src/Model/Product/Flag/PromotionFlagFacade.php
grossmannmartin: Can you double-check this point, please? 

Currently:
- after loading demo data, there are two rows representing 3+1, even though only one product has such a promotion (one is assigned to the product, the other one to a flag)
- when I add a promotion to the product, two rows are still created

grossmannmartin (approved):

grossmannmartin (approved):

pk16011990 (approved):

packages/framework/src/Model/AdvancedSearch/Filter/ProductCalculatedSellingDeniedFilter.php
```diff
@@ -54,17 +64,29 @@ public function getValueFormOptions(): array
...
+     * @param \Shopsys\FrameworkBundle\Model\AdvancedSearch\AdvancedSearchRuleData[] $rulesData
      */
     #[Override]
     public function extendQueryBuilder(QueryBuilder $queryBuilder, array $rulesData): void
     {
-        foreach ($rulesData as $index => $ruleData) {
-            $sellingDenied = $ruleData->operator === self::OPERATOR_IS;
+        $subQueryBuilder = $this->em->createQueryBuilder();
+        $subQueryBuilder
+            ->select('COUNT(pds.id)')
```
grossmannmartin: We could use 
`select('1')` and later in rulesData foreach 
```
if ($ruleData->operator === self::OPERATOR_IS) {
        $qb->andWhere("EXISTS ($existsDql)");
    }
// NOT EXISTS in the other branch
```

packages/framework/src/Resources/views/Admin/Content/Product/sellability.html.twig
```diff
@@ -0,0 +1,4 @@
+<div class="in-product-visibility__title text-center">
+    {{ 'Sellability on each domain'|trans }}
+</div>
+{% include '@ShopsysFramework/Admin/Content/Product/sellabilityItems.html.twig' with { product: product, domains: domains, sellabilityPerDomainIds: sellabilityPerDomainIds } only %}
```
grossmannmartin: Why is it split into separate templates?

grossmannmartin (approved):

packages/framework/src/Resources/views/Admin/Content/Product/sellability.html.twig
vitek-rostislav: not anymore :slightly_smiling_face:

packages/framework/src/Model/AdvancedSearch/Filter/ProductCalculatedSellingDeniedFilter.php
vitek-rostislav: nice, rewritten in e04762511d

packages/administration/templates/content/product/listGrid.html.twig
```diff
@@ -38,6 +38,34 @@
...
+        {% set sellabilityClass = 'text-green' %}
+        {% set sellabilityIcon = 'cart' %}
+        {% set sellabilityTitle = 'Sellable'|trans %}
+    {% elseif row.sellingDeniedDomainsCount < getDomainsCount() and row.sellingDeniedDomainsCount > 0 %}
+        {% set sellabilityClass = 'text-orange' %}
+        {% set sellabilityIcon = 'cart' %}
+        {% set sellabilityTitle = 'Sellable'|trans %}
+    {% else %}
+        {% set sellabilityClass = 'text-secondary' %}
+        {% set sellabilityIcon = 'cart' %}
```
grossmannmartin: shoudn't we use cart-off icon here? so it's unified with the icon in the dropdown

packages/administration/templates/content/product/visibilityItems.html.twig
```diff
@@ -30,10 +47,16 @@
             <div class="col">
                 {% if isVisible %}
                     <a href="{{ url }}" class="" target="_blank">
```
grossmannmartin: this was probably my error from earlier, but empty class does not make sense imho

grossmannmartin (approved):

packages/administration/templates/content/product/listGrid.html.twig
vitek-rostislav: yes, good catch, fixed, thanks :ok_hand:

PR #4209, opened 2025-09-29
Name is no longer mandatory for Category, Blog Category and Blog Article

TomasLudvik:

- We had weird behavior where some form required name for entities but by current locale of administratort. This behavior has been now unified and all such entities now no longer require name, but are not visible in such case.

- Blog categories now have same visibility icons as product categories.

- Blog category visibility is now also calculated for root categories, that until now has always been visible.

- Unified displaying of information, about missing translation for current locale of all entities.

----
  - https://tl-fix-requiring-name.odin.shopsys.cloud
  - https://cz.tl-fix-requiring-name.odin.shopsys.cloud
  - https://tl-fix-requiring-name.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
+++ b/packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
@@ -170,9 +170,7 @@ final class BlogArticleFormType extends AbstractType
 
         $builderSettingsGroup
             ->add('names', LocalizedType::class, [
-                'main_constraints' => [
-                    new Constraints\NotBlank(['message' => 'Please enter name']),
-                ],
+                'required' => false,
                 'entry_options' => [
                     'required' => false,
                     'constraints' => [
--- a/packages/framework/src/Form/Admin/Blog/BlogCategoryFormType.php
+++ b/packages/framework/src/Form/Admin/Blog/BlogCategoryFormType.php
@@ -124,9 +124,7 @@ final class BlogCategoryFormType extends AbstractType
 
         $builderSettingsGroup
             ->add('names', LocalizedType::class, [
-                'main_constraints' => [
-                    new Constraints\NotBlank(['message' => 'Please enter name']),
-                ],
+                'required' => false,
                 'entry_options' => [
                     'required' => false,
                     'constraints' => [
--- a/packages/framework/src/Form/Admin/Category/CategoryFormType.php
+++ b/packages/framework/src/Form/Admin/Category/CategoryFormType.php
@@ -119,9 +119,7 @@ final class CategoryFormType extends AbstractType
 
         $builderSettingsGroup
             ->add('name', LocalizedType::class, [
-                'main_constraints' => [
-                    new Constraints\NotBlank(['message' => 'Please enter name']),
-                ],
+                'required' => false,
                 'entry_options' => [
                     'required' => false,
                     'constraints' => [
--- a/packages/framework/src/Twig/ProductExtension.php
+++ b/packages/framework/src/Twig/ProductExtension.php
@@ -32,7 +32,6 @@ class ProductExtension extends AbstractExtension
     {
         return [
             new TwigFilter('productDisplayName', $this->getProductDisplayName(...)),
-            new TwigFilter('productListDisplayNameByName', $this->getProductListDisplayNameByName(...)),
         ];
     }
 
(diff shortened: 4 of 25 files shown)


packages/framework/src/Twig/ProductExtension.php
```diff
@@ -85,19 +84,6 @@ public function getProductDisplayName(Product $product)
         return $product->getName();
     }
 
-    /**
-     * @param string|null $name
-     * @return string
-     */
-    public function getProductListDisplayNameByName(?string $name): string
```
vitek-rostislav, line 92: missing upgrade notes for this removal

packages/framework/src/Form/Admin/Category/CategoryFormType.php
```diff
@@ -120,9 +120,6 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
 
         $builderSettingsGroup
             ->add('name', LocalizedType::class, [
-                'main_constraints' => [
-                    new Constraints\NotBlank(['message' => 'Please enter name']),
```
vitek-rostislav, line 123: it should have now `'required' => false,` even in the options for the `LocalizedType` field itself, to avoid rendering the `*` (see https://prnt.sc/Nnzl2wdbBnrA)

vitek-rostislav: the screenshot is actually from BlogCategory form type, however, it applies to both of them :)

packages/framework/src/Form/Admin/Blog/BlogCategoryFormType.php
```diff
@@ -126,9 +126,6 @@ private function createSettingsGroup(FormBuilderInterface $builder, array $optio
 
         $builderSettingsGroup
             ->add('names', LocalizedType::class, [
-                'main_constraints' => [
-                    new Constraints\NotBlank(['message' => 'Please enter name']),
```
vitek-rostislav, line 128: should be now set `'required' => false` here

packages/framework/src/Model/Blog/Category/BlogCategoryVisibilityRepository.php
```diff
@@ -46,38 +46,13 @@ public function refreshCategoriesVisibility(): void
      */
     protected function refreshBlogCategoriesVisibilityOnDomain(DomainConfig $domainConfig): void
     {
-        $this->setRootBlogCategoryVisibleOnDomain($domainConfig);
-
```
vitek-rostislav, line 50: it feels like this whole commit is not necessary, let's talk about it, please. The thing is, the root category is the one that is never presented to users, it is just the root of the nested set tree in the database...

packages/framework/src/Model/Blog/Category/BlogCategoryVisibilityRepository.php
```diff
@@ -108,12 +83,12 @@ protected function refreshBlogCategoriesVisibilityOnDomainAndLevel(DomainConfig
                     AND
                     bct.name IS NOT NULL
                     AND
-                    parent_bcd.visible = TRUE
+                    (parent_bcd.visible = TRUE OR parent_bcd.visible IS NULL)
                 )
```
vitek-rostislav, line 112: this `parent_bcd.visible IS NULL` is now quite confusing as `visible` is not nullable column (I know it comes from the JOIN clause, right But still, I qould rather not do this change)

packages/framework/src/Model/Blog/Category/BlogCategoryVisibilityRepository.php
vitek-rostislav: btw until now, it was unified with CategoryVisibilityRepository

packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
```diff
@@ -182,9 +182,6 @@ private function createSettingsGroup(
 
         $builderSettingsGroup
             ->add('names', LocalizedType::class, [
-                'main_constraints' => [
-                    new Constraints\NotBlank(['message' => 'Please enter name']),
```
vitek-rostislav, line 174: so please add `'required' => false` here as well

vitek-rostislav (changes requested):

upgrade-notes/backend_20251002_092325.md
```diff
@@ -0,0 +1,3 @@
+#### Name is no longer mandatory for Category, Blog Category and Blog Article ([#4209](https://github.com/shopsys/shopsys/pull/4209))
+
+- Twig filter `productListDisplayNameByName` has been removed use `nameWithFallbackOnEmpty`
```
vitek-rostislav:
```suggestion
- Twig filter `productListDisplayNameByName` has been removed, use `nameWithFallbackOnEmpty` instead
```

vitek-rostislav (approved):

pk16011990 (approved):

PR #4216, opened 2025-10-03
Heureka categories for SK domains

sspooky13:
I implemented Heureka categories for SK domains, where these categories should be downloaded from [SK Heureka](https://www.heureka.sk/direct/xml-export/shops/heureka-sekce.xml) instead of [CZ Heureka](https://www.heureka.cz/direct/xml-export/shops/heureka-sekce.xml). Logic is separated from CZ heureka categories, and admin can set SK heureka categories if eshop has at least one SK domain

#### Fixes issues
... 

----
  - https://pt-heureka-categories.odin.shopsys.cloud
  - https://cz.pt-heureka-categories.odin.shopsys.cloud
  - https://pt-heureka-categories.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Migrations/Version20260223133152.php
+++ b/packages/framework/src/Migrations/Version20260223133152.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Override;
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20260223133152 extends AbstractMigration
+{
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE heureka_category ADD locale VARCHAR(255) NOT NULL DEFAULT \'cs\'');
+        $this->sql('ALTER TABLE heureka_category ALTER locale DROP DEFAULT');
+
+        $this->sql('ALTER TABLE heureka_category ADD heureka_id INT NOT NULL DEFAULT 0');
+        $this->sql('ALTER TABLE heureka_category ALTER heureka_id DROP DEFAULT');
+        $this->sql('UPDATE heureka_category SET heureka_id = id');
+
+        $this->sql('CREATE SEQUENCE heureka_category_id_seq');
+        $this->sql('SELECT setval(\'heureka_category_id_seq\', (SELECT MAX(id) FROM heureka_category))');
+        $this->sql('ALTER TABLE heureka_category ALTER id SET DEFAULT nextval(\'heureka_category_id_seq\')');
+
+        $this->sql('DROP INDEX uniq_fe112a6912469de2;');
+
+        $this->sql('CREATE UNIQUE INDEX uq_heureka_category_heureka_id_locale ON heureka_category (locale, heureka_id)');
+        $this->sql('CREATE INDEX IDX_FE112A6912469DE2 ON heureka_category_categories (category_id)');
+    }
+}
--- a/packages/product-feed-heureka/src/Form/CategoryCrudExtension.php
+++ b/packages/product-feed-heureka/src/Form/CategoryCrudExtension.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace Shopsys\ProductFeed\HeurekaBundle\Form;
 
 use Override;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\Plugin\PluginCrudExtensionInterface;
 use Shopsys\ProductFeed\HeurekaBundle\Model\HeurekaCategory\HeurekaCategory;
+use Shopsys\ProductFeed\HeurekaBundle\Model\HeurekaCategory\HeurekaCategoryDownloader;
 use Shopsys\ProductFeed\HeurekaBundle\Model\HeurekaCategory\HeurekaCategoryFacade;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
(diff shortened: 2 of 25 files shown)


packages/framework/src/Migrations/Version20260223133152.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE heureka_category ADD locale VARCHAR(255) NOT NULL DEFAULT \'cs\'');
+        $this->sql('ALTER TABLE heureka_category ALTER locale DROP DEFAULT');
+
+        $this->sql('ALTER TABLE heureka_category ADD heureka_id INT NOT NULL DEFAULT 0');
+        $this->sql('ALTER TABLE heureka_category ALTER heureka_id DROP DEFAULT');
+        $this->sql('UPDATE heureka_category SET heureka_id = id');
```
vitek-rostislav, line 21: this seems bit dangerous, is not it? I would rather keep the default 0 and let the ids be properly set by the download. But if I understand it correctly, it must be unique because of the composite index below, right? But what if our internal id then conflits with the real heureka ID? Or do we delete all and insert again during download? If so, I have no further questions :D

vitek-rostislav: ok, now I see id was actually heureka id, ok :D what about renaming the column and make composite primary key?

packages/product-feed-heureka/src/Form/CategoryCrudExtension.php
```diff
@@ -27,18 +31,24 @@ public function getFormTypeClass(): string
...
     {
-        $heurekaCategory = $this->heurekaCategoryFacade->findByCategoryId($categoryId);
-
         $pluginData = [];
 
-        if ($heurekaCategory !== null) {
-            $pluginData['heureka_category'] = $heurekaCategory;
+        foreach ($this->heurekaCategoryDownloader->getSupportedLocales() as $locale) {
+            if (!$this->domain->hasLocale($locale)) {
+                continue;
```
vitek-rostislav, line 44: hasLocale might be confusing as getter in Domain usually serve as proxies to getCurrentDomainConfig..I would maybe just rename it to something like hasAnyDomainLocale or something like that, what do you think?

packages/product-feed-heureka/src/Form/CategoryFormType.php
```diff
@@ -16,22 +18,36 @@ final class CategoryFormType extends AbstractType
...
+        }
+    }
+
+    private function getLabelForLocale(string $locale): string
+    {
+        return match ($locale) {
+            'cs' => $this->translator->trans('Heureka.cz category'),
+            'sk' => $this->translator->trans('Heureka.sk category'),
+            default => $this->translator->trans('Heureka category'),
+        };
```
vitek-rostislav, line 51: if we hypothetically added new supported locale, this one might be omitted, what about throwing exception here? Or is it too ofensive?

packages/product-feed-heureka/src/Model/HeurekaCategory/HeurekaCategoryRepository.php
```diff
@@ -21,40 +21,46 @@ protected function getHeurekaCategoryRepository(): EntityRepository
...
+            ->andWhere('hc.locale = :locale')
+            ->setParameter('heurekaId', $heurekaId)
+            ->setParameter('locale', $locale);
         $heurekaCategory = $queryBuilder->getQuery()->getOneOrNullResult();
 
         if ($heurekaCategory === null) {
             throw new HeurekaCategoryNotFoundException(
-                'Heureka category with ID ' . $id . ' does not exist.',
+                'Heureka category with ID ' . $heurekaId . ' does not exist.',
             );
```
vitek-rostislav, line 64: "...and locale..."?

vitek-rostislav (changes requested):

packages/framework/src/Migrations/Version20260223133152.php
grossmannmartin: Yes, the id was previously the heureka id in fact. 

About the composite primary key. I don't know... 
I still remember how my great grandma told me: "Honey, don't use composite primary keys in Doctrine, it's a devil's creation."

But maybe that time has passed, and it's ok now 🤔

packages/product-feed-heureka/src/Form/CategoryCrudExtension.php
grossmannmartin: Yes, I had the same twitching. I have renamed it to `anyDomainHasLocale`. It feels a little more natural to me (49da90ef21). We may discuss the naming further.

packages/product-feed-heureka/src/Form/CategoryFormType.php
grossmannmartin: Honestly, this is potato, potahto to me. 
If you feel more like it, I can throw an exception. 

My first thoughts were:
- I don't think this will be updated in the future
- If it will, I get a sane default. If we want a better name, we can find it pretty quickly

packages/product-feed-heureka/src/Model/HeurekaCategory/HeurekaCategoryRepository.php
grossmannmartin: nice 👍 600684ffe24ac6556392008c4846246439a7db12

packages/product-feed-heureka/src/Form/CategoryFormType.php
vitek-rostislav: yeah I can live with that

packages/framework/src/Migrations/Version20260223133152.php
vitek-rostislav: image

ok then, lets keep it as is :)

vitek-rostislav (approved):

PR #4238, opened 2025-10-17
fixed superadmin behavior

grossmannmartin:

- superadmin users are now displayed in the administrator list grid if the currently logged admin is a superadmin
- superadmin can promote regular admin to superadmin
- edit superadmin now show that the currently edited admin is superadmin
- back-fixed RBAC migration - will be backported to 17.0.1
- fixed access to reset password page - it's now public as it's used as unlogged admin

#### Fixes issues
... 

----
  - https://mg-fix-superadmin-role.odin.shopsys.cloud
  - https://cz.mg-fix-superadmin-role.odin.shopsys.cloud
  - https://mg-fix-superadmin-role.odin.shopsys.cloud/sk

### DIFF
--- a/docs/introduction/role-based-access-control.md
+++ b/docs/introduction/role-based-access-control.md
@@ -91,13 +91,13 @@ The system automatically manages role inheritance:
 
 ```
 ROLE_SUPER_ADMIN (highest)
-    └── ROLE_ADMIN
-        └── ROLE_ALL (all FULL permissions)
-            ├── ROLE_PRODUCT_FULL
-            │   ├── ROLE_PRODUCT_EDIT (includes VIEW)
-            │   ├── ROLE_PRODUCT_CREATE (includes VIEW)
-            │   └── ROLE_PRODUCT_DELETE (includes VIEW)
-            └── [other role hierarchies...]
+    └── ROLE_ALL (all FULL permissions)
+        ├── ROLE_PRODUCT_FULL
+        │   ├── ROLE_PRODUCT_EDIT (includes VIEW)
+        │   ├── ROLE_PRODUCT_CREATE (includes VIEW)
+        │   └── ROLE_PRODUCT_DELETE (includes VIEW)
+        └── [other role hierarchies...]
+ROLE_ADMIN (lowest) - allows login and access to basic admin functions
 ```
 
 !!! note
--- a/packages/administration/templates/content/administrator/edit.html.twig
+++ b/packages/administration/templates/content/administrator/edit.html.twig
@@ -3,7 +3,18 @@
 {% block title %}- {{ 'Editing administrator'|trans }} - {{ administrator.realName }}{% endblock %}
 
 {% block pre_title %}{{ 'Edit administrator'|trans }}{% endblock %}
-{% block h1 %}{{ administrator.realName }}{% endblock %}
+{% block h1 %}
+    {{ administrator.realName }}
+    {% if administrator.isSuperadmin() %}
+        <span class="badge fs-5 ms-2 bg-yellow text-yellow-fg"
+              data-bs-toggle="tooltip"
+              title="{{ 'Superadmins have all permissions including changing system configuration and access to hidden administration sections.'|trans }}"
+        >
+            {{ ux_icon('crown') }}
+            {{ 'Super Administrator'|trans }}
+        </span>
+    {% endif %}
+{% endblock %}
 
 {% block admin_log %}
     {% if lastAdminActivities|length > 0 %}
--- a/packages/administration/templates/content/administrator/listGrid.html.twig
+++ b/packages/administration/templates/content/administrator/listGrid.html.twig
@@ -4,6 +4,12 @@
     <a href="{{ url('admin_administrator_edit', { id: row.a.id }) }}">{{ value }}</a>
 {% endblock %}
 
+{% block grid_value_cell_id_superadmin %}
+    {% if value %}
+        <span class="badge bg-red-lt text-red-lt-fg">{{ 'Yes'|trans }}</span>
+    {% endif %}
+{% endblock %}
+
 {% block grid_pager_totalcount %}
     {% set entityName = 'administrators'|trans %}
     {{ parent() }}
--- a/packages/framework/src/Controller/Admin/AdministratorController.php
+++ b/packages/framework/src/Controller/Admin/AdministratorController.php
@@ -79,7 +79,11 @@ class AdministratorController extends AdminBaseController
     #[CanView]
     public function listAction(): Response
     {
-        $queryBuilder = $this->administratorFacade->getAllListableQueryBuilder();
+        if ($this->getCurrentAdministrator()->isSuperadmin()) {
+            $queryBuilder = $this->administratorFacade->getAllQueryBuilder();
+        } else {
+            $queryBuilder = $this->administratorFacade->getAllListableExcludingSuperadminQueryBuilder();
+        }
         $dataSource = $this->queryBuilderDataSourceFactory->create($queryBuilder, 'a.id');
 
         $grid = $this->gridFactory->create('administratorList', $dataSource, AdminRoleConstant::ROLE_ADMINISTRATOR);
@@ -88,6 +92,11 @@ class AdministratorController extends AdminBaseController
         $grid->addColumn('realName', 'a.realName', t('Full name'), true);
         $grid->addColumn('userName', 'a.username', t('Username'), true);
         $grid->addColumn('email', 'a.email', t('Email'));
+
+        if ($this->getCurrentAdministrator()->isSuperadmin()) {
+            $grid->addColumn('superadmin', 'is_superadmin', t('Superadmin'))
+                ->setClassAttribute('text-center w-1');
+        }
 
         $grid->addEditActionColumn('admin_administrator_edit', ['id' => 'a.id']);
         $grid->addDeleteActionColumn('admin_administrator_delete', ['id' => 'a.id'])
--- a/packages/framework/src/Controller/Admin/AdministratorController.php
+++ b/packages/framework/src/Controller/Admin/AdministratorController.php
@@ -11,6 +11,7 @@ use Shopsys\FrameworkBundle\Component\Security\Attribute\CanCreate;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\CanDelete;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\CanView;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\ForRole;
+use Shopsys\FrameworkBundle\Component\Security\Attribute\PublicAccess;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\RequireRole;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\SuperAdminOnly;
 use Shopsys\FrameworkBundle\Component\Security\Role\AdminRoleConstant;
@@ -598,7 +599,7 @@ class AdministratorController extends AdminBaseController
      * @return \Symfony\Component\HttpFoundation\Response
      */
     #[Route(path: '/administrator/set-new-password/', name: 'admin_administrator_set-new-password')]
-    #[CanView]
+    #[PublicAccess]
     public function setNewPasswordAction(Request $request): Response
     {
         $email = $request->query->get('email', '');
--- a/packages/framework/src/Controller/Admin/AdministratorController.php
+++ b/packages/framework/src/Controller/Admin/AdministratorController.php
@@ -12,6 +12,7 @@ use Shopsys\FrameworkBundle\Component\Security\Attribute\CanDelete;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\CanView;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\ForRole;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\RequireRole;
+use Shopsys\FrameworkBundle\Component\Security\Attribute\SuperAdminOnly;
 use Shopsys\FrameworkBundle\Component\Security\Role\AdminRoleConstant;
 use Shopsys\FrameworkBundle\Component\Security\Role\SystemRole;
 use Shopsys\FrameworkBundle\Form\Admin\Administrator\AdministratorFormType;
@@ -560,6 +561,33 @@ class AdministratorController extends AdminBaseController
             [
                 'email' => $administrator->getEmail(),
             ],
+        );
+
+        return $this->redirectToRoute('admin_administrator_edit', ['id' => $id]);
+    }
+
+    /**
+     * @CsrfProtection
+     * @param int $id
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    #[Route(path: '/administrator/promote-to-superadmin/{id}', name: 'admin_administrator_promote-to-superadmin', requirements: ['id' => '\d+'])]
+    #[SuperAdminOnly]
+    public function promoteToSuperadminAction(int $id): Response
+    {
+        $administrator = $this->administratorFacade->getById($id);
+        $administratorData = $this->administratorDataFactory->createFromAdministrator($administrator);
+
+        $administratorData->roleGroup = null;
+        $administratorData->roles = [SystemRole::SUPER_ADMIN];
+
+        $this->administratorFacade->edit($id, $administratorData);
+
+        $this->addSuccessFlash(
+            t(
+                'Administrator "%administrator_name%" now has superadmin permissions.',
+                ['%administrator_name%' => $administrator->getRealName()],
+            ),
         );
 
         return $this->redirectToRoute('admin_administrator_edit', ['id' => $id]);
--- a/packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
+++ b/packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
@@ -160,6 +160,35 @@ final class AdministratorFormType extends AbstractType
                 'back_route' => 'admin_administrator_list',
                 'entity' => $adminToEdit,
             ]);
+
+        if ($options['scenario'] !== self::SCENARIO_EDIT || !$this->canWorkWithRoles($adminToEdit) || !$this->getCurrentAdministrator()->isSuperadmin()) {
+            return;
+        }
+
+        $builderPromoteGroup = $builder->create('promote', GroupType::class, [
+            'label' => 'Super Administrator',
+        ]);
+
+        $builderPromoteGroup->add('promoteToSuperadmin', DisplayOnlyUrlType::class, [
+            'route' => 'admin_administrator_promote-to-superadmin',
+            'route_params' => [
+                'id' => $adminToEdit->getId(),
+                RouteCsrfProtector::CSRF_TOKEN_REQUEST_PARAMETER => $this->routeCsrfProtector->getCsrfTokenByRoute(
+                    'admin_administrator_promote-to-superadmin',
+                ),
+            ],
+            'route_attr' => [
+                'data-confirm-window' => null,
+                'data-confirm-style' => 'warning',
+                'data-confirm-message' => t('Are you sure you want to promote this user to Superadmin?'),
+            ],
+            'label' => 'Superadmin permissions',
+            'help' => 'Superadmins have all permissions including changing system configuration and access to hidden administration sections.',
+            'route_label' => t('Promote user to Superadmin'),
+            'link_target' => '_self',
+        ]);
+
+        $builder->add($builderPromoteGroup);
     }
 
     /**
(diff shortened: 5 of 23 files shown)


vitek-rostislav (approved):

packages/framework/src/Model/Administrator/Administrator.php
```diff
@@ -426,12 +420,12 @@ public function eraseCredentials(): void
     #[Override]
     public function getRoles(): array
     {
+        $roles = [];
+
         if ($this->roleGroup !== null) {
-            return $this->roleGroup->getRoles();
+            $roles = $this->roleGroup->getRoles();
```
henzigo: @grossmannmartin, we should not combine roles from the role group with roles defined directly. This is not safe as it would give you access to parts that you should not have. Also, UI is defined that you cannot use both at the same time so this will create new hidden behavior.

henzigo: Why not create a new action in the admin UI that promotes an administrator to Superadmin (visible only to SUPER_ADMIN) and correctly removes the role group ID and sets the individual SUPER_ADMIN role only? 

The official way to create a superadmin would be much better than editing the database directly :)

grossmannmartin: Maybe it would be better to have a button.
But I don't understand why combining roles is a big no-no. I can easily imagine assigning a role and adding a single permission extra. I know the ui does not allow it now.

Arbitrary precedence of one permission mechanism does not seem intuitive to me. 
Besides assigning superadmin role should overwrite everything else, in my opinion.

henzigo: > Besides assigning superadmin role should overwrite everything else, in my opinion.

I can live with checking if superadmin exists with early return before processing RoleGroup and directly defined roles.

> But I don't understand why combining roles is a big no-no. I can easily imagine assigning a role and adding a single permission extra. I know the ui does not allow it now.

You are creating hidden behaviour, which directly contradicts how admin rights currently work. In the admin, you clearly can select only one (RoleGroup or custom roles). This behaviour should also be reflected in the internal code; otherwise, we’re creating weak code that’s prone to bugs.

If we define combining roles as a feature, we should verify and define this behaviour consistently throughout the codebase, not only in one place. Moreover, this behaviour can only be exploited by directly modifying the database, and it will work only until someone updates that administrator, since the system will automatically remove all direct roles if the RoleGroup is not null.

grossmannmartin: Ok. I have changed the direction of this PR. see updated description

docs/introduction/role-based-access-control.md
```diff
@@ -91,13 +91,13 @@ The system automatically manages role inheritance:
...
-            │   ├── ROLE_PRODUCT_CREATE (includes VIEW)
-            │   └── ROLE_PRODUCT_DELETE (includes VIEW)
-            └── [other role hierarchies...]
+    └── ROLE_ALL (all FULL permissions)
+        ├── ROLE_PRODUCT_FULL
+        │   ├── ROLE_PRODUCT_EDIT (includes VIEW)
+        │   ├── ROLE_PRODUCT_CREATE (includes VIEW)
+        │   └── ROLE_PRODUCT_DELETE (includes VIEW)
+        └── [other role hierarchies...]
+ROLE_ADMIN (lowest - login only)
```
vitek-rostislav: it is also necessary for `efconnect` and `elfinder` routes, and required for some other basic actions, see `SystemRole::ADMIN` usages. Maybe we could mention it here :thinking: Also, we could also mention, that each administrator has at least this role, right?

packages/administration/templates/content/administrator/edit.html.twig
```diff
@@ -3,7 +3,18 @@
 {% block title %}- {{ 'Editing administrator'|trans }} - {{ administrator.realName }}{% endblock %}
 
 {% block pre_title %}{{ 'Edit administrator'|trans }}{% endblock %}
-{% block h1 %}{{ administrator.realName }}{% endblock %}
+{% block h1 %}
+    {{ administrator.realName }}
+    {% if administrator.isSuperadmin() %}
```
vitek-rostislav, line 8: love it!

packages/administration/templates/content/administrator/listGrid.html.twig
```diff
@@ -4,6 +4,12 @@
     <a href="{{ url('admin_administrator_edit', { id: row.a.id }) }}">{{ value }}</a>
 {% endblock %}
 
+{% block grid_value_cell_id_superadmin %}
+    {% if value %}
+        <span class="badge bg-red-lt text-red-lt-fg">{{ 'Yes'|trans }}</span>
```
vitek-rostislav, line 9: I am not sure about the red color, it looks like a warning, or seems like something is wrong with that :smile: On the other hand, superadmin sections has red color in the menu, so at least this is unified, I do not know

packages/framework/src/Controller/Admin/AdministratorController.php
```diff
@@ -565,6 +565,33 @@ public function sendResetPasswordAction(int $id): Response
...
     }
 
+    /**
+     * @CsrfProtection
+     * @param int $id
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    #[Route(path: '/administrator/promote-to-superadmin/{id}', name: 'admin_administrator_promote-to-superadmin', requirements: ['id' => '\d+'])]
+    #[RequireRole(SystemRole::SUPER_ADMIN)]
+    public function promoteToSuperadminAction(int $id): Response
```
vitek-rostislav, line 577: This is a nice feature, however, it would be nice to also have an option to create superadmin directly. :innocent: Would it be too bold of me to want this? :smile:

vitek-rostislav: It would probably lead to a complete reworking of the current implementation, right? :sweat_smile: I mean, I can imagine a simple checkbox "superadmin" in the form. If that was checked, the roles and role groups sections would be hidden then, and vice versa. This could then be used in both create and edit scenarios. Let's talk about it, please :innocent:

project-base/app/tests/App/Smoke/Http/RouteConfigCustomization.php
```diff
@@ -195,6 +195,28 @@ private function configureAdminRoutes(RouteConfigCustomizer $routeConfigCustomiz
                     ->setAuth(new NoAuth())
                     ->setExpectedStatusCode(200);
             })
+            ->customizeByRouteName('admin_administrator_promote-to-superadmin', function (RouteConfig $config) {
+                $config->changeDefaultRequestDataSet('Standard admin is not allowed to promote to superadmin')
+                    ->addCallDuringTestExecution(
```
vitek-rostislav, line 200: you can add the route to the regular expression [here](https://github.com/shopsys/shopsys/blob/18.0/project-base/app/tests/App/Smoke/Http/RouteConfigCustomization.php#L130) instead of duplicating the CSRF setting logic, or?

packages/framework/src/Migrations/Version20250711031748.php
```diff
@@ -51,6 +51,8 @@ private function migrateAdministratorSystemRolesToRoleGroups(): void
             ['name' => SystemRole::ALL_VIEW],
         );
 
+        $this->migrateSuperAdministrators();
```
vitek-rostislav, line 54: cool that you addressed this :ok_hand: :green_heart:

vitek-rostislav (changes requested):

packages/framework/src/Controller/Admin/AdministratorController.php
```diff
@@ -556,6 +565,33 @@ public function sendResetPasswordAction(int $id): Response
         return $this->redirectToRoute('admin_administrator_edit', ['id' => $id]);
     }
 
+    /**
+     * @CsrfProtection
+     * @param int $id
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    #[Route(path: '/administrator/promote-to-superadmin/{id}', name: 'admin_administrator_promote-to-superadmin', requirements: ['id' => '\d+'])]
+    #[RequireRole(SystemRole::SUPER_ADMIN)]
```
henzigo: You can use #[SuperAdminOnly] instead of this

packages/administration/templates/content/administrator/edit.html.twig
henzigo: Love it more :D

packages/framework/src/Controller/Admin/AdministratorController.php
```diff
@@ -79,12 +79,21 @@ public function __construct(
...
+            $queryBuilder = $this->administratorFacade->getAllQueryBuilder();
+        } else {
+            $queryBuilder = $this->administratorFacade->getAllListableExcludingSuperadminQueryBuilder();
+        }
         $dataSource = $this->queryBuilderDataSourceFactory->create($queryBuilder, 'a.id');
 
         $grid = $this->gridFactory->create('administratorList', $dataSource, AdminRoleConstant::ROLE_ADMINISTRATOR);
         $grid->setDefaultOrder('realName');
 
+        if ($this->getCurrentAdministrator()->isSuperadmin()) {
```
henzigo, line 98: I would move this column to the end after the email column. It looks weird to have this information as the first column. The first should be a row identifier, and after that, any other information associated with it.

packages/framework/src/Controller/Admin/AdministratorController.php
henzigo: If you are going to discuss this, I would like to give you my opinion. 

Currently, "Superadmin" role is something in the middle of custom roles and "role group". In my opinion, we should create a RoleGroup "Super admin" that will act the same. But from the user (administrator) perspective, it will be unified with other behaviours.

packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
```diff
@@ -160,6 +160,35 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                'data-confirm-style' => 'warning',
+                'data-confirm-message' => t('Are you sure you want to promote this user to Superadmin?'),
+            ],
+            'label' => 'Superadmin permissions',
+            'help' => 'Superadmins have all permissions including changing system configuration and access to hidden administration sections.',
+            'route_label' => t('Promote user to Superadmin'),
+            'link_target' => '_self',
+        ]);
+
+        $builder->add($builderPromoteGroup);
```
henzigo, line 191: This looks cool, but why don't you just create that action as a grid action? It can show all that information in that confirmation window.

Also, this is just another "Role Group", but in UI it's separated, and it looks like this is something different. 

I understand why you implemented it this way, but isn't that overkill?

packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
```diff
@@ -160,6 +160,35 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+
+        if ($options['scenario'] !== self::SCENARIO_EDIT || !$this->canWorkWithRoles($adminToEdit) || !$this->getCurrentAdministrator()->isSuperadmin()) {
+            return;
+        }
+
+        $builderPromoteGroup = $builder->create('promote', GroupType::class, [
+            'label' => 'Super Administrator',
+        ]);
+
+        $builderPromoteGroup->add('promoteToSuperadmin', DisplayOnlyUrlType::class, [
```
henzigo, line 172: This is weird to use `DisplayOnlyUrlType` and show something totally different in that type :D

packages/framework/src/Migrations/Version20250711031748.php
```diff
@@ -84,4 +86,27 @@ private function migrateAdministratorSystemRolesToRoleGroups(): void
...
+    public function migrateSuperAdministrators(): void
+    {
+        $superAdministratorIds = $this->connection->fetchAllAssociative(
+            'SELECT administrator_id FROM administrator_roles WHERE role = :superadmin_role',
+            ['superadmin_role' => SystemRole::SUPER_ADMIN],
+        );
+
+        foreach ($superAdministratorIds as $superAdministratorId) {
+            $this->sql(
+                'UPDATE administrators SET role_group_id = NULL WHERE id = :administrator_id',
```
henzigo: Nitpick

`UPDATE administrators SET role_group_id = NULL WHERE id = :administrator_id AND role_group_id IS NOT NULL`

We shouldn't update rows when nothing has changed. Like in this case, it's irrelevant, but we all should remember that.

packages/framework/src/Model/Administrator/Administrator.php
```diff
@@ -430,13 +430,17 @@ public function getRoles(): array
...
 
-        $roles = [SystemRole::ADMIN];
+        $roles = [];
 
         /** @var \Shopsys\FrameworkBundle\Model\Administrator\Role\AdministratorRole $role */
         foreach ($this->roles->getValues() as $role) {
             $roles[] = $role->getRole();
         }
 
+        if (!in_array(SystemRole::ADMIN, $roles, true) && !in_array(SystemRole::SUPER_ADMIN, $roles, true)) {
```
henzigo: `!in_array(SystemRole::ADMIN, $roles, true)` 

This part of the condition is irrelevant, as 4 lines below, unique roles are returned.

But I don't understand what this change fixed. Could you describe it to me please?

packages/framework/src/Model/Administrator/AdministratorRepository.php
```diff
@@ -111,21 +112,33 @@ public function getByEmail(string $administratorEmail): Administrator
...
+
+    /**
+     * @return \Doctrine\ORM\QueryBuilder
+     */
+    public function getAllQueryBuilder(): QueryBuilder
+    {
+        return $this->getAdministratorRepository()->createQueryBuilder('a')
+            ->addSelect('CASE WHEN ar.role = :superRole THEN true ELSE false END as is_superadmin')
+            ->leftJoin('a.roles', 'ar')
+            ->setParameter('superRole', SystemRole::SUPER_ADMIN);
```
henzigo: Nitpick (Irrelevant for this scenario)

Left join 1 to many will multiply the original data (administrator), and it will slow down Doctrine hydration (Also query response is bigger).

henzigo: Also, if you will use aggregation like in the next method:

```
return (int)($this->getAllListableExcludingSuperadminQueryBuilder()
            ->select('COUNT(a)')
            ->getQuery()->getSingleScalarResult());
```

It can return an incorrect number due to that left join

packages/administration/templates/content/administrator/listGrid.html.twig
grossmannmartin: Well, it's quite dangerous to have a superadmin role, so... 😆

packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
grossmannmartin: I didn't want to put it in the grid to avoid having such a niche scenario available from the list. 
I see "promote to superadmin" as something that should be carefully thought through, and the separation helps understand that it's not "just another role", but something really dangerous.

packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
grossmannmartin: It only displays the url that promotes user to superadmin 😆

packages/framework/src/Migrations/Version20250711031748.php
grossmannmartin: Good to know, thanks...

packages/framework/src/Model/Administrator/AdministratorRepository.php
grossmannmartin: Yes, I'm aware of that. But it's better than sequentially loading roles for each row (due to the isSuperamin implementation), or?

packages/framework/src/Model/Administrator/Administrator.php
grossmannmartin: Otherwise, ROLE_ADMIN is added to superadmin when updated.

packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
henzigo: Yes, I understand that, but it is still just another role :D

packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
henzigo: In this way yes :D

packages/framework/src/Model/Administrator/Administrator.php
henzigo: Is it a problem? Or if you are doing those changes, can we avoid conditions like this and instead just accept that SUPERADMIN can also have ADMIN role ?

packages/framework/src/Model/Administrator/AdministratorRepository.php
henzigo: If you need to know if the admin has the superadmin role, you can join just this role based on condition and then only check if that join is not null or create subquery like:

```
SELECT a, (EXISTS (
            	SELECT 1 FROM administrator_roles AS ar 
            	WHERE ar.administrator_id = a.id AND ar.role = 'ROLE_SUPERADMIN'
            )) as is_superadmin
...
```

henzigo: Or you can use `QueryBuilderWithRowManipulatorDataSourceFactory` to load this for all administrators as 1 query

docs/introduction/role-based-access-control.md
grossmannmartin: I didn't want to bulk up the documentation... how about this? 2dd90ccd25

packages/framework/src/Controller/Admin/AdministratorController.php
grossmannmartin: I see creating a superadmin directly as quite a niche use case — I don’t really mind creating the administrator first and then assigning the superadmin permission afterwards.

Integrating the superadmin permission into the regular role system would definitely be nice, but it would require a larger refactor to make sure the superadmin role isn’t visible or assignable by regular administrators. That doesn’t feel like an efficient change to make at the moment.

I also avoided adding a simple “Superadmin” checkbox to the form on purpose. Granting superadmin access is a potentially risky action, and I wanted to keep a small mental barrier there — a moment to stop and think, “Do I really want to do this?” 😄

project-base/app/tests/App/Smoke/Http/RouteConfigCustomization.php
grossmannmartin: I don't think so... I also need to add an extra request dataset (for superadmin) and check for a different status code for regular admin. 
But maybe I'm overlooking some possibilities 🤔

packages/framework/src/Controller/Admin/AdministratorController.php
grossmannmartin: I don't like two ways of doing things. 
This raises questions like "Does it behave differently?" "Should I use `#[AdminIsEnough]` instead of `#[RequireRole(SystemRole::ADMIN)]`?"

But it passed through review, so changed - 027b710ae9

henzigo: No, it's just a different way to do the same thing. But from the perspective of how Superadmin role can be dangerous, it's just better readable for programmers

packages/framework/src/Controller/Admin/AdministratorController.php
grossmannmartin: 🤷‍♂️  changed ea92bba420

packages/framework/src/Controller/Admin/AdministratorController.php
henzigo: So that's why `#[SuperAdminOnly]` attribute exists. We could add more specific attributes like `#[AdminIsEnough]`, but we still need to have an option for other possible roles configured on project.

packages/framework/src/Form/Admin/Administrator/AdministratorFormType.php
grossmannmartin: We don't treat it as any other role. Not in the matter of code, nor in the matter of thinking

packages/framework/src/Migrations/Version20250711031748.php
grossmannmartin: fixed a1dff53167

packages/framework/src/Model/Administrator/Administrator.php
grossmannmartin: Well, it behaved just weird. It's also a default state when you load demo data. 
Besides, it makes sense to me — if I’m a superadmin, the default role is Superadmin, and if not, it defaults to Admin.

henzigo: So what about this?

```diff
    public function getRoles(): array
    {
+        if ($this->roles->exists(fn ($key, AdministratorRole $role) => $role->getRole() === SystemRole::SUPER_ADMIN)) {
+            return [SystemRole::SUPER_ADMIN];
+        }
        
        if ($this->roleGroup !== null) {
            return $this->roleGroup->getRoles();
        }

        $roles = [SystemRole::ADMIN];

        /** @var \Shopsys\FrameworkBundle\Model\Administrator\Role\AdministratorRole $role */
        foreach ($this->roles->getValues() as $role) {
            $roles[] = $role->getRole();
        }

        return array_unique($roles);
    }
```

It's more readable, you will always get the correct role, and it's also more readable.

packages/framework/src/Model/Administrator/AdministratorRepository.php
grossmannmartin: rewritten without left join in e22ef94281

packages/framework/src/Model/Administrator/Administrator.php
grossmannmartin: 🤷  -> eb94ac9ab9

project-base/app/tests/App/Smoke/Http/RouteConfigCustomization.php
vitek-rostislav: ok, no biggie, I believe you :smile:

vitek-rostislav (approved):

pk16011990 (approved):

PR #4246, opened 2025-10-23
order withdrawal from contract 

vitek-rostislav:

Based on the legislative conditions, a customer should be able to request a withdrawal from a contract when certain criteria are met (the order is not cancelled, the order was not delivered yet, or the deadline for the withdrawal request after the order delivery has not exceeded). 

#### Fixes issues
... 

## Summary by CodeRabbit

* **New Features**
  * Added order withdrawal functionality allowing customers to request contract withdrawal with deadline tracking
  * Added withdrawal request form capturing customer contact information and notes
  * Added withdrawal instructions and deadline information to order details
  * Added delivery date tracking for orders

----
  - https://rv-withdrawal.odin.shopsys.cloud
  - https://cz.rv-withdrawal.odin.shopsys.cloud
  - https://rv-withdrawal.odin.shopsys.cloud/sk


vitek-rostislav:
@coderabbitai review

project-base/storefront/components/Pages/Order/ContactInformation/FormBlocks/ContactInformationDeliveryAddress.tsx
```diff
@@ -185,7 +185,7 @@ export const ContactInformationDeliveryAddress: FC = () => {
                                         render={(textInput) => <FormLine bottomGap>{textInput}</FormLine>}
                                         textInputProps={{
                                             label: formMeta.fields.deliveryTelephone.label,
-                                            required: true,
+                                            required: false,
```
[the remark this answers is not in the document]
chlebektomas: agree

project-base/storefront/components/Forms/validationRules.ts
```diff
@@ -24,18 +24,19 @@ export const validateCustomer = (): Schema => {
 
 export const validateTelephone = (t: Translate): Schema => {
     return Yup.string()
-        .matches(
-            /^(\+[1-9]\d{0,3})?[0-9]+$/,
-            t('Please enter a valid phone number. Examples: 123456789, +420123456789'),
-        )
+        .optional()
```
chlebektomas: is it necessary to have it globally?

chlebektomas: I am a little bit confused about all these phone validation updates.

chlebektomas (changes requested):

packages/framework/src/Component/Setting/Setting.php
```diff
@@ -28,6 +28,8 @@ class Setting
     public const string CUSTOMER_USER_DEFAULT_GROUP_ROLE_ID = 'customerUserDefaultGroupRoleId';
     public const string FILE_STRUCTURE_MIGRATED_FOR_RELATIONS = 'fileStructureMigratedForRelations';
     public const string CSP_HEADER = 'cspHeader';
+    public const string WITHDRAWAL_DEADLINE_DAYS = 'withdrawalDeadlineDays';
+    public const string WITHDRAWAL_INSTRUCTIONS = 'withdrawalInstructions';
```
grossmannmartin: :bulb: What about create a separate setting facade for withdrawal? Similar to the `PricingSetting`

packages/framework/src/Form/WysiwygTypeExtension.php
```diff
@@ -54,6 +59,44 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+
+            if (array_key_exists('help', $view->vars) && $view->vars['help'] !== null) {
+                $view->vars['help'] .= $variablesHtml;
+            } else {
+                $view->vars['help'] = $variablesHtml;
+            }
+
+            $view->vars['help_html'] = true;
+        }
+    }
```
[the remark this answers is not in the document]
grossmannmartin: this seems reasonable

packages/framework/src/Model/Order/Mail/WithdrawalMail.php
```diff
@@ -0,0 +1,107 @@
...
+     * @param \Shopsys\FrameworkBundle\Model\Mail\MailTemplate $mailTemplate
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @return \Shopsys\FrameworkBundle\Model\Mail\MessageData
+     */
+    #[Override]
+    public function createMessage(MailTemplate $mailTemplate, $order): MessageData
+    {
+        return new MessageData(
+            $order->getWithdrawalEmail(),
+            array_filter([$mailTemplate->getBccEmail(), $order->getEmail()]),
```
grossmannmartin: Is it a problem when the same e-mail address is in `to` and `bcc` fields?
I mean when I withdraw with the same e-mail as I made an order (which may be quite common scenario)

packages/framework/src/Model/Order/Withdrawal/WithdrawalDeadlineCalculation.php
```diff
@@ -0,0 +1,40 @@
...
+        if ($deliveredAt === null) {
+            return null;
+        }
+
+        $withdrawalDeadlineDays = $this->setting->getForDomain(
+            Setting::WITHDRAWAL_DEADLINE_DAYS,
+            $order->getDomainId(),
+        );
+
+        return (clone $deliveredAt)->modify(sprintf('+%d days', $withdrawalDeadlineDays));
```
grossmannmartin, line 34: I think DateTimeImmutable should always be the first choice to avoid the necessity of manual cloning. But keep it as it is

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestData.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    public string $firstName;
+
+    public string $lastName;
+
+    public string $email;
+
+    public ?string $telephone = null;
+
+    public ?string $note = null;
```
grossmannmartin: Should be these properties in the data object typed?

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestDataFactory.php
```diff
@@ -0,0 +1,25 @@
...
+
+class WithdrawalRequestDataFactory
+{
+    /**
+     * @param array<string, mixed> $data
+     * @return \Shopsys\FrameworkBundle\Model\Order\Withdrawal\WithdrawalRequestData
+     */
+    public function createFromArray(array $data): WithdrawalRequestData
+    {
+        $withdrawalRequestData = new WithdrawalRequestData();
```
grossmannmartin: maybe createInstance method?

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestFacade.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @param array<string, mixed> $withdrawalData
+     */
+    public function createWithdrawalRequest(Order $order, array $withdrawalData): void
+    {
+        $withdrawalRequestData = $this->withdrawalRequestDataFactory->createFromArray($withdrawalData);
+
+        $order->setWithdrawalData($withdrawalRequestData);
```
grossmannmartin: coderabbit has a point in one thing - we should not allow creating withdrawal for orders that won't pass the check

edit: oh, it's guarded by the validation in yaml file, isn't it? If so, I think it's ok then

edit2: I think it's counterintuitive... I can call facade method that completely bypasses any validation process...

packages/framework/src/Model/Order/OrderDeliveryDateFacade.php
```diff
@@ -0,0 +1,55 @@
...
+        return $order->getDeliveredAt() === null && ($order->getStatus()->getType() === OrderStatusTypeEnum::TYPE_DONE);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     */
+    protected function setDeliveredAt(Order $order): void
+    {
+        $domainTimeZone = $this->displayTimeZoneProvider->getDisplayTimeZoneByDomainId($order->getDomainId());
+        $deliveredAt = $this->dateTimeHelper->createUtcDateTimeByTimeZoneAndString('now', $domainTimeZone);
```
grossmannmartin: What's the point of the conversion? isn't `new DateTime('now')` enough?

packages/framework/src/Model/Order/OrderDeliveryDateFacade.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    protected function shouldSetDeliveredAt(Order $order): bool
+    {
+        return $order->getDeliveredAt() === null && ($order->getStatus()->getType() === OrderStatusTypeEnum::TYPE_DONE);
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     */
+    protected function setDeliveredAt(Order $order): void
```
grossmannmartin: `setDeliveredAtToNow`?

packages/framework/src/Resources/translations/messages.cs.po
```diff
@@ -3448,6 +3472,24 @@ msgstr "Hlídací psi"
...
+msgid "Withdrawal Request"
+msgstr "Žádost o odstoupení od smlouvy"
+
+msgid "Withdrawal deadline (days)"
+msgstr "Lhůta pro odstoupení (dny)"
+
+msgid "Withdrawal from a contract"
+msgstr "Odstoupení od smlouvy"
+
+msgid "Withdrawal from contract"
```
grossmannmartin, line 3529: this and the one above should be unified, or?

packages/framework/src/Resources/translations/validators.cs.po
```diff
@@ -688,6 +694,9 @@ msgstr "Název skladu nesmí být delší než {{ limit }} znaků"
 msgid "Warehouse with this external code already exists"
 msgstr "Sklad s tímto externím kódem již existuje"
 
+msgid "Withdrawal deadline must be a positive number."
+msgstr "Lhůta pro odstoupení musí být kladné číslo"
```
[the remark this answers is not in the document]
grossmannmartin: true

packages/framework/src/Form/Admin/LegalConditions/TermsAndConditionsSettingFormType.php
```diff
@@ -46,8 +50,40 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                    new Constraints\NotBlank([
+                        'message' => 'Please enter withdrawal deadline.',
+                    ]),
+                    new Constraints\Positive([
+                        'message' => 'Withdrawal deadline must be a positive number.',
+                    ]),
+                ],
+            ])
+            ->add('withdrawalInstructions', CKEditorType::class, [
+                'required' => true,
```
grossmannmartin, line 71: can we please add help text? I have absolutely no idea why I should set it and where is it then displayed.

packages/framework/src/Resources/views/Mail/Order/withdrawalAdminMail.html.twig
```diff
@@ -0,0 +1,27 @@
+<h1>{{ 'New Withdrawal Request'|trans }}</h1>
+
+<p>{{ 'A customer has requested withdrawal from contract for the following order'|trans }}</p>
+
+<h2>{{ 'Order information'|trans }}</h2>
+<ul>
+    <li><strong>{{ 'Order number'|trans }}:</strong> {{ order.number }}</li>
+    <li><strong>{{ 'Customer email'|trans }}:</strong> {{ order.email }}</li>
+    <li><strong>{{ 'Order date'|trans }}:</strong> {{ order.createdAt|formatDateTime }}</li>
```
grossmannmartin: should it be formatted for the timezone of the order domain?

packages/framework/src/Resources/views/Mail/Order/withdrawalAdminMail.html.twig
```diff
@@ -0,0 +1,27 @@
...
+</ul>
+
+<h2>{{ 'Withdrawal request details'|trans }}</h2>
+<ul>
+    <li><strong>{{ 'Requested at'|trans }}:</strong> {{ order.withdrawalRequestedAt|formatDateTime }}</li>
+    <li><strong>{{ 'First name'|trans }}:</strong> {{ order.withdrawalFirstName }}</li>
+    <li><strong>{{ 'Last name'|trans }}:</strong> {{ order.withdrawalLastName }}</li>
+    <li><strong>{{ 'Email'|trans }}:</strong> {{ order.withdrawalEmail }}</li>
+    {% if order.withdrawalTelephone %}
+        <li><strong>{{ 'Telephone'|trans }}:</strong> {{ order.withdrawalTelephone }}</li>
```
grossmannmartin: I would really like to start using more modern alternative like 'Phone' or 'Phone number' 😁

packages/frontend-api/src/Component/Constraints/OrderWithdrawalRequest.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @var array<string, string>
+     */
+    protected const array ERROR_NAMES = [
+        self::ORDER_NOT_FOUND_ERROR => 'ORDER_NOT_FOUND_ERROR',
+        self::ORDER_CANCELLED_ERROR => 'ORDER_CANCELLED_ERROR',
+        self::WITHDRAWAL_DEADLINE_PASSED_ERROR => 'WITHDRAWAL_DEADLINE_PASSED_ERROR',
+        self::ALREADY_REQUESTED_ERROR => 'ALREADY_REQUESTED_ERROR',
+    ];
```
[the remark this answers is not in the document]
grossmannmartin: 👎🏿

packages/frontend-api/src/Model/Mutation/Order/OrderWithdrawalRequestMutation.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $input = $argument['input'];
+
+        try {
+            $order = $this->orderFacade->getByUrlHashAndDomain(
+                $input['orderUrlHash'],
+                $this->domain->getId(),
+            );
+        } catch (OrderNotFoundException) {
+            throw new OrderNotFoundUserError('Order not found.');
```
grossmannmartin: This seems a little inconsistent. We specifically check for the existence of the order (even though it's checked in the validator), but we don't check the prerequisities for creating the withdrawal

packages/framework/src/Model/Order/Withdrawal/WithdrawalFacade.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     * @return string
+     */
+    public function getWithdrawalInstructions(Order $order): string
+    {
+        $withdrawalInstructions = $this->setting->getForDomain(
+            Setting::WITHDRAWAL_INSTRUCTIONS,
+            $this->domain->getId(),
```
grossmannmartin: 🤔 do we want the instructions for the current domain, or the domain of the order?

grossmannmartin (changes requested):

project-base/storefront/components/Forms/validationRules.ts
vitek-rostislav: well the thing is, despite having two methods for validation (`validateTelephone` and `validateTelephoneRequired`, they both require to input a telephone because of the regex validation that does not allow empty field.
image

So I updated the `validateTelephone` so that it can be used in forms to validate telephone but not require it (this is what I needed in the withdrawal form, and what should be used in the delivery address as well because the telephone should actually not be required in the delivery address (it is not required in API nor database)

chlebektomas: I see now, thanks. All good then.

project-base/storefront/components/Pages/Order/ContactInformation/FormBlocks/ContactInformationDeliveryAddress.tsx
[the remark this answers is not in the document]
vitek-rostislav: yeah, as explained in the comment below, the telephone in the delivery address should actually be optional indeed. I consider this a bug that I originally wanted to solve in a separate issue, however, as the commit was quite simple, and I needed that behavior in the withdrawal form, I dared to fix it directly in this pull request :innocent:

project-base/storefront/components/Forms/validationRules.ts
```diff
@@ -24,18 +24,19 @@ export const validateCustomer = (): Schema => {
...
+        .matches(/^(\+[1-9]\d{0,3})?[0-9]+$/, {
+            excludeEmptyString: true,
+            message: t('Please enter a valid phone number. Examples: 123456789, +420123456789'),
+        })
```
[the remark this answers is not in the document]
vitek-rostislav: fixed in 07c435ee6e

packages/framework/src/Form/WysiwygTypeExtension.php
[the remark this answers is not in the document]
vitek-rostislav: fixed in 2a4d60cd6e

packages/framework/src/Migrations/Version20251021122100.php
```diff
@@ -0,0 +1,109 @@
...
+
+            $this->sql(
+                'INSERT INTO mail_templates (name, domain_id, send_mail) VALUES (:mailTemplateName, :domainId, :sendMail)',
+                [
+                    'mailTemplateName' => $mailTemplateName,
+                    'domainId' => $domainId,
+                    'sendMail' => true,
+                ],
+            );
+        }
```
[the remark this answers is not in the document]
vitek-rostislav: fixed in 7d4b2dedf8

packages/framework/src/Model/Mail/MailDisplayPriceResolver.php
```diff
@@ -0,0 +1,39 @@
...
+    public function getDisplayPrice(): string
+    {
+        if ($this->mailTemplateDisplayPrice === static::DISPLAY_PRICE_SELLING) {
+            return $this->pricingSetting->getSellingPriceType() === PricingSetting::PRICE_TYPE_WITH_VAT
+                ? static::DISPLAY_PRICE_WITH_VAT
+                : static::DISPLAY_PRICE_WITHOUT_VAT;
+        }
+
+        return static::DISPLAY_PRICE_BOTH;
+    }
```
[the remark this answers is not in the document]
vitek-rostislav: No, see https://docs.shopsys.com/en/17.0/introduction/faq-and-common-issues/#how-do-i-configure-prices-in-e-mail-templates - the only supported values are `selling_price` and `both`

packages/framework/src/Model/Mail/MailTemplateSender/WithdrawalMailTemplateSender.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param \Shopsys\FrameworkBundle\Model\Mail\MailTemplate $mailTemplate
+     * @param string $mailTo
+     * @param int|null $entityId
+     */
+    #[Override]
+    public function sendTemplate(MailTemplate $mailTemplate, string $mailTo, ?int $entityId): void
+    {
+        $order = $this->orderFacade->getById($entityId);
+        $this->withdrawalCustomerMailFacade->sendMailTemplate($mailTemplate, $order, $mailTo);
+    }
```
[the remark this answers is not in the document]
vitek-rostislav: fixed in ac8781162b

packages/framework/src/Model/Order/Mail/WithdrawalMail.php
vitek-rostislav: yeah, good catch, fixed in 2e754e24ef

packages/framework/src/Model/Mail/MessageData.php
```diff
@@ -24,8 +24,8 @@ class MessageData
      * @param string|null $replyTo
      */
     public function __construct(
-        public string $toEmail,
-        public ?string $bccEmail,
+        public string|array $toEmail,
+        public string|array|null $bccEmail,
         public string $body,
```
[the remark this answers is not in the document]
vitek-rostislav: no

packages/framework/src/Model/Order/Order.php
```diff
@@ -590,6 +633,13 @@ protected function fillCommonFields(OrderData $orderData): void
 
         $this->status = $orderData->status;
         $this->heurekaAgreement = $orderData->heurekaAgreement;
+        $this->deliveredAt = $orderData->deliveredAt;
+        $this->withdrawalFirstName = $orderData->withdrawalFirstName;
+        $this->withdrawalLastName = $orderData->withdrawalLastName;
+        $this->withdrawalTelephone = $orderData->withdrawalTelephone;
+        $this->withdrawalEmail = $orderData->withdrawalEmail;
+        $this->withdrawalNote = $orderData->withdrawalNote;
+        $this->withdrawalRequestedAt = $orderData->withdrawalRequestedAt;
```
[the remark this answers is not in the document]
vitek-rostislav: no, it can be modified in the administration anytime

packages/framework/src/Model/Order/Order.php
```diff
@@ -853,6 +903,84 @@ public function setOrderPaymentStatusPageValidFromNow(): void
...
+    public function setWithdrawalData(
+        WithdrawalRequestData $withdrawalRequestData,
+    ): void {
+        $this->withdrawalFirstName = $withdrawalRequestData->firstName;
+        $this->withdrawalLastName = $withdrawalRequestData->lastName;
+        $this->withdrawalEmail = $withdrawalRequestData->email;
+        $this->withdrawalTelephone = $withdrawalRequestData->telephone;
+        $this->withdrawalNote = $withdrawalRequestData->note;
+        $this->withdrawalRequestedAt = new DateTime();
+    }
```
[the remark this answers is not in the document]
vitek-rostislav: no, it is covered by validation in a higher layer

packages/framework/src/Model/Order/Order.php
```diff
@@ -853,6 +903,84 @@ public function setOrderPaymentStatusPageValidFromNow(): void
...
+        return $this->deliveredAt;
+    }
+
+    /**
+     * @param \DateTime $deliveredAt
+     */
+    public function setDeliveredAt($deliveredAt)
+    {
+        $this->deliveredAt = $deliveredAt;
+    }
```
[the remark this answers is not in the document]
vitek-rostislav: no, we do not do that by purpose in the packages

packages/framework/src/Model/Order/Withdrawal/WithdrawalChecker.php
```diff
@@ -0,0 +1,54 @@
...
+
+        if ($withdrawalDeadline === null) {
+            return;
+        }
+
+        $now = new DateTime();
+
+        if ($now > $withdrawalDeadline) {
+            throw new WithdrawalDeadlinePassedException('Withdrawal deadline has passed for this order');
+        }
```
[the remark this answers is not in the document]
vitek-rostislav: we have `CheckTimezonesCommand` that ensures timezones in PHP and database configuration are identical

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestFacade.php
```diff
@@ -0,0 +1,43 @@
...
+
+        $this->em->flush();
+
+        $this->withdrawalCustomerMailFacade->sendEmail($order);
+        $this->withdrawalAdminMailFacade->sendEmail($order);
+    }
```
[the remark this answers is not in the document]
vitek-rostislav: I will rewrite it a little bit, the facade will use the checker

packages/framework/src/Component/Setting/Setting.php
vitek-rostislav: great idea, 15175cd465

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestData.php
vitek-rostislav: hm, probably should not :smile: bc851ef0ec

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestDataFactory.php
vitek-rostislav: ok, bed2b724a6

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestFacade.php
vitek-rostislav: refactored in b3386c33d5

packages/frontend-api/src/Model/Mutation/Order/OrderWithdrawalRequestMutation.php
vitek-rostislav: refactored in b3386c33d5

packages/framework/src/Model/Order/Withdrawal/WithdrawalFacade.php
vitek-rostislav: rewritten in 15175cd4659cfe2b9038cbef5357941261a3101d to use domain of the order

packages/framework/src/Model/Order/OrderDeliveryDateFacade.php
vitek-rostislav: yes, simplified in 31f2729f33

packages/framework/src/Model/Order/OrderDeliveryDateFacade.php
vitek-rostislav: I have removed the method completely in 31f2729f33

packages/framework/src/Form/Admin/LegalConditions/TermsAndConditionsSettingFormType.php
vitek-rostislav: :+1: a4d494e098

packages/framework/src/Resources/views/Mail/Order/withdrawalAdminMail.html.twig
vitek-rostislav: I am not sure, I do not think so. It is formatted for the admin timezone, which seems to be the best option to me, or? We can talk about it :slightly_smiling_face:

vitek-rostislav: Anyway, I have added some improvements to the template in 3897261ea4 and 7d82e5a31f

packages/framework/src/Resources/views/Mail/Order/withdrawalAdminMail.html.twig
vitek-rostislav: :shrug: c453c7bd92

packages/framework/src/Resources/translations/messages.cs.po
vitek-rostislav: Yeah, it really feels like translation messages are my new archenemy :melting_face:

vitek-rostislav: 7b565a8db9 and e025db7876

packages/framework/src/Model/Order/OrderDeliveryDateFacade.php
vitek-rostislav: eh, fixed tests in 304ee3eb6e

grossmannmartin (approved):

chlebektomas (approved):

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -77,8 +77,13 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->add('orderItems', OrderItemsType::class, [
                 'order' => $order,
             ])
-            ->add($this->createPaymentTransactionsGroup($builder, $order))
-            ->add($this->createWithdrawalRequestGroup($builder))
+            ->add($this->createPaymentTransactionsGroup($builder, $order));
+
+        if ($order->getWithdrawalRequest() !== null) {
```
grossmannmartin: Withdrawal cannot be created from the administration anymore, but can be edited, when present.
Is it ok from the business perspective?

project-base/app/src/DataFixtures/Demo/OrderWithdrawalRequestDataFixture.php
```diff
@@ -0,0 +1,67 @@
...
+                    'lastName' => 'Bořič',
+                    'telephone' => '+420755496328',
+                    'email' => 'no-reply@shopsys.com',
+                    'note' => t('Product does not match description, I want to return the entire order.', domain: Translator::DATA_FIXTURES_TRANSLATION_DOMAIN, locale: $domainConfig->getLocale()),
+                ]);
+                $withdrawalRequestData->requestedAt = (new DateTime('now -1 day'))->setTime(14, 30, 15);
+
+                $this->withdrawalRequestFacade->createOnly($orderReference, $withdrawalRequestData);
+            } catch (PersistentReferenceNotFoundException) {
+                continue;
```
grossmannmartin: Is this safe? Couldn't it just hide an errors and making debugging harder?

packages/framework/src/Resources/translations/validators.cs.po
```diff
@@ -472,6 +475,9 @@ msgstr "Vyplňte prosím předmět"
 msgid "Please enter telephone number"
 msgstr "Vyplňte prosím telefon"
 
+msgid "Please enter the date and time of the withdrawal request"
+msgstr "Zadejte prosím datum a čas požadavku na odstoupení"
```
grossmannmartin: `... od smlouvy`?

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestFacade.php
```diff
@@ -37,11 +39,21 @@ public function createWithdrawalRequest(Order $order, array $withdrawalData): vo
 
         $withdrawalRequestData = $this->withdrawalRequestDataFactory->createFromArray($withdrawalData);
```
grossmannmartin: I'm thinking whether the `createWithdrawalRequest` shouldn't accept the WithdrawalRequestData instead of an array. 

That would require data preparation to be moved to the mutation, but it would "standardize" the facade behavior for creating the withdrawal from the different sources (erp/crm/...)

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequest.php
```diff
@@ -0,0 +1,147 @@
...
+    protected $email;
+
+    /**
+     * @var string|null
+     * @ORM\Column(type="text", nullable=true)
+     */
+    protected $note;
+
+    /**
+     * @var \DateTime
```
grossmannmartin: I would prefer DateTimeImmutable, but i can live with it

packages/framework/src/Model/Order/OrderFacade.php
```diff
@@ -158,6 +158,10 @@ public function edit(int $orderId, OrderData $orderData): Order
         $orderTotalPrice = $this->orderPriceCalculation->getOrderTotalPrice($order);
         $order->setTotalPrices($orderTotalPrice->getPrice(), $orderTotalPrice->getProductPrice());
 
+        if ($orderData->withdrawalRequestData !== null && $order->getWithdrawalRequest() !== null) {
+            $order->getWithdrawalRequest()->edit($orderData->withdrawalRequestData);
```
grossmannmartin: When I call 

```
$orderData = $this->orderDataFactory->createFromOrder($order);

$orderData->withdrawalRequestData = null;

// edit/flush
```

The result is unexpected - it's ignored instead of set to null

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: After a discussion with @pk16011990, it is now even simpler - the withdrawal is displayed only, no need for creating, editing, or whatever in the administration, just display :slightly_smiling_face: 

reworked within eca3bfe98c

project-base/app/src/DataFixtures/Demo/OrderWithdrawalRequestDataFixture.php
vitek-rostislav: yeah, the thing is (was) that we are creating distinct data for the 2nd domain in the `OrderDataFixture` and that was missing the reference. I fixed that by adding the reference even for the 2nd domain. IMHO it is better now, thanks for pointing this out.

eca3bfe98c

packages/framework/src/Resources/translations/validators.cs.po
vitek-rostislav: completely removed in eca3bfe98c

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequestFacade.php
vitek-rostislav: agree, reworked in eca3bfe98c

packages/framework/src/Model/Order/Withdrawal/WithdrawalRequest.php
vitek-rostislav: changed in eca3bfe98c

packages/framework/src/Model/Order/OrderFacade.php
vitek-rostislav: after eca3bfe98c, not anymore :smile:

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -605,17 +608,22 @@ private function createPaymentTransactionsGroup(FormBuilderInterface $builder, O
...
+    ): FormBuilderInterface {
         $builderWithdrawalRequestGroup = $builder->create('withdrawalRequestGroup', GroupType::class, [
             'label' => 'Withdrawal Request',
         ]);
 
         $builderWithdrawalRequestGroup
-            ->add('withdrawalRequestData', OrderWithdrawalFormType::class, [
+            ->add('withdrawalRequest', OrderWithdrawalDisplayType::class, [
                 'label' => false,
+                'mapped' => false,
```
grossmannmartin: is it not possible to set this as default in the configureOptions method of the OrderWithdrawalDisplayType class?

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Order/WithdrawalRequest/OrderWithdrawalRequestDecorator.types.yaml
```diff
@@ -0,0 +1,23 @@
...
+    config:
+        fields:
+            firstName:
+                type: "String!"
+                description: "First name for withdrawal request contact"
+            lastName:
+                type: "String!"
+                description: "Last name for withdrawal request contact"
+            telephone:
+                type: "String!"
```
grossmannmartin: this shouldn't be required fie as the entity field is nullable, or?

packages/framework/src/Model/Order/Mail/WithdrawalAdminMailFacade.php
```diff
@@ -52,8 +52,11 @@ public function sendEmail(Order $order): void
      */
     protected function getMailBody(Order $order): string
     {
+        $withdrawalRequest = $order->getWithdrawalRequestThrowExceptionWhenNull();
```
grossmannmartin: The method name feels overly descriptive. Usually, I expect that the `get<Subject>()` method returns the object or throw exception. In rare cases, when it may return null, I can see it from the method return value.

packages/framework/src/Model/Order/Order.php
```diff
@@ -891,6 +892,20 @@ public function getWithdrawalRequest()
         return $this->withdrawalRequest;
     }
 
+    /**
+     * @return \Shopsys\FrameworkBundle\Model\Order\Withdrawal\WithdrawalRequest
+     */
+    public function getWithdrawalRequestThrowExceptionWhenNull()
```
grossmannmartin: Oh, now I see why the wordy name 😄 

What about `getWithdrawalRequest()` / `findWithdrawalRequest()`?

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/AdvancedSearch/AdvancedSearchOperatorTranslation.php
```diff
@@ -29,6 +29,8 @@ public function __construct()
             AdvancedSearchFilterInterface::OPERATOR_LT => t('lower than'),
             AdvancedSearchFilterInterface::OPERATOR_GTE => t('higher or equal'),
             AdvancedSearchFilterInterface::OPERATOR_LTE => t('lower or equal'),
+            AdvancedSearchFilterInterface::OPERATOR_EXISTS => t('exists'),
+            AdvancedSearchFilterInterface::OPERATOR_DOES_NOT_EXIST => t('does not exist'),
```
grossmannmartin, line 33: This is nice :) 

Maybe we could use already existing operators. like "Withdrawn from the contract -> is / is not"
But on the other hand, these new operators would be better in product advanced search, field "Has X+Y promotion" 😂

What was I saying? Oh, right - This is nice :)

packages/framework/src/Model/AdvancedSearchOrder/Filter/OrderWithdrawalRequestFilter.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function extendQueryBuilder(QueryBuilder $queryBuilder, array $rulesData): void
+    {
+        foreach ($rulesData as $ruleData) {
+            if ($ruleData->operator === self::OPERATOR_EXISTS) {
+                $queryBuilder->andWhere('o.withdrawalRequest IS NOT NULL');
```
grossmannmartin: I would like to discuss one more thing - the owning side of the 1:1 relation.
I have noticed we have foreign key on both sides (order and withdrawal_request). 

This is brittle (on conflict, what is the source of truth?), cost more on insert and it's not even possible to delete a withdrawal (ok, that is not probably needed anyway). 

I would keep the FK only in the withdrawal_requests table with onDelete cascade set (because that prevent existence of the "orphaned" withdrawals)
(1:1 relation is enforced by unique index on the order_id column)

upgrade-notes/backend_20251111_135924.md
```diff
@@ -8,4 +8,6 @@
 - `Shopsys\FrameworkBundle\Model\Order\Mail` class was changed:
     - `getDisplayPrice()` method was moved to the new `MailDisplayPriceResolver` class
     - `DISPLAY_PRICE_WITH_VAT`, `DISPLAY_PRICE_WITHOUT_VAT`, `DISPLAY_PRICE_BOTH`, and `DISPLAY_PRICE_SELLING` constants were moved to the new `MailDisplayPriceResolver` class
+- `Shopsys\FrontendApiBundle\Model\Order\OrderItemApiFacade::createOrderItemQueryBuilder()` was renamed to `createOrderItemExcludingOrdersWithWithdrawalQueryBuilder()`
+    - the method now excludes orders with withdrawal requests from the result
```
grossmannmartin: nitpick - this method always excluded orders with withdrawal requests as no withdrawal requests existed before

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: :ok_hand: 8189b3593e

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Order/WithdrawalRequest/OrderWithdrawalRequestDecorator.types.yaml
vitek-rostislav: yes, good catch, thanks, 07aac0c8c1

upgrade-notes/backend_20251111_135924.md
vitek-rostislav: true, removed in 828bcde40d

packages/framework/src/Model/Order/Mail/WithdrawalAdminMailFacade.php
vitek-rostislav: reworked within 6513d115b9

packages/framework/src/Model/Order/Order.php
vitek-rostislav: reworked within 6513d115b9

packages/framework/src/Model/AdvancedSearchOrder/Filter/OrderWithdrawalRequestFilter.php
vitek-rostislav: reworked within 6513d115b9

packages/framework/src/Model/Order/Withdrawal/WithdrawalChecker.php
```diff
@@ -18,11 +18,13 @@ class WithdrawalChecker
      * @param \Shopsys\FrameworkBundle\Model\Order\Withdrawal\WithdrawalDeadlineCalculation $withdrawalDeadlineCalculation
      * @param \Shopsys\FrameworkBundle\Model\Order\OrderFacade $orderFacade
      * @param \Shopsys\FrameworkBundle\Component\Domain\Domain $domain
+     * @param \Shopsys\FrameworkBundle\Model\Order\Withdrawal\WithdrawalRequestRepository $withdrawalRequestRepository
      */
     public function __construct(
         protected readonly WithdrawalDeadlineCalculation $withdrawalDeadlineCalculation,
         protected readonly OrderFacade $orderFacade,
         protected readonly Domain $domain,
+        protected readonly WithdrawalRequestRepository $withdrawalRequestRepository,
```
grossmannmartin, line 24: I would expect facade here tbh. Is a reason to use directly a repository?

packages/frontend-api/src/Model/Order/OrderItemApiFacade.php
```diff
@@ -45,7 +46,8 @@ protected function createOrderItemExcludingOrdersWithWithdrawalQueryBuilder(): Q
             ->select('oi')
             ->from(OrderItem::class, 'oi')
             ->join('oi.order', 'o')
-            ->andWhere('o.withdrawalRequest IS NULL');
+            ->leftJoin(WithdrawalRequest::class, 'wr', 'WITH', 'wr.order = o')
```
grossmannmartin: I would use 
 ```
->join('oi.order', 'o')
->andWhere('NOT EXISTS(
    SELECT 1 FROM ' . WithdrawalRequest::class . ' wr 
    WHERE wr.order = o
)')
```

this should be slightly faster as it's not necessary to build a full join result before filtering (ADD - assumption driven development 😁)

File: packages/frontend-api/src/Model/Resolver/Order/OrderResolverMap.php [hunk 45]
```diff
@@ -54,6 +57,9 @@ protected function map(): array
                 'withdrawalDeadline' => function (Order $order) {
                     return $this->withdrawalFacade->getWithdrawalDeadline($order);
                 },
+                'withdrawalRequest' => function (Order $order) {
+                    return $this->withdrawalRequestFacade->findByOrder($order);
```
grossmannmartin: this is slightly confusing. 
What is the difference between `WithdrawalFacade` and `WithdrawalRequestFacade`? (this is rhetorical question :D)

project-base/app/config/packages/messenger.yaml
```diff
@@ -15,6 +15,7 @@ framework:
             Shopsys\FrameworkBundle\Model\Order\Messenger\PlacedOrderMessage: placed_order_transport
             Symfony\Component\Mailer\Messenger\SendEmailMessage: send_email_transport
             Shopsys\FrameworkBundle\Model\Article\Messenger\ArticleExportMessage: article_export_transport
+            Shopsys\FrameworkBundle\Model\Order\Withdrawal\Messenger\WithdrawalRequestMessage: send_email_transport
```
grossmannmartin, line 18: My preference is slightly in favor of a separate queue. But very slightly as it's possible to change from project and new queue would require many more changes.

Maybe I'm leaning back to keep it unified 😁

grossmannmartin (approved):

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -76,7 +81,15 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
                 'order' => $order,
             ])
-            ->add($this->createPaymentTransactionsGroup($builder, $order))
+            ->add($this->createPaymentTransactionsGroup($builder, $order));
+
+        $withdrawalRequest = $this->withdrawalRequestFacade->findByOrder($order);
+
+        if ($withdrawalRequest !== null) {
+            $builder->add($this->createWithdrawalRequestGroup($builder, $withdrawalRequest));
+        }
```
henzigo: What about doing something like:

```php
final class OrderWithdrawalDisplayType extends AbstractType
{
    public function __construct(
        private readonly DateTimeFormatterExtension $dateTimeFormatterExtension,
        private readonly WithdrawalRequestFacade $withdrawalRequestFacade,
    ) {
    }

    #[Override]
    {
        /** @var \Shopsys\FrameworkBundle\Model\Order\Order $order */
        $order = $options['order'];
        $withdrawalRequest = $this->withdrawalRequestFacade->findByOrder($order);

        if ($withdrawalRequest === null) {
            $builder->addEventListener(FormEvents::PRE_SET_DATA, static function (FormEvent $event): void {
                $form = $event->getForm();
                if ($parent = $form->getParent()) {
                    $parent->remove($form->getName());
                }
            });
            return;
        }

        $group = $builder->create('withdrawalRequestGroup', GroupType::class, [
            'label' => 'Withdrawal Request',
        ]);

        $group
            ->add('firstName', DisplayOnlyType::class, [
                'label' => 'First name',
                'data' => $withdrawalRequest->getFirstName(),
            ])
            ->add('lastName', DisplayOnlyType::class, [
                'label' => 'Last name',
                'data' => $withdrawalRequest->getLastName(),
            ])
            ->add('telephone', DisplayOnlyType::class, [
                'label' => 'Phone',
                'data' => $withdrawalRequest->getTelephone(),
            ])
            ->add('email', DisplayOnlyType::class, [
                'label' => 'Email',
                'data' => $withdrawalRequest->getEmail(),
            ])
            ->add('note', DisplayOnlyType::class, [
                'label' => 'Note',
                'data' => $withdrawalRequest->getNote(),
            ])
            ->add('requestedAt', DisplayOnlyType::class, [
                'label' => 'Requested at',
                'data' => $this->dateTimeFormatterExtension->formatDateTime($withdrawalRequest->getRequestedAt()),
            ]);

        $builder->add($group);
    }

    #[Override]
    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver
            ->setRequired('order')
            ->setAllowedTypes('order', Order::class)
            ->setDefaults([
                'label' => false,
                'mapped' => false,
            ]);
    }
}
```

And here just load that type:

```
$builder->add('withdrawal', OrderWithdrawalDisplayType::class, [
    'order' => $order,
]);
```

Looks cleaner and more encapsulated for me

packages/framework/src/Model/Order/Withdrawal/WithdrawalChecker.php
vitek-rostislav: yes, originally, there was a facade, but this was another place where the circular reference occurs (`WithdrawalRequestFacade` has a dependency on `WihdrawalChecker`, it uses it in `createWithdrawalRequest` method, and the checker needs the facade to find the withdrawal by order :melting_face: Another (and maybe better) solution would be splitting `WithdrawalRequestFacade` into two services - one for requests creating, another for fetching them :shrug: But at this point, I am totaly ok with using the repository within the checker :smile:

packages/frontend-api/src/Model/Order/OrderItemApiFacade.php
vitek-rostislav: yeah, I had both solutions and the left join way seemed better readable to me. The performance difference should be insignificant AFAIK. But the truth is, my original implementation of the subselect was using DQL and that was definitely worse readable than your suggestion, and as there are two people who prefer the subselect way, here it is :slightly_smiling_face: 9a09b3d955

packages/frontend-api/src/Model/Resolver/Order/OrderResolverMap.php
vitek-rostislav: in the context of this pull request, there are no rhetorical questions. What about this? a24b75a06d

And please, do not ask another rhetorical question "what is the difference between `WithdrawalSettingFacade` and `WithdrawalSetting` classes?" :joy:

project-base/app/config/packages/messenger.yaml
vitek-rostislav: ok, at the moment, I am keeping it as-is. I would create a separate queue in the platform once we need it (when we use the message for other-than-sending-mails purpose). Also, I am putting a faith into the project developers - if they need to use the message for something else, they will create a separate queue for that for sure :crossed_fingers:

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: I like my way better :shrug: :innocent:

grossmannmartin (approved):

packages/administration/translations/messages.sk.po
```diff
@@ -1936,6 +1936,9 @@ msgstr "Tento stav nemôže byť odstránený, pretože je naň viakaná funkcio
 msgid "This status can't be deleted, because there is a functionality bounded to it - it identifies orders that were successfully processed."
 msgstr "Tento stav nemôže byť odstránený, pretože je naň viakaná funkcionalita - identifikuje objednávky, ktoré boli úspešne spracované."
 
+msgid "This status can't be deleted, because there is a functionality bounded to it - it identifies orders with withdrawal requests."
+msgstr "Tento stav nemôže byť odstránený, pretože je naň viakaná funkcionalita - určuje objednávky s požiadavkou na odstúpenie od zmluvy."
```
grossmannmartin: Isn't there a typo in the first part of the message? (same for other messages)

```
Tento stav nie je možné zmazať, pretože je naň naviazaná funkčnosť
```

packages/framework/src/Migrations/Version20251124135249.php
```diff
@@ -0,0 +1,87 @@
...
+                'INSERT INTO order_status_translations (translatable_id, name, locale) VALUES (:id, :name, :locale)',
+                [
+                    'id' => $nextId,
+                    'name' => t('Withdrawn', [], Translator::DEFAULT_TRANSLATION_DOMAIN, $locale),
+                    'locale' => $locale,
+                ],
+            );
+        }
+
+        $this->sql('SELECT setval(\'order_statuses_id_seq\', GREATEST(MAX(id), :nextId)) FROM order_statuses', ['nextId' => $nextId]);
```
grossmannmartin: Wouldn't it be better to let PostgreSQL handle sequences automatically and just select the associated ID instead of forcing it and setval the sequence?

I mean 
`insert into order_statuses (type)`
`$orderStatusId = $this->connection->lastInsertId();`

packages/framework/src/Model/Order/Mail/WithdrawalMailFacade.php
```diff
@@ -0,0 +1,35 @@
...
+    public function __construct(
+        protected readonly OrderMailFacade $orderMailFacade,
+        protected readonly OrderStatusFacade $orderStatusFacade,
+    ) {
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Order\Order $order
+     */
+    public function sendMail(Order $order): void
```
grossmannmartin: I'm looking at the usage. Now the flow is actually doubled. When I create the withdrawal, and when I change the status in the administration. 
Couldn't we unify it so request withdrawal will execute the same code as changing the status in the administration?

Also, now the impossibility of creating a withdrawal in the administration really stands out. 
When I change the status to "withdrawn", I cannot add the withdrawal info, and the order looks different from when the customer requested it. 

I would rather 
- allow adding the withdrawal info in the admin (here we go again)
- disallow set withdrawn status in the administration

packages/framework/src/Migrations/Version20251124135249.php
vitek-rostislav: Ah, of course, that is way better, thank you! 136322ecb5

packages/administration/translations/messages.sk.po
vitek-rostislav: yes :sweat_smile: a315aa0264

packages/framework/src/Migrations/Version20251124135249.php
vitek-rostislav: The only thing that bothers me - because of that already removed vintage "over limit" order status, the next sequence number is actually 6, so we now have a gap there. Should we care? 
image

packages/framework/src/Model/Order/Mail/WithdrawalMailFacade.php
vitek-rostislav: Yes, the way of changing the status in the administration messes things up horribly :sweat_smile:  Let's talk about it, please :pray:

packages/framework/src/Migrations/Version20251124135249.php
grossmannmartin: no, we don't care :)

packages/framework/assets/js/admin/components/OrderWithdrawal.js
```diff
@@ -0,0 +1,37 @@
+import Register from '../../common/utils/Register';
+
+export default class OrderWithdrawal {
+    constructor($container) {
+        this.$statusSelect = $container.find('.js-order-status-select');
+        this.$withdrawalGroup = $container.find('.js-order-withdrawal-group');
+        this.handleStatusChange();
+
+        if ($(this.$withdrawalGroup).data('withdrawal-request-exists')) {
```
grossmannmartin: this.$withdrawalGroup is already a jQuery object. It's probably not necessary to wrap it again

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -135,6 +143,15 @@ private function createBasicInformationGroup(FormBuilderInterface $builder, Orde
         $builderBasicInformationGroup = $builder->create('basicInformationGroup', GroupType::class, [
             'label' => 'Basic information',
         ]);
+        $withdrawalRequest = $this->withdrawalRequestFacade->findByOrder($order);
```
grossmannmartin, line 146: nitpick

I have a small concern for this method being called twice (maybe more, not sure how the computing validation group works exactly). 
Maybe we could set an option in configureDefaults and then get the withdrawalRequest from the option here and in the validation group callback?

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -173,10 +190,20 @@ private function createBasicInformationGroup(FormBuilderInterface $builder, Orde
...
                 'choice_value' => 'id',
+                'choice_attr' => function (OrderStatus $orderStatus) {
+                    return [
+                        'data-js-order-status-type' => $orderStatus->getType(),
+                    ];
+                },
                 'multiple' => false,
                 'expanded' => false,
+                'attr' => [
+                    'class' => 'js-order-status-select',
```
grossmannmartin: I would use a separate data attribute here. 
When we set a class here, we may overwrite the default classes rendered in the template.

```
'attr' => [
    'data-js-order-status-select' => null
```

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -612,20 +639,30 @@ private function createPaymentTransactionsGroup(FormBuilderInterface $builder, O
...
      * @return \Symfony\Component\Form\FormBuilderInterface
      */
     private function createWithdrawalRequestGroup(
         FormBuilderInterface $builder,
-        WithdrawalRequest $withdrawalRequest,
+        ?WithdrawalRequest $withdrawalRequest,
     ): FormBuilderInterface {
+        $rowAttr = [
+            'class' => 'js-order-withdrawal-group',
+            'data-withdrawal-request-exists' => $withdrawalRequest !== null,
```
grossmannmartin: Do we need both? I think we can safely use just the data attribute as a selector, or?

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -612,20 +639,30 @@ private function createPaymentTransactionsGroup(FormBuilderInterface $builder, O
...
+
         $builderWithdrawalRequestGroup = $builder->create('withdrawalRequestGroup', GroupType::class, [
             'label' => 'Withdrawal Request',
+            'row_attr' => $rowAttr,
         ]);
 
         $builderWithdrawalRequestGroup
-            ->add('withdrawalRequest', OrderWithdrawalDisplayType::class, [
-                'withdrawal_request' => $withdrawalRequest,
+            ->add('withdrawalRequestData', OrderWithdrawalFormType::class, [
```
grossmannmartin, line 663: Do we really want a group inside a group? I don't find it pleasing to the eye 

Snímek obrazovky 2025-11-25 v 21 07 48

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: Yes, I have had the exact train of thought during the implementation. I checked the DB queries, there are 2 of them, each takes a few milliseconds (~2ms). I also thought about the option, but it seemed a little bit weird to require the withdrawal request in the options of the order form :shrug: I also do not like the fact that the form type is created in two controller actions (`editAction` and for some reason, `addProductAction`) so I would have to modify both of them. So if you do not mind, I am keeping it as is for now

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: yeah, well, ... :sweat_smile: I discussed that with the product owner during the implementation, and he was ok with that. 

Another option would be moving that group under the basic information group, but that would be too far from the order status select box. 

Or I could move the items from the group, but that would probably mean adding the word "withdrawal" to all the related fields to preserve clarity. 

So in the end, the group in group solution seems like the lesser evil :smile: And maybe, one day, we will reorganise the form completely so :shrug: :smile:

packages/framework/assets/js/admin/components/OrderWithdrawal.js
vitek-rostislav: fixed in c23f7b8776

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: fixed in c23f7b8776

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: fixed in c23f7b8776

grossmannmartin (approved):

JanMolcik (approved):

JanMolcik (approved):

project-base/storefront/components/Pages/Customer/OrderDetail/OrderDetailWithdrawalSection.tsx
```diff
@@ -41,78 +40,69 @@ export const OrderDetailWithdrawalSection: FC<OrderDetailWithdrawalSectionProps>
...
+                        {t('If you have an issue with this order, you can create a')}{' '}
+                        <ExtendedNextLink
+                            aria-label={t('Go to create complaint page', { ns: 'accessibility' })}
+                            className="text-sm"
+                            href={newComplaintUrl}
+                            type="complaintNew"
+                        >
+                            {t('complaint')}
+                        </ExtendedNextLink>
+                        .
```
JanMolcik: It would be nice to use the `<Trans ...>` component with `link` key here for better readability (similar to `@project-base/storefront/components/Pages/Order/ContactInformation/FormBlocks/ContactInformationSendOrderButton.tsx`).

JanMolcik (changes requested):

JanMolcik (approved):

PR #4250, opened 2025-10-27
[CRUD] Access control implementation

henzigo:

Implement RBAC system to CRUD to provide automatic security features.

- Role is created for enabled actions automatically if not configured
- RouteAccessControl now also checks CRUD routes
- Internal working with CRUD was simplified to provide an easier way for future implementation
- Some bugs in RBAC and CRUD were fixed

Testing CRUD controller: https://github.com/shopsys/shopsys/commit/a03313a1877e36ed75aa0d15314031aaf0ea77a4

#### Fixes issues
... 

----
  - https://jg-crud-access-control.odin.shopsys.cloud
  - https://cz.jg-crud-access-control.odin.shopsys.cloud
  - https://jg-crud-access-control.odin.shopsys.cloud/sk


packages/administration/src/Component/Crud/Helper/CrudTransformationHelper.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Crud\Helper;
+
+use Shopsys\AdministrationBundle\Component\Config\ActionType;
+
+final class CrudTransformationHelper
```
vitek-rostislav, line 9: this class deserves tests IMHO

packages/administration/src/Component/Crud/CrudAdminRoleProvider.php
```diff
@@ -0,0 +1,73 @@
...
+
+            $role = new Role(
+                $crudControllerDefinition->getRoleConstant(),
+                $config->getMenuTitle(),
+                Permission::getHighestLevelPermissions($requiredPermissions),
+            );
+
+            $role->setRoleSection($config->getCustomRoleSection() ?? $config->getMenuSection());
+            $role->setOverwritable(false);
+
```
vitek-rostislav, line 69: what if they need to change it on a project? Should they do that by reconfiguring given CRUD controller in project-base via extension? Will this work?

packages/administration/tests/Unit/Component/Security/Attribute/AttributeProcessorTest.php
```diff
@@ -388,8 +388,8 @@ public function testMethod(): void
             }
         };
 
-        $method = new ReflectionMethod($testClass, 'testMethod');
-        $rules = $this->processor->processMethod($method);
+        $reflectionClass = new ReflectionClass($testClass);
+        $rules = $this->processor->processMethod($reflectionClass, $reflectionClass->getMethod('testMethod'));
```
vitek-rostislav, line 393: this repeats so many times that it might deserve a helper method. But it is rather a nitpick

packages/administration/src/Component/Router/CrudControllerRouteLoader.php
```diff
@@ -80,20 +68,20 @@ public function setResolver(LoaderResolverInterface $resolver)
...
 
-        if ($config->isFullDisabled()) {
-            return;
-        }
+            $route = $routeItem->getRoute();
 
-        foreach ($config->getActions() as $actionType) {
-            $routeItem = $this->crudRouteProvider->generate($item, $actionType);
+            $route->setDefault('_crud_controller', true);
+            $route->setDefault('_crud_action', $actionType->value);
```
vitek-rostislav: maybe there could be constants for these names?

vitek-rostislav (changes requested):

packages/administration/src/Component/Crud/CrudAdminRoleProvider.php
henzigo: This `Role` is configured after are extensions are loaded, so you can always enable/disable actions or do what you want.

And if you want to use your custom-defined `Role`, you can set it via `setCustomRoleConstant` method. In that case this Role is not generated.

packages/administration/src/Component/Crud/Helper/CrudTransformationHelper.php
henzigo: Ok why not

packages/administration/src/Component/Router/CrudControllerRouteLoader.php
henzigo: fixed

packages/administration/tests/Unit/Component/Security/Attribute/AttributeProcessorTest.php
henzigo: I don't think that's entirely necessary.

packages/administration/src/Component/Crud/Helper/CrudTransformationHelper.php
henzigo: Added as last fixup commit

packages/administration/tests/Unit/Component/Security/Attribute/AttributeProcessorTest.php
vitek-rostislav: ok

vitek-rostislav:
CR ok :heavy_check_mark: 

Testing:
- [current issue, but related] I noticed on the dashboard, there are links to pages that an admin does not have access to ("product search + advanced search, transfer issues)
- https://docs.shopsys.com/en/17.0/administration/crud-controller/getting-started/creating-a-new-crud-controller/#2-configure-crud-controller needs to be updated, `configure` is public now
- [current issue, but related] in `configureDatagrid`, when I use non existing route (e.g. in `$action->linkToRoute`), I do not get any error that the route does not exist
- is it a wanted behavior that when I create CRUD controller, it has all the actions in the config, however, I need to explicitly allow them using `$config->enableAction(ActionType::DELETE);` in `configure` method? It is quite unintuitive that I see all the permissions  in the roles grid for the new controller, however I do not see the row actions in the grid unless I enable them...ok update, I see them in the roles but I can not assign them, that is ok. But still, I would expect them to be enabled by default (when they are present in the configuration, or?

henzigo:
> [current issue, but related] I noticed on the dashboard, there are links to pages that an admin does not have access to ("product search + advanced search, transfer issues)

Yes, also `New` buttons are visible even if you don't have permission to do that. This is the current state, but we should look at it soon.

> https://docs.shopsys.com/en/17.0/administration/crud-controller/getting-started/creating-a-new-crud-controller/#2-configure-crud-controller needs to be updated, configure is public now

Fixed

> [current issue, but related] in configureDatagrid, when I use non existing route (e.g. in $action->linkToRoute), I do not get any error that the route does not exist

Fixed. The URL is generated and validated before permissions are checked. But there should be a different exception currently, when you call `$this->generateUrl('asd')` in any admin controller, it's trying to resolve the current domain config on the chain router, which should not be used. We should look at and fix it. 

Look: [1921318](https://github.com/shopsys/shopsys/pull/4250/commits/19213180b1cdb4dba022331b9516061005d2a511)

I also added a hint for Symfony plugin to autocomplete routes for `linkToRoute` method: [a1d8c08](https://github.com/shopsys/shopsys/pull/4250/commits/a1d8c08f00241c477647dd370df41350510dbc41)

> is it a wanted behavior that when I create CRUD controller, it has all the actions in the config, however, I need to explicitly allow them using $config->enableAction(ActionType::DELETE); in configure method? It is quite unintuitive that I see all the permissions in the roles grid for the new controller, however I do not see the row actions in the grid unless I enable them...ok update, I see them in the roles but I can not assign them, that is ok. But still, I would expect them to be enabled by default (when they are present in the configuration, or?

In the next PRs, I will introduce `Handler` classes that must be implemented to process those actions. It will also automatically enable actions for the implemented `Handler`. Without that, you cannot process that action correctly. Currently, other actions are not implemented, so disabling those actions seems reasonable to me.

vitek-rostislav (approved):

PR #4255, opened 2025-10-31
Correctly escape the src attribute in img tag

florianjiri:

Correctly escape the src attribute in img tag

### DIFF
--- a/packages/framework/src/Resources/views/Common/image.html.twig
+++ b/packages/framework/src/Resources/views/Common/image.html.twig
@@ -2,7 +2,13 @@
     <picture>
         {% apply spaceless %}
             {% set attr = attr|merge({'class': (attr.class|default('') ~ ' ' ~ imageCssClass)|trim}) %}
-            <img {% for attrname, attrvalue in attr %} {{ attrname }}="{{ attrvalue }}"{% endfor %} itemprop="image" />
+            <img {%- for attrname, attrvalue in attrForMS %}
+            {%- if attrname == 'src' -%}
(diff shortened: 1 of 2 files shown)


packages/framework/src/Resources/views/Common/image.html.twig
```diff
@@ -2,7 +2,13 @@
     <picture>
         {% apply spaceless %}
             {% set attr = attr|merge({'class': (attr.class|default('') ~ ' ' ~ imageCssClass)|trim}) %}
-            <img {% for attrname, attrvalue in attr %} {{ attrname }}="{{ attrvalue }}"{% endfor %} itemprop="image" />
+            <img {%- for attrname, attrvalue in attrForMS %}
+            {%- if attrname == 'src' -%}
+                {{- " " ~ attrname }}="{{ attrvalue|escape_but_amp }}"
```
pk16011990, line 7: I think we should use `{{ attrvalue|e('html_attr') }}`. Char `&` must be escaped in HTML attributes by specification. Or not?

pk16011990: Or we can use custom escapers https://twig.symfony.com/doc/3.x/filters/escape.html#custom-escapers

grossmannmartin:
According to the html spec, ampersand should be converted to html entity in the attribute values to avoid possible [ambiguous ampersand](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2)

Any html complaint client has to decode the value before interpreting.

closing

PR #4257, opened 2025-11-03
ensure AccessChecker::hasPermission works properly with single roles

vitek-rostislav:
When a single role is passed to `AccessChecker::hasPermission`, we must not append the permission value to the role constant. 

It should not happen very often, however, we had one bug caused by this. `CurrencyInlineEdit` defines its role constant as `SystemRole::SUPER_ADMIN` (ensuring only superadmins can work with the currencies in the inline grid), however, since the permission value was appended internally to the constant name, superadministrators were not able to create/edit/delete currencies at all.

Moreover, `SystemRole::ALL` and `SystemRole::ALL_VIEW` were removed from the role collection to avoid their misuse. Such roles serve only to mimic role groups covering all child roles and should not be used within security attributes like `CanCreate`, `CanEdit`, `ForRole`, etc.

#### Fixes issues
... 

## Release Notes

* **Bug Fixes**
  * Improved permission checking performance for single-role access scenarios.

* **Chores**
  * Removed "Full access" and "Full access (view-only)" system roles from default admin role configuration.

----
  - https://rv-currency.odin.shopsys.cloud
  - https://cz.rv-currency.odin.shopsys.cloud
  - https://rv-currency.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/translations/messages.cs.po
+++ b/packages/administration/translations/messages.cs.po
@@ -907,12 +907,6 @@ msgstr "Pátek"
 msgid "From"
 msgstr "Od"
 
-msgid "Full access"
-msgstr "Plný přístup"
-
-msgid "Full access (view only)"
-msgstr "Plný přístup (pouze prohlížení)"
-
 msgid "Future"
 msgstr "Budoucí"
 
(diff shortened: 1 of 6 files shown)


vitek-rostislav:
@coderabbitai review

packages/administration/translations/messages.cs.po
```diff
@@ -907,12 +907,6 @@ msgstr "Pátek"
 msgid "From"
 msgstr "Od"
 
-msgid "Full access"
-msgstr "Plný přístup"
-
-msgid "Full access (view only)"
-msgstr "Plný přístup (pouze prohlížení)"
-
```
henzigo, line 915: Why are those messages removed? They are used for Role group selectbox

vitek-rostislav: They are never used within the `administration` package anymore. They are still used and translated in the `framework` package :wink:

grossmannmartin (approved):

PR #4261, opened 2025-11-04
fixed group admin form fields in cards

grossmannmartin:

If form contain fields that are not inside a group, they are now still rendered inside a card.
Form types that renders card by their nature (ProductsType, ...) now use interface to mark this behavior (and disable auto-placing inside a card)
+ Fixed few visual incosistencies across the administration 

#### Fixes issues
... 

## Summary by CodeRabbit

* **New Features**
  * Added automatic card-based grouping system for form fields with improved layout organization.

* **Style**
  * Restructured admin interface templates with Bootstrap card layouts for better visual organization across multiple admin pages.
  * Refined spacing and margins on various admin interface cards for improved visual hierarchy.

* **Refactor**
  * Reorganized form field structure and grouping to support improved card-based rendering throughout the admin panel.

----
  - https://mg-admin-auto-card.odin.shopsys.cloud
  - https://cz.mg-admin-auto-card.odin.shopsys.cloud
  - https://mg-admin-auto-card.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/templates/content/administrator/edit.html.twig
+++ b/packages/administration/templates/content/administrator/edit.html.twig
@@ -32,13 +32,13 @@
                             <div class="card timeline-event-card">
                                 <div class="card-body">
                                     <div class="text-secondary float-end">
-                                        {{ 'Last action' }}:
+                                        {{ 'Last action'|trans }}:
                                         {{ adminActivity.lastActionTime|formatDateTime }}
                                     </div>
-                                    <h4 class="mb-2">{{ 'Administrator logged in' }}</h4>
+                                    <h4 class="mb-2">{{ 'Administrator logged in'|trans }}</h4>
                                     <div class="text-secondary">
                                         {{ adminActivity.loginTime|formatDateTime }}
-                                        {{ 'from IP' }}
+                                        {{ 'from IP'|trans }}
                                         {{ adminActivity.ipAddress }}
                                     </div>
                                 </div>
--- a/packages/administration/templates/content/administrator/edit.html.twig
+++ b/packages/administration/templates/content/administrator/edit.html.twig
@@ -18,29 +18,35 @@
 
 {% block admin_log %}
     {% if lastAdminActivities|length > 0 %}
-        <h2>{{ 'Administrator activities history'|trans }}</h2>
-        <ul class="timeline">
-            {% for adminActivity in lastAdminActivities %}
-                <li class="timeline-event">
-                    <div class="timeline-event-icon">
-                        {{ ux_icon('lock') }}
-                    </div>
-                    <div class="card timeline-event-card">
-                        <div class="card-body">
-                            <div class="text-secondary float-end">
-                                {{ 'Last action' }}:
-                                {{ adminActivity.lastActionTime|formatDateTime }}
+        <div class="card">
+            <div class="card-header">
+                <h3 class="card-title">{{ 'Administrator activities history'|trans }}</h3>
+            </div>
+            <div class="card-body">
+                <ul class="timeline">
+                    {% for adminActivity in lastAdminActivities %}
+                        <li class="timeline-event">
+                            <div class="timeline-event-icon">
+                                {{ ux_icon('lock') }}
                             </div>
-                            <h4 class="mb-2">{{ 'Administrator logged in' }}</h4>
-                            <div class="text-secondary">
-                                {{ adminActivity.loginTime|formatDateTime }}
-                                {{ 'from IP' }}
-                                {{ adminActivity.ipAddress }}
+                            <div class="card timeline-event-card">
+                                <div class="card-body">
+                                    <div class="text-secondary float-end">
+                                        {{ 'Last action' }}:
+                                        {{ adminActivity.lastActionTime|formatDateTime }}
+                                    </div>
+                                    <h4 class="mb-2">{{ 'Administrator logged in' }}</h4>
+                                    <div class="text-secondary">
+                                        {{ adminActivity.loginTime|formatDateTime }}
+                                        {{ 'from IP' }}
+                                        {{ adminActivity.ipAddress }}
+                                    </div>
+                                </div>
                             </div>
-                        </div>
-                    </div>
-                </li>
-            {% endfor %}
-        </ul>
+                        </li>
+                    {% endfor %}
+                </ul>
+            </div>
+        </div>
     {% endif %}
 {% endblock %}
--- a/packages/administration/templates/content/transportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
+++ b/packages/administration/templates/content/transportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
@@ -28,7 +28,6 @@
                             <tr id="{{ priceLimitForm.vars.id }}" class="js-free-transport-and-payment-price-limit" data-domain-id="{{ domainId }}">
                                 {% if isMultidomain() %}
                                     <td>
-                                        {{ form_errors(priceLimitForm.priceLimit) }}
                                         {{ domainName }}
                                     </td>
                                 {% endif %}
@@ -37,6 +36,7 @@
                                 </td>
                                 <td>
                                     {{ form_widget(priceLimitForm.priceLimit, { symbolAfterInput: currencySymbolByDomainId(domainId), attr: { class: 'js-free-transport-and-payment-price-limit-input' } }) }}
+                                    {{ form_errors(priceLimitForm.priceLimit) }}
                                 </td>
                             </tr>
                         {% endfor %}
--- a/packages/administration/templates/content/transportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
+++ b/packages/administration/templates/content/transportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
@@ -6,40 +6,44 @@
 {% block h1 %}{{ 'Free shipping and payment'|trans }}{% endblock %}
 
 {% block main_content %}
-    {{ form_start(form) }}
-    {{ form_errors(form) }}
-    <div class="table-responsive">
-        <table class="table table-vcenter">
-            <thead>
-                {% if isMultidomain() %}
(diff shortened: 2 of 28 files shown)


packages/framework/src/Form/AutoCardGroupableExtension.php
```diff
@@ -0,0 +1,107 @@
...
+
+use Override;
+use Shopsys\FormTypesBundle\ActionBarType;
+use Symfony\Component\Form\AbstractTypeExtension;
+use Symfony\Component\Form\Extension\Core\Type\FormType;
+use Symfony\Component\Form\Extension\Core\Type\HiddenType;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+
+final class AutoCardGroupableExtension extends AbstractTypeExtension
```
vitek-rostislav, line 16: should it be final? What if they want to add another type to invisibleTypes on a project?

vitek-rostislav: if it should be final indeed, then protected properties/methods do no make any sense, right?

vitek-rostislav:
@coderabbitai review

packages/administration/templates/content/administrator/edit.html.twig
```diff
@@ -7,29 +7,35 @@
...
+                                    <div class="text-secondary float-end">
+                                        {{ 'Last action' }}:
+                                        {{ adminActivity.lastActionTime|formatDateTime }}
+                                    </div>
+                                    <h4 class="mb-2">{{ 'Administrator logged in' }}</h4>
+                                    <div class="text-secondary">
+                                        {{ adminActivity.loginTime|formatDateTime }}
+                                        {{ 'from IP' }}
+                                        {{ adminActivity.ipAddress }}
+                                    </div>
```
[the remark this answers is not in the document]
vitek-rostislav: the trans filter was missing even in the original code, however, should it be added?

packages/administration/templates/content/uploadedFile/edit.html.twig
```diff
@@ -24,29 +24,33 @@
...
+                        <div class="card-footer">
+                            {{ entity.nameWithExtension }}
+                        </div>
+                    </div>
                 </div>
-
-                <div class="card-footer">
-                    {{ entity.nameWithExtension }}
+                <div class="col">
+                    <label class="form-label">Replace file</label>
```
vitek-rostislav: should we add trans here as well?

packages/administration/templates/content/uploadedFile/edit.html.twig
```diff
@@ -24,29 +24,33 @@
...
-
-            <div class="card card-sm" style="max-width: 17.5rem;">
-                <div class="position-relative ratio ratio-16x9 overflow-hidden">
-                    {{ uploadedFilePreviewWithLink(entity) }}
+    <div class="card">
+        <div class="card-body">
+
+            <div class="row mb-3">
+                <div class="col-3">
+                    <label class="form-label">File</label>
```
vitek-rostislav: ...and here...
image

packages/framework/src/Form/AutoCardGroupableExtension.php
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Invisible/utility types that don't break auto-card sequences
+     * These fields are skipped during grouping and rendered by form_rest()
+     *
+     * @var array<class-string>
+     */
+    protected array $invisibleTypes = [
+        ActionBarType::class,
+        HiddenType::class,
+    ];
```
henzigo: What about SubmitType itself?

packages/framework/src/Form/AutoCardGroupableExtension.php
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            $child = $form[$name];
+            $childFormType = $child->getConfig()->getType()->getInnerType();
+
+            if ($this->isInvisibleType($childFormType::class)) {
+                continue;
+            }
+
+            if ($childFormType instanceof FormTypeRenderedWithCardInterface) {
```
henzigo: What about defining a global form variable like `auto_group` or something like that, and using it instead of the interface? Default can be true and for specific groupTypes we can set it as false in default options.

It seems a more native way of how we should work with forms.

packages/administration/templates/form/theme.html.twig
```diff
@@ -71,3 +71,34 @@
 
     {{ parent('number_widget') }}
 {% endblock number_widget %}
+
+{%- block form_widget_compound -%}
```
henzigo, line 75: We don't need to change the behaviour of all compound widgets. We can set:

```
if ($form->getParent() === null) {
    $view->vars['block_prefixes'][] = 'auto_group';
}
```

and then use:

```
{% block auto_group_widget %}
    ...
{% endblock %}
```

What do you think?

packages/framework/src/Form/AutoCardGroupableExtension.php
grossmannmartin: good thinking, removed final in e5f0c60f35

packages/administration/templates/content/administrator/edit.html.twig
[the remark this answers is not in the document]
grossmannmartin: added in b3870b6c00

packages/administration/templates/content/uploadedFile/edit.html.twig
grossmannmartin: added in b3870b6c00

packages/administration/templates/content/transportAndPayment/freeTransportAndPaymentLimitSetting.html.twig
```diff
@@ -6,40 +6,44 @@
...
+                                        {{ domainName }}
+                                    </td>
+                                {% endif %}
+                                <td>
+                                    {{ form_widget(priceLimitForm.enabled, { attr: { class: 'js-free-transport-and-payment-price-limit-enabled' } } ) }}
+                                </td>
+                                <td>
+                                    {{ form_widget(priceLimitForm.priceLimit, { symbolAfterInput: currencySymbolByDomainId(domainId), attr: { class: 'js-free-transport-and-payment-price-limit-input' } }) }}
+                                </td>
+                            </tr>
```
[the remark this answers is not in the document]
grossmannmartin: fixed in 493327eaa4

packages/administration/templates/content/uploadedFile/edit.html.twig
```diff
@@ -24,29 +24,33 @@
...
+                    </div>
                 </div>
-
-                <div class="card-footer">
-                    {{ entity.nameWithExtension }}
+                <div class="col">
+                    <label class="form-label">Replace file</label>
+                    {{ form_widget(form.files) }}
+                    {{ form_errors(form.files) }}
                 </div>
```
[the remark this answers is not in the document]
grossmannmartin: wont fix. form_label does not render for binding anyway and the whole purpose is to not have col in class to avoid unnecessary wrapping

packages/administration/templates/content/uploadedFile/edit.html.twig
grossmannmartin: ... and here 4f9c200898 ...

packages/framework/src/Form/AutoCardGroupableExtension.php
grossmannmartin: I don't thing the submit button should be added. It will be rendered normally (e.g. wrapped in card if not, or attached into previously auto-created group). 

ActionBar is mentioned as it has a fixed position and hidden type because it's not visually rendered at all

packages/framework/src/Form/AutoCardGroupableExtension.php
grossmannmartin: that's pretty good idea 👍 
d66c8ecad5

packages/administration/templates/form/theme.html.twig
grossmannmartin: It looks safer. Fixed d2f9be3c25

packages/framework/src/Form/AutoCardGroupableExtension.php
grossmannmartin: Oh... we have standard that enforces final in form types and extensions... 
I changed it to private then. 
I guess if someone would want to extend the list of invisibleTypes he should create a new extension. 
But I don't think it will be a common scenario. I think use of a form option `render_in_own_card` should cover mos od use cases. 

-> 9d47709a69

packages/framework/src/Form/Admin/Customer/BillingAddressFormType.php
```diff
@@ -183,6 +182,7 @@ public function configureOptions(OptionsResolver $resolver): void
                 'data_class' => BillingAddressData::class,
                 'disableCompanyCustomerCheckbox' => false,
                 'attr' => ['novalidate' => 'novalidate'],
+                'render_in_own_card' => true,
```
vitek-rostislav: I like the new solution; however, `render_in_own_card` seems confusing to me as looking at the option, I am not sure whether it says the field **is rendered** in its own card, or it **needs to be rendered** in its own card. I would prefer something like `(is)_rendered_in_own_card` (or maybe `renders_in_own_card`) which seems more explicit. Moreover, it would be nice to add information into docs about this, probably [here](https://docs.shopsys.com/en/17.0/introduction/using-form-types/)?

packages/administration/templates/form/theme.html.twig
grossmannmartin: Reverted. It breaks rendering of simple elements (especially hidden fields)

packages/framework/src/Form/Admin/Customer/BillingAddressFormType.php
grossmannmartin: renamed to `renders_in_own_card ` in 67dbdb94be

vitek-rostislav (approved):

pk16011990 (approved):

PR #4263, opened 2025-11-05
Replace `CsrfProtection` annotation with attribute

henzigo:

This PR originally introduced a method to check whether a controller action is protected.
However, I started having bad dreams about still needing to use an annotation instead of an attribute.
Fortunately, those dreams are gone now. Good night 😆 

#### Fixes issues
... 

## Summary by CodeRabbit

* **New Features**
  * Admin: added ability to delete salesmen from the admin UI with confirmation and success feedback.

* **Refactor**
  * Migrated CSRF protection from legacy annotations to PHP 8 attributes across admin endpoints.
  * CSRF route protector now uses in-memory caching for faster protection checks.

* **Chores**
  * Updated documentation and framework wiring to support attribute-based CSRF declarations.

----
  - https://jg-csrf-attribute.odin.shopsys.cloud
  - https://cz.jg-csrf-attribute.odin.shopsys.cloud
  - https://jg-csrf-attribute.odin.shopsys.cloud/sk


packages/framework/src/Component/Router/Security/RouteCsrfProtector.php
```diff
@@ -106,9 +110,27 @@ protected function isProtected(ControllerEvent $event): bool
...
+        $controllerName = is_object($controller) ? get_class($controller) : $controller;
+
+        if ($this->inMemoryCache->hasItem(static::CSRF_ROUTES_CACHE_NAMESPACE, $controllerName, $actionMethod)) {
+            return $this->inMemoryCache->getItem(static::CSRF_ROUTES_CACHE_NAMESPACE, $controllerName, $actionMethod);
+        }
+
+        $method = new ReflectionMethod($controller, $actionMethod);
+        $isProtected = $this->annotationReader->getMethodAnnotation($method, CsrfProtection::class) !== null;
+
+        $this->inMemoryCache->save(static::CSRF_ROUTES_CACHE_NAMESPACE, $isProtected, $controllerName, $actionMethod);
```
grossmannmartin: nitpick 

```php
return $this->inMemoryCache->getOrSaveValue(
    static::CSRF_ROUTES_CACHE_NAMESPACE,
    function () use ($controllerName, $actionMethod) {
        $method = new ReflectionMethod($controllerName, $actionMethod);

        return ReflectionHelper::getMethodAttribute($method, CsrfProtection::class) !== null;
    },
    $controllerName,
    $actionMethod,
);
```

seems more readable, it avoids error prone duplication of keyparts in several cache calls and seems more readable

composer.json
```diff
@@ -209,6 +209,7 @@
         "phpstan/phpstan-symfony": "^2.0.6",
         "phpunit/phpunit": "^11.2.1",
         "psr/event-dispatcher": "^1.0.0",
+        "rector/rector": "^2.2",
```
grossmannmartin: I'm struggling quite a bit with adding a tool that will probably break with every next release.

There is a great chance it won't be used anytime soon and when someone would need it, he would have to fix it first (and that is more work than start from scratch imho)

Besides, the change made with rector is much quicker to do with two regular expression replaces

grossmannmartin (changes requested):

composer.json
henzigo: Ok removed

packages/framework/src/Component/Router/Security/RouteCsrfProtector.php
henzigo: Makes sense. Refactored

grossmannmartin (approved):

PR #4275, opened 2025-11-20
bump doctrine/persistence and twig/twig

grossmannmartin:

Bump necessary packages to fix builds on 14.5 branch
 

#### Fixes issues
... 

----
  - https://mg-14-fix-build.odin.shopsys.cloud
  - https://cz.mg-14-fix-build.odin.shopsys.cloud


composer.json
```diff
@@ -181,7 +181,7 @@
         "symfony/workflow": "^5.4",
         "tracy/tracy": "^2.4.13",
         "tuscanicz/enum": "^2.1",
-        "twig/twig": "^3.5.0 <3.10.0",
+        "twig/twig": "^3.14.0",
```
vitek-rostislav, line 184: "the latest" version is actually `3.22`, even though I suppose it is installed one way or another, I would probably be happier if it was defined here. On the other hand, I can live with the current state :smile:

vitek-rostislav: btw, IMHO, adding something like "cherry-picked/taken from https://github.com/shopsys/shopsys/pull/3443" into the commit messages for these kinds of commits could be helpful for future generations (as well as the current generation of the code-reviewers)  :slightly_smiling_face:

composer.json
```diff
@@ -97,7 +97,7 @@
         "doctrine/doctrine-migrations-bundle": "^3.2.2",
         "doctrine/migrations": "^3.4.1",
         "doctrine/orm": "^2.11.2",
-        "doctrine/persistence": "^2.4",
+        "doctrine/persistence": "^3.3.3",
```
vitek-rostislav, line 100: ...footnote as an addition to the last comment: Not that all the original pull requests are super-descriptive :grimacing: https://github.com/shopsys/shopsys/pull/3498

phpstan.neon
```diff
@@ -59,6 +59,8 @@ parameters:
             # in fact, anything can be passed to the load function, it just has a bad annotation
             message: '#^Parameter \#1 \$key of method Overblog\\DataLoader\\DataLoaderInterface::load\(\) expects string#'
             path: %currentWorkingDirectory%/*
+        -
```
vitek-rostislav, line 62: Do we still call the clear method with a parameter anywhere? Ah, I see, in 14.5 version, we do. I was trying to find when did we got rid of that, but I was not successful. I would love to resolve the issue rather than ignoring it, however, I can see it is out of the scope of this pull request :shrug:

vitek-rostislav (approved):

vitek-rostislav:
Could you take a look at the Cypress tests as well, please? Just to be sure :innocent:

phpstan.neon
grossmannmartin: I've tried to fix this. 
The original fix in the newer version was here https://github.com/shopsys/shopsys/commit/f9a1d2ec012721b154a4d3e2993bd78fbc79c622#diff-7d97ca4459ef4a5a8f2666c1031daac7f83168828aeb2a843a1d40fed2325150
(and one more occurrence elsewhere, don't know where). 

But this was a change of application, and I wanted to avoid this as it's out of scope of the 14.5 version.

composer.json
grossmannmartin: added

composer.json
grossmannmartin: added

pk16011990 (approved):

PR #4317, opened 2025-12-08
improved withdrawal deadline calculation

vitek-rostislav:
Admin can now import public holidays and select whether a closed day is a public holiday. The setting is then used for order withdrawal calculation along with weekends. If the withdrawal deadline results in a public holiday or weekend, it is shifted to the next closest business day. 

#### Fixes issues
... 

----
  - https://rv-withdrawal-returns.odin.shopsys.cloud
  - https://cz.rv-withdrawal-returns.odin.shopsys.cloud
  - https://rv-withdrawal-returns.odin.shopsys.cloud/sk


vitek-rostislav:
@coderabbitai review

vitek-rostislav:
@coderabbitai review

packages/administration/templates/content/closedDay/list.html.twig
```diff
@@ -5,13 +5,21 @@
...
+{% endblock %}
+
 {% block main_content %}
     {{ render(controller('Shopsys\\FrameworkBundle\\Controller\\Admin\\DomainController::domainTabsAction')) }}
 
     {{ gridView.render() }}
 {% endblock %}
 
 {% block btn %}
+    <a href="{{ url('admin_closedday_holidaysimport') }}" class="btn btn-primary">
```
grossmannmartin: I wouldn't make both buttons primary. It's screaming too much, in my opinion. I would use btn-secondary.

packages/administration/templates/layout/layout_with_panel.html.twig
```diff
@@ -55,6 +55,9 @@
                             {% block h1 %}{{ 'Administration'|trans }}{% endblock %}
                             {% include '@ShopsysAdministration/partial/display_uuid.html.twig' %}
                         </h2>
+                        <div class="text-secondary">
```
grossmannmartin: what about `page-subtitle` instead?

We may omit empty div for example 
```
  {% set sub_title_content %}
      {% block sub_title %}{% endblock %}
  {% endset %}

  {% if sub_title_content|trim is not empty %}
      <div class="page-subtitle">{{ sub_title_content|raw }}</div>
  {% endif %}
  ```

but I'm not sure if I like the solution 😅 We should switch to latte 🫣

File: packages/administration/translations/messages.cs.po [hunk 3]
```diff
@@ -1903,6 +1909,9 @@ msgstr "Nastavení obchodních podmínek"
 msgid "The current list of URL addresses can be found in the project's repository in the \"storefront/config/staticRewritePaths.js\" file."
 msgstr "Aktuální seznam URL adres lze nalézt v repozitáři projektu v souboru \"storefront/config/staticRewritePaths.js\"."
 
+msgid "The days are used to display proper stores opening hours and to calculate order withdrawal deadline."
+msgstr "Dny se používají k zobrazení správné otevírací doby prodejny a k výpočtu lhůty pro odstoupení od smlouvy u objednávky."
```
grossmannmartin: `prodejen`? Even english text has plural

packages/framework/src/Form/Admin/Country/CountryFormType.php
```diff
@@ -75,7 +75,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                     ]),
                 ],
                 'label' => 'Code',
-                'help' => t('Country code in ISO 3166-1 alpha-2'),
+                'help_html' => true,
+                'help' => t('Country code in <a href="%wikiLink%" target="_blank">ISO 3166-1 alpha-2</a>', [
+                    '%wikiLink%' => 'https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2',
```
grossmannmartin, line 80: 💚

packages/framework/src/Form/Admin/Holidays/HolidaysImportFormType.php
```diff
@@ -0,0 +1,110 @@
...
+     * @param \Symfony\Component\Form\FormBuilderInterface $builder
+     * @param array $options
+     */
+    #[Override]
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('countryCode', ChoiceType::class, [
+                'label' => 'Country',
+                'choices' => $this->holidaysImportFacade->getValidCountryChoices(),
```
grossmannmartin: we could just use countryFacade->getAll() and choice filter https://symfony.com/doc/current/reference/forms/types/choice.html#choice-filter 

I'm not sure if it's better... it's a method less 🤷‍♂️

packages/framework/src/Form/Admin/Holidays/HolidaysImportFormType.php
```diff
@@ -0,0 +1,110 @@
...
+                'placeholder' => 'Select country',
+                'required' => true,
+                'help_html' => true,
+                'help' => t('If the country is not listed, you have to <a href="%countryCreateUrl%" target="_blank">create</a> it first. Or, in rare scenarios, your country code might not be <a href="%supportedCountriesUrl%" target="_blank">supported</a>.', [
+                    '%countryCreateUrl%' => $this->router->generate('admin_country_new'),
```
grossmannmartin, line 57: just a question - If I don't have a permission, the link will be properly generated, I just get 403 error, right?
I can live with that

packages/framework/src/Form/DomainsType.php
```diff
@@ -33,4 +34,15 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefaults([
+            'error_bubbling' => false,
```
grossmannmartin, line 45: why do we need this? Can it break something else?

packages/framework/src/Model/Order/Withdrawal/WithdrawalDeadlineCalculation.php
```diff
@@ -31,6 +34,10 @@ public function getWithdrawalDeadline(Order $order): ?DateTimeInterface
 
         $withdrawalDeadlineDays = $this->withdrawalSetting->getDeadlineDays($order->getDomainId());
 
-        return (clone $deliveredAt)->modify(sprintf('+%d days', $withdrawalDeadlineDays));
+        $deadline = $deliveredAt
+            ->modify(sprintf('+%d days', $withdrawalDeadlineDays))
+            ->setTime(0, 0);
```
grossmannmartin: Shouldn't it be 23:59? I assume the last day is also allowed to withdraw

packages/framework/src/Model/Store/ClosedDay/ClosedDayRepository.php
```diff
@@ -70,6 +71,35 @@ public function getFollowingWeekClosedDaysNotExcludedForStore(Store $store): arr
...
+            ->andWhere('cd.date >= :startDate')
+            ->andWhere('cd.date <= :endDate')
+            ->setParameter('domainId', $domainId)
+            ->setParameter('isPublicHoliday', true)
+            ->setParameter('startDate', $startDate)
+            ->setParameter('endDate', $endDate)
+            ->getQuery()
+            ->getResult();
+
+        return array_map(static fn (array $row) => $row['date'], $closedDays);
```
grossmannmartin: can't we use this?

```php
return array_column($closedDays, 'date');
```

packages/framework/src/Twig/RequiredSettingExtension.php
```diff
@@ -221,4 +231,48 @@ protected function checkAllSliderNumericValuesAreSet(): void
...
+    /**
+     * @param int $year
+     */
+    protected function checkPublicHolidaysForYearAreSet(int $year): void
+    {
+        $yearStart = new DateTimeImmutable($year . '-01-01');
+        $yearEnd = new DateTimeImmutable($year . '-12-31');
+
+        foreach ($this->domain->getAdminEnabledDomainIds() as $domainId) {
+            $holidays = $this->closedDayFacade->getPublicHolidays($domainId, $yearStart, $yearEnd);
```
grossmannmartin: We hydrate the entities for each request just to check if any record exists.  
I would rather just call `->hasPublicHolidays(...)` which just do EXISTS in the postgres.

Maybe even further optimize it, so it will make just one query for both years, but as it's a thing just a 30 days in a year, I don't insist on that

packages/php-image/Dockerfile
```diff
@@ -65,6 +65,7 @@ RUN apk add --no-cache --virtual .build-deps \
     docker-php-ext-configure gd --with-freetype --with-jpeg && \
     docker-php-ext-install \
         bcmath \
+        calendar \
```
grossmannmartin, line 68: just a reminder to run rebuild of php-image after merge, otherwise all builds will fail for the rest of the day

grossmannmartin (changes requested):

packages/administration/templates/content/closedDay/list.html.twig
vitek-rostislav: :+1: 93f3d0bd10

packages/administration/templates/layout/layout_with_panel.html.twig
vitek-rostislav: `page-subtitle` class works better, thanks. 8ed4a5464f

Regarding the empty div, I was concerned with that originally during the implementation; however, I think I am ok with that now (this is not the only empty div we have in the layout) as the solution seems too complicated :shrug:

vitek-rostislav: btw, how would latte help?

packages/administration/translations/messages.cs.po
vitek-rostislav: good catch :ok_hand: f48061418c

packages/framework/src/Form/Admin/Holidays/HolidaysImportFormType.php
vitek-rostislav: I think it is quite a neat solution, thanks :ok_hand: f6539aaa75

vitek-rostislav: However, I had to do a little bit more... c470d5f4a5

packages/framework/src/Form/Admin/Holidays/HolidaysImportFormType.php
vitek-rostislav: yes, I can also live with that :grin:

packages/framework/src/Form/DomainsType.php
vitek-rostislav: Without this setting, the validation error related to `DomainsType` is rendered at the top of its parent form. We can either set it with every usage of the type (this is the first occurrence of the validation of such a field in our codebase), or use it as a default value of the type to avoid future confusion or DRY violations. I am pretty sure (and I hope) it does not break anything :slightly_smiling_face: 

With error bubbling set to `false`:
image

Without error bubbling set to `false`:
image

packages/framework/src/Model/Order/Withdrawal/WithdrawalDeadlineCalculation.php
vitek-rostislav: yes, good catch, fixed in f0b84207ca

packages/framework/src/Twig/RequiredSettingExtension.php
vitek-rostislav: Good point, I have added `hasPublicHolidays` method for this scenario - 0853f35f4d. 

If I wanted to do just 1 query for 2 years, I would have to do more changes to somehow distinguish the year for the final message(s). I do not think it is crucial, so I am keeping the two queries :slightly_smiling_face:

packages/administration/templates/layout/layout_with_panel.html.twig
grossmannmartin: latte has n:macros, we could use `n:ifcontent` and if tag content is empty, it will not be rendered

packages/framework/src/Form/DomainsType.php
grossmannmartin: neat 💚

grossmannmartin (approved):

packages/framework/src/Model/Order/Withdrawal/WithdrawalChecker.php
```diff
@@ -46,9 +46,10 @@ public function checkOrderWithdrawal(Order $order): void
             return;
         }
 
-        $now = $this->clock->now();
+        $today = $this->clock->now()->format('Y-m-d');
+        $deadlineDate = $withdrawalDeadline->format('Y-m-d');
 
-        if ($now > $withdrawalDeadline) {
+        if ($today > $deadlineDate) {
```
grossmannmartin: I'm not sure about the string comparison here... I mean, it works, but is it clear enough that we compare lexicographically, not as dates? 

For example, changing the format sounds innocent, but in reality, it breaks comparison.
Why don't we set the same time instead and compare datetimes?

grossmannmartin (changes requested):

packages/framework/src/Model/Order/Withdrawal/WithdrawalChecker.php
vitek-rostislav: you are absolutely right! :slightly_smiling_face:  What about this? 84b3e39642

grossmannmartin (approved):

PR #4335, opened 2025-12-17
updated fixed bar on product detail

chlebektomas:
Added product info to the fixed bar on the product detail page in administration with responsive styling.

----
  - https://tc-ssp-3745-admin-product-detail.odin.shopsys.cloud
  - https://cz.tc-ssp-3745-admin-product-detail.odin.shopsys.cloud
  - https://tc-ssp-3745-admin-product-detail.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/src/Form/Admin/Type/ActionBarType.php
+++ b/packages/administration/src/Form/Admin/Type/ActionBarType.php
@@ -9,7 +9,10 @@ use Shopsys\FormTypesBundle\ActionBarType as BaseActionBarType;
 use Shopsys\FrameworkBundle\Component\Security\AccessControl\RouteAccessCheckerInterface;
 use Symfony\Component\Form\AbstractTypeExtension;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 use Symfony\Component\Routing\RouterInterface;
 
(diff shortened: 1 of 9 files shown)


packages/administration/templates/form/type/ActionBarType.html.twig
```diff
@@ -1,5 +1,11 @@
 {% block action_bar_row %}
-    {% embed '@ShopsysAdministration/partial/_fixed_bar.html.twig' %}
+    {% embed '@ShopsysAdministration/partial/_fixed_bar.html.twig' with {product_info: form.vars.product_info} %}
+        {% block fixed_bar_product_info %}
+            <div class="action-bar-product_info small">
+                <span class="fw-semibold text-truncate" style="max-width: 300px;">{{ product_info.name }}</span>
+                <span class="text-muted">|&nbsp;{{ product_info.catNum }}</span>
```
grossmannmartin: we use lowercased `catnum` everywhere in the codebase

```suggestion
                <span class="text-muted">|&nbsp;{{ product_info.catnum }}</span>
```

packages/administration/src/Form/Admin/Type/ActionBarType.php
```diff
@@ -36,6 +39,27 @@ public static function getExtendedTypes(): iterable
...
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setDefined(['product_info'])
```
grossmannmartin: I would generalize the feature slightly.  
What about:
- split product_info into two options - `entity_name` and `entity_identifier`
- if one of the options is null, just the specific option will not be rendered (for example, in orders I use only entity_identifier)
- if both are null, nothing will be rendered

packages/administration/templates/form/type/ActionBarType.html.twig
```diff
@@ -1,5 +1,11 @@
 {% block action_bar_row %}
-    {% embed '@ShopsysAdministration/partial/_fixed_bar.html.twig' %}
+    {% embed '@ShopsysAdministration/partial/_fixed_bar.html.twig' with {product_info: form.vars.product_info} %}
```
grossmannmartin: I'm not sure why this works without an explicit check for null value 😅
But I think it would be safer to do something like 

```twig
{% embed '@ShopsysAdministration/partial/_fixed_bar.html.twig' %}
    {% block fixed_bar_product_info %}
        {% if form.vars.entity_name or form.vars.entity_identifier %}
             ....
             {% if form.vars.entity_name %}
                 html with name
             {% endif %}

             {% if form.vars.entity_identifier %}
                 html with identifier
             {% endif %}
```

grossmannmartin (changes requested):

upgrade-notes/backend_20251217_084346.md
```diff
@@ -0,0 +1,3 @@
+#### Product info in fixed bar ([#4335](https://github.com/shopsys/shopsys/pull/4335))
```
grossmannmartin: I think we don't need upgrade notes at all

grossmannmartin (approved):

pk16011990 (approved):

PR #4339, opened 2025-12-18
changed heureka key in data fixtures

grossmannmartin:

The originally set key gave the impression that it was real, and that could have been confusing.
Now it's clear that the key is made up. 

_We want a key set, so the checkbox for heureka agreement shows._ 

#### Fixes issues
... 

----
  - https://mg-remove-key.odin.shopsys.cloud
  - https://cz.mg-remove-key.odin.shopsys.cloud
  - https://mg-remove-key.odin.shopsys.cloud/sk

### DIFF
--- a/project-base/app/src/DataFixtures/Demo/SettingValueDataFixture.php
+++ b/project-base/app/src/DataFixtures/Demo/SettingValueDataFixture.php
@@ -147,7 +147,7 @@ class SettingValueDataFixture extends AbstractReferenceFixture implements Depend
             if ($this->heurekaShopCertificationLocaleHelper->isDomainLocaleSupported($locale)) {
                 $this->setting->setForDomain(
                     HeurekaSetting::HEUREKA_API_KEY,
-                    '96411416349324269511946875061235',
+                    'JustMadeUpDemoKeyToEnableHeureka',
                     $domainId,
                 );
             }


vitek-rostislav:
I think (no guarantee here :smile: ) that the key was actually just some random sequence to enable the heureka behavior by default with demo data (IMHO you need to have the key filled in somehow, otherwise heureka is disabled). We should check that, at least :slightly_smiling_face:

vitek-rostislav:
> I think (no guarantee here 😄 ) that the key was actually just some random sequence to enable the heureka behavior by default with demo data (IMHO you need to have the key filled in somehow, otherwise heureka is disabled). We should check that, at least 🙂

I have checked that, and it indeed influences the appearance of the "heureka agreement" checkbox within the order form in the storefront (when the key is not filled, the checkbox is not displayed).
image

Maybe we could add some "thisisjustademokey" (it needs to have 32 characters) string to demo data to preserve the checkbox on the supported domains?

vitek-rostislav (changes requested):
lets talk about it :) also, there were wrong upgrade notes (duplicated from https://github.com/shopsys/shopsys/pull/4338)

vitek-rostislav (approved):

pk16011990 (approved):

PR #4370, opened 2026-01-05
ProductTest tweaks

vitek-rostislav:
- use DQL files instead of hardcoded queries in strings
- add test asserting all product fields are covered
- cover all the product fields
- refactor the test to iteratively assert the expected data instead of comparing two huge arrays at once

#### Fixes issues
... 

----
  - https://rv-product-files-test.odin.shopsys.cloud
  - https://cz.rv-product-files-test.odin.shopsys.cloud
  - https://rv-product-files-test.odin.shopsys.cloud/sk

----
  - https://rv-product-files-test.odin.shopsys.cloud
  - https://cz.rv-product-files-test.odin.shopsys.cloud
  - https://rv-product-files-test.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Model/ProductVideo/ProductVideoDataFactory.php
+++ b/packages/framework/src/Model/ProductVideo/ProductVideoDataFactory.php
@@ -23,6 +23,14 @@ class ProductVideoDataFactory
     }
 
     /**
+     * @return \Shopsys\FrameworkBundle\Model\ProductVideo\ProductVideoData
+     */
+    public function create(): ProductVideoData
+    {
+        return $this->createInstance();
+    }
+
+    /**
      * @param \Shopsys\FrameworkBundle\Model\ProductVideo\ProductVideo $productVideo
      * @return \Shopsys\FrameworkBundle\Model\ProductVideo\ProductVideoData
      */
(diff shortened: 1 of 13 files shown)


packages/framework/src/Model/ProductVideo/ProductVideoDataFactory.php
```diff
@@ -17,7 +17,7 @@ public function __construct(
     /**
      * @return \Shopsys\FrameworkBundle\Model\ProductVideo\ProductVideoData
      */
-    protected function createInstance(): ProductVideoData
+    public function createInstance(): ProductVideoData
```
grossmannmartin: I slightly struggle with the createInstance being public. 

I'm afraid that this could lead to undesired extension of this method. For example, if we add a field that should have a default value during creation, we may tend to add it to a createInstance method. 

It's longer, but I think the more "our-way" is 

```php
protected function createInstance(): ProductVideoData
{
    return new ProductVideoData();
}

public function create(): ProductVideoData
{
    return $this->createInstance();
}
```

project-base/app/src/DataFixtures/Demo/ProductDataFixture.php
```diff
@@ -74,6 +77,9 @@ public function load(ObjectManager $manager): void
 
         $parameterValues = [];
 
+        $productVideoData = $this->productVideoDataFactory->createInstance();
+        $productVideoData->videoToken = 'Wz9zttavXpo';
```
grossmannmartin, line 81: Great choice 💚

grossmannmartin (approved):

packages/framework/src/Model/ProductVideo/ProductVideoDataFactory.php
vitek-rostislav: When will I learn? I think this is at least the second time you have addressed such a thing in my pull request. I try to remember this for the next time :sweat_smile: fixed in d7b34b86fa

pk16011990 (approved):

PR #4377, opened 2026-01-07
Shopsys CLI: A new project bootstrapper

grossmannmartin:

This PR introduces Shopsys CLI, a standalone CLI tool for automating the initialization and configuration of new Shopsys Platform projects.
Shopsys CLI eliminates manual setup work by providing interactive domain configuration, automatic file generation, and multi-domain setup capabilities.

- Adds a new packages/forge package with a CLI application
- Implements a worker-based architecture for modular, extensible configuration tasks
- Supports both interactive prompts and YAML-based configuration
- Provides extensible configuration sections for domain and project-level settings

CLI Commands:

- shopsys init [project-name] - Clone project-base and run configuration
- shopsys configure <path> - Configure an existing project (interactive or YAML)
- shopsys workers - List all available configuration workers

----
  - https://mg-forge.odin.shopsys.cloud
  - https://cz.mg-forge.odin.shopsys.cloud
  - https://mg-forge.odin.shopsys.cloud/sk

----
  - https://mg-forge.odin.shopsys.cloud
  - https://cz.mg-forge.odin.shopsys.cloud
  - https://mg-forge.odin.shopsys.cloud/sk

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,7 @@
             "Shopsys\\AdministrationBundle\\": "packages/administration/src/",
             "Shopsys\\CodingStandards\\": "packages/coding-standards/src/",
             "Shopsys\\FormTypesBundle\\": "packages/form-types-bundle/src/",
+            "Shopsys\\Cli\\": "packages/cli/src/",
             "Shopsys\\FrontendApiBundle\\": "packages/frontend-api/src/",
             "Shopsys\\GoogleCloudBundle\\": "packages/google-cloud-bundle/src/",
             "Shopsys\\HttpSmokeTesting\\": "packages/http-smoke-testing/src/",
--- a/docs/installation/installation-guide.md
+++ b/docs/installation/installation-guide.md
@@ -1,8 +1,24 @@
 # Installation Guide
 
-This document will provide you with information about 2 ways of developing and running Shopsys Platform project and the services that it depends on.  
-The first option [using docker](#installation-using-docker) is **highly recommended** since it is the easiest and fastest way to start Shopsys Platform.
-In the case the operating system does not support docker, or you are not able to use Docker (e.g., due to the performance problems with Docker-sync), we prepared also second section with document about project [installation without docker](#installation-without-docker), however this way is slower and harder to configure and maintain because of different operating systems and their versions.
+This document will provide you with information about ways of developing and running Shopsys Platform project and the services that it depends on.
+
+## Quick Start with Shopsys CLI
+
+For new projects, we recommend using **Shopsys CLI** - a command-line tool that automates project initialization and configuration:
+
+> **Windows Users:** Run these commands inside WSL2 (Windows Subsystem for Linux).
+
+```bash
+# Download Shopsys CLI
+curl -L https://github.com/shopsys/cli/releases/latest/download/shopsys.phar -o shopsys
+chmod +x shopsys
+
+# Initialize a new project
+./shopsys init my-project
+```
+
+The CLI will guide you through domain configuration, locale settings, and more.
+See [Project Initialization with Shopsys CLI](./project-initialization-with-shopsys-cli.md) for detailed documentation.
 
 ## Installation using Docker
 
--- a/docs/installation/installation-guide.md
+++ b/docs/installation/installation-guide.md
@@ -4,12 +4,20 @@ This document will provide you with information about ways of developing and run
 
 ## Quick Start with Shopsys CLI
 
-For new projects, we recommend using **Shopsys CLI** - a command-line tool that automates project initialization and configuration:
+For new projects, the **Shopsys CLI** is automatically downloaded when you create a project using Composer:
 
 > **Windows Users:** Run these commands inside WSL2 (Windows Subsystem for Linux).
 
 ```bash
-# Download Shopsys CLI
+composer create-project shopsys/project-base --no-install --keep-vcs --ignore-platform-reqs
+```
+
+The CLI (`shopsys.phar`) is automatically downloaded and the configuration wizard launches, guiding you through domain configuration, locale settings, and more.
+
+Alternatively, you can download the CLI manually and initialize a project:
+
+```bash
+# Download Shopsys CLI manually
 curl -L https://github.com/shopsys/cli/releases/latest/download/shopsys.phar -o shopsys
 chmod +x shopsys
 
@@ -17,7 +25,6 @@ chmod +x shopsys
 ./shopsys init my-project
 ```
 
-The CLI will guide you through domain configuration, locale settings, and more.
 See [Project Initialization with Shopsys CLI](./project-initialization-with-shopsys-cli.md) for detailed documentation.
 
 ## Installation using Docker
--- a/docs/installation/project-initialization-with-shopsys-cli.md
+++ b/docs/installation/project-initialization-with-shopsys-cli.md
@@ -9,18 +9,24 @@ It simplifies multi-domain setup by providing interactive configuration prompts 
 
 > **Windows Users:** The Shopsys CLI must be run inside WSL2 (Windows Subsystem for Linux). Open your WSL2 terminal before running these commands.
 
-Download the PHAR file from the [shopsys/cli releases](https://github.com/shopsys/cli/releases):
+### Automatic Installation (Recommended)
+
+When you create a new Shopsys Platform project using `composer create-project`, the CLI is **automatically downloaded** and the configuration wizard is launched:
+
+```bash
+composer create-project shopsys/project-base --no-install --keep-vcs --ignore-platform-reqs
+```
+
+The `post-create-project-cmd` hook downloads `shopsys.phar` and runs `php shopsys.phar configure .` automatically.
+
+### Manual Installation
+
+If you need to download the CLI manually (e.g., to use with an existing project), download the PHAR file from the [shopsys/cli releases](https://github.com/shopsys/cli/releases):
 
 ```bash
 # Download the latest release
 curl -L https://github.com/shopsys/cli/releases/latest/download/shopsys.phar -o shopsys
 chmod +x shopsys
-```
-
-You can also install it globally:
-
-```bash
-sudo mv shopsys /usr/local/bin/shopsys
 ```
 
 ## Commands
--- a/docs/installation/project-initialization-with-shopsys-cli.md
+++ b/docs/installation/project-initialization-with-shopsys-cli.md
@@ -1,0 +1,92 @@
+# Project Initialization with Shopsys CLI
+
+Shopsys CLI (package name: `shopsys/cli`) is a command-line tool that automates the initialization and configuration of new Shopsys Platform projects.
+It simplifies multi-domain setup by providing interactive configuration prompts or YAML-based configuration files.
+
+> **Important:** The Shopsys CLI version is coupled with the Shopsys Platform version. Always use the CLI version that matches your target Platform version (e.g., use CLI 18.0 for Platform 18.0).
+
+## Installation
+
+> **Windows Users:** The Shopsys CLI must be run inside WSL2 (Windows Subsystem for Linux). Open your WSL2 terminal before running these commands.
+
+Download the PHAR file from the [shopsys/cli releases](https://github.com/shopsys/cli/releases):
+
+```bash
+# Download the latest release
+curl -L https://github.com/shopsys/cli/releases/latest/download/shopsys.phar -o shopsys
+chmod +x shopsys
+```
+
+You can also install it globally:
+
+```bash
+sudo mv shopsys /usr/local/bin/shopsys
+```
+
+## Commands
+
+### Initialize a New Project
+
+The `init` command clones the project-base repository and runs the configuration wizard:
+
+```bash
+# Interactive mode (recommended)
+./shopsys init my-project
+
+# With a specific version (use matching CLI version!)
+./shopsys init my-project --branch=18.0.0
+
+# Non-interactive mode with YAML configuration
+./shopsys init my-project --config=project-config.yaml
+
+# Use the latest stable release (default)
+./shopsys init my-project --branch=stable
+```
+
+**Options:**
+
+- `--branch` (`-b`): Git branch or tag to clone (default: `stable`)
+- `--config` (`-c`): Path to YAML configuration file for non-interactive mode
+
+### Configure an Existing Project
+
+The `configure` command applies configuration to an existing project-base:
+
+```bash
+# Interactive mode
+./shopsys configure /path/to/project
+
+# Non-interactive mode
+./shopsys configure /path/to/project --config=project-config.yaml
+```
+
+### List Available Workers
+
+View all configuration workers and their descriptions:
+
+```bash
+./shopsys workers
+```
+
+## Interactive Configuration
+
+When running in interactive mode, the CLI will guide you through configuring your domains and project settings step by step.
+After collecting all information, it displays a YAML summary and asks for confirmation before applying changes.
+
+## YAML Configuration
+
+For automated or repeatable setups, you can provide a YAML configuration file instead of using interactive mode.
+Run the interactive mode once to see the expected YAML structure, then save it for future use:
+
+```bash
+./shopsys init my-project --config=project-config.yaml
+```
+
+## Next Steps
+
+After running the CLI:
+
+1. **Review changes**: Check the generated configuration files
+2. **Commit to git**: Save your configuration
+3. **Install the project**: Run `./scripts/install.sh`
+4. **Start development**
--- a/packages/cli/src/Command/ConfigureCommand.php
+++ b/packages/cli/src/Command/ConfigureCommand.php
@@ -1,0 +1,203 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\Cli\Command;
+
+use Exception;
+use Override;
+use Shopsys\Cli\Exception\CancelledException;
+use Shopsys\Cli\Input\InteractiveInputCollector;
+use Shopsys\Cli\Input\YamlConfigLoader;
+use Shopsys\Cli\Worker\WorkerInterface;
+use Shopsys\Cli\Worker\WorkerResult;
+use Shopsys\Cli\Worker\WorkerRunner;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+#[AsCommand(
+    name: 'configure',
+    description: 'Configure an existing Shopsys Platform project',
+)]
+class ConfigureCommand extends Command
+{
+    private const string OPTION_CONFIG = 'config';
+    private const string OPTION_MODIFICATIONS = 'modifications';
+    private const string ARGUMENT_PATH = 'path';
+
+    /**
+     * @param \Shopsys\Cli\Worker\WorkerRunner $workerRunner
+     * @param \Shopsys\Cli\Input\InteractiveInputCollector $interactiveInputCollector
+     * @param \Shopsys\Cli\Input\YamlConfigLoader $yamlConfigLoader
+     */
+    public function __construct(
+        private readonly WorkerRunner $workerRunner,
+        private readonly InteractiveInputCollector $interactiveInputCollector,
+        private readonly YamlConfigLoader $yamlConfigLoader,
+    ) {
+        parent::__construct();
+    }
+
+    #[Override]
+    protected function configure(): void
+    {
+        $optionConfig = self::OPTION_CONFIG;
+        $optionModifications = self::OPTION_MODIFICATIONS;
+
+        $this
+            ->addArgument(
+                self::ARGUMENT_PATH,
+                InputArgument::REQUIRED,
+                'Path to the project directory',
+            )
+            ->addOption(
+                $optionConfig,
+                $optionConfig[0],
+                InputOption::VALUE_REQUIRED,
+                'Path to YAML configuration file (for non-interactive mode)',
+            )
+            ->addOption(
+                $optionModifications,
+                $optionModifications[0],
+                InputOption::VALUE_NONE,
+                'Display modified files',
+            )
+            ->setHelp(
+                <<<HELP
+The <info>%command.name%</info> command configures an existing Shopsys Platform project.
+
+<info>Interactive mode:</info>
+    <comment>%command.full_name% /path/to/project</comment>
+
+<info>Non-interactive mode (from YAML file):</info>
+    <comment>%command.full_name% --{$optionConfig}=project-config.yaml</comment>
+
+<info>Show modified files for each step:</info>
+    <comment>%command.full_name% --{$optionModifications}</comment>
+HELP,
+            );
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Input\InputInterface $input
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     * @return int
+     */
+    #[Override]
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $projectPath = realpath($input->getArgument(self::ARGUMENT_PATH));
+
+        if ($projectPath === false) {
+            $io->error(sprintf('Project path not found: %s', $input->getArgument(self::ARGUMENT_PATH)));
+
+            return Command::FAILURE;
+        }
+
+        try {
+            $configFile = $input->getOption(self::OPTION_CONFIG);
+
+            if ($configFile !== null) {
+                $config = $this->yamlConfigLoader->load($configFile, $projectPath);
+            } else {
+                $config = $this->interactiveInputCollector->collect($projectPath, $io);
+            }
+        } catch (CancelledException) {
+            $io->warning('Configuration cancelled by user.');
+            $io->block(
+                sprintf('Copy the configuration printed above into a YAML file, fix it, and rerun with the --%s option', self::OPTION_CONFIG),
+                'TIP',
+                'fg=bright-cyan',
+            );
+
+            return Command::SUCCESS;
+        } catch (Exception $e) {
+            $io->error($e->getMessage());
+
+            return Command::FAILURE;
+        }
+
+        $io->title('Applying Configuration');
+
+        $results = $this->workerRunner->run(
+            $config,
+            $projectPath,
+            function (WorkerInterface $worker, WorkerResult $result) use ($io, $input) {
+                if ($result->success) {
+                    $io->section(sprintf('<fg=green>✓</> %s', $result->message));
+
+                    if ($input->getOption(self::OPTION_MODIFICATIONS)) {
+                        if (count($result->filesModified) > 0) {
+                            $io->writeln('✏️ <fg=green>Modified files:</>');
+                            $io->listing($result->filesModified);
+                        }
+
+                        if (count($result->filesCreated) > 0) {
+                            $io->writeln('️➕ <fg=green>Created files:</>');
+                            $io->listing($result->filesCreated);
+                        }
+
+                        if (count($result->filesDeleted) > 0) {
+                            $io->writeln('️🗑️ <fg=green>Deleted files:</>');
+                            $io->listing($result->filesDeleted);
+                        }
+                    }
+                } else {
+                    $io->writeln(sprintf('<fg=red>✗</> %s: %s', $worker->getName(), $result->message));
+                }
+            },
+        );
+
+        $io->newLine();
+
+        if ($this->workerRunner->allSuccessful($results)) {
+            $io->success('Configuration complete!');
+            $this->displayHints($io, $results);
+            $this->displayNextSteps($io);
+
+            return Command::SUCCESS;
+        }
+
+        $io->warning('Configuration completed with some errors. Please review the output above.');
+
+        return Command::FAILURE;
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Style\SymfonyStyle $io
+     * @param array<\Shopsys\Cli\Worker\WorkerResult> $results
+     */
+    private function displayHints(SymfonyStyle $io, array $results): void
+    {
+        $hints = $this->workerRunner->collectHints($results);
+
+        if (count($hints) === 0) {
+            return;
+        }
+
+        $io->writeln('<info>Important notes:</info>');
+        $io->listing($hints);
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Style\SymfonyStyle $io
+     */
+    private function displayNextSteps(SymfonyStyle $io): void
+    {
+        $io->writeln('<info>Next steps:</info>');
+        $io->listing([
+            'Install project: <comment>./scripts/install.sh</comment>',
+            'Fix backend standards: <comment>docker compose exec php-fpm php phing standards-fix</comment>',
+            'Dump backend translations: <comment>docker compose exec php-fpm php phing translations-dump</comment>',
+            'Fix storefront standards: <comment>docker compose exec storefront pnpm run check--fix</comment>',
+            'Dump storefront translations: <comment>docker compose exec storefront pnpm run translate</comment>',
+            'Verify and commit changes to git',
+        ]);
+    }
+}
--- a/packages/cli/src/Command/InitCommand.php
+++ b/packages/cli/src/Command/InitCommand.php
@@ -1,0 +1,152 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\Cli\Command;
+
+use Override;
+use Shopsys\Cli\Exception\GitException;
+use Shopsys\Cli\Model\GitHandler;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+#[AsCommand(
+    name: 'init',
+    description: 'Initialize a new Shopsys Platform project',
+)]
+class InitCommand extends Command
+{
+    private const string REPOSITORY_URL = 'https://github.com/shopsys/project-base.git';
+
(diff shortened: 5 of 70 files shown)


grossmannmartin:
@coderabbitai review

composer.json
```diff
@@ -10,6 +10,7 @@
             "Shopsys\\AdministrationBundle\\": "packages/administration/src/",
             "Shopsys\\CodingStandards\\": "packages/coding-standards/src/",
             "Shopsys\\FormTypesBundle\\": "packages/form-types-bundle/src/",
+            "Shopsys\\Forge\\": "packages/forge/src/",
```
vitek-rostislav: should we add the forge to the packages lists? in monorepo_functions, etc....

packages/cli/src/Command/InitCommand.php
```diff
@@ -0,0 +1,148 @@
...
+
+<info>Interactive mode:</info>
+    <comment>%command.full_name% my-project</comment>
+
+<info>Non-interactive mode (from YAML file):</info>
+    <comment>%command.full_name% my-project --{$optionConfig}=project-config.yaml</comment>
+
+<info>Init a specific version:</info>
+    <comment>%command.full_name% my-project --{$optionBranch}=18.0.0</comment>
+HELP,
```
vitek-rostislav, line 82: by default, there should be a tag, right? in an ideal world :smile: I would consider put the tag into the help

packages/cli/src/Command/InitCommand.php
```diff
@@ -0,0 +1,148 @@
...
+            $io->error(sprintf('Directory already exists: %s', $projectName));
+
+            return Command::FAILURE;
+        }
+
+        $io->title('Shopsys Forge - Project Initialization');
+        $io->section('Cloning Shopsys Platform repository');
+
+        $io->writeln(sprintf('Cloning branch <info>%s</info>...', $branch));
+
```
vitek-rostislav, line 116: also here, branch or tag should be somehow taken into account IMHO, or is it a nitpick?

packages/cli/src/Command/ListWorkersCommand.php
```diff
@@ -0,0 +1,89 @@
...
+        } catch (Exception $e) {
+            $io->error('Failed to load workers: ' . $e->getMessage());
+
+            return Command::FAILURE;
+        }
+
+        $io->section('Backend Workers');
+        $io->table(
+            ['Worker', 'Description', 'Priority'],
+            $this->getWorkersInfo($workers, 'Backend'),
```
vitek-rostislav, line 54: I would make an enum for the worker types, or constants, at least :innocent:

vitek-rostislav: ...but now I see it is resolved by the namespace, so ok, I can live with that as is :)

packages/cli/src/Worker/WorkerResult.php
```diff
@@ -0,0 +1,58 @@
...
+            filesDeleted: $filesDeleted,
+        );
+    }
+
+    /**
+     * @param string $message
+     * @return self
+     */
+    public static function failure(string $message): self
+    {
```
vitek-rostislav, line 57: is it possible that worker fails but creates/edits/deletes some files before that? Is there some revert mechanism implemented?

packages/cli/src/Command/ConfigureCommand.php
```diff
@@ -0,0 +1,162 @@
...
+     * @param \Symfony\Component\Console\Input\InputInterface $input
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     * @return int
+     */
+    #[Override]
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $projectPath = realpath($input->getArgument('path'));
+
```
vitek-rostislav, line 96: constant for the argument name?

packages/cli/src/Command/ConfigureCommand.php
```diff
@@ -0,0 +1,162 @@
...
+        if ($projectPath === false) {
+            $io->error(sprintf('Project path not found: %s', $input->getArgument('path')));
+
+            return Command::FAILURE;
+        }
+
+        try {
+            $configFile = $input->getOption('config');
+
```
vitek-rostislav, line 105: constant for the option name?

packages/cli/src/Command/ConfigureCommand.php
```diff
@@ -0,0 +1,162 @@
...
+        $io->warning('Configuration completed with some errors. Please review the output above.');
+
+        return Command::FAILURE;
+    }
+
+    /**
+     * @param \Symfony\Component\Console\Style\SymfonyStyle $io
+     */
+    private function displayNextSteps(SymfonyStyle $io): void
```
vitek-rostislav, line 191: do not we want workers for these steps as well?

packages/cli/src/Config/ConfigSectionRegistry.php
```diff
@@ -0,0 +1,50 @@
...
+        private readonly ServiceLocator $domainConfigSections,
+        private readonly ServiceLocator $projectConfigSections,
+    ) {
+    }
+
+    /**
+     * @return \Shopsys\Forge\Config\DomainConfigSectionInterface[]
+     */
+    public function getDomainConfigSections(): array
+    {
```
vitek-rostislav, line 25: could be unified with `getProjectConfigSections` into one method taking `ServiceLocator` as a parameter to avoid DRY

vitek-rostislav: I mean, we can preserve the two public methods, just extract their common logic into a private method

packages/cli/src/Config/CoreDomainConfig.php
```diff
@@ -0,0 +1,92 @@
...
+     * @param bool $loadDemoData
+     */
+    public function __construct(
+        public readonly int $id,
+        public readonly string $name,
+        public readonly string $locale,
+        public readonly string $timezone,
+        public readonly string $type,
+        public readonly string $currencyCode,
+        public readonly bool $loadDemoData,
```
vitek-rostislav, line 32: we do not have currency code in domains yaml. But we do have it next config...but what about `mapSetting` then?

packages/cli/src/Config/CoreDomainConfigValidator.php
```diff
@@ -0,0 +1,88 @@
...
+use DateTimeZone;
+use Exception;
+use RuntimeException;
+
+final class CoreDomainConfigValidator
+{
+    /**
+     * @var string[]
+     */
+    public const array SUPPORTED_DOMAIN_TYPES = ['b2c', 'b2b'];
```
vitek-rostislav, line 16: there are constants for this defined in `DomainConfig` (I would appreciate enum, but :shrug:)

packages/cli/src/Worker/Backend/DomainConfigWorker.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $filePath = $projectPath . '/' . self::DOMAINS_URLS_FILE_PATH;
+
+        $domainsUrls = [];
+
+        foreach ($config->domains as $domain) {
+            $domainsUrls[] = [
+                'id' => $domain->id,
+                'url' => sprintf('http://127.0.0.%d:8000', $domain->id),
+            ];
```
vitek-rostislav, line 78: do not we want to have the url configurable?

packages/forge/src/Worker/AbstractWorker.php
```diff
@@ -0,0 +1,19 @@
...
+
+abstract class AbstractWorker implements WorkerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function getName(): string
+    {
+        return substr(static::class, strrpos(static::class, '\\') + 1);
```
vitek-rostislav: there is `ReflectionHelper::getShortClassName` but I understand we do not want a dependency on the package, right? What about the reflection itself? But I can live with it as is

packages/cli/src/Worker/Backend/RoutingFilesWorker.php
```diff
@@ -0,0 +1,118 @@
...
+                $fullFilePath,
+            );
+
+            $filesCreated[] = $filePath;
+        }
+
+        $filesDeleted = $this->deleteUnnecessaryRoutingFiles($config, $projectPath);
+
+        return WorkerResult::success(
+            'Routing files created',
```
vitek-rostislav, line 81: Should we remind the users not to forget adding the proper routes translations to the routing files?

packages/cli/src/Worker/Storefront/DomainsEnvWorker.php
```diff
@@ -0,0 +1,111 @@
...
+        );
+    }
+
+    /**
+     * @param int $domainId
+     * @return string
+     */
+    private function getPlaceholderUrl(int $domainId): string
+    {
+        return sprintf('http://127.0.0.%d:8000/', $domainId);
```
vitek-rostislav, line 109: here, if we added the URL to the configuration, we could dump it here :)

vitek-rostislav: but when I think about it, this should be always be the default config in git...Anyhow, what about possible /locale urls on domains? :smile: (if there was such an URL in the project-config, it would then require a few manual edits within the initial setup)

packages/cli/src/Config/Section/MapSettingsSection.php
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function collectInteractive(SymfonyStyle $io, int $domainId): void
+    {
+        $io->section(sprintf('Map Settings for Domain %d', $domainId));
+
```
vitek-rostislav, line 45: ah, here is the map config, ok :+1:

packages/cli/src/Worker/Storefront/DomainsPublicRuntimeConfigWorker.php
```diff
@@ -0,0 +1,187 @@
...
+        );
+    }
+
+    /**
+     * @param string $content
+     * @param \Shopsys\Forge\Config\CoreProjectConfig $config
+     * @return string
+     */
+    private function replaceRemotePatterns(string $content, CoreProjectConfig $config): string
+    {
```
vitek-rostislav, line 73: unrelated note - until we implement some automatic action to verify the forge, we should at least add an instruction to the releaser to verify the forge manually, what do you think?

packages/cli/src/Worker/Storefront/RoutesWorker.php
```diff
@@ -0,0 +1,149 @@
...
+        $filePath = $projectPath . '/' . self::FILE_PATH;
+
+        $existingContent = $this->fileHandler->readFile($filePath);
+        $this->parseExistingRoutes($existingContent);
+
+        $routesContent = $this->generateRoutesContent($config);
+        $this->fileHandler->writeFile($filePath, $routesContent);
+
+        return WorkerResult::success(
+            'Generated routes.ts',
```
vitek-rostislav, line 64: I have already mentioned this, however, in general, we could update the worker outputs with some additional instructions somehow, maybe. Users should know thay need to update the route translations properly after processing

packages/cli/src/Worker/Backend/resources/elastic-locales.yaml
```diff
@@ -0,0 +1,130 @@
+# Locale-specific Elasticsearch configuration
+# This file contains settings for stemming and stopwords per locale
```
vitek-rostislav, line 2: love it!

packages/forge/src/Worker/Backend/resources/elastic-locales.yaml
```diff
@@ -0,0 +1,130 @@
...
+        stemmer: portuguese
+        stopwords: _portuguese_
+
+    ro:
+        prefix: romanian
+        stemmer: romanian
+        stopwords: _romanian_
+
+    ru:
+        prefix: russian
```
vitek-rostislav: I know it is rather a political question, and a personal opinion, but I do not want that here

packages/cli/src/Config/CoreDomainConfig.php
```diff
@@ -0,0 +1,92 @@
...
+            'timezone' => $this->timezone,
+            'type' => $this->type,
+            'currency_code' => $this->currencyCode,
+            'load_demo_data' => $this->loadDemoData,
+        ];
+
+        foreach ($this->configSections as $section) {
+            $data[$section::getKey()] = $section->toArray();
+        }
+
```
vitek-rostislav, line 83: I have just noticed one thing, it is probably not needed, however, it is a little bit confusing, at least. The config can have an array of another configSections, but these are not taken into account within `fromArray` method at all, unlike the `toArray` method. Let's talk about it, please.

docs/installation/installation-guide.md
```diff
@@ -1,8 +1,22 @@
 # Installation Guide
 
-This document will provide you with information about 2 ways of developing and running Shopsys Platform project and the services that it depends on.  
-The first option [using docker](#installation-using-docker) is **highly recommended** since it is the easiest and fastest way to start Shopsys Platform.
-In the case the operating system does not support docker, or you are not able to use Docker (e.g., due to the performance problems with Docker-sync), we prepared also second section with document about project [installation without docker](#installation-without-docker), however this way is slower and harder to configure and maintain because of different operating systems and their versions.
+This document will provide you with information about ways of developing and running Shopsys Platform project and the services that it depends on.
+
+## Quick Start with Forge
+
```
vitek-rostislav, line 6: IMHO it is worth adding links here (or directly to the new forge page) to multiple places in the docs (e.g. particular platforms installation guides, how-to-setup-domains-and-locales, etc.). If anyone is reading such a page (that he might be redirected to from a web search, lets say), he deserves to know, "hey, you do not need to do any of that manually if you use this awesome tool" :)

docs/installation/project-initialization-with-forge.md
```diff
@@ -0,0 +1,88 @@
...
+```bash
+# Download the latest release
+curl -L https://github.com/shopsys/forge/releases/latest/download/forge.phar -o forge
+chmod +x forge
+```
+
+You can also install it globally:
+
+```bash
+sudo mv forge /usr/local/bin/forge
```
vitek-rostislav: What about the poor Windows users? :smile:

docs/installation/project-initialization-with-shopsys-cli.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `init` command clones the project-base repository and runs the configuration wizard:
+
+```bash
+# Interactive mode (recommended)
+forge init my-project
+
+# With a specific version
+forge init my-project --branch=18.0.0
+
```
vitek-rostislav, line 44: as mentioned earlier in the review, I would somehow stress out, that a tag should be used, or?

docs/installation/project-initialization-with-forge.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+## Next Steps
+
+After running Forge:
+
+1. **Review changes**: Check the generated configuration files
+2. **Commit to git**: Save your configuration
+3. **Install the project**: Run `./scripts/install.sh`
+4. **Start development**: Follow the [Installation Guide](./installation-guide.md)
```
vitek-rostislav: is the installation guide linked here by purpose? As it does not make sense to me :thinking:

vitek-rostislav (changes requested):
Awesome job there, hats off! I believe the developers will be thrilled with this :rocket:

docs/installation/project-initialization-with-forge.md
vitek-rostislav: it should have been `installation-using-docker-application-setup.md` IMHO

composer.json
grossmannmartin: You are absolutely right! 

Added to the monorepo_functions.sh (for split) and CHANGELOG (for changelog 😆) in 16d7538181

Other places are not required imho as Forge is not a dependency, nor does it have dependencies on other Shopsys Platform packages.

packages/cli/src/Command/InitCommand.php
grossmannmartin: What about this? cf76fc72f4

packages/cli/src/Command/InitCommand.php
grossmannmartin: taken into account in dde7c33e34 :)

packages/cli/src/Command/ListWorkersCommand.php
grossmannmartin: me too 😆

packages/cli/src/Worker/WorkerResult.php
grossmannmartin: No revert mechanism is implemented. The code is under version control in Git, so the user immediately sees the changes, and it's possible that if any worker fails, the so-far made changes make sense, and it would be beneficial to keep them.

packages/cli/src/Command/ConfigureCommand.php
grossmannmartin: true - 54cf81585f

packages/cli/src/Command/ConfigureCommand.php
grossmannmartin: true - 7a23410090

packages/cli/src/Command/ConfigureCommand.php
grossmannmartin: I'm not sure. Personally, I would rather do these steps by myself, so I would be in control if any steps fail.
But maybe I'm a bit old-school 😁
Let's discuss it, please.

packages/cli/src/Config/ConfigSectionRegistry.php
grossmannmartin: good idea - 9e0fcd37b6

packages/cli/src/Config/CoreDomainConfig.php
grossmannmartin: I consider a currencyCode as an Core definition (even though we don't have it on backend in domains). 
I don't think the CoreDomainConfig have to necessarily mirror the backend domain class/config.

On the other hand, MapSetting is (at least in my head 😁) more of an extension of configuration, so it's put extra

packages/cli/src/Config/CoreDomainConfigValidator.php
grossmannmartin: Yes, constants are in DomainConfig, but Forge is independent of the other packages, so I cannot use them. 

I can extract this to the enum, if you want. But TBH I see a little added value in that here.

packages/cli/src/Worker/Backend/DomainConfigWorker.php
grossmannmartin: I'm not sure... I think 98% of project use the default setup (http://127.0.0.1:8000, http://127.0.0.2:8000, ...)

Maybe for the path fragments in URL, but I'm not sure if the benefit balances the increased complexity of setup (I mean from the user perspective)

packages/cli/src/Worker/Backend/RoutingFilesWorker.php
grossmannmartin: I hope for the GI of our users (general intelligence) 😅

I, myself, don't read most of such recommendations when presented (even in the Releaser, I find much information unnecessary 🫣)

packages/cli/src/Worker/Storefront/DomainsEnvWorker.php
grossmannmartin: yes. 

What was the question? 😁

As in some of my previous comments:

> I'm not sure... I think 98% of project use the default setup (http://127.0.0.1:8000/, http://127.0.0.2:8000/, ...)

> Maybe for the path fragments in URL, but I'm not sure if the benefit balances the increased complexity of setup (I mean from the user perspective)

packages/forge/src/Worker/AbstractWorker.php
grossmannmartin: It's nicer with reflection 👍 - 7c9ef79fe2

packages/cli/src/Worker/Storefront/RoutesWorker.php
grossmannmartin: Personally, I consider the changed files in git as good enough guidance. 

Let's discuss it, I can add some "guidance info" to workers and print them quite easily.

packages/cli/src/Worker/Storefront/DomainsPublicRuntimeConfigWorker.php
grossmannmartin: added note to releaser b43176677d

packages/forge/src/Worker/Backend/resources/elastic-locales.yaml
grossmannmartin: 908c169a36 #StandWithUkraine

packages/cli/src/Config/CoreDomainConfig.php
grossmannmartin: I'm not sure, what do you mean... fromArray() for each section is located in each section class and loaded in yamlConfig loader along with core. Is that it?

docs/installation/project-initialization-with-forge.md
grossmannmartin: They have chosen their fate 😁

But really. I have absolutely no idea how this works in WSL2. I assume it's possible to do the same as it has a debian linux underneath.

docs/installation/installation-guide.md
grossmannmartin: added in e3755f65

docs/installation/project-initialization-with-shopsys-cli.md
grossmannmartin: Here, I think, the comment summarizes it well. `version`, I believe, is descriptive enough (I don't inherently need to know that it means git tag)

docs/installation/project-initialization-with-forge.md
grossmannmartin: 🤔  You're right, but even `installation-using-docker-application-setup.md` does not fit here... We already have the application installed and configured at this point. 

I have removed the link altogether - 73d9489eed - we may discuss it further.

composer.json
vitek-rostislav: and will it be released separately? So it should not be included within the releaser packages? `AbstractShopsysReleaseWorker::EXCLUDED_PACKAGES`

Let's talk about it, please

packages/cli/src/Config/CoreDomainConfigValidator.php
vitek-rostislav: yes, I did not realize the forge's independence in the first place

packages/cli/src/Worker/Backend/DomainConfigWorker.php
vitek-rostislav: lets talk about it (do not be afraid, I believe I can live with the current solution, I would just like to brainstorm the pros and cons :slightly_smiling_face: )

packages/cli/src/Worker/Backend/RoutingFilesWorker.php
vitek-rostislav: ok :slightly_frowning_face:

packages/cli/src/Config/CoreDomainConfig.php
vitek-rostislav: I am not sure either now :smile: No, my concern is about the usage - one might expect he can call `fromArray` with an array that could hypothetically contain the data for the other sections, but that would not be taken into account in the resulting object. Instead, the approach used in `YamlConfigLoader` must be used. I do not think it is a big deal, but as mentioned above, it is at least inconsistent with the reverse `toArray` function. Let's talk about it briefly and then close this thread :sweat_smile:

docs/installation/project-initialization-with-forge.md
vitek-rostislav: >They have chosen their fate 😁

Yes, I do agree, however

> But really. I have absolutely no idea how this works in WSL2. I assume it's possible to do the same as it has a debian linux underneath.

The problem is, IMHO, this looks like (and is probably mentioned) it should be run within the terminal in your system directly, right? It will not work on Windows (as well as the `chmod` command above).

packages/cli/src/Worker/Backend/DomainConfigWorker.php
henzigo: From my perspective, I understand both of you. From my point of view, if the project uses path fragments, you want to mirror that setting also on localhost. 

But I don't need to have domains URLs configurable. I'm absolutely ok if the script shows a message where you can change default URLs manually. So complexity will not change, and devs will get the info they need

packages/cli/src/Worker/Backend/RoutingFilesWorker.php
henzigo: Sorry for opening this comment, but when all localized files are generated, couldn't we show a list of files that must be translated as some message at the end of this script? 

I can imagine that you will simply forget to translate one of them, as there are multiple files with different purposes (translations, routes FE, routes BE, ...).

> I, myself, don't read most of such recommendations when presented (even in the Releaser, I find much information unnecessary 🫣)

Imagine that I will be releaser of the next version of Shopsys platform. In that case, I will read all messages from Releaser twice before i will move to the next step because it will be my first time. Let's make it simple for everyone and don't handicap newbies.

docs/installation/installation-guide.md
```diff
@@ -1,8 +1,22 @@
...
+For new projects, we recommend using **Shopsys Forge** - a CLI tool that automates project initialization and configuration:
+
+```bash
+# Download Forge
+curl -L https://github.com/shopsys/forge/releases/latest/download/forge.phar -o forge
+chmod +x forge
+
+# Initialize a new project
+./forge init my-project
```
henzigo: Hey, just an idea. `Forge` is a cool name for this, but what about renaming it just to `shopsys` CLI tool? (`shopsys init my-project`)

It would be easier to remember it for future usage and automatically increases brand awareness among programmers who will use this script. 

At the same time, we have an open path to add more features. For example, something like `shopsys start`, `shopsys stop` and/or some specific features.

What do you think?

docs/installation/installation-guide.md
```diff
@@ -1,8 +1,22 @@
...
-In the case the operating system does not support docker, or you are not able to use Docker (e.g., due to the performance problems with Docker-sync), we prepared also second section with document about project [installation without docker](#installation-without-docker), however this way is slower and harder to configure and maintain because of different operating systems and their versions.
+This document will provide you with information about ways of developing and running Shopsys Platform project and the services that it depends on.
+
+## Quick Start with Forge
+
+For new projects, we recommend using **Shopsys Forge** - a CLI tool that automates project initialization and configuration:
+
+```bash
+# Download Forge
+curl -L https://github.com/shopsys/forge/releases/latest/download/forge.phar -o forge
```
henzigo: We are recommending to download latest version, but the script has an option to specify version `forge init my-project --branch=18.0.0`.

This can create a bunch of errors, as files that are accessed/modified by this script can be changed between versions. Workers should be defined per version, or we should be able to configure which workers should run under a specified version to prevent any problems.

For example, look at `DomainConfigWorker`. In case we rename the  `domains_urls.yaml.dist` file, we will break this script for older versions.   

I like the idea of having a single Forge `app` that is independent of any Shopsys version, but to do that, we need to think a little differently.

packages/cli/src/Worker/Backend/DomainConfigWorker.php
grossmannmartin: I'll add some hints about necessary next steps

packages/cli/src/Worker/Backend/RoutingFilesWorker.php
grossmannmartin: I'll add some hints.

docs/installation/project-initialization-with-forge.md
grossmannmartin: I'll add note to remember running it in WSL

docs/installation/installation-guide.md
grossmannmartin: I can rename the binary and the phar file. The package name I don't want to rename as shopsys/shopsys already exists

henzigo: Exactly, my suggestion was to rename phar, not the repo itself.

packages/cli/src/Config/CoreDomainConfig.php
```diff
@@ -35,11 +35,12 @@ public function __construct(
 
     /**
      * @param array<string, mixed> $data
+     * @param \Shopsys\Forge\Config\ConfigSectionRegistry|null $registry
      * @return self
      */
-    public static function fromArray(array $data): self
+    public static function fromArray(array $data, ?ConfigSectionRegistry $registry = null): self
```
vitek-rostislav, line 41: :green_heart:

packages/forge/src/Worker/Storefront/RoutesWorker.php
```diff
@@ -60,9 +60,14 @@ public function execute(CoreProjectConfig $config, string $projectPath): WorkerR
         $routesContent = $this->generateRoutesContent($config);
         $this->fileHandler->writeFile($filePath, $routesContent);
 
+        $locales = implode(', ', $config->getUniqueLocales());
```
vitek-rostislav: this one is unused :slightly_smiling_face:

vitek-rostislav (changes requested):
Hi, as we discussed, there are two more things to implement:

- remove that global installation hint, and add the cli configuration run after the `composer create-project`
- rename even the package name from forge to CLI for unity sake

docs/installation/installation-guide.md
grossmannmartin: Now, the configuration is run automatically after composer create-project. The script will attempt to download the shopsys cli (formerly forge) at the requested version of the project base. 
I believe this is a nice compromise between complexity and usability.

henzigo: Oukej

vitek-rostislav:
# Testing and additional thoughts
- [x] Please check the "forge -> cli" renaming as there are some "clit"s in the codebase now :smile: 
- [x] The introduction table is a bit broken (probably because of the renaming as well), see 
image

- [x] During the testing of the "composer  creat-project" way, I started writing this summary, and got a timeout. I am not sure whether we can/want to address this, just reporting :innocent: 
image

- [x]  one of the acceptance criteria says that we can get rid of unnecessary configuration right away - what about that `frontend_api.domains` setting? Don't we want to remove it? Or do you not want to bother with that within this pull request? On one hand, I am ok with doing that in a separate task, on the other hand, that would mean removing `FrontendApiDomainsWorker` that you have just introduced, so I am thinking, we could "not-introduce" that worker at all if we got rid of the setting here :innocent: 
- [x] There are some unnecessary changes in the YAML files' formatting after running the CLI configuration (e.g., removed quotes, reformatted arrays, etc.). YAML standards do not mind the changed formatting, so I do not think it is a problem. Thanks for reading this note :smile: 
image

- [x] After configuring a single-domain "cs" application, I expected the other locales' translation files (messages, validators, etc.) to be removed, but they are still there even after running `translations-dump`. It would be nice to remove them, what do you think? (this applies to both BE and SF translations)
- [x] what am I doing wrong? // EDIT - ah, now i know, I need to run it as `./shopsys` - should we mention that in `project-initialization-with-shopsys-cli.md` for amateurs like me? :smile: 
image

- [x] nice to have - when user cancels the configuration, maybe we could hint him to save the presented configuration into yaml file, fix the configuration, and run the CLI again with `--config` argument? So that users know they do not need to go through the whole configuration interactively again when they find out there was a typo or so
image

- [x] I tried creating a typo within the currency name in the configuration. I got a weird error (in the code, I see there should be the exception message in the output, but I see just the currency code :confused: 
image

- [x] Btw, this is what I see when trying to use a new existing currency (`HUF`) in the configuration // EDIT - ah, it will be because of the fact I cloned the project and then ran the shopsys CLI configure before installing the project dependencies, right? But would not this be the common approach? (even with the "create-project" way?) // EDIT2 - no, it is not dependent on the project dependencies, the vendor folder is supposed to be baked within the phar, right?
image

- [x] maybe it is not a good idea to use `composer require shopsys/cli`, see https://github.com/shopsys/cli

grossmannmartin:
@vitek-rostislav 

> During the testing of the "composer creat-project" way, I started writing this summary, and got a timeout. I am not sure whether we can/want to address this, just reporting 😇

I have increased the timout to 20 minutes

> There are some unnecessary changes in the YAML files' formatting after running the CLI configuration (e.g., removed quotes, reformatted arrays, etc.). YAML standards do not mind the changed formatting, so I do not think it is a problem. Thanks for reading this note 😄

Sadly, I don't have many options for changing the output, and I didn't find any better library. We could adapt the original files, so the changes will be minimal, but I'm not sure if it's worth it.

> one of the acceptance criteria says that we can get rid of unnecessary configuration right away - what about that frontend_api.domains setting? Don't we want to remove it? Or do you not want to bother with that within this pull request? On one hand, I am ok with doing that in a separate task, on the other hand, that would mean removing FrontendApiDomainsWorker that you have just introduced, so I am thinking, we could "not-introduce" that worker at all if we got rid of the setting here 😇

separated to #4423, deleted worker here.

> After configuring a single-domain "cs" application, I expected the other locales' translation files (messages, validators, etc.) to be removed, but they are still there even after running translations-dump. It would be nice to remove them, what do you think? (this applies to both BE and SF translations)

I'm not sure. When a project decides to add a new domain with the locale that was previously deleted, it has to restore the previously deleted file or fill in the translations by hand from scratch. I am also considering the future plan to have many more locales by default in the platform.
Let's discuss it further, please.

vitek-rostislav:
> I'm not sure. When a project decides to add a new domain with the locale that was previously deleted, it has to restore the previously deleted file or fill in the translations by hand from scratch. I am also considering the future plan to have many more locales by default in the platform.
Let's discuss it further, please.

Well, the unused route translations are actually deleted during the project configuration at the moment, so it is not consistent right now, if I am ot mistaken. Moreover, the translations would still be dumped into the `po` files during the project development, which would be quite annoying for the project developers, IMHO. At least, I would maybe consider excluding the unused locales from the `translation-dump` phing target, but that would mean extending the target in the project's `build.xml`, right?

Currently, I still consider removing the unnecessary `po` files as the lesser evil. However, I really like the plans for having the platform ready for more locales by default. 

One way or another, I do not want to postpone the task with the discussions, as we will probably not meet until Friday, so we can address it in a separate task :wink: 

// Edit, Ah, no, the locales are loaded from domains config, so no unnecessary dumps, ok. That makes me less bothered with the current state :slightly_smiling_face:

vitek-rostislav (approved):

pk16011990 (approved):

PR #4389, opened 2026-01-14
filtered seo category for limited user

chlebektomas:
A limited user can no longer see SEO categories that contain a price filter. This prevents sorting products by price.

----
  - https://tc-ssp-2739-seo-category-for-limited-user.odin.shopsys.cloud
  - https://cz.tc-ssp-2739-seo-category-for-limited-user.odin.shopsys.cloud
  - https://tc-ssp-2739-seo-category-for-limited-user.odin.shopsys.cloud/sk

----
  - https://tc-ssp-2739-seo-category-for-limited-user.odin.shopsys.cloud
  - https://cz.tc-ssp-2739-seo-category-for-limited-user.odin.shopsys.cloud
  - https://tc-ssp-2739-seo-category-for-limited-user.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Model/CategorySeo/ReadyCategorySeoMix.php
+++ b/packages/framework/src/Model/CategorySeo/ReadyCategorySeoMix.php
@@ -9,6 +9,7 @@ use Doctrine\ORM\Mapping as ORM;
 use Ramsey\Uuid\Uuid;
 use Shopsys\FrameworkBundle\Model\Category\Category;
 use Shopsys\FrameworkBundle\Model\Product\Flag\Flag;
+use Shopsys\FrameworkBundle\Model\Product\Listing\ProductListOrderingConfig;
 
 #[ORM\Table(name: 'ready_category_seo_mixes')]
 #[ORM\UniqueConstraint(name: 'selected_category_seo_mix_combination_json', columns: ['selected_category_seo_mix_combination_json'])]
@@ -189,6 +190,17 @@ class ReadyCategorySeoMix
     public function setOrdering($ordering)
     {
         $this->ordering = $ordering;
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasPriceBasedOrdering(): bool
+    {
+        return in_array($this->ordering, [
+            ProductListOrderingConfig::ORDER_BY_PRICE_ASC,
+            ProductListOrderingConfig::ORDER_BY_PRICE_DESC,
+        ], true);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


packages/frontend-api/src/Model/Category/ReadyCategorySeoMixBatchLoader.php
```diff
@@ -47,10 +57,22 @@ public function loadByCategoryIds(array $categoryIds): Promise
...
         }
 
         return $this->promiseAdapter->all($result);
     }
+
+    /**
+     * @param string|null $ordering
+     * @return bool
+     */
+    protected function isPriceBasedOrdering(?string $ordering): bool
```
vitek-rostislav: ordering will never be null so you can pu just `string $iórdering` here

vitek-rostislav: I would make this method public and pu it into `ReadyCategorySeoMix` entity. (it will not need the parameter anymore then :)) Then, we need to modify `ReadyCategorySeoMixQuery` and ensure user gets access denied (or 404) when he can not see prices (to disallow direct URL access to such a category mix). Then, it would be nice to add tests for both these features (it will be a new test extending `GraphQlB2bDomainWithLoginTestCase`). Let me know if you need help with that :wink:

upgrade-notes/backend_20260114_085419.md
```diff
@@ -0,0 +1,3 @@
+#### Seo category for limited user ([#4389](https://github.com/shopsys/shopsys/pull/4389))
+
+- a limited user can no longer see SEO categories that contain a price filter
```
vitek-rostislav, line 3: after adding the tests, there should be see project base diff here :)

vitek-rostislav (changes requested):

packages/framework/src/Model/CategorySeo/ReadyCategorySeoMix.php
```diff
@@ -191,6 +192,17 @@ public function setOrdering($ordering)
         $this->ordering = $ordering;
     }
 
+    /**
+     * @return bool
+     */
+    public function hasPriceBasedOrdering()
```
vitek-rostislav: here can be a typehint :slightly_smiling_face:

upgrade-notes/backend_20260114_085419.md
vitek-rostislav: this is not addressed, yet

.claude/skills/backend-php/SKILL.md
```diff
@@ -0,0 +1,89 @@
+# Backend PHP Development Patterns
```
vitek-rostislav: I am not sure about this skill, it seems too specific but within the `.claude` folder, I am never sure what is correct :smile: so probably, we will see

vitek-rostislav (approved):

pk16011990 (approved):

PR #4392, opened 2026-01-15
Fix email template images for outlook

### LINKED ISSUE #4256: Outlook displays emails incorrectly due to images in WebP format
Some emails are rendered incorrectly in Outlook because our CDN serves images in WebP format. Outlook does not support WebP, which causes images to break or not display at all.

### Is your feature request related to a problem?
Images should be displayed correctly in all major email clients, including Outlook.

### Additional context
For Outlook, it is possible to use  and return an image src that is not converted, while leaving the others as is. This should be applied only for emails.

(Somehow) Solved in the project GOSSP-824

JanMolcik:
... 

#### Fixes issues
Closes [#4256](https://github.com/shopsys/shopsys/issues/4256) 

----
  - https://jm-ssp-3803-fix-outlook-webp-images.odin.shopsys.cloud
  - https://cz.jm-ssp-3803-fix-outlook-webp-images.odin.shopsys.cloud
  - https://jm-ssp-3803-fix-outlook-webp-images.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Resources/views/Common/imageEmail.html.twig
+++ b/packages/framework/src/Resources/views/Common/imageEmail.html.twig
@@ -1,0 +1,29 @@
+{% block image_html %}
+    <picture>
+        {% apply spaceless %}
+            {% set mergedClass = (attr.class|default('') ~ ' ' ~ imageCssClass)|trim %}
+            <!--[if !mso]><!-->
+            {% set attr = attr|merge({'class': mergedClass}) %}
+            <img
+            {%- for attrname, attrvalue in attr %}
+                {%- if attrname == 'src' -%}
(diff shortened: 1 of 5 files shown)


grossmannmartin:
VSH CDN  (at which this PR focuses based on `vshcdn-webp-noautoconvert` param) serves webp only if accepted by the client (the `Accept` header). 

Outlook may block the autoloading of images, but it has nothing to do with the format
image001

Images in the url still have the jpg (png) extension, so the only way for outlook to determine the content type is webp is to download the image (and send Accept header with image/webp).

One possible solution to avoid this is to inline images, but this has different drawbacks (mail size, possible higher spam score).

closing

PR #4416, opened 2026-01-26
fixed add newly uploaded image in elfinder behavior

grossmannmartin:

There were two main issues:

## 1.
On CI, where php-fpm and webserver share the content directory via volume (for example, `docker/conf/docker-compose.github-actions.review.simplified.yml.dist`).
If the admin created a folder in elfinder, the folder was created with strict permissions (0700). Due to a different uid in containers, webserver didn't had a permission to enter and read the content of the newly created folder -> the uploaded file didn't show up. 
Fixed by loosening permissions on the created folder.

## 2. 
Sometimes, when the image was added immediately after upload, it was added with a URL with an incorrect double slash, which causes problems, especially in combination with the CDN. 
This was caused by a bug in the latest version of the elfinder library - https://github.com/Studio-42/elFinder/issues/3725
Fixed by patching the volume driver method

#### Fixes issues
... 

----
  - https://mg-wysiwyg-folder.odin.shopsys.cloud
  - https://cz.mg-wysiwyg-folder.odin.shopsys.cloud
  - https://mg-wysiwyg-folder.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/Filesystem/Flysystem/VolumeDriver.php
+++ b/packages/framework/src/Component/Filesystem/Flysystem/VolumeDriver.php
@@ -196,6 +196,116 @@ class VolumeDriver extends Driver
 
         return $stat;
     }
+
+      /**
+     * Return content URL (for netmout volume driver)
+     * If file.url == 1 requests from JavaScript client with XHR
+     *
+     * @param string $hash    file hash
+     * @param array  $options options array
+     *
+     * @return boolean|string
+     * @author Naoki Sawada
+     */
+    #[Override]
+    public function getContentUrl($hash, $options = array())
+    {
+        if (($file = $this->file($hash)) === false) {
+            return false;
+        }
+        if (!empty($options['onetime']) && $this->options['onetimeUrl']) {
+            if (is_callable($this->options['onetimeUrl'])) {
+                return call_user_func_array($this->options['onetimeUrl'], array($file, $options, $this));
+            } else {
+                $ret = false;
+                if ($tmpdir = elFinder::getStaticVar('commonTempPath')) {
+                    if ($source = $this->open($hash)) {
+                        if ($_dat = tempnam($tmpdir, 'ELF')) {
+                            $token = md5($_dat . session_id());
+                            $dat = $tmpdir . DIRECTORY_SEPARATOR . 'ELF' . $token;
+                            if (rename($_dat, $dat)) {
+                                $info = stream_get_meta_data($source);
+                                if (!empty($info['uri'])) {
+                                    $tmp = $info['uri'];
+                                } else {
+                                    $tmp = tempnam($tmpdir, 'ELF');
+                                    if ($dest = fopen($tmp, 'wb')) {
+                                        if (!stream_copy_to_stream($source, $dest)) {
+                                            $tmp = false;
+                                        }
+                                        fclose($dest);
+                                    }
+                                }
+                                $this->close($source, $hash);
+                                if ($tmp) {
+                                    $info = array(
+                                        'file' => base64_encode($tmp),
+                                        'name' => $file['name'],
+                                        'mime' => $file['mime'],
+                                        'ts' => $file['ts']
+                                    );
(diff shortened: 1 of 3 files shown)


packages/framework/src/Component/Filesystem/Flysystem/VolumeDriver.php
```diff
@@ -196,6 +196,115 @@ protected function _stat($path, $hash = '')
...
+      /**
+     * Return content URL (for netmout volume driver)
+     * If file.url == 1 requests from JavaScript client with XHR
+     *
+     * @param string $hash    file hash
+     * @param array  $options options array
+     *
+     * @return boolean|string
+     * @author Naoki Sawada
+     */
```
vitek-rostislav, line 209: #[Override] is missing here

vitek-rostislav: I was wondering why this was not reported by `ecs` but I see that the whole file is skipped so that would be the reason probably :smile:

vitek-rostislav: Could we somehow emphasise why/where this method is overridden? At least I would put some comment inside the method, something like "/** here starts the custom code... ...end of custom code **/

vitek-rostislav: btw, I got a felling we do not have to copy-paste the whole parent/grandparent logic, so I asked claude and maybe it could be simplified. What do you think?

```
Yes, you can simplify significantly by calling parent::getContentUrl()   
  and only keeping your custom URL building logic:                         
                                                                           
  #[Override]                                                              
  public function getContentUrl($hash, $options = array())                 
  {                                                                        
      if (($file = $this->file($hash)) === false) {                        
          return false;                                                    
      }                                                                    
                                                                           
      // Handle onetime URLs via parent chain                              
      if (!empty($options['onetime']) && $this->options['onetimeUrl']) {   
          return parent::getContentUrl($hash, $options);                   
      }                                                                    
                                                                           
      // Custom URL building when file has no URL but base URL is set      
      // This fixes double-slash issues in the grandparent's implementation
      if (empty($file['url']) && $this->URL) {                             
          $decoded = $this->decode($hash);                                 
                                                                           
          // safely strip root prefix if present                           
          if (str_starts_with($decoded, $this->root)) {                    
              $path = substr($decoded, strlen($this->root));               
          } else {                                                         
              $path = $decoded;                                            
          }                                                                
                                                                           
          // remove leading separator to avoid double slashes              
          $path = ltrim($path, '/');                                       
                                                                           
          // normalize separators                                          
          $path = str_replace($this->separator, '/', $path);               
                                                                           
          if ($this->encoding) {                                           
              $path = $this->convEncIn($path, true);                       
          }                                                                
                                                                           
          $path = str_replace('%2F', '/', rawurlencode($path));            
                                                                           
          return rtrim($this->URL, '/') . '/' . $path;                     
      }                                                                    
                                                                           
      // Handle all other cases (file['url'] set, temporary) via parent    
      return parent::getContentUrl($hash, $options);                       
  }    
```

vitek-rostislav (changes requested):

packages/framework/src/Component/Filesystem/Flysystem/VolumeDriver.php
grossmannmartin: I'm not sure if it's actually better... 
This is tightly coupled with the parent implementation. We use some conditions, and we're counting on the parent to do the same conditions too. 

I don't like the elfinder implementation at all, but I don't think we make it better with this 😔

grossmannmartin: Also, I have excluded this class from the standards completely, because it forces us to rewrite the code, and it's harder to compare the implementation with the original. 

I have added the override attribute and the custom code marking, though. It's a good idea to have it there

pk16011990 (approved):

vitek-rostislav (approved):

PR #4418, opened 2026-01-26
[SSP-3796] Fix product availability store count

machacjan:
The original number returned by `getAvailableStoresCount` did not take into account the visibility by domain of related stores, which resulted in contradicting information displayed to customer.

----
  - https://jm-ssp-3796-fix-product-availability-store-count.odin.shopsys.cloud
  - https://cz.jm-ssp-3796-fix-product-availability-store-count.odin.shopsys.cloud
  - https://jm-ssp-3796-fix-product-availability-store-count.odin.shopsys.cloud/sk

----
  - https://jm-ssp-3796-fix-product-availability-store-count.odin.shopsys.cloud
  - https://cz.jm-ssp-3796-fix-product-availability-store-count.odin.shopsys.cloud
  - https://jm-ssp-3796-fix-product-availability-store-count.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Model/Product/Availability/ProductAvailabilityFacade.php
+++ b/packages/framework/src/Model/Product/Availability/ProductAvailabilityFacade.php
@@ -65,12 +65,15 @@ class ProductAvailabilityFacade
         }
 
         $productStocks = $this->productStockFacade->getProductStocksByProduct($product);
+        $storeCountsByStockId = $this->storeFacade->getStoreCountsByDomainIdIndexedByStockId($domainId);
 
         $count = 0;
 
         foreach ($productStocks as $productStock) {
+            $stock = $productStock->getStock();
+
             if ($productStock->getProductQuantity() > 0 && $productStock->getStock()->isEnabled($domainId)) {
-                $count += count($productStock->getStock()->getStores());
+                $count += $storeCountsByStockId[$stock->getId()] ?? 0;
             }
         }
 
--- a/packages/framework/src/Model/Store/StoreFacade.php
+++ b/packages/framework/src/Model/Store/StoreFacade.php
@@ -126,6 +126,14 @@ class StoreFacade
     }
 
     /**
+     * @return int[]
+     */
+    public function getStoreCountsByDomainIdIndexedByStockId(int $domainId): array
+    {
+        return $this->storeRepository->getStoreCountsByDomainIdIndexedByStockId($domainId);
+    }
+
+    /**
      * @param \Shopsys\FrameworkBundle\Model\Store\OpeningHours\OpeningHoursData[] $openingHoursDataArray
      * @return \Shopsys\FrameworkBundle\Model\Store\OpeningHours\OpeningHours[]
      */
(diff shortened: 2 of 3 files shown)


packages/framework/src/Model/Product/Availability/ProductAvailabilityFacade.php
```diff
@@ -96,9 +96,13 @@ public function getAvailableStoresCount(Product $product, int $domainId): ?int
         $count = 0;
 
         foreach ($productStocks as $productStock) {
-            if ($productStock->getProductQuantity() > 0 && $productStock->getStock()->isEnabled($domainId)) {
-                $count += count($productStock->getStock()->getStores());
+            $stock = $productStock->getStock();
+
+            if ($productStock->getProductQuantity() <= 0 || $stock->isEnabled($domainId) === false) {
+                continue;
             }
```
grossmannmartin: Personally, I find "positive" condition a bit more readable. 
Early exit (or continue in this case) with a negated condition makes me stop and invert it.

I suggest 
```php
if ($productStock->getProductQuantity() > 0 && $productStock->getStock()->isEnabled($domainId)) {
    $count += count($stock->getStoresByDomainId($domainId));
```

packages/framework/src/Model/Stock/Stock.php
```diff
@@ -225,4 +225,16 @@ public function getStores()
     {
         return $this->stores->getValues();
     }
+
+    /**
+     * @param int $domainId
+     * @return \Shopsys\FrameworkBundle\Model\Store\Store[]
+     */
+    public function getStoresByDomainId(int $domainId): array
```
grossmannmartin: Can we take it a little further? This solution is still (as the previous one) prone to the N+1 problem. Meaning each stock will call an extra query to get stores. 

What about preloading all store counts on the domain and then just add the count. 

something like 

```php
public function getAvailableStoresCount(Product $product, int $domainId): ?int
{
    if ($product->isMainVariant()) {
        return null;
    }

    $productStocks = $this->productStockFacade->getProductStocksByProduct($product);
    $storeCountsByStockId = $this->storeFacade->getStoreCountsByDomainIdIndexedByStockId($domainId);

    $count = 0;

    foreach ($productStocks as $productStock) {
        $stock = $productStock->getStock();

        if ($productStock->getProductQuantity() > 0 && $stock->isEnabled($domainId)) {
            $count += $storeCountsByStockId[$stock->getId()] ?? 0;
        }
    }

    return $count;
}
```

and getStoreCountsByDomainIdIndexedByStockId in the repository could look like 

```php
public function getStoreCountsByDomainIdIndexedByStockId(int $domainId): array
{
    $rows = $this->getQueryBuilder()
        ->select('IDENTITY(s.stock) AS stockId, COUNT(s.id) AS storeCount')
        ->where('s.domainId = :domainId')
        ->setParameter('domainId', $domainId)
        ->groupBy('s.stock')
        ->getQuery()
        ->getArrayResult();

    $result = [];

    foreach ($rows as $row) {
        $result[(int)$row['stockId']] = (int)$row['storeCount'];
    }

    return $result;
}
```

That way, we don't need an extra query for each stock, and we avoid hydrating store entities altogether.
What do you think?

grossmannmartin (changes requested):

packages/framework/src/Model/Stock/Stock.php
grossmannmartin: This method is no longer necessary

grossmannmartin (approved):

PR #4419, opened 2026-01-26
[SSP-1684] Simplify category SEO creation

machacjan:
Using simple grouped form with selectboxes instead of combination table.

----
  - https://jm-ssp-1684-simplify-category-seo-creation.odin.shopsys.cloud
  - https://cz.jm-ssp-1684-simplify-category-seo-creation.odin.shopsys.cloud
  - https://jm-ssp-1684-simplify-category-seo-creation.odin.shopsys.cloud/sk


grossmannmartin (changes requested):
Please take a look at the result. In the selectbox we have values from all locales, which contradicts the previous behavior. 

Image

vs original

Image

packages/framework/src/Form/Admin/CategorySeo/ReadyCategorySeoCombinationSelectorFormType.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    #[Override]
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setRequired(['categorySeoFiltersData', 'parameterValueChoices', 'backLink'])
+            ->setDefaults([
+                'flagChoices' => [],
+                'orderingChoices' => [],
```
grossmannmartin: it seems a little weird that parameterValueChoices are mandatory, while orderingChoices and flagChoices have default value. 
Shouldn't it be unified?

packages/framework/src/Form/Admin/CategorySeo/ReadyCategorySeoCombinationSelectorFormType.php
```diff
@@ -0,0 +1,103 @@
...
+            $builder->add($parametersGroup);
+        }
+
+        $basicInformationGroup = $builder->create('general', GroupType::class, [
+            'label' => 'Basic information',
+        ]);
+
+        $basicInformationGroup
+            ->add('domainId', HiddenType::class)
+            ->add('categoryId', HiddenType::class);
```
grossmannmartin: are those hidden fields (domainId and categoryId) necessary? the usage in the newCombinationsAction can be replaced. 
CategoryId is there available directly, domainId can be obtained the same way as it's passed to the form type ($this->adminDomainTabsFacade->getSelectedDomainId())

packages/framework/src/Form/Admin/CategorySeo/ReadyCategorySeoCombinationSelectorFormType.php
```diff
@@ -0,0 +1,103 @@
...
+                'back_url' => $options['backLink'],
+                'back_label' => t('Back to parameter selection'),
+            ]);
+    }
+
+    #[Override]
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setRequired(['categorySeoFiltersData', 'parameterValueChoices', 'backLink'])
```
grossmannmartin: not even parametervalue choices are actually required, or? 
I mean, I can create seomix without any parameters.

packages/framework/src/Form/Admin/CategorySeo/ReadyCategorySeoCombinationSelectorFormType.php
grossmannmartin, line 91: please add 
`'attr' => ['novalidate' => 'novalidate'],`
to avoid triggering browser validation

packages/framework/src/Model/CategorySeo/SelectedCategorySeoMixCombinationFactory.php
```diff
@@ -8,6 +8,27 @@
...
+     * @param int[] $parameterValueIdsByParameterIds
+     */
+    public function create(
+        int $domainId,
+        int $categoryId,
+        ?int $flagId,
+        ?string $ordering,
+        array $parameterValueIdsByParameterIds,
+    ): SelectedCategorySeoMixCombination {
+        ksort($parameterValueIdsByParameterIds);
```
grossmannmartin, line 21: why is this ksort needed?

grossmannmartin (changes requested):

packages/framework/src/Model/CategorySeo/SelectedCategorySeoMixCombinationFactory.php
machacjan: At this point it's _not_ needed, however it's a public method usable wherever, and down the line we rely on the `selectedCategorySeoMixCombinationJson` being structurally and functionally unique, so I would say a little bit of redundancy does no harm.

grossmannmartin (approved):

PR #4149, opened 2025-08-19
upgrade postgreSQL version to 18.x

vitek-rostislav:
#### Description, the reason for the PR
The new is always better :slightly_smiling_face: :rocket: 

For keeping the Postgres version fresh, we now use  `postgres:18-alpine` image without specifying the minor version, to get the minor upgrades automatically when available.

#### Fixes issues
... 

#### License Agreement for contributions

- [x] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://rv-update-postgres.odin.shopsys.cloud
  - https://cz.rv-update-postgres.odin.shopsys.cloud
  - https://rv-update-postgres.odin.shopsys.cloud/sk

----
:globe_with_meridians: Live Preview:
  - https://rv-update-postgres.odin.shopsys.cloud
  - https://cz.rv-update-postgres.odin.shopsys.cloud
  - https://rv-update-postgres.odin.shopsys.cloud/sk

PR #4420, opened 2026-01-26
Improve Payment GTM flow

JanMolcik:
- Add GTM `ec.payment` event tracking for external payment gateways (GoPay)
- Track payment success, failure, and abandonment scenarios
- Expose `paymentTransactionsCount` in Order GraphQL type for retry tracking 

#### Fixes issues
... 

----
  - https://tc-ssp-3808-gopay-gtm.odin.shopsys.cloud
  - https://cz.tc-ssp-3808-gopay-gtm.odin.shopsys.cloud
  - https://tc-ssp-3808-gopay-gtm.odin.shopsys.cloud/sk


docker/conf/docker-compose-orbstack.yml.dist
```diff
@@ -0,0 +1,255 @@
+x-variables:
```
grossmannmartin: Is it possible to split this commit into a separate PR? I have many questions and comments 😅

(Like, should it also be in the project-base? Do we want to add an option to an install script? Why are more ports in the storefront container? We should sync it with #4149, etc.)

project-base/storefront/pages/order-confirmation.tsx
```diff
@@ -76,6 +83,27 @@ const OrderConfirmationPage: FC<ServerSidePropsType> = () => {
...
     }, []);
 
+    // Handle GTM payment event when user returns from gateway via browser back button
+    // The pending payment is saved in localStorage when user initiates payment via GoPayGateway
+    useEffect(() => {
+        if (!orderData?.order || wasGtmPaymentEventFiredRef.current) {
+            return;
+        }
+
+        const pendingPayment = getGtmPendingPaymentFromLocalStorage(url);
```
grossmannmartin: What if the pending payment gets stuck in the local storage? 

By mistake, or some reload, ... Do we have a detection/recovery mechanism?

JanMolcik: There is a `usePaymentReturnHandler` in `Loaders` that is triggered while loading any page. This handler is responsible for checking for stale payment entries. 

Each local storage payment entry has a timestamp and we only consider recent payments (< 30 mins, configurable by `MAX_PENDING_PAYMENT_AGE_MS`). Older entries are discarded.

docker/conf/docker-compose-orbstack.yml.dist
JanMolcik: Sure I can split it. It is in a testing mode currently.

chlebektomas (approved):

docs/research/agent-ready-specification/index.md
```diff
@@ -0,0 +1,90 @@
+# Agent Ready Specification
```
chlebektomas: is `index.md` the right naming for this file?

project-base/storefront/pages/order-confirmation.tsx
```diff
@@ -60,112 +71,225 @@ const OrderConfirmationPage: FC<ServerSidePropsType> = () => {
...
-                        heading={t('Your order was created')}
+                        heading={pageHeading}
+                        headingClassName={undefined}
                         orderDetailUrl={orderDetailUrl}
                     >
-                        {orderPaymentType === TypePaymentTypeEnum.GoPay && (
+                        {!!order && orderPaymentType === TypePaymentTypeEnum.GoPay && (
                             <div className="mt-4">
                                 {isMaxTransactionCountReached ? (
+                                    maxTransactionOrderDetailUrl ? (
```
chlebektomas: would be nice to extract this translation logic for better readability, but it's up to you

project-base/storefront/pages/order-payment-confirmation.tsx
```diff
@@ -82,7 +145,24 @@ const OrderPaymentConfirmationPage: FC<ServerSidePropsType> = () => {
...
-        return null;
+        return (
+            <CommonLayout isFetchingData={isFetchingData} pageTypeOverride="order-confirmation" title={t('Order sent')}>
+                <Webline>
+                    <ConfirmationPageContent
+                        error={orderDetailError || isOrderPaymentPageContentError}
+                        heading={t('Order sent')}
+                        orderDetailUrl={orderDetailUrlFromQuery}
+                        content={
+                            !orderDetailError && !isOrderPaymentPageContentError
```
chlebektomas: would be nice something like `const text =  !orderDetailError && !isOrderPaymentPageContentError ?t("We couldn't .... : undefined`

chlebektomas (changes requested):

project-base/storefront/components/Pages/App/Loaders.tsx
```diff
@@ -18,6 +19,7 @@ export const Loaders = () => {
     useRefetchComparedProducts();
     useRefetchWishedProducts();
     useHtmlLangLoader();
+    usePaymentReturnHandler();
```
chlebektomas: is it necessary to have this globally on every page?

JanMolcik: Good point. We removed this from global loaders in follow-up. `usePaymentReturnHandler` was removed from `Loaders.tsx` and the hook/tests were deleted. Payment GTM is now final-only from confirmation flow, and we no longer emit synthetic `payment_abandoned` events globally.

docs/research/agent-ready-specification/index.md
JanMolcik: Kept intentionally. Here `index.md` acts as the section landing page and is referenced directly in `docs/research/agent-ready-specification/navigation.yml` (`arrange: index.md, ...`). So this naming is currently intentional for this folder structure.

project-base/storefront/pages/order-confirmation.tsx
JanMolcik: I extracted the max-transaction translation/rendering branch into `maxTransactionCountReachedContent` and now only render that variable in JSX. Behavior is unchanged; readability is better.

project-base/storefront/pages/order-payment-confirmation.tsx
JanMolcik: I introduced named variables `orderPaymentPageError` and `missingOrderFallbackContent` and pass them to `ConfirmationPageContent`. This keeps the same behavior but makes the rendering branch easier to read.

chlebektomas (approved):

project-base/storefront/components/Pages/Order/PaymentConfirmation/useInlinePaymentBackGuard.ts
```diff
@@ -0,0 +1,77 @@
+import { RefObject, useCallback, useEffect } from 'react';
+import {
+    buildPaymentConfirmationUrlFromSession,
+    markGoPayPaymentSessionForRedirectOnly,
+} from 'utils/goPayPaymentSessionStorage';
+
+// Handles navigation events (popstate, pageshow) during active inline GoPay payment.
+// Browser back from within GoPay's cross-origin iframe is NOT interceptable — the iframe
+// creates its own history entries, so back first traverses iframe history before reaching
+// our page. Session-based recovery on other pages ensures the user always reaches
```
JanMolcik, line 11: The recovery approach makes sense, but I am not fully convinced by the "regardless of how they navigate away" part yet. I can see recovery on `order-confirmation`, `order-payment-confirmation`, and order-detail pages, but I do not see equivalent session recovery on earlier checkout steps (`/cart`, `/order/transport-and-payment`, `/order/contact-information`). If real browser-back from the inline iframe lands on one of those pages, is it guaranteed that we still route the user back through `/order-payment-confirmation` instead of leaving them on a stale checkout step?

project-base/storefront/components/Pages/Customer/OrderDetail/OrderDetailContent.tsx
```diff
@@ -2,12 +2,37 @@ import { OrderDetailBasicInfo } from './OrderDetailBasicInfo';
...
 
 type OrderDetailContentProps = {
     order: TypeOrderDetailFragment;
 };
 
 export const OrderDetailContent: FC<OrderDetailContentProps> = ({ order }) => {
+    const { tryEmitPaymentEvent } = useEmitPendingPaymentEvent();
+    const orderPaymentItem = getOrderPaymentItem(order.items);
+
+    useEffect(() => {
```
JanMolcik, line 17: Nice reuse of the shared emitter. One thing I am still unsure about here: on order-detail pages this fires immediately from the first `order` payload, while `useRefreshOrderPaymentStatus()` only revalidates when that payload already says `hasPaymentInProcess === true`. If a retry has just finished and the first order query is briefly stale, we could emit the final `ec.payment` from outdated status data before any explicit status refresh happens. Could we gate this behind a confirmed status refresh when there is a pending GoPay attempt for the same order, or is there another guarantee that the first order payload is always fresh enough here?

packages/framework/src/Model/Order/Order.php
```diff
@@ -523,9 +529,20 @@ public function getPaymentTransactionsCount(): int
         return $this->paymentTransactions->count();
     }
 
+    /**
+     * Returns the external payment URL only if the last transaction belongs to the current payment method.
+     * After a payment method change, the old transaction URL is no longer valid
+     * and null forces the frontend to create a new transaction via PayOrder mutation.
+     */
```
grossmannmartin: I would not describe behavior outside this class. It may become obsolete easily. 

I think we may remove the comment altogether.

packages/framework/src/Model/Order/Order.php
```diff
@@ -588,6 +605,7 @@ protected function fillCommonFields(OrderData $orderData): void
 
         $this->promoCode = $orderData->promoCode;
         $this->freeTransportAndPaymentApplied = $orderData->freeTransportAndPaymentApplied;
+        $this->goPayBankSwift = $orderData->goPayBankSwift;
```
grossmannmartin: this line should be then removed from the __construct method (now this is called twice in some scenarios)

PR #4421, opened 2026-01-26
[SSP-1983] Move image config from YAML to attributes

machacjan:

----
  - https://jm-ssp-1983-move-image-config-from-yaml-to-attributes.odin.shopsys.cloud
  - https://cz.jm-ssp-1983-move-image-config-from-yaml-to-attributes.odin.shopsys.cloud
  - https://jm-ssp-1983-move-image-config-from-yaml-to-attributes.odin.shopsys.cloud/sk


packages/framework/src/Component/Image/Config/ImageConfigLoader.php
```diff
@@ -27,119 +24,101 @@ class ImageConfigLoader
...
-                    $entityConfig[ImageConfigDefinition::CONFIG_CLASS],
-                    $e,
-                );
-            }
+        if (count($attributes) === 0) {
+            return null;
         }
 
-        return $this->foundEntityConfigs;
+        return $attributes[0]->newInstance();
```
TomasLudvik: We are returning always only first attribute - in my scenario this would not work.

packages/framework/src/Component/Image/Config/Attributes/AsImageEntity.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Image\Config\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+class AsImageEntity
```
TomasLudvik: I would like to suggest a bit different approach. Current configuration was not ideal, we had something like:
- entity config (that included hidden default type)
- optional additional types

Now I would like to make it a bit simpler:
- we would have optional attribute `EntityImageFolder(?string $folderName = null)`  (feel free to suggest better naming) that would do same as `name` in your `AsImageEntity` and if it would not be defined, it would be set as camelCase of entity name. So for `TransportDomain` entity it would be named `transportDomain`. In current code it would be set in `Advert` as there is used name `noticer` so it would not change, but for others you could ommit it.
- we would have mandatory attribute `EntityImage(string $name = 'default', bool $multiple = false)` or `EntityImageType`. We would have to check if two or more `EntityImages` were not defined without name and throw error if such situation occured.

packages/framework/src/DependencyInjection/Compiler/RegisterImageEntitiesCompilerPass.php
```diff
@@ -0,0 +1,86 @@
...
+class RegisterImageEntitiesCompilerPass implements CompilerPassInterface
+{
+    /**
+     * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
+     */
+    #[Override]
+    public function process(ContainerBuilder $container): void
+    {
+        /** @var \Doctrine\ORM\Mapping\Driver\AttributeDriver $attributeReader */
+        $attributeReader = $container->get('doctrine.orm.default_metadata_driver'); // @phpstan-ignore-line Service is private but accessible during container compilation
```
TomasLudvik: Ignore this same way in `phpstan.neon` is ignored RegisterExtendedEntitiesCompilerPass

packages/framework/src/DependencyInjection/Compiler/RegisterImageEntitiesCompilerPass.php
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        return $classesWithAttribute;
+    }
+
+    /**
+     * @param string[] $classesWithAttribute
+     * @return string[]
+     */
+    protected function filterOutParentClasses(array $classesWithAttribute): array
```
TomasLudvik: Does this mean, that classes extended in project_base must have the attribute defined again?

TomasLudvik: Also this function could be improved like this:

```
  {                                                                                                                                                                                                                                                                                                                
      $parentClasses = [];                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                   
      foreach ($classesWithAttribute as $class) {                                                                                                                                                                                                                                                                  
          foreach (class_parents($class) as $parent) {                                                                                                                                                                                                                                                             
              $parentClasses[$parent] = true;                                                                                                                                                                                                                                                                      
          }                                                                                                                                                                                                                                                                                                        
      }                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                   
      return array_filter(                                                                                                                                                                                                                                                                                         
          $classesWithAttribute,                                                                                                                                                                                                                                                                                   
          static fn (string $class) => !isset($parentClasses[$class]),                                                                                                                                                                                                                                             
      );                                                                                                                                                                                                                                                                                                           
  }    
```

packages/framework/src/Component/Image/Config/Attributes/AsImageEntity.php
TomasLudvik: So for `Advert` it would look like this:

```
#[EntityImageFolder('noticer')]
#[EntityImage] // default would fallback as name: 'default'
#[EntityImage('web')]
#[EntityImage('mobile')]
```

TomasLudvik (changes requested):

packages/framework/src/Component/Image/Config/ImageConfigLoader.php
```diff
@@ -4,142 +4,152 @@
...
+        $reflectionClass = new ReflectionClass($entityClass);
+
+        do {
+            $attributes = $reflectionClass->getAttributes(EntityImage::class, ReflectionAttribute::IS_INSTANCEOF);
+
+            if ($attributes !== []) {
+                return array_map(static fn (ReflectionAttribute $attribute) => $attribute->newInstance(), $attributes);
             }
-        }
+        } while ($reflectionClass = $reflectionClass->getParentClass());
```
TomasLudvik, line 69: I like this approach, we will need to document it so developers know how to use it

packages/framework/src/Model/Advert/Advert.php
```diff
@@ -7,10 +7,16 @@
...
 use Doctrine\ORM\Mapping as ORM;
 use Ramsey\Uuid\Uuid;
+use Shopsys\FrameworkBundle\Component\Image\Config\Attributes\EntityImage;
+use Shopsys\FrameworkBundle\Component\Image\Config\Attributes\EntityImageFolder;
 use Shopsys\FrameworkBundle\Model\Category\Category;
 
 #[ORM\Table(name: 'adverts')]
 #[ORM\Entity]
+#[EntityImageFolder('noticer')]
+#[EntityImage()]
```
TomasLudvik:
```suggestion
#[EntityImage]
```

packages/framework/src/Component/Image/Config/ImageConfigLoader.php
```diff
@@ -4,142 +4,152 @@
...
-
-            if (array_key_exists($typeName, $result)) {
-                throw new DuplicateTypeNameExceptionInvalid($typeName);
+        foreach ($imageAttributes as $imageAttribute) {
+            if (array_key_exists($imageAttribute->name, $result)) {
+                throw new DuplicateTypeNameExceptionInvalid($imageAttribute->name);
             }
 
-            $result[$typeName] = $typeName;
+            $result[$imageAttribute->name] = $imageAttribute->name;
```
TomasLudvik: I feel like the key is not used anywhere - am I right?

packages/framework/tests/Unit/Component/Image/Config/ImageConfigLoaderTest.php
```diff
@@ -21,124 +24,82 @@ class ImageConfigLoaderTest extends TestCase
...
-        $imageEntityConfig = $preparedConfig[$inputConfig[0][ImageConfigDefinition::CONFIG_CLASS]];
-        $this->assertSame('Class_1', $imageEntityConfig->getEntityClass());
-        $this->assertSame('Name_1', $imageEntityConfig->getEntityName());
-        $this->assertFalse($imageEntityConfig->isMultiple(null));
-        $this->assertTrue($imageEntityConfig->isMultiple('TypeName_1'));
-        $this->assertFalse($imageEntityConfig->isMultiple('TypeName_2'));
+        $config = $this->imageConfigLoader->loadFromEntityClasses([
+            TestEntityDuplicateNameA::class,
+            TestEntityDuplicateNameB::class,
+        ]);
```
TomasLudvik: I expected this test to throw error on duplicate folder name

upgrade-notes/backend_20260203_100856.md
```diff
@@ -0,0 +1,62 @@
...
+
+- image configuration has been moved from YAML file (`images.yaml`) to PHP attributes on entity classes
+    - the `images.yaml` file is no longer used and can be deleted
+    - the parameter `shopsys.image_config_filepath` has been removed from `paths.yaml`
+
+- if you have extended `\Shopsys\FrameworkBundle\Component\Image\Config\ImageConfigLoader`, you need to update your code
+    - the constructor signature has changed - `Filesystem` parameter was removed
+    - the `loadFromYaml()` method has been removed
+    - the `loadFromArray()` method has been removed
+    - the new method `loadFromEntityClasses()` is now used to load image configuration from entity attributes
```
TomasLudvik: This should be omitted

upgrade-notes/backend_20260203_100856.md
```diff
@@ -0,0 +1,62 @@
...
+
+- the following classes have been removed:
+    - `\Shopsys\FrameworkBundle\Component\Image\Config\Exception\EntityParseException`
+    - `\Shopsys\FrameworkBundle\Component\Image\Config\ImageConfigDefinition`
+
+- if you have custom entities with image support configured in `images.yaml`, you need to migrate them to use the new PHP attributes:
+    - add `#[EntityImage]` attribute to your entity class to define an image type
+        - use `name` parameter to specify a custom type name (e.g., `#[EntityImage(name: 'gallery')]`)
```
TomasLudvik:
```suggestion
        - use `name` parameter to specify a custom type name (e.g., `#[EntityImage(name: 'gallery')]`), otherwise default name `default` will be used
```

upgrade-notes/backend_20260203_100856.md
```diff
@@ -0,0 +1,62 @@
...
+        - use `multiple: true` to allow multiple images for the type (e.g., `#[EntityImage(multiple: true)]`)
+        - the attribute is repeatable - you can add multiple `#[EntityImage]` attributes for different image types
+    - add `#[EntityImageFolder(name: 'folderName')]` attribute to customize the folder name for storing images
+        - if not specified, the folder name defaults to the lowercase class short name (e.g., `Product` → `product`)
```
TomasLudvik:
```suggestion
        - if not specified, the folder name defaults to the camelCase class short name (e.g., `SalesRepresentative` → `salesRepresentative`)
```

upgrade-notes/backend_20260203_100856.md
```diff
@@ -0,0 +1,62 @@
...
+              - name: mobile
+        ```
+
+        ```php
+        // After (entity class)
+        use Shopsys\FrameworkBundle\Component\Image\Config\Attributes\EntityImage;
+        use Shopsys\FrameworkBundle\Component\Image\Config\Attributes\EntityImageFolder;
+
+        #[EntityImageFolder('noticer')]
+        #[EntityImage()]
```
TomasLudvik:
```suggestion
        #[EntityImage]
```

packages/framework/src/Component/Image/Config/ImageConfigLoader.php
TomasLudvik: Also these tests would be nice:
- parent entity with set types and folder + child entity that has nothing set - check that all types and folder are set correctly
- parent entity with set types and folder + child entity with types set - check that types from child entity and folder from parent entity are set correctly
- parent entity with set types and folder + child entity with folder set - check that types from parent entity and folder from child entity are set correctly

packages/framework/src/Component/Image/Config/ImageConfigLoader.php
TomasLudvik: I found this in `ImageConfigLoaderTest:75` that seemed weird and led me to this thought.

TomasLudvik (changes requested):
Very nice job, I like this. I think that this way the necessity of configuration is minimal and setting is now clear. 

We should write some documentation for this so developers know how to set necessary attributes.

upgrade-notes/backend_20260203_100856.md
```diff
@@ -0,0 +1,62 @@
+#### Move image config from YAML to attributes ([#4421](https://github.com/shopsys/shopsys/pull/4421))
+
+- image configuration has been moved from YAML file (`images.yaml`) to PHP attributes on entity classes
+    - the `images.yaml` file is no longer used and can be deleted
+    - the parameter `shopsys.image_config_filepath` has been removed from `paths.yaml`
```
TomasLudvik: We do not explain what changed, but only what should developers do. That way we keep upgrade notes as short and straightforward as possible.

In this case I would not mention this block at all as it is covered by 
```
```
line, that we use in most of our upgrade notes at the end, where developers then see direct diff what to do in theirs project.

packages/framework/src/Component/Image/Config/ImageConfigLoader.php
machacjan: Test cases covered.

packages/framework/src/Component/Image/Config/ImageConfigLoader.php
machacjan: Unused, removed, tests updated.

packages/framework/tests/Unit/Component/Image/Config/ImageConfigLoaderTest.php
machacjan: Correct, this would previously throw. Updated the loader so that it throws in case of duplicate folder names with classes that have no relation.

upgrade-notes/backend_20260203_100856.md
machacjan: Understood, upgrade notes updated.

TomasLudvik (approved):

pk16011990 (approved):

PR #4357, opened 2025-12-29
14.5 update of dependencies

TomasLudvik:
#### Description, the reason for the PR
This is 14.5 version of
https://github.com/shopsys/shopsys/pull/3478
https://github.com/shopsys/shopsys/pull/3496
https://github.com/shopsys/shopsys/pull/3539
https://github.com/shopsys/shopsys/pull/3213

#### License Agreement for contributions

- [x] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://tl-14-5-update-dependencies.odin.shopsys.cloud
  - https://cz.tl-14-5-update-dependencies.odin.shopsys.cloud

PR #4430, opened 2026-02-03
Release version 14.5.1

grossmannmartin:

New version of 14.5 tech version.

#### Fixes issues
... 

----
  - https://rc-14-5-1.odin.shopsys.cloud
  - https://cz.rc-14-5-1.odin.shopsys.cloud


utils/releaser/phpunit.xml
```diff
@@ -7,7 +7,7 @@
     stopOnFailure="false"
     failOnWarning="true"
     beStrictAboutTestsThatDoNotTestAnything="false"
-    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/11.2/phpunit.xsd"
+    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/11.1/phpunit.xsd"
```
vitek-rostislav: it is rather a nitpick, however, this was updated to `11.2` in `14.5` branch within phpunit upgrade (https://github.com/shopsys/shopsys/commit/c308b60c0e2c89aa293e780b643442485a6f0962). I do not think we want to "downgrade" the schema here (we should rather update it in the main branch when I look at it...sooo - https://github.com/shopsys/shopsys/pull/4435

UPGRADE-14.5-tech.md
```diff
@@ -167,45 +169,45 @@ For commands not covered by Make targets (e.g., `exec`, `logs`, `restart`), use
...
+#### Replace shopsys/ordered-form package with becklyn/ordered-form-bundle ([#4357](https://github.com/shopsys/shopsys/pull/4357))
 
 -   calling `setPosition()` directly on existing form fields is not supported anymore, otherwise, ordering form fields works the same way
 -   see [becklyn/ordered-form-bundle](https://github.com/Becklyn/OrderedFormBundle) for the documentation
--   see #project-base-diff to update your project
+-   see [project-base diff](https://www.github.com/shopsys/project-base/commit/4b636e4c85c6f4c846f9f3ac1f5075b2cd6cca10) to update your project
 
 #### Upgrade Sentry package ([#4357](https://github.com/shopsys/shopsys/pull/4357))
 
--   see #project-base-diff to update your project
```
vitek-rostislav, line 184: something to consider - the upgrade line says "upgrade sentry package", but actually, there are multiple packages upgraded in https://github.com/shopsys/shopsys/pull/4357, so I would maybe improve the headline here to avoid confusion.

vitek-rostislav: Ah, ok, now I remember, there are multiple upgrade headlines that point to the same PR and the same project-base diff.  
image

I believe we can live with that, or if you think it would be less confusing if those were merged together with just one link to the project base diff, feel free to update it :grin:

vitek-rostislav (approved):

utils/releaser/phpunit.xml
grossmannmartin: cherry picked :) thanks

UPGRADE-14.5-tech.md
grossmannmartin: cefbeb92d1 better?

vitek-rostislav: yes, awesome :ok_hand:

vitek-rostislav (approved):

PR #4445, opened 2026-02-07
ensure friendly url slug is always url-encoded

vitek-rostislav:
When using URL-encoded slugs, the case-sensitive search might not work because, within the encoding, the case does not matter (e.g., both `caf%C3%A9` and `caf%c3%a9` are equivalent codes for `café`). 

To avoid the potential mismatched search, we now ensure the slug is always normalized and encoded in the database, and always encode (and normalize) the input slug before searching.

* normalization basically means we run `rawurlencode(rawurldecode())` on the given slug to ensure it is always encoded the same way

#### Fixes issues
... 

----
  - https://rv-friendly-urls.odin.shopsys.cloud
  - https://cz.rv-friendly-urls.odin.shopsys.cloud
  - https://rv-friendly-urls.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Form/FriendlyUrlType.php
+++ b/packages/framework/src/Form/FriendlyUrlType.php
@@ -5,12 +5,15 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Form;
 
 use Override;
+use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\FriendlyUrlSlugNormalizer;
 use Shopsys\FrameworkBundle\Component\Router\FriendlyUrl\UrlListData;
+use Shopsys\FrameworkBundle\Component\Translation\Translator;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Validator\Constraints;
+use Symfony\Component\Validator\Context\ExecutionContextInterface;
 
 final class FriendlyUrlType extends AbstractType
 {
@@ -33,10 +36,7 @@ final class FriendlyUrlType extends AbstractType
             'help' => t('If the slug contains non-ASCII characters (e.g. "café"), encode it as URL-encoded form (e.g. "caf%C3%A9"). You can use https://www.urlencoder.org/.'),
             'constraints' => [
                 new Constraints\NotBlank(),
-                new Constraints\Regex(
-                    pattern: static::SLUG_REGEX,
-                    message: 'Slug containing non-ASCII characters must be URL-encoded. You can use https://www.urlencoder.org/.',
-                ),
+                new Constraints\Callback(callback: [$this, 'validateSlugEncoding']),
             ],
         ]);
     }
@@ -49,5 +49,24 @@ final class FriendlyUrlType extends AbstractType
                 'limit_domains_by_ids' => [],
             ])
             ->setAllowedTypes('limit_domains_by_ids', 'array');
+    }
+
+    public function validateSlugEncoding(mixed $slug, ExecutionContextInterface $context): void
+    {
+        if (!is_string($slug) || $slug === '') {
+            return;
+        }
+
+        if (preg_match(static::SLUG_REGEX, $slug) === 1) {
+            return;
+        }
+
+        $context->buildViolation(
+            t('Slug containing non-ASCII characters must be URL-encoded. The encoded value of "%enteredValue%" is "%normalizedValue%".', [
+                '%enteredValue%' => $slug,
+                '%normalizedValue%' => FriendlyUrlSlugNormalizer::normalize($slug),
+            ], Translator::VALIDATOR_TRANSLATION_DOMAIN),
+        )
+            ->addViolation();
     }
 }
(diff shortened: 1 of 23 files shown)


packages/framework/src/Migrations/Version20260207120000.php
```diff
@@ -0,0 +1,68 @@
...
+     * @param \Doctrine\DBAL\Schema\Schema $schema
+     */
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $encodedRows = $this->sql(
+            "SELECT domain_id, slug, route_name, entity_id, main, redirect_to, redirect_code, last_modification
+             FROM friendly_urls
+             WHERE slug LIKE '%\%%' ESCAPE '\\'",
+        )->fetchAllAssociative();
```
grossmannmartin: It's not uncommon to have a friendly_urls table with hundreds of thousands or even milion records. 
Will loading all into memory cope with this?

edit: I see it's selecting only friendly_urls with the escaping sequence, I'm still curious about a limit, because when memory limit is hit on this, it's not pleasant to make it work from project

packages/framework/src/Migrations/Version20260207120000.php
```diff
@@ -0,0 +1,68 @@
...
+                    'slug' => $decodedSlug,
+                ],
+            )->fetchOne();
+
+            if ($decodedSlugExists) {
+                continue;
+            }
+
+            $this->sql(
+                'INSERT INTO friendly_urls (domain_id, slug, route_name, entity_id, main, redirect_to, redirect_code, last_modification)
```
grossmannmartin, line 38: We could skip the select exists and do `insert ... on conflict do nothing` instead, what do you think?

grossmannmartin (changes requested):

packages/framework/src/Migrations/Version20260207120000.php
vitek-rostislav: that is a relevant concern, I will try to look into it

packages/framework/src/Migrations/Version20260207120000.php
vitek-rostislav: Somebody told me it is bad for the primary index to do the updates, so :shrug: I will double check that

packages/framework/src/Migrations/Version20260207120000.php
vitek-rostislav: rewritten to while loop with `fetchAssociative` in a69349ceca

packages/framework/src/Migrations/Version20260207120000.php
vitek-rostislav: ok, so the different instance of the same person told me it is ok, so :smile: a69349ceca

grossmannmartin (approved):

grossmannmartin:
I suggest testing the following slugs - with client-side navigation, page refresh; when inserted into form and directly into the database

- `á-carte`
- `%c3%a1-carte`
- `%25c3%25a1-carte`
- `10%`

packages/framework/tests/Unit/Component/Router/FriendlyUrl/FriendlyUrlFactoryTest.php
```diff
@@ -63,6 +63,22 @@ public function testCreateForAllDomainsFailsForSingleDomainRoute(): void
...
 
+    public function testCreateNormalizesPercentEncodedSlug(): void
+    {
+        $friendlyUrlFactory = $this->getFriendlyUrlFactory();
+        $friendlyUrl = $friendlyUrlFactory->create('front_product_detail', 1, 1, 'caf%c3%a9');
+
+        $this->assertSame('caf%C3%A9', $friendlyUrl->getSlug());
+    }
+
+    public function testCreatePreservesPlainAsciiSlug(): void
```
grossmannmartin: nitpick: we could unify the test cases and use dataprovider (maybe even widen the test scenarios)

project-base/app/web/resolveFriendlyUrl.php
```diff
@@ -48,6 +48,16 @@ function returnSlug(string $route, ?string $redirectTo, ?int $redirectCode = nul
     exit;
 }
 
+function normalizeSlug(string $slug): string
```
grossmannmartin, line 51: nitpick: 
Maybe we could directly require the FriendlyUrlSlugNormalizer file in this script to avoid duplicate method declaration 🤔

packages/framework/src/Form/FriendlyUrlType.php
```diff
@@ -30,9 +30,13 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             'attr' => [
                 'placeholder' => 'slug',
             ],
+            'help' => t('If the slug contains non-ASCII characters (e.g. "café"), encode it as URL-encoded form (e.g. "caf%C3%A9"). You can use https://www.urlencoder.org/.'),
             'constraints' => [
                 new Constraints\NotBlank(),
-                new Constraints\Regex(static::SLUG_REGEX),
+                new Constraints\Regex(
```
grossmannmartin: nitpick, idea:
Maybe we could add a little javascript that encodes the slug on key up?

grossmannmartin (approved):

project-base/app/web/resolveFriendlyUrl.php
vitek-rostislav: yeah, I do not like the duplication, but I also do not like the higher complexity of monorepo vs project-base import logic, so keeping as-is

packages/framework/src/Form/FriendlyUrlType.php
vitek-rostislav: I have added the proper encoded value into the validation message

image

pk16011990 (approved):

PR #4461, opened 2026-02-17
reworked cron runner

grossmannmartin:

cron modules now run in isolated subprocesses

  - added CronModuleProcessRunner to spawn each module as a separate PHP process via shopsys:cron:internal-runner
  - added CronModuleRunnerFacade to handle in-process execution inside each subprocess
  - added CronInternalRunnerCommand (shopsys:cron:internal-runner, hidden) as the subprocess entry point
  - added stop_on_failure configuration option per cron instance
  - CronModule: suspend() now also sets $scheduled to true
  - CronModuleFacade: markCronAsFailed() now handles null $lastStartedAt safely

#### Fixes issues
... 

----
  - https://mg-cron.odin.shopsys.cloud
  - https://cz.mg-cron.odin.shopsys.cloud
  - https://mg-cron.odin.shopsys.cloud/sk

----
  - https://mg-cron.odin.shopsys.cloud
  - https://cz.mg-cron.odin.shopsys.cloud
  - https://mg-cron.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Command/CronCommand.php
+++ b/packages/framework/src/Command/CronCommand.php
@@ -46,25 +46,32 @@ class CronCommand extends Command
     protected function configure(): void
     {
         $this
-            ->addOption(self::OPTION_LIST, null, InputOption::VALUE_NONE, 'List all Service commands')
-            ->addOption(self::OPTION_MODULE, null, InputOption::VALUE_OPTIONAL, 'Service ID')
+            ->addOption(
+                self::OPTION_LIST,
+                null,
+                InputOption::VALUE_NONE,
+                'List all Service commands',
+            )
+            ->addOption(
+                self::OPTION_MODULE,
+                null,
+                InputOption::VALUE_REQUIRED,
+                'Run only specific module (service ID)',
+            )
             ->addOption(
                 self::OPTION_INSTANCE_NAME,
                 null,
                 InputOption::VALUE_REQUIRED,
-                'specific cron instance identifier',
+                'Specific cron instance identifier',
             )
             ->addOption(
                 self::OPTION_RUN_ALL_SERIALLY,
                 null,
                 InputOption::VALUE_NONE,
-                'run all crons serially (this is not safe to run in production environment)',
+                'Run all cron jobs serially (this is not safe to run in production environment)',
             );
     }
 
-    /**
-     * {@inheritdoc}
-     */
     #[Override]
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
--- a/packages/framework/src/Command/CronCommand.php
+++ b/packages/framework/src/Command/CronCommand.php
@@ -145,13 +145,6 @@ class CronCommand extends Command
      */
     protected function getCronCommands(array $cronModuleConfigs, bool $includeInstance = false): array
     {
-        uasort(
-            $cronModuleConfigs,
-            function (CronModuleConfig $cronModuleConfigA, CronModuleConfig $cronModuleConfigB) {
-                return strcmp($cronModuleConfigA->getServiceId(), $cronModuleConfigB->getServiceId());
-            },
-        );
-
         $commands = [];
 
         foreach ($cronModuleConfigs as $cronModuleConfig) {
--- a/packages/framework/src/Command/CronCommand.php
+++ b/packages/framework/src/Command/CronCommand.php
@@ -6,8 +6,8 @@ namespace Shopsys\FrameworkBundle\Command;
 
 use DateTimeZone;
 use NinjaMutex\Lock\LockInterface;
+use NinjaMutex\Mutex;
 use Override;
-use Shopsys\FrameworkBundle\Command\Exception\CronCommandException;
 use Shopsys\FrameworkBundle\Command\Exception\CronIsLockedException;
 use Shopsys\FrameworkBundle\Component\Cron\Config\CronModuleConfig;
 use Shopsys\FrameworkBundle\Component\Cron\CronFacade;
@@ -17,9 +17,11 @@ use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+use Symfony\Component\Process\Process;
 
 #[AsCommand(
     name: 'shopsys:cron',
@@ -78,6 +80,7 @@ class CronCommand extends Command
         $optionList = $input->getOption(self::OPTION_LIST);
         $optionInstanceName = $input->getOption(self::OPTION_INSTANCE_NAME);
         $optionRunAllSerially = $input->getOption(self::OPTION_RUN_ALL_SERIALLY);
+        $optionModule = $input->getOption(self::OPTION_MODULE);
 
         if ($optionList === true) {
             $this->listAllCronModulesSortedByServiceId($input, $output, $this->cronFacade);
@@ -92,9 +95,7 @@ class CronCommand extends Command
         }
 
         if ($optionRunAllSerially === true) {
-            foreach ($this->cronFacade->getAll() as $cronModuleConfig) {
-                $this->cronFacade->runModuleByServiceId($cronModuleConfig->getServiceId());
-            }
+            $this->runAllModulesSerially($output);
 
             return Command::SUCCESS;
         }
@@ -102,15 +103,26 @@ class CronCommand extends Command
         $instanceName = $optionInstanceName ?? $this->chooseInstance($input, $output);
 
         try {
-            $this->runCron($input, $this->cronFacade, $this->mutexFactory, $instanceName);
+            $mutex = $this->acquireInstanceMutex($instanceName);
         } catch (CronIsLockedException $e) {
             $output->writeln($e->getMessage());
 
             return Command::SUCCESS;
-        } catch (CronCommandException $e) {
-            $output->writeln($e->getMessage());
+        }
 
-            return Command::FAILURE;
+        try {
+            if ($optionModule !== null) {
+                $this->cronFacade->runSingleModule(
+                    $optionModule,
+                    $instanceName,
+                    $this->getProcessOutputCallback($output),
+                    $output->isDecorated(),
+                );
+            } else {
+                $this->runModulesInInstance($instanceName, $output);
+            }
+        } finally {
+            $mutex->releaseLock();
         }
 
         return Command::SUCCESS;
@@ -165,26 +177,9 @@ class CronCommand extends Command
         return $commands;
     }
 
-    protected function runCron(
-        InputInterface $input,
-        CronFacade $cronFacade,
-        MutexFactory $mutexFactory,
-        string $instanceName,
-    ): void {
-        $requestedModuleServiceId = $input->getOption(self::OPTION_MODULE);
-        $runAllModules = $requestedModuleServiceId === null;
-        $cronInstances = $this->parameterBag->get('cron_instances');
-        $instanceRunEveryMin = $cronInstances[$instanceName]['run_every_min'] ?? CronModuleConfig::RUN_EVERY_MIN_DEFAULT;
-
-        if ($instanceRunEveryMin < 0 || $instanceRunEveryMin > 30) {
-            $instanceRunEveryMin = CronModuleConfig::RUN_EVERY_MIN_DEFAULT;
-        }
-
-        if ($runAllModules) {
-            $cronFacade->scheduleModulesByTime($this->dateTimeHelper->getCurrentRoundedTimeForIntervalAndTimezone($instanceRunEveryMin, $this->getCronTimeZone()));
-        }
-
-        $mutex = $mutexFactory->getPrefixedCronMutex($instanceName);
+    protected function acquireInstanceMutex(string $instanceName): Mutex
+    {
+        $mutex = $this->mutexFactory->getPrefixedCronMutex($instanceName);
 
         if (!$mutex->acquireLock(0)) {
             throw new CronIsLockedException(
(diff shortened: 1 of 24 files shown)


grossmannmartin:
@coderabbitai review

packages/framework/src/Component/Cron/CronFacade.php
```diff
@@ -148,8 +131,18 @@ public function getAllForInstance(string $instanceName): array
...
-        }, $this->getAll()));
+        return array_unique(
+            array_map(
+                static fn (CronModuleConfig $config) => $config->getInstanceName(),
+                $this->getAll(),
+            ),
+        );
+    }
+
+    protected function isInstanceStopOnFailure(string $instanceName): bool
```
TomasLudvik: `shouldInstanceStopOnFailure` sounds more natural

packages/framework/src/Component/Cron/CronModuleRunnerFacade.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->cronModuleFacade->markCronAsStarted($cronModuleConfig);
+
+        try {
+            $status = $this->cronModuleExecutor->runModule(
+                $cronModuleService,
+                $this->cronModuleFacade->isModuleSuspended($cronModuleConfig),
+            );
+        } catch (Throwable $throwable) {
+            $this->cronModuleFacade->markCronAsFailed($cronModuleConfig);
+            $this->cronModuleFacade->unscheduleModule($cronModuleConfig);
```
TomasLudvik: Should it be unscheduled? What if the error was of some kind, that could be solved in next iteration? I thought that until know the cron remained scheduled.

packages/framework/src/Component/Cron/CronModuleProcessRunner.php
```diff
@@ -0,0 +1,64 @@
...
+    protected function createProcess(
+        string $serviceId,
+        string $instanceName,
+        bool $useAnsiOutput,
+        ?string $runId = null,
+    ): Process {
+        $command = [
+            PHP_BINARY,
+            $this->projectDir . '/bin/console',
+            'shopsys:cron:internal-runner',
```
TomasLudvik: This would be better to run via constant, that would be created in the Command

packages/framework/src/Command/CronCommand.php
```diff
@@ -165,39 +177,74 @@ protected function getCronCommands(array $cronModuleConfigs, bool $includeInstan
...
+        );
+
+        $this->cronFacade->runScheduledModulesForInstance(
+            $instanceName,
+            $this->getProcessOutputCallback($output),
+            $output->isDecorated(),
+        );
+    }
+
+    protected function runAllModulesSerially(OutputInterface $output): void
```
TomasLudvik, line 222: This method lacks any cron locking, is that ok?

packages/framework/src/DependencyInjection/Compiler/RegisterCronModulesCompilerPass.php
```diff
@@ -46,7 +46,7 @@ public function process(ContainerBuilder $container): void
     }
 
     /**
-     * @param array<string, array{run_every_min: int|null, timeout_iterated_cron_sec: int|null}> $cronInstances
+     * @param array<string, array<string, mixed>> $cronInstances
```
TomasLudvik: Cannot we keep more descriptive docblock?

TomasLudvik (changes requested):

packages/framework/src/Component/Cron/CronModuleRunnerFacade.php
```diff
@@ -0,0 +1,105 @@
...
+        } elseif ($status === CronModuleExecutor::RUN_STATUS_SUSPENDED) {
+            $this->cronModuleFacade->suspendModule($cronModuleConfig);
+        }
+
+        $this->logger->info('Cron module ended', [
+            'status' => $status,
+        ]);
+
+        return $status;
```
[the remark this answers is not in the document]
grossmannmartin: fixed in a3c711835a

packages/framework/src/Component/Cron/CronFacade.php
grossmannmartin: fixed in a60d0d1177

packages/framework/src/Component/Cron/CronModuleRunnerFacade.php
grossmannmartin: That's a good question. It's probably safer to keep it scheduled in case of dependent modules in the instance.
fixed in 3308824c48

packages/framework/src/Component/Cron/CronModuleProcessRunner.php
grossmannmartin: good idea 0bc555a6ef

packages/framework/src/Command/CronCommand.php
grossmannmartin: It's the same behavior as before. It's just more obvious now. 

Implementing the cron lock is quite tricky. We have to lock the module with the appropriate instance (lock is done at the instance level), so we have to iterate over instances and run modules from the instance. 
The tricky part is - what to do when we meet the already acquired lock in the middle of the processing 

service -> ok
export -> instance is locked
gopay -> ??

Should we stop? Silently skip?

packages/framework/src/DependencyInjection/Compiler/RegisterCronModulesCompilerPass.php
grossmannmartin: Every day I learn something new - a91f80c08b

packages/framework/tests/Unit/Component/Cron/CronFacadeTest.php
```diff
@@ -87,42 +48,54 @@ function (CronModuleConfig $cronModuleConfig) use ($validServiceId) {
...
+            ->with('service.id', 'default', $this->isCallable(), true, null)
+            ->willReturn(CronModuleProcessRunner::RESULT_SUCCESS);
+
+        $cronConfig = $this->createCronConfigWithRegisteredServices([]);
+
+        $result = $this->createCronFacade($cronConfig, $cronModuleFacadeMock, cronModuleProcessRunner: $cronModuleProcessRunnerMock)
+            ->runSingleModule('service.id', 'default', static function (): void {}, true);
+
+        $this->assertSame(CronModuleProcessRunner::RESULT_SUCCESS, $result);
+    }
```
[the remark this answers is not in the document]
grossmannmartin: added tests in ccb71be533

TomasLudvik (approved):

packages/framework/src/Command/CronCommand.php
henzigo: Is there any use case for this option? If not, let's remove it :)

packages/framework/src/Component/Cron/CronModuleFacade.php
```diff
@@ -130,7 +132,7 @@ public function markCronAsFailed(CronModuleConfig $cronModuleConfig): void
             [
                 'errorStatus' => CronModule::CRON_STATUS_ERROR,
                 'serviceId' => $cronModuleConfig->getServiceId(),
-                'now' => $this->clock->now()->format('Y-m-d H:i:s'),
+                'now' => $finishedAt->format('Y-m-d H:i:s'),
                 'lastDuration' => $lastCronDuration,
             ],
```
henzigo, line 137: Just a note.

That information is duplicated with `cron_module_runs` table.  Cannot we use only that table instead of updating the `cron_module` rows multiple times for every cron run?

packages/framework/src/Component/Cron/CronModuleProcessRunner.php
```diff
@@ -0,0 +1,65 @@
...
+        ?string $runId = null,
+    ): string {
+        $process = $this->createProcess(
+            $serviceId,
+            $instanceName,
+            $useAnsiOutput,
+            $runId,
+        );
+
+        $process->run($processOutputCallback);
```
henzigo, line 34: Did you consider using [ProcessHelper](https://symfony.com/doc/current/components/console/helpers/processhelper.html)?

packages/framework/src/Component/Cron/CronModuleRunnerFacade.php
```diff
@@ -0,0 +1,107 @@
...
+            $this->logger->popProcessor();
+        }
+    }
+
+    protected function runModuleByServiceId(string $serviceId): string
+    {
+        $cronModuleConfig = $this->cronConfig->getCronModuleConfigByServiceId($serviceId);
+
+        return $this->runSingleModule($cronModuleConfig);
+    }
```
henzigo: This method is probably useless as it's called only in `runModuleByServiceIdInContext()`

packages/framework/src/Component/Cron/CronFacade.php
```diff
@@ -52,79 +65,49 @@ protected function runModules(array $cronModuleConfigs, string $instanceName): v
         $this->logger->info('Start of cron instance');
 
         try {
+            $stopOnFailure = $this->shouldInstanceStopOnFailure($instanceName);
+
             foreach ($cronModuleConfigs as $cronModuleConfig) {
-                $this->runSingleModule($cronModuleConfig);
+                if ($this->cronModuleFacade->isModuleDisabled($cronModuleConfig)) {
+                    $this->cronModuleFacade->unscheduleModule($cronModuleConfig);
```
henzigo, line 72: Hm. Shouldn't this behaviour be defined in the entity itself in the `disable()` method? 

Probably, this method shouldn't iterate over disabled modules in the first place.

packages/framework/src/Command/CronCommand.php
henzigo: Oh, never mind. I found the use case 👍 

It almost works perfectly, but there is one little issue:

```
INFO      [cron] Cron module started ["module" => "FeedCronModule","instance" => "export"]
INFO      [cron] Queue is empty, no feeds to process. ["module" => "FeedCronModule","instance" => "export"]
INFO      [cron] Cron module ended ["status" => "ok"] ["module" => "FeedCronModule","instance" => "export"]
```

So feeds are not generated.

packages/framework/src/Command/CronInternalRunnerCommand.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+#[AsCommand(
+    name: self::COMMAND_NAME,
+    description: 'Runs a single cron module in isolated internal process.',
+    hidden: true,
+)]
+class CronInternalRunnerCommand extends Command
```
henzigo: Why don't we name it `CronModuleRunnerCommand`?

It's hidden, so I think it's safe and probably more descriptive :) What do you think?

packages/framework/src/Command/CronCommand.php
```diff
@@ -165,39 +177,74 @@ function (CronModuleConfig $cronModuleConfigA, CronModuleConfig $cronModuleConfi
...
-        }
+        return $instanceRunEveryMin;
+    }
 
-        $mutex = $mutexFactory->getPrefixedCronMutex($instanceName);
+    protected function runModulesInInstance(
+        string $instanceName,
+        OutputInterface $output,
+    ): void {
+        $instanceRunEveryMin = $this->getInstanceRunEveryMin($instanceName);
```
henzigo, line 209: This retarded hack should be completely rewritten into something more common, like a normal Unix cron string format (`*/5 * * * *`). 

It has only advantages:

- More scheduling options
- No hacks
- Multiple packages for validating and interpreting (Also working with time zones)
- Standard in the IT world

packages/framework/src/Component/Cron/CronFacade.php
```diff
@@ -148,8 +131,18 @@ public function getAllForInstance(string $instanceName): array
...
+            array_map(
+                static fn (CronModuleConfig $config) => $config->getInstanceName(),
+                $this->getAll(),
+            ),
+        );
+    }
+
+    protected function shouldInstanceStopOnFailure(string $instanceName): bool
+    {
+        $cronInstances = $this->parameterBag->get('cron_instances');
```
henzigo, line 144: I know this is not new code, but do you realize that the `framework` bundle is dependent on a parameter that is defined in `project-base`? 


It should be rewritten into the standard configuration

project-base/app/config/cron.yaml
```diff
@@ -3,6 +3,13 @@ parameters:
         default:
             run_every_min: 5
             timeout_iterated_cron_sec: 240
+            stop_on_failure: true
+        service:
+            stop_on_failure: false
+        export:
+            stop_on_failure: false
+        gopay:
+            stop_on_failure: false
```
henzigo, line 12: I'm wondering if this label is sufficient. It would be great to add some visual information, at least to the administration, so that there is a place where I can see how each instance works.

henzigo:
If you'd like, feel free to ignore my comments. But I need to tell you something. I don't know what to think about this `rework`.  

On the one hand, I'm glad that someone has started digging into this. On the other hand, however, I'm sad that you haven't dug very deep or thought it through very carefully.

Why can't we rewrite all those cron jobs to make sense in this day and age?

Why not just use Symfony Scheduler and build custom behavior around it? It should not be that hard. Also, it can open other thoughts about how we work and how we should work with cron modules.

packages/framework/src/Command/CronCommand.php
grossmannmartin: feeds have their own timing, so when the time is right, they're generated properly. Or?

packages/framework/src/Component/Cron/CronModuleProcessRunner.php
grossmannmartin: I did, and I don't know the benefit of it

packages/framework/src/Component/Cron/CronModuleRunnerFacade.php
grossmannmartin: inlined.

packages/framework/src/Component/Cron/CronFacade.php
grossmannmartin: That would not handle the edge case of the disabled, but scheduled module.

henzigo: Like after upgrading? That could be fixed by migration

packages/framework/src/Command/CronInternalRunnerCommand.php
grossmannmartin: renamed

packages/framework/src/Command/CronCommand.php
grossmannmartin: I would not go with such drastic change

packages/framework/src/Component/Cron/CronFacade.php
grossmannmartin: Yes, we have that in various places

packages/framework/src/Command/CronCommand.php
henzigo: I understand. But I would love to see it as the following task

project-base/app/config/cron.yaml
grossmannmartin: good idea, added badge on the dashboard

pk16011990 (approved):

PR #4463, opened 2026-02-19
added feed recovery for corrupted or nonexistent files during feed ge…

TomasLudvik:
We had problem on our servers that sometimes on S3 files could not be moved and that broke feed generation. This should fix such problem.

----
  - https://tl-fix-uneditable-file-in-feed.odin.shopsys.cloud
  - https://cz.tl-fix-uneditable-file-in-feed.odin.shopsys.cloud
  - https://tl-fix-uneditable-file-in-feed.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Model/Feed/FeedExport.php
+++ b/packages/framework/src/Model/Feed/FeedExport.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Model\Feed;
 
 use Doctrine\ORM\EntityManagerInterface;
+use League\Flysystem\FilesystemException;
 use League\Flysystem\FilesystemOperator;
 use League\Flysystem\MountManager;
+use Psr\Log\LoggerInterface;
 use Shopsys\FrameworkBundle\Component\Domain\Config\DomainConfig;
 use Shopsys\FrameworkBundle\Component\String\TransformStringHelper;
 use Shopsys\FrameworkBundle\DependencyInjection\ServicesResetter;
@@ -31,6 +33,7 @@ class FeedExport
         protected readonly string $feedFilepath,
         protected readonly string $feedLocalFilepath,
         protected readonly ServicesResetter $servicesResetter,
+        protected readonly LoggerInterface $logger,
         protected ?int $lastSeekId = null,
     ) {
     }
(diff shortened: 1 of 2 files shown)


packages/framework/src/Model/Feed/FeedExport.php
```diff
@@ -31,19 +33,65 @@ public function __construct(
...
+                    'temporaryFilepath' => $this->getTemporaryFilepath(),
+                    'exception' => $deleteException,
+                ],
             );
-        } else {
-            $this->localFilesystem->touch($this->getTemporaryLocalFilepath());
+        }
+    }
+
+    protected function tryDeleteLocalTemporaryFile(): void
```
grossmannmartin: This method is almost identical to `clearTemporaryFile()`

packages/framework/src/Model/Feed/FeedExport.php
```diff
@@ -31,19 +33,65 @@ public function __construct(
...
+
+        $this->tryDeleteRemoteTemporaryFile();
+        $this->tryDeleteLocalTemporaryFile();
+        $this->lastSeekId = null;
+    }
+
+    protected function tryDeleteRemoteTemporaryFile(): void
+    {
+        try {
+            $this->filesystem->delete($this->getTemporaryFilepath());
```
grossmannmartin: Maybe we could unify this with the `clearTemporaryFile()` method? 
I don't think there will be a scenario where we want to delete only one of the temporary files

grossmannmartin (changes requested):

grossmannmartin (approved):

packages/framework/src/Model/Feed/FeedExport.php
```diff
@@ -31,20 +33,43 @@ public function __construct(
...
+            try {
+                $this->mountManager->move(
+                    'main://' . $this->getTemporaryFilepath(),
+                    'local://' . $this->transformStringHelper->removeDriveLetterFromPath($this->getTemporaryLocalFilepath()),
+                );
+
+                return;
+            } catch (FilesystemException $exception) {
+                $this->handleWakeUpFailure($exception);
+            }
```
henzigo, line 53: I feel like this solution only addresses the consequences, not the cause. We should probably also handle `sleep()` and `finishFile()` methods.

What about implementing a "try repeater" that will try to move those files, for example, 3 times before it fails. It could prevent the regeneration whole feed.

packages/framework/src/Model/Feed/FeedExport.php
```diff
@@ -31,20 +33,43 @@ public function __construct(
...
+            [
+                'feedName' => $this->feed->getInfo()->getName(),
+                'domainId' => $this->domainConfig->getId(),
+                'temporaryFilepath' => $this->getTemporaryFilepath(),
+                'exception' => $exception,
+            ],
+        );
+
+        $this->clearTemporaryFile();
+        $this->lastSeekId = null;
```
henzigo, line 73: Extracting this into a separate method seems unnecessary.

packages/framework/src/Model/Feed/FeedExportFactory.php
```diff
@@ -23,6 +24,7 @@ public function __construct(
         protected readonly MountManager $mountManager,
         protected readonly ServicesResetter $servicesResetter,
         protected readonly TransformStringHelper $transformStringHelper,
+        protected readonly LoggerInterface $logger,
```
henzigo, line 27: Shouldn't those changes be addressed in the upgrade notes?

henzigo (commented):
Just some notes to consider

packages/framework/src/Model/Feed/FeedExport.php
TomasLudvik: As we discussed this previously on Slack - there are no errors in Sentry that point to your paradigma, that problem is during sleep. I want to implement this fix and see if it solves the problem. If another error will occur in the future, we can improve it further.

packages/framework/src/Model/Feed/FeedExport.php
TomasLudvik: I do not feel like it

packages/framework/src/Model/Feed/FeedExportFactory.php
TomasLudvik: We do not write changes of method interfaces to the upgrade notes anymore

PR #4465, opened 2026-02-20
claude: added test writing skill

grossmannmartin:

Added skill for better writing tests

#### Fixes issues
... 

----
  - https://mg-claude-test-skill.odin.shopsys.cloud
  - https://cz.mg-claude-test-skill.odin.shopsys.cloud
  - https://mg-claude-test-skill.odin.shopsys.cloud/sk

### DIFF
--- a/.claude/skills/test-writing/SKILL.md
+++ b/.claude/skills/test-writing/SKILL.md
@@ -1,0 +1,433 @@
+---
+name: test-writing
+description: >
+  This skill MUST be used when the user asks to write tests, add tests, create test cases,
+  test a specific class or method, or when working on any *Test.php file in this Shopsys
+  application. Applies codebase-specific best practices across unit, functional, GraphQL API,
+  smoke, and acceptance test layers.
+version: 1.0.0
+---
+
+# Shopsys Test Writing Guide
+
+## 1. Layer Selection — Pick the Right Base Class First
+
+```
+What am I testing?
+│
+├─ Pure logic, no I/O, all deps injectable or mockable?
+│   └─► Unit Test
+│       extends PHPUnit\Framework\TestCase
+│       namespace mirrors the source location:
+│         project-base/src/...  → Tests\App\Unit\...
+│         packages/framework/   → Tests\FrameworkBundle\Unit\...
+│         packages/frontend-api/→ Tests\FrontendApiBundle\Unit\...
+│         packages/<foo>/       → Tests\<FooBundle>\Unit\...
+│
+├─ Service that reads+writes the DB (mutations, creates, deletes)?
+│   └─► TransactionFunctionalTestCase  ← auto-rollback after each test
+│       extends Tests\App\Test\TransactionFunctionalTestCase
+│       namespace Tests\App\Functional\...
+│
+├─ Service that only reads the DB (no mutations)?
+│   └─► FunctionalTestCase
+│       extends Tests\App\Test\FunctionalTestCase
+│       namespace Tests\App\Functional\...
+│
+├─ GraphQL API endpoint?
+│   ├─ Anonymous request            → GraphQlTestCase
+│   ├─ Authenticated B2C customer   → GraphQlWithLoginTestCase  (auto-calls login())
+│   ├─ B2B domain, anonymous        → GraphQlB2bDomainTestCase
+│   └─ B2B domain + authenticated   → GraphQlB2bDomainWithLoginTestCase
+│   All in namespace Tests\FrontendApiBundle\Functional\...
+│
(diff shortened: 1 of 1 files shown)


.claude/skills/test-writing/SKILL.md
```diff
@@ -0,0 +1,318 @@
...
+
+## 1. Layer Selection — Pick the Right Base Class First
+
+```
+What am I testing?
+│
+├─ Pure logic, no I/O, all deps injectable or mockable?
+│   └─► Unit Test
+│       extends PHPUnit\Framework\TestCase
+│       namespace Tests\App\Unit\...
```
vitek-rostislav: I would be careful with this. There are unit tests in packages as well (so the namespace is then e.g. `Tests\FrameworkBundle\Unit`, etc.)

.claude/skills/test-writing/SKILL.md
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * @inject
+ */
+private ProductDataFactory $productDataFactory;
+```
+
+Rules:
+- `private` visibility in `project-base` tests (final classes, strict)
+- `protected` visibility in `packages` tests (extendable)
```
vitek-rostislav: I do not think this is true within the tests

vitek-rostislav (changes requested):

.claude/skills/test-writing/SKILL.md
grossmannmartin: I have addressed this, along with some improvements from the first usage of the skill

.claude/skills/test-writing/SKILL.md
grossmannmartin: I have addressed this, along with some improvements from the first usage of the skill

.claude/skills/test-writing/SKILL.md
```diff
@@ -316,3 +342,54 @@ final class ProductQueryTest extends GraphQlTestCase
 - No return type on `setUp()` — it is `void` but the annotation is inherited via `#[Override]`
 - `#[Override]` attribute on every `setUp()` / `tearDown()` override
 - Full type hints and docblocks on all injected properties (aids IDE autocompletion)
+- **Don't promote `setUp()` locals to class properties unless test methods need them.** If a variable is only used inside `setUp()` to build shared state (e.g. a stub used solely to register a service and extract its class name), keep it local. Only promote to a property when at least one test method references it directly.
+
+---
+
+## 11. Verifying Tests — Always Run via Phing
```
vitek-rostislav: Interesting, even I do not do that 😄 I mean, when I debug a single test, I do not want to wait for the whole suite. Maybe we could add instructions for running the single test with the configuration defined within the phpunit command? I use this:

```
vendor/bin/phpunit --colors=always --configuration /var/www/html/project-base/app/phpunit.xml --testsuite Functional,FrontendApiFunctional,FrontendApiFunctionalB2b,Unit --filter <testName>
```

This can be further adjusted based on what test needs to be run. I believe Claude will understand that when running a test from the frontend-api package, he (she? it? 😄) needs to define the proper config and testsuite.

.claude/skills/test-writing/SKILL.md
```diff
@@ -316,3 +342,54 @@ final class ProductQueryTest extends GraphQlTestCase
...
+// GOOD — createStub() for pure return-value fakes
+$facadeStub = $this->createStub(SomeFacade::class);
+$facadeStub->method('find')->willReturn($entity);
+
+// GOOD — createMock() when verifying a call was made
+$mailerMock = $this->createMock(Mailer::class);
+$mailerMock->expects($this->once())->method('send');
+```
+
+### Two stubs of the same interface share the same class name
```
vitek-rostislav, line 418: This looks like feedback from a very specific problem, but I guess we should get used to such things in the AI md files, right?

vitek-rostislav (changes requested):

.claude/skills/test-writing/SKILL.md
grossmannmartin: We (Claude and I) tried to improve this. What about now? 7d105feca2

.claude/skills/test-writing/SKILL.md
grossmannmartin: I don't know TBH 🤔
I stumbled upon this problem, but it's true that this should be common knowledge (I mean, array indexing behavior). 
I am really surprised that he had this problem in the first place, but I followed the rule - "if you correct him, make it known to avoid this problem in the future."

vitek-rostislav (approved):

PR #4448, opened 2026-02-09
Update to Symfony 7.4

TomasLudvik:
#### Description, the reason for the PR
We want to use latest LTS from Symfony so we are upgrading to Symfony 7.4 and alongside upgrading most of the dependencies so our technical debt is minimal as possible.

#### License Agreement for contributions

- [x] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://tl-update-symfony-7.odin.shopsys.cloud
  - https://cz.tl-update-symfony-7.odin.shopsys.cloud
  - https://tl-update-symfony-7.odin.shopsys.cloud/sk

----
:globe_with_meridians: Live Preview:
  - https://tl-update-symfony-7.odin.shopsys.cloud
  - https://cz.tl-update-symfony-7.odin.shopsys.cloud
  - https://tl-update-symfony-7.odin.shopsys.cloud/sk

PR #4466, opened 2026-02-21
phpstan: move particular file ignores from configs to inline suppressions

vitek-rostislav:
- We forgot to update the `packages/framework/phpstan.neon` in https://github.com/shopsys/shopsys/pull/4448
- It is better to have the inline suppressions where possible to avoid DRY in configuration (monorepo vs packages)

#### Fixes issues
... 

----
  - https://rv-fix-phpstan.odin.shopsys.cloud
  - https://cz.rv-fix-phpstan.odin.shopsys.cloud
  - https://rv-fix-phpstan.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/phpstan.neon
+++ b/packages/framework/phpstan.neon
@@ -5,12 +5,6 @@ parameters:
     ignoreErrors:
         # Don't forget to add these rules to phpstan.neon in monorepo
         -
-            message: '#Property Shopsys\\FrameworkBundle\\Model\\Order\\Order::\$transport \(Shopsys\\FrameworkBundle\\Model\\Transport\\Transport\) does not accept null\.#'
-            path: %currentWorkingDirectory%/src/Model/Order/Order.php
-        -
-            message: '#Property Shopsys\\FrameworkBundle\\Model\\Order\\Order::\$payment \(Shopsys\\FrameworkBundle\\Model\\Payment\\Payment\) does not accept null\.#'
-            path: %currentWorkingDirectory%/src/Model/Order/Order.php
-        -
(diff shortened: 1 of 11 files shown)


packages/framework/phpstan.neon
```diff
@@ -13,6 +13,14 @@ parameters:
         -
             message: '#^Unsafe usage of new static\(\).#'
             path: %currentWorkingDirectory%/src/*
+        -
+            # PHPStan inference issue with Yaml::parse flags in Symfony 7 - runtime behavior is correct
+            message: '#^Parameter \#2 \$flags of method Symfony\\Component\\Yaml\\Parser::parse\(\) expects#'
+            path: %currentWorkingDirectory%/src/Command/YamlLintCommand.php
+        -
+            # Symfony 7.4 forward compatibility - parent method will require 2 parameters in Symfony 8.0
+            message: '#^Method Symfony\\Component\\Security\\Core\\User\\InMemoryUserChecker::checkPostAuth\(\) invoked with 2 parameters, 1 required\.$#'
```
grossmannmartin: If we move the PHPStan error ignore from phpstan.neon directly into the specific file, we avoid having to maintain the ignore rule in two separate places. At the same time, when refactoring the affected code, it naturally prompts us to reconsider whether the ignore is still necessary.

grossmannmartin (changes requested):

packages/framework/phpstan.neon
vitek-rostislav: reworked :wink:

grossmannmartin (approved):

PR #2515, opened 2022-11-19
[#2375] Graphql request with error must roll back db changes

stanoMilan:
| Q             | A
| ------------- | ---
|Description, reason for the PR| ...
|New feature| No 
|[BC breaks](https://docs.shopsys.com/en/latest/contributing/backward-compatibility-promise/)| No 
|Fixes issues| ... closes #2375
|Have you read and signed our [License Agreement for contributions](https://www.shopsys.com/license-agreement)?| Yes

PR #4479, opened 2026-02-25
FE API: rollback mutation & discard async messages on error

### LINKED ISSUE #2375: (GRAPQHL) FRONTEND API request is not in transaction (if throw exception)
## Describe the bug

Query from FRONTEND API (graphql) is not in transaction. 
Example:
Throw expcetion after create Order.
Create order by GRAPHQL . You have to see new order in adminstratrion. 

### Additional context
It is because, all graphql request finished with code 200. Overbolog bundle catch errors.

vitek-rostislav:

GQL requests always return 200 OK (the errors are handled internally), so the error handlers listening on kernel errors are not called. This PR addresses that problem, so that DB transactions are rolled back when an error occurs during GQL mutation. Moreover, RabbitMQ messages are now discarded on an error within GQL request.

#### Fixes issues
closes #2515
closes #2375 

----
  - https://rv-mutation-rollback.odin.shopsys.cloud
  - https://cz.rv-mutation-rollback.odin.shopsys.cloud
  - https://rv-mutation-rollback.odin.shopsys.cloud/sk

----
  - https://rv-mutation-rollback.odin.shopsys.cloud
  - https://cz.rv-mutation-rollback.odin.shopsys.cloud
  - https://rv-mutation-rollback.odin.shopsys.cloud/sk

### DIFF
--- a/packages/frontend-api/src/Component/HttpFoundation/GraphqlMutationErrorRollbackSubscriber.php
+++ b/packages/frontend-api/src/Component/HttpFoundation/GraphqlMutationErrorRollbackSubscriber.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Component\HttpFoundation;
+
+use Overblog\GraphQLBundle\Event\Events;
+use Overblog\GraphQLBundle\Event\ExecutorResultEvent;
+use Override;
+use Shopsys\FrameworkBundle\Component\HttpFoundation\SilencedExceptionEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+
+class GraphqlMutationErrorRollbackSubscriber implements EventSubscriberInterface
+{
+    protected const string MUTATION_OPERATION = 'mutation';
+
+    protected bool $shouldRollbackTransaction = false;
+
+    public function __construct(
+        protected readonly GraphqlOperationTypeResolver $graphqlOperationTypeResolver,
+        protected readonly EventDispatcherInterface $eventDispatcher,
+    ) {
+    }
+
+    /**
+     * @return array<string, string|array<int, array{0: string, 1?: int}>>
+     */
+    #[Override]
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            Events::POST_EXECUTOR => [
+                ['onPostExecutor'],
+            ],
+            KernelEvents::RESPONSE => [
+                ['onKernelResponse', 10],
+            ],
+        ];
+    }
+
+    public function onPostExecutor(ExecutorResultEvent $event): void
+    {
+        if (count($event->getResult()->errors) === 0) {
+            return;
+        }
+
+        $operationType = $this->graphqlOperationTypeResolver->resolveOperationTypeFromPayload([
+            'query' => $event->getExecutorArguments()->getRequestString(),
+            'operationName' => $event->getExecutorArguments()->getOperationName(),
+        ]);
+
+        if ($operationType !== static::MUTATION_OPERATION) {
+            return;
+        }
+
+        $this->shouldRollbackTransaction = true;
+    }
+
+    public function onKernelResponse(ResponseEvent $event): void
+    {
+        if (!$event->isMainRequest() || !$this->shouldRollbackTransaction) {
+            return;
+        }
+
+        $this->shouldRollbackTransaction = false;
+        $this->eventDispatcher->dispatch(new SilencedExceptionEvent());
+    }
+}
--- a/packages/frontend-api/src/Component/Messenger/GraphqlErrorResetDelayedEnvelopesSubscriber.php
+++ b/packages/frontend-api/src/Component/Messenger/GraphqlErrorResetDelayedEnvelopesSubscriber.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Component\Messenger;
+
+use Overblog\GraphQLBundle\Event\Events;
+use Overblog\GraphQLBundle\Event\ExecutorResultEvent;
+use Override;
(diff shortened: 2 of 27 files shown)


vitek-rostislav:
@coderabbitai review

packages/frontend-api/src/Component/HttpFoundation/GraphqlMutationErrorRollbackSubscriber.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function onKernelResponse(ResponseEvent $event): void
+    {
+        if (!$this->shouldRollbackTransaction || !$event->isMainRequest() || !$this->em->getConnection()->isTransactionActive()) {
+            return;
+        }
+
+        TransactionalMasterRequestListener::setTransactionManuallyRollbacked();
+        $this->em->rollback();
+    }
```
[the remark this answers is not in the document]
vitek-rostislav: implemented in 470fb5099c

packages/frontend-api/src/Component/HttpFoundation/GraphqlMutationErrorRollbackSubscriber.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function onKernelRequest(RequestEvent $event): void
+    {
+        if (!$event->isMainRequest()) {
+            return;
+        }
+
+        $this->shouldRollbackTransaction = false;
+    }
```
grossmannmartin: isn't this event superfluous? 
$shouldRollbackTransaction is already set to false from the object initialization. 

Maybe we could keep it as a safeguard (or if we adopt frankenphp). But I don't think the check for the main/sub request serves any practical purpose here

packages/frontend-api/src/Component/HttpFoundation/GraphqlMutationErrorRollbackSubscriber.php
grossmannmartin, line 70: I would reset the `shouldRollbackTransaction`

packages/frontend-api/src/Component/Messenger/GraphqlErrorResetDelayedEnvelopesSubscriber.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function onPostExecutor(ExecutorResultEvent $event): void
+    {
+        if (count($event->getResult()->errors) === 0) {
+            return;
+        }
+
+        $this->delayedEnvelopesCollector->resetEnvelopes();
+    }
```
grossmannmartin, line 41: what about dispatching an event and handle this event in the `DispatchCollectedEnvelopesSubscriber`? that way we would have unified the place, where we handle/reset the collected envelopes

grossmannmartin (changes requested):

packages/frontend-api/src/Component/Messenger/GraphqlErrorResetDelayedEnvelopesSubscriber.php
vitek-rostislav: That seems nice, however, `frontend-api` vs `framework` package problem comes to ruin our lives once again :smile: I mean, `DispatchCollectedEnvelopesSubscriber` is in `framework`, `GraphqlErrorResetDelayedEnvelopesSubscriber` is in `frontend-api`, and if I wanted to create a "graphQL error occurred" event to dispatch here, it should also be in `frontend-api` package, which is the `framework` package not aware of. 

We could create a general "collected messages should be reset" event in the `framework` package that could be then dispatched from the `frontend-api` package, but I am not sure whether that would be better than the current approach :shrug: IMHO, an event should represent what happened, not what should happen, right?

packages/frontend-api/src/Component/HttpFoundation/GraphqlMutationErrorRollbackSubscriber.php
vitek-rostislav: Makes sense, I have removed that in a171c98db5, along with resetting `shouldRollbackTransaction` in `onKernelResponse`

packages/frontend-api/src/Component/HttpFoundation/GraphqlMutationErrorRollbackSubscriber.php
vitek-rostislav: added in a171c98db5aa91483f5e87356f8f5252e1e048da

packages/frontend-api/src/Component/Messenger/GraphqlErrorResetDelayedEnvelopesSubscriber.php
henzigo: Hm, are we talking about something like [this](https://github.com/shopsys/shopsys/pull/4017/changes/157976c9bb7ea69d8678822ab9f77e0bfc53137e)?

vitek-rostislav: Well, that is an antipattern in my opinion, it seems like reversed logic, listeners should be aware of events, not vice versa, I think :shrug: Lets discuss that tomorrow

vitek-rostislav: Btw, after asking Codex, whether `ManualRequestRollbackEvent` is a good practice, I got the following answer:
```
• Short answer: usually no.

  ManualRequestRollbackEvent is effectively a command disguised as an event (“listener,
  do rollback now”).
  That creates hidden coupling and makes flow harder to reason about.
```

vitek-rostislav: on the other hand, if we renamed `ManualRequestRollbackEvent` to something like `SilencedExceptionEvent`,  it could be a nice fit :+1: :slightly_smiling_face:

vitek-rostislav: We could get rid of `TransactionalMasterRequestListener::setTransactionManuallyRollbacked();` completely if we had such an event. I like that idea

grossmannmartin (approved):

PR #4490, opened 2026-02-27
[SSP-3083] Add blog article status

machacjan:

----
  - https://jm-ssp-3083-add-blog-article-status.odin.shopsys.cloud
  - https://cz.jm-ssp-3083-add-blog-article-status.odin.shopsys.cloud
  - https://jm-ssp-3083-add-blog-article-status.odin.shopsys.cloud/sk


packages/framework/src/Model/Blog/Article/Elasticsearch/FilterQuery.php
```diff
@@ -56,6 +57,20 @@ public function filterByCategory(BlogCategory $blogCategory): static
         return $clone;
     }
 
+    public function filterByPublishDateUpToNow(DateTimeInterface $now): static
```
vitek-rostislav: I would etither: 
1) change the method name so that it does not speak about "now" but rather a general date time (e.g. something like `filterByPublishDateUpTo(DateTimeInterface $dateTime)`
or, 
2) remove the method parameter and use `$this->clock->now()` directly inside the method

packages/framework/src/Model/Blog/Article/BlogArticle.php
```diff
@@ -298,13 +283,22 @@ protected function createTranslation(): BlogArticleTranslation
         return new BlogArticleTranslation();
     }
 
+    /**
+     * @return \Doctrine\Common\Collections\Collection<int, \Shopsys\FrameworkBundle\Model\Blog\Article\BlogArticleDomain>
+     */
+    public function getDomains()
```
vitek-rostislav, line 289: just a note to myself, maybe to discuss later when I understand the code better. 

I do not think we should expose the domains this way. The common practise is to use an entity as a proxy to a domain, not dealing with domain entities directly

packages/administration/templates/content/blog/article/statusBadge.html.twig
```diff
@@ -0,0 +1,28 @@
+{% if status == constant('Shopsys\\FrameworkBundle\\Model\\Blog\\Article\\BlogArticleStatusEnum::STATUS_DRAFT') %}
```
vitek-rostislav: I prefer `is same as` rather than non-strict `==` in the templates, however, it is a nitpick

packages/framework/src/Model/Blog/Article/BlogArticleGridFactory.php
```diff
@@ -21,15 +22,27 @@ public function __construct(
     ) {
     }
 
-    public function create(QueryBuilder $queryBuilder): Grid
+    public function create(QueryBuilder $queryBuilder, ?int $selectedDomainId = null): Grid
     {
+        if ($selectedDomainId !== null) {
```
vitek-rostislav, line 27: I know this was not explicit in the user story, but it would be nice to have that "status" column even if the domain tabs show "all domains". Lets get inspiration from the product list - could we maybe dynamically load that statusItems block after clicking on "preview" in the grid? Lets maybe brainstorm this on a quick call with PK

packages/administration/templates/content/blog/article/statusItems.html.twig
```diff
@@ -0,0 +1,42 @@
...
+        <div class="row align-items-center mb-3">
+
+            {% if isMultidomain() %}
+                <div class="col-auto">
+                    {{ domain_icon(domain.id) }}
+                </div>
+            {% endif %}
+
+            <div class="col">
+                {% if url is not null and (status == constant('Shopsys\\FrameworkBundle\\Model\\Blog\\Article\\BlogArticleStatusEnum::STATUS_PREVIEW') or (status == constant('Shopsys\\FrameworkBundle\\Model\\Blog\\Article\\BlogArticleStatusEnum::STATUS_PUBLISHED') and visible and (publishDate is null or publishDate <= date()))) %}
```
vitek-rostislav: we could extract the whole second part of the condition (starting with `and (status ==)...`) into a public function directly in the `BlogArticle` entity, what do you think? It could hel the readability and potential reusability IMHO

vitek-rostislav: I suggest something like `BlogArticle::isVisibleNow`. Btw, I think I will have another thoughts regarding what "visible" means for the blog articles through the codebase.... :smile:

packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
```diff
@@ -136,11 +139,28 @@ private function createSettingsGroup(
...
         $categoriesOptionsByDomainId = [];
+        $publishDatesOptionsByDomainId = [];
+        $statusesOptionsByDomainId = [];
 
         foreach ($this->domain->getAllIds() as $domainId) {
             $categoriesOptionsByDomainId[$domainId] = [
                 'domain_id' => $domainId,
             ];
+            $publishDatesOptionsByDomainId[$domainId] = [
+                'data' => $blogArticle?->getPublishDate($domainId),
```
vitek-rostislav: this should not be necessary, the data are set from the data object (e.g. `statuses` field works out of the box as well without setting the `data` option)

packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
```diff
@@ -160,24 +180,60 @@ private function createSettingsGroup(
...
         return $builderSettingsGroup;
     }
 
+    private function getArticlePreviewUrl(?BlogArticle $blogArticle, int $domainId): ?string
+    {
+        if ($blogArticle === null) {
+            return null;
+        }
+
+        $friendlyUrl = $this->friendlyUrlFacade->findMainFriendlyUrl($domainId, 'front_blogarticle_detail', $blogArticle->getId());
```
vitek-rostislav: you can use `getAbsoluteUrlByRouteNameAndEntityId` to avoid duplicate slug concatenation

vitek-rostislav: but the truth is, you will probably need to try-catch `FriendlyUrlNotFoundException`

packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
```diff
@@ -160,24 +180,60 @@ private function createSettingsGroup(
...
+        $previewDescription = t('Article is accessible via direct URL only, with noindex/nofollow.');
+
+        if ($previewUrl !== null) {
+            $previewDescription .= ' <a href="' . htmlspecialchars($previewUrl) . '" target="_blank">' . htmlspecialchars($previewUrl) . '</a>';
+        }
+
+        return [
+            BlogArticleStatusEnum::STATUS_DRAFT => t('Article is only visible in administration.'),
+            BlogArticleStatusEnum::STATUS_PREVIEW => $previewDescription,
+            BlogArticleStatusEnum::STATUS_PUBLISHED => t('Article is fully visible on storefront.'),
```
vitek-rostislav: I would maybe add something like "...when all relevant conditions are met". Not sure whether we should mention all the conditions (non-empty name, assignment into a visible category + publish date in the past), but it could help the administrators, lets maybe ask PK :shrug:

packages/framework/src/Migrations/Version20260227164539.php
```diff
@@ -0,0 +1,38 @@
...
+        ');
+        $this->sql('ALTER TABLE blog_article_domains ALTER COLUMN status SET NOT NULL');
+
+        $this->sql('ALTER TABLE blog_article_domains ADD publish_date TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL');
+        $this->sql('
+            UPDATE blog_article_domains bad
+            SET publish_date = ba.publish_date
+            FROM blog_articles ba
+            WHERE bad.blog_article_id = ba.id
+                AND ba.hidden = false
```
vitek-rostislav: is this wanted? I mean, f there is a hidden article with publish date set on a project, the publish date will get lost after the migration, right?

packages/framework/src/Model/Blog/Article/Elasticsearch/BlogArticleElasticsearchRepository.php
```diff
@@ -42,6 +48,8 @@ public function getAllBlogArticles(int $offset, int $limit, bool $onlyVisibleOnH
     {
         $filterQuery = $this->filterQueryFactory
             ->create($offset, $limit)
+            ->filterByStatus(BlogArticleStatusEnum::STATUS_PUBLISHED)
```
vitek-rostislav, line 49: I would appreciate a bunch of tests that would describe the scenarios. Either  in `BlogArticleElasticsearchRepositoryTest`, or directly in the API `BlogArticle(s)Test` - we should cover the scenarios like "preview article is accessible via direct link (getBySlug/uuid) only", "published article without publish date in the past is not visible at all", and maybe some other cases ("get articles by blog category retuns only truly visbile articles")? We can discuss that.

packages/framework/src/Model/Blog/Article/Elasticsearch/FilterQuery.php
vitek-rostislav: I prefer no. 2 :)

packages/framework/src/Model/Blog/Article/Elasticsearch/FilterQuery.php
```diff
@@ -56,6 +58,63 @@ public function filterByCategory(BlogCategory $blogCategory): static
...
+        $clone->filters[] = [
+            'term' => [
+                'status' => $status,
+            ],
+        ];
+
+        return $clone;
+    }
+
+    public function excludePublishedWithFutureDate(DateTimeInterface $now): static
```
vitek-rostislav: same as above, I do not think you need to pass "now" here - you can call clock->now() directly in the method

packages/framework/src/Model/Blog/Article/BlogArticleFacade.php
```diff
@@ -80,6 +92,49 @@ public function edit(int $blogArticleId, BlogArticleData $blogArticleData): Blog
         return $blogArticle;
     }
 
+    protected function applyStatusTransitions(
+        BlogArticle $blogArticle,
+        BlogArticleData $blogArticleData,
+    ): void {
+        foreach ($blogArticle->getDomains() as $blogArticleDomain) {
```
vitek-rostislav, line 105: I would love to talk about that state machine usage. On one hand, it could help with extension of the behavior on a project, hypothetically (but it would require better extension points for the form type - the statuses with their explanations are now wired inside `BlogArticleFormType`). On the other hand, it feels quite complex and we do not have a use case for that - I mean, currently, `$this->blogArticlePublishingStateMachine->can($blogArticleDomain, $transitionName)` never returns false, right?

vitek-rostislav: ok now I see `BlogArticlePublishingGuardSubscriber` which ensures name is filled in before changing status to "preview" (should we also check the category assignment there?). 

I noticed one problem - when tyou try to create an article without specifying the name in the "preview" status, the guard will not let you do that, presenting "Unable to change status for domain B2C - EN: Name must be defined.", however, the article is actually created along with the domains and translations as flush is called before `applyStatusTransitions` and because `BlogArticleStatusTransitionException` is then caught in the controller causing `TransactionalMasterRequest` mechnism is never triggered. You can solve this by firing `SilencedExceptionEvent` in the `catch` block in the controller.

project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx
```diff
@@ -45,30 +46,39 @@ const BlogArticleDetailPage: NextPage<ServerSidePropsType> = () => {
         return <Error404Content />;
     }
 
+    const isDraft = blogArticleData?.blogArticle?.status === 'draft';
+    const isPreview = blogArticleData?.blogArticle?.status === 'preview';
+    const isFuturePublishDate =
+        !!blogArticleData?.blogArticle?.publishDate && new Date(blogArticleData.blogArticle.publishDate) > new Date();
+    const shouldNoIndex = isDraft || isPreview || isFuturePublishDate;
```
Comment (vitek-rostislav) [inline project-base/storefront/pages/blogArticles/[blogArticleSlug].tsx:49]: draft/future publish date should never be present on the storefront right? I mean, I can live with this safety net, but maybe it is too complicated in vain

upgrade-notes/backend_20260228_093412.md
```diff
@@ -0,0 +1,19 @@
...
+- blog article status transitions are now managed by a Symfony Workflow state machine (`blog_article_publishing`)
+    - copy `project-base/app/config/packages/workflow.yaml` to your project to register the workflow
+    - the workflow defines three places (`draft`, `preview`, `published`) and three transitions (`publish`, `to_preview`, `to_draft`) on `Shopsys\FrameworkBundle\Model\Blog\Article\BlogArticleDomain`
+    - `Shopsys\FrameworkBundle\Model\Blog\Article\BlogArticlePublishingGuardSubscriber` guards the `publish` transition — it blocks publishing if the article name is not defined for the domain's locale
+    - if you need custom transition guards or additional statuses, extend the workflow configuration and/or subscribe to `workflow.blog_article_publishing.guard.*` events
+- published blog articles with a future `publishDate` are now accessible via direct URL but excluded from listings and sitemaps — the storefront should render `noindex, nofollow` for such articles
```
vitek-rostislav: actually, this should be a behavior of "preview" articles, right?

vitek-rostislav (changes requested):

packages/framework/src/Model/Blog/Article/BlogArticleGridFactory.php
machacjan: I would simply add the appropriate badge if _all_ domains of that blog article have the same status, otherwise show a gray badge simply labeled `Mixed`. Adding the same functionality as the _Preview_ button in edit page would, in my opinion, make the grid more cluttered.

packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
pk16011990: I prefer "Article is ready to be displayed on the storefront." It is not visible, but only ready and then it depends on other settings/fields.

project-base/app/tests/FrontendApiBundle/Functional/Blog/Article/BlogArticleVisibilityTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->assertNotContains($this->previewArticle->getUuid(), $uuids);
+    }
+
+    public function testDraftArticleIsNotAccessibleByUuid(): void
+    {
+        $response = $this->getResponseContentForQuery(
+            $this->getBlogArticleByUuidQuery($this->draftArticle->getUuid()),
+        );
+
+        $this->assertResponseContainsArrayOfErrors($response);
```
vitek-rostislav: I would keep this assertion, but moreover, I would appreciate another assertion for the particular error here (`BlogArticleNotFoundUserError`?). We could assert the message or code of the error

project-base/app/tests/FrontendApiBundle/Functional/Blog/Article/BlogArticleVisibilityTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->assertNotContains($this->publishedFutureArticle->getUuid(), $uuids);
+    }
+
+    public function testPublishedArticleWithFutureDateIsNotAccessibleByUuid(): void
+    {
+        $response = $this->getResponseContentForQuery(
+            $this->getBlogArticleByUuidQuery($this->publishedFutureArticle->getUuid()),
+        );
+
+        $this->assertResponseContainsArrayOfErrors($response);
```
vitek-rostislav: Same as above, I would keep this assertion, but moreover, I would appreciate another assertion for the particular error here (`BlogArticleNotFoundUserError`?). We could assert the message or code of the error. 

So you could even create a private method here, something like `assertArticleNotFoundError`

project-base/app/tests/FrontendApiBundle/Functional/Blog/Article/BlogArticleVisibilityTest.php
```diff
@@ -0,0 +1,111 @@
...
+                blogArticles {
+                    edges {
+                        node {
+                            uuid
+                        }
+                    }
+                }
+            }
+        ';
+    }
```
vitek-rostislav: please, use gql files instead of hardcoded strings 🙏

packages/framework/src/Form/Admin/Blog/BlogArticleFormType.php
vitek-rostislav: Moreover, let's extract this into `BlogArticleStatusEnum` class, please, so when they add a new status on a project, they do not have to extend the form type

packages/framework/src/Model/Blog/Article/BlogArticle.php
vitek-rostislav: I am now strongly thinking about removing the workflow completely. It somewhat forces us to stop using `BlogArticle` as the proxy for  `BlogArticleDomain` and use the domain entities directly instead. Moreover, even after [some simplifications](https://github.com/shopsys/shopsys/commit/1152df2c9815799c9b8abde3631ea6fa453dc162), it still feels quite complicated for what it does without any visible additional value. But I need to discuss this further with other colleagues...

I have also drafted the version without workflow - https://github.com/shopsys/shopsys/commit/dc5fa3b4c5aba87a80c7a13ea04150d0eaeacf14

I need some time to make the final decission :smile:

packages/framework/src/Model/Blog/Article/BlogArticleGridFactory.php
vitek-rostislav: All I am saying is that changing the grid columns based on the domains tab selection could be confusing for admins :slightly_smiling_face: I suggest you discuss this with the product owner :slightly_smiling_face:

vitek-rostislav (changes requested):
What I also noticed during some testing:
- when a blog article is created without name, its editation then fails on 500 error because of the missing name that the app needs for generating the link

Image

- when trying to create a new article with "preview" status without name, I got 500 error. It seems like the SilencedExceptionEvent
 should not be dispatched in newAction 🫠

packages/framework/src/Model/Blog/Article/BlogArticle.php
vitek-rostislav: ok, here is the _final_ (draft) solution that I suggest - https://github.com/shopsys/shopsys/commit/e47636dd021486fda78851dc56e77c706b6c7032

Basically, let's keep the workflow, but lets not re-define the "transition - places" relations in PHP again, yaml should be the only source of the truth. Hence, the WIP commit drops `BlogArticleTransitionEnum` usage completely. Instead, `TransitionNameByTargetPlaceResolver` is drafted - it resolves the transition name programmatically using the workflow component itself. It would be nice to add some unit tests for that, btw.

The `BlogArticleFacade::applyStatusTransitions()` is then simplified to try "get transition name" -> "apply transition" -> "catch `NotEnabledTransitionException` and get blocker list".

The commit will probably need some further polishing, let me know if you will need a call, we can discuss it :wink: 

Moreover, I moved the publish date setting to `BlogArticlePublishingGuardSubscriber` which should then probably be renamed to something like `BlogArticleStatus(Workflow)Subscriber` (the publish date check must have been dropped from the `guard.publish` as the feature was in contradiction)

packages/administration/templates/content/blog/article/statusItems.html.twig
```diff
@@ -0,0 +1,46 @@
+<p>
+    {% for domain in domains %}
+        {% set status = blogArticle.getStatus(domain.id) %}
+        {% set publishDate = blogArticle.getPublishDate(domain.id) %}
+        {% set visible = blogArticle.isVisible(domain.id) %}
+        {% set url = null %}
+
+        {% if blogArticle.name is not same as(null) %}
```
vitek-rostislav: I am afraid this will not be sufficient - invisible article will still be a problem (e.g. because of missing category). What about re-using `getArticlePreviewUrl` from `BlogArticleFormType`? Lets put it into soem common helper/facade, and use it in the form and in the template (e.g. using new twig extension?)

vitek-rostislav: ah, ok, invisible article should be probably covered by `findUrlByDomainId` itself. You can close this then

packages/framework/src/Form/BlogCategoriesType.php
```diff
@@ -31,6 +33,20 @@ public function buildView(FormView $view, FormInterface $form, array $options):
     public function buildForm(FormBuilderInterface $builder, array $options): void
     {
         $builder->addViewTransformer($this->blogCategoriesTypeTransformer);
+
+        $builder->addEventListener(FormEvents::PRE_SUBMIT, static function (FormEvent $event): void {
+            $submittedData = $event->getData() ?? [];
```
vitek-rostislav: why is this needed, please?

vitek-rostislav (approved):

packages/framework/src/Form/BlogCategoriesType.php
TomasLudvik: It is not

vitek-rostislav (approved):

pk16011990 (approved):

PR #4495, opened 2026-03-04
admin: menu search now highlights active item when navigating with keyboard

chlebektomas:
Fix search menu focus state.

----
  - https://tc-ssp-3920-fix-admin-focus.odin.shopsys.cloud
  - https://cz.tc-ssp-3920-fix-admin-focus.odin.shopsys.cloud
  - https://tc-ssp-3920-fix-admin-focus.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/assets/src/styles/custom.scss
+++ b/packages/administration/assets/src/styles/custom.scss
@@ -76,3 +76,7 @@
 .white-space-normal {
     white-space: normal !important;
 }
+
+.js-search-item.active {
+    background-color: var(--tblr-primary-lt);
+}


packages/administration/assets/src/styles/custom.scss
```diff
@@ -76,3 +76,7 @@
 .white-space-normal {
     white-space: normal !important;
 }
+
+.js-search-item.active {
```
grossmannmartin, line 80: `js-` prefixed classes shouldn't be used for styling, but purely for JS functionality. (And yes, I realized now, we broke this already several times)


Also, I wonder why the default  active state does not work - as seen in the preview https://preview.tabler.io/lists.html

grossmannmartin (changes requested):

packages/administration/assets/src/styles/custom.scss
chlebektomas: solved by setting up a new display correctly 🤣

PR #4501, opened 2026-03-05
Add to cart race condition

chlebektomas:
Fixes 500 errors on AddToCartMutation caused by a race condition when rapidly adding multiple products to cart. The add-to-cart button is now disabled during the request, and the backend uses a pessimistic lock to safely serialize concurrent cart modifications.  

----
  - https://tc-ssp-3909-add-to-cart-race-condition.odin.shopsys.cloud
  - https://cz.tc-ssp-3909-add-to-cart-race-condition.odin.shopsys.cloud
  - https://tc-ssp-3909-add-to-cart-race-condition.odin.shopsys.cloud/sk

----
  - https://tc-ssp-3909-add-to-cart-race-condition.odin.shopsys.cloud
  - https://cz.tc-ssp-3909-add-to-cart-race-condition.odin.shopsys.cloud
  - https://tc-ssp-3909-add-to-cart-race-condition.odin.shopsys.cloud/sk


JanMolcik (approved):
Reviewed, approved.

JanMolcik (changes requested):
I reproduced the broader product-list issue on the review preview via `playwright-cli`.

What I tested:

- reset browser state
- open homepage on the review preview
- click wishlist/comparison on product A and immediately on product B in the same product list

Observed result:

- `Wishlist`: only `1x AddProductToListMutation` was sent, only the first product switched to `Remove from wishlist`, the second one stayed `Add to wishlist`
- `Comparison`: only `1x AddProductToListMutation` was sent, only the first product switched to `Remove from comparison`, the second one stayed `Add to comparison`

So the current frontend guard does not only suppress repeated clicks on the same control, it also drops legitimate clicks on a different product while the first mutation is still in flight.

Playwright repro bundle for follow-up analysis:

- report + snapshot: https://gist.github.com/JanMolcik/8f73e4ded11ecfcb30f426ebb4f69850

Relevant code paths:

- `project-base/storefront/utils/productLists/useProductList.ts`
- `project-base/storefront/components/Blocks/Product/ProductsList/ProductsListContent.tsx`

JanMolcik (commented):
I have one low-severity recommendation after the latest review.

[P3] The new `useRefetchOnTabFocus()` bypasses the original `pause` guard for product list query hooks. In `project-base/storefront/utils/productLists/useProductList.ts` and `project-base/storefront/utils/productLists/useProductListCount.ts`, focus triggers `reexecuteQuery({ requestPolicy: 'network-only' })` unconditionally, while `urql` explicitly states that `reexecuteQuery` will execute even when `pause: true`. That means an anonymous user without a wishlist/comparison UUID may still send unnecessary `productList` / `productListCount` requests on tab focus, even though these hooks were intentionally paused.

I do not see this as a functional blocker or a high-severity issue. The main previous review finding is fixed, and the preview behavior now looks correct:
- the former page-wide wishlist/comparison lock is gone and has been replaced with a per-product guard in `project-base/storefront/utils/productLists/useProductList.ts`
- the global block remains only for the no-UUID create case, which makes sense for avoiding duplicate list creation
- the new Cypress coverage for wishlist/comparison race conditions matches this behavior
- on the review preview, wishlist and comparison now behave correctly after the first list creation, and the header badges update without reload
- cross-tab product-list refresh also worked correctly in manual Playwright verification

So this is a recommendation for cleanup/network efficiency, not a high-severity bug.

JanMolcik (approved):

pk16011990 (approved):

PR #4507, opened 2026-03-16
order relations that can influence order item price calculations are now stored during order creation

TomasLudvik:
- Having direct relation until now was causing problems when settings that influence price calculation changed over time. If order was saved after some time, it could cause changes in order that were not expected.

----
  - https://tl-freeze-order-prices-before-setting-change.odin.shopsys.cloud
  - https://cz.tl-freeze-order-prices-before-setting-change.odin.shopsys.cloud
  - https://tl-freeze-order-prices-before-setting-change.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Form/OrderItemsType.php
+++ b/packages/framework/src/Form/OrderItemsType.php
@@ -83,9 +83,10 @@ final class OrderItemsType extends AbstractType
         $view->vars['transportVatPercentsByTransportId'] = $this->transportFacade->getTransportVatPercentsByDomainIdIndexedByTransportId(
             $order->getDomainId(),
         );
-        $view->vars['paymentPricesWithVatByPaymentId'] = $this->paymentFacade->getPaymentPricesWithVatByCurrencyAndDomainIdIndexedByPaymentId(
-            $order->getCurrency(),
+        $view->vars['paymentPricesWithVatByPaymentId'] = $this->paymentFacade->getPaymentPricesWithVatByDomainIdIndexedByPaymentId(
             $order->getDomainId(),
+            $order->getCurrencyRoundingType(),
+            $order->getCurrencyRoundingPlacesPriceWithoutVat(),
         );
         $view->vars['paymentVatPercentsByPaymentId'] = $this->paymentFacade->getPaymentVatPercentsByDomainIdIndexedByPaymentId(
             $order->getDomainId(),
--- a/packages/framework/src/Migrations/Version20260310120000.php
+++ b/packages/framework/src/Migrations/Version20260310120000.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Override;
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20260310120000 extends AbstractMigration
+{
+    #[Override]
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE orders ADD currency_code VARCHAR(3) NOT NULL DEFAULT \'\'');
+        $this->sql('ALTER TABLE orders ADD currency_rounding_type VARCHAR(15) NOT NULL DEFAULT \'\'');
+        $this->sql('ALTER TABLE orders ADD currency_rounding_places_price_without_vat INT NOT NULL DEFAULT 2');
+        $this->sql('ALTER TABLE orders ADD currency_min_fraction_digits INT NOT NULL DEFAULT 2');
+        $this->sql('ALTER TABLE orders ADD payment_czk_rounding BOOLEAN NOT NULL DEFAULT FALSE');
+
+        $this->sql('
+            UPDATE orders o
+            SET
+                currency_code = c.code,
+                currency_rounding_type = c.rounding_type,
+                currency_rounding_places_price_without_vat = c.rounding_places_price_without_vat,
+                currency_min_fraction_digits = c.min_fraction_digits,
+                payment_czk_rounding = COALESCE(
+                    (SELECT p.czk_rounding
+                     FROM order_items oi
+                     JOIN payments p ON p.id = oi.payment_id
+                     WHERE oi.order_id = o.id AND oi.type = \'payment\'
+                     LIMIT 1),
+                    FALSE
+                )
+            FROM currencies c
+            WHERE c.id = o.currency_id
+        ');
+
+        $this->sql('ALTER TABLE orders ALTER COLUMN currency_code DROP DEFAULT');
+        $this->sql('ALTER TABLE orders ALTER COLUMN currency_rounding_type DROP DEFAULT');
+        $this->sql('ALTER TABLE orders ALTER COLUMN currency_rounding_places_price_without_vat DROP DEFAULT');
+        $this->sql('ALTER TABLE orders ALTER COLUMN currency_min_fraction_digits DROP DEFAULT');
+        $this->sql('ALTER TABLE orders ALTER COLUMN payment_czk_rounding DROP DEFAULT');
+
+        $this->sql('ALTER TABLE orders DROP CONSTRAINT IF EXISTS fk_e52ffdee38248176');
+        $this->sql('DROP INDEX IF EXISTS idx_e52ffdee38248176');
+        $this->sql('ALTER TABLE orders DROP COLUMN currency_id');
+
+        $this->sql('ALTER TABLE currencies ADD CONSTRAINT UNIQ_37C4469377153098 UNIQUE (code)');
+        $this->sql('CREATE INDEX idx_orders_currency_code ON orders (currency_code)');
+        $this->sql('ALTER TABLE orders ADD CONSTRAINT fk_orders_currency_code FOREIGN KEY (currency_code) REFERENCES currencies (code)');
+    }
+
+    #[Override]
+    public function down(Schema $schema): void
+    {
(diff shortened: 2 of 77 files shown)


TomasLudvik:
@coderabbitai

packages/framework/src/Form/OrderItemsType.php
```diff
@@ -83,8 +85,9 @@ public function buildView(FormView $view, FormInterface $form, array $options):
         $view->vars['transportVatPercentsByTransportId'] = $this->transportFacade->getTransportVatPercentsByDomainIdIndexedByTransportId(
             $order->getDomainId(),
         );
+        $currency = $this->currencyFacade->getByCode($order->getCurrencyCode());
         $view->vars['paymentPricesWithVatByPaymentId'] = $this->paymentFacade->getPaymentPricesWithVatByCurrencyAndDomainIdIndexedByPaymentId(
-            $order->getCurrency(),
+            $currency,
```
vitek-rostislav: we should pass the scalars into this method as well, right? The thing is - the currency with the stored code might not exist anymore (I know, edge case, but still...), and moreover, the currency setting might have changed so we should not rely on the "live" currency entity

packages/framework/src/Model/Order/OrderData.php
```diff
@@ -402,4 +422,15 @@ public function getPromoCodeDiscountPrice(): PriceInterface
     {
         return $this->getTotalPriceForItemTypes([OrderItemTypeEnum::TYPE_DISCOUNT])->inverse();
     }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Pricing\Currency\Currency $currency
+     */
+    public function fillCurrencyFieldsFromCurrency($currency): void
+    {
```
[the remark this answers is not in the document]
vitek-rostislav: I agree, this could be typed

packages/framework/src/Model/Order/OrderRepository.php
```diff
@@ -252,20 +250,19 @@ public function getCurrenciesUsedInOrders(): array
         return $this->em->createQueryBuilder()
             ->select('c')
             ->from(Currency::class, 'c')
-            ->join(Order::class, 'o', Join::WITH, 'o.currency = c.id')
```
vitek-rostislav, line 255: IMHO the whole feature that uses `getCurrenciesUsedInOrders` can be removed now, or? I mean, these currencies were not allowed for deletion because of their usage in orders, but that is not a blocker anymore :)

packages/framework/src/Model/Statistics/StatisticsRepository.php
```diff
@@ -129,7 +129,7 @@ public function getOrdersValueBetweenDates(DateTimeImmutable $startDateTime, Dat
         $query = $this->em->createNativeQuery(
             'SELECT SUM(o.total_price_with_vat * c.exchange_rate) AS total_price
             FROM orders o
-            JOIN currencies c ON o.currency_id = c.id
+            JOIN currencies c ON c.code = o.currency_code
```
vitek-rostislav, line 132: have you considered freezing the currency exchange rate as well? I do not think it is necessary, just need to ask :)

packages/framework/src/Migrations/Version20260310120000.php
```diff
@@ -0,0 +1,59 @@
...
+        $this->sql('ALTER TABLE orders ALTER COLUMN currency_min_fraction_digits DROP DEFAULT');
+        $this->sql('ALTER TABLE orders ALTER COLUMN payment_czk_rounding DROP DEFAULT');
+
+        $this->sql('ALTER TABLE orders DROP CONSTRAINT IF EXISTS fk_e52ffdee38248176');
+        $this->sql('DROP INDEX IF EXISTS idx_e52ffdee38248176');
+        $this->sql('ALTER TABLE orders DROP COLUMN currency_id');
+
+        $this->sql('ALTER TABLE currencies ADD CONSTRAINT UNIQ_37C4469377153098 UNIQUE (code)');
+        $this->sql('CREATE INDEX idx_orders_currency_code ON orders (currency_code)');
+        $this->sql('ALTER TABLE orders ADD CONSTRAINT fk_orders_currency_code FOREIGN KEY (currency_code) REFERENCES currencies (code)');
```
vitek-rostislav, line 53: why do we have this FK setting? It basically ensures a currency still can not be deleted when used in an order, right? I mean, it probably makes sense, I just wouldn't have thought of that :+1: So my previous comment about currency deletion is irelevant then, right? :smile:

packages/framework/src/Migrations/Version20260310120000.php
```diff
@@ -0,0 +1,59 @@
...
+    public function up(Schema $schema): void
+    {
+        $this->sql('ALTER TABLE orders ADD currency_code VARCHAR(3) NOT NULL DEFAULT \'\'');
+        $this->sql('ALTER TABLE orders ADD currency_rounding_type VARCHAR(15) NOT NULL DEFAULT \'\'');
+        $this->sql('ALTER TABLE orders ADD currency_rounding_places_price_without_vat INT NOT NULL DEFAULT 2');
+        $this->sql('ALTER TABLE orders ADD currency_min_fraction_digits INT NOT NULL DEFAULT 2');
+        $this->sql('ALTER TABLE orders ADD payment_czk_rounding BOOLEAN NOT NULL DEFAULT FALSE');
+
+        $this->sql('
+            UPDATE orders o
```
vitek-rostislav, line 23: I am quite concerned the migration could take a significant amount of time when there are high numbers of orders in DB, or? We should definitely mention the migration in the upgrade notes at least. You can also do merge the two update commands into one, Codex suggests :)

vitek-rostislav:
```
$this->sql('
      UPDATE orders o
      SET
          currency_code = c.code,
          currency_rounding_type = c.rounding_type,
          currency_rounding_places_price_without_vat =
  c.rounding_places_price_without_vat,
          currency_min_fraction_digits = c.min_fraction_digits,
          payment_czk_rounding = p.czk_rounding
      FROM currencies c, payments p
      WHERE c.id = o.currency_id
          AND p.id = o.payment_id
  ');
```

vitek-rostislav (changes requested):

vitek-rostislav:
Also, some interesting Codexes findings:

1)
```
Medium: the new payment_czk_rounding snapshot is not
    synchronized when payment is changed through the standard admin
    edit flow. The controller posts straight into edit() via
    OrderController.php:72 (packages/
    framework/src/Controller/Admin/OrderController.php:72) and
    OrderController.php:79 (packages/
    framework/src/Controller/Admin/OrderController.php:79). Inside
    the entity, Order.php:603 (packages/framework/src/Model/Order/Order.php:603) takes the
    newly selected payment, but Order.php:605 (packages/framework/src/Model/Order/
    Order.php:605) persists paymentCzkRounding from stale
    OrderData, which is preloaded from the old order. After a
    payment change, orders.payment_id and
    orders.payment_czk_rounding can immediately disagree. Suggested
    fix: derive paymentCzkRounding from $orderPaymentData->payment
    during editOrderPayment(), or refresh it in OrderFacade::edit()
    before calling $order->edit().
```
2) 
```
Medium: the migration makes orders.currency_code a foreign key
    to currencies.code, but currency code is still editable in
    admin with no guard. The FK is added at
    Version20260310120000.php:52 (
    project-base/app/src/Migrations/Version20260310120000.php:52),
    the form still exposes editable code at CurrencyFormType.php:49
    (packages/framework/src/Form/
    Admin/Pricing/Currency/CurrencyFormType.php:49), and
    CurrencyFacade.php:50 (packages/
    framework/src/Model/Pricing/Currency/CurrencyFacade.php:50)
    persists edits without checking whether orders already
    reference that code. In PostgreSQL that turns a previously
    valid currency-code edit into a flush-time FK failure.
    Suggested fix: either drop the FK on orders.currency_code, or
    explicitly forbid code changes once a currency is used by any
    order
```

vitek-rostislav:
Ah, and I just noticed, `Version20260310120000` should be in the framework package, not in `project-base`, right?

packages/framework/src/Migrations/Version20260310120000.php
TomasLudvik: Yeah, it was originally in one, but I have decided to merge them into one for better readability, but as you wish

packages/framework/src/Model/Statistics/StatisticsRepository.php
TomasLudvik: Well I do not think it is important in any way, as you are selling order in such currency and no exchange is made anytime

packages/framework/src/Migrations/Version20260310120000.php
TomasLudvik: It is weird situation, when you resolve your questions by yourself, but keep the comments open 😆

packages/framework/src/Model/Statistics/StatisticsRepository.php
vitek-rostislav: well actually, the exchange is made here, for the statistics. But I think it is ok as is, it actually still makes sense to present the total price based on the current exchange rate :shrug:

packages/framework/src/Migrations/Version20260310120000.php
vitek-rostislav: I need to add this instruction to my "code review alter ego" skill: Prompt: "ask a bunch of questions in the review, then answer them yourself so that the PR author is confused what to do" 😄

vitek-rostislav (approved):

pk16011990 (approved):

PR #4512, opened 2026-03-17
SSR and Error handling

chlebektomas:
Improve SSR and error handling.

----
  - https://tc-ssp-2260-ssr-redirect.odin.shopsys.cloud
  - https://cz.tc-ssp-2260-ssr-redirect.odin.shopsys.cloud
  - https://tc-ssp-2260-ssr-redirect.odin.shopsys.cloud/sk

----
  - https://tc-ssp-2260-ssr-redirect.odin.shopsys.cloud
  - https://cz.tc-ssp-2260-ssr-redirect.odin.shopsys.cloud
  - https://tc-ssp-2260-ssr-redirect.odin.shopsys.cloud/sk


JanMolcik:
I do not have any blocker findings, but I am leaving two non-blocking notes worth double-checking:

1. `prefetchLayoutQueries()` still branches on `isFullPageRequest()`, so layout queries (`Navigation`, `Settings`, `SeoPage`, footer articles, adverts...) are still not prefetched during CSR. I could not break this in the preview at runtime, but compared to the original Jira intent of "removing SSR/CSR divergence", it would be good to confirm that this is an intentional compromise and not just a renamed remainder of the previous behavior.

2. Changing the footer articles limit from `100` to `30` is a behavior change. I could not reproduce a problem with the demo data at runtime, but on projects with a larger number of footer articles this may start silently truncating content. If this is intentional, it would be helpful to call it out explicitly in the PR description or in the commit message.

Other than that, after the runtime pass through the Branch Review URL I do not have anything blocking.

JanMolcik (approved):

chlebektomas:
> I do not have any blocker findings, but I am leaving two non-blocking notes worth double-checking:
> 
> 1. `prefetchLayoutQueries()` still branches on `isFullPageRequest()`, so layout queries (`Navigation`, `Settings`, `SeoPage`, footer articles, adverts...) are still not prefetched during CSR. I could not break this in the preview at runtime, but compared to the original Jira intent of "removing SSR/CSR divergence", it would be good to confirm that this is an intentional compromise and not just a renamed remainder of the previous behavior.
> 2. Changing the footer articles limit from `100` to `30` is a behavior change. I could not reproduce a problem with the demo data at runtime, but on projects with a larger number of footer articles this may start silently truncating content. If this is intentional, it would be helpful to call it out explicitly in the PR description or in the commit message.
> 
> Other than that, after the runtime pass through the Branch Review URL I do not have anything blocking.


1. `isFullPageRequest()` branch in `prefetchLayoutQueries()` is intentional.
We kept this behavior on purpose: for `/_next/data` CSR navigations we do not server-prefetch layout queries (Navigation/Settings/SeoPage/footer articles/adverts), to avoid increasing server work/TTFB on client transitions.
Scope of this PR was parallel SSR + error-handling consistency, not changing CSR prefetch strategy.
So this is an intentional compromise, not an accidental rename of previous behavior.

2. Footer articles `first` limit:
Agreed this would be a behavior change if reduced. We are keeping `first: 100` (same as current behavior), so no intentional truncation change is introduced by this PR.

Also, could you please do one more quick CR pass over the latest 3 commits? We added follow-up adjustments there and want to confirm everything still looks correct.

project-base/storefront/utils/auth/isUserLoggedInSSR.ts
```diff
@@ -1,16 +1,28 @@
...
+} from 'graphql/requests/customer/queries/CurrentCustomerUserAuthQuery.generated';
+import {
     CurrentCustomerUserQueryDocument,
+    TypeCurrentCustomerUserQuery,
 } from 'graphql/requests/customer/queries/CurrentCustomerUserQuery.generated';
 import { Client } from 'urql';
 
 export const isUserLoggedInSSR = (currentClient: Client): boolean => {
-    const customerQueryResult = currentClient.readQuery<TypeCurrentCustomerUserQuery>(
+    const authCustomerQueryResult = currentClient.readQuery<TypeCurrentCustomerUserAuthQuery>(
```
JanMolcik, line 12: This helper is now a pure cache read. On a fresh SSR client that means it returns `false` until either `CurrentCustomerUserAuthQuery` or `CurrentCustomerUserQuery` has already been prefetched. That is fine inside `buildServerSideProps()`, but it regresses earlier call sites that use it before `initServerSideProps()` fills the cache.

project-base/storefront/pages/customer/order-detail.tsx
```diff
@@ -122,6 +122,7 @@ export const getServerSideProps = getServerSidePropsWrapper(
                 client,
                 ssrExchange,
                 domainConfig,
+                currentCustomerUserPrefetchMode: 'full',
```
JanMolcik: `currentCustomerUserPrefetchMode: 'full'` is useful here, but it only takes effect inside `initServerSideProps()`. The `isUserLoggedInSSR(client)` check above runs before that prefetch happens, so on a cold SSR client it always behaves as "not logged in" and skips the intended SSR prefetch of `OrderDetailQuery` / `OrderAvailablePaymentsQuery`. Could we either prefetch the lightweight auth query before that branch or make `isUserLoggedInSSR()` fetch when the cache is cold?

JanMolcik (approved):

PR #4514, opened 2026-03-18
Improve work with phone numbers

grossmannmartin:
  Phone numbers across the platform were previously stored as plain strings without any structure. This made it impossible to reliably identify country prefixes, validate numbers per country, or filter by prefix.

  This PR introduces international phone prefix support. Administrators can now configure allowed phone prefixes per domain (with a default), and all phone inputs — admin forms, frontend API, and storefront — decompose phone numbers into three parts: country code (ISO alpha-2), prefix (dial code like `+420`), and national number.

  **Key changes:**

  - **New `PhoneData` value object** — universal transfer format for phone data across all layers (country code, prefix, number)
  - **Domain-aware phone prefix settings** — each domain can configure its own set of allowed prefixes and a default; admin dashboard warns when prefixes are not configured
  - **New `PhoneType` form type** — compound Symfony form with prefix selector and number input, supports legacy/unknown prefixes for migrated data
  - **Entity changes** — `Order`, `DeliveryAddress`, and `CustomerUser` now store phone as 3 separate columns (`telephone_prefix`, `telephone_prefix_country_code`, `telephone_number`)
  - **Database migrations** — schema migration splits existing `telephone` column into 3 columns; data migration parses existing numbers using `libphonenumber` with country-aware context
  - **Frontend API (GraphQL)** — new `PhoneData` output type and `PhoneDataInput` input type; `phonePrefixes` query returns enabled prefixes for a domain
  - **Storefront** — phone prefix selector component integrated into all forms (registration, order, profile, complaints, inquiries, delivery addresses)
  - **Enhancements** — suspicious phone numbers (with unknown/mismatched prefixes) are visually marked in admin; phone strings are formatted according to country conventions

#### Fixes issues
... 

----
  - https://mg-phone-prefix.odin.shopsys.cloud
  - https://cz.mg-phone-prefix.odin.shopsys.cloud
  - https://mg-phone-prefix.odin.shopsys.cloud/sk


grossmannmartin:
@coderabbitai review

packages/framework/src/Form/Admin/PhonePrefix/PhonePrefixSettingsFormType.php
```diff
@@ -0,0 +1,141 @@
...
+            ->add('actionBar', ActionBarType::class, [
+                'save_label' => t('Save changes'),
+            ]);
+    }
+
+    #[Override]
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setRequired('domain_id')
```
vitek-rostislav, line 88: just an idea, what about making the form multidomain? Would not that be better ux?

packages/framework/src/Form/Admin/PhonePrefix/PhonePrefixSettingsFormType.php
```diff
@@ -0,0 +1,141 @@
...
+
+    public function validateDefaultPrefixIsEnabled(
+        PhonePrefixSettingsData $phonePrefixSettingsData,
+        ExecutionContextInterface $executionContext,
+    ): void {
+        if ($phonePrefixSettingsData->defaultCode !== null
+            && !in_array($phonePrefixSettingsData->defaultCode, $phonePrefixSettingsData->enabledCodes, true)
+        ) {
+            $executionContext->buildViolation('Default phone prefix must be enabled')
+                ->atPath('defaultCode')
```
vitek-rostislav, line 107: translation is missing

packages/framework/src/Form/Admin/PhonePrefix/PhonePrefixSettingsFormType.php
```diff
@@ -0,0 +1,141 @@
...
+                ->addViolation();
+        }
+    }
+
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\PhonePrefix\CountryDialCode[] $countryDialCodes
+     * @return \Shopsys\FrameworkBundle\Model\PhonePrefix\CountryDialCode[]
+     */
+    public function getPreferredDialCodesByCountriesOnDomain(int $domainId, array $countryDialCodes): array
+    {
```
vitek-rostislav, line 117: could be protected

packages/framework/src/Model/PhonePrefix/Settings/PhonePrefix.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    #[ORM\Column(type: 'string', length: 2)]
+    protected $code;
+
+    /**
+     * @var bool
+     */
+    #[ORM\Column(type: 'boolean')]
+    protected $isDefault;
+
```
vitek-rostislav, line 39: would it make sense to create index here?

packages/framework/src/Model/PhonePrefix/Settings/PhonePrefixSettingsFacade.php
```diff
@@ -0,0 +1,46 @@
...
+        protected readonly EntityManagerInterface $em,
+        protected readonly PhonePrefixRepository $phonePrefixRepository,
+        protected readonly PhonePrefixFactory $phonePrefixFactory,
+    ) {
+    }
+
+    public function getByDomainId(int $domainId): PhonePrefixSettingsData
+    {
+        $phonePrefixSettingsData = new PhonePrefixSettingsData();
+
```
vitek-rostislav, line 22: should not we use data factory for this?

project-base/app/src/DataFixtures/Demo/PhonePrefixDataFixture.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    #[Override]
+    public function load(ObjectManager $manager): void
+    {
+        foreach ($this->domain->getAll() as $domainConfig) {
+            $enabledCountryCodes = $this->getEnabledCountriesOnDomain($domainConfig->getId());
+
+            $phonePrefixSettingsData = new PhonePrefixSettingsData();
+            $phonePrefixSettingsData->enabledCodes = $enabledCountryCodes;
```
vitek-rostislav, line 32: factory?

packages/framework/src/Twig/RequiredSettingExtension.php
```diff
@@ -148,6 +151,23 @@ protected function checkAtLeastOneCountryExists(): void
...
     }
 
+    protected function checkPhonePrefixIsConfigured(): void
+    {
+        $notConfiguredDomainIds = $this->phonePrefixSettingsFacade->filterOutConfiguredDomainIds($this->domain->getAdminEnabledDomainIds());
+
+        foreach ($notConfiguredDomainIds as $domainId) {
+            $domainConfig = $this->domain->getDomainConfigById($domainId);
+
+            $this->requiredSettingsMessages[] = t(
```
vitek-rostislav, line 161: maybe this could be just one message with a list of domains?

vitek-rostislav: but I see this is a common pattern, so :shrug:

packages/framework/src/Model/PhonePrefix/CountryDialCodeProvider.php
```diff
@@ -0,0 +1,84 @@
...
+            return null;
+        }
+
+        $phoneUtil = PhoneNumberUtil::getInstance();
+        $countryCallingCode = ltrim($dialCode, '+');
+
+        $countryCode = $phoneUtil->getRegionCodeForCountryCode((int)$countryCallingCode);
+
+        return $countryCode !== 'ZZ' ? $countryCode : null;
+    }
```
vitek-rostislav, line 73: we can use the constant for "ZZ"

packages/framework/src/Form/Admin/Order/OrderFormType.php
```diff
@@ -319,15 +320,10 @@ private function createPersonalDataGroup(FormBuilderInterface $builder): FormBui
                     ),
                 ],
             ])
-            ->add('telephone', TextType::class, [
+            ->add('telephone', PhoneType::class, [
                 'label' => 'Telephone',
```
vitek-rostislav, line 324: is `TelType` validated for length?

packages/framework/src/Migrations/Version20260312111907.php
```diff
@@ -0,0 +1,147 @@
...
+namespace Shopsys\FrameworkBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use libphonenumber\NumberParseException;
+use libphonenumber\PhoneNumberUtil;
+use Override;
+use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
+
+class Version20260312111907 extends AbstractMigration
+{
```
vitek-rostislav, line 14: we should add warning about this migration to the upgrade notes as it might take some time on big table, right?

packages/framework/src/Model/Order/Order.php
```diff
@@ -149,11 +150,23 @@ class Order
     #[ORM\Column(type: 'string', length: 255)]
     protected $email;
 
+    /**
+     * @var string|null
+     */
+    #[ORM\Column(type: 'string', length: 10, nullable: true)]
+    protected $telephonePrefix;
+
```
vitek-rostislav, line 157: JG had this idea about creating custom embed for the field, which might be interesting

project-base/app/tests/FrontendApiBundle/Functional/Order/OrderWithPersonalPickupStoreTest.php
```diff
@@ -51,42 +46,29 @@ public function testCreateOrderWithPersonalPickupStore(): void
...
+                'street' => '123 Fake Street',
+                'city' => 'Springfield',
+                'postcode' => '12345',
+                'country' => 'CZ',
+                'isDeliveryAddressDifferentFromBilling' => false,
+            ],
+        );
 
-    private function getMutation(string $cartUuid): string
-    {
```
vitek-rostislav, line 58: love to see these refactors to gql files!

packages/framework/src/Model/SalesRepresentative/SalesRepresentativeGridFactory.php
```diff
@@ -27,6 +27,7 @@ public function __construct(
     public function create(?string $roleConstant): Grid
     {
         $queryBuilder = $this->salesRepresentativeFacade->getAllQueryBuilder();
+        $queryBuilder->addSelect('CONCAT(COALESCE(sr.telephonePrefix, \'\'), sr.telephoneNumber) as telephone');
         $dataSource = $this->queryBuilderDataSourceFactory->create($queryBuilder, 'sr.id');
```
vitek-rostislav, line 32: as I saw this concatenation multiple times, maybe it would be worth creating some helper SQL function fo that?

project-base/app/tests/FrontendApiBundle/Functional/Customer/User/CurrentCustomerUserTest.php
```diff
@@ -81,7 +82,7 @@ public function testCurrentCustomerUser(): void
                 'firstName' => 'Jan',
                 'lastName' => 'Dvořák',
                 'email' => 'no-reply99@shopsys.com',
-                'telephone' => '585425321',
+                'telephone' => '+420585425321',
             ],
```
vitek-rostislav, line 106: should we cover the phone data field as well? (maybe in multiple other tests as well)

packages/frontend-api/src/Model/Resolver/Settings/PhonePrefixesQuery.php
```diff
@@ -0,0 +1,41 @@
...
+    public function phonePrefixesQuery(): array
+    {
+        $locale = $this->domain->getLocale();
+
+        $countryDialCodes = $this->countryDialCodeProvider->getAllEnabledOnDomainWithDefaultFirst($this->domain->getId());
+
+        return array_map(
+            static fn (CountryDialCode $countryDialCode): array => [
+                'code' => $countryDialCode->code,
+                'dialCode' => $countryDialCode->dialCode,
```
vitek-rostislav, line 34: I would prefefr a data object rather than an array but is rather a matter of taste, right?

vitek-rostislav (changes requested):

packages/framework/src/Form/Admin/Order/OrderFormType.php
vitek-rostislav: now I know it is :slightly_smiling_face:

packages/framework/src/Model/Order/Order.php
grossmannmartin: Embed is... politely speaking, a big pile of goo. 

Nullability cannot be implemented reliably. Meaning storing `$phoneData = null` will get hydrated back as `new PhoneData(null, null, null)`. Mandatoriness on the database level cannot be enforced.

packages/framework/src/Form/Admin/PhonePrefix/PhonePrefixSettingsFormType.php
grossmannmartin: nice, thanks - c07822baf5

packages/framework/src/Form/Admin/PhonePrefix/PhonePrefixSettingsFormType.php
grossmannmartin: made private as the class is final 2185265586

packages/administration/src/Component/Security/Role/CoreAdminRoleProvider.php
```diff
@@ -142,6 +142,7 @@ protected function getSettingsAndConfigurationRoles(): array
             new Role(AdminRoleConstant::ROLE_FREE_TRANSPORT_AND_PAYMENT, t('Free transport and payment'), [Permission::EDIT]),
             new Role(AdminRoleConstant::ROLE_TRANSPORT_AND_PAYMENT, t('Transports and payments'), [Permission::FULL]),
             new Role(AdminRoleConstant::ROLE_COUNTRY, t('Countries'), [Permission::CREATE, Permission::EDIT]),
+            new Role(AdminRoleConstant::ROLE_PHONE_PREFIX, t('Phone prefixes'), [Permission::FULL]),
```
henzigo: You defined `Permission::FULL`, but only `Permission::VIEW` and `Permission::EDIT` are used.

packages/framework/src/Model/PhonePrefix/Settings/PhonePrefix.php
grossmannmartin: There is already the partial index `phone_prefixes_domain_default` that should cover the queries made. So it seems like another index on isDefault only would not be helpful, as it's not queried solely by this column.

packages/framework/src/Model/Order/Order.php
henzigo: So if we cannot use the Embedded type, we should extract this into a separate table. What do you think?

packages/framework/src/Model/PhonePrefix/Settings/PhonePrefixSettingsFacade.php
grossmannmartin: done 5f2f18ef0d

project-base/app/src/DataFixtures/Demo/PhonePrefixDataFixture.php
grossmannmartin: done 5f2f18ef0d

packages/framework/src/Twig/RequiredSettingExtension.php
grossmannmartin: Let's focus on this in a separate issue, please.

packages/framework/src/Model/PhonePrefix/CountryDialCodeProvider.php
grossmannmartin: reworked in 78c04f7c00

packages/framework/src/Migrations/Version20260312111907.php
grossmannmartin: I have added some relevant warnings in 24448fe62c

project-base/app/tests/FrontendApiBundle/Functional/Customer/User/CurrentCustomerUserTest.php
grossmannmartin: covered by 281cbbef59, good idea 👍

packages/frontend-api/src/Model/Resolver/Settings/PhonePrefixesQuery.php
grossmannmartin: it is 🤷‍♂️ if you insist, I can add the object, but I don't have the need, tbh

packages/framework/src/Model/SalesRepresentative/SalesRepresentativeGridFactory.php
grossmannmartin: fixed (along with fixes caused by wrong rebase) in c8f0238d27

packages/framework/src/Model/Order/Order.php
vitek-rostislav: it would be 1:1relation :shrug:

vitek-rostislav (approved):

JanMolcik (changes requested):
Storefront runtime looks OK. I could not reproduce an infinite re-render loop in `PhonePrefixSelect` on the review, but I still cannot approve this from FE side because the test contract is not finished: Cypress helpers/fixtures still send `telephone` as a string instead of `PhoneDataInput`, so auth/order CI jobs are failing. Details are in inline comments.

project-base/storefront/components/providers/PersistStoreProvider.tsx
```diff
@@ -87,6 +88,26 @@ const createPersistStore = (domainId: number) => {
                         };
                     }
 
+                    // Migration from version 1 to version 2
+                    if (version < 2) {
+                        migratedPersistedState = {
+                            ...migratedPersistedState,
+                            ...defaultContactInformationState,
```
JanMolcik: This migration resets the whole `contactInformation` object, so users can lose a partially filled checkout form. Could this merge only the new phone fields, similar to the follow-up migration below?

project-base/storefront/components/Forms/PhonePrefixSelect/PhonePrefixSelect.tsx
```diff
@@ -0,0 +1,188 @@
...
+        prefixByCountryCode.get(countryCode) ??
+        normalizedPrefixes.find((phonePrefix) => phonePrefix.dialCode === normalizedDialCode) ??
+        (shouldBeReadonly ? normalizedPrefixes[0] : undefined);
+    const activeOption = activePhonePrefix
+        ? (selectOptions.find((option) => option.value === activePhonePrefix.code) ?? null)
+        : null;
+    const activePrefixLabel = activePhonePrefix ? getPhonePrefixDisplayLabel(activePhonePrefix) : undefined;
+
+    // Keep the hidden form fields aligned with the displayed selection.
+    useEffect(() => {
```
JanMolcik, line 103: I could not reproduce an infinite loop at runtime, but this sync logic plus the two effects below is still quite fragile. Could this be simplified so the select has less intertwined state-sync behavior?

project-base/storefront/graphql/requests/orders/mutations/CreateOrderMutation.graphql
```diff
@@ -2,7 +2,7 @@ mutation CreateOrderMutation(
     $firstName: String!
     $lastName: String!
     $email: String!
-    $telephone: String!
+    $telephone: PhoneDataInput!
```
JanMolcik, line 5: The new `PhoneDataInput` contract is correct, but FE Cypress helpers/fixtures need to be updated too. They still send `telephone` as a string, which is why auth/order CI jobs are currently failing.

JanMolcik (approved):

packages/framework/src/Model/PhonePrefix/PhoneData.php
```diff
@@ -0,0 +1,50 @@
...
+            prefix: $data['prefix'],
+            number: $data['number'],
+        );
+    }
+
+    public function toLibPhoneNumberObject(): libphonenumber\PhoneNumber
+    {
+        return new libphonenumber\PhoneNumber()
+            ->setCountryCode((int)($this->prefix ? ltrim($this->prefix, '+') : 0))
+            ->setNationalNumber($this->number);
```
pk16011990: `$this->number` is nullable and it works only because setNationalNumber is not strict...

project-base/storefront/components/Forms/PhonePrefixSelect/PhoneNumberInputControlled.tsx
```diff
@@ -0,0 +1,211 @@
...
+        fieldState: { error: telephoneError },
+    } = useController({ name: telephoneName, control: formProviderMethods.control });
+
+    const hasError = Boolean(prefixError) || Boolean(telephoneError);
+    const activeError = telephoneError ?? prefixError;
+    const telephoneInputId = formName + '-' + telephoneName;
+    const phoneInputErrorMessageId = `${telephoneInputId}-error`;
+    const prefixTid = formName + '-' + prefixName;
+
+    const onPrefixSelectChange = useCallback(
```
pk16011990, line 63: Wrap in useCallback to prevent unnecessary re-renders in PhonePrefixSelect.tsx useEffect (line 102-112).
Current implementation causes the effect to re-run on every parent render because onChange is not a stable reference.

packages/framework/src/Form/Constraints/PhoneNumberValidator.php
```diff
@@ -0,0 +1,55 @@
...
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+
+class PhoneNumberValidator extends ConstraintValidator
+{
+    protected const int MAX_NUMBER_LENGTH = 30;
+
+    #[Override]
+    public function validate(mixed $value, Constraint $constraint): void
```
pk16011990, line 18: `prefix` and `countryCode` do not need a length validation?

packages/framework/src/Model/PhonePrefix/Settings/PhonePrefix.php
```diff
@@ -0,0 +1,78 @@
...
+     * @var int
+     */
+    #[ORM\Column(type: 'integer')]
+    protected $domainId;
+
+    /**
+     * @var string
+     */
+    #[ORM\Column(type: 'string', length: 2)]
+    protected $code;
```
pk16011990, line 32: Idea: Is it Country Code? Better naming or foreign key to `countries.code`?

packages/framework/src/Model/PhonePrefix/PhoneData.php
```diff
@@ -0,0 +1,50 @@
...
+final class PhoneData
+{
+    public function __construct(
+        public ?string $countryCode = null,
+        public ?string $prefix = null,
+        public ?string $number = null,
+    ) {
+    }
+
+    public function toPhoneNumber(): ?string
```
pk16011990: Idea for a more universal solution

```suggestion
    public function toPhoneNumber(int $format = PhoneNumberFormat::INTERNATIONAL): ?string
```

project-base/storefront/utils/phonePrefix/usePhonePrefixes.ts
```diff
@@ -0,0 +1,18 @@
...
+    usePhonePrefixesQuery,
+} from 'graphql/requests/phonePrefixes/queries/PhonePrefixesQuery.generated';
+
+export type PhonePrefixesType = NonNullable<TypePhonePrefixesQuery['settings']>['phonePrefixes'];
+const EMPTY_PHONE_PREFIXES: PhonePrefixesType = [];
+
+export const usePhonePrefixes = () => {
+    const [{ data: phonePrefixesData }] = usePhonePrefixesQuery();
+    const phonePrefixes = phonePrefixesData?.settings?.phonePrefixes ?? EMPTY_PHONE_PREFIXES;
+    const defaultPhonePrefix = phonePrefixes.at(0);
```
pk16011990: To prevent undefined

```suggestion
    const defaultPhonePrefix = phonePrefixes[0] ?? null;
```

upgrade-notes/backend_20260319_135122.md
```diff
@@ -0,0 +1,18 @@
+#### Improve work with phone numbers ([#4514](https://github.com/shopsys/shopsys/pull/4514))
+
+- phone numbers across all entities are now stored as structured data using `Shopsys\FrameworkBundle\Model\PhonePrefix\PhoneData` instead of plain strings
+    - the `$telephone` property on `OrderData`, `CustomerUserData`, `DeliveryAddressData`, `ComplaintData` (`$deliveryTelephone`), `InquiryData`, `SalesRepresentativeData`, `WithdrawalRequestData`, and `RegistrationData` now expects `PhoneData` instead of `string`
+    - if you set telephone data directly on these data objects, use `new PhoneData($countryCode, $prefix, $number)` instead of a plain string
+    - existing `getTelephone()` methods on entities still return a formatted `string`, use the new `getTelephoneData()` methods to get the structured `PhoneData` object
```
pk16011990: It can be tricky. Numbers are formatted using `PhoneNumberFormat::INTERNATIONAL`, which means, e.g., `+420 111 222 333`. If the project had no spaces before at all, it would be a problem in some cases.

pk16011990 (changes requested):

packages/framework/src/Form/Constraints/PhoneNumberValidator.php
grossmannmartin: As it's used in the form type where the data comes from the choice type, I would say they don't. 
Validity is ensured by the internal working of the choice type when an invalid value is denied.

packages/framework/src/Model/PhonePrefix/Settings/PhonePrefix.php
grossmannmartin: It's possible that the code will not match any stored country (old data, migration from ERP, ...). So I would not use a foreign key here. 
Based on this, I didn't want it to be confused with the countries of the e-shop, hence the name. This value is used merely for distinguishing the same prefixes for several regions, so I think it's acceptable.
But I am open to discussion.

packages/framework/src/Model/PhonePrefix/PhoneData.php
grossmannmartin: good idea 👍

upgrade-notes/backend_20260319_135122.md
grossmannmartin: I see. Honestly, I think the only thing to do here is a tweak of the upgrade notes a bit. 
What do you think? c3ad92a101

pk16011990 (approved):

PR #4522, opened 2026-03-23
added new rounding to 0.05 for order total price

TomasLudvik:
- In Slovakia there is need to round order prices to five cents for cash payment, we already had setting (czkRounding) for whole numbers
- This was reworked to ommit information about for what currency it is used and made more generally
- New solution is prepared for other possible needs in the future

----
  - https://tl-rounding-to-fives.odin.shopsys.cloud
  - https://cz.tl-rounding-to-fives.odin.shopsys.cloud
  - https://tl-rounding-to-fives.odin.shopsys.cloud/sk

----
  - https://tl-rounding-to-fives.odin.shopsys.cloud
  - https://cz.tl-rounding-to-fives.odin.shopsys.cloud
  - https://tl-rounding-to-fives.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Form/Admin/Payment/PaymentFormType.php
+++ b/packages/framework/src/Form/Admin/Payment/PaymentFormType.php
@@ -20,6 +20,7 @@ use Shopsys\FrameworkBundle\Form\MessageType;
 use Shopsys\FrameworkBundle\Form\PriceAndVatTableByDomainsType;
 use Shopsys\FrameworkBundle\Model\GoPay\PaymentMethod\GoPayPaymentMethod;
 use Shopsys\FrameworkBundle\Model\GoPay\PaymentMethod\GoPayPaymentMethodFacade;
+use Shopsys\FrameworkBundle\Model\Payment\OrderRoundingTypeEnum;
 use Shopsys\FrameworkBundle\Model\Payment\Payment;
 use Shopsys\FrameworkBundle\Model\Payment\PaymentData;
 use Shopsys\FrameworkBundle\Model\Payment\PaymentFacade;
@@ -47,6 +48,7 @@ final class PaymentFormType extends AbstractType
         private readonly Domain $domain,
         private readonly PaymentTypeProvider $paymentTypeProvider,
         private readonly PaymentInstructionFacade $paymentInstructionFacade,
+        private readonly OrderRoundingTypeEnum $orderRoundingTypeEnum,
     ) {
     }
 
@@ -163,9 +165,18 @@ final class PaymentFormType extends AbstractType
         ]);
 
         $builderPriceGroup
-            ->add('czkRounding', YesNoType::class, [
-                'label' => 'Order in CZK round to whole crowns',
-                'help' => t('Rounding item with 0 % VAT will be added to your order. It is used for payment in cash.'),
+            ->add('orderRoundingTypeByDomainId', MultidomainType::class, [
+                'entry_type' => ChoiceType::class,
+                'entry_options' => [
+                    'choices' => $this->orderRoundingTypeEnum->getAllIndexedByTranslations(),
+                ],
+                'display_mode' => 'columns',
+                'row_attr' => [
+                    'class' => 'mb-3',
+                ],
+                'label' => 'Order total rounding',
+                'required' => true,
+                'help' => t('Rounding item with 0 % VAT will be added to the order. It is used for payment in cash.'),
             ])
             ->add('pricesByDomains', PriceAndVatTableByDomainsType::class, [
                 'pricesIndexedByDomainId' => $this->paymentFacade->getPricesIndexedByDomainId($payment),
--- a/packages/framework/src/Model/Order/OrderPriceCalculation.php
+++ b/packages/framework/src/Model/Order/OrderPriceCalculation.php
@@ -6,7 +6,8 @@ namespace Shopsys\FrameworkBundle\Model\Order;
 
 use Shopsys\FrameworkBundle\Component\Money\Money;
 use Shopsys\FrameworkBundle\Model\Order\Item\OrderItemPriceCalculation;
-use Shopsys\FrameworkBundle\Model\Pricing\Currency\Currency;
+use Shopsys\FrameworkBundle\Model\Payment\OrderRoundingTypeEnum;
+use Shopsys\FrameworkBundle\Model\Payment\Payment;
 use Shopsys\FrameworkBundle\Model\Pricing\Price;
 use Shopsys\FrameworkBundle\Model\Pricing\PriceInterface;
 use Shopsys\FrameworkBundle\Model\Pricing\Rounding;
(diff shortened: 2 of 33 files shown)


packages/framework/src/Form/Admin/Payment/PaymentFormType.php
```diff
@@ -163,9 +165,11 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
 
         $builderPriceGroup
-            ->add('czkRounding', YesNoType::class, [
-                'label' => 'Order in CZK round to whole crowns',
-                'help' => t('Rounding item with 0 % VAT will be added to your order. It is used for payment in cash.'),
+            ->add('orderRoundingType', ChoiceType::class, [
+                'required' => true,
+                'choices' => $this->orderRoundingTypeEnum->getAllIndexedByTranslations(),
+                'label' => 'Order total rounding',
+                'help' => t('Rounding item with 0 % VAT will be added to the order. It is used for payment in cash. The final rounding also depends on the currency rounding settings.'),
```
grossmannmartin: Don't forget to dump translations, please

grossmannmartin: `The final rounding also depends on the currency rounding settings.`

Does it? I thought currency rounding only affects the price going into the calculation, not the rounding item itself.

project-base/app/src/DataFixtures/Demo/PaymentDataFixture.php
```diff
@@ -94,19 +98,48 @@ public function load(ObjectManager $manager): void
...
+        $paymentData = $this->paymentDataFactory->create();
+        $paymentData->type = PaymentTypeEnum::TYPE_BASIC;
+
+        foreach ($this->domainsForDataFixtureProvider->getAllowedDemoDataDomains() as $domainConfig) {
+            $locale = $domainConfig->getLocale();
+
+            $paymentData->name[$locale] = t('Cash', [], Translator::DATA_FIXTURES_TRANSLATION_DOMAIN, $locale);
+        }
+
+        $paymentData->enabled[Domain::SECOND_DOMAIN_ID] = true;
```
grossmannmartin: Now we have to create a separate `cash` payment for each domain. What about moving the setting to per-domain, so we can make the administration clearer?

Something like this, for instance

Image

or maybe to the right, empty space, after VAT settings.

packages/framework/src/Model/Payment/OrderRoundingTypeEnum.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Shopsys\FrameworkBundle\Model\Payment;
+
+use Shopsys\FrameworkBundle\Component\Enum\AbstractEnum;
+
+class OrderRoundingTypeEnum extends AbstractEnum
+{
+    public const string NONE = 'none';
+    public const string FIVES = 'fives';
+    public const string INTEGER = 'integer';
```
grossmannmartin: I'm not sure about the names. 
Integer seems to be too much of a "dev" option compared to the other options. A
And FIVES is ambiguous to me (I would like to rename even the user-facing texts).

What about `NONE`, `WHOLE`, and `NICKEL`|`STEP_0_05`|`FIVE_CENTS`|`NEAREST_0_05`.

grossmannmartin (changes requested):

packages/framework/src/Model/Order/OrderPriceCalculation.php
```diff
@@ -52,12 +54,16 @@ public function calculateOrderRoundingPrice(
...
             return null;
         }
 
         $priceWithVat = $orderTotalPrice->getPriceWithVat();
-        $roundedPriceWithVat = $priceWithVat->round(0);
+        $roundedPriceWithVat = match ($payment->getOrderRoundingType()) {
+            OrderRoundingTypeEnum::FIVES => $priceWithVat->multiply(20)->round(0)->divide(20, 2),
+            OrderRoundingTypeEnum::INTEGER => $priceWithVat->round(0),
+            default => throw new InvalidEnumCaseException(OrderRoundingTypeEnum::class, $payment->getOrderRoundingType()),
+        };
```
henzigo: Definition and calculation are separated. I would move this match into `OrderRoundingTypeEnum` class as a new method.

packages/framework/src/Form/Admin/Payment/PaymentFormType.php
```diff
@@ -163,9 +165,14 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
 
         $builderPriceGroup
-            ->add('czkRounding', YesNoType::class, [
-                'label' => 'Order in CZK round to whole crowns',
-                'help' => t('Rounding item with 0 % VAT will be added to your order. It is used for payment in cash.'),
+            ->add('orderRoundingTypeByDomainId', MultidomainType::class, [
+                'entry_type' => ChoiceType::class,
+                'entry_options' => [
+                    'choices' => $this->orderRoundingTypeEnum->getAllIndexedByTranslations(),
+                ],
```
grossmannmartin, line 172: What if we added

```
'display_mode' => 'columns',
'row_attr' => [
    'class' => 'mb-3',
],
```

It looks better imho

Image

packages/framework/src/Model/Order/Processing/OrderProcessorMiddleware/AddPaymentMiddleware.php
```diff
@@ -57,7 +57,7 @@ public function handle(
         $orderData->addTotalPrice($paymentPrice, OrderItemTypeEnum::TYPE_PAYMENT);
 
         $orderData->orderPayment = $orderItemData;
-        $orderData->paymentCzkRounding = $payment->isCzkRounding();
+        $orderData->paymentCzkRounding = $payment->hasOrderRoundingForDomain($domainId);
```
grossmannmartin: I'm a little surprised that we still have `paymentCzkRounding` in the app (sorry, I missed this in the first round 🫠) 

I think this property could be removed from the order altogether. Or do we use it for something?

packages/framework/src/Model/Payment/OrderRoundingTypeEnum.php
```diff
@@ -0,0 +1,38 @@
...
+    public function getAllIndexedByTranslations(): array
+    {
+        return [
+            t('None') => self::NONE,
+            t('To five cents (0.05)') => self::FIVE_CENTS,
+            t('To whole numbers (1.00)') => self::WHOLE,
+        ];
+    }
+
+    public function roundPrice(string $roundingType, Money $priceWithVat): Money
```
grossmannmartin: I would name it just Price. It does not matter if it's with or without VAT in this context.

grossmannmartin (approved):

packages/framework/src/Model/Order/Processing/OrderProcessorMiddleware/AddPaymentMiddleware.php
TomasLudvik: Yeah I think that change was merged when this was in some state already, so it was not caught. Nice catch!

grossmannmartin (approved):

PR #4526, opened 2026-03-27
Auto remove of stale annotations

### LINKED ISSUE #2615: Annotation fixer don't remove overridden removed methods
## Describe the bug
If the method is removed from vendor class, annotation remains on child classes. Phpstan can't find out that this method is missing. There is `Call to undefined method` error when you try to call these methods. 

For example, there are methods on CC in `App\Model\Order\OrderFacade`, which no longer exists in vendor:
- getCustomerUserOrderLimitedList,
- getByUuidAndUrlHash,
- getByUuidAndCustomerUser.

Methods are removed in ed95cb0.

grossmannmartin:

This PR improves maintenance of generated `@method` and `@property` annotations in project classes that extend the packages.
Until now, the annotation fixing process could add or update annotations, but it did not clean up entries that had become outdated after the parent framework classes changed.
As a result, some project classes still contained annotations for methods or properties that no longer existed, which made the generated metadata less trustworthy for IDEs and static analysis.

This change completes annotation maintenance by removing stale annotations in the same workflow, and it also cleans up outdated annotations in `project-base`, so the repository reflects the actual API of the extended classes.

#### Fixes issues

closes #2615 

----
  - https://mg-remove-superfl-anot.odin.shopsys.cloud
  - https://cz.mg-remove-superfl-anot.odin.shopsys.cloud
  - https://mg-remove-superfl-anot.odin.shopsys.cloud/sk


grossmannmartin:
@coderabbitai review

packages/framework/src/Component/ClassExtension/StaleAnnotationsRemover.php
```diff
@@ -0,0 +1,190 @@
...
+        ];
+    }
+
+    /**
+     * @param array<string, true> $validFrameworkMethodNames
+     * @param array<string, true> $validFrameworkPropertyNames
+     * @param array<string, true> $projectMethodNames
+     * @param array<string, true> $projectPropertyNames
+     */
+    protected function isAnnotationStale(
```
TomasLudvik, line 172: One possible problem of this method is that manually added methods and properties will be removed. I want to just mention this to ensure we are ok with such behavior.

TomasLudvik (approved):

packages/framework/src/Component/ClassExtension/StaleAnnotationsRemover.php
```diff
@@ -0,0 +1,190 @@
...
+        $allLines = explode("\n", $projectClassDocComment);
+        $filteredLines = [];
+
+        foreach ($allLines as $line) {
+            if (!isset($staleLinesSet[$line])) {
+                $filteredLines[] = $line;
+            }
+        }
+
+        $newDocComment = implode("\n", $filteredLines);
```
TomasLudvik: This code can potentially generate empty codeblock if no lines remain in $filteredLines

TomasLudvik (changes requested):

packages/framework/src/Component/ClassExtension/StaleAnnotationsRemover.php
grossmannmartin: Yes, I'm aware of that. 
Sadly, I don't have idea how to prevent that

packages/framework/src/Component/ClassExtension/StaleAnnotationsRemover.php
grossmannmartin: fixed in 629948c3e9

packages/framework/src/Component/ClassExtension/StaleAnnotationsRemover.php
```diff
@@ -0,0 +1,200 @@
...
+                $filteredLines[] = $line;
+            }
+        }
+
+        $contentLines = array_filter(
+            $filteredLines,
+            static fn (string $line) => !preg_match('~^\s*/?\*{1,2}\s*/?\s*$~', $line),
+        );
+
+        if ($contentLines === []) {
```
TomasLudvik:
```suggestion
        if (count($filteredLines) === 2) {
```

TomasLudvik: I find this much easier to read, what do you think?

TomasLudvik (changes requested):

TomasLudvik (approved):

pk16011990 (approved):

PR #4527, opened 2026-03-27
[CRUD] fix translations

henzigo:

Separates CRUD route loading and role provision from request-time definition building to fix locale/translation issues. Previously, `CrudControllerRouteLoader::load()` triggered `getItems()` which built full
  Definition objects — including `t()` translations — during route compilation before the request locale was set. This caused all translated strings to be cached with the default locale.

#### Fixes issues
... 

----
  - https://jg-crud-translations-fix.odin.shopsys.cloud
  - https://cz.jg-crud-translations-fix.odin.shopsys.cloud
  - https://jg-crud-translations-fix.odin.shopsys.cloud/sk

----
  - https://jg-crud-translations-fix.odin.shopsys.cloud
  - https://cz.jg-crud-translations-fix.odin.shopsys.cloud
  - https://jg-crud-translations-fix.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/src/Component/Crud/CrudControllerRegistry.php
+++ b/packages/administration/src/Component/Crud/CrudControllerRegistry.php
@@ -21,9 +21,19 @@ final class CrudControllerRegistry
     public const string CRUD_CONTROLLERS_EXTENSIONS_PARAMETER = 'shopsys.admin.crud_controllers_extensions';
 
     /**
-     * @var \Shopsys\AdministrationBundle\Component\Crud\Definition[]|null
+     * @var array<class-string<\Shopsys\AdministrationBundle\Controller\AbstractCrudController>, array{controllerName: string, entityClass: class-string, entityName: string}>|null
      */
-    private ?array $items = null;
+    private ?array $resolvedMetadata = null;
+
+    /**
+     * @var array<class-string<\Shopsys\AdministrationBundle\Controller\AbstractCrudController>, \Shopsys\AdministrationBundle\Controller\AbstractCrudControllerExtension[]>|null
+     */
+    private ?array $resolvedExtensions = null;
+
+    /**
+     * @var array<class-string<\Shopsys\AdministrationBundle\Controller\AbstractCrudController>, \Shopsys\AdministrationBundle\Component\Crud\Definition>
+     */
+    private array $definitions = [];
 
     /**
      * @param array<int, array{class: class-string<\Shopsys\AdministrationBundle\Controller\AbstractCrudController>, entityClass: string}> $crudControllers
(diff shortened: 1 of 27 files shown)


packages/administration/src/Component/Crud/Helper/CrudTransformationHelper.php
```diff
@@ -37,15 +38,45 @@ public static function generateController(string $controllerClass, ActionType $p
...
+    /**
+     * Converts PascalCase entity class short name to human-readable plural form (ucfirst).
+     *
+     * Example:
+     *     OrderItem => Order items
+     *     Category => Categories
+     */
+    public static function toPluralEntityName(string $entityName): string
+    {
+        $humanized = (string)u($entityName)->snake()->replace('_', ' ');
```
grossmannmartin: I would rather use `new UnicodeString(` everywhere. `u(` is short, but it doesn't say anything

packages/administration/src/Component/Crud/Helper/CrudTransformationHelper.php
```diff
@@ -37,15 +38,45 @@ public static function generateController(string $controllerClass, ActionType $p
      */
     public static function transformToRouteUrl(string $controllerName): string
     {
-        return $controllerName
-            |> self::getCleanControllerName(...)
-            |> (fn ($v) => preg_replace('/(?<!^)[A-Z]/', '-$0', $v))
-            |> strtolower(...);
+        return (string)u(self::getCleanControllerName($controllerName))->kebab();
```
grossmannmartin: getCleanControllerName could(should?) be private. I don't expect it to be called outside the scope of the helper.

packages/administration/src/Component/Crud/CrudControllerRegistry.php
```diff
@@ -39,85 +49,71 @@ public function __construct(
...
     /**
-     * @param class-string<\Shopsys\AdministrationBundle\Controller\AbstractCrudController> $controllerClass
-     * @param class-string $entityClass
-     * @param \Shopsys\AdministrationBundle\Controller\AbstractCrudControllerExtension[] $extensions
+     * Returns all registered CRUD controllers with metadata and config (including disabled).
+     *
+     * @return \Shopsys\AdministrationBundle\Component\Crud\CrudRegistryItem[]
      */
-    private function addItem(string $controllerClass, string $entityClass, array $extensions): void
+    public function getAll(): array
```
grossmannmartin, line 57: This method is called several times during the request, and each time it rebuilds the config using reflection. 
Couldn't we cache it?

packages/administration/src/Component/Router/CrudRouteProvider.php
```diff
@@ -46,35 +51,96 @@ final class CrudRouteProvider
...
+    public function getAll(): array
     {
-        return new CrudRouteItem(
-            controller: CrudTransformationHelper::generateController($item->controllerClass, $pageType),
-            route: $this->generateRoute($item, $pageType, $item->getConfig()->getRoutePrefix()),
-            routeName: CrudTransformationHelper::generateRouteName($item->controllerName, $pageType),
-            pageType: $pageType,
-        );
+        if ($this->cachedRouteItems !== null) {
+            return $this->cachedRouteItems;
```
grossmannmartin: We should use `InMemoryCache` instead of plain array

grossmannmartin (changes requested):

grossmannmartin:
also, the upgrade notes are missing

packages/administration/src/Component/Crud/Helper/CrudTransformationHelper.php
henzigo: Ok, why not

packages/administration/src/Component/Crud/Helper/CrudTransformationHelper.php
henzigo: Ok made it private

packages/administration/src/Component/Router/CrudRouteProvider.php
henzigo: Refactored 👍

packages/administration/src/Component/Crud/CrudControllerRegistry.php
henzigo: No, this method cannot be cached. That's exactly why translations didn't work. 

The first call to this method occurs before the language is set. This would cause the translations to be in the default language, and they would not be translated again.

henzigo: But I made a few changes. Instead of a full `Container`, the `ServiceLocator` is used, so it iterates only over CrudControllers. This should be better. Look: https://github.com/shopsys/shopsys/commit/1663cfb979f64cc6cf93cf8510336f019e09439d

henzigo:
> also, the upgrade notes are missing

I don't think it's needed ... or?

grossmannmartin (approved):

PR #4447, opened 2026-02-08
security headers

vitek-rostislav:
#### Description, the reason for the PR

- all the static security headers are now set at nginx level
- superadmin set CSP value is now applied again, even on storefront, thanks to exposing the value via FE API
- added docs about security headers
- the following headers are configured:
    - `Content-Security-Policy`
    - `Access-Control-Allow-Origin`
    - `Access-Control-Allow-Credentials`
    - `Referrer-Policy`
    - `Strict-Transport-Security`
    - `X-Frame-Options`
    - `X-Content-Type-Options`
    - `X-XSS-Protection` (explicitly turned off and replaced by `frame-ancestors` CSP setting)
- related PR for production setup: https://github.com/shopsys/deployment/pull/66

#### Fixes issues
... 

#### License Agreement for contributions

- [x] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://rv-security-headers.odin.shopsys.cloud
  - https://cz.rv-security-headers.odin.shopsys.cloud
  - https://rv-security-headers.odin.shopsys.cloud/sk

----
:globe_with_meridians: Live Preview:
  - https://rv-security-headers.odin.shopsys.cloud
  - https://cz.rv-security-headers.odin.shopsys.cloud
  - https://rv-security-headers.odin.shopsys.cloud/sk

PR #4531, opened 2026-03-27
[GrapesJS] fix cross-domain template loading blocked by CSP

grossmannmartin:

When editing multilocale blog article descriptions (or any GrapesJS field), the template URL was resolved per-locale to the corresponding domain (e.g., `https://example.cz/grapesjs-template`).
Since the admin always runs on the first domain (`https://example.com`), this resulted in a cross-origin AJAX request that was blocked by the `Content-Security-Policy` `connect-src` directive introduced in #4447.

Loading an arbitrary domain template based on the locale is just slightly better than use first domain every time.
The fix changes the template URL to a relative path (`/grapesjs-template`), which always resolves against the current (admin) domain, avoiding any cross-origin requests.

The now-unused `getDomainUrlByLocale` Twig function and its backing method in `DomainExtension` have been removed.

#### Fixes issues
... 

----
  - https://mg-grapes-other-domain.odin.shopsys.cloud
  - https://cz.mg-grapes-other-domain.odin.shopsys.cloud
  - https://mg-grapes-other-domain.odin.shopsys.cloud/sk

----
  - https://mg-grapes-other-domain.odin.shopsys.cloud
  - https://cz.mg-grapes-other-domain.odin.shopsys.cloud
  - https://mg-grapes-other-domain.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/templates/form/type/GrapesJsType.html.twig
+++ b/packages/administration/templates/form/type/GrapesJsType.html.twig
@@ -1,16 +1,10 @@
 {% block grapes_js_widget %}
-    {% set domainByLocale = '' %}
-
-    {% if attr['data-locale'] is defined %}
-        {% set domainByLocale = getDomainUrlByLocale(attr['data-locale']) %}
-    {% endif %}
-
     <div class="d-none">{{ form_widget(form) }}</div>
     <button
             type="button"
             class="btn btn-info js-grapesjs-button"
             data-textarea-id="{{ form.vars.id }}"
-            data-template-url="{{ domainByLocale ~ constant('Shopsys\\FrameworkBundle\\Form\\GrapesJsType::GRAPESJS_TEMPLATE_PATH') }}"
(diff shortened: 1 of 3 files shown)


vitek-rostislav (approved):
I just need to ask, as I do not remember that - have we lost some localization of the template by this change? Looking at the code, I am not able to tell. In the SF template, no translation is used, but `useNavigationQuery` should return proper navigation items (including translations) on the given domain, right? If this won't work anymore, it feels like it might be a problem from a business point of view. But as you say, it was just slightly better. I can live with that 🤷

pk16011990 (approved):

grossmannmartin:
> I just need to ask, as I do not remember that - have we lost some localization of the template by this change? Looking at the code, I am not able to tell. In the SF template, no translation is used, but `useNavigationQuery` should return proper navigation items (including translations) on the given domain, right? If this won't work anymore, it feels like it might be a problem from a business point of view. But as you say, it was just slightly better. I can live with that 🤷

Yes, we have lost some localization in the process... But I think it's acceptable

PR #4543, opened 2026-04-03
Replace Eslint and Prettier with Biome

chlebektomas:
Replace Eslint and Prettier with Biome.

----
  - https://tc-ssp-3845-biome.odin.shopsys.cloud
  - https://cz.tc-ssp-3845-biome.odin.shopsys.cloud
  - https://tc-ssp-3845-biome.odin.shopsys.cloud/sk

----
  - https://tc-ssp-3845-biome.odin.shopsys.cloud
  - https://cz.tc-ssp-3845-biome.odin.shopsys.cloud
  - https://tc-ssp-3845-biome.odin.shopsys.cloud/sk


project-base/storefront/biome/plugin-valid-interactive-content.grit
```diff
@@ -0,0 +1,74 @@
...
+        )
+    },
+    `<button $props>$children</button>` as $match where {
+        $match <: contains `<a $anchorProps>$anchorChildren</a>` as $invalid,
+        register_diagnostic(
+            span = $invalid,
+            message = "Button elements cannot contain anchor elements. Interactive elements cannot be nested inside other interactive elements."
+        )
+    },
+    `<a $props>$before <button $buttonProps>$buttonChildren</button> $after</a>` as $match where {
```
JanMolcik: This rule does not currently catch the `<a><button /></a>` case. Minimal repro:

```tsx
export const X = () => (
    <a href="/">
        <button>Go</button>
    </a>
);
```

Current result: no plugin diagnostic. Expected diagnostic: `Anchor elements cannot contain button elements. Interactive elements cannot be nested inside other interactive elements.`

Please rewrite this in the same descendant-matching style as the working `button ...` cases (match `<a $props>$children</a>` and then use `contains <button ...> as $invalid`) and add a regression fixture for it.

project-base/storefront/biome.json
```diff
@@ -0,0 +1,137 @@
...
+        "actions": {
+            "source": {
+                "organizeImports": "on"
+            }
+        }
+    },
+    "linter": {
+        "enabled": true,
+        "rules": {
+            "recommended": true,
```
JanMolcik, line 54: The previous ESLint setup had custom `jsx-a11y/anchor-is-valid` coverage for `Link` / `ExtendedNextLink`, but I don’t see an equivalent replacement in the current Biome setup. Built-in Biome `useValidAnchor` covers raw `<a>` elements, not these custom wrappers, so this migration currently loses part of the previous storefront-specific validation.

project-base/storefront/biome.json
```diff
@@ -0,0 +1,137 @@
...
+            "a11y": {
+                "noStaticElementInteractions": "off",
```
JanMolcik: This weakens accessibility lint parity compared to the previous ESLint setup. `noStaticElementInteractions` and `useSemanticElements` are both turned off, and I also don’t see an equivalent replacement for the previous `jsx-a11y/no-noninteractive-element-interactions` coverage. If this is intentional, it should be explicitly justified; otherwise we should restore equivalent protection.

docs/storefront/coding-standards.md
```diff
@@ -191,14 +191,14 @@ Interactive elements cannot be nested inside other interactive elements.
 
 #### Automatic Detection
 
-Your ESLint setup automatically catches:
+Your Biome setup automatically catches:
```
JanMolcik, line 194: This section is now misleading. It still references old ESLint rule names and also claims coverage that does not match the current Biome config. Please update this section to describe the actual enforced Biome rules (and the local Grit plugin checks), rather than the previous ESLint naming/behavior.

JanMolcik (changes requested):
This is close, but there are still a few parity gaps that should be fixed before approval:

- the custom Grit rule meant to catch `<a><button /></a>` does not currently fire, so the migration regresses an existing invalid-interactive-structure check,
- the previous custom `anchor-is-valid` coverage for `Link` / `ExtendedNextLink` is no longer replaced,
- the current Biome config weakens some accessibility protection compared to the previous ESLint setup,
- and `docs/storefront/coding-standards.md` still describes outdated / inaccurate rule coverage.

Upgrade notes and the `@biomejs/biome` license entry are already present, so those are no longer concerns.

project-base/storefront/biome/plugin-valid-interactive-content.grit
```diff
@@ -0,0 +1,122 @@
...
+        )
+    },
+    `<a $props>$children</a>` as $match where {
+        $match <: contains `<button $buttonProps>$buttonChildren</button>` as $invalid,
+        register_diagnostic(
+            span = $invalid,
+            message = "Anchor elements cannot contain button elements. Interactive elements cannot be nested inside other interactive elements."
+        )
+    },
+    `<button $props>$before <button $buttonProps>$buttonChildren</button> $after</button>` as $match where {
```
JanMolcik: This still does not catch nested buttons. I reproduced this fixture and Biome reported no plugin diagnostic:

```tsx
export const X = () => (
    <button>
        <button>Nested</button>
    </button>
);
```

Please rewrite this like the fixed `a > button` rule: match `<button $props>$children</button>` and use `contains <button ...> as $invalid`, instead of matching `$before <button ...> $after` inline.

project-base/storefront/biome/plugin-valid-interactive-content.grit
```diff
@@ -0,0 +1,122 @@
...
+        $match <: contains `<button $buttonProps>$buttonChildren</button>` as $invalid,
+        register_diagnostic(
+            span = $invalid,
+            message = "Button elements cannot contain other button elements. Interactive elements cannot be nested."
+        )
+    },
+    `<ExtendedNextLink href="#" $props>$children</ExtendedNextLink>` as $invalid where {
```
JanMolcik: These `ExtendedNextLink` checks are too shape-dependent. They catch e.g. `<ExtendedNextLink href="#" className="x">...`, but they do not catch `<ExtendedNextLink href="#">...` or `<ExtendedNextLink className="x" href="#">...`. Please make the rule independent of extra props / prop order, otherwise the replacement for the old link validation is easy to bypass.

JanMolcik: The previous ESLint `anchor-is-valid` customization covered both `Link` and `ExtendedNextLink`, but the new custom checks only cover `ExtendedNextLink`. I verified `<Link href="#" className="x">Go</Link>` does not emit a plugin diagnostic. Please either add equivalent `Link` coverage or explicitly document that we intentionally drop that part of the old parity.

project-base/storefront/biome/plugin-valid-interactive-content.grit
```diff
@@ -0,0 +1,122 @@
+engine biome(1.0)
+language js(typescript, jsx)
+
+or {
```
JanMolcik, line 4: Please add regression fixtures for this plugin. At minimum they should cover `button > div`, `button > a`, `a > button`, `button > button`, invalid `ExtendedNextLink href="#"` without extra props, invalid `ExtendedNextLink` with reordered props, and `Link href="#"` if we keep the old `Link` parity. The current `button > button` miss shows why this needs fixture coverage.

JanMolcik (changes requested):
Thanks for the follow-up fixes. The `a > button` case and the docs/a11y config are now improved, but the custom Grit plugin still has parity gaps: `button > button` is not reported, `ExtendedNextLink` invalid href checks only match a narrow prop shape, and the old `Link` wrapper coverage is still missing. I would keep this as request changes until those are fixed and covered by regression fixtures.

JanMolcik (approved):
The follow-up fixes address my review comments. I re-checked the Grit plugin behavior, regression fixtures, docs, and CI status; this is ready to move forward from my side.

pk16011990 (approved):

PR #4549, opened 2026-04-13
Fix last visited products showing fewer items when some are unavailable

grossmannmartin:

The "Last Visited Products" slider could show fewer than 10 products when some of the stored items became hidden or unavailable. This happened because the feature stored exactly 10 product catalogue numbers — the same number it displayed. When the API filtered out unavailable products, there were no extras to fill the gap.

Now, 15 items are stored in the cookie while only 10 are displayed. The extra buffer ensures that even if a few products become unavailable, the slider still shows the expected number of items.

#### Fixes issues

N/A

----
  - https://mg-last-visited.odin.shopsys.cloud
  - https://cz.mg-last-visited.odin.shopsys.cloud
  - https://mg-last-visited.odin.shopsys.cloud/sk


chlebektomas:
The code changes look good. Regression tests for the new 15-stored / 10-displayed behavior would be nice, but it’s up to you.

chlebektomas (approved):

pk16011990 (approved):

PR #4560, opened 2026-04-14
Preserve expired-token error code across backend and storefront

grossmannmartin:

Until now, every token authentication failure in the Frontend API was reported to the storefront with the same `invalid-token` user code and an HTTP 401 response. That meant an expired token — a normal, expected situation after a period of inactivity — was indistinguishable from a genuinely invalid token, so the storefront could not react to it appropriately (e.g. trigger a silent token refresh or show a friendly "please sign in again" flow).

This PR makes the backend preserve the original error code thrown during token processing. When a token is expired, the API now returns the `expired-token` user code; when it is malformed or otherwise invalid, it keeps returning `invalid-token`. The authentication endpoint also returns HTTP 200 with the GraphQL error payload instead of HTTP 401, so the error travels through the GraphQL pipeline like any other user error.

On the storefront side, auth-error detection was updated to recognize these GraphQL user codes in addition to the legacy HTTP 401 check. As a result, expired sessions are detected reliably, tokens are cleared, server-side requests redirect to the login page, and the `expired-token` code is classified as a no-log error so expired sessions no longer produce noise in error logs.

#### Fixes issues

N/A

----
  - https://mg-http-auth.odin.shopsys.cloud
  - https://cz.mg-http-auth.odin.shopsys.cloud
  - https://mg-http-auth.odin.shopsys.cloud/sk

----
  - https://mg-http-auth.odin.shopsys.cloud
  - https://cz.mg-http-auth.odin.shopsys.cloud
  - https://mg-http-auth.odin.shopsys.cloud/sk

### DIFF
--- a/upgrade-notes/backend_20260414_180244.md
+++ b/upgrade-notes/backend_20260414_180244.md
@@ -1,0 +1,3 @@
+#### token auth errors now preserve expired-token across backend and storefront ([#4560](https://github.com/shopsys/shopsys/pull/4560))
+
+- Frontend API token authentication failures are now returned with HTTP code 200 instead of previous 401 code with GraphQL error `extensions.userCode` values `expired-token` and `invalid-token`, update your application to reflect it
--- a/packages/frontend-api/src/Model/Token/Exception/ExpiredTokenUserMessageException.php
+++ b/packages/frontend-api/src/Model/Token/Exception/ExpiredTokenUserMessageException.php
@@ -6,4 +6,5 @@ namespace Shopsys\FrontendApiBundle\Model\Token\Exception;
 
 class ExpiredTokenUserMessageException extends TokenUserMessageException
 {
+    protected const string CODE = 'expired-token';
 }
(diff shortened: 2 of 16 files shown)


upgrade-notes/backend_20260414_180244.md
```diff
@@ -0,0 +1,5 @@
+#### token auth errors now preserve expired-token across backend and storefront ([#4560](https://github.com/shopsys/shopsys/pull/4560))
+
+- Frontend API token authentication failures now return `HTTP 200` with GraphQL errors instead of `HTTP 401`
+- Frontend API token authentication failures can now return `extensions.userCode = expired-token` for expired JWTs instead of always returning `invalid-token`
+- if your custom Frontend API client or middleware handles authentication failures by checking only HTTP status codes, update it to inspect GraphQL error `extensions.userCode` values `expired-token` and `invalid-token`
```
TomasLudvik: These are not upgrade notes

TomasLudvik (changes requested):

chlebektomas (approved):

pk16011990 (approved):

TomasLudvik (approved):

PR #4561, opened 2026-04-14
Improve GraphQL query performance with batch loading and Redis caching

TomasLudvik:

Several frequently used GraphQL queries suffered from N+1 query problems and unnecessary database roundtrips, increasing page load times on the storefront.

This PR addresses the most impactful bottlenecks:

- **Slug resolution** for brands, stores, flags, categories, blog categories, and products now uses DataLoader batch loading instead of individual database queries per entity. A single query resolves all slugs in a GraphQL response at once.
- **Promoted products** are now fetched with a single Elasticsearch query instead of first loading top products from the database and then looking them up in Elasticsearch. The `is_promoted` flag and `top_product_position` are exported directly into the product ES index.
- **Store opening hours and closed days** are batch-preloaded when listing stores, eliminating per-store queries for opening hours and closed day exclusions.
- **SeoPage lookup** uses an optimized query with a new composite database index on `(domain_id, page_slug)`.
- **Empty array short-circuits** in `ParameterRepository` and `FlagRepository` skip unnecessary queries when there are no UUIDs to look up.
- **Redis caching** is enabled for `AdvertsQuery`, `BrandsQuery`, `StoresQuery`, `SeoPageQuery`, and `PromotedProductsQuery` with automatic cache invalidation when data changes in admin.

----
  - https://tl-graphql-performance-improvements.odin.shopsys.cloud
  - https://cz.tl-graphql-performance-improvements.odin.shopsys.cloud
  - https://tl-graphql-performance-improvements.odin.shopsys.cloud/sk

----
  - https://tl-graphql-performance-improvements.odin.shopsys.cloud
  - https://cz.tl-graphql-performance-improvements.odin.shopsys.cloud
  - https://tl-graphql-performance-improvements.odin.shopsys.cloud/sk


packages/framework/src/Model/Store/OpeningHours/StoreOpeningHoursProvider.php
```diff
@@ -41,6 +42,18 @@ public function getOpeningHoursSetting(Store $store): SpatieOpeningHours
         );
     }
 
+    /**
+     * @param \Shopsys\FrameworkBundle\Model\Store\Store[] $stores
+     */
+    public function preloadClosedDaysForStores(array $stores): void
```
grossmannmartin: This whole flow of stores and excluded days is quite unpredictable and unpredictably coupled, imho.

1. I have to manually call this preload somewhere in the code, so the getExceptions() method elsewhere gets preloaded data
2. I'm not sure why opening hours are preloaded directly in the repository, and closed days are externally. 

As a bonus, for example, for this simple call 
```
{
	stores {
    edges{
      node{
        name
      }
    }
  }
}
```
actually has the performance worse

packages/framework/src/Model/Store/ClosedDay/ClosedDayRepository.php
```diff
@@ -111,6 +99,74 @@ protected function createPublicHolidaysQueryBuilder(
...
+    public function getFollowingWeekClosedDaysForStoresIndexedByStoreId(array $stores): array
+    {
+        if ($stores === []) {
+            return [];
+        }
+
+        $firstStore = array_first($stores);
+
+        $closedDays = $this
+            ->createFollowingWeekClosedDaysQueryBuilder($firstStore->getDomainId())
```
grossmannmartin: Implicitly assuming all stores will have the same domain is not a safe pattern, imho.
As it's public method, I can call it with any stores I want

project-base/storefront/graphql/requests/adverts/queries/AdvertsQuery.graphql
```diff
@@ -1,4 +1,4 @@
-query AdvertsQuery($categoryUuid: Uuid, $positionNames: [String!]) {
+query AdvertsQuery($categoryUuid: Uuid, $positionNames: [String!]) @redisCache(ttl: 300) {
```
grossmannmartin: If adverts are rarely changed, why not use 3600 as the rest of the queries do?

packages/frontend-api/src/Model/Blog/Category/BlogCategoryResolverMap.php
```diff
@@ -21,6 +22,7 @@ public function __construct(
         protected readonly BlogCategoryFacade $blogCategoryFacade,
         protected readonly BlogArticleElasticsearchFacade $blogArticleElasticsearchFacade,
         protected readonly HreflangLinksFacade $hreflangLinksFacade,
+        protected readonly DataLoaderInterface $blogCategorySlugBatchLoader,
```
grossmannmartin, line 25: I would re-validate this whole commit. 

It's true that on the first load, the slugs are loaded sequentially, introducing the N+1 problem. But each consecutive call gets the slug from the Redis cache. 
So, instead of a slower first load, we’ll have slower every subsequent call. Making the `main_friendly_url_slug_cache` pointless.

packages/framework/src/Model/Product/TopProduct/TopProductFacade.php
```diff
@@ -49,7 +56,10 @@ public function saveTopProductsForDomain(int $domainId, array $products): void
         foreach ($products as $product) {
             $topProduct = $this->topProductFactory->create($product, $domainId, $position++);
             $this->em->persist($topProduct);
+            $affectedProductIds[] = $product->getId();
         }
         $this->em->flush();
+
+        $this->productRecalculationDispatcher->dispatchProductIds(array_unique($affectedProductIds));
```
grossmannmartin: It's not necessary to dispatch whole product, we should create a new scope for top products and dispatch just this scope

project-base/storefront/graphql/requests/products/queries/PromotedProductsQuery.graphql
```diff
@@ -1,4 +1,4 @@
-query PromotedProductsQuery {
+query PromotedProductsQuery @redisCache(ttl: 300) {
```
grossmannmartin: I think 3600 is ok. Especially if the admin change invalidates it

packages/framework/src/Model/Seo/Page/SeoPageFacade.php
```diff
@@ -59,6 +65,8 @@ public function delete(int $seoPageId): void
 
         $this->em->remove($seoPage);
         $this->em->flush();
+
+        $this->cleanStorefrontCacheFacade->cleanStorefrontGraphqlQueryCache(CleanStorefrontCacheFacade::SEO_PAGE_QUERY_KEY_PART);
```
grossmannmartin, line 69: It's somehow related to the seo pages, but not to this line.

If you enter the page with the seo page (cart, homepage), the unknown error pops up.

grossmannmartin (changes requested):

packages/framework/src/Model/Seo/Page/SeoPageRepository.php
```diff
@@ -73,6 +73,7 @@ public function findByDomainIdAndPageSlug(int $domainId, string $pageSlug): ?Seo
             ->andWhere('spd.pageSlug = :pageSlug')
             ->setParameter('domainId', $domainId)
             ->setParameter('pageSlug', $pageSlug)
+            ->setMaxResults(1)
```
grossmannmartin: Shouldn't the domain_id + slug be a unique index? There is a validation in the form, so I think it's fair to back it with the database constraint.

packages/framework/src/Model/Product/TopProduct/TopProductFacade.php
```diff
@@ -62,7 +63,10 @@ public function saveTopProductsForDomain(int $domainId, array $products): void
         }
         $this->em->flush();
 
-        $this->productRecalculationDispatcher->dispatchProductIds(array_unique($affectedProductIds));
+        $this->productRecalculationDispatcher->dispatchProductIds(
+            array_unique($affectedProductIds),
```
grossmannmartin: array_unique is called already in the dispatchProductIds method.

packages/framework/src/Model/Product/TopProduct/TopProductFacade.php
```diff
@@ -39,6 +44,11 @@ public function saveTopProductsForDomain(int $domainId, array $products): void
     {
         $oldTopProducts = $this->topProductRepository->getAll($domainId);
 
+        $affectedProductIds = array_map(
+            static fn (TopProduct $topProduct) => $topProduct->getProduct()->getId(),
+            $oldTopProducts,
+        );
+
         foreach ($oldTopProducts as $oldTopProduct) {
```
grossmannmartin, line 49: nitpick, we don't need a separate array_map, but we may fill the $affectedProductIds array in this loop to avoid iterating the array twice

packages/framework/src/Model/Product/TopProduct/TopProductRepository.php
```diff
@@ -34,6 +34,21 @@ public function getAll(int $domainId): array
...
+    public function getTopProductPositionsIndexedByProductIdForDomain(int $domainId): array
+    {
+        $topProducts = $this->getAll($domainId);
+        $positions = [];
+
+        foreach ($topProducts as $topProduct) {
+            $positions[$topProduct->getProduct()->getId()] = $topProduct->getPosition();
+        }
+
+        return $positions;
```
grossmannmartin, line 56: We could use DQL directly to avoid entity hydration when everything we need is just product_id -> position

```php
$rows = $this->em->createQueryBuilder()
        ->select('IDENTITY(tp.product) AS productId', 'tp.position AS position')
        ->from(TopProduct::class, 'tp')
        ->andWhere('tp.domainId = :domainId')
        ->setParameter('domainId', $domainId)
        ->getQuery()
        ->getArrayResult();

    $positions = [];

    foreach ($rows as $row) {
        $positions[(int) $row['productId']] = $row['position'];
    }

    return $positions;
```

grossmannmartin (changes requested):

grossmannmartin (approved):

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Settings/SettingsDecorator.types.yaml
```diff
@@ -52,3 +52,7 @@ SettingsDecorator:
                 type: "[PhonePrefix!]!"
                 resolve: "@=query('phonePrefixesQuery')"
                 description: "Returns available phone prefixes for current domain"
+            defaultPricingGroupId:
+                type: "Int!"
+                resolve: "@=query('defaultPricingGroupIdQuery')"
+                description: "ID of the default pricing group for the current domain. Used by the storefront to scope the public storefront-query cache for anonymous traffic so it shares a bucket with logged-in users in the default pricing group."
```
grossmannmartin: API should not describe how the field is used in the client, as the implementation may change.

project-base/storefront/urql/fetcher.ts
```diff
@@ -9,14 +9,54 @@ const getHash = async (data: string): Promise<string> => {
...
+    const authToken = headers.get('X-Auth-Token')?.replace(/^Bearer\s+/i, '');
+    const defaultPricingGroupId = (domainId && defaultPricingGroupByDomain.get(domainId)) ?? 0;
+    let pricingGroupId: number | string = defaultPricingGroupId;
+    let roles: string[] = [];
+
+    if (authToken) {
+        try {
+            const [, payloadSegment] = authToken.split('.');
+            const payload = JSON.parse(Buffer.from(payloadSegment, 'base64url').toString('utf8'));
+            pricingGroupId = payload?.pricingGroupId ?? defaultPricingGroupId;
```
chlebektomas, line 36: Critical: the per-pricing-group cache bucket is selected from the unverified JWT payload, and cache hits are returned directly from the storefront before backend authentication validates the token. A client can forge X-Auth-Token with arbitrary pricingGroupId/roles; if that bucket already exists in Redis, the response is served without signature, expiry, or revocation checks. Please move this cache boundary after trusted auth, or bypass Redis for authenticated/per-pricing-group requests unless the bucket comes from a validated auth context.

project-base/storefront/graphql/requests/products/queries/PromotedProductsQuery.graphql
```diff
@@ -1,4 +1,4 @@
-query PromotedProductsQuery {
+query PromotedProductsQuery @redisCachePerPricingGroup(ttl: 3600) {
```
chlebektomas, line 1: Warning: the source query uses ttl: 3600, but the generated runtime document in PromotedProductsQuery.generated.tsx still uses ttl: 300. Runtime currently follows the generated file, and the next GraphQL regeneration will silently change cache behavior. Please align the source and generated document.

packages/frontend-api/src/Model/Token/TokenCustomerUserTransformer.php
```diff
@@ -16,6 +16,7 @@ public function transform(CustomerUser $user): array
             FrontendApiUser::CLAIM_EMAIL => $user->getEmail(),
             FrontendApiUser::CLAIM_FULL_NAME => $user->getFullName(),
             FrontendApiUser::CLAIM_ROLES => $user->getRoles(),
+            FrontendApiUser::CLAIM_PRICING_GROUP_ID => $user->getPricingGroup()->getId(),
```
chlebektomas, line 22: Suggestion: transform() returns a structured claims array, but the method only declares array. Please add a docblock shape that includes pricingGroupId and the other claim keys, in line with the project rule to specify non-obvious array contents.

project-base/storefront/urql/fetcher.ts
chlebektomas, line 36: Critical: authenticated tokens that do not yet contain pricingGroupId fall back to the default pricing group. Existing access tokens created before this deploy remain valid for a short time, so a non-default pricing-group user can miss the cache, receive correct backend prices, and then store them under the default/pg0 bucket for anonymous or default-pricing users. If an auth token exists but pricingGroupId/roles are missing or invalid, per-pricing-group responses should not read from or write to Redis.

project-base/storefront/urql/fetcher.ts
```diff
@@ -9,14 +9,54 @@ const getHash = async (data: string): Promise<string> => {
...
+    }
+
+    const id = (data as { settings?: { defaultPricingGroupId?: unknown } }).settings?.defaultPricingGroupId;
+
+    if (typeof id === 'number') {
+        defaultPricingGroupByDomain.set(domainId, id);
+    }
+};
+
+const getAuthBucketFromHeaders = async (headers: Headers, domainId: string | null): Promise<string> => {
```
chlebektomas, line 26: Warning: there are no tests covering @redisCachePerPricingGroup behavior. Please add coverage for cache hit/miss keying, sorted roles, malformed or forged JWT payloads, missing pricingGroupId in an authenticated token, and directive stripping. The auth/cache leakage risk here is exactly the kind of regression that should be locked by focused fetcher tests.

chlebektomas (changes requested):

grossmannmartin (approved):

project-base/storefront/urql/fetcher.ts
```diff
@@ -102,13 +142,37 @@ export const fetcher =
...
+            const isPerPricingGroup = init.body.match(PER_PRICING_GROUP_CACHE_REGEXP) !== null;
+
+            if (isPerPricingGroup && hasAuthToken(headers)) {
+                return fetch(createCleanedInput(input), {
+                    ...init,
+                    body,
+                });
+            }
+
+            const authBucket = isPerPricingGroup ? `${getAnonymousPricingGroupBucket(domainId)}:` : '';
```
JanMolcik: This per-pricing-group cache bucket can keep using a stale defaultPricingGroupId after the admin default pricing group is changed, so anonymous requests remain in the old bucket. I verified this is reproducible with the Playwright CLI. Please invalidate or refresh this dependency when the default pricing group changes.

Steps to reproduce:
1. Open the review storefront homepage at https://tl-graphql-performance-improvements.odin.shopsys.cloud/ and the admin pricing groups page at https://tl-graphql-performance-improvements.odin.shopsys.cloud/admin/pricing/group/list/.
2. In admin, note the current “Default pricing group” value.
3. Clear the Storefront Query Cache from the Redis/cache management screen so the next anonymous storefront request seeds a fresh baseline.
4. In an anonymous browser session, reload the storefront homepage and inspect the page `__NEXT_DATA__` payload; the `urqlState` entry for `SettingsQuery` shows the current `settings.defaultPricingGroupId`.
5. Back in admin, change “Default pricing group” to a different group and save.
6. Without clearing the Storefront Query Cache, reload the anonymous storefront homepage and inspect `__NEXT_DATA__` again.
7. Observe that `settings.defaultPricingGroupId` still has the previous value even though admin shows the new default. Clearing the Storefront Query Cache makes the new value appear, confirming this is stale cache invalidation that needs to be fixed.

JanMolcik (changes requested):
Requesting changes for the verified cache invalidation issue below.

project-base/storefront/urql/fetcher.ts
chlebektomas: @TomasLudvik Could you please check the latest fixup? It should fix this issue.

packages/framework/src/Model/Pricing/Group/PricingGroupSettingFacade.php
```diff
@@ -56,11 +58,21 @@ public function getDefaultPricingGroupByDomain(DomainConfig $domainConfig): Pric
...
-            $domainConfig->getId(),
+            $domainId,
         );
+
+        if ((int)$currentDefaultPricingGroupId === $pricingGroup->getId()) {
+            return;
+        }
+
+        $this->cleanStorefrontCacheFacade->cleanStorefrontGraphqlQueryCache(CleanStorefrontCacheFacade::SETTINGS_QUERY_KEY_PART);
+        $this->cleanStorefrontCacheFacade->cleanStorefrontGraphqlQueryCache(CleanStorefrontCacheFacade::PROMOTED_PRODUCTS_QUERY_KEY_PART);
```
TomasLudvik: Drop this - no need to clean cache of products as it is still correct

packages/framework/tests/Unit/Model/Pricing/Group/PricingGroupSettingFacadeTest.php
```diff
@@ -0,0 +1,126 @@
+<?php
```
TomasLudvik: This tests seem to be unnecessary, drop them

JanMolcik (approved):

PR #4564, opened 2026-04-16
Add MCP server for AI-based read-only database exploration 

vitek-rostislav:

This PR adds a production MCP (Model Context Protocol) server to Shopsys Platform, letting superadministrators connect external AI clients such as Claude Code or Codex CLI directly to the project database for read-only exploration. No LLM or prompt orchestration is embedded in Shopsys itself - the server is only a protocol, authentication, and data-access boundary.

The MCP server exposes three tools: `getDatabaseTableNames`, `getDatabaseSchema`, and `executeSql`. SQL is parsed into a PostgreSQL AST via `ext-pg_query` and validated structurally against an explicitly exposed schema. Queries run through a dedicated read-only Doctrine connection with row limits and a statement timeout, and every tool call is audit-logged to a dedicated Monolog channel.

Schema exposure is explicit and deny-by-default. Entities and columns become queryable only when annotated with `#[AsMcpTable]` / `#[AsMcpColumn]` (shipped in a lightweight `shopsys/mcp-attributes` package so `framework` can depend on them without pulling in the server runtime). A custom PHPStan rule enforces that every Doctrine entity and mapped field in `App\` and `Shopsys\` namespaces makes an explicit exposure decision, so new fields cannot silently leak to AI clients. The generated schema artifact at `project-base/app/mcp-schema.json` is refreshed automatically after migrations and is the single runtime source of truth.

Administrators manage the feature under their user menu, under "My MCP server"

image

The page supports two authentication paths: a manual token (shown once, hashed at rest) for clients that cannot use OAuth, and a browser-based OAuth authorization flow for clients that can - the administrator approves or denies access in the administration UI, and connected clients are listed by their registered `client_name` and can be revoked individually. MCP requests are authenticated through a dedicated Bearer-token firewall, independent from the administration browser session.

In production, the feature is enabled only if proper database credentials (`MCP_DATABASE_USER` and `MCP_DATABASE_PASSWORD`) are set.

The feature ships as two new packages - `shopsys/mcp-attributes` and `shopsys/mcp` - and includes documentation, Claude Code / Codex onboarding snippets, and a browser auth flow reference.

#### Fixes issues

N/A

----
  - https://rv-shopsys-mcp-2.odin.shopsys.cloud
  - https://cz.rv-shopsys-mcp-2.odin.shopsys.cloud
  - https://rv-shopsys-mcp-2.odin.shopsys.cloud/sk


grossmannmartin:
@coderabbitai review

composer.json
```diff
@@ -29,6 +29,8 @@
             "Shopsys\\S3Bridge\\": "packages/s3-bridge/src/",
             "Shopsys\\FrameworkBundle\\": "packages/framework/src/",
             "Shopsys\\LuigisBoxBundle\\": "packages/luigis-box/src/",
+            "Shopsys\\McpAttributes\\": "packages/mcp-attributes/src/",
+            "Shopsys\\McpBundle\\": "packages/mcp/src/",
```
grossmannmartin, line 33: Shouldn't be these packages added also to the replace section?

project-base/app/config/packages/mcp.yaml
```diff
@@ -0,0 +1,12 @@
+mcp:
+    app: shopsys-database
+    client_transports:
+        http: true
+        stdio: false
+    description: 'Shopsys database MCP server'
+    http:
```
grossmannmartin, line 7: Just an idea - mostly it's named simply as shopsys mcp, the database in the name seems redundant to me for future sake

packages/mcp/src/Component/Security/McpTokenAuthenticator.php
```diff
@@ -0,0 +1,107 @@
...
+use Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+use Symfony\Component\Security\Http\Authenticator\Passport\Passport;
+use Symfony\Component\Security\Http\Authenticator\Passport\SelfValidatingPassport;
+use Symfony\Component\Security\Http\EntryPoint\AuthenticationEntryPointInterface;
+
+class McpTokenAuthenticator extends AbstractAuthenticator implements AuthenticationEntryPointInterface
+{
+    protected const string HEADER_AUTHORIZATION = 'Authorization';
+    protected const string BEARER_PREFIX = 'Bearer ';
```
grossmannmartin: I would rather use a different header from `Authorization`, if possible, to avoid conflict with the HTTP auth.

packages/mcp/src/Resources/config/services.yaml
```diff
@@ -0,0 +1,8 @@
+services:
+    _defaults:
+        autoconfigure: true
+        autowire: true
+        public: false
+
+    Shopsys\McpBundle\:
+        resource: '../../**/*{Authenticator,DataFactory,Facade,Factory,Generator,Hasher,Lookup,Repository}.php'
```
grossmannmartin: shouldn't we autoconfigure only classes from the src subfolder?

packages/mcp/src/Migrations/Version20260401100401.php
```diff
@@ -0,0 +1,36 @@
...
+                secret_hash VARCHAR(255) NOT NULL,
+                created_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL,
+                last_used_at TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL,
+                revoked_at TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL,
+                replaced_at TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL,
+                administrator_id INT NOT NULL,
+                PRIMARY KEY (id)
+            )');
+        $this->sql('CREATE UNIQUE INDEX UNIQ_5A30067464EAEB46 ON administrator_mcp_tokens (public_token_id)');
+        $this->sql('CREATE INDEX IDX_5A3006744B09E92C ON administrator_mcp_tokens (administrator_id)');
```
grossmannmartin, line 32: I would check if the lookup for the token uses indexes properly. 
Lookup uses `public_token_id`, `revoked_at`, and `replaced_at` to find token. I'm not sure about expected row count, but I assume it can be a large number, so the lookup should use indexes properly.

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpToken.php
```diff
@@ -0,0 +1,139 @@
...
+     */
+    #[ORM\Column(type: 'datetime_immutable', nullable: true)]
+    protected $replacedAt;
+
+    public function __construct(AdministratorMcpTokenData $administratorMcpTokenData)
+    {
+        $this->administrator = $administratorMcpTokenData->administrator;
+        $this->publicTokenId = $administratorMcpTokenData->publicTokenId;
+        $this->secretHash = $administratorMcpTokenData->secretHash;
+        $this->createdAt = $administratorMcpTokenData->createdAt;
```
grossmannmartin, line 99: I think it would be safer and better aligned with the business model if the secretHash were set only once during entity creation and not exposed in the data object. 
But it's rather a nitpick.

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpTokenFacade.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public function generateTokenForAdministrator(Administrator $administrator): string
+    {
+        $now = $this->clock->now();
+        $existingAdministratorMcpToken = $this->administratorMcpTokenRepository->findActiveByAdministrator($administrator);
+
+        if ($existingAdministratorMcpToken !== null) {
+            $existingAdministratorMcpToken->replace($now);
+            $this->entityManager->flush();
```
grossmannmartin: What is the idea behind the replace? 

So we keep a history of all issued tokens? 
I don't see (yet) what it's used for, except for the token lookup

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpTokenLookup.php
```diff
@@ -0,0 +1,35 @@
...
+    public function __construct(
+        protected readonly AdministratorMcpTokenRepository $administratorMcpTokenRepository,
+        protected readonly AdministratorMcpTokenHasher $administratorMcpTokenHasher,
+    ) {
+    }
+
+    public function findValidTokenByTokenString(string $tokenString): ?AdministratorMcpToken
+    {
+        $tokenParts = explode('.', $tokenString, 2);
+
```
grossmannmartin, line 22: the logic behind how are parts combined together is spread in two classes (in the final implementation in `\Shopsys\McpBundle\Component\Security\McpBearerToken::parseTokenString()` and `\Shopsys\McpBundle\Model\Administrator\McpToken\AdministratorMcpIssuedToken::getTokenString()`. 

I think it would be nice to have it unified in one place, wdyt?

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpTokenRepository.php
```diff
@@ -0,0 +1,49 @@
...
+            ->getOneOrNullResult();
+    }
+
+    public function findActiveByPublicTokenId(string $publicTokenId): ?AdministratorMcpToken
+    {
+        return $this->getAdministratorMcpTokenEntityRepository()->createQueryBuilder('amt')
+            ->where('amt.publicTokenId = :publicTokenId')
+            ->andWhere('amt.revokedAt IS NULL')
+            ->andWhere('amt.replacedAt IS NULL')
+            ->setParameter('publicTokenId', $publicTokenId)
```
grossmannmartin, line 48: nitpick: we could extract a common query builder parts to the protected method

packages/mcp/src/ShopsysMcpBundle.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Shopsys\McpBundle;
+
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass;
+use Override;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+class ShopsysMcpBundle extends Bundle
+{
```
grossmannmartin, line 20: Maybe we could use AbstractBundle (see https://symfony.com/doc/7.4/bundles/configuration.html).
I have another idea about it later in review

packages/administration/assets/src/js/utils/confirmWindow.js
```diff
@@ -10,23 +10,55 @@ export default class ConfirmWindow {
...
                 content,
                 style,
                 continueUrl,
+                continueEvent,
             });
         });
     }
 
+    static resolveContinueEvent($element, continueUrl) {
+        if (continueUrl !== '#') {
```
grossmannmartin: I think this overcomplicates things. I see the usage only to generate the MCP token, but this may be also a link (with button style), or?

packages/mcp/src/Controller/Admin/McpServerController.php
```diff
@@ -0,0 +1,116 @@
...
+        $this->administratorMcpTokenFacade->revokeTokenForAdministrator($administrator);
+        $this->addSuccessFlash(t('MCP token revoked'));
+
+        return $this->redirectToRoute('admin_superadmin_mcp_token');
+    }
+
+    protected function createGenerateForm(bool $hasActiveToken): FormInterface
+    {
+        return $this->createForm(GenerateMcpServerTokenFormType::class, null, [
+            'action' => $this->generateUrl('admin_superadmin_mcp_token_generate'),
```
grossmannmartin: Right now I can have multiple authorized mcp clients (claude code), but just single manually generated mcp token. Isn't this limiting? 

Why not unify the list of tokens? Now I have to expand the last section to know that I have some manually generated mcp token. 

I would like to have a list of all active tokens, to be able to generate more manual tokens (I have one set in my automated tool, and now I need ad-hoc access). 

And it would be beautiful to be able to set custom expiration (like 1 day, 7 days, 14 days, 31 days, custom)  😇

docs/ai/mcp/browser-auth-flow.md
```diff
@@ -0,0 +1,99 @@
...
+3. it opens the Shopsys browser authorization page
+4. the administrator allows or denies access
+5. the client exchanges the authorization code for an MCP bearer token
+6. the client uses that bearer token for MCP requests
+
+## Step-by-step
+
+### 1. Authorization server metadata
+
+| Item | Value |
```
grossmannmartin: I would like to improve the formatting a bit... Item/value seems strangely vague. 
Maybe definition list, or at least remove the header completely?

docs/ai/mcp/browser-auth-flow.md
```diff
@@ -0,0 +1,99 @@
...
+| --- | --- |
+| Route name | `mcp_oauth_metadata` |
```
grossmannmartin: some routes are prefixed with admin_superadmin_, some not. is this intentional?

docs/ai/mcp/browser-auth-flow.md
```diff
@@ -0,0 +1,99 @@
+# MCP browser authentication flow
```
grossmannmartin: heading should match the file name (easier to look up the file)

docs/ai/mcp/index.md
```diff
@@ -0,0 +1,198 @@
+# MCP server
```
grossmannmartin: I think this should be a separate article and index.md should be just a table of contents.

docs/ai/mcp/index.md
```diff
@@ -0,0 +1,198 @@
...
+## Available MCP tools
+
+1. `getDatabaseTableNames`
+    Returns the list of exposed database tables available to MCP clients.
+2. `getDatabaseSchema`
+    Returns the generated schema details for selected exposed tables, including columns, primary keys, and foreign keys.
+3. `executeSql`
+    Executes a validated read-only SQL query against the exposed schema through the dedicated read-only connection.
+
+SQL validation uses the [`ext-pg_query`](https://flow-php.com/documentation/components/extensions/pg-query-ext/) PHP extension built on [`libpg_query`](https://github.com/pganalyze/libpg_query), so the query is parsed into a PostgreSQL AST and checked structurally against the exposed schema instead of relying on simple keyword matching.
```
grossmannmartin: I would remove `instead ...`. I don't think we should mention the dead ends we explored during implementation

docs/ai/mcp/index.md
```diff
@@ -0,0 +1,198 @@
...
+The current implementation is restricted to superadministrators even though the flow is otherwise described in general administrator terms.
+
+The page currently provides:
+
+- overview of authorized MCP clients
+- revoke action for connected-client tokens
+- manual token generation and revocation
+- Claude Code setup instructions
+- generic setup instructions for other MCP clients
+
```
grossmannmartin: This is over-describing and will require updates on future app changes

docs/ai/mcp/index.md
```diff
@@ -0,0 +1,198 @@
...
+        "referencedTable": "administrator_role_groups",
+        "referencedColumnNames": ["id"],
+        "onDelete": "NO ACTION"
+      }
+    ]
+  }
+}
+```
+
+This artifact is then used by:
```
grossmannmartin: This whole article imho focuses more on how it works internally. I think the main focus should be on "what it means for the programmer and the end-user"

docs/ai/mcp/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+- `getDatabaseTableNames`
+- `getDatabaseSchema`
+- SQL validation for `executeSql`
+
+## Exposure configuration
+
+MCP exposure is explicit. Doctrine entities and their mapped fields are exposed only when they are marked with MCP attributes.
+
+### Table exposure
```
grossmannmartin: I'm not sure if this section is understandable enough. 

> Doctrine entities and their mapped fields are exposed only when they are marked with MCP attributes

With any mcp attributes? Is AsMcpColumn complementary or replacement for AsMcpTable? How the attributes with the `false` value works? 
And are they mandatory?

docs/ai/mcp/index.md
```diff
@@ -0,0 +1,198 @@
...
+`fieldName` is only valid on the class-level attribute. It must reference an existing mapped field or owning-side association.
+
+## Static enforcement
+
+The exposure coverage is enforced by the custom PHPStan rule `McpEntityExposureAttributeRule`.
+
+That rule checks ORM entities in the `App\\` and `Shopsys\\` namespaces and requires:
+
+- `#[AsMcpTable(exposed: bool)]` on every checked Doctrine entity
+- `#[AsMcpColumn(exposed: bool)]` on every mapped scalar field, embedded field, and relevant owning-side association of an exposed entity
```
grossmannmartin: Ok, this answers most of the questions above, but maybe we could describe the expected setting and then just mention the phpstan rule as an automation?

docs/integration/mcp-browser-auth-flow.md
```diff
@@ -0,0 +1,274 @@
+# MCP browser authentication flow
```
grossmannmartin: file should be named according to the main title

docs/integration/mcp-browser-auth-flow.md
```diff
@@ -0,0 +1,274 @@
...
+2. Claude dynamically registers itself as an OAuth client.
+3. Claude opens the browser with an authorization request.
+4. The administrator approves or denies access in the Shopsys administration.
+5. On approval, Shopsys issues a short-lived authorization code.
+6. Claude exchanges the authorization code for a Shopsys MCP bearer token.
+7. Claude uses that bearer token for subsequent MCP calls.
+
+## Sequence diagram
+
+```mermaid
```
grossmannmartin: we don't have mermaid to visualize sequence diagrams

docs/integration/mcp-browser-auth-flow.md
```diff
@@ -0,0 +1,274 @@
...
+        Claude->>OAuth: MCP requests with Bearer token
+    else Deny access
+        Admin-->>Browser: redirect_uri?error=access_denied&state=...
+        Browser->>Claude: loopback callback
+    end
+```
+
+## Step-by-step
+
+### 1. Claude discovers OAuth metadata
```
grossmannmartin: Isn't this unnecessarily in-depth? Isn't the browser-auth-flow in mcp folder enough?

docs/integration/mcp-browser-auth-flow.md
```diff
@@ -0,0 +1,274 @@
...
+
+Claude first fetches the OAuth authorization-server metadata:
+
+- endpoint: `GET /.well-known/oauth-authorization-server`
+- controller: [McpOAuthMetadataController.php](/home/rosta/Projects/shopsys/packages/mcp/src/Controller/OAuth/McpOAuthMetadataController.php)
```
grossmannmartin: I think the links should point to the github (multiple occurrences, maybe across files)

docs/ai/mcp/index.md
```diff
@@ -0,0 +1,198 @@
...
+#[AsMcpTable]
+class Product
+{
+}
+```
+
+Or explicitly disable exposure:
+
+```php
+#[AsMcpTable(false)]
```
grossmannmartin: I'm not sure about the `false` as the parameter (for example `#[AsMcpTable(false)]`). While reading the entity, I have no clue what this actually does. 

I would use named parameter (like `#[AsMcpTable(exposed: false)]`), or maybe have two separate attributes? `AsMcpTable` and `McpIgnore` (but this has its own set of problems)

packages/mcp/src/Command/GenerateSchemaCommand.php
```diff
@@ -0,0 +1,48 @@
...
+     * {@inheritdoc}
+     */
+    #[Override]
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $schemaFilePath = $this->mcpSchemaFileGenerator->getSchemaFilePath();
+
+        if (!$this->mcpSchemaFileGenerator->generateSchemaFile()) {
+            $io->success(sprintf('MCP schema is up to date: %s', $schemaFilePath));
```
grossmannmartin, line 50: I assume we should run the check on gitlab/github to verify if schema is up to date. We could create `shopsys:mcp:verify-schema` command for that, or make an option to this command that switch behavior/return values (0 ok, 1 generated ?)

packages/mcp/config/services.yaml
```diff
@@ -0,0 +1,34 @@
...
+        resource: '../../**/*{Authenticator,DataFactory,Executor,Facade,Factory,Generator,Hasher,Loader,Lookup,Normalizer,Parser,Provider,Repository,Resolver,Subscriber,Tool,Validator}.php'
+
+    Shopsys\McpBundle\Command\:
+        resource: '../../Command/'
+
+    Shopsys\McpBundle\Component\Database\Schema\ExposedSchemaProvider:
+        arguments:
+            $mcpConnection: '@doctrine.dbal.mcp_connection'
+            $schemaFilePath: '%shopsys_mcp.schema.file_path%'
+
```
grossmannmartin, line 34: What about not having it as a parameter as it's just an internal detail of the mcp bundle?

packages/mcp/src/Component/Database/Schema/SchemaNameNormalizer.php
```diff
@@ -0,0 +1,35 @@
...
+        return $this->normalizeIdentifier($name->getIdentifier());
+    }
+
+    public function normalizeTableName(OptionallyQualifiedName $name): string
+    {
+        return $this->normalizeIdentifier($name->getUnqualifiedName());
+    }
+
+    public function normalizeIdentifier(Identifier $identifier): string
+    {
```
grossmannmartin, line 30: protected?

packages/framework/src/Component/Image/ImageTranslation.php
```diff
@@ -8,21 +8,28 @@
 use Override;
 use Prezent\Doctrine\Translatable\Attribute as Prezent;
 use Prezent\Doctrine\Translatable\Entity\AbstractTranslation;
+use Shopsys\McpAttributes\Attribute\AsMcpColumn;
+use Shopsys\McpAttributes\Attribute\AsMcpTable;
 
+#[AsMcpTable]
+#[AsMcpColumn(fieldName: 'id')]
+#[AsMcpColumn(fieldName: 'locale')]
 #[ORM\Table(name: 'images_translations')]
```
grossmannmartin, line 18: Just slight concern - the AsMcpColumn behaves very differently when put above class and property. Even some parameter become mandatory. 

Wouldn't be better to have:
- an extra attribute for class-defined columns?
- option with extra fields directly on AsMcpTable?

project-base/app/src/Model/Customer/User/CustomerUserRefreshTokenChain.php
```diff
@@ -6,12 +6,14 @@
...
 use Shopsys\FrameworkBundle\Model\Customer\User\CustomerUserRefreshTokenChain as BaseCustomerUserRefreshTokenChain;
+use Shopsys\McpAttributes\Attribute\AsMcpTable;
 
 /**
  * @property \App\Model\Customer\User\CustomerUser $customerUser
  * @property \App\Model\Administrator\Administrator|null $administrator
  * @method \App\Model\Administrator\Administrator|null getAdministrator()
  */
+#[AsMcpTable(false)]
 #[ORM\Table(name: 'customer_user_refresh_token_chain')]
```
grossmannmartin, line 17: The default behavior is not inherited from the parent class? Should it be? Can it be?

packages/framework/src/Model/Product/Unit/UnitTranslation.php
```diff
@@ -8,21 +8,28 @@
 use Override;
 use Prezent\Doctrine\Translatable\Attribute as Prezent;
 use Prezent\Doctrine\Translatable\Entity\AbstractTranslation;
+use Shopsys\McpAttributes\Attribute\AsMcpColumn;
+use Shopsys\McpAttributes\Attribute\AsMcpTable;
 
+#[AsMcpTable]
+#[AsMcpColumn(fieldName: 'id')]
+#[AsMcpColumn(fieldName: 'locale')]
 #[ORM\Table(name: 'unit_translations')]
```
grossmannmartin, line 18: Are those two fields reported by phpstan when missing Mcp attribute? 

If I saw correctly, this is used only for translatable entity. I understand this is more general solution, but for everyday use, wouldn't be better to have some shortcut? 
For example something like `#[AsMcpTranslatableTable]` (instead of AsMcpTable)

packages/php-image/Dockerfile
```diff
@@ -49,18 +49,23 @@ RUN apk add --no-cache --virtual .build-deps \
...
+    protobuf-c-dev \
     rabbitmq-c-dev && \
+    curl -fsSL https://github.com/php/pie/releases/latest/download/pie.phar -o /usr/local/bin/pie && \
+    chmod +x /usr/local/bin/pie && \
     pecl install redis && \
     docker-php-ext-enable redis && \
     pecl install amqp && \
     docker-php-ext-enable amqp && \
+    pie install flow-php/pg-query-ext && \
     docker-php-ext-configure gd --with-freetype --with-jpeg && \
```
grossmannmartin: please mention all installed packages in the comment above, so we know why we have it

packages/php-image/Dockerfile
```diff
@@ -49,18 +49,23 @@ RUN apk add --no-cache --virtual .build-deps \
     autoconf \
     freetype-dev \
     g++ \
+    git \
     icu-dev \
```
grossmannmartin: git is already added later. It should not be mentioned twice

packages/php-image/Dockerfile
```diff
@@ -49,18 +49,23 @@ RUN apk add --no-cache --virtual .build-deps \
...
     jpeg-dev \
     libpng-dev \
     libpq-dev \
     libzip-dev \
     make \
     openssl-dev \
+    protobuf-c-dev \
     rabbitmq-c-dev && \
```
grossmannmartin: are protobuf-c-dev and protobuf-c needed at all?

packages/php-image/Dockerfile
grossmannmartin: Build of php image warns about some missing build dependencies 

```
#13 39.24 🥧 PHP Installer for Extensions (PIE) 1.4.2, from The PHP Foundation
#13 39.35 You are running PHP 8.5.5
#13 39.37 Target PHP installation: 8.5.5 nts, on Linux/OSX/etc x86_64 (from /usr/local/bin/php)
#13 39.53 The following build tools are missing: bison, re2c, libtoolize/glibtoolize
#13 39.53 You are not running in interactive mode, and you did not provide the --auto-install-build-tools flag.
#13 39.53 You may need to run: apk add --no-cache --virtual .php-pie-deps bison re2c libtool
```

see https://github.com/shopsys/shopsys/actions/runs/24983057151/job/73149612855#step:4:633

I'm not sure if they're necessary. 

Also, I would compare the size of the image before and after this PR. To avoid accidental installation of some huge package

packages/mcp/src/DependencyInjection/ShopsysMcpExtension.php
```diff
@@ -0,0 +1,37 @@
...
+
+    #[Override]
+    public function load(array $configs, ContainerBuilder $container): void
+    {
+        $configuration = new Configuration();
+        $config = $this->processConfiguration($configuration, $configs);
+
+        $container->setParameter('shopsys_mcp.query.statement_timeout_ms', $config['query']['statement_timeout_ms']);
+        $container->setParameter('shopsys_mcp.query.lock_timeout_ms', $config['query']['lock_timeout_ms']);
+        $loader = new YamlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
```
grossmannmartin: I would like to avoid setting the parameter and set the values to the services directly. It should be fairly easy with the AbstractBundle (yes, this is the comment about AbstractBundle I teased about earlier)

packages/administration/templates/partial/flash_message/_messages.html.twig
```diff
@@ -3,7 +3,7 @@
         <div>
             {{ ux_icon(icon, {class: 'icon alert-icon'}) }}
         </div>
-        <div>
+        <div class="flex-fill">
             {% for message in messages %}
```
grossmannmartin, line 7: flash messages are supposed to change in #4546 to toast-like style. I think it contradicts with the desired usage here (to render the copyable secret)

project-base/docker/nginx/nginx.conf
```diff
@@ -68,6 +68,18 @@ server {
...
+    location ~ ^/_mcp(?:/|$) {
+        return 470; # send to @app
+    }
+
+    location = /.well-known/oauth-authorization-server {
+        return 470; # send to @app
+    }
+
+    location ~ ^/mcp/oauth(?:/|$) {
+        return 470; # send to @app
```
grossmannmartin, line 85: shouldn't it be `_mcp/...`? also, in that case, the first _mcp location should cover it

grossmannmartin (changes requested):

packages/mcp/src/Component/Logger/McpRequestLogProcessor.php
```diff
@@ -0,0 +1,39 @@
...
+        $administratorMcpToken = $request->attributes->get(McpTokenAuthenticator::REQUEST_ATTRIBUTE_ADMINISTRATOR_MCP_TOKEN);
+
+        if (!$administratorMcpToken instanceof AdministratorMcpToken) {
+            return $record;
+        }
+
+        $record['extra']['administrator_id'] = $administratorMcpToken->getAdministrator()->getId();
+        $record['extra']['administrator_username'] = $administratorMcpToken->getAdministrator()->getUsername();
+
+        return $record;
```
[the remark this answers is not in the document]
vitek-rostislav: solved in 6409056423

packages/framework/src/Component/ArrayUtils/ArrayHelper.php
```diff
@@ -10,4 +10,30 @@ public static function haveArraysDifferentValues(array $array1, array $array2):
...
+    }
+
+    /**
+     * @param array<mixed> $data
+     * @return array<string>
+     */
+    public static function getArrayOrEmpty(array $data, string $key): array
+    {
+        return is_array($data[$key] ?? null) ? $data[$key] : [];
+    }
```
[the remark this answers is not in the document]
vitek-rostislav: solved in 258744d33e

packages/mcp/src/Component/Database/Schema/McpSchemaFileGenerator.php
```diff
@@ -0,0 +1,45 @@
...
+        $existingSchemaJson = is_file($this->schemaFilePath) ? file_get_contents($this->schemaFilePath) : false;
+
+        if ($existingSchemaJson === $generatedSchemaJson) {
+            return false;
+        }
+
+        if (@file_put_contents($this->schemaFilePath, $generatedSchemaJson) === false) {
+            throw new RuntimeException(sprintf(
+                'Generated MCP schema file could not be written: %s.',
+                $this->schemaFilePath,
```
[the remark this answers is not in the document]
vitek-rostislav: reworked within b9f917ee9a

project-base/app/deploy/deploy-project.sh
```diff
@@ -32,6 +32,8 @@ function deploy() {
         ["DATABASE_NAME"]=${PROJECT_NAME}
         ["DATABASE_USER"]=${PROJECT_NAME}
         ["DATABASE_PASSWORD"]=${POSTGRES_DATABASE_PASSWORD}
+        ["MCP_DATABASE_USER"]=${MCP_DATABASE_USER:-${PROJECT_NAME}_mcp}
+        ["MCP_DATABASE_PASSWORD"]=${MCP_DATABASE_PASSWORD:-}
```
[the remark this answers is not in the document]
vitek-rostislav: I do not want to change it, we will ensure the user and proper password exists in Gitlab ENV variables

composer.json
vitek-rostislav: yes, good catch :+1:  7165490739

project-base/app/config/packages/mcp.yaml
vitek-rostislav: :+1: e30250368f

packages/mcp/src/Component/Security/McpTokenAuthenticator.php
vitek-rostislav: I am not able to do that, I would not satisfy the protocol if the header was different :frowning_face: see https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#token-requirements

I think I can live with the current state (http auth is not required when accessing the web from a whitelisted IP, e.g., from VPN by default, if I am not mistaken - that is good enough for me, but I am open to discussion, of course).

packages/mcp/src/Resources/config/services.yaml
vitek-rostislav: well we do - `resource: '../../**/*{...}.php'` should translate to `src/**/*...php`, or?

packages/mcp/src/Migrations/Version20260401100401.php
vitek-rostislav: it was reworked in 2620fa0ae1 and the lookups should be ok now, see the analysis :innocent: 

```
The current lookup paths on administrator_mcp_tokens are only these three:

  1. findActiveByPublicTokenId(string $publicTokenId, DateTimeImmutable $dateTime)

  - used during bearer token authentication
  - filters by:
      - public_token_id = ?
      - revoked_at IS NULL
      - expires_at > ?
  - the existing unique index on public_token_id is sufficient
  - reason:
      - it finds at most one row
      - revoked_at and expires_at are then just post-filters on that single candidate

  2. findActiveTokensByAdministrator(Administrator $administrator, DateTimeImmutable $dateTime)

  - used for the admin token overview
  - filters by:
      - administrator_id = ?
      - revoked_at IS NULL
      - expires_at > ?
      - ORDER BY created_at DESC
  - the existing index on administrator_id is the relevant one here
  - reason:
      - this is the only query that may return multiple rows
  - I would not add anything more at this point:
      - there is no evidence yet that token counts per administrator are large enough to justify a composite index

  3. findActiveByIdAndAdministrator(Administrator $administrator, int $id, DateTimeImmutable $dateTime)

  - used in the admin revoke flow
  - filters by:
      - id = ?
      - administrator_id = ?
      - revoked_at IS NULL
      - expires_at > ?
  - the primary key on id is sufficient
  - reason:
      - id is unique, so the database finds a single row and only validates the remaining conditions
```

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpTokenFacade.php
vitek-rostislav: removed within 2620fa0ae1

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpTokenRepository.php
vitek-rostislav: :+1: added in 2620fa0ae1

packages/administration/assets/src/js/utils/confirmWindow.js
vitek-rostislav: after reworking the whole thing in 2620fa0ae1, yes this is unnecessary and I will try not to forget to drop that commit during future rebase :smile: Lets keep this thread open until then

packages/mcp/src/Controller/Admin/McpServerController.php
vitek-rostislav: reworked in 2620fa0ae1, custom expiration will be added shortly (tomorrow :smile: )

vitek-rostislav: ok, here comes the enhanced custom token settings, easypick :smile: fdeb39bc09

docs/integration/mcp-browser-auth-flow.md
vitek-rostislav: the file was just a leftover, sorry for that, removed in 9378239f2c

docs/integration/mcp-browser-auth-flow.md
vitek-rostislav: the file was just a leftover, sorry for that, removed in 9378239f2c

docs/integration/mcp-browser-auth-flow.md
vitek-rostislav: the file was just a leftover, sorry for that, removed in 9378239f2c

docs/integration/mcp-browser-auth-flow.md
vitek-rostislav: the file was just a leftover, sorry for that, removed in 9378239f2c

packages/mcp/src/Command/GenerateSchemaCommand.php
vitek-rostislav: the file is gitignored :slightly_smiling_face:

packages/mcp/config/services.yaml
vitek-rostislav: yes, reworked within b9f917ee9abdb6e14847cd86b1aaae0a90d5c555

packages/php-image/Dockerfile
vitek-rostislav: Well I followed some instructions when adding it, now the backwards analysis says:
```
Yes, both packages are needed ? at least according to the current config.m4 in flow-php/pg-query-ext (1.x):                                                                        
                                                                                                                                                                                     
  What ext/config.m4 does:                                                                                                                                                           
                                                                                                                                                                                     
  1. libpg_query.a is linked statically (LDFLAGS="$LDFLAGS .../libpg_query.a").                                                                                                      
  2. For a shared build (which is what pie/pecl always produce), the script additionally adds PHP_ADD_LIBRARY(protobuf-c, ...). The comment in the config says it explicitly:        
  "protobuf-c is required for shared builds (libpg_query.a needs it)".                                                                                                               
  3. It uses pkg-config --libs libprotobuf-c for detection ? which needs protobuf-c-dev (that's where the .pc file and headers live), otherwise it falls back to a weaker search for
  .so in /usr/local, /usr, etc.                                                                                                                                                      
                                                                       
  Implication for our Dockerfile:                                                                                                                                                    
                                                                       
  - protobuf-c-dev (build-time): required for pkg-config during pie install; without it the dynamic linking gets messy.                                                              
  - protobuf-c (runtime): required because the resulting pg_query.so has libprotobuf-c.so as a dynamic dependency; without it, php fails to load the extension.
```

Hope you do not mind I copy paste these agent answers here :innocent:

packages/php-image/Dockerfile
vitek-rostislav: :+1: done in e3e55fea2c

packages/php-image/Dockerfile
vitek-rostislav: yes, fixed in e3e55fea2c

packages/php-image/Dockerfile
vitek-rostislav: They say the warnings are rather suggestions for tools that are actually not needed for us :shrug:

project-base/docker/nginx/nginx.conf
vitek-rostislav: it is because of the firewall (see `security.yaml`) -  `_mcp` location is behind the mcp firewall (requires bearer token), while `/mcp/oauth` must not be behind that firewall as it is used for the auth flow to get the token :slightly_smiling_face:

packages/php-image/Dockerfile
vitek-rostislav: regarding the sizes, it is `355 MB` for the current `19.0` image vs `358 MB` now :slightly_smiling_face: Here is a breakdown we can yet discuss - the interesting thing is the removed git from runtime, I am not sure whether it is ok, but Claude tells me it is ok :smile: 

image

docs/ai/mcp/browser-auth-flow.md
vitek-rostislav: yes yes, some of the routes are intentionally without the prefix because they are used within the oath flow, basically outside of the administration, or as my friend says:

- `admin_superadmin_mcp_oauth_authorize`
      - this is the browser consent page inside Shopsys administration
      - it is an admin route
      - it is restricted to superadmins
- `mcp_oauth_metadata`
- `mcp_oauth_register`
- `mcp_oauth_token`
      - these are not admin pages at all
      - they are public OAuth protocol endpoints used by the MCP client itself
      - they must be reachable before the client has an admin session or bearer token
      - so they intentionally do not live under the admin route namespace

docs/ai/mcp/index.md
vitek-rostislav: docs cleaned up in 9bef211c4b

docs/ai/mcp/index.md
vitek-rostislav: docs cleaned up in 9bef211c4b

docs/ai/mcp/index.md
vitek-rostislav: docs cleaned up in 9bef211c4b

docs/ai/mcp/index.md
vitek-rostislav: docs cleaned up in 9bef211c4b

docs/ai/mcp/index.md
vitek-rostislav: docs cleaned up in 9bef211c4b

docs/ai/mcp/index.md
vitek-rostislav: docs cleaned up in 9bef211c4b

docs/ai/mcp/browser-auth-flow.md
vitek-rostislav: docs cleaned up in 9bef211c4b

docs/ai/mcp/browser-auth-flow.md
vitek-rostislav: docs cleaned up in 9bef211c4b

project-base/app/src/Model/Customer/User/CustomerUserRefreshTokenChain.php
vitek-rostislav: No, it is not. It's the same as ORM attributes - we also have to copy them when extending an entity so I figured this is the default state that I am ok with, as it is aligned and explicit

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpToken.php
vitek-rostislav: makes sense, refactored in 69560113eb

packages/mcp/src/Model/Administrator/McpToken/AdministratorMcpTokenLookup.php
vitek-rostislav: makes sense, I just want to keep `AdministratorMcpIssuedToken::getTokenString()` for convenience, but now it is just a proxy to the new `McpBearerToken::createTokenString()`

so, d46d9f372d - better?

packages/mcp/src/DependencyInjection/ShopsysMcpExtension.php
vitek-rostislav: ok, reworked in 35165c567f but I am not sure whether this is the desired state, so we can discuss it further

packages/mcp/src/ShopsysMcpBundle.php
vitek-rostislav: ok, reworked in 35165c567f but I am not sure whether this is the desired state, so we can discuss it further

docs/ai/mcp/index.md
vitek-rostislav: I have added the named parameters for these in edb65c278c

packages/mcp/src/Component/Database/Schema/SchemaNameNormalizer.php
vitek-rostislav: yes, 230b207f38

packages/administration/templates/partial/flash_message/_messages.html.twig
vitek-rostislav: thanks for the heads up, reworked in 75fa2decd2

packages/framework/src/Model/Product/Unit/UnitTranslation.php
vitek-rostislav: yes, they are reported by phpstan when missing the attribute. I have reworked the attributes a little in 18ffdd4264, however, I have not added `AsMcpTranslatableTable` shortcut in the end. I think it would complicate things in the background, as well as the phpstan rule, tests and docs. I am open to discussion of course

packages/framework/src/Component/Image/ImageTranslation.php
vitek-rostislav: yes, good point, I have refactored that a bit in 18ffdd4264 - a new attribute for this purpose was created

packages/mcp/src/Resources/config/services.yaml
grossmannmartin: I don't think so, but it's rewritten during simplifying to abstract bundle, so good job 👍😁

packages/php-image/Dockerfile
grossmannmartin: Well, he's not wrong, but only because we install git extra in the `project-base/app/docker/php-fpm/Dockerfile` in the development stage. 

That way, we still get git installed where needed (during release, some phing targets depend on git), but the production image will not have it. 
I think it's ok.

packages/administration/assets/src/js/utils/modalWindow.js
```diff
@@ -34,14 +36,16 @@ export default class ModalWindow {
             buttons: [],
             style: null,
             borderless: false,
+            backdrop: true,
+            keyboard: true,
```
grossmannmartin: Personally, I would use just a single boolean configuration value. I don't think there will be a use case for disallowing backdrop click, but allowing escape dismissal. 


If you disagree, I am strongly inclined to rename the keyboard property, as it's too general, and it's not clear what this does.
Also, the backdrop should be boolean instead of boolean|string, because ` ${this.options.backdrop === 'static' ? 'data-bs-backdrop="static"' : ''` does not feel right

project-base/app/config/packages/monolog.yaml
```diff
@@ -41,6 +43,14 @@ monolog:
             path: "%shopsys.log_stream%"
             bubble: false
             channels: [ "cron", "queue" ]
+        mcp:
```
grossmannmartin, line 46: please verify with @henzigo if it's not necessary to add this somewhere into our production logging stack

project-base/app/webpack.config.js
```diff
@@ -73,6 +73,7 @@ Encore
 
 Encore.addAliases({
     'framework': '@shopsys/framework/js',
+    'mcp': path.resolve(path.join(__dirname, sources.getPackageNodeModulesDir('mcp'), 'js')),
```
grossmannmartin: Idea: 
Isn't it possible to get it working without an alias? We probably had to create the mcp's package.json file, right?

packages/mcp/translations/messages.cs.po
```diff
@@ -0,0 +1,161 @@
...
+msgid "Expires at"
+msgstr "Vyprší dne"
+
+msgid "Export the manual token in your current shell, then add the MCP server once."
+msgstr "Exportujte manuální token v aktuálním shellu a poté jednou přidejte MCP server."
+
+msgid "Generate MCP token"
+msgstr "Vygenerovat MCP token"
+
+msgid "Generated tokens are shown only once. Regenerating a token revokes the previous token immediately."
```
grossmannmartin: it seems like we need to dump translations

packages/mcp/src/Form/Admin/Mcp/McpOauthAuthorizationFormType.php
```diff
@@ -0,0 +1,68 @@
...
+                ],
+            ]);
+    }
+
+    #[Override]
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefaults([
+            'attr' => [
+                'class' => 'd-flex gap-2',
```
grossmannmartin: I would avoid the visual tuning in the form types if possible and rather placed it to the template

packages/mcp/templates/content/superadmin/_codexSetup.html.twig
```diff
@@ -0,0 +1,38 @@
...
+                </span>
+            </div>
+        </div>
+
+        <p class="text-muted mb-3">{{ 'Environment variables exported this way are available only in the current shell session unless you add them to your shell profile.'|trans }}</p>
+
+        <div>
+            <label class="form-label">{{ 'One-time setup command'|trans }}</label>
+            <div class="input-group input-group-flat w-100">
+                <input type="text" value="{{ codexAddCommand }}" class="form-control font-monospace" readonly>
```
grossmannmartin: The instructions are not fully visible in the input. 
What about using pre like here https://preview.tabler.io/logs.html ?

But we would have to figure out how to position the copy button. Personally, I would like something like on the documentation page 

for example

Image

What do you think?

packages/mcp/templates/content/superadmin/listGrid.html.twig
```diff
@@ -0,0 +1,57 @@
...
+{% block grid_value_cell_id_lastUsedAt %}
+    {% if value is null %}
+        {{ 'Never used'|trans }}
+    {% else %}
+        {{ value|formatDateTime }}
+    {% endif %}
+{% endblock %}
+
+{% block grid_action_cell_type_delete %}
+    {% if row.amt.revokedAt is null and row.amt.expiresAt|date('U') > 'now'|date('U') %}
```
grossmannmartin, line 41: For some reason, I was not able to delete the manually generated token. Can you verify it, please?

grossmannmartin (changes requested):

packages/administration/assets/src/js/utils/confirmWindow.js
vitek-rostislav: dropped during rebase :heavy_check_mark:

packages/administration/assets/src/js/utils/modalWindow.js
vitek-rostislav: good point :+1: what about this? 49d48f8676

project-base/app/config/packages/monolog.yaml
vitek-rostislav: question raised, answer is pending :slightly_smiling_face:

project-base/app/webpack.config.js
vitek-rostislav: yes, great idea! fixed in eb81f1965706ade42d54cf7814f85a57f35ba395 and also added biome in 2e5c66719130cc4bd1fee935c9687e3e1c7b8acd

packages/mcp/templates/content/superadmin/listGrid.html.twig
vitek-rostislav: fixed in d54a96dfb4

packages/mcp/templates/content/superadmin/_codexSetup.html.twig
vitek-rostislav: good idea, implemented in e8a475451b

packages/mcp/src/Form/Admin/Mcp/McpOauthAuthorizationFormType.php
vitek-rostislav: yes, good catch, fixed in 262a5a6446

packages/php-image/Dockerfile
vitek-rostislav: hm, in the end, it looks like we have to keep the git there as well, see https://github.com/shopsys/shopsys/actions/runs/25800334624/job/75788393472?pr=4564

Sooo, db16d6dab9. I am open to discussion as this kind of reverts one of the previous fixups based on your earlier comment, right?

packages/mcp/translations/messages.cs.po
vitek-rostislav: ah, yes, fixed in b92388fc6f and 4102415caa, thanks :+1:

project-base/app/webpack.config.js
grossmannmartin: love it!

grossmannmartin (approved):

project-base/app/config/packages/monolog.yaml
henzigo: Logs are formatted as `json` in the production environment, so we don't need to adjust anything :)

packages/mcp/src/Component/Logger/McpToolCallLogger.php
```diff
@@ -0,0 +1,36 @@
...
+
+        if ($inputContext !== []) {
+            $context['input'] = $inputContext;
+        }
+
+        if ($resultContext !== []) {
+            $context['result'] = $resultContext;
+        }
+
+        $this->logger->info('MCP tool executed', $context);
```
henzigo, line 34: Should we log those messages as `info`?

vitek-rostislav: why not? :eyes:

packages/mcp/templates/content/superadmin/_codexSetup.html.twig
```diff
@@ -1,11 +1,7 @@
 {% import '@ShopsysAdministration/partial/copyable_code_block.html.twig' as copyableCodeBlock %}
 
 {% set codexExportCommand = 'export ' ~ bearerTokenEnvVarName ~ '="' ~ (tokenString ?? '<token>') ~ '"' %}
-{%- set codexAddCommand -%}
-codex mcp add {{ mcpServerName }} \
---url {{ mcpEndpointUrl }} \
---bearer-token-env-var {{ bearerTokenEnvVarName }}
-{%- endset -%}
+{% set codexAddCommand = 'codex mcp add ' ~ mcpServerName ~ ' --url ' ~ mcpEndpointUrl ~ ' --bearer-token-env-var ' ~ bearerTokenEnvVarName %}
```
grossmannmartin: Even though it's not necessary to manually wrap the command, I think the previous version looked better. 

Now: 

Image

previously:

Image

grossmannmartin (approved):

packages/mcp/templates/content/superadmin/_codexSetup.html.twig
vitek-rostislav: ok, reverted in 2c01f1950a

docs/administration/administration-menu.md
```diff
@@ -74,3 +74,39 @@ Routes with mandatory route parameters but without concrete values usually have
...
+                'puzzle',
+                '/admin/custom-page/',
+                0,
+            ),
+        ];
+    }
+}
+```
+
+The item consists of a title, an icon alias, a link, and a priority.
```
grossmannmartin, line 108: I don't think we should use the path. 
A. It's possible to change the admin url in the future
B. We don't have a formal way to detect usage, so when I delete/rename an action, this may become obsolete.

edit: oh, I see now that in the actual code we use UrlGenerator. I think the documentation should reflect it

packages/administration/src/Component/Menu/Dropdown/AdminDropdownMenuItemsRenderer.php
```diff
@@ -0,0 +1,49 @@
...
+        protected readonly Environment $twigEnvironment,
+    ) {
+    }
+
+    public function render(): string
+    {
+        $items = $this->getItems();
+
+        return $this->twigEnvironment->render('@ShopsysAdministration/partial/admin_dropdown_menu_items.html.twig', [
+            'items' => $items,
```
grossmannmartin: nitpick - I find this variable superfluous

docs/administration/administration-menu.md
```diff
@@ -74,3 +74,39 @@ Routes with mandatory route parameters but without concrete values usually have
...
+Higher priority items are rendered first.
+If the item should be visually separated from the previous items, set the optional `renderDividerBefore` constructor argument to `true`.
+
+Access checks should be handled inside the provider itself by returning an empty array for users that should not see the item.
```
grossmannmartin: `By returning an empty array` may be misleading. If I want to render two items, one for superadmin and the second for all, I don't return an empty array, but simply not include forbidden item

grossmannmartin (approved):

PR #4574, opened 2026-04-22
fixed some next static files not being served in prod / devel

TomasLudvik:
- Not set CORS for static next files caused fonts not being downloaded

----
  - https://tl-fix-cors-headers.odin.shopsys.cloud
  - https://cz.tl-fix-cors-headers.odin.shopsys.cloud
  - https://tl-fix-cors-headers.odin.shopsys.cloud/sk

----
  - https://tl-fix-cors-headers.odin.shopsys.cloud
  - https://cz.tl-fix-cors-headers.odin.shopsys.cloud
  - https://tl-fix-cors-headers.odin.shopsys.cloud/sk

### DIFF
--- a/project-base/app/orchestration/kubernetes/configmap/nginx.yaml
+++ b/project-base/app/orchestration/kubernetes/configmap/nginx.yaml
@@ -261,7 +261,6 @@ data:
 
             location ^~ /_next/static/ {
                 add_header Access-Control-Allow-Origin "*" always;
-                add_header Access-Control-Allow-Credentials "false" always;
                 add_header X-Powered-By "Shopsys Platform" always;
                 add_header Referrer-Policy "strict-origin-when-cross-origin" always;
                 add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
--- a/project-base/app/orchestration/kubernetes/configmap/nginx.yaml
+++ b/project-base/app/orchestration/kubernetes/configmap/nginx.yaml
@@ -259,6 +259,25 @@ data:
                 try_files $uri @app;
             }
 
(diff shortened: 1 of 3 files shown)


project-base/app/orchestration/kubernetes/configmap/nginx.yaml
```diff
@@ -259,6 +259,25 @@ data:
                 try_files $uri @app;
             }
 
+            location ^~ /_next/static/ {
```
grossmannmartin: Instead of duplicating many lines in the config, I suggest adding a map for conditional add_header. 

In the http context, I would add a map (in the nginx.yaml, there is a map already; this one should go right next  to it)

```
map $request_uri $static_cors {
      default "";
      ~^/_next/static/ "*";
  }
```

then in the `@storefront` location 

```
add_header Access-Control-Allow-Origin $static_cors always;
```

The whole thing works like this:
- if request_uri match the `~^/_next/static/`, set `*` to `$static_cors`
  - if not, set an empty string
- `@storefront` adds the header Access-Control-Allow-Origin with the previously set value
  - If the value is an empty string, nginx doesn't add this header at all (i.e., add_header with empty value is no-op)


Note 1: Fonts are broken only in Chrome/Firefox (for the sake of future testing)
Note 2: Tested on devel
Image
Note 3: It's necessary to clear the font from the CDN for this change to take effect (otherwise CDN will still use previously cached font without cors set)

grossmannmartin (changes requested):

grossmannmartin (approved):

pk16011990 (approved):

PR #4584, opened 2026-04-29
[CRUD] Create and edit actions

henzigo:

This PR adds full create and edit functionality to the CRUD controller framework. Until now, CRUD controllers only supported listing and deleting entities — administrators had no way to create or edit
  records through the generic CRUD system and had to fall back to custom controllers.                                                                                                                            
   
  With this change, developers can implement `EditHandlerInterface`, `CreateHandlerInterface`, or the combined `CrudHandlerInterface` to enable form-based create and edit pages. Forms are configured through a new   
  `CrudFormConfigurator` that supports two approaches: using an existing Symfony `FormType` class, or building the form inline with a builder. The configurator automatically provides the current action type
  (create/edit) to both the controller and the form, making it easy to adjust form behavior based on the operation being performed. Extensions can hook into the create/edit lifecycle with                      
  `beforeCreate/afterCreate/onCreateError` and `beforeEdit/afterEdit/onEditError` callbacks, and can also extend forms when the builder mode is used.

  An example `AdministratorCrudController` in project-base demonstrates the full setup including a CRUD handler and form configuration.

#### Fixes issues
... 

----
  - https://jg-crud-forms.odin.shopsys.cloud
  - https://cz.jg-crud-forms.odin.shopsys.cloud
  - https://jg-crud-forms.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/src/Component/Crud/Handler/CreateHandlerInterface.php
+++ b/packages/administration/src/Component/Crud/Handler/CreateHandlerInterface.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\Component\Crud\Handler;
+
+/**
+ * Handler interface for create operations
+ *
+ * This interface extends ReadHandlerInterface and adds create functionality for CRUD controllers.
+ */
+interface CreateHandlerInterface extends ReadHandlerInterface
+{
+    /**
+     * Creates an empty data transfer object with default values
+     *
+     * @return object The new data transfer object
+     */
+    public function createData(): object;
+
+    /**
+     * Creates a new entity from the provided data
+     *
+     * @param object $data The data transfer object with values for the new entity
+     * @return \Shopsys\FrameworkBundle\Component\Utils\Presentable The newly created entity
+     */
+    public function create(object $data): object;
+}
--- a/packages/administration/src/Controller/AbstractCrudController.php
+++ b/packages/administration/src/Controller/AbstractCrudController.php
@@ -115,7 +115,7 @@ abstract class AbstractCrudController extends AdminBaseController
         $entity = $handler->getById($id);
         $data = $handler->createDataFromEntity($entity);
 
-        $formConfigurator = new CrudFormConfigurator($this->formFactory, $data);
+        $formConfigurator = new CrudFormConfigurator($this->formFactory, $data, ActionType::EDIT);
         $this->configureForm($formConfigurator, $entity);
         $this->executeExtensions(fn (AbstractCrudControllerExtension $extension) => $extension->configureForm($formConfigurator, $entity));
 
@@ -184,7 +184,7 @@ abstract class AbstractCrudController extends AdminBaseController
         $handler = $this->definition->getHandlerForAction(ActionType::CREATE);
         $data = $handler->createData();
 
-        $formConfigurator = new CrudFormConfigurator($this->formFactory, $data);
+        $formConfigurator = new CrudFormConfigurator($this->formFactory, $data, ActionType::CREATE);
         $this->configureForm($formConfigurator, null);
         $this->executeExtensions(fn (AbstractCrudControllerExtension $extension) => $extension->configureForm($formConfigurator, null));
 
--- a/packages/administration/src/Controller/AbstractCrudController.php
+++ b/packages/administration/src/Controller/AbstractCrudController.php
@@ -11,7 +11,11 @@ use Shopsys\AdministrationBundle\Component\Config\ActionsConfig;
 use Shopsys\AdministrationBundle\Component\Config\ActionType;
 use Shopsys\AdministrationBundle\Component\Config\CrudConfig;
 use Shopsys\AdministrationBundle\Component\Crud\Definition;
+use Shopsys\AdministrationBundle\Component\Crud\Extension\CrudCreateHookExtensionInterface;
 use Shopsys\AdministrationBundle\Component\Crud\Extension\CrudDeleteHookExtensionInterface;
+use Shopsys\AdministrationBundle\Component\Crud\Extension\CrudEditHookExtensionInterface;
+use Shopsys\AdministrationBundle\Component\Crud\Handler\CreateHandlerInterface;
+use Shopsys\AdministrationBundle\Component\Crud\Handler\EditHandlerInterface;
 use Shopsys\AdministrationBundle\Component\Crud\Helper\CrudTransformationHelper;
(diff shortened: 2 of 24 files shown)


project-base/app/src/Controller/Admin/AdministratorCrudController.php
```diff
@@ -0,0 +1,52 @@
...
+use Override;
+use Shopsys\AdministrationBundle\Component\Attributes\CrudController;
+use Shopsys\AdministrationBundle\Component\Config\CrudConfig;
+use Shopsys\AdministrationBundle\Component\Crud\Form\CrudFormConfigurator;
+use Shopsys\AdministrationBundle\Component\Datagrid\Datagrid;
+use Shopsys\AdministrationBundle\Controller\AbstractCrudController;
+use Shopsys\FrameworkBundle\Form\Admin\Administrator\AdministratorFormType;
+
+#[CrudController(entityClass: Administrator::class)]
+final class AdministratorCrudController extends AbstractCrudController
```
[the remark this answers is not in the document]
henzigo: RBAC attributes should not be used for CrudControllers. Roles are generated automatically for them. So those Attributes are mandatory only for normal controllers

packages/administration/src/Component/Crud/Handler/CreateHandlerInterface.php
```diff
@@ -0,0 +1,28 @@
...
+     *
+     * @return object The new data transfer object
+     */
+    public function createData(): object;
+
+    /**
+     * Creates a new entity from the provided data
+     *
+     * @param object $data The data transfer object with values for the new entity
+     * @return object&\Shopsys\FrameworkBundle\Component\Utils\Presentable The newly created entity
```
vitek-rostislav: why do we have `object&\Shopsys\FrameworkBundle\Component\Utils\Presentable` here? It feels redundant. Could not we remove it and change the return typehint directly to `Presentable` (or just `object` return typehint?)

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -93,19 +103,137 @@ public function detailAction(int $id): Response
         ]);
     }
 
-    public function editAction(int $id): Response
+    public function editAction(Request $request, int $id): Response
     {
+        /** @var \Shopsys\AdministrationBundle\Component\Crud\Handler\EditHandlerInterface $handler */
+        $handler = $this->definition->getHandlerForAction(ActionType::EDIT);
+        $entity = $handler->getById($id);
```
vitek-rostislav, line 115: we could add proxy method into Definition that would have precise return typehint -> we would not need inline annotation then. Something like `getHandlerForEditAction(): EditHandlerInterface`, what do you think?

vitek-rostislav: (the same for `newAction` below)

packages/administration/src/Controller/AbstractCrudController.php
```diff
@@ -93,19 +105,137 @@ public function detailAction(int $id): Response
...
+        $data = $handler->createDataFromEntity($entity);
+
+        $formConfigurator = new CrudFormConfigurator($this->formFactory, $data, ActionType::EDIT);
```
vitek-rostislav, line 118: I find the solution with configurator unnecessarily complicated, I think we do not need the configurator at all, as we can use `FormBuilderInterface` directly. Check this WIP - https://github.com/shopsys/shopsys/commit/8a08abaf534477e6878096015f68b272fd2956d1 - we can talk about it - I think it simplifies the whole thing heavily.

vitek-rostislav: I would yet then rename `AbstractCrudControllerExtension::configureForm` to something like `extendForm`

vitek-rostislav: yes, I know you kind of had that approach earlier in this branch

vitek-rostislav (changes requested):

packages/administration/src/Component/Crud/Handler/CreateHandlerInterface.php
henzigo: Good catch — the intersection type is indeed redundant. `Definition::__construct` already requires `entityClass` to implement `Presentable` whenever a handler implementing `ReadHandlerInterface` is registered. I've removed it, the method now returns `Presentable`.

packages/administration/src/Controller/AbstractCrudController.php
henzigo: The mapping between `ActionType` and the handler interface is already encoded in the `ActionType` enum (via `handlerInterface()`). Adding `getHandlerForEditAction()` / `getHandlerForCreateAction()` would duplicate that mapping and grow the API surface for every new action type. The inline `@var` annotations are only at the two call sites and keep the type disclosure local, so I'd prefer to keep them as-is.

packages/administration/src/Controller/AbstractCrudController.php
henzigo: Thanks for the WIP — I understand the motivation. The configurator however provides a unified API for both scenarios — using an existing `FormType` class as well as the inline builder. The WIP variant removes the extra layer, but it exposes `FormFactory` in the extension API and loses the `setFormOption`/`getFormOption` mechanism that lets the controller and extensions share form options without directly manipulating the builder. On top of that, this layer gives us control and the ability to extend or adjust the API in the future without forcing individual CRUD controller implementations to change. For consistency, separation of concerns, and maintainability, I'd prefer to keep the configurator in this form.

henzigo: Extensions follow a consistent naming pattern mirroring the controller methods: `configureDatagrid`, `configureActions`, `configureQuery`, `configureForm`. Renaming just this one would break that consistency. The "extending" semantic is already implicit in the fact that the method lives on an extension class.

packages/administration/src/Controller/AbstractCrudController.php
vitek-rostislav: ok, I can live with that

packages/administration/src/Controller/AbstractCrudController.php
vitek-rostislav: ok ok :slightly_smiling_face:

vitek-rostislav:
From testing, I have noticed just this one thing:
- The CRUD-configured page is displayed twice in the search for some reason (the first entry is the current non-crud page, but the other two are duplicate with each other)
image
- sad

pk16011990 (approved):

vitek-rostislav (approved):

PR #4596, opened 2026-05-04
Fix annotation-fixer whitespace handling in generic types

vitek-rostislav:
There was a bug in the annotations fixer: 

When stringifying generic types like `Collection<int, CartItem>`, phpDocumentor drops the whitespace after the comma. The fixer therefore wrote `Collection<int,CartItem>` instead of the canonical `Collection<int, CartItem>`, leaving the codebase with mixed formatting. As a side-effect, the duplicate-detection step (which uses `str_contains` to compare a candidate annotation against the existing docblock) failed to recognise differently-formatted equivalents, so duplicate annotation lines could accumulate. A separate regex that extracts the annotation name from a method signature also failed on signatures with multiple type-prefix words, preventing duplicate detection in those cases too.

After the fix, the tool produces consistently formatted generic types and reliably detects existing annotations regardless of whitespace. Running the fixed tool across `project-base` cleans up the accumulated inconsistencies - adding the missing space in many `array<int, ?>` and `Collection<int, ?>` annotations and removing one duplicate `@method findMailTemplateForOrderStatus` in `OrderMail.php`.

#### Fixes issues

N/A

----
  - https://rv-annotations-fixer-fix.odin.shopsys.cloud
  - https://cz.rv-annotations-fixer-fix.odin.shopsys.cloud
  - https://rv-annotations-fixer-fix.odin.shopsys.cloud/sk

----
  - https://rv-annotations-fixer-fix.odin.shopsys.cloud
  - https://cz.rv-annotations-fixer-fix.odin.shopsys.cloud
  - https://rv-annotations-fixer-fix.odin.shopsys.cloud/sk


vitek-rostislav:
@coderabbitai review

packages/framework/src/Component/ClassExtension/AnnotationsReplacer.php
```diff
@@ -43,7 +43,7 @@ public function replaceInMethodReturnType(ReflectionMethod $reflectionMethod): s
         $replacedReturnTypes = [];
 
         foreach ($methodReturnTypes as $methodReturnType) {
-            $replacedReturnTypes[] = $this->replaceIn((string)$methodReturnType);
+            $replacedReturnTypes[] = $this->replaceIn($this->normalizeStringifiedType((string)$methodReturnType));
```
TomasLudvik: This is done in three places, it could be moved in the `replaceIn` method directly

packages/framework/tests/Unit/Component/ClassExtension/AnnotationsReplacerTest.php
```diff
@@ -106,6 +107,7 @@ public static function getTestReplaceInMethodReturnTypeDataProvider(): array
             [$reflectionClass->getMethod('returnsFrameworkArticleDataArray'), '\App\Model\Article\ArticleData[]'],
             [$reflectionClass->getMethod('returnsInt'), 'int'],
             [$reflectionClass->getMethod('returnsFrontendApiProductRepository'), '\App\FrontendApi\Model\Product\ProductRepository'],
+            [$reflectionClass->getMethod('returnsArrayOfFrameworkCategories'), 'array<int, \App\Model\Category\Category>'],
```
TomasLudvik, line 110: Parameter and properties could be also tested here.

upgrade-notes/backend_20260504_124357.md
```diff
@@ -0,0 +1,3 @@
+#### fix annotation-fixer to always add whitespace after comma in generic types ([#4596](https://github.com/shopsys/shopsys/pull/4596))
+
+- see #project-base-diff to update your project
```
TomasLudvik, line 4: There is no other change in the project-base but the applied fixer changes. I think it would be more DX friendly to just suggest running fixer or omitting upgrade note completely.

TomasLudvik (changes requested):

packages/framework/src/Component/ClassExtension/AnnotationsReplacer.php
vitek-rostislav: fixed in e27aec4d8c. I can not put it directly in `replaceIn` as the method is called from `ExtendedClassesAnnotationsCommand` directly with the whole class content, and it would then add space after any comma within the class - this is not what we want.

Yes, the whole fixer is really complicated :smile:

packages/framework/tests/Unit/Component/ClassExtension/AnnotationsReplacerTest.php
vitek-rostislav: added in 9d1b9359a3

upgrade-notes/backend_20260504_124357.md
vitek-rostislav: fair point, added in 9d1b9359a3. I would like to keep the upgrade notes and the link to project-base diff as well, IMHO it helps the project developers also for the sanity check (to easily get the answer to the question "how should the changes look like?")

TomasLudvik (approved):

pk16011990 (approved):

PR #4604, opened 2026-05-06
Remove broken FrontendApi namespace sniffs

vitek-rostislav:

Two custom PHPCS sniffs in `project-base/app/sniffer/` were removed:
- `FrontendApiNamespaceSniffer`
- `ExtendedApiClassNamespaceSniffer`

#####  They were broken and unused:

1. **Reversed `str_starts_with` arguments** in both `processResolver()` and `processMutation()` — `!str_starts_with('App\\', $actualNamespace)` always evaluated to `true` for any real `App\…` namespace, so the function exited immediately and the entire check body never ran. The sniff was effectively dead code for years.
2. **Wrong expected namespace part** — `RESOLVER_NAMESPACE_PART = 'Query'`, but the actual project convention is `…\FrontendApi\Resolver\…\FooQuery`. If bug no. 1 were fixed, every legitimate resolver in the repo would be flagged as a violation.
3. **Infinite loop in `processParentFrontendApiNamespacePart()`** — `while (\$namespacePart !== null)` paired with `next()` returning `false` (not `null`) at end-of-array. When the target part was not found in the namespace, the loop never exited and ECS hung.

##### Why the rules are not worth keeping even when fixed

- **Discovery does not depend on namespace.** Frontend API resolvers and mutations are auto-registered by Symfony via filename globs (`**/*{Query,Mutation,…}.php`) in `services.yaml` with `autoconfigure: true`. The path under `FrontendApi\…` is purely organizational, never load-bearing for runtime behavior.
- **Empirical signal.** Both sniffs were silently broken since their original implementation and nobody experienced pain from their absence — the protection they offered was not valued in practice.

#### Fixes issues

N/A

----
  - https://rv-fix-sniffer-warnings.odin.shopsys.cloud
  - https://cz.rv-fix-sniffer-warnings.odin.shopsys.cloud
  - https://rv-fix-sniffer-warnings.odin.shopsys.cloud/sk

----
  - https://rv-fix-sniffer-warnings.odin.shopsys.cloud
  - https://cz.rv-fix-sniffer-warnings.odin.shopsys.cloud
  - https://rv-fix-sniffer-warnings.odin.shopsys.cloud/sk


vitek-rostislav:
@coderabbitai review

upgrade-notes/backend_20260514_135845.md
```diff
@@ -0,0 +1,4 @@
+#### Remove ExtendedApiClassNamespaceSniffer and FrontendApiNamespaceSniffer ([#4604](https://github.com/shopsys/shopsys/pull/4604))
+
+- two custom ECS sniffers (`Sniffer\FrontendApiNamespaceSniffer`, `Sniffer\ExtendedApiClassNamespaceSniffer`) under `project-base/app/sniffer/` were removed together with the `Sniffer\` PSR-4 autoload mapping in `composer.json`
```
grossmannmartin: I'm not sure if we should describe the change in the upgrade notes or just provide instructions (in this case, just `see project-base diff`)... I will bring that up on the team meeting

grossmannmartin (approved):

PR #4610, opened 2026-05-12
Cache storefront translations across route changes

chlebektomas:

Storefront translations are now reused during client-side route changes instead of being sent again in every Next route-data response.

The first page load still contains the required translation namespaces so the page renders translated immediately. Subsequent client-side navigation now receives only translation metadata, while the browser keeps using cached translations. When translations are changed in administration, the translation version changes and the storefront refreshes the cached JSON files automatically.

Missing user translation override files are treated as an empty JSON layer, so fresh installations without administration overrides do not produce unnecessary 404 responses.

#### Fixes issues
N/A

----
  - https://tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud
  - https://cz.tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud
  - https://tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud/sk

----
  - https://tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud
  - https://cz.tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud
  - https://tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud/sk

----
  - https://tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud
  - https://cz.tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud
  - https://tc-ssp-4029-cache-storefront-translations.odin.shopsys.cloud/sk


JanMolcik (changes requested):
Requesting changes for an important locale caching regression risk.

`project-base/storefront/utils/i18n/useCachedI18n.ts:80` checks cached translations only by namespace name. `cachedNamespaces` is keyed as `common` / `accessibility`, without the locale. After an initial EN page load, if a project has a client-side language switcher and navigates to another locale with the same `__translationVersion`, the route-data response contains the new `__lang` but no `__namespaces`; `missingNamespaceNames` is empty and `shouldRefreshNamespaces` is false, so the provider can render with `lang` set to the new locale while still using the previous locale dictionaries.

This is important for e-shop implementations that use language switchers or same-host locale routes. Please key cached namespaces by locale, or reset/refetch namespaces whenever `pageProps.__lang` differs from the cached locale. Please also add a test covering EN cached namespaces -> route-data with `__lang: 'sk'`, same version, no `__namespaces`.

JanMolcik (approved):
The locale-cache issue from my previous review is fixed.

I rechecked the fixup commit and verified that cached namespaces now track the language, route-data with the same `__translationVersion` but a different `__lang` triggers a namespace refresh, and the new test covers the EN -> SK same-version case without `__namespaces` in page props.

I also verified the live preview: client-side EN -> SK navigation now fetches the SK `_next/data` payload and the SK `common`/`accessibility` locale JSON files, while the route-data response still omits `__namespaces` and keeps the expected translation metadata.

No further blocking comments from me.

project-base/docker/nginx/nginx.conf
```diff
@@ -100,10 +100,16 @@ server {
...
+    # Return an empty JSON object so server-side and browser fetches can treat them as an empty override layer.
     location ~ ^/(?:[^/]+/)?content(-test)?/locales/ {
-        try_files $uri =404;
+        default_type application/json;
+        try_files $uri @emptyJson;
+    }
+
+    location @emptyJson {
+        default_type application/json;
+        return 200 '{}';
```
grossmannmartin, line 124: This has to be done in project-base/app/orchestration/kubernetes/configmap/nginx.yaml too. 

probably with `error_page              404 = @emptyJson;` as there is a proxy_pass, not the file access

grossmannmartin (changes requested):

grossmannmartin (approved):

packages/framework/tests/Unit/Model/LanguageConstant/LanguageConstantFacadeTest.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * @param array<int, array{0: string, 1: string, 2: array<string, string>}> $userTranslationsByLocaleAndNamespace
+     * @param callable(string, string): void $writeCallback
+     */
+    private function createLanguageConstantFacade(
+        array $userTranslationsByLocaleAndNamespace,
+        callable $writeCallback,
+        ?CleanStorefrontCacheFacade $cleanStorefrontCacheFacade = null,
+    ): LanguageConstantFacade {
+        $languageConstantRepository = $this->createMock(LanguageConstantRepository::class);
```
grossmannmartin: this should be `createStub` instead of `createMock` to avoid warning during test run

grossmannmartin (approved):

pk16011990 (approved):

PR #4613, opened 2026-05-13
Update administrator activity once per request

grossmannmartin:

This PR reduces unnecessary administrator activity updates during a single administration request.

Previously, administrator activity was updated from the user refresh flow, which can be executed multiple times during one request-response lifecycle. This caused repeated database writes for the same request. Administrator inactivity validation still happens during user refresh, but the activity timestamp is now updated once for the main administration request.

As a result, administrator sessions still expire correctly after inactivity, while the current activity record is kept up to date with fewer redundant writes.

This PR also adjusts the local commit workflow instructions so that upgrade notes are kept separate by default, but can be included in the same commit when a user explicitly asks for a single commit.

#### Fixes issues

N/A

----
  - https://mg-admin-last-action.odin.shopsys.cloud
  - https://cz.mg-admin-last-action.odin.shopsys.cloud
  - https://mg-admin-last-action.odin.shopsys.cloud/sk

----
  - https://mg-admin-last-action.odin.shopsys.cloud
  - https://cz.mg-admin-last-action.odin.shopsys.cloud
  - https://mg-admin-last-action.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/tests/Unit/Model/Administrator/Activity/AdministratorActivityFacadeTest.php
+++ b/packages/framework/tests/Unit/Model/Administrator/Activity/AdministratorActivityFacadeTest.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\FrameworkBundle\Unit\Model\Administrator\Activity;
+
+use DateTimeImmutable;
+use Doctrine\ORM\EntityManagerInterface;
+use PHPUnit\Framework\TestCase;
+use Psr\Clock\ClockInterface;
+use Shopsys\FrameworkBundle\Model\Administrator\Activity\AdministratorActivity;
+use Shopsys\FrameworkBundle\Model\Administrator\Activity\AdministratorActivityFacade;
+use Shopsys\FrameworkBundle\Model\Administrator\Activity\AdministratorActivityFactory;
+use Shopsys\FrameworkBundle\Model\Administrator\Activity\AdministratorActivityRepository;
+use Shopsys\FrameworkBundle\Model\Administrator\Administrator;
+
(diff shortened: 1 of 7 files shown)


grossmannmartin:
@coderabbitai review

packages/framework/tests/Unit/Model/Administrator/Activity/AdministratorActivityFacadeTest.php
```diff
@@ -0,0 +1,50 @@
...
+use Shopsys\FrameworkBundle\Model\Administrator\Activity\AdministratorActivityFacade;
+use Shopsys\FrameworkBundle\Model\Administrator\Activity\AdministratorActivityFactory;
+use Shopsys\FrameworkBundle\Model\Administrator\Activity\AdministratorActivityRepository;
+use Shopsys\FrameworkBundle\Model\Administrator\Administrator;
+
+class AdministratorActivityFacadeTest extends TestCase
+{
+    public function testUpdateCurrentActivityUpdatesAdministratorActivityAndFlushes(): void
+    {
+        $lastActivity = new DateTimeImmutable('2026-05-13 10:00:00');
```
vitek-rostislav: I would consider naming this as something like `$mockedNow` - it helps the test clarity IMHO. But I can live with the current state

vitek-rostislav (approved):
One suggestion, otherwise ok. Reviewed and tested ✔️

pk16011990 (approved):

packages/framework/tests/Unit/Model/Administrator/Activity/AdministratorActivityFacadeTest.php
grossmannmartin: renamed

PR #4617, opened 2026-05-15
Fix language constants with boundary spaces

grossmannmartin:

Administrators can now work with language constant translations that intentionally start or end with a regular space.

Previously, boundary spaces were hard to notice in the language constant list and detail views, and translations with leading or trailing spaces could not be saved because the form trimmed the value before validation. This was problematic for translations where the surrounding space is part of the intended output.

This change keeps regular leading and trailing spaces in the submitted translation, rejects other whitespace characters such as tabs and newlines, and makes boundary spaces visible in administration using dot markers with an explanatory tooltip.

#### Fixes issues
... 

----
  - https://mg-fix-language-constant-space.odin.shopsys.cloud
  - https://cz.mg-fix-language-constant-space.odin.shopsys.cloud
  - https://mg-fix-language-constant-space.odin.shopsys.cloud/sk

----
  - https://mg-fix-language-constant-space.odin.shopsys.cloud
  - https://cz.mg-fix-language-constant-space.odin.shopsys.cloud
  - https://mg-fix-language-constant-space.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/templates/content/languageConstant/renderValueWithWhitespace.html.twig
+++ b/packages/administration/templates/content/languageConstant/renderValueWithWhitespace.html.twig
@@ -1,0 +1,19 @@
+{% macro renderValueWithWhitespace(value) %}
+    {% if value starts with ' ' or value ends with ' ' %}
+        {% set trimmedValue = value|trim(' ') %}
+
+        {%- if trimmedValue is empty -%}
+            {{- value|replace({' ': '&middot;'})|raw -}}
+        {%- else -%}
+            {%- if value starts with ' ' -%}&middot;{%- endif -%}
+            {{- trimmedValue -}}
+            {%- if value ends with ' ' -%}&middot;{%- endif -%}
+        {%- endif -%}
+
+        <span>
+            {{ info_icon('String contains leading/trailing spaces represented by &middot; sign'|trans) }}
+        </span>
+    {% else %}
+        {{ value }}
+    {% endif %}
+{% endmacro %}
--- a/packages/framework/src/Form/Admin/LanguageConstant/LanguageConstantFormType.php
+++ b/packages/framework/src/Form/Admin/LanguageConstant/LanguageConstantFormType.php
@@ -6,6 +6,7 @@ namespace Shopsys\FrameworkBundle\Form\Admin\LanguageConstant;
 
 use Override;
 use Shopsys\FormTypesBundle\ActionBarType;
+use Shopsys\FrameworkBundle\Component\Translation\Translator;
 use Shopsys\FrameworkBundle\Form\DisplayOnlyType;
 use Shopsys\FrameworkBundle\Model\LanguageConstant\LanguageConstantData;
 use Symfony\Component\Form\AbstractType;
@@ -13,6 +14,7 @@ use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Validator\Constraints;
+use Symfony\Component\Validator\Context\ExecutionContextInterface;
 
 final class LanguageConstantFormType extends AbstractType
 {
@@ -37,8 +39,15 @@ final class LanguageConstantFormType extends AbstractType
             ->add('userTranslation', TextType::class, [
                 'required' => true,
                 'label' => 'User translation',
+                'trim' => false,
                 'constraints' => [
                     new Constraints\NotBlank(),
+                    new Constraints\Callback(static function ($value, ExecutionContextInterface $context): void {
+                        if (is_string($value) && preg_match('/[^\S ]/u', $value) === 1) {
+                            $context->buildViolation(t('Only regular spaces are allowed as whitespace characters.', domain: Translator::VALIDATOR_TRANSLATION_DOMAIN))
+                                ->addViolation();
+                        }
+                    }),
                 ],
             ])
             ->add('actionBar', ActionBarType::class, [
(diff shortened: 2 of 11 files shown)


packages/framework/src/Form/Admin/LanguageConstant/LanguageConstantFormType.php
```diff
@@ -37,8 +38,15 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->add('userTranslation', TextType::class, [
                 'required' => true,
                 'label' => 'User translation',
+                'trim' => false,
                 'constraints' => [
                     new Constraints\NotBlank(),
+                    new Constraints\Callback(static function ($value, ExecutionContextInterface $context): void {
+                        if (is_string($value) && preg_match('/[^\S ]/u', $value) === 1) {
+                            $context->buildViolation('Only regular spaces are allowed as whitespace characters.')
```
vitek-rostislav: I think you will have to define the proper translation domain here, or?

packages/framework/tests/Unit/Form/Admin/LanguageConstant/LanguageConstantFormTypeTest.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        return [
+            'actionBar' => [
+                'save' => '',
+            ],
+            'userTranslation' => 'user translation',
+        ];
+    }
+
+    private function createLanguageConstantForm(): FormInterface
```
vitek-rostislav, line 129: Just an idea, if we delegated the logic into a separate tiny service instead of inlining the callback in the form type, we could get rid of that heavy testing orchestration. However, it is not a common practice to delegate the form validation logic outside the form, so 🤷

vitek-rostislav: The truth is, I have just noticed that besides the validity, we actually test the submitted data, so the `TypeTestCase` is justified here

packages/administration/templates/content/languageConstant/renderValueWithWhitespace.html.twig
```diff
@@ -0,0 +1,19 @@
...
+            {{- value|replace({' ': '&middot;'})|raw -}}
+        {%- else -%}
+            {%- if value starts with ' ' -%}&middot;{%- endif -%}
+            {{- trimmedValue -}}
+            {%- if value ends with ' ' -%}&middot;{%- endif -%}
+        {%- endif -%}
+
+        <span>
+            {{ info_icon('String contains leading/trailing spaces represented by &middot;'|trans) }}
+        </span>
```
vitek-rostislav:
```suggestion
            {{ info_icon('String contains leading/trailing spaces represented by &middot; sign'|trans) }}
```

I would put " sign" after minidot, so the sentence is clearer and does not end with the minidot character.

vitek-rostislav: (also the Czech/Slovak translations should be adjusted in this manner IMHO)

vitek-rostislav (changes requested):

packages/framework/src/Form/Admin/LanguageConstant/LanguageConstantFormType.php
vitek-rostislav: btw I also do not see the string in translations at all, so we need to use a translator here? Or improve the extraction so it automatically extracts strings from `buildViolation` into the `validators` domain :innocent:

vitek-rostislav:
one more thing (maybe just my local but worth mentioning) - I encountered an error when trying to edit
the first (`·for {{ price }}`) constant in cs/sk locale (`admin/constant/edit/accessibility/%20for%20%7B%7B%20price%20%7D%7D/`)
```
An exception occurred while executing a query: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "translation" of relation "language_constant_translations" violates not-null constraint
DETAIL:  Failing row contains (6, 1, null, sk)
```

packages/framework/src/Form/Admin/LanguageConstant/LanguageConstantFormType.php
grossmannmartin: Thanks! 
I have added the translation in the adafc1082d. I look upon the auto-extraction of buildViolation in the separate PR.

packages/framework/tests/Unit/Form/Admin/LanguageConstant/LanguageConstantFormTypeTest.php
grossmannmartin: I'm not quite sure what I should do with this 😁 Can we discuss, please?

packages/administration/templates/content/languageConstant/renderValueWithWhitespace.html.twig
grossmannmartin: Fair, what about fc521a7191?

grossmannmartin:
> one more thing (maybe just my local but worth mentioning) - I encountered an error when trying to edit the first (`·for {{ price }}`) constant in cs/sk locale (`admin/constant/edit/accessibility/%20for%20%7B%7B%20price%20%7D%7D/`)
> 
> ```
> An exception occurred while executing a query: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "translation" of relation "language_constant_translations" violates not-null constraint
> DETAIL:  Failing row contains (6, 1, null, sk)
> ```

Yeah... me too... But I wasn't able to reproduce the behavior 🤔

pk16011990 (approved):

vitek-rostislav (approved):

PR #4618, opened 2026-05-15
broaden protected-visibility rule to all Shopsys code

vitek-rostislav:

This PR strengthens the Shopsys coding standards so that private-visibility and class-finality rules are enforced uniformly across the whole `Shopsys\` namespace, instead of relying on a hand-maintained per-subnamespace allowlist that quietly let new code slip through.

`ForbiddenPrivateVisibilityFixer` now applies to all `Shopsys\` code by default; the allowlist in `ecs.php` is removed, and `utils/releaser` and `packages/cli` are explicitly excluded because they are internal tooling rather than extension points. Two refinements come with this: `final readonly class` is now correctly recognized as final (the `readonly` modifier between `final` and `class` previously broke the check), so members of those classes can stay private; and `abstract` classes are now treated as out of scope, because their authors deliberately design for inheritance and consciously choose what stays `private` vs `protected`. Several `private` members in framework commands that the old allowlist had missed — notably under `FrameworkBundle\Command` — are now `protected`, so projects can override command behavior.

Doctrine migrations are guaranteed to be final going forward. A new `FinalMigrationFixer` (sharing logic with `FinalFormTypeFixer` via the new `AbstractFinalClassByParentFixer`) adds the `final` modifier to every class extending `AbstractMigration`, the migration generator template emits `final class` directly, and all existing `Version20*` migrations are updated. With migrations now final, helpers and constants inside them return to `private`.

Also added new `FinalCompilerPassFixer` that ensure compiler pass classes are declared `final`.

Existing project overrides keep working, and from now on any new code under `Shopsys\` is automatically subject to the same visibility-and-finality rules without further ECS configuration.

#### Fixes issues
N/A

----
  - https://rv-private-protected-fixer-fix.odin.shopsys.cloud
  - https://cz.rv-private-protected-fixer-fix.odin.shopsys.cloud
  - https://rv-private-protected-fixer-fix.odin.shopsys.cloud/sk


packages/coding-standards/tests/Unit/CsFixer/ForbiddenPrivateVisibilityFixer/correct/correct-final-readonly.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace TestNamespace\TestSubNamespace;
+
+final readonly class SomeFinalReadonlyClass
```
grossmannmartin, line 5: Can you also add a test case for `readonly final class ...` as it's technically a valid order of modifiers.

packages/coding-standards/src/CsFixer/FinalFormTypeFixer.php
```diff
@@ -5,168 +5,31 @@
...
-
-        // Not found in imports - return as is (could be relative to current namespace)
-        return $className;
-    }
-
-    private function isFormTypeClass(string $className): bool
-    {
-        return in_array($className, [
+        return [
             'AbstractType',
```
grossmannmartin: I know it's not caused by this change. I just noticed it, and I can't help myself 😇

If I create class `\Shopsys\FrameworkBundle\Model\Administration\AbstractType` and extend it like this 

```
namespace Shopsys\FrameworkBundle\Model\Administration;

class FooClass extends AbstractType
{

}
```

Then this fixer forces me to make this FooClass final. 

Maybe we could use \PhpCsFixer\Tokenizer\Analyzer\FullyQualifiedNameAnalyzer to get FQCN for better comparison?

ecs.php
```diff
@@ -226,5 +201,13 @@
                 __DIR__ . '/project-base',
                 __DIR__ . '/packages/framework/src/Form/Locale/LocalizedType.php',
             ],
+            FinalMigrationFixer::class => [
+                __DIR__ . '/project-base',
```
grossmannmartin: Is there a reason for not having migrations in the project-base final? 

And the same question for Form types...

ecs.php
```diff
@@ -226,5 +201,13 @@
                 __DIR__ . '/project-base',
                 __DIR__ . '/packages/framework/src/Form/Locale/LocalizedType.php',
```
grossmannmartin: This is no longer needed imho as this class is already final

packages/migrations/src/Resources/views/Migration/migration.php.twig
```diff
@@ -8,7 +8,7 @@ use Doctrine\DBAL\Schema\Schema;
 use Override;
 use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
 
-class {{ migrationClassName }} extends AbstractMigration
+final class {{ migrationClassName }} extends AbstractMigration
```
grossmannmartin, line 11: This supports my previous point about final migrations in project-base. We generate them as final anyway.

packages/administration/src/DependencyInjection/Compiler/InitializeControllersCompilerPass.php
```diff
@@ -43,7 +43,7 @@ public function process(ContainerBuilder $container): void
      *
      * @return array{class: string, entityClass: string}|null
      */
-    private function processService(ContainerBuilder $container, string $serviceId): ?array
+    protected function processService(ContainerBuilder $container, string $serviceId): ?array
```
grossmannmartin: Does it make sense not to have CompilerPass classes final? As they are created with the new keyword in the Bundle class?

grossmannmartin (approved):

packages/coding-standards/tests/Unit/CsFixer/ForbiddenPrivateVisibilityFixer/correct/correct-final-readonly.php
vitek-rostislav: :+1: added in e69e90b551

ecs.php
vitek-rostislav: adjusted in ff8a6c55ae

ecs.php
vitek-rostislav: yes, that was already removed

packages/migrations/src/Resources/views/Migration/migration.php.twig
vitek-rostislav: yes, adjusted in ff8a6c55ae

packages/administration/src/DependencyInjection/Compiler/InitializeControllersCompilerPass.php
vitek-rostislav: true, implemented in 5651469ce0, 978b5ca505, and 279a692b2c :innocent:

packages/coding-standards/src/CsFixer/FinalFormTypeFixer.php
vitek-rostislav: yes, good point. Implemented in cd927f9211

grossmannmartin (approved):

PR #4630, opened 2026-05-21
Fix storefront domain resolution behind proxies

chlebektomas:

Storefront domain resolution is now more tolerant of requests coming through reverse proxies and scanners. Requests with public forwarded host headers are resolved against that public host instead of an internal upstream host, and default HTTP/HTTPS ports no longer make an otherwise configured domain look unknown.

When a known host is requested with an unsupported locale prefix, storefront falls back to the configured domain for that host so the request can be handled by the normal page/error flow instead of failing during server-side domain resolution.

#### Fixes issues
N/A

----
  - https://tc-ssp-3995-fix-storefront-domain-resolution.odin.shopsys.cloud
  - https://cz.tc-ssp-3995-fix-storefront-domain-resolution.odin.shopsys.cloud
  - https://tc-ssp-3995-fix-storefront-domain-resolution.odin.shopsys.cloud/sk


JanMolcik (approved):
Approved. Logic correctly addresses all three production failure modes behind SSP-3995 (proxy host SSP-4019/4020, default port `:443` SSP-3995, unsupported locale `/en` SSP-4057/4058). Full storefront vitest suite passes (1133 tests), and I verified all three Jira test scenarios locally against this branch code — no `Domain ... is not configured` crashes.

A few non-blocking warnings worth addressing:

**1. `normalizeDomainHost` — imprecise substring port replace**
`domainUrl.replace(':3000', ':8000')` matches `:3000` *inside* other ports, e.g. `host:30000` → `host:80000`. Pre-existing behavior, low real-world risk, but if you touch this line:
```ts
const normalizedDomain = domainUrl.replace(/:3000$/, ':8000');
```

**2. `findDomainConfigByHost` now runs for ALL locales**
The old fallback was guarded by `if (isDefaultLocale)`; the new host-match fallback runs unconditionally after the loop. This is intended (the SSP-4057/4058 `/en` fallback), but it's a broader behavior change than it looks — any unknown locale on a known host now falls back instead of returning an undefined config. Suggest an inline comment documenting that this fallback is intentional for unsupported-locale requests, so a future reader doesn't "fix" it back to default-only.

**3. Test gap — `X-Forwarded-Host` as a real array**
`getFirstHeaderValue` handles `string[]`, but tests only cover the comma-string form (`'secure.example.com, proxy.internal'`). Node can deliver repeated headers as an array. Suggest adding:
```ts
test('uses the first forwarded host value from an array header', () => {
    const result = getDomainConfig({
        locale: 'default',
        req: {
            headers: {
                host: 'webserver-php-fpm:8080',
                'x-forwarded-host': ['secure.example.com', 'proxy.internal'],
                'x-forwarded-proto': 'https',
            },
        },
    } as any);

    expect(result.domainId).toBe(4);
});
```

JanMolcik (approved):
Re-CR after fixup commits `74950c22d4` + `55f30032e2`. **Approved.**

### BV concerns resolved

**A — Middleware asymmetry (most important):** ✓ Fixed in `requestParser.ts:15`. Middleware now uses the same `getFirstHeaderValue(x-forwarded-host) ?? host` precedence as SSR. The two layers are now symmetric — no more silent "ultimate fallback" to `domainEntries[0]` when `Host: webserver-php-fpm:8080` arrives without forwarded headers being read.

**B — Protocol default fragility:** ✓ Fixed for the middleware path in `requestParser.ts:23-24`. When `X-Forwarded-Proto` is missing, falls back to `originalUrl.protocol` (the real connection protocol) instead of hardcoded `'http'`. Edge case from BV-B (`https example.com:443` with no XFP → wrong-protocol URL → `:443` not stripped → no match) is closed for middleware.

  Minor follow-up (non-blocking): the SSR side `getRequestProtocol` in `domainConfig.ts:21-23` still defaults to `'http'`. Different constraint there — Node's `req` doesn't carry the connection protocol the same way — so this is acceptable, but worth a brief inline comment explaining the asymmetry so future readers don't see "fix only applied to middleware" and wonder.

### Prior warnings from first CR also addressed

- ✓ `:3000 → :8000` replace anchored as `/:3000$/` regex (`domainConfig.ts:26`)
- ✓ Array-typed `X-Forwarded-Host` test added (`uses the first forwarded host value from an array header`)
- ✓ Locale-fallback intent documented inline (`domainConfig.ts:88-89`: "Unsupported locale paths should still use the configured domain for the host...")

### Verification on latest PR code

- Unit tests: **38/38 pass** across the three affected files (`resolveDomainConfigByHost` 15, `requestParser` 13, `getHostAndDomainFromRequest` 10) — including all new middleware + protocol cases.
- All 4 Jira test scenarios (comment 215637) re-run locally against `http://127.0.0.1:8000/`:
  1. targeted unit test — pass
  2. `:443` + `X-Forwarded-Proto: https` — 308, no crash
  3. proxy `Host: webserver-php-fpm:8080` + `X-Forwarded-Host: 127.0.0.1:8000` — **200 OK** (resolver now matches via XFH; previously redirected via fallback) — observable behavior change confirms BV-A fix
  4. `/en/non-existing` — 404, normal page flow
- Zero `Domain ... is not configured` / `Internal Server Error` in storefront logs throughout.

### Summary

Fixup commits cleanly address both BV findings with matching tests and updated upgrade notes that now explicitly mention middleware. Asymmetric SSR/middleware behavior eliminated. Recommend merging.

pk16011990 (approved):

PR #4632, opened 2026-05-21
Allow quantity changes from product cards

chlebektomas:

This PR lets customers manage product quantity directly from product cards and product detail once the product is already in the cart. On a fresh page load, the action area shows the same skeleton pattern used by product-list filtering while cart data is being resolved, so the UI no longer flickers between "Add to cart" and quantity controls.

The mini-cart now follows the same behavior for regular products, while gift items keep the existing gift badge styling and no longer show a remove control in the compact header cart. This makes gifts with measured quantities read as a cart benefit instead of an editable/removable item.

#### Fixes issues
N/A

----
  - https://tc-ssp-4060-adjust-cart-quantity-from-product-card.odin.shopsys.cloud
  - https://cz.tc-ssp-4060-adjust-cart-quantity-from-product-card.odin.shopsys.cloud
  - https://tc-ssp-4060-adjust-cart-quantity-from-product-card.odin.shopsys.cloud/sk

----
  - https://tc-ssp-4060-adjust-cart-quantity-from-product-card.odin.shopsys.cloud
  - https://cz.tc-ssp-4060-adjust-cart-quantity-from-product-card.odin.shopsys.cloud
  - https://tc-ssp-4060-adjust-cart-quantity-from-product-card.odin.shopsys.cloud/sk


project-base/storefront/components/Blocks/Product/CartItemQuantityControls.tsx
```diff
@@ -0,0 +1,125 @@
...
+    gtmMessageOrigin,
+    gtmProductListName,
+    listIndex,
+    size = 'medium',
+}) => {
+    const spinboxRef = useRef<HTMLInputElement>(null);
+    const lastSubmittedQuantityRef = useRef<number | null>(null);
+    const [visibleQuantity, setVisibleQuantity] = useState(cartItem.quantity);
+    const [spinboxValue, setSpinboxValue] = useState<number>();
+    const debouncedSpinboxValue = useDebounce(spinboxValue, 500);
```
JanMolcik: **Redundant debounce — removes ~500ms latency per click.**

Live timing on this branch: click → mutation = **1051ms**. `Spinbox` already debounces internally (500ms) and dedup-guards via `lastReportedValueRef`. Removing this outer debounce + threading the value directly cuts latency in half with no behavioral change.

Drop `spinboxValue`, `debouncedSpinboxValue`, and the trailing `useEffect`. Submit straight from `Spinbox.onChangeValueCallback`:

```suggestion
```
(delete this line — replaced by direct submission below)

JanMolcik (commented):
## Review notes — Spinbox debounce + dedup

I ran a live Playwright test on the deployed PR branch (rebuilt local DB on `tc-ssp-4060-adjust-cart-quantity-from-product-card`) to time the click→mutation latency. Results below; suggestions inline.

### Live measurements

```
EXP 1 — single + click:        click → AddToCartMutation = 1051ms
EXP 2 — burst of 4 clicks:     last click → AddToCartMutation = 1010ms (coalesced to 1 mutation)
```

Expected single-debounce latency would be ~500–700ms. Observed ~1000ms confirms two debounces fire serially:

- `Spinbox.tsx:43` — `useDebounce(value, 500)`
- `CartItemQuantityControls.tsx:34` — `useDebounce(spinboxValue, 500)`

### Why the outer debounce is unnecessary

Three dedup layers already exist downstream of the inner debounce:

1. `Spinbox.tsx:183-184` — `lastReportedValueRef` skips repeat reports (introduced in commit `ecc715ed58` for React 18+ Strict Mode safety)
2. `CartItemQuantityControls.tsx:51-54` — `if (quantity === cartItem.quantity) return;`
3. `CartItemQuantityControls.tsx:57-59` — `if (lastSubmittedQuantityRef.current === quantity) return;`

So even under StrictMode/rapid re-renders, duplicate `addToCart` calls are already prevented. The second `useDebounce` doesn't protect anything — it just adds ~500ms latency per click.

I searched git history, upgrade-notes, PR description/comments, and the Spinbox test suite for a documented "double-fire" / "fires twice" scenario that would justify the second debounce. Found none.

### Suggestions

See inline comments for: (1) collapsing to a single debounce in `CartItemQuantityControls`, and (2) test gaps that should be filled (no test currently asserts `toHaveBeenCalledTimes(1)` for the callback, which is why the redundant debounce wasn't caught).

### Test coverage audit

**Vitest** — only `Spinbox.test.tsx` modified. The new 125-line `CartItemQuantityControls.tsx` has **zero unit tests**. New tests added in `Spinbox.test.tsx` cover the new props (`onMinValueDecrease`, `maxValueReachedTitle`, `defaultValue` rerender) but **none assert callback fire-count** — which is why the double-debounce wasn't caught.

**Cypress** — only `cartInHeader.cy.ts` + `cartSupport.ts` + `tids.ts` modified. Changes update the mini-cart remove flow to click the spinbox decrement instead of the old remove button. **No new e2e test exists for the core PR behavior**:

- Add-to-cart from product card → spinbox transition
- `+` / `−` updates qty (and one AddToCartMutation per debounce window)
- Trash icon appears at qty 1 with "Remove from cart" tooltip
- Trash click removes item and restores "Add to cart" button
- AddToCartPopup appears only on first add (not on subsequent `+`)
- SSR skeleton renders before cart hydration (no flicker)

See the inline test-gap comment on `Spinbox.test.tsx` for suggested Vitest additions; a new `CartItemQuantityControls.test.tsx` would also be wise. For Cypress, suggest adding a flow in `cartPage.cy.ts` or a new `productCardQuantityAdjust.cy.ts` exercising the click sequence above with snapshot verification.

project-base/storefront/components/Blocks/Product/CartItemQuantityControls.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+export const CartItemQuantityControls: FC<CartItemQuantityControlsProps> = ({
+    cartItem,
+    className,
+    gtmMessageOrigin,
+    gtmProductListName,
+    listIndex,
+    size = 'medium',
+}) => {
+    const spinboxRef = useRef<HTMLInputElement>(null);
```
JanMolcik, line 28: Consider dropping the `spinboxValue` state entirely (see comment on line 34). If you keep `lastSubmittedQuantityRef` + the `cartItem.quantity` guard in `onSubmitCartChange`, the dedup remains correct without the second debounce.

project-base/storefront/components/Blocks/Product/CartItemQuantityControls.tsx
```diff
@@ -0,0 +1,125 @@
...
+        setVisibleQuantity(cartItem.quantity);
+        lastSubmittedQuantityRef.current = null;
+    }, [cartItem.quantity]);
+
+    useEffect(() => {
+        if (debouncedSpinboxValue === undefined) {
+            return;
+        }
+
+        onSubmitCartChange(debouncedSpinboxValue);
```
JanMolcik: Suggested refactor — call `onSubmitCartChange` directly from `Spinbox.onChangeValueCallback`, removing the second debounce hop:

```suggestion
                onChangeValueCallback={(quantity) => {
                    setVisibleQuantity(quantity);
                    onSubmitCartChange(quantity);
                }}
```

This removes ~500ms of latency per click (verified via Playwright). The dedup guards inside `onSubmitCartChange` (`quantity === cartItem.quantity` + `lastSubmittedQuantityRef.current === quantity`) already protect against duplicate submissions during StrictMode double-render or rapid React re-renders.

project-base/storefront/vitest/components/Spinbox/Spinbox.test.tsx
```diff
@@ -103,6 +103,30 @@ describe('Spinbox Component', () => {
             expect(input).toHaveValue(1);
         });
```
JanMolcik, line 104: **Missing test coverage: callback fire count.**

No existing test asserts `expect(onChangeCallback).toHaveBeenCalledTimes(1)` — that's why the redundant outer debounce in `CartItemQuantityControls` wasn't caught.

Suggested additions to the Spinbox test suite (or a new `CartItemQuantityControls.test.tsx`):

1. **Single click → exactly one report**
   ```ts
   test('reports value only once per debounced change', async () => {
     const cb = vi.fn();
     const { container } = render(<Spinbox {...defaultProps} defaultValue={1} onChangeValueCallback={cb} />);
     const inc = container.querySelectorAll('button')[1];
     await user.click(inc);
     // useDebounce is mocked to passthrough, so cb fires synchronously
     expect(cb).toHaveBeenCalledTimes(1);
     expect(cb).toHaveBeenCalledWith(2);
   });
   ```

2. **Same value never reported twice (dedup guard)**
   ```ts
   test('does not report when value returns to last reported value', async () => {
     const cb = vi.fn();
     const { container } = render(<Spinbox {...defaultProps} defaultValue={5} onChangeValueCallback={cb} />);
     const [dec, inc] = container.querySelectorAll('button');
     await user.click(inc); // 5→6
     await user.click(dec); // 6→5
     // Only the 6 should be reported; 5 matches the initialized lastReportedValueRef
     expect(cb.mock.calls.map(c => c[0])).toEqual([6]);
   });
   ```
   ⚠️ Note: this test exposes a **separate bug** — initializing `lastReportedValueRef` to `defaultValue` (PR change at `Spinbox.tsx:46`) silently skips reporting when the user reverts to the original quantity (e.g. on the cart page where `defaultValue={quantity}`). The base behavior (`useRef(undefined)`) would report `5`. If the new init is intentional, document why; otherwise revert.

3. **Burst clicks coalesce to one submission**
   ```ts
   test('rapid clicks fire onChangeValueCallback only once after debounce settles', async () => {
     // Requires real timers + the real useDebounce (don't mock it for this test)
     vi.useFakeTimers();
     const cb = vi.fn();
     const { container } = render(<Spinbox {...defaultProps} defaultValue={1} onChangeValueCallback={cb} />);
     const inc = container.querySelectorAll('button')[1];
     fireEvent.click(inc); fireEvent.click(inc); fireEvent.click(inc); fireEvent.click(inc);
     expect(cb).not.toHaveBeenCalled();
     vi.advanceTimersByTime(600);
     expect(cb).toHaveBeenCalledTimes(1);
     expect(cb).toHaveBeenCalledWith(5);
   });
   ```

4. **(Companion: `CartItemQuantityControls.test.tsx`) — exactly one `addToCart` call per change**
   Mock `useAddToCart` + `useRemoveFromCart` and assert `addToCart` is called once for a `+` click, once for `−`, zero times for a click that returns to `cartItem.quantity`.

project-base/storefront/components/Forms/Spinbox/Spinbox.tsx
```diff
@@ -41,7 +70,17 @@ export const Spinbox = forwardRef<HTMLInputElement, SpinboxProps>(
         const spinboxRef = useForwardedRef<HTMLInputElement>(spinboxForwardedRef);
         const intervalRef = useRef<NodeJS.Timeout | null>(null);
         const debouncedValue = useDebounce(value, 500);
-        const lastReportedValueRef = useRef<number | undefined>(undefined);
+        const lastReportedValueRef = useRef<number | undefined>(defaultValue);
```
JanMolcik, line 92: **Behavioral change for cart page — verify intent.**

This init changed from `useRef<number|undefined>(undefined)` (base) to `useRef<number|undefined>(defaultValue)` (PR). Combined with the dedup guard at line 184 (`debouncedValue !== lastReportedValueRef.current`), this means: **a user edit that returns to the original value never fires `onChangeValueCallback`**.

On the cart page (`CartListItem.tsx:159-167`) where `defaultValue={quantity}` and `onChangeValueCallback={setSpinboxValue}`, the sequence `5 → 7 → 5` will silently drop the final `5` — no `ChangeQuantityInCart` mutation, cart stays inconsistent with server.

If this is intentional (e.g. to avoid no-op mutations on the cart page), please add a comment explaining the trade-off. Otherwise revert to `useRef(undefined)` and add the test case from comment #4 to lock it in.

project-base/storefront/components/Forms/Spinbox/Spinbox.tsx
```diff
@@ -41,7 +70,17 @@ export const Spinbox = forwardRef<HTMLInputElement, SpinboxProps>(
...
+
+        useEffect(() => {
```
JanMolcik, line 95: **`defaultValue` reset effect can clobber in-flight edits.**

This effect resets `value`, `lastValidValue`, `lastReportedValueRef`, and the DOM input synchronously whenever `defaultValue` changes. In `CartItemQuantityControls` this happens on every cart refetch (via `setVisibleQuantity(cartItem.quantity)`). If the user clicks `+` while a refetch is in flight, the optimistic state is wiped.

Suggested guard: only reset when `defaultValue` differs from the last reported value, or skip the reset while `lastSubmittedQuantityRef.current !== null` upstream.

At minimum, add a test:
```ts
test('does not reset internal value when defaultValue prop matches current edit', () => {
  const { container, rerender } = render(<Spinbox {...defaultProps} defaultValue={1} />);
  const inc = container.querySelectorAll('button')[1];
  fireEvent.click(inc); // 1→2 locally
  rerender(<Spinbox {...defaultProps} defaultValue={1} />); // server still says 1
  // current code clobbers to 1; desired: preserve 2 (in-flight edit)
  const input = container.querySelector('input[type=number]') as HTMLInputElement;
  expect(input.valueAsNumber).toBe(2);
});
```

project-base/storefront/components/Blocks/Product/CartItemQuantityControls.tsx
```diff
@@ -0,0 +1,125 @@
+import { TrashCanIcon } from 'components/Basic/Icon/TrashCanIcon';
```
JanMolcik, line 1: **Missing unit tests for this component.**

125 lines of new state-management logic (visible vs spinbox vs submitted quantity, three dedup guards, error rollback, optimistic update) and zero Vitest coverage. Suggest a new `vitest/components/Blocks/Product/CartItemQuantityControls.test.tsx` covering:

- `+` click triggers exactly one `addToCart` (mocked) with absolute quantity
- `−` click at qty 1 triggers `removeFromCart`, not `addToCart`
- Click sequence that returns to `cartItem.quantity` (e.g. 5→6→5) triggers zero mutations (guard at line 51-54)
- Failed `addToCart` restores previous quantity via `restorePreviousQuantity`
- `cartItem.quantity` prop change updates `visibleQuantity` and clears `lastSubmittedQuantityRef`
- AddToCartPopup is **not** shown on `+` clicks (only the `useAddToCartHandler` path opens it, this component uses `useAddToCart` directly)

With these tests in place, removing the redundant outer debounce (see comment on line 34) becomes a safe refactor — the tests guarantee dedup still works.

project-base/storefront/cypress/e2e/cart/cartInHeader.cy.ts
```diff
@@ -26,7 +26,7 @@ describe('Cart In Header Tests', () => {
     it('[Cart Header Remove] should remove products from cart using cart in header and then display empty cart message', function () {
```
JanMolcik, line 26: **Cypress coverage gap — core PR behavior not e2e-tested.**

Only the existing mini-cart-remove flow was adapted to use the spinbox decrement. The PR's primary user-facing behavior has no e2e coverage:

1. Click "Add to cart" on a product card → spinbox `[−][1][+]` appears in that card
2. Click `+` → quantity increments, no AddToCartPopup
3. Decrease to 1 → minus icon swaps to trash icon (verify title `"Remove from cart"`)
4. Click trash → spinbox disappears, "Add to cart" button restored
5. (SSR) On fresh load with item already in cart, skeleton renders before quantity controls (no flicker between button and spinbox)

Suggested location: extend `cartPage.cy.ts` or create `cypress/e2e/cart/productCardQuantityAdjust.cy.ts`. Acceptance tests in this repo use `cy.getByTID` with the TIDs from `cypress/tids.ts` — `blocks_product_addtocart` + `forms_spinbox_increase` / `_decrease` / `spinbox_input` are already available.

Without this coverage, future refactors (e.g. removing the redundant outer debounce or changing the SSR skeleton timing) could silently regress the feature.

project-base/storefront/components/Blocks/Product/CartItemQuantityControls.tsx
chlebektomas: Done, I removed the extra debounce.

project-base/storefront/vitest/components/Spinbox/Spinbox.test.tsx
chlebektomas: Added tests for rapid spinbox clicks.

project-base/storefront/components/Forms/Spinbox/Spinbox.tsx
chlebektomas: Fixed for product card controls, stale cart data should not overwrite the pending quantity there.

project-base/storefront/components/Blocks/Product/CartItemQuantityControls.tsx
chlebektomas: Added unit tests.

project-base/storefront/cypress/e2e/cart/cartInHeader.cy.ts
chlebektomas: Added Cypress test for add-to-cart button changing to spinbox and back. SSR skeleton tests are flaky, I would not add them here.

project-base/storefront/components/Forms/Spinbox/Spinbox.tsx
chlebektomas: Fixed, changing back to the original value is still sent when needed.

JanMolcik (approved):

upgrade-notes/storefront_20260521_143352.md
```diff
@@ -0,0 +1,7 @@
+#### Allow quantity changes from product cards ([#4632](https://github.com/shopsys/shopsys/pull/4632))
```
fujta, line 1: Nemělo by tu přibýt info i o tom `knip`?

fujta (approved):

upgrade-notes/storefront_20260521_143352.md
chlebektomas: Je to přidáno jako extra commit. Není to součástí úpravy. Za mě není třeba upravovat upgrade notes.

JanMolcik:
## Review finding after retest

- first add from a product card changes the card to a spinbox
- qty `1` shows the trash icon
- the first AddToCart popup appears
- clicking `+` from the spinbox does not reopen the AddToCart popup
- the pending state no longer renders a spinner; the live DOM showed no spinner and only the quieter pending look

### Blocking issue: overlapping quantity changes can be lost

There is still a slow-network bug in the quantity controls.

Live repro with a delayed `AddToCartMutation`:

1. Add product from a card, quantity is `1`.
2. Delay the next `AddToCartMutation` by ~3 seconds.
3. Click `+`, so the visible quantity becomes `2` and the absolute quantity `2` mutation is in flight.
4. Before it finishes, click `-`, so the user intent is back to `1`.
5. Wait for the request/refetch.

Actual result: only one mutation is sent and the final cart becomes quantity `2` / `€279.92`. The user's final intent (`1`) is lost.

The static code path matches the runtime behavior:

- `project-base/storefront/utils/cart/useAddToCart.ts` returns `null` when the same product UUID is already in `addingToCartProductUuidsRef`.
- `project-base/storefront/components/Blocks/Product/CartItemQuantityControls.tsx` treats the `null` result as failure/stale restore, but the earlier in-flight absolute mutation still wins when it resolves.

Suggested fix: queue/serialize the latest absolute quantity per product while a mutation is in flight, so the last user intent wins. Alternatively, disable the controls until the pending mutation/refetch completes, but queueing is nicer UX.

Please also add coverage with a deferred `addToCart` promise or Cypress network delay: click `+`, then click `-` before the first mutation resolves, and assert the final cart quantity matches the last user intent.

project-base/storefront/components/Blocks/Product/ProductsList/ProductListItem.tsx
```diff
@@ -72,22 +74,31 @@ export const ProductListItem = forwardRef<HTMLLIElement, ProductItemProps>(
...
         const { t } = useTranslation();
-        const { canSeePrices } = useAuthorization();
+        const { canCreateOrder, canSeePrices } = useAuthorization();
+        const { cart, isCartFetchingOrUnavailable } = useCurrentCart();
+        const isProductActionDependentOnCart =
+            visibleItemsConfig.addToCart &&
+            canCreateOrder &&
+            !product.isSellingDenied &&
+            !product.isCurrentlyOutOfStock &&
+            !(!product.isMainVariant && product.isInquiryType) &&
```
fujta: Použil bych DeMorgana a přepsal na (product.isMainVariant || !product.isInquiryType))

Dle mě bude čitelnější :D

project-base/storefront/components/Pages/Cart/CartList/CartListItem.tsx
```diff
@@ -34,39 +34,52 @@ export const CartListItem: FC<CartListItemProps> = ({
...
-        if (lastSubmittedQuantityRef.current === qty) {
-            return;
-        }
-        lastSubmittedQuantityRef.current = qty;
-        onAddToCart(productUuid, qty, idx, true);
-    });
+    const onSubmitCartChange = useCallback(
+        async (qty: number) => {
+            const currentSubmittedQuantity = lastSubmittedQuantityRef.current;
+            const quantityToCompare = currentSubmittedQuantity ?? quantity;
```
fujta: Vím, že je to mimo scope, ale na první pohled mám celkem guláš v tom pojmenování množství:

- qty
- quantity
- currentSubmittedQuantity
- lastSubmittedQuantityRef
- quantityToCompare

Minimálně bych qty a quantity pojmenoval nějak trefněji. Ale nechám na tvoje uvážení.

JanMolcik (changes requested):
## Blockers

- I cannot approve the current head `0c74ea066800e00e3b3a5dd1f3443537efe6766e`: the PR is still blocked by failing required checks.
- `Check Storefront standards` fails because `docker compose exec storefront pnpm compile-tailwind-for-admin` changes `project-base/storefront/public/tailwind-for-admin/style.css`. Please regenerate and commit the generated admin Tailwind CSS.
- `Run Cypress tests (cart)` still fails, including a PR-owned missing baseline for `snapshots/e2e/cart/productAddToCart.cy.ts/2-3-6 product card quantity spinbox.png`, plus visual diffs around the add-to-cart popup/header cart and a clipped gift-row visibility assertion.
- Additional Cypress visual suites are still red (`comparison`, `filterAndSort`, `limitedUser`, plus the remaining failed suites in the check list), so the current branch is not merge-ready.

## Critical findings

- None found in the updated quantity-queueing code. The previous slow/overlapping quantity-change issue appears addressed by the new serialized submission flow in `CartItemQuantityControls` and the added Vitest coverage.

## Warnings

- Several visual snapshots are now different from the baseline. Since this PR changes product-card cart controls globally, please review and commit the intended snapshot/generated-output changes instead of relying on CI reruns.

## Suggestions

- Consider adding/keeping a focused regression for the `2 -> 1 -> trash` path while an absolute quantity update is still pending. The live preview behaved correctly for me, but it is close to the tester-reported rapid-decrement class of issues.

## Verified correct

- Jira tester feedback is reflected in the current implementation: the earlier max tooltip, rapid decrement, mobile drawer, trash-at-one, and spinner/pending-look rounds were all adjusted and marked OK/accepted in Jira comments.
- Static review: `CartItemQuantityControls` now queues the latest requested absolute quantity while a same-product add-to-cart mutation is in flight, then submits that latest value after the current request completes.
- Static review: the added Vitest test covers `+` followed by `-` before the first cart update resolves.
- Live preview: first add switches the product card into quantity controls; a rapid `+` then `-` sequence ended at visible quantity `1`; a `2 -> 1 -> trash` sequence removed the spinbox and restored the add button.

## Playwright verification

- Review URL tested: `https://tc-ssp-4060-adjust-cart-quantity-from-product-card.odin.shopsys.cloud`
- Scenarios checked: first add/product-card spinbox, rapid increment/decrement from the product card, and decrement-to-trash removal.
- Result: the live feature behavior looked correct in these flows, but CI/generated artifacts still block approval.

## Summary

The previous functional blocker from my last review appears fixed. I am requesting changes rather than approving because the current head still has red required checks, including an uncommitted generated CSS change and a missing Cypress baseline introduced by this PR.

JanMolcik (approved):
Approving from code-review perspective. The previously reported overlapping quantity-change issue is addressed in the current head, the reviewed code looks clean, and the remaining failing checks/generated snapshots were explicitly accepted as non-blocking for this review.

pk16011990 (approved):

PR #4637, opened 2026-05-29
Improve administration order detail editing

grossmannmartin:

This PR improves the administration order detail so administrators can work with orders in smaller, focused parts instead of one long edit form. Order data is split into editable sections and lazy-loaded tabs, making the page easier to scan and reducing the risk of changing unrelated data while handling a specific task.

The order detail now supports contextual editing of basic order data, addresses, transport and payment, notes, withdrawal state, and order items. It also adds dedicated areas for order history and payment transactions, including clearer GoPay refund handling and manual refunded amount correction directly from the order detail.

The change also introduces extension points and documentation for adding custom order detail sections and tabs, so projects can extend the page without copying the whole order detail template.

#### Fixes issues

closes #4546 

----
  - https://mg-order-detail.odin.shopsys.cloud
  - https://cz.mg-order-detail.odin.shopsys.cloud
  - https://mg-order-detail.odin.shopsys.cloud/sk

----
  - https://mg-order-detail.odin.shopsys.cloud
  - https://cz.mg-order-detail.odin.shopsys.cloud
  - https://mg-order-detail.odin.shopsys.cloud/sk


project-base/app/assets/controllers.json
```diff
@@ -0,0 +1,14 @@
+{
```
vitek-rostislav, line 1: I am not happy that we have to add all these bootstraping files (controllers.json, stimulus_bootstrap.js, ...) into `project-base`, not into the `administration` package. But I suppose we have no other option?

packages/administration/src/Component/Security/AccessControl/LiveComponentAccessControlDataProvider.php
```diff
@@ -0,0 +1,95 @@
...
+
+class LiveComponentAccessControlDataProvider
+{
+    /**
+     * @param iterable<object> $twigComponents
+     */
+    public function __construct(
+        protected readonly AttributeProcessor $attributeProcessor,
+        #[AutowireIterator('twig.component')]
+        protected readonly iterable $twigComponents,
```
vitek-rostislav, line 23: I wonder, if we used an existing vendor component (e.g. from https://ux.symfony.com/packages), would it be then required to define the RBAC attribute there? Would we be able to achieve that somehow?

packages/administration/src/Component/FlashMessage/LiveComponentFlashMessageSubscriber.php
```diff
@@ -0,0 +1,86 @@
...
+        if ($toastHtml === '') {
+            return;
+        }
+
+        $event->getResponse()->headers->set(self::FLASH_MESSAGES_HEADER, base64_encode($toastHtml));
+    }
+
+    protected function shouldHandleResponse(Response $response): bool
+    {
+        if (!$response->isSuccessful()) {
```
vitek-rostislav: Claude says: (:smile:)
> 422 responses are skipped. The guard if (!$response->isSuccessful()) returns early. Symfony UX LiveComponent returns
    HTTP 422 for validation-error re-renders, with the content type still application/vnd.live-component+html. Any flash message added during such a request won't be emitted in the header; it
    stays in the flash bag and then surfaces unexpectedly on the next full page load. If flash messages can be added on a validation-failure path, handle 422 explicitly (or confirm this is
    intentional and document it).

packages/administration/templates/partial/flash_message/all_messages.html.twig
```diff
@@ -0,0 +1,15 @@
+<div {{ attributes }}>
```
vitek-rostislav, line 1: IMHO, the `live` name makes you think about things you should not think about :smile: I would be totally ok with `index.html.twig` here or something like `all_messages.html.twig`

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
```diff
@@ -0,0 +1,86 @@
...
+use Symfony\UX\LiveComponent\ComponentToolsTrait;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+#[AsLiveComponent(
+    name: 'OrderDetail:SectionEditor',
+    template: '@ShopsysAdministration/content/order/detail/components/section_editor.html.twig',
+)]
+#[ForRole(AdminRoleConstant::ROLE_ORDER)]
+class SectionEditorComponent
+{
```
vitek-rostislav, line 26: I would like to talk about reusability of the components - e.g. this one does not need to belong to the order namespace at all, or? I can imagine addressing the generalization once it is needed (when redesigning a product edit page?) or in a separate task (preferably), but I could imagine that as a part of this task as well :)

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
```diff
@@ -0,0 +1,86 @@
...
+
+    #[LiveListener(self::OPEN_ORDER_DETAIL_SECTION_EDITOR)]
+    #[CanEdit]
+    public function openSectionEditor(#[LiveArg] string $section): void
+    {
+        $this->assertValidSectionId($section);
+
+        $this->sectionId = $section;
+        $this->open = true;
+        $this->dispatchBrowserEvent('order-detail-section-editor:open');
```
vitek-rostislav: I wanted to ask to use constants for those, but they are used in twig only, right? Even with that said, it could help with the readability of the code. But I am not sure, I do not like the constant usages in templates, so :D

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
```diff
@@ -0,0 +1,86 @@
...
+    template: '@ShopsysAdministration/content/order/detail/components/section_editor.html.twig',
+)]
+#[ForRole(AdminRoleConstant::ROLE_ORDER)]
+class SectionEditorComponent
+{
+    use ComponentToolsTrait;
+    use DefaultActionTrait;
+
+    public const string CLOSE_ORDER_DETAIL_SECTION_EDITOR = 'closeOrderDetailSectionEditor';
+    public const string OPEN_ORDER_DETAIL_SECTION_EDITOR = 'openOrderDetailSectionEditor';
```
vitek-rostislav: maybe we could use `_EVENT` and `Event` suffixes so the constants are clearer?

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
```diff
@@ -0,0 +1,86 @@
...
+        if ($this->sectionId === null) {
+            return null;
+        }
+
+        return $this->orderDetailSectionRegistry->getSection($this->sectionId);
+    }
+
+    protected function assertValidSectionId(string $sectionId): void
+    {
+        if (ctype_alnum($sectionId)) {
```
vitek-rostislav: is this necessary? Why do we need to restrict the section identifier this way?

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorFormComponent.php
```diff
@@ -0,0 +1,126 @@
...
+            [
+                'order' => $order,
+                'attr' => [
+                    'novalidate' => 'novalidate',
+                ],
+            ],
+        );
+    }
+
+    protected function assertValidSectionId(string $sectionId): void
```
vitek-rostislav: this is duplicated with the method in `SectionEditorComponent`

packages/administration/src/Component/OrderDetail/OrderDetailSectionRegistry.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function getSection(string $sectionId): OrderDetailSection
+    {
+        return $this->getSections()[$sectionId] ?? throw new NotFoundHttpException();
+    }
+
+    /**
+     * @return array<string, \Shopsys\AdministrationBundle\Component\OrderDetail\OrderDetailSection>
+     */
+    public function getSections(): array
```
vitek-rostislav: this could be protected

docs/cookbook/extending-order-detail-page.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+and add the new field where it should appear:
+
+```twig
+{{ form_row(form.personalDataGroup.firstName) }}
+{{ form_row(form.personalDataGroup.lastName) }}
+{{ form_row(form.personalDataGroup.email) }}
+{{ form_row(form.personalDataGroup.telephone) }}
+{{ form_row(form.personalDataGroup.deliveryInstructions) }}
```
vitek-rostislav: just a note, here is a room for improvement imho. For regular forms, we achieved that neat thing that developers do not need to adjust the templates when adding new fields. Now, they have to update two templates in such a scenario. Could not we somehow generalize the section fields rendering so that we do not have to explicitly enumerate them in the templates? :eyes:

packages/administration/templates/content/order/detail/sections/personal_form.html.twig
```diff
@@ -0,0 +1,4 @@
+{{ form_row(form.personalDataGroup.firstName) }}
```
vitek-rostislav: as mentioned in the comment in the docs - could not we generalize/get rid of these templates? Lets talk about it. 

Btw, we use camelCase for the form templates, and snake_case for the view templates, it worsens their coupling imho

docs/cookbook/extending-order-detail-page.md
```diff
@@ -0,0 +1,302 @@
...
+For example, override the existing personal section form template:
+
+```text
+app/templates/bundles/ShopsysAdministrationBundle/content/order/detail/sections/personalForm.html.twig
+```
+
+and add the new field where it should appear:
+
+```twig
+{{ form_row(form.personalDataGroup.firstName) }}
```
vitek-rostislav: another thought - would it make sense to prepare a block here, so that the extending template uses renders the parent block and then adds its own fields (sure, that would help only if the new field had to be rendered first/last, but still...)

docs/cookbook/extending-order-detail-page.md
```diff
@@ -0,0 +1,302 @@
...
+
+```twig
+{# app/templates/Admin/Content/Order/Detail/Sections/internalNoteForm.html.twig #}
+
+{{ form_row(form.noteGroup.internalNote) }}
+```
+
+The form field itself still has to be added to `OrderFormType` through a form extension, as shown in the previous section.
+
+Registering a section does not render it anywhere automatically.
```
vitek-rostislav, line 195: could it? Should it? :smile:

vitek-rostislav:
Random note from testing - seems like the popover is not displayed properly in the modal, see.

image


"This phone number may not be valid. Please verify it." should be displayed there, but it is not. It happens when the warning is added dynamically, my steps to repreduce:
1) edit delivery addres, set same as billing, save
2) open the delivery address form again, uncheck "same as billing", select telephone prefix
3) see dynamically added warning icon in the phone number input, without popover

packages/administration/templates/content/order/detail/components/section_editor.html.twig
```diff
@@ -0,0 +1,54 @@
...
+         data-form-modal-pending-actions-value="{{ ['save']|json_encode }}">
+        <div class="modal-dialog {{ this.section ? this.section.getModalDialogClass() : 'modal-xl' }} modal-dialog-centered modal-dialog-scrollable">
+            <div class="modal-content">
+                {% if this.section %}
+                    <div class="modal-header">
+                        <h5 class="modal-title">{{ this.section.getModalTitle() }}</h5>
+                        <button type="button"
+                                class="btn-close"
+                                data-action="form-modal#requestClose"
+                                aria-label="{{ 'Close'|trans }}"></button>
```
vitek-rostislav: there could be a line break before the closing tag

vitek-rostislav:
Another note from testing: 

I noticed the submit button is always active, and it is possible to "submit" the form even if there are validation errors. Could we improve this UX?


https://github.com/user-attachments/assets/db3b0293-52d1-4d67-a707-2674211cb25d

packages/administration/templates/content/order/detail/components/section_editor.html.twig
```diff
@@ -0,0 +1,44 @@
...
+                                class="btn btn-ghost-secondary"
+                                data-action="form-modal#requestClose">
+                            {{ 'Cancel'|trans }}
+                        </button>
+                        <button type="button"
+                                class="d-none"
+                                data-form-modal-target="closeTrigger"
+                                data-action="live#action"
+                                data-live-action-param="close"></button>
+                        <button type="submit" class="btn btn-primary" form="js-order-detail-section-editor-form">
```
vitek-rostislav: here is also missing a linebreak before closing tag

packages/administration/assets/src/controllers/delivery_address_form_controller.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export default class extends Controller {
+    static targets = ['sameAsBillingAddress', 'deliveryFields'];
+
+    connect() {
+        this.updateDeliveryFieldsState();
+    }
+
+    updateDeliveryFieldsState() {
+        const disabled = this.sameAsBillingAddressTarget.checked;
```
vitek-rostislav, line 15: when I check the "same as billing" checkbox, the delivery fields disappear from the form, leaving just the checkbox visible. Is that a wanted behavior? From this piece of code I would expect them to be just disabled but still visible :thinking:

packages/administration/src/Component/OrderDetail/OrderDetailSectionProvider.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * @return iterable<\Shopsys\AdministrationBundle\Component\OrderDetail\OrderDetailSection>
+     */
+    #[Override]
+    public function getSections(): iterable
+    {
+        yield new OrderDetailSection(
+            'tracking',
+            '@ShopsysAdministration/content/order/detail/sections/_payment_transport_view.html.twig',
+            '@ShopsysAdministration/content/order/detail/sections/paymentTransportForm.html.twig',
```
vitek-rostislav: I think I have mentioned that before already - the inconsistency in naming for read/edit templates seems quite confusing to me. I would rather have something like `paymentTransportView` and `paymentTransportForm` pairs

packages/administration/src/Controller/OrderDetailController.php
```diff
@@ -0,0 +1,68 @@
...
+        protected readonly OrderFacade $orderFacade,
+        protected readonly BreadcrumbOverrider $breadcrumbOverrider,
+        protected readonly OrderDataFactory $orderDataFactory,
+        protected readonly OrderStatusFacade $orderStatusFacade,
+    ) {
+    }
+
+    #[Route(path: '/order/edit/{id}', requirements: ['id' => '\d+'], name: 'admin_order_edit')]
+    #[CanEdit(methods: [HttpMethod::POST])]
+    #[CanView(methods: [HttpMethod::GET])]
```
vitek-rostislav, line 38: will `POST` be even allowed in this scenario? There is no form to submit anymore, right?

packages/administration/src/Controller/OrderDetailController.php
```diff
@@ -0,0 +1,68 @@
...
+    #[CanEdit]
+    #[CsrfProtection]
+    public function changeStatusAction(int $id, int $statusId): Response
+    {
+        $status = $this->orderStatusFacade->getById($statusId);
+
+        $orderData = $this->orderDataFactory->createFromOrder($this->orderFacade->getById($id));
+        $orderData->status = $status;
+        $this->orderFacade->edit($id, $orderData);
+        $this->addSuccessFlash(t('Order status has been changed.'));
```
vitek-rostislav, line 153: Just an idea, would it make sense to create (extract from edit) a thinner public method just for changing the status? The edit method itself is quite heavy. On the other hand, part of that heaviness is because of the status changes, so...

packages/administration/templates/content/order/detail/sections/_basic_information_view.html.twig
```diff
@@ -0,0 +1,78 @@
...
+
+<div class="subheader d-flex justify-content-between align-items-center mb-1">
+    <span>{{ 'Basic information'|trans }}</span>
+    <a href="#"
+       class="text-secondary small d-inline-block"
+       data-bs-toggle="collapse"
+       data-bs-target="#basicInfoMore"
+       data-text-more="{{ 'Show more'|trans }}"
+       data-text-less="{{ 'Show less'|trans }}"
+       onclick="var d=this.dataset; this.textContent = this.textContent.trim() === d.textMore ? d.textLess : d.textMore;"
```
vitek-rostislav: I know it is just a tiny piece of code, but do we really want to inline JS this way? I would say it is not a common practice. Moreover, it is not re-usable, so IMHO we should either create a small JS toggle component, or, get rid of the SJ completely and implement this using css - see https://github.com/shopsys/shopsys/commit/06130379025d171b494c13a2a478ac93c1587da1 - what do you think?

vitek-rostislav:
Another note from testing - when changing the status to withdrawn, I see a modal where I need to set the withdrawal data. This makes sense but I am afraid it might be confusing  - when admin closes the modal, the status is not changed. Maybe we could add a warning or something like thta, lets talk about it, please

packages/administration/templates/content/order/detail/sections/_basic_information_view.html.twig
```diff
@@ -0,0 +1,78 @@
+{% set canEdit = canEdit|default(can_edit(constant('Shopsys\\FrameworkBundle\\Component\\Security\\Role\\AdminRoleConstant::ROLE_ORDER'))) %}
+{% set order_status_color = order.status.type|order_status_color %}
+
```
vitek-rostislav: I would use camelCase for the variable name so it is unified with variable above and distinguished from the macro itself

packages/administration/templates/content/order/detail/sections/billing_address_view.html.twig
```diff
@@ -0,0 +1,54 @@
+<div class="subheader d-flex justify-content-between align-items-center mb-1">
+    {{ 'Billing address'|trans }}
+    {% include '@ShopsysAdministration/content/order/detail/partials/_button_edit.html.twig' with {
```
vitek-rostislav: maybe it is an overkill, but maybe, it would make sense to introduce a component for the edit button? Lets discuss it - https://github.com/shopsys/shopsys/commit/71a706d911ee49d744894fb665ee90879b4550fc

packages/administration/src/Component/OrderDetail/OrderDetailTabProvider.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @return iterable<\Shopsys\AdministrationBundle\Component\OrderDetail\OrderDetailTab>
+     */
+    #[Override]
+    public function getTabs(): iterable
+    {
+        yield new OrderDetailTab(
+            'withdrawal',
+            t('Withdrawal'),
+            'OrderDetail:WithdrawalTab',
```
vitek-rostislav: I would appreciate constants for these

packages/administration/src/Controller/OrderDetailController.php
```diff
@@ -0,0 +1,156 @@
...
+        $this->breadcrumbOverrider->overrideLastItem(
+            t('Editing order - Nr. %number%', ['%number%' => $order->getNumber()]),
+        );
+
+        $orderDetailTabs = $this->orderDetailTabRegistry->getTabs($order);
+
+        return $this->render('@ShopsysAdministration/content/order/detail/edit.html.twig', [
+            'orderDetailTabs' => $orderDetailTabs,
+            'order' => $order,
+            'activeTab' => $this->getActiveTabId($orderDetailTabs, $request->query->getString('activeTab'), $order),
```
vitek-rostislav, line 52: application never sets the `activeTab` to the query string? We should implement that, or?

packages/administration/assets/src/controllers/order_detail_tabs_controller.js
```diff
@@ -0,0 +1,59 @@
+import { Controller } from '@hotwired/stimulus';
+import Translator from 'bazinga-translator';
+import Register from 'framework/common/utils/Register';
+
+export default class extends Controller {
```
vitek-rostislav, line 5: the tabs component is one of the things that could definitelly be generalized

packages/administration/templates/content/order/detail/edit.html.twig
```diff
@@ -0,0 +1,85 @@
...
+
+    {% if orderDetailTabs %}
+        <div class="card" data-controller="order-detail-tabs">
+            <div class="card-header">
+                <ul class="nav nav-tabs card-header-tabs" role="tablist">
+                    {% for tab in orderDetailTabs %}
+                        {% set isActive = activeTab == tab.id %}
+                        {% set isDisabled = tab.isDisabled(order) %}
+                        <li class="nav-item" role="presentation">
+                            <button class="nav-link js-order-detail-tab-link{% if isActive %} active{% endif %}"
```
vitek-rostislav: it seems like `js-order-detail-tab-link` is never used?

packages/administration/src/Component/OrderDetail/LiveComponent/WithdrawalStatusModalComponent.php
```diff
@@ -0,0 +1,133 @@
...
+                    OrderWithdrawalFormType::VALIDATION_GROUP_WITHDRAWAL_REQUIRED,
+                ],
+                'attr' => [
+                    'novalidate' => 'novalidate',
+                ],
+            ],
+        );
+    }
+
+    protected function createWithdrawalRequestData(
```
vitek-rostislav: I can imagine putting this method directly into the factory, something like `createFromWithdrawalRequestOrFallbackToOrder` (probably some better name :D )

packages/administration/src/Controller/OrderDetailController.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        $order = $this->orderFacade->getById($id);
+        $status = $this->orderStatusFacade->getById($statusId);
+
+        if ($status->getType() !== OrderStatusTypeEnum::TYPE_WITHDRAWN) {
+            throw $this->createNotFoundException();
+        }
+
+        $this->breadcrumbOverrider->overrideLastItem(
+            t('Editing order - Nr. %number%', ['%number%' => $order->getNumber()]),
```
vitek-rostislav, line 73: hmm it looks like the breadcrumb does not work at all for this route (you probably need to add it to the menu configuration as hidden item?)

image

packages/administration/src/Controller/OrderDetailController.php
```diff
@@ -0,0 +1,156 @@
...
+    #[CanEdit]
+    #[CsrfProtection]
+    public function changeStatusAction(int $id, int $statusId): Response
+    {
+        $order = $this->orderFacade->getById($id);
+        $status = $this->orderStatusFacade->getById($statusId);
+
+        if ($status->getType() === OrderStatusTypeEnum::TYPE_WITHDRAWN) {
+            return new RedirectResponse($this->generateUrl('admin_order_edit_withdrawal_status', [
+                'id' => $id,
```
vitek-rostislav: you can use `redirectToRoute` here and also a few lines below

packages/administration/tests/Unit/Component/EntityLog/Timeline/TwigComponent/EntityLogTimelineComponentTest.php
```diff
@@ -0,0 +1,228 @@
...
+        $component = $this->createComponent([$firstLog, $secondLog]);
+
+        $groups = $component->getGroups();
+
+        $this->assertCount(2, $groups);
+        $this->assertSame($firstLog, $groups[0][0]);
+        $this->assertSame($secondLog, $groups[1][0]);
+    }
+
+    public function testActionDisplayHelpersUseActionSpecificValues(): void
```
vitek-rostislav: the assertions in this test are quite random, right? They cover just one case in each of the three match constructs behind, I think I do not like and I think I want to remove the whole test

packages/administration/tests/Unit/Component/EntityLog/Timeline/TwigComponent/EntityLogTimelineComponentTest.php
```diff
@@ -0,0 +1,228 @@
...
+        $resolvedChangesFormatterStub
+            ->method('formatResolvedChanges')
+            ->willReturnCallback(static function (array $changeSet): string {
+                if ($changeSet === []) {
+                    return '';
+                }
+
+                $firstChange = reset($changeSet);
+
+                return 'formatted:' . $firstChange['newReadableValue'];
```
vitek-rostislav: extracting the "formatted:" string into a constant could be a nice little improvement towards the test readability

vitek-rostislav:
Yet another thing I noticed - the history tab never gets updated when performing partial edits in the modals. Admin has to reload the whole page to see the changes in the tab. Do we want to improve that? :innocent: 


https://github.com/user-attachments/assets/549c8d78-74b3-4448-bc35-2faad841f021

docs/cookbook/extending-order-detail-page.md
```diff
@@ -0,0 +1,302 @@
+# Extending Order Detail Page
+
+This cookbook describes how to extend the administration order detail page with new features.
+The order detail page is split into editable sections and lazy-loaded tabs:
+
+- sections are small editable parts of the order detail and are rendered with the `OrderDetail:Section` component
+- tabs are larger page parts rendered through a registered component and loaded only when opened
+
```
vitek-rostislav, line 12: Maybe a screenshot with an explanation what is what could be helpful here to make things absolutely clear

docs/cookbook/extending-order-detail-page.md
```diff
@@ -0,0 +1,302 @@
+# Extending Order Detail Page
+
```
vitek-rostislav, line 2: the page deserves a TOC

packages/administration/templates/form/vertical.html.twig
```diff
@@ -0,0 +1,68 @@
+{% extends 'bootstrap_5_layout.html.twig' %}
+
+{%- block textarea_widget -%}
+    {% set attr = attr|merge({'data-bs-toggle': 'autosize'}) %}
+    {{- parent() -}}
+{%- endblock -%}
+
+{%- block form_widget_simple -%}
+    {% if symbolAfterInput is defined or symbolBeforeInput is defined or rawSymbolAfterInput is defined or rawSymbolBeforeInput is defined %}
+        <div class="input-group">
```
vitek-rostislav, line 10: the vertical theme has a lot of code in common with the default `theme.html.twig`. Could not we somehow avoid the DRYness?

packages/administration/templates/form/vertical.html.twig
```diff
@@ -0,0 +1,68 @@
+{% extends 'bootstrap_5_layout.html.twig' %}
```
vitek-rostislav, line 1: is it intentional that this theme is located in a separate folder (`form_theme`) and not in the same (`form`) folder as `theme.html.twig`?

packages/administration/src/Component/OrderDetail/LiveComponent/PaymentTransactionsTabComponent.php
```diff
@@ -0,0 +1,259 @@
...
+            throw new InvalidArgumentException('Payment transaction refund modal is not open.');
+        }
+
+        return $this->getPaymentTransaction($this->refundModalPaymentTransactionId);
+    }
+
+    protected function instantiateForm(): FormInterface
+    {
+        if ($this->refundModalPaymentTransactionId === null) {
+            return $this->formFactory->createNamed('payment_transaction_refund', FormType::class);
```
vitek-rostislav: the second parameter (`FormType::class`) could be omitted, it matches the default value

packages/administration/templates/content/order/detail/components/payment_transactions_tab.html.twig
```diff
@@ -0,0 +1,220 @@
...
+                                },
+                            }) }}
+                            {% if this.refundModalErrorMessage %}
+                                <div class="alert alert-danger mb-4" role="alert">
+                                    {{ this.refundModalErrorMessage }}
+                                </div>
+                            {% endif %}
+
+                            <div class="border-top pt-4">
+                                <h4>{{ 'Refund via GoPay'|trans }}</h4>
```
vitek-rostislav, line 157: goPay vs other provider transactions is a separate topic, right? But still, do we want to hardcode gopay here ion the label?

packages/administration/templates/form/content/PaymentTransactionsType.html.twig
```diff
@@ -1,81 +0,0 @@
...
-                <br>
-                ({{ paymentTransaction.externalPaymentSubStatus|translate_payment_transaction_sub_status }})
-            {% endif %}
-        </td>
-        <td class="text-end">
-            {{ paymentTransaction.paidAmount|priceWithCurrencyByOrder(order) }}
-        </td>
-        <td class="text-end">
-            <span class="js-refunded-amount">
-            {{ paymentTransaction.refundedAmount|priceWithCurrencyByOrder(order) }}
```
vitek-rostislav, line 47: `ExecuteRefund.js` can be now removed as it is unused, right?

packages/administration/src/Component/OrderDetail/LiveComponent/PaymentTransactionsTabComponent.php
```diff
@@ -0,0 +1,259 @@
...
+        InvalidArgumentException $exception,
+        PaymentTransaction $paymentTransaction,
+    ): string {
+        if ($exception->getMessage() === PaymentTransactionRefundFacade::ERROR_REFUNDED_AMOUNT_GREATER_THAN_PAID_AMOUNT) {
+            return t('You can set refunded amount only up to %paidAmount%.', [
+                '%paidAmount%' => $this->priceExtension->priceWithCurrencyByOrderFilter($paymentTransaction->getPaidAmount(), $this->getOrder()),
+            ]);
+        }
+
+        return $exception->getMessage();
```
vitek-rostislav: this (actually the messages defined in constants in `PaymentTransactionRefundFacade` ) means some validation messages are not translated. E.g. when I switch to Czech, and try to refund `-10` CZK, I got "Refund amount must be greater than zero." error message

packages/framework/tests/Unit/Model/Payment/Transaction/PaymentTransactionTest.php
```diff
@@ -0,0 +1,41 @@
...
+use Shopsys\FrameworkBundle\Component\Money\Money;
+use Shopsys\FrameworkBundle\Model\Payment\Payment;
+use Shopsys\FrameworkBundle\Model\Payment\Transaction\PaymentTransactionData;
+use Shopsys\FrameworkBundle\Model\Payment\Transaction\PaymentTransactionDataFactory;
+use Shopsys\FrameworkBundle\Model\Payment\Transaction\PaymentTransactionFactory;
+
+class PaymentTransactionTest extends TestCase
+{
+    public function testPaymentTransactionWithoutPaymentIsNotRefundable(): void
+    {
```
vitek-rostislav, line 22: maybe we could add a few more cases to also cover the rest of the logic? Not gopay/not the right status is also considered not refundable. (I know it was already uncovered, but... :innocent:)

packages/administration/src/Component/OrderDetail/LiveComponent/ItemsTabComponent.php
```diff
@@ -0,0 +1,290 @@
...
+    {
+        $order = $this->getOrder();
+        $this->orderData = $this->orderDataFactory->createFromOrder($order);
+
+        return $this->formFactory
+            ->createNamedBuilder('order_form', FormType::class, $this->orderData, [
+                'data_class' => OrderData::class,
+                'attr' => [
+                    'novalidate' => 'novalidate',
+                    'data-model' => 'norender|*',
```
vitek-rostislav: `getDataModelValue` method should be used here or removed as unused :)

packages/administration/src/Component/OrderDetail/LiveComponent/ItemsTabComponent.php
```diff
@@ -0,0 +1,290 @@
...
+            $this->newItemCounter = max($this->newItemCounter, $newItemNumber + 1);
+        }
+
+        return OrderData::NEW_ITEM_PREFIX . $this->newItemCounter++;
+    }
+
+    /**
+     * @return array<string, string|null>
+     */
+    protected function createEmptyItemFormValues(): array
```
vitek-rostislav: I would like to brainstorm, whether we are able to better connect this componet to `OrderItemFormType`...(I have no idea how)

project-base/app/tests/App/Functional/Component/OrderDetail/ItemsTabComponentTest.php
```diff
@@ -0,0 +1,214 @@
...
+
+        $this->assertTrue($component->editMode);
+        $this->assertCount($originalItemsCount + 1, $component->formValues['orderItems']['itemsWithoutTransportAndPayment']);
+        $this->assertSame(
+            'Edited before adding item',
+            $component->formValues['orderItems']['itemsWithoutTransportAndPayment'][$itemKey]['name'],
+        );
+        $this->assertArrayHasKey('new_0', $component->formValues['orderItems']['itemsWithoutTransportAndPayment']);
+        $this->assertSame('', $component->formValues['orderItems']['itemsWithoutTransportAndPayment']['new_0']['unitPriceWithoutVat']);
+        $this->assertSame('', $component->formValues['orderItems']['itemsWithoutTransportAndPayment']['new_0']['vatPercent']);
```
vitek-rostislav: maybe we can add a nicely named variable for `$component->formValues['orderItems']['itemsWithoutTransportAndPayment']['new_0']`

project-base/app/tests/App/Functional/Component/OrderDetail/ItemsTabComponentTest.php
```diff
@@ -0,0 +1,214 @@
...
+    public function testPrefillTransportAndPaymentUpdatesFormValues(): void
+    {
+        $transport = $this->getReference(TransportDataFixture::TRANSPORT_CZECH_POST, Transport::class);
+        $payment = $this->getReference(PaymentDataFixture::PAYMENT_CASH_ON_DELIVERY, Payment::class);
+        $component = $this->createItemsTabComponent($this->getOrder());
+
+        $component->prefillTransport($transport->getId());
+        $component->prefillPayment($payment->getId());
+
+        $this->assertSame('1', $component->formValues['orderItems']['orderTransport']['setPricesManually']);
```
vitek-rostislav: we could also assert proper VAT is prefilled (expected value could be taken from the transport/payment)

packages/administration/translations/messages.cs.po
```diff
@@ -2263,12 +2446,21 @@ msgstr "Zde můžete přetáhnout další obsah."
 msgid "You can drag articles here."
 msgstr "Zde můžete přetáhnout články"
 
+msgid "You can refund only %refundableAmount%."
+msgstr "Můžete refundovat pouze %refundableAmount%."
+
+msgid "You can set refunded amount only up to %paidAmount%."
+msgstr "Refundovanou částku můžete nastavit maximálně na %paidAmount%."
```
vitek-rostislav: this should be in validators domain, right?

packages/administration/translations/messages.cs.po
```diff
@@ -2263,12 +2446,21 @@ msgstr "Zde můžete přetáhnout další obsah."
 msgid "You can drag articles here."
 msgstr "Zde můžete přetáhnout články"
 
+msgid "You can refund only %refundableAmount%."
+msgstr "Můžete refundovat pouze %refundableAmount%."
```
vitek-rostislav: this should be in validators domain, right?

packages/administration/templates/content/order/detail/components/withdrawal_status_modal.html.twig
```diff
@@ -0,0 +1,53 @@
...
+    {% embed '@ShopsysAdministration/content/order/detail/components/_modal.html.twig' with {
+        title: 'Withdrawal request'|trans,
+        dialogClass: 'modal-xl',
+    } %}
+        {% block close %}
+            <a href="{{ url('admin_order_edit', { id: this.orderId }) }}"
+               class="btn-close"
+               aria-label="{{ 'Close'|trans }}"
+               data-loading="addClass(disabled)"
+            ></a>
```
vitek-rostislav, line 11: when I see so many `data-loading="addClass(disabled)"` additions, I wonder, could not we introduce some component (whatever "component" means here) that would ancapsulate this behavior? As it seems we basically want it for all UX buttons, right?

vitek-rostislav (changes requested):
Hi, very impressive job there! I have still some doubts about the new stack accessibility to regular devs, though :smile: It will need some time for the onboarding. During the review, I have also drafted some explanation of the ecosystem to understand the concepts directly on our code. It is now in a form of HTML, but I am convinced we will need something like this in our docs, so I am planning to polish it and convert into markdown. You can see the preview here - https://refined-github-html-preview.kidonng.workers.dev/shopsys/shopsys/raw/8d9e09cafaf0fa8709bd9ab497f384aa8d4acffd/symfony-ux-in-admin.html - it basically represents things I had to understand and learn during the review and I can imagine it will be in some form usefull for our colleagues as well. (Yes, we should not replace the official docs, but I believe if we keep it lean and tight to our implementation, it could have a great additional value). 

There are tons of comments in the review, but do not worry, I am open to discussion of course, and ready to go through them together. 

There is one big concern about the re-usability of some components, I believe we will be able to at least prepare a well-structured task for this so we can address it in a separate issue, lets talk about it, we will see :)

packages/administration/templates/form/vertical.html.twig
grossmannmartin: Yes, it's intentional. This template must not be auto-registered with `\Shopsys\AdministrationBundle\DependencyInjection\ShopsysAdministrationExtension::autoRegisterFormThemes()` method.

project-base/app/assets/controllers.json
grossmannmartin: I agree it is not ideal, but I think this is the right boundary with Encore/Symfony UX. 
The app owns the Webpack build and `enableStimulusBridge()` points to an app-level controllers.json, while the administration package only contributes controllers loaded from @shopsys/administration/src/controllers. 

Keeping the bootstrap in project-base also preserves the place for project-specific Stimulus controllers.

packages/administration/src/Component/Security/AccessControl/LiveComponentAccessControlDataProvider.php
grossmannmartin: Vendor components without live actions will work just fine (for example, an icon component). 
I did not find the vendor component already with live action. But I assume we would have to figure out how, when necessary.

But my first instinct is that we would not just drop in use such a component, as we don't have any way to define the permissions for it.

packages/administration/src/Component/FlashMessage/LiveComponentFlashMessageSubscriber.php
grossmannmartin: Implemented, and I was able to simplify things thanks to this:
739d444447, 42ad839f1c, f046aaffdd, 51c45f522f, b7a5b44f24

packages/administration/templates/partial/flash_message/all_messages.html.twig
grossmannmartin: renamed to all_messages.html.twig in dc60427331

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
grossmannmartin: When we need this on another page, we can generalize it. 
Now, I don't think it's achievable easily. It's quite bound to the orders (getting/storing data, permissions, ...). 

I think the order page is quite unique, and other pages will be more form-based.

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
grossmannmartin: They're now constansts - 09fea43978

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
grossmannmartin: renamed and somehow unified in baf8db94a2, 23620335a3, 368fe8c10b, 2aca3542f4, d8632a089e, 4db2d6f674 and 524bb6be48

The last one has constants for consistency, but JS is still coupled by strings.

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
grossmannmartin: Because invalid characters later fails in symfony code on 

```
An exception has been thrown during the rendering of a template 
("The name "order_edit_+ě-g8965`#@$~^&*~$#@" contains illegal characters. 
Names should start with a letter, digit or underscore and only contain letters, digits, numbers, underscores ("_"), hyphens ("-") and colons (":").") in
 "@ShopsysAdministration/content/order/detail/components/section_editor.html.twig" at line 20.
```

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorFormComponent.php
grossmannmartin: Good catch. I have moved it upper in the layers in 078c4a3a94

packages/administration/src/Component/OrderDetail/OrderDetailSectionRegistry.php
grossmannmartin: yes - 67f3a5b24a

docs/cookbook/extending-order-detail-page.md
grossmannmartin: I'm aware of this limitation. From my point of view, it's a price to pay for custom-designed forms. 
We may somehow unify the rendering of the whole group, but there are a lot of exceptions:
- basicInformationGroup.status is not rendered (Ok, this may be removed, maybe)
- deliveryAddressForm.html.twig render two columns
- billingAddressForm.html.twig renders fields from two groups (and in columns)

I'm afraid that unifying it just somewhere is more harmful than explicitly listing them everywhere.

But I have realized that we don't need most of the fields in the orderformtype anymore. So I did a little cleanup in the de65178. We can discuss it from the point of BC breaks

packages/administration/templates/content/order/detail/sections/personal_form.html.twig
grossmannmartin: About templates - see previous comment, please. We can discuss ofc.

I have unified the view/form naming in the b42d2e627e

docs/cookbook/extending-order-detail-page.md
grossmannmartin: It makes sense. I have wrapped the content in a block, so it's now possible to do this

```twig
{% extends '@!ShopsysAdministration/content/order/detail/sections/payment_transport_form.html.twig' %}

{% block section_content %}
    {# pre-vendor form fields #}

    {{ parent() }}

    {# post-vendor form fields #}
{% endblock %}
```

see 3553a1bd6e

docs/cookbook/extending-order-detail-page.md
grossmannmartin: I am not able to render it nicely. Auto render "somewhere" is against the purpose of the "nice and usable order detail page" imho

packages/administration/templates/content/order/detail/components/section_editor.html.twig
grossmannmartin: 0de2d18f97

packages/administration/templates/content/order/detail/components/section_editor.html.twig
grossmannmartin: 0de2d18f97

packages/administration/src/Component/OrderDetail/OrderDetailSectionProvider.php
grossmannmartin: renamed in earlies commits

grossmannmartin:
> Random note from testing - seems like the popover is not displayed properly in the modal, see.
> 
> image
> "This phone number may not be valid. Please verify it." should be displayed there, but it is not. It happens when the warning is added dynamically, my steps to repreduce:
> 
> 1. edit delivery addres, set same as billing, save
> 2. open the delivery address form again, uncheck "same as billing", select telephone prefix
> 3. see dynamically added warning icon in the phone number input, without popover

Fixed in b8a112a0ea. The main problem was that the component was re-rendered on each field change.

packages/administration/assets/src/controllers/delivery_address_form_controller.js
grossmannmartin: Great catch. Fixed in be427cb07d

packages/administration/src/Controller/OrderDetailController.php
grossmannmartin: removed in 4dfe295c4a

packages/administration/src/Controller/OrderDetailController.php
grossmannmartin: I would not do it. There may be a complex logic tight to the status change

packages/administration/templates/content/order/detail/sections/_basic_information_view.html.twig
grossmannmartin: I have re-implemented it using CSS in ed8b5fded4

packages/administration/templates/content/order/detail/sections/_basic_information_view.html.twig
grossmannmartin: renamed 80fc0b3ff5

packages/administration/templates/content/order/detail/sections/billing_address_view.html.twig
grossmannmartin: good idea - ef505d8c5f, 7b16340081, and c264dbd8f3

packages/administration/src/Component/OrderDetail/OrderDetailTabProvider.php
grossmannmartin: makes sense - f8fdbac45a, ddd3ec31f3, 86793ff9b1, d3abadb8d1

packages/administration/assets/src/controllers/order_detail_tabs_controller.js
grossmannmartin: I would like to keep generalization to the point where we actually use tab elsewhere, wdyt?

packages/administration/src/Controller/OrderDetailController.php
grossmannmartin: I knew I forgot something 😁 
Thanks, implemented in c72a0d99e0

packages/administration/templates/content/order/detail/edit.html.twig
grossmannmartin: removed 9fa5f05bbd

packages/administration/src/Component/OrderDetail/LiveComponent/WithdrawalStatusModalComponent.php
grossmannmartin: make sense 94081203ae

packages/administration/src/Controller/OrderDetailController.php
grossmannmartin: added in 1a1edb7ac1

packages/administration/src/Controller/OrderDetailController.php
grossmannmartin: eec6770781 and fae204ec0d

packages/administration/tests/Unit/Component/EntityLog/Timeline/TwigComponent/EntityLogTimelineComponentTest.php
grossmannmartin: removed 15ad65f624

packages/administration/tests/Unit/Component/EntityLog/Timeline/TwigComponent/EntityLogTimelineComponentTest.php
grossmannmartin: 297a7da756

docs/cookbook/extending-order-detail-page.md
grossmannmartin: In 40ac4b7687 along with other changes from the CR reflected

docs/cookbook/extending-order-detail-page.md
grossmannmartin: 👍  40ac4b7687

packages/administration/templates/form/vertical.html.twig
grossmannmartin: I tried... and all attempts were actually worse. We can discuss

packages/administration/src/Component/OrderDetail/LiveComponent/PaymentTransactionsTabComponent.php
grossmannmartin: 55c48e28a0

packages/administration/templates/content/order/detail/components/payment_transactions_tab.html.twig
grossmannmartin: The whole component is bound to the gopay. I would generalize it with another payment provider implemented, if you agree.

packages/administration/templates/form/content/PaymentTransactionsType.html.twig
grossmannmartin: Yes, yes it can. 0e1d60156c

packages/administration/src/Component/OrderDetail/LiveComponent/PaymentTransactionsTabComponent.php
grossmannmartin: Yes, this code was a coupled, redundant, ambiguous patchwork (crap for short). 

Rewritten in 3bca63706e

packages/framework/tests/Unit/Model/Payment/Transaction/PaymentTransactionTest.php
grossmannmartin: added in e5f80f6de3

packages/administration/src/Component/OrderDetail/LiveComponent/ItemsTabComponent.php
grossmannmartin: getDataModelValue is now used. Also, I made a few fixes in JS. It should be now more reliable on re-renders - 14cb5339db

project-base/app/tests/App/Functional/Component/OrderDetail/ItemsTabComponentTest.php
grossmannmartin: I've made a little cleanup in b7cc72514d

project-base/app/tests/App/Functional/Component/OrderDetail/ItemsTabComponentTest.php
grossmannmartin: Added in b7cc72514d

packages/administration/translations/messages.cs.po
grossmannmartin: changed in earlier commits

packages/administration/translations/messages.cs.po
grossmannmartin: changed in earlier commits

packages/administration/src/Component/OrderDetail/LiveComponent/ItemsTabComponent.php
grossmannmartin: What about this 6f71874c1c?
It removes most of the duplication and string-based coupling.

packages/administration/templates/content/order/detail/components/withdrawal_status_modal.html.twig
grossmannmartin: What about this f4778d2692? I did not add it to all buttons, because it seemed unnecessary for buttons just having data-loading attr

grossmannmartin:
> Another note from testing:
> 
> I noticed the submit button is always active, and it is possible to "submit" the form even if there are validation errors. Could we improve this UX?
> 
>  Screencast.from.2026-06-09.12-39-08.mp4

I'm not sure about this. I think it could overcomplicate things, and it would be really easy to lock the user in a disabled state. Validations are server-side, so we would have had to duplicate the logic on the client or re-render the component on each keystroke.

grossmannmartin:
> Another note from testing - when changing the status to withdrawn, I see a modal where I need to set the withdrawal data. This makes sense but I am afraid it might be confusing - when admin closes the modal, the status is not changed. Maybe we could add a warning or something like thta, lets talk about it, please

Yes, that makes sense. What about this? 1c165db555

grossmannmartin:
> Yet another thing I noticed - the history tab never gets updated when performing partial edits in the modals. Admin has to reload the whole page to see the changes in the tab. Do we want to improve that? 😇
> 
>  Screencast.from.2026-06-09.16-53-47.mp4

What about this? 92b29e0ea1

packages/administration/src/Component/OrderDetail/LiveComponent/PaymentTransactionsTabComponent.php
```diff
@@ -0,0 +1,262 @@
...
+        try {
+            $this->paymentTransactionRefundFacade->changeManualRefundedAmount($paymentTransaction, $this->paymentTransactionRefundData->refundedAmount);
+            $this->flashMessageService->addSuccessFlash(t('Refunded amount has been changed.'));
+            $this->emit(SectionEditorFormComponent::ORDER_DETAIL_SECTION_UPDATED_EVENT);
+            $this->dispatchBrowserEvent(SectionEditorFormComponent::ORDER_DETAIL_UPDATED_BROWSER_EVENT);
+            $this->closeRefundModal();
+        } catch (RefundedAmountGreaterThanPaidAmountException) {
+            $this->addFormError(
+                'refundedAmount',
+                t('You can set refunded amount only up to %paidAmount%.', [
```
vitek-rostislav: these messages should rather be in the validtors domain, but I can live with the current state :)

vitek-rostislav: Ah, now I see this is the only one missing the domain, was that on purpose?

packages/administration/templates/content/order/detail/components/items_tab.html.twig
```diff
@@ -0,0 +1,33 @@
...
+                    label: 'Cancel'|trans,
+                    buttonClass: 'btn btn-outline-secondary js-order-items-edit-only' ~ (not this.editMode ? ' d-none' : ''),
+                    preventDefault: false,
+                }) }}
+                {{ component('Admin:LiveActionButton', {
+                    action: 'save',
+                    label: 'Save items'|trans,
+                    loadingLabel: 'Saving items...'|trans,
+                    icon: 'checked',
+                    buttonClass: 'btn btn-primary js-order-items-edit-only' ~ (not this.editMode ? ' d-none' : ''),
```
vitek-rostislav: it seems `js-order-items-edit-only` is not used anywhere

packages/administration/src/Component/OrderDetail/LiveComponent/SectionEditorComponent.php
vitek-rostislav: ah, ok :+1:

docs/cookbook/extending-order-detail-page.md
vitek-rostislav: > But I have realized that we don't need most of the fields in the orderformtype anymore. So I did a little cleanup in the https://github.com/shopsys/shopsys/commit/de65178ef01a027dd36417d4b316552ec53a5a1a. We can discuss it from the point of BC breaks

Interesting. Yeah, regarding BC breaks, this will probably be hurtful. I did not notice whether it is explicitly mentioned in the upgrade notes now?

Maybe we could yet discuss another approach - defining the fields classes (or let's say appearance) directly in the form type. But I feel like it will not be easy anyway.

packages/administration/assets/src/controllers/order_detail_tabs_controller.js
vitek-rostislav: yes :)

packages/administration/templates/form/vertical.html.twig
vitek-rostislav: ok, lets talk about it

packages/administration/templates/form/vertical.html.twig
vitek-rostislav: So it means it could "override" the default theme, then? We just want to enforce the theme on particular forms by explicitly using `form_theme`, right?

packages/administration/src/Component/OrderDetail/LiveComponent/ItemsTabComponent.php
vitek-rostislav: yes! :smile: :ok_hand:

vitek-rostislav (approved):
There are just a few comments left 👌 The threads are somehow broken, so I will add a new comment here - as we discussed, it would be nice to move the `vertical.html.twig` template so it is in the same folder as `theme.html.twig`, and adjust `ShopsysAdministrationExtension::autoRegisterFormThemes()` method. I do not need to see the task again, you can move it to tests after the remaining comments are resolved 😉

packages/administration/src/Component/OrderDetail/LiveComponent/PaymentTransactionsTabComponent.php
grossmannmartin: No, it's a mistake. Fixed

packages/administration/src/Component/OrderDetail/LiveComponent/HistoryTabComponent.php
```diff
@@ -0,0 +1,35 @@
...
+use Shopsys\FrameworkBundle\Model\Order\Order;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+#[AsLiveComponent(
+    name: self::COMPONENT_NAME,
+    template: '@ShopsysAdministration/content/order/detail/components/history_tab.html.twig',
+)]
+class HistoryTabComponent
```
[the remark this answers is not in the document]
grossmannmartin: rejected. This is not a real CWE, as the component has NO live action.
Adding class-level #[ForRole] alone would not protect default renders anyway, because the live-component access subscriber skips _component_default_action

packages/administration/src/Component/OrderDetail/LiveComponent/WithdrawalStatusModalComponent.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        $this->submitForm();
+
+        if ($this->withdrawalRequestData === null) {
+            return null;
+        }
+
+        $order = $this->getOrder();
+        $orderData = $this->orderDataFactory->createFromOrder($order);
+        $orderData->status = $this->orderStatusFacade->getById($this->statusId);
```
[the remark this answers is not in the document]
grossmannmartin: Rejected.

The key premise is wrong: statusId is not client-controllable in the useful sense. It is #[LiveProp], but not writable: true, so the client cannot update it as model data. If the client tampers with the original prop value, the Live Component checksum fails. The only current place that creates a valid signed statusId for this modal is [editWithdrawalStatusAction (line 63)](/Volumes/work/shopsys/framework/packages/administration/src/Controller/OrderDetailController.php:63), and that already rejects non-TYPE_WITHDRAWN statuses.

pk16011990 (approved):

PR #4640, opened 2026-06-02
Allow disabling MCP in production by omitting database credentials

vitek-rostislav:

The MCP integration connects to the database using its own dedicated credentials (`MCP_DATABASE_USER` / `MCP_DATABASE_PASSWORD`). These were defined in the main `.env` file, which is loaded in every environment — including production. As a result, the feature was effectively always enabled, with no straightforward way to turn it off on a production deployment.

This PR moves the default MCP credentials out of the main `.env` and into the development (`.env.dev`) and test (`.env.test`) environments only, and sets them explicitly in the CI and review Docker Compose configurations. Production deployments no longer receive these credentials by default, so a project can now disable MCP simply by not providing them — while local development, tests, and review apps keep working out of the box.

Additionally, a failure to generate the MCP schema during cache warmup is now logged as an error instead of a warning, so the problem becomes visible in Sentry instead of being silently buried among warnings.

#### Fixes issues

N/A

----
  - https://rv-mcp-vars.odin.shopsys.cloud
  - https://cz.rv-mcp-vars.odin.shopsys.cloud
  - https://rv-mcp-vars.odin.shopsys.cloud/sk

----
  - https://rv-mcp-vars.odin.shopsys.cloud
  - https://cz.rv-mcp-vars.odin.shopsys.cloud
  - https://rv-mcp-vars.odin.shopsys.cloud/sk


docker/conf/docker-compose.github-actions.review.yml.dist
```diff
@@ -140,6 +142,8 @@ services:
             GOOGLE_CLIENT_SECRET: "${GOOGLE_CLIENTSECRET}"
             SEZNAM_CLIENT_ID: "${SEZNAM_CLIENTID}"
             SEZNAM_CLIENT_SECRET: "${SEZNAM_CLIENTSECRET}"
+            MCP_DATABASE_USER: shopsys_mcp
+            MCP_DATABASE_PASSWORD: shopsys_mcp
```
grossmannmartin: Do we need mcp enabled on the consumer? (same in the simplified version and gitlab review)

grossmannmartin (approved):

docker/conf/docker-compose.github-actions.review.yml.dist
vitek-rostislav: Good point, probably not at all, removing

pk16011990 (approved):

PR #4641, opened 2026-06-03
Add customer account order and complaint filtering

chlebektomas:

This PR improves the customer account order and complaint history so customers can find past records faster and understand the current result set more clearly.

Customers can now search orders and complaints by number or product, filter them by date, switch between status tabs with counts, and clear filters consistently. The order and complaint views now also share the same visual language for lists, details, empty states, loading states, and responsive behavior.

The API now exposes the filtering and counting data needed by the storefront, so filtered views remain shareable through the URL and can be reused consistently across customer account pages.

#### Fixes issues
N/A

----
  - https://tc-ssp-4087-order-list-filtering.odin.shopsys.cloud
  - https://cz.tc-ssp-4087-order-list-filtering.odin.shopsys.cloud
  - https://tc-ssp-4087-order-list-filtering.odin.shopsys.cloud/sk

----
  - https://tc-ssp-4087-order-list-filtering.odin.shopsys.cloud
  - https://cz.tc-ssp-4087-order-list-filtering.odin.shopsys.cloud
  - https://tc-ssp-4087-order-list-filtering.odin.shopsys.cloud/sk

----
  - https://tc-ssp-4087-order-list-filtering.odin.shopsys.cloud
  - https://cz.tc-ssp-4087-order-list-filtering.odin.shopsys.cloud
  - https://tc-ssp-4087-order-list-filtering.odin.shopsys.cloud/sk


project-base/storefront/graphql/requests/orders/queries/OrdersQuery.graphql
```diff
@@ -1,8 +1,33 @@
...
+        totalCount
+    }
+    doneOrders: orders (first: 1, filter: $doneStatusFilter) {
+        totalCount
+    }
+    canceledOrders: orders (first: 1, filter: $canceledStatusFilter) {
+        totalCount
+    }
+    withdrawnOrders: orders (first: 1, filter: $withdrawnStatusFilter) {
+        totalCount
```
grossmannmartin: I don't have a good feeling about these aliases. 

For newly added status, this is another place that has to be updated. Also, each alias bring performance overhead. 
Complaints are solved differently, can't we use the same pattern here?

project-base/storefront/components/Pages/Customer/Orders/OrdersStatusTabs.tsx
```diff
@@ -0,0 +1,93 @@
...
+        />
+    );
+};
+
+const getStatusTabs = (t: ReturnType<typeof useTranslation>['t'], orderStatusCounts: OrderStatusCounts) => [
+    { status: null, label: t('All'), count: orderStatusCounts.all },
+    { status: TypeOrderStatusEnum.New, label: t('New'), count: orderStatusCounts[TypeOrderStatusEnum.New] },
+    {
+        status: TypeOrderStatusEnum.InProgress,
+        label: t('In progress'),
```
grossmannmartin: Labels are possible to set in the administration, I don't think we should hardcode default values here.

Also, it's possible to create new statuses in the administration. They will have type "in_process", but I'm not sure if such status should appear separately or not.

packages/frontend-api/src/Model/Resolver/Complaint/ComplaintsQuery.php
```diff
@@ -45,29 +49,70 @@ public function complaintResolutionQuery(): array
...
 
+    public function complaintStatusesQuery(): array
+    {
+        return $this->complaintApiFacade->getComplaintStatuses($this->domain->getLocale());
+    }
+
+    public function complaintStatusCountsQuery(Argument $argument): array
+    {
+        $customerUser = $this->currentCustomerUser->getCurrentCustomerUser();
+        $filter = $this->complaintFilterFactory->createFromArgument($argument)->withoutStatus();
```
grossmannmartin: Silently removing some field from the filter may be surprising from the API's consumer point. 

I would rather make the filter without status and pass status extra only where needed. 
Or make it optional, but then this method should somehow reflect the passed status

packages/frontend-api/src/Model/Resolver/Complaint/ComplaintsQuery.php
```diff
@@ -45,29 +49,70 @@ public function complaintResolutionQuery(): array
         return $this->complaintResolutionEnum->serialize();
     }
 
+    public function complaintStatusesQuery(): array
```
grossmannmartin: this query does not seem used on the storefront. I would remove it

grossmannmartin (changes requested):

packages/frontend-api/src/Model/Complaint/ComplaintApiFacade.php
```diff
@@ -155,23 +163,88 @@ public function getCustomerComplaintsLimitedList(
...
+                'count' => $countsByStatusId[$complaintStatus->getId()] ?? 0,
+            ],
+            $this->complaintStatusFacade->getAll(),
+        );
+    }
+
+    /**
+     * @return array{id: int, name: string}
+     */
+    protected function getComplaintStatusData(ComplaintStatus $complaintStatus, string $locale): array
```
grossmannmartin: nitpick: this should be rather `createComplaintStatusData`

packages/frontend-api/src/Model/Complaint/ComplaintApiFacade.php
```diff
@@ -155,23 +163,88 @@ public function getCustomerComplaintsLimitedList(
...
+    {
+        return [
+            'id' => $complaintStatus->getId(),
```
grossmannmartin: exposing internal auto-incremented ID is highly discouraged. The value may, in rare occasions, change and it's not considered a best practice for external system to rely on it.

packages/frontend-api/src/Model/Complaint/ComplaintFilter.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    /**
+     * @param string[]|null $statuses
+     * @param int[]|null $statusIds
+     */
+    public function __construct(
+        protected ?DateTimeImmutable $createdAfter = null,
+        protected ?DateTimeImmutable $createdBefore = null,
+        protected ?array $statuses = null,
+        protected ?array $statusIds = null,
```
grossmannmartin: Filter rely on two input fields representing the same data. What will happen if those two array will not align?
Just one should be enough (preferred), or a wrapping DTO (data transfer object) should be introduced.

packages/frontend-api/src/Model/Order/OrderFilterFactory.php
```diff
@@ -22,13 +22,31 @@ public function createFromArgument(Argument $argument): OrderFilter
...
 
-        $status = isset($filter['status']) ? $this->orderStatusFacade->getAllByType($filter['status']) : null;
+        $statuses = [];
+
+        if (isset($filter['status'])) {
+            $statuses[] = $filter['status'];
+        }
+
+        if (isset($filter['statuses'])) {
+            $statuses = array_merge($statuses, $filter['statuses']);
```
grossmannmartin: `status` and `statuses` seem redundant. Why not just statuses and pass single-value array for one?

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Complaint/Input/ComplaintFilterInputDecorator.types.yaml
```diff
@@ -0,0 +1,24 @@
...
+                type: "String"
+                description: "Filter complaints by complaint number or product"
+            status:
+                type: "ComplaintStatusEnum"
+                description: "Filter complaints by status"
+            statuses:
+                type: "[ComplaintStatusEnum!]"
+                description: "Filter complaints by statuses"
+            statusIds:
+                type: "[Int!]"
```
grossmannmartin: Those three seem hard to understand from the consumer's point of view. 
I would recommend just single 
```graphql
statuses:
    type: "[ComplaintStatusEnum!]!"
```

I would make it mandatory unless there is a distinction between `[]` and `null`

grossmannmartin (changes requested):

packages/framework/src/Model/Complaint/Status/ComplaintStatusData.php
```diff
@@ -11,8 +11,14 @@ class ComplaintStatusData
...
 
+    /**
+     * @var string|null
+     */
+    public $code;
+
     public function __construct()
     {
         $this->name = [];
+        $this->code = null;
```
grossmannmartin: This initialization is not necessary, nor expected. 
A) untyped property has null as a default value implicitly.
B) explicitly set null tells me that null is allowed value, which is not.

TLDR: please remove this line 😁

packages/framework/src/Model/Complaint/Status/ComplaintStatusFacade.php
```diff
@@ -16,11 +17,13 @@ public function __construct(
         protected readonly ComplaintStatusRepository $complaintStatusRepository,
         protected readonly MailTemplateFacade $mailTemplateFacade,
         protected readonly ComplaintMail $complaintMail,
+        protected readonly TransformStringHelper $transformStringHelper,
     ) {
     }
 
     public function create(ComplaintStatusData $complaintStatusData): ComplaintStatus
     {
+        $complaintStatusData->code = $this->createUniqueCode($complaintStatusData);
```
grossmannmartin: I would rather remove $code from ComplaintStatusData completely and pass it as another argument to the ComplaintStatusFactory::create(). 
That way it will stays immutable and the class API will be honest about immutability.

packages/framework/src/Model/Order/Status/OrderStatusFacade.php
```diff
@@ -17,11 +18,13 @@ public function __construct(
         protected readonly OrderRepository $orderRepository,
         protected readonly MailTemplateFacade $mailTemplateFacade,
         protected readonly OrderStatusFactory $orderStatusFactory,
+        protected readonly TransformStringHelper $transformStringHelper,
     ) {
     }
 
     public function create(OrderStatusData $orderStatusFormData): OrderStatus
     {
+        $orderStatusFormData->code = $this->createUniqueCode($orderStatusFormData);
```
grossmannmartin: The same comments as for complaints are valid even for orders.

project-base/app/tests/FrontendApiBundle/Functional/Order/graphql/getOrders.graphql
```diff
@@ -1,4 +1,4 @@
-query getOrders ($first: Int, $last: Int, $filter: OrderFilterInput) {
+query getOrders ($first: Int, $last: Int, $filter: OrderFilterInput, $statuslessFilter: OrderFilterInput) {
```
grossmannmartin, line 1: can't this remain just $filter? or it's somehow specific here?

packages/frontend-api/src/Model/Order/OrderRepository.php
```diff
@@ -6,18 +6,18 @@
...
 
 class OrderRepository
 {
-    protected EntityManagerInterface $em;
-
-    public function __construct(EntityManagerInterface $entityManager)
-    {
-        $this->em = $entityManager;
+    public function __construct(
+        protected EntityManagerInterface $em,
```
grossmannmartin: this may be readonly

project-base/storefront/components/Pages/Customer/Orders/OrdersStatusTabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+                        );
+                    })}
+                </div>
+            )}
+        />
+    );
+};
+
+const getStatusTabs = (t: ReturnType<typeof useTranslation>['t'], orderStatusCounts: OrderStatusCount[]) => {
+    const allOrdersCount = orderStatusCounts.reduce((totalCount, { count }) => totalCount + count, 0);
```
grossmannmartin, line 74: maybe we could hide statuses that does not have any orders, what do you think? 

So instead of this 

Image

we could have this

Image

grossmannmartin (changes requested):

grossmannmartin (approved):

project-base/storefront/components/Forms/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,231 @@
...
+    day_button:
+        'flex size-9 cursor-pointer items-center justify-center rounded-button font-semibold text-sm transition hover:bg-fill-accent-less hover:text-link-hovered focus:bg-fill-accent-less focus:outline-hidden',
+    selected:
+        '[&_button]:bg-link-default [&_button]:text-text-inverted [&_button]:hover:bg-link-hovered [&_button]:hover:text-text-inverted',
+    today: '[&_button]:ring-2 [&_button]:ring-link-default [&_button]:ring-inset',
+    outside: 'text-input-placeholder-default opacity-50',
+    disabled: 'pointer-events-none opacity-40',
+};
+
+const DatePickerChevron = ({ className, orientation }: { className?: string; orientation?: string }) => {
```
fujta: Nezasloužila by si tato komponenta a funkce pod ní vlastní soubor? + ty funkce nějaké unit testy?

project-base/storefront/components/Pages/Customer/Complaints/ComplaintsFilter.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+export const ComplaintsFilter = forwardRef<ComplaintsFilterHandle>((_, ref) => {
+    const { t } = useTranslation();
+    const router = useRouter();
+    const [values, setValues] = useState<ComplaintsFilterFormValues>(() =>
+        getComplaintsFilterFormValuesFromUrlQuery(router.query),
+    );
+    const debouncedSearch = useDebounce(values.search, COMPLAINT_FILTER_SEARCH_DEBOUNCE_DELAY);
+
+    useImperativeHandle(ref, () => ({
```
fujta: Nejsem vyloženě proti tomu, ale nebylo by čistější použít globální stav místo imperative handlů?

project-base/storefront/components/Pages/Customer/Orders/OrdersFilter.tsx
```diff
@@ -0,0 +1,141 @@
...
+export const OrdersFilter = forwardRef<OrdersFilterHandle>((_, ref) => {
+    const { t } = useTranslation();
+    const router = useRouter();
+    const [values, setValues] = useState<OrdersFilterFormValues>(() =>
+        getOrdersFilterFormValuesFromUrlQuery(router.query),
+    );
+    const debouncedSearch = useDebounce(values.search, ORDER_FILTER_SEARCH_DEBOUNCE_DELAY);
+
+    useImperativeHandle(ref, () => ({
+        clear: () => setValues(emptyOrdersFilterFormValues),
```
fujta: Tady taky

project-base/storefront/pages/customer/complaints.tsx
```diff
@@ -42,18 +65,33 @@ const ComplaintsPage: FC = () => {
         { __typename: 'Link', name: t('My complaints'), slug: customerComplaintsUrl },
     ];
 
-    const [searchQueryValue, setSearchQueryValue] = useState('');
-    const { mappedComplaints, complaintsTotalCount, complaintsDataFetching, complaintsData } =
-        useComplaintsData(searchQueryValue);
+    const { mappedComplaints, complaintsTotalCount, complaintStatusCounts, complaintsDataFetching, complaintsData } =
+        useComplaintsData(filter, statuslessFilter);
```
fujta: Vím, že to není scope tohoto tasku, ale tato komponenta by si dle mě zasloužila větší code split. Teď se dle mě bude zbytečně rerenderovat celá stránka i s layoutem, jak se bude měnit filtr, ne?

project-base/storefront/pages/customer/orders.tsx
```diff
@@ -17,36 +19,85 @@ import { TypeCustomerUserRoleEnum } from 'graphql/types';
...
+    const router = useRouter();
+    const paginationScrollTargetRef = useRef<HTMLDivElement>(null);
+    const ordersFilterRef = useRef<OrdersFilterHandle>(null);
     const currentPage = useCurrentPageQuery();
     const { url } = useDomainConfig();
+    const filter = getOrdersFilterFromUrlQuery(router.query);
+    const activeStatusCode = getOrderStatusCodeFromUrlQuery(router.query[ORDER_STATUS_QUERY_PARAMETER_NAME]);
+    const hasActiveFilters = hasActiveOrderListFiltersFromUrlQuery(router.query);
+    const statuslessFilter = getOrdersStatuslessFilterFromUrlQuery(router.query);
     const [{ data: ordersData, fetching: areOrdersFetching }] = useOrdersQuery({
```
fujta: Tady taky

project-base/storefront/utils/parsing/getDateTimeFromDateQuery.ts
```diff
@@ -0,0 +1,29 @@
...
+    const date = getStringFromUrlQuery(dateQuery);
+
+    if (!getIsDateQueryValid(date)) {
+        return null;
+    }
+
+    return `${date}T${isEndOfDay ? '23:59:59' : '00:00:00'}+00:00`;
+};
+
+export const getIsDateQueryValid = (date: string): boolean => {
```
fujta: Ultranitpick

```suggestion
export const isDateQueryValid = (date: string): boolean => {
```

fujta (changes requested):

fujta (approved):

packages/frontend-api/src/Model/DateFilter/DateFilterDateTimeFactory.php
```diff
@@ -0,0 +1,39 @@
...
+        return $this->createDateTimeInUtc($date, '23:59:59');
+    }
+
+    protected function createDateTimeInUtc(?string $date, string $time): ?DateTimeImmutable
+    {
+        if ($date === null) {
+            return null;
+        }
+
+        return (new DateTimeImmutable(sprintf('%s %s', $date, $time), $this->domain->getDateTimeZone()))
```
grossmannmartin: Please use Symfony\Component\Clock\DatePoint instead. 
See https://docs.shopsys.com/en/20.0/introduction/working-with-date-time-values/#parsing-external-dates

edit: see my other comment first, please as I suggest a siginficant change there

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Order/Filter/OrderFilterInputDecorator.types.yaml
```diff
@@ -5,11 +5,17 @@ OrderFilterInputDecorator:
         description: "Filter orders"
         fields:
             createdAfter:
-                type: "DateTime"
-                description: "Filter orders created after this date"
-            status:
-                type: "OrderStatusEnum"
+                type: "Date"
```
grossmannmartin: Now we have difference between input and output of the API. 
Imagine the complaint created at `2024-09-21 1:15:43 CEST` (`2024-09-20 23:15:43 UTC`)

Then this query
```
{
  complaints(first: 10, filter: {createdAfter: "2024-09-21", createdBefore: "2024-09-21"}) {
    edges {
      node {
        createdAt
      }
    }
  }
}
```

will return 

```
{
  "data": {
    "complaints": {
      "edges": [
        {
          "node": {
            "createdAt": "2024-09-20T23:15:43+00:00"
          }
        }
      ]
    }
  }
}
```

Notice the discrepancy of the createdAfter filter date and returned createdAt date (1 day shift).

I would much rather keep the DateTime as input type (I would drop DateType completely as it's confusing from my pov), convert to UTC on the storefront (as the output dates are converted anyway) and keep the straight narrative that API communicate date time values in the UTC.
And I'm fine with the API accepting date-time ranges (meaning I can request complaints between 11 and 13 hours)

grossmannmartin (changes requested):

grossmannmartin (approved):

fujta (approved):

fujta (approved):

fujta (approved):

packages/frontend-api/src/Model/Complaint/ComplaintFilterFactory.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function __construct(
+        protected readonly ComplaintStatusFacade $complaintStatusFacade,
+    ) {
+    }
+
+    public function createFromArgument(Argument $argument): ComplaintFilter
+    {
+        if (!isset($argument['filter'])) {
+            $search = isset($argument['searchInput']['search']) ? trim($argument['searchInput']['search']) : null;
```
grossmannmartin: Do we need to keep the `searchIpnut` field? It seems like it's superseded by the `filter` field.

packages/frontend-api/src/Model/Order/OrderFilterFactory.php
```diff
@@ -22,13 +22,19 @@ public function createFromArgument(Argument $argument): OrderFilter
 
         $filter = $argument['filter'];
 
-        $status = isset($filter['status']) ? $this->orderStatusFacade->getAllByType($filter['status']) : null;
+        $statusCodes = $filter['statusCodes'] ?? [];
+
+        $orderStatuses = $this->orderStatusFacade->getAllByCodes(array_unique($statusCodes));
```
grossmannmartin: array_unique has a really small practical effect here. It does not alter the results (nor the sort order, nor the duplicities of returned rows).
If necessary, it should be the responsibility of the facade method itself (so it's not necessary to normalize input on each usage), but here I would say it may be omitted.

project-base/app/src/DataFixtures/Demo/ComplaintDataFixture.php
```diff
@@ -80,6 +88,91 @@ public function getDependencies(): array
         ];
     }
 
+    private function createAdditionalCustomerComplaints(): void
+    {
+        if ($this->environment === 'test') {
+            return;
+        }
```
grossmannmartin: I would really like to avoid setting different data for tests and demo data. 
It will be really hard to debug why some tests are failing.

project-base/app/src/DataFixtures/Demo/ComplaintDataFixture.php
```diff
@@ -80,6 +88,91 @@ public function getDependencies(): array
...
+            $customerUser = $this->customerUserRepository->findCustomerUserByEmailAndDomain(
+                'no-reply@shopsys.com',
+                $domainId,
+            );
+
+            if ($customerUser === null) {
+                continue;
+            }
+
+            assert($customerUser instanceof CustomerUser);
```
grossmannmartin: there is no way $customerUser could be anything else than CustomerUser class at this point.
I would remove this assert

project-base/app/src/DataFixtures/Demo/ComplaintStatusDataFixture.php
```diff
@@ -24,6 +28,7 @@ public function load(ObjectManager $manager): void
     {
         $this->createComplaintStatusReference(1, self::COMPLAINT_STATUS_NEW);
         $this->createComplaintStatusReference(2, self::COMPLAINT_STATUS_RESOLVED);
+        $this->createInProgressComplaintStatusReference();
```
grossmannmartin: If we need a new status, I would rather edit the `packages/framework/src/Migrations/Version20240816221930.php` migration and create a reference here. So it's unified

project-base/app/src/DataFixtures/Demo/OrderDataFixture.php
```diff
@@ -637,6 +640,58 @@ private function loadDefault(int $domainId): void
             TransportDataFixture::TRANSPORT_DRONE,
             PaymentDataFixture::PAYMENT_CARD,
         );
+
+        if ($this->environment !== 'test' && $noReplyCustomerUser !== null) {
```
grossmannmartin: same as for complaints. Let's keep it unfied and predictable for test and demo environment

project-base/app/src/DataFixtures/Demo/OrderDataFixture.php
```diff
@@ -637,6 +640,58 @@ private function loadDefault(int $domainId): void
...
+            $orderData = $this->orderDataFactory->create();
+            $orderData->status = $this->getReference(
+                $statusReferenceNames[($i - 1) % count($statusReferenceNames)],
+                OrderStatus::class,
+            );
+            $orderData->firstName = 'Generated';
+            $orderData->lastName = 'Order' . $i;
+            $orderData->email = 'no-reply@shopsys.com';
+            $orderData->telephone = new PhoneData('CZ', '+420', (string)(700000000 + $i));
+            $orderData->street = 'Generated Street ' . $i;
```
grossmannmartin: what about use Faker and have more realistic data?

project-base/app/src/Model/Order/Status/OrderStatus.php
```diff
@@ -6,9 +6,11 @@
 
 use Doctrine\ORM\Mapping as ORM;
 use Shopsys\FrameworkBundle\Model\Order\Status\OrderStatus as BaseOrderStatus;
+use Shopsys\McpAttributes\Attribute\AsMcpInheritedColumn;
 use Shopsys\McpAttributes\Attribute\AsMcpTable;
 
 #[AsMcpTable]
+#[AsMcpInheritedColumn(fieldName: 'code', exposed: true)]
```
grossmannmartin: This is not necessary if parent class exposes the column.

Also, in the parent class we can use (without explicit `exposed: true`)

```
#[AsMcpColumn]
#[ORM\Column(type: 'string', length: 255, unique: true)]
protected $code;
```

grossmannmartin (changes requested):

grossmannmartin (approved):

PR #4642, opened 2026-06-04
Unify order confirmation and payment return flow

grossmannmartin:

This PR unifies the customer experience after order creation and payment attempts under the order confirmation page. Customers now return from GoPay through a short-lived payment return hash instead of exposing order-identifying data in the payment gateway return URL, and the confirmation page can show the correct order sent, payment success, in-process, or failed state from the current order data.

The change also makes repeated payment attempts behave more consistently. The storefront can update the payment status directly on the confirmation page, preserve the order context during the flow, and report GTM payment events with retry counts derived from persisted payment transactions instead of temporary local storage. Obsolete payment-fail GTM events are removed, so analytics now uses the unified payment event for successful, in-process, and failed payment states.

For API consumers, the order now includes confirmation page content and payment transaction count directly, so clients can render the complete post-order state without relying on separate legacy order sent/payment content queries.

#### Fixes issues

closes #4420 

----
  - https://mg-order-confirmation-page.odin.shopsys.cloud
  - https://cz.mg-order-confirmation-page.odin.shopsys.cloud
  - https://mg-order-confirmation-page.odin.shopsys.cloud/sk


grossmannmartin:
@coderabbitai review

TomasLudvik (approved):

project-base/storefront/public/locales/en/common.json
```diff
@@ -662,6 +663,7 @@
     "Write what you are looking for...": "Write what you are looking for...",
     "Yes": "Yes",
     "You": "You",
+    "You are being redirected to the payment gateway.": "",
```
JanMolcik: This new translation key is currently mapped to an empty string. Please fill the value so the payment-gateway redirect state has visible localized copy instead of relying on fallback behavior.

project-base/storefront/public/locales/cs/common.json
```diff
@@ -698,6 +699,7 @@
     "Write what you are looking for...": "Napište, co hledáte...",
     "Yes": "Ano",
     "You": "Vy",
+    "You are being redirected to the payment gateway.": "",
```
JanMolcik: This new translation key is currently mapped to an empty string. Please provide the Czech copy for the payment-gateway redirect state.

project-base/storefront/public/locales/sk/common.json
```diff
@@ -698,6 +699,7 @@
     "Write what you are looking for...": "Napíšte, čo hľadáte...",
     "Yes": "Áno",
     "You": "Vy",
+    "You are being redirected to the payment gateway.": "",
```
JanMolcik: This new translation key is currently mapped to an empty string. Please provide the Slovak copy for the payment-gateway redirect state.

JanMolcik (approved):
Approved. I only left inline comments for the missing storefront translations.

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Order/OrderDecorator.types.yaml
```diff
@@ -106,6 +106,10 @@ OrderDecorator:
             lastExternalPaymentUrl:
                 type: "String"
                 description: "URL for accessing the last payment transaction on a gateway without invoking the new payment transaction. Depending on the payment status, user might see the payment details or even retry the transaction if possible."
+            paymentStatus:
```
grossmannmartin, line 109: I would name it `lastExternalPaymentStatus` so it's unified with the field above

PR #4643, opened 2026-06-05
github actions: use `xargs -r` to safely handle empty input in workflows

vitek-rostislav:

The "Clean up after branch merge" workflow started failing with exit code 123 after the recent `actions/checkout` upgrade to v6 (https://github.com/shopsys/shopsys/pull/4535). The cleanup step strips the checkout's auth header before pushing branch deletions to the split package repositories, using `git config | grep | xargs git config --unset-all`. Under v6 the checkout no longer leaves a matching `extraheader` config entry, so `grep` produced no output — and because `xargs` was running without `--no-run-if-empty`, it still invoked `git config --unset-all` once with no argument, which fails and aborts the whole step before the branch removal could run.

Adding `-r` (`--no-run-if-empty`) makes `xargs` a no-op when there is nothing to unset, so the workflow proceeds correctly. The same robustness fix is applied to the two monorepo split workflows that share the pattern; they were already guarded with `|| true`, but the flag avoids needlessly spawning a failing command.

#### Fixes issues

N/A

----
  - https://rv-fix-gh-actions.odin.shopsys.cloud
  - https://cz.rv-fix-gh-actions.odin.shopsys.cloud
  - https://rv-fix-gh-actions.odin.shopsys.cloud/sk

----
  - https://rv-fix-gh-actions.odin.shopsys.cloud
  - https://cz.rv-fix-gh-actions.odin.shopsys.cloud
  - https://rv-fix-gh-actions.odin.shopsys.cloud/sk

### DIFF
--- a/.github/workflows/monorepo-force-split-branch.yaml
+++ b/.github/workflows/monorepo-force-split-branch.yaml
@@ -28,5 +28,5 @@ jobs:
                     sudo apt-get install -y git-filter-repo
             -   name: Split repositories
                 run: |
-                    git config -l | grep 'http\..*\.extraheader' | cut -d= -f1 | xargs -L1 git config --unset-all || true
+                    git config -l | grep 'http\..*\.extraheader' | cut -d= -f1 | xargs -r -L1 git config --unset-all
                     bash ./.github/monorepo/split-repositories.sh "${{ inputs.branch_name }}" "${REMOTE_TEMPLATE}" true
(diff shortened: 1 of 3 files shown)


.github/workflows/monorepo-force-split-branch.yaml
```diff
@@ -28,5 +28,5 @@ jobs:
                     sudo apt-get install -y git-filter-repo
             -   name: Split repositories
                 run: |
-                    git config -l | grep 'http\..*\.extraheader' | cut -d= -f1 | xargs -L1 git config --unset-all || true
+                    git config -l | grep 'http\..*\.extraheader' | cut -d= -f1 | xargs -r -L1 git config --unset-all || true
```
grossmannmartin: || true is not necessary now, right? (even in the second split)

grossmannmartin (approved):

.github/workflows/monorepo-force-split-branch.yaml
vitek-rostislav: true indeed, removed so it is unified :slightly_smiling_face: :+1:

PR #4645, opened 2026-06-08
github actions: set `persist-credentials` to false and remove redundant git config unsetting steps

vitek-rostislav:

The monorepo-split workflows push the split package repositories using a dedicated push token (`ACTIONS_GIT_PUSH_TOKEN`). By default, `actions/checkout` persists the workflow `GITHUB_TOKEN` into the local git config as an auth header, which collides with that push token. Until now the workflows worked around this by scrubbing the persisted credentials from the git config at runtime before every push.

A previous attempt (#4643) hardened that workaround by adding `xargs -r` so the scrubbing step would not fail on empty input — but that was the wrong direction, fixing a symptom of the workaround instead of its cause.

This sets `persist-credentials: false` on the checkout step instead, so the conflicting credentials are never written in the first place. The runtime config-unsetting steps then become redundant and are removed, leaving the split and branch-cleanup workflows simpler and less fragile.

#### Fixes issues

N/A

----
  - https://rv-fix-monorepo-split-credentials.odin.shopsys.cloud
  - https://cz.rv-fix-monorepo-split-credentials.odin.shopsys.cloud
  - https://rv-fix-monorepo-split-credentials.odin.shopsys.cloud/sk

----
  - https://rv-fix-monorepo-split-credentials.odin.shopsys.cloud
  - https://cz.rv-fix-monorepo-split-credentials.odin.shopsys.cloud
  - https://rv-fix-monorepo-split-credentials.odin.shopsys.cloud/sk

### DIFF
--- a/.github/workflows/clean-up-after-branch-merge.yaml
+++ b/.github/workflows/clean-up-after-branch-merge.yaml
@@ -24,6 +24,7 @@ jobs:
                 with:
                     ref: 'refs/heads/${{ github.event.repository.default_branch }}'
                     path: cancel-review-after-branch-merge
+                    persist-credentials: false
             -   name: Cancel review
                 working-directory: cancel-review-after-branch-merge
                 run: |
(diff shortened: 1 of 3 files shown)


vitek-rostislav:
- Failing force split branch job on `20.0`: https://github.com/shopsys/shopsys/actions/runs/27131700633/job/80073989219
- Successful split with the fix applied: https://github.com/shopsys/shopsys/actions/runs/27131801024/job/80074322973

vitek-rostislav:
@coderabbitai review

TomasLudvik (approved):

pk16011990 (approved):

PR #4650, opened 2026-06-10
[SSP-4095] store filter improvement

RostislavKreisinger:

This PR improves store and pickup place search on the storefront.

Customers can now search stores by city or postcode even when the searched location does not directly match an existing store address. The search text can be resolved to coordinates through Google geocoding, so store lists can be ordered by distance from the searched location. Store lists also support pagination with loading more results, which keeps the initial response smaller and improves usability for larger store datasets.

The pickup place selection popup now uses the same searchable store list with map support, distance-based filtering, loading more results, and clearer selection behavior. In administration, store coordinates can be loaded automatically from the entered store address, reducing manual work when configuring stores.

#### Fixes issues
... 

----
  - https://rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud
  - https://cz.rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud
  - https://rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud/sk

----
  - https://rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud
  - https://cz.rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud
  - https://rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud/sk

----
  - https://rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud
  - https://cz.rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud
  - https://rk-ssp-4095-store-filter-improvement.odin.shopsys.cloud/sk


packages/framework/src/Form/Admin/Store/StoreFormType.php
```diff
@@ -180,7 +181,12 @@ private function createMapGroup(FormBuilderInterface $builder): FormBuilderInter
             ->add('longitude', NumberType::class, [
                 'required' => false,
                 'scale' => 10,
-            ]);
+            ])
+            ->add('loadCoordinates', ButtonType::class, [
+                'label' => 'Load Coordinates',
```
TomasLudvik: This is not consistent with labeling, it should be `Load coordinates`

packages/framework/src/Controller/Admin/StoreController.php
```diff
@@ -140,6 +141,26 @@ public function editAction(Request $request, int $id): Response
         ]);
     }
 
+    #[Route(
+        path: '/store/load-coordinates',
+        name: 'admin_store_loadcoordinates',
+        methods: ['post'],
+        condition: 'request.isXmlHttpRequest()',
+    )]
+    #[CanView]
```
TomasLudvik: #[CanView(methods: [HttpMethod::POST])]

packages/framework/assets/js/admin/components/StoreCoordinates.js
```diff
@@ -0,0 +1,47 @@
...
+            loaderElement: $button,
+            data: {
+                street: $form.find('.js-store-address-street').val(),
+                city: $form.find('.js-store-address-city').val(),
+                postcode: $form.find('.js-store-address-postcode').val(),
+                countryCode: $form.find('.js-store-address-country option:selected').data('country-code'),
+            },
+            success: data => {
+                if (data.latitude === undefined || data.longitude === undefined) {
+                    return;
```
TomasLudvik, line 38:
```suggestion
          new ModalWindow({
              content: Translator.trans('Coordinates could not be loaded for the entered address.'),
              style: 'danger'
          });
          return;
```

packages/framework/assets/js/admin/components/StoreCoordinates.js
```diff
@@ -0,0 +1,47 @@
...
+                }
+
+                $form.find('.js-store-coordinate-latitude').val(data.latitude);
+                $form.find('.js-store-coordinate-longitude').val(data.longitude);
+                FormChangeInfo.showInfo();
```
TomasLudvik, line 43:
```suggestion
                FormChangeInfo.showInfo();
                this.showSuccessState($button);
```

packages/framework/assets/js/admin/components/StoreCoordinates.js
```diff
@@ -0,0 +1,47 @@
+import Ajax from '../../common/utils/Ajax';
+import Register from '../../common/utils/Register';
+import FormChangeInfo from './FormChangeInfo';
```
TomasLudvik, line 6:
```suggestion
import FormChangeInfo from './FormChangeInfo';
import Check from 'icons/tabler/check.svg';
import Translator from 'bazinga-translator';
```

packages/framework/assets/js/admin/components/StoreCoordinates.js
```diff
@@ -0,0 +1,47 @@
...
+                    return;
+                }
+
+                $form.find('.js-store-coordinate-latitude').val(data.latitude);
+                $form.find('.js-store-coordinate-longitude').val(data.longitude);
+                FormChangeInfo.showInfo();
+            },
+        });
+    }
+
```
TomasLudvik, line 48:
```suggestion
  showSuccessState($button) {
      const originalHtml = $button.html();
      $button.removeClass('btn-primary').addClass('btn-success').html(`${Check} ${Translator.trans('Coordinates loaded')}`);

      setTimeout(() => {
          $button.removeClass('btn-success').addClass('btn-primary').html(originalHtml);
      }, 3000);
  }
```

packages/framework/assets/js/admin/components/StoreCoordinates.js
TomasLudvik: In this file I have made few suggestions, that will let user know what is happening

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Shopsys\FrameworkBundle\Component\AddressCoordinates;
+
+use Symfony\Contracts\HttpClient\Exception\DecodingExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class GoogleAddressCoordinatesFacade
+{
+    protected const string GEOCODE_ADDRESS_URL = 'https://geocode.googleapis.com/v4/geocode/address';
```
TomasLudvik, line 15: This should be set as env variable

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        try {
+            $response = $this->httpClient->request('GET', static::GEOCODE_ADDRESS_URL, [
+                'query' => [
+                    'key' => $this->googleMapApiKey,
+                    'address.addressLines' => $street,
+                    'address.locality' => $city,
+                    'address.regionCode' => strtoupper($countryCode),
+                    'address.postalCode' => $postcode,
```
TomasLudvik: Should there be any validation for postcode or google does not care?

packages/framework/src/Form/Admin/Store/StoreFormType.php
```diff
@@ -176,11 +180,25 @@ private function createMapGroup(FormBuilderInterface $builder): FormBuilderInter
...
             ])
             ->add('longitude', NumberType::class, [
                 'required' => false,
                 'scale' => 10,
-            ]);
+                'attr' => [
+                    'class' => 'js-store-coordinate-longitude',
+                ],
+            ])
+            ->add('loadCoordinates', ButtonType::class, [
```
TomasLudvik: This should be added only if the google map api key is set

packages/framework/src/Controller/Admin/StoreController.php
```diff
@@ -140,6 +143,35 @@ public function editAction(Request $request, int $id): Response
...
+        $addressCoordinatesData = $this->addressCoordinatesFacade->getCoordinatesByAddress(
+            $request->request->getString('street'),
+            $request->request->getString('city'),
+            $request->request->getString('countryCode'),
+            $request->request->getString('postcode'),
+        );
+
+        if ($addressCoordinatesData === null) {
+            return new JsonResponse([
+                'success' => false,
```
TomasLudvik: No need to send success - you are not using it anywhere, same for next return

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
TomasLudvik: I would expect some logging here, so we can find out why it is not functioning

packages/frontend-api/src/Model/Resolver/Store/StoresQuery.php
```diff
@@ -33,6 +35,12 @@ public function storesQuery(
         /** @var array{latitude: string, longitude: string}|null $coordinates */
         $coordinates = $argument->offsetGet('coordinates');
 
+        $searchCoordinates = $this->storeSearchTextCoordinatesProvider->getCoordinatesFromSearchText($searchText);
```
TomasLudvik: The https://developers.google.com/maps/documentation/geocoding/reference/rest/v4beta/geocode.address/geocodeAddress
does the trick, no need to parse it on our own, Google can do that for us.

TomasLudvik: @pk16011990 should we also add address autocompletion for search form on the storefront? I think it would improve this significantly

packages/frontend-api/src/Model/Store/StoreRepository.php
```diff
@@ -42,7 +42,7 @@ protected function getBasicFilteredQueryBuilder(
             ->andWhere('s.domainId = :domainId')
             ->setParameter('domainId', $domainId);
 
-        if ($storesFilterOptions->getSearchText() !== null) {
+        if ($storesFilterOptions->getSearchText() !== null && $storesFilterOptions->getCoordinates() === null) {
```
TomasLudvik, line 76: If user uses location in the browser and just sets search text like "Ostrava", this will not search and show all results, right?

project-base/app/config/packages/cache.yaml
```diff
@@ -8,6 +8,10 @@ framework:
                 adapter: cache.adapter.redis
                 provider: snc_redis.bestselling_products
                 default_lifetime: 43200
+            store_search_coordinates_cache:
+                adapter: cache.adapter.redis
+                provider: snc_redis.store_search_coordinates
+                default_lifetime: 43200
```
TomasLudvik: I wonder why lifetime of 14hours. The coordinates should never change right, so we do not need to delete them. Other thing might be that they will never be reused, so we are keeping them only if user reuses them in short time? So I would consider one hour or unlimited here.

project-base/app/.env
```diff
@@ -79,3 +79,5 @@ MAKER_PHP_CS_FIXER_CONFIG_PATH=vendor/shopsys/maker/config/maker-php-cs-fixer.co
 LOCAL_PATH_TO_PROJECT_ROOT=/var/www/html
 SYMFONY_IDE_URL_FORMAT=phpstorm://open?file=%f&line=%l
 TRACY_DEBUGGER_IDE_URL_FORMAT=phpstorm://%action/?file=%file&line=%line&search=%search&replace=%replace
+
+GOOGLE_MAP_API_KEY=
```
TomasLudvik, line 83: Please set some value for our review server, so this functionality can be tested

upgrade-notes/backend_20260610_092359.md
```diff
@@ -0,0 +1,5 @@
+#### [SSP-4095] store filter improvement ([#4650](https://github.com/shopsys/shopsys/pull/4650))
+
+- add `GOOGLE_MAP_API_KEY` to your application environment; keep it empty to disable backend Google coordinate lookup or set it to a Google API key that can use Geocoding API
+- add `store_search_coordinates_cache` cache pool and `snc_redis.store_search_coordinates` Redis client configuration when you use the default Frontend API store search coordinates provider
```
TomasLudvik: You can ommit these as they are covered by project-base-diff

Makefile
```diff
@@ -72,7 +72,7 @@ generate-schema-native: ## Generates GraphQL schema and frontend types (natively
 # ✅ Code Checks and Fixes (PHP and JS/TS)
 # ------------------------------------------------------------------------------
 
-check-fix: generate-schema php-checks php-translations storefront-checks storefront-translations storefront-styles-for-admin check-licenses generate-icons-for-styleguide ## Runs all code checks (backend & storefront) and attempts to fix issues
+check-fix: generate-schema php-checks php-translations storefront-checks storefront-translations storefront-styles-for-admin check-licenses generate-icons-for-styleguide storefront-styles-for-admin ## Runs all code checks (backend & storefront) and attempts to fix issues
```
TomasLudvik: This commit says it removes duplicate entry, but it does exactly opposite - it adds already existing target here

packages/frontend-api/src/Model/Store/StoreSearchTextCoordinatesProvider.php
```diff
@@ -112,7 +113,15 @@ protected function getCachedCoordinatesBySearchText(
     ): ?array {
         return $this->storeSearchCoordinatesCache->get(
             $this->getCacheId($searchText, $countryCode),
-            fn () => $this->formatCoordinatesData($coordinatesProvider()),
+            function (CacheItemInterface $cacheItem, bool &$save) use ($coordinatesProvider): ?array {
+                $coordinates = $this->formatCoordinatesData($coordinatesProvider());
```
TomasLudvik: This is nice, but we should take into account two things:
- cache null, because google returned not found
- do not cache null, because google returned error

project-base/storefront/public/locales/cs/common.json
```diff
@@ -523,7 +524,9 @@
     "secs": "sek",
     "Select from existing addresses": "Vybrat z existujících adres",
     "Select image": "Vybrat obrázek",
+    "Select store": "",
```
TomasLudvik: You are missing some translations here

TomasLudvik (changes requested):

packages/framework/src/Form/Admin/Store/StoreFormType.php
RostislavKreisinger: ok

packages/framework/src/Controller/Admin/StoreController.php
RostislavKreisinger: ok

packages/framework/assets/js/admin/components/StoreCoordinates.js
RostislavKreisinger: ok

packages/framework/assets/js/admin/components/StoreCoordinates.js
RostislavKreisinger: ok

packages/framework/assets/js/admin/components/StoreCoordinates.js
RostislavKreisinger: ok

packages/framework/assets/js/admin/components/StoreCoordinates.js
RostislavKreisinger: ok

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
RostislavKreisinger: i don't thing so. Ask yourself whether it is possible to use a different address in the context of this class.

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
RostislavKreisinger: input validation is the responsibility of StoreSearchTextCoordinatesProvider

packages/framework/src/Form/Admin/Store/StoreFormType.php
RostislavKreisinger: ok

packages/framework/src/Controller/Admin/StoreController.php
RostislavKreisinger: ok

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
RostislavKreisinger: ok, thanx.

packages/frontend-api/src/Model/Resolver/Store/StoresQuery.php
RostislavKreisinger: I know that Google provides processing of unstructured data, but I deliberately chose to work with structured data.

packages/frontend-api/src/Model/Store/StoreRepository.php
RostislavKreisinger: yes, filtering results is in this case disabled and is returned list sorted by distance, so Ostrava will be first item.

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
TomasLudvik: Definitely yes as there is version of API in the URL, so you might need to update to version 5 etc.  Also I think it is our practice to put URLs in the .env same as packetery for example.

project-base/app/config/packages/cache.yaml
RostislavKreisinger: 14 hours is really not enough and it's a mistake, but after discussing with PK and MG, we didn't want to have infinity so that Redis wouldn't bloat unnecessarily. I'll set it to a week, after a week of not being used, the cache will be released and infrequent hits will disappear, leaving only the frequently searched ones.

project-base/storefront/public/locales/cs/common.json
RostislavKreisinger: ok

packages/frontend-api/src/Model/Store/StoreRepository.php
TomasLudvik: Ok, what about text "Ostr". I think it is error to not use search text in case that users location from browser is used.

packages/frontend-api/src/Model/Resolver/Store/StoresQuery.php
pk16011990: A nice feature idea, thanks. However, not relevant for this PR.

TomasLudvik: Can I ask what is benefit of doing this on our own? I am affraid that current state is not sufficient, as it will not work outside of Czech republic and Slovakia. Also tomorrow someone might say we want to support streets in the search and we will have to update the code again. 

All of this can be easily covered by Google, so I think it is easier way for us, or am I missing something?

project-base/app/.env
RostislavKreisinger: in public .env file?? realy? Nooooo :-)

Makefile
RostislavKreisinger: ok remove duplicited target

packages/frontend-api/src/Model/Resolver/Store/StoresQuery.php
RostislavKreisinger: updated

packages/frontend-api/src/Model/Store/StoreRepository.php
RostislavKreisinger: If you don't have geolocation turned on and Google doesn't find any results for you, the filter is applied and the term Ostr finds Ostrava. If you have obtained coordinates in any way, the result is just sorted.

project-base/app/.env
TomasLudvik: I meant to the GitHub Action secrets ofc, but I have done it :)

packages/frontend-api/src/Model/Store/StoreSearchTextCoordinatesProvider.php
```diff
@@ -33,36 +30,9 @@ public function getCoordinatesFromSearchText(?string $searchText): ?array
             return null;
         }
 
-        $postcode = $this->getPostcode($normalizedSearchText);
         $defaultCountryCode = $this->getDefaultCountryCode();
```
TomasLudvik, line 34: This method is still limiting to Czech Republic and Slovakia. You could remove this completely right? Why do we need to add this?

TomasLudvik:
@coderabbitai review

packages/frontend-api/src/Model/Resolver/Store/StoresQuery.php
henzigo: I think a few issues could arise here.

1. If we call the API for every query, the query can be very easily abused and repeated calls could burn through quite a lot of money. There should therefore be some kind of rate limit.
2. If I find a match in the database (e.g. by city), is there any point in calling this API? I'd probably only call it if the database returns no data. What do you think?
3. In case the user has geolocation enabled, it shouldn't be overwritten anymore, right?

packages/frontend-api/src/Model/Store/StoreSearchTextCoordinatesProvider.php
```diff
@@ -0,0 +1,105 @@
...
+
+        $searchText = trim((string)preg_replace('/\s+/', ' ', $searchText));
+
+        return $searchText !== '' ? $searchText : null;
+    }
+
+    /**
+     * @return array{latitude: string, longitude: string}|null
+     */
+    protected function getCachedCoordinatesBySearchText(string $searchText, string $countryCode): ?array
```
henzigo, line 57: I don't think caching the result is the best idea, if only because $searchText can be anything. We could very quickly run out of space in Redis that way.

henzigo (commented):
Please look

upgrade-notes/backend_20260610_092359.md
RostislavKreisinger: ok

packages/frontend-api/src/Model/Store/StoreSearchTextCoordinatesProvider.php
RostislavKreisinger: this was removed

packages/frontend-api/src/Model/Resolver/Store/StoresQuery.php
RostislavKreisinger: ad 1. cache and in documentation are instructions about limits
ad 2. Preselecting a city to get its coordinates if found in the DB could replace the API call. I'll work that into it.
ad 3. why not? I'm in Prague but I want to find the closest store to Opava where my mother-in-law lives :-D

packages/frontend-api/src/Model/Store/StoreSearchTextCoordinatesProvider.php
RostislavKreisinger: We brainstormed this problem and agreed on the current solution. Subsequent tuning can be done using TTL.

packages/frontend-api/src/Model/Store/StoreSearchTextCoordinatesProvider.php
RostislavKreisinger: now is Country Code depend on first country by current domain from admin settings.

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
```diff
@@ -0,0 +1,145 @@
...
+    public function getCoordinatesByUnstructuredAddress(string $address): ?AddressCoordinatesData
+    {
+        if (!$this->isGoogleApiAvailable() || $address === '') {
+            return null;
+        }
+
+        return $this->getCoordinatesByQuery([
+            'key' => $this->googleMapApiKey,
+            'addressQuery' => $address,
+        ]);
```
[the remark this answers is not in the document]
RostislavKreisinger: ale three points are done

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * @param array<string, string> $query
+     */
```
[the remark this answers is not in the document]
RostislavKreisinger: done

packages/framework/src/Component/AddressCoordinates/GoogleAddressCoordinatesFacade.php
```diff
@@ -0,0 +1,145 @@
...
+    protected function getCoordinatesByQuery(array $query): ?AddressCoordinatesData
+    {
+        try {
+            $response = $this->httpClient->request('GET', static::GEOCODE_ADDRESS_URL, [
+                'query' => $query,
+                'headers' => [
+                    'X-Goog-FieldMask' => 'results.location',
+                ],
+            ]);
```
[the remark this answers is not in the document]
RostislavKreisinger: done

packages/frontend-api/src/Model/Resolver/Store/StoresQuery.php
henzigo: 1. Cache won't help you there. I can easily generate, for example, 1M hashes as $searchText, so every call will also call the Google API. You can of course limit the maximum number of calls, but then it will stop working for all.
3. Oh, sorry, I was a bit confused there, because when I entered a city, it kept showing the distance in km from me, which didn't make sense.

packages/frontend-api/src/Model/Store/StoreSearchTextCoordinatesProvider.php
henzigo: Yeah, I know that you brainstormed it, but I don't think this is the best way. I suggest implementing a rate limit so that the search query cannot be abused

TomasLudvik (approved):

project-base/storefront/components/Basic/GoogleMap/GoogleMap.tsx
```diff
@@ -184,6 +216,17 @@ export const GoogleMap: FC<GoogleMapProps> = ({
                         />
                     );
                 })}
+
+                {additionalMarker?.latitude !== null &&
```
chlebektomas: Suggestion: The `additionalMarker` validity check is repeated inline in JSX and then parsed inside the render. Please extract this into a single derived value above the return, e.g. `additionalMarkerCoordinates`, and render based on that. It would keep the JSX focused on rendering and make the nullability/parsing rules easier to read.

project-base/storefront/components/Blocks/StoreList/StoresWrapper.tsx
```diff
@@ -3,79 +3,205 @@ import { StoreList } from 'components/Blocks/StoreList/StoreList';
...
                 </div>
                 <div className="basis-1/2" data-tid={TIDs.stores_map}>
                     <div className="flex aspect-square rounded-xl bg-background-more p-5 lg:sticky lg:top-5">
                         <GoogleMap
-                            activeMarkerHandler={(uuid) => clickOnMarkerHandler(uuid)}
-                            markers={mappedStores}
-                            shouldCenterToUserCoordinates={debouncedSearchTextValue === ''}
-                            userCoordinates={userCoordinates}
+                            activeMarkerHandler={clickOnMarkerHandler}
+                            additionalMarker={
```
chlebektomas: Suggestion: The map props are doing several derived decisions inline (`additionalMarker`, `latitude`, `longitude`, `shouldCenterToUserCoordinates`). Please consider extracting these above the return, for example `additionalMapMarker`, `mapLatitude`, `mapLongitude`, and `shouldCenterMapToUserCoordinates`. The `GoogleMap` call would then read as a simple declaration of already-prepared map state.

project-base/storefront/components/Blocks/Popup/PickupPlacePopup.tsx
```diff
@@ -1,64 +1,202 @@
...
     const { pickupPlace } = useCurrentCart();
     const [selectedStoreUuid, setSelectedStoreUuid] = useState(pickupPlace?.identifier ?? '');
+    const [selectedPickupPlace, setSelectedPickupPlace] = useState<StoreOrPacketeryPoint | null>(pickupPlace ?? null);
+    const [searchTextValue, setSearchTextValue] = useState<string>('');
+    const defaultUserCoordinates = useSessionStore((s) => s.coordinates);
+    const [userCoordinates, setUserCoordinates] = useState<TypeCoordinates | null>(defaultUserCoordinates);
+    const [transportStores, setTransportStores] = useState<TypeListedStoreConnectionFragment | null>(null);
+    const [isLoadingMoreTransportStores, setIsLoadingMoreTransportStores] = useState(false);
     const updatePortalContent = useSessionStore((s) => s.updatePortalContent);
+    const debouncedSearchTextValue = useDebounce(searchTextValue, 700);
```
chlebektomas: Suggestion: The store page and pickup popup now duplicate the same search/debounce/query-key/load-more connection state. A small hook, for example `usePaginatedStoreConnection` or a pair of narrower hooks for search state and load-more behavior, would reduce the chance that the standalone stores page and checkout popup drift apart later.

project-base/storefront/components/Blocks/Popup/PickupPlacePopup.tsx
```diff
@@ -1,64 +1,202 @@
...
+    }, [transportStoresData?.transport?.stores]);
+
+    useEffect(() => {
+        if (selectedStoreUuid === '') {
+            setSelectedPickupPlace(null);
+
+            return;
+        }
+
+        const selectedStore = transportStores?.edges?.find(
```
chlebektomas: Suggestion: Selected store lookup is repeated here and in the select handler. Please extract a small helper like `findStoreByUuid(transportStores, uuid)` or derive the selected store once with `useMemo`, then reuse it in both places. It would make the selection state easier to follow.

project-base/storefront/components/Basic/GoogleMap/GoogleMap.tsx
```diff
@@ -66,7 +73,10 @@ export const GoogleMap: FC<GoogleMapProps> = ({
     const mapRef = useRef<any>(null);
     const [isGoogleApiLoaded, setIsGoogleApiLoaded] = useState(false);
 
-    const validMarkers = (markers?.filter((marker) => marker.latitude !== null && marker.longitude !== null) ??
+    const [{ data: mapStoresData }] = useMapStoresQuery({ pause: markers !== undefined });
+    const effectiveMarkers = markers ?? mapConnectionEdges<TypeMapStoreFragment>(mapStoresData?.stores.edges);
+
+    const validMarkers = (effectiveMarkers?.filter((marker) => marker.latitude !== null && marker.longitude !== null) ??
```
chlebektomas: Suggestion: `effectiveMarkers`, `validMarkers`, and `markersClusterConfig` are rebuilt on every render and `markersClusterConfig` is also used as a dependency for the map-centering effect. Please consider memoizing these derived marker collections so the map effect is driven by actual marker changes rather than incidental rerenders.

chlebektomas (commented):
Readability suggestions for the new store-search/map flow.

project-base/storefront/public/locales/en/common.json
```diff
@@ -601,6 +605,7 @@
     "Toggle blog category": "Toggle blog category",
     "Tomorrow": "Tomorrow",
     "Too many login attempts. Try again later.": "Too many login attempts. Try again later.",
+    "Too many store search attempts. Try again later.": "",
```
chlebektomas: Warning: The new `"Too many store search attempts. Try again later."` translation value is an empty string here, and the same empty value is also present in the Czech and Slovak storefront locale files. This error is mapped to a flash message in `errorMessageMapper.ts`, so `t(...)` can resolve to an empty string and the user may see an empty toast when the store-search rate limit is hit. Please fill the locale values, at least the English fallback plus Czech/Slovak translations. It would also be useful to add a small assertion using real locale resources, because the current mocked `t` tests would not catch this.

project-base/storefront/components/Blocks/Popup/PickupPlacePopup.tsx
```diff
@@ -1,64 +1,202 @@
...
     onChangePickupPlaceCallback: (transportUuid: string, selectedPickupPlace: StoreOrPacketeryPoint | null) => void;
 };
 
+const PICKUP_PLACE_POPUP_STORES_SCROLL_TARGET_ID = 'pickup-place-popup-stores-scroll';
+
 export const PickupPlacePopup: FC<PickupPlacePopupProps> = ({ transportUuid, onChangePickupPlaceCallback }) => {
     const { t } = useTranslation();
+    const client = useClient();
     const { pickupPlace } = useCurrentCart();
     const [selectedStoreUuid, setSelectedStoreUuid] = useState(pickupPlace?.identifier ?? '');
```
chlebektomas, line 43: Warning: The popup initializes `selectedStoreUuid` from the cart pickup place, but the rendered list below is just the current `TransportStoresQuery` page. I do not see logic that prepends the selected/last-order pickup place or asks the backend to return it first, so if the selected pickup place is not first in the current ordering, Jira's requirement that it is selected and displayed first is not met. Please either pass the selected pickup identifier into the query and return it first, or merge/prepend `pickupPlace` client-side with de-duping. A focused Cypress/component test should cover a preselected pickup place that is not first in the returned stores and assert it renders first and selected.

chlebektomas (commented):
Adding the two warning findings from the earlier review.

project-base/storefront/components/Basic/GoogleMap/GoogleMap.tsx
RostislavKreisinger: ok,

project-base/storefront/components/Blocks/StoreList/StoresWrapper.tsx
RostislavKreisinger: ok

project-base/storefront/components/Blocks/Popup/PickupPlacePopup.tsx
RostislavKreisinger: ok

project-base/storefront/components/Blocks/Popup/PickupPlacePopup.tsx
RostislavKreisinger: ok

project-base/storefront/public/locales/en/common.json
RostislavKreisinger: ok

project-base/storefront/components/Blocks/Popup/PickupPlacePopup.tsx
RostislavKreisinger: ok

project-base/storefront/components/Blocks/StoreList/usePaginatedStoreConnection.ts
```diff
@@ -0,0 +1,134 @@
...
+                searchText: debouncedSearchTextValue || null,
+                coordinates: userCoordinates,
+                first: STORE_LIST_PAGE_SIZE,
+                after: null,
+            }) as QueryVariables,
+        [additionalQueryVariables, debouncedSearchTextValue, userCoordinates],
+    );
+    const queryKey = JSON.stringify(queryVariables);
+    const queryKeyRef = useRef(queryKey);
+    const [{ data, fetching: isStoreConnectionFetching }] = useQuery<TData, QueryVariables>({
```
chlebektomas: Warning: The shared store-connection hook drops the `error` returned by `useQuery`, and the `loadMoreStores` path also ignores `storesResponse.error`. If the initial `/stores` or pickup-place query fails, the UI can stay blank with no user-facing feedback; if a later query fails, it can keep stale results while appearing idle. Please expose the query error from this hook and render an error/empty state in the callers, or clear/replace stale stores when a new query fails. A focused test should mock `StoresQuery`/`TransportStoresQuery` failures and assert that the user sees a meaningful error state instead of an empty list.

chlebektomas (commented):
One remaining storefront review note from the focused re-check.

project-base/storefront/components/Blocks/StoreList/usePaginatedStoreConnection.ts
RostislavKreisinger: @chlebektomas please check last fix

chlebektomas (approved):

TomasLudvik (approved):

PR #4652, opened 2026-06-10
Fix Luigi's Box secondary search results

grossmannmartin:

Search result pages now return Luigi's Box articles, categories, and brands from their own search requests instead of relying on product quicksearch results.

Previously, the search page used product search as the main Luigi's Box request and displayed secondary results from `quicksearch_types`. That could make articles, categories, or brands differ from the results returned by Luigi's Box for the same type directly, even though autocomplete showed the expected matches.

This keeps the storefront GraphQL search contract unchanged, but makes the displayed secondary search sections match Luigi's Box fulltext search behavior more reliably. The PR also removes a misleading Luigi's Box tracker ID example from the storefront `.env` file.

#### Fixes issues

N/A

----
  - https://mg-luigisbox-search-secondary.odin.shopsys.cloud
  - https://cz.mg-luigisbox-search-secondary.odin.shopsys.cloud
  - https://mg-luigisbox-search-secondary.odin.shopsys.cloud/sk

----
  - https://mg-luigisbox-search-secondary.odin.shopsys.cloud
  - https://cz.mg-luigisbox-search-secondary.odin.shopsys.cloud
  - https://mg-luigisbox-search-secondary.odin.shopsys.cloud/sk


grossmannmartin:
@coderabbitai review

upgrade-notes/backend_20260610_134623.md
```diff
@@ -0,0 +1,5 @@
+#### Separate Luigi's Box search requests ([#4652](https://github.com/shopsys/shopsys/pull/4652))
+
+- if your project extends `Shopsys\LuigisBoxBundle\Component\LuigisBox\LuigisBoxClient`, update overrides or calls of `addSearchSpecificParametersToUrl()` to remove the `$limitsByType` argument
+- replace usages of removed methods `Shopsys\LuigisBoxBundle\Component\LuigisBox\LuigisBoxClient::getQuicksearchTypesWithLimits()`, `Shopsys\LuigisBoxBundle\Component\LuigisBox\LuigisBoxClient::getMainType()`, and `Shopsys\LuigisBoxBundle\Component\LuigisBox\LuigisBoxClient::getMainTypeLimit()` with per-type Luigi's Box search request handling
+- if your project extends `Shopsys\LuigisBoxBundle\Model\Batch\LuigisBoxBatchLoader`, replace usages of removed property `$mainBatchLoadData` and removed methods `getMainBatchLoadData()` and `getMainType()` with handling based on each `Shopsys\LuigisBoxBundle\Model\Batch\LuigisBoxBatchLoadData` item separately
```
TomasLudvik, line 6: This note documents the symbols removed in the first commit, but the fixup adds new public/static API that downstream projects extending these classes will want documented too: `LuigisBoxClient::getDataForMultiple()` and the new accessors `LuigisBoxBatchLoader::getTotalByMetadataKey()` / `getFacetsByMetadataKey()`. Could we extend this note to mention them before squashing?

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -56,72 +69,273 @@ public static function getFacets(): array
...
 
+    /**
+     * @return array<int, array<string, mixed>>
+     */
+    public static function getFacetsByMetadataKey(string $metadataKey): array
+    {
+        return static::$facetsByMetadataKey[$metadataKey] ?? [];
+    }
+
+    public static function getMetadataKey(LuigisBoxBatchLoadData $luigisBoxBatchLoadData): string
```
TomasLudvik: The metadata-key machinery here (this hash, plus the parallel `$totalsByType`/`$totalsByMetadataKey` and `$facets`/`$facetsByMetadataKey` pairs and the `$updateTypeMetadata` flag below) is essentially working around the fact that totals/facets live in static mutable state — the collision it defends against (e.g. autocomplete-product + search-product in one batch) only exists because both write the same static slot.

Did you consider returning totals/facets together with the loaded result instead of via static side-channels? That would let us delete most of this.

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -56,72 +69,273 @@ public static function getFacets(): array
...
+
+        foreach ($luigisBoxBatchLoadData as $key => $batchLoadData) {
+            $mappedData[] = $this->mapDataByType(
+                $batchLoadData->getType(),
+                $batchLoadData->getLimit(),
+                $luigisBoxResultsByKey[$key],
+                $batchLoadData->getEndpoint(),
+                static::getMetadataKey($batchLoadData),
+                $batchLoadData->getEndpoint() === LuigisBoxEndpointEnum::SEARCH
+                    || array_key_exists($batchLoadData->getType(), $searchTypes) === false,
```
TomasLudvik: This flag (`endpoint === SEARCH || !array_key_exists(type, $searchTypes)`), combined with the early `return` on `$updateTypeMetadata === false` in `mapDataByType()`, is hard to follow — the intent (when a type is requested via both autocomplete and search, let the search result own the legacy `$totalsByType[type]`) isn't obvious from the boolean. Could we extract it into a named predicate, e.g. `shouldUpdateLegacyTotalForType(...)`, in line with our preference for self-descriptive code over comments?

packages/luigis-box/src/Model/Batch/LuigisBoxBatchLoader.php
```diff
@@ -56,72 +69,273 @@ public static function getFacets(): array
...
+            $luigisBoxBatchLoadData->getUserIdentifier(),
+            $luigisBoxBatchLoadData->getQuery(),
+        ]);
+    }
+
+    /**
+     * @param array<int, array<int|string, \Shopsys\LuigisBoxBundle\Model\Batch\LuigisBoxBatchLoadData>> $groups
+     * @param array<int, array<string, bool>> $typesByGroup
+     */
+    protected function getAutocompleteGroupIndex(
```
TomasLudvik, line 203: Minor: this re-derives each group's key via `getAutocompleteGroupKey(array_first($group))` on every lookup, making the grouping O(n^2) over autocomplete items. Batch sizes are tiny so it's not a real perf concern, but the group key is already known when the group is created — storing it alongside `$typesByGroup` would avoid the repeated recomputation and read a bit clearer.

TomasLudvik (commented):
A few comments from reviewing the search-request refactor (mainly the fixup commit).

TomasLudvik (approved):

pk16011990 (approved):

PR #4653, opened 2026-06-10
mcp: prevent reading hidden columns via subselects

vitek-rostislav:

The MCP SQL tool lets an AI assistant query the database while restricting which tables and columns it is allowed to read. The query validator enforced these restrictions on the top-level query, but did not validate columns referenced inside a derived table (a subquery in the `FROM` clause). As a result, wrapping a non-exposed column in a subquery bypassed the restriction entirely — for example, `SELECT x.password FROM (SELECT password FROM administrators) x` returned administrator password hashes that should never be reachable through MCP. Because the same gap also allowed reaching non-exposed tables, this was a data-exposure security issue.

This change makes the validator fully validate every derived table's inner query against the exposed schema before its columns become available to the outer query, so hidden columns and non-exposed tables are rejected no matter how deeply they are nested. The accompanying test suite is extended to lock down this and the related vectors that reference a hidden column from a different position — `WHERE`, `ORDER BY`, an allowed function argument, a `UNION` branch, a `CASE` expression, a CTE body, and an output alias — each of which now reliably returns a "column not exposed" error.

#### Fixes issues

N/A

----
  - https://rv-fix-sql-validator-subselect.odin.shopsys.cloud
  - https://cz.rv-fix-sql-validator-subselect.odin.shopsys.cloud
  - https://rv-fix-sql-validator-subselect.odin.shopsys.cloud/sk

----
  - https://rv-fix-sql-validator-subselect.odin.shopsys.cloud
  - https://cz.rv-fix-sql-validator-subselect.odin.shopsys.cloud
  - https://rv-fix-sql-validator-subselect.odin.shopsys.cloud/sk

### DIFF
--- a/packages/mcp/src/Component/Database/Query/SqlQueryValidator.php
+++ b/packages/mcp/src/Component/Database/Query/SqlQueryValidator.php
@@ -452,6 +452,11 @@ class SqlQueryValidator
             );
         }
 
+        if ($wrappedNodeTag === self::NODE_TAG_RANGE_SUBSELECT) {
+            // Range subselects are validated in collectRelationsFromRangeSubselect(), before their derived alias is visible.
+            return null;
+        }
+
         if ($wrappedNodeTag === self::NODE_TAG_FUNC_CALL && !$this->isAllowedFunctionCall($wrappedNode)) {
             return self::ERROR_UNSUPPORTED_READ_WRITE_CONSTRUCT;
         }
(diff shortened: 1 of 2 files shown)


grossmannmartin (approved):
Maybe we could add multiple subqueries to test. Like:

```
SELECT
    (SELECT username FROM administrators LIMIT 1) AS visible_username,
    (SELECT password FROM administrators LIMIT 1) AS hidden_password
LIMIT 1;
```

and

```
SELECT x.username
FROM (
    SELECT username
    FROM (SELECT username, password FROM administrators) y
) x
LIMIT 1;
```

pk16011990 (approved):

PR #4656, opened 2026-06-15
Limit GraphQL batch request operations

grossmannmartin:

This PR limits how many operations can be sent in a single GraphQL batch request to the Frontend API.

Without this limit, clients could send oversized batch payloads that would all be passed to GraphQL execution at once. The new behavior rejects batch requests above the configured limit with a clear `400 Bad Request` response, while requests within the limit continue to work normally.

The default limit is 20 operations and can be adjusted through configuration when a project needs a different threshold.

#### Fixes issues
N/A

----
  - https://mg-limit-batch-query.odin.shopsys.cloud
  - https://cz.mg-limit-batch-query.odin.shopsys.cloud
  - https://mg-limit-batch-query.odin.shopsys.cloud/sk

----
  - https://mg-limit-batch-query.odin.shopsys.cloud
  - https://cz.mg-limit-batch-query.odin.shopsys.cloud
  - https://mg-limit-batch-query.odin.shopsys.cloud/sk

### DIFF
--- a/packages/frontend-api/src/Resources/config/parameters.yaml
+++ b/packages/frontend-api/src/Resources/config/parameters.yaml
@@ -1,2 +1,3 @@
 parameters:
+    shopsys.frontend_api.max_batch_operations: 20
     shopsys.frontend_api.validation_logged_as_error: false
--- a/packages/frontend-api/src/Component/HttpFoundation/GraphqlBatchRequestValidator.php
+++ b/packages/frontend-api/src/Component/HttpFoundation/GraphqlBatchRequestValidator.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\FrontendApiBundle\Component\HttpFoundation;
+
+use JsonException;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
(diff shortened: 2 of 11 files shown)


packages/frontend-api/src/Resources/config/parameters.yaml
```diff
@@ -1,2 +1,3 @@
 parameters:
+    shopsys.frontend_api.max_batch_operations: 20
```
vitek-rostislav, line 2: I wonder, maybe we could link this setting from the security section in `project-base/app/config/packages/overblog_graphql.yaml` - some comment that would mention something like "if you need to adjust the max batch size, you can set the `shopsys.frontend_api.max_batch_operations`custom  parameter in your `app/config/parameters_common.yaml` as it is not supported natively in the overblog package". Something like that. Because honestly, that was the first config I tried to find this setting in :)

vitek-rostislav (approved):
Just one idea, otherwise looks good to me, reviewed and tested, nice job :ok_hand:

pk16011990 (approved):

packages/frontend-api/src/Resources/config/parameters.yaml
grossmannmartin: fixed

PR #4660, opened 2026-06-18
nginx: uploaded files now download as attachments

grossmannmartin:

Uploaded files served from `/content/uploadedFiles/` are now delivered as downloadable attachments instead of being rendered directly in the browser.

This prevents uploaded HTML or SVG files from executing active content when someone opens the raw uploaded-file URL. Legitimate file access remains available, but browsers are instructed to download these files consistently in both Docker and Kubernetes nginx configurations.

#### Fixes issues

N/A

----
  - https://mg-svg-attachment.odin.shopsys.cloud
  - https://cz.mg-svg-attachment.odin.shopsys.cloud
  - https://mg-svg-attachment.odin.shopsys.cloud/sk

----
  - https://mg-svg-attachment.odin.shopsys.cloud
  - https://cz.mg-svg-attachment.odin.shopsys.cloud
  - https://mg-svg-attachment.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/FileUpload/FileUpload.php
+++ b/packages/framework/src/Component/FileUpload/FileUpload.php
@@ -5,15 +5,17 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Component\FileUpload;
 
 use InvalidArgumentException;
+use League\Flysystem\Config;
 use League\Flysystem\FilesystemException;
 use League\Flysystem\FilesystemOperator;
 use League\Flysystem\MountManager;
 use League\Flysystem\StorageAttributes;
+use League\Flysystem\Visibility;
 use Shopsys\FrameworkBundle\Component\Cache\InMemoryCache;
 use Shopsys\FrameworkBundle\Component\CustomerUploadedFile\CustomerUploadedFile;
 use Shopsys\FrameworkBundle\Component\CustomerUploadedFile\CustomerUploadedFileRepository;
 use Shopsys\FrameworkBundle\Component\Doctrine\Exception\UnexpectedTypeException;
-use Shopsys\FrameworkBundle\Component\FileUpload\Exception\MissingFileClassDirectoryMappingException;
+use Shopsys\FrameworkBundle\Component\FileUpload\Exception\MissingFileConfigByFileClassException;
 use Shopsys\FrameworkBundle\Component\FileUpload\Exception\MoveToEntityFailedException;
 use Shopsys\FrameworkBundle\Component\FileUpload\Exception\UploadFailedException;
 use Shopsys\FrameworkBundle\Component\Image\Image;
@@ -30,9 +32,15 @@ class FileUpload
     protected const int DELETE_OLD_FILES_SECONDS = 86400;
     protected const string POSITION_BY_ENTITY_AND_TYPE_CACHE_NAMESPACE = 'positionByEntityAndType';
 
+    /**
+     * @param array<class-string<\Shopsys\FrameworkBundle\Component\FileUpload\EntityFileUploadInterface>, array{
+     *      directory: non-empty-string,
+     *      visibility: 'public'|'private',
+     *  }> $filesConfigByFileClass
+     */
     public function __construct(
         protected readonly string $temporaryDir,
-        protected array $directoriesByFileClass,
+        protected array $filesConfigByFileClass,
         protected readonly FileNamingConvention $fileNamingConvention,
         protected readonly MountManager $mountManager,
         protected readonly FilesystemOperator $filesystem,
@@ -54,6 +62,9 @@ class FileUpload
         $this->mountManager->move(
             'local://' . $file->getRealPath(),
             'main://' . $this->getTemporaryDirectory() . '/' . $temporaryFilename,
+            [
+                Config::OPTION_VISIBILITY => Visibility::PRIVATE,
+            ],
         );
 
         return $temporaryFilename;
@@ -174,7 +185,13 @@ class FileUpload
                     $this->filesystem->delete($targetFilename);
                 }
 
-                $this->mountManager->move('main://' . $sourceFilepath, 'main://' . $targetFilename);
+                $this->mountManager->move(
+                    'main://' . $sourceFilepath,
+                    'main://' . $targetFilename,
+                    [
+                        Config::OPTION_VISIBILITY => $this->getVisibilityByFileClass($fileForUpload->getFileClass()),
+                    ],
+                );
                 $entity->setFileKeyAsUploaded($key);
             } catch (IOException $ex) {
                 $message = 'Failed to rename file from temporary directory to entity';
@@ -266,20 +283,36 @@ class FileUpload
         return $uploadEntityType;
     }
 
-    protected function getDirectoryByFileClass(string $fileClass): string
+    /**
+     * @return array{
+     *      directory: non-empty-string,
+     *      visibility: 'public'|'private',
+     *  }
+     */
+    protected function getFileConfigByFileClass(string $fileClass): array
     {
-        if (array_key_exists($fileClass, $this->directoriesByFileClass)) {
-            return $this->directoriesByFileClass[$fileClass];
+        if (array_key_exists($fileClass, $this->filesConfigByFileClass)) {
+            return $this->filesConfigByFileClass[$fileClass];
         }
 
-        foreach ($this->directoriesByFileClass as $class => $dir) {
+        foreach ($this->filesConfigByFileClass as $class => $config) {
             if (is_subclass_of($fileClass, $class)) {
-                return $dir;
+                return $config;
             }
         }
 
-        throw new MissingFileClassDirectoryMappingException(
-            sprintf('Missing directory mapping for file class "%s"', $fileClass),
+        throw new MissingFileConfigByFileClassException(
+            sprintf('Missing file config mapping for file class "%s"', $fileClass),
         );
+    }
+
+    protected function getDirectoryByFileClass(string $fileClass): string
+    {
+        return $this->getFileConfigByFileClass($fileClass)['directory'];
+    }
+
+    protected function getVisibilityByFileClass(string $fileClass): string
+    {
+        return $this->getFileConfigByFileClass($fileClass)['visibility'];
     }
 }
--- a/packages/framework/src/Resources/config/services.yaml
+++ b/packages/framework/src/Resources/config/services.yaml
@@ -297,10 +297,16 @@ services:
     Shopsys\FrameworkBundle\Component\FileUpload\FileUpload:
         arguments:
             $temporaryDir: '%shopsys.var_dir%'
-            $directoriesByFileClass:
-                Shopsys\FrameworkBundle\Component\Image\Image: '%shopsys.image_dir%'
-                Shopsys\FrameworkBundle\Component\UploadedFile\UploadedFile: '%shopsys.uploaded_file_dir%'
-                Shopsys\FrameworkBundle\Component\CustomerUploadedFile\CustomerUploadedFile: '%shopsys.customer_uploaded_file_dir%'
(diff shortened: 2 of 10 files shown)


project-base/docker/nginx/nginx.conf
```diff
@@ -25,6 +25,11 @@ map $request_uri $static_cors {
     ~^/_next/static/ "*";
 }
 
+map $request_uri $uploaded_file_content_disposition {
```
TomasLudvik: Consider keying this map on `$uri` rather than `$request_uri`.

`$request_uri` is the raw request line — slashes aren't merged and `%XX` isn't decoded — whereas the location that actually serves the file matches against the normalized, decoded `$uri`. That gap is bypassable in the fail-**open** direction:

- `//content/uploadedFiles/x.svg` → `merge_slashes` (on by default) collapses this for location matching so the file is served, but `$request_uri` stays `//content/...`, this regex doesn't match → no `Content-Disposition` → the SVG renders inline.
- `/content/uploaded%46iles/x.svg` → decoded `$uri` resolves to the real file and serves it, but `$request_uri` still contains `%46` and won't match the literal `uploadedFiles`.

Both re-open exactly the inline-rendering vector this PR closes. `$uri` (decoded + slash-merged) avoids both. The existing `$static_cors` map uses `$request_uri`, but there a miss fails *closed* (header omitted = safer); here a miss fails *open*, so the variable choice matters more.

TomasLudvik (commented):
Reviewed the nginx hardening — solid, correctly-placed fix that closes the direct-static-access gap left by `UploadedFileController::viewAction()`. Four inline notes below: one substantive (`$request_uri` vs `$uri`, bypassable in the fail-open direction), one coverage gap (SVG via the image system stays inline-renderable, so the PR's stated goal is only partly met), and a minor K8s duplicate-header note. Details inline.

project-base/app/orchestration/kubernetes/configmap/nginx.yaml
```diff
@@ -73,6 +73,11 @@ data:
             ~^/_next/static/ "*";
         }
 
+        map $request_uri $uploaded_file_content_disposition {
```
TomasLudvik: Same point as the Docker `nginx.conf` map: consider `$uri` instead of `$request_uri`. `$request_uri` is raw (slashes not merged, `%XX` not decoded) while the `content` location serves against the normalized `$uri`, so `//content/uploadedFiles/x.svg` or `/content/uploaded%46iles/x.svg` would still be served but skip the `attachment` disposition. See the detailed note on the Docker config for the full reasoning.

project-base/app/orchestration/kubernetes/configmap/nginx.yaml
```diff
@@ -139,6 +144,7 @@ data:
             add_header X-Content-Type-Options "nosniff" always;
             add_header X-XSS-Protection "0" always;
             add_header Permissions-Policy "camera=(), geolocation=(self), microphone=(), payment=(), usb=()" always;
+            add_header Content-Disposition $uploaded_file_content_disposition always;
```
TomasLudvik, line 147: Minor, K8s-specific: in this config the `content` location proxies to S3 (`proxy_pass`), and `add_header` *appends* rather than replaces. If an S3 object ever carries its own `Content-Disposition`, the client receives two values, which browsers handle inconsistently. Adding `proxy_hide_header Content-Disposition;` in the `location ~ ^/(content)/` block would make the outcome deterministic. Low risk given the current bucket setup — just flagging it.

project-base/docker/nginx/nginx.conf
```diff
@@ -25,6 +25,11 @@ map $request_uri $static_cors {
...
+    default "";
+    ~^/(?:[^/]+/)?content(?:-test)?/uploadedFiles/ "attachment";
```
TomasLudvik, line 30: This scopes forced-download to `uploadedFiles/` only, but SVG is an allowed upload type that lands under `content/images/` — so the active-content vector this PR targets isn't fully closed.

- Parameter-value icons explicitly accept `image/svg+xml` (`ParameterValueFormType.php:40`, *"You can upload a PNG, JPG, GIF, or SVG file"*), and `ImageProcessor::$supportedImageExtensions` includes `svg`.
- The image location only matches `jpe?g|png|gif`, so a stored `.svg` (e.g. `/content/images/parameterValue/….svg`) falls through to the generic static/S3 location and is served raw as `image/svg+xml` → scripts run on direct navigation. Reachable by any admin or a CSRF'd/hijacked admin session, so "admin-managed" doesn't neutralize it.
- `wysiwyg/` is lower-risk (elFinder's `upload_allow` excludes svg/html) but that's a MIME filter and bypassable — same defense-in-depth rationale.

Suggestion: force download for active-content extensions (`svg`, `svgz`, `html`, `htm`, `xml`, `xhtml`, …) across the whole `content/` tree, not just `uploadedFiles/`. `Content-Disposition: attachment` is ignored for subresource loads (``), so legitimately-embedded images keep displaying — it only forces download on direct top-level navigation (worth a quick verify in your target browsers). At minimum this is worth an explicit scope decision rather than leaving `images/`/`wysiwyg/` implicitly out.

project-base/docker/nginx/nginx.conf
grossmannmartin: After a careful re-evaluation of the URI normalization semantics involved here, including the distinction between the raw request target and the path representation Nginx uses after decoding and slash normalization for location/file resolution, I agree that the previous implementation left an avoidable mismatch in the security decision boundary.

The uploaded-file Content-Disposition map has therefore been updated to use the normalized $uri value instead of $request_uri, so the attachment header is now decided from the same effective path shape that Nginx uses when serving the file. This closes the bypass cases involving duplicate slashes and percent-encoded path segments.

In other words: fixed in 07857856a6.

project-base/app/orchestration/kubernetes/configmap/nginx.yaml
grossmannmartin: fixed in  07857856a6

project-base/app/orchestration/kubernetes/configmap/nginx.yaml
grossmannmartin: This is a fair point for the Kubernetes path, where uploaded content is served through the S3 proxy rather than directly from the local filesystem. Since add_header does not replace an upstream Content-Disposition, an object-level header from S3 could otherwise be forwarded alongside the Nginx-added one, leaving the final browser behavior dependent on duplicate-header handling.

I’ve made the proxy behavior deterministic by hiding any upstream Content-Disposition from S3 in the relevant content location, so the response now consistently uses the disposition controlled by our Nginx config.

Fixed in f0afbd6198.

project-base/docker/nginx/nginx.conf
grossmannmartin: I think the premise here is not correct for parameter-value icons.

ParameterValueFormType does allow SVG, but it uses FileUploadType, not ImageUploadType. On save, ParameterFacade::editParameterValue() passes colorIcon to UploadedFileFacade::manageFiles(), so these files are stored as UploadedFile records under the uploaded-file storage, whose configured path is /web/content/uploadedFiles/.

So a parameter-value SVG should not land under /content/images/parameterValue/...; it should be covered by the existing content/uploadedFiles/ rule.

Image upload does not allow SVG files at all.

packages/framework/src/Resources/config/services.yaml
```diff
@@ -293,10 +293,16 @@ services:
...
+            $filesConfigByFileClass:
+                Shopsys\FrameworkBundle\Component\Image\Image:
+                    directory: '%shopsys.image_dir%'
+                    visibility: !php/const League\Flysystem\Visibility::PUBLIC
+                Shopsys\FrameworkBundle\Component\UploadedFile\UploadedFile:
+                    directory: '%shopsys.uploaded_file_dir%'
+                    visibility: !php/const League\Flysystem\Visibility::PUBLIC
+                Shopsys\FrameworkBundle\Component\CustomerUploadedFile\CustomerUploadedFile:
+                    directory: '%shopsys.customer_uploaded_file_dir%'
+                    visibility: !php/const League\Flysystem\Visibility::PRIVATE
```
TomasLudvik, line 309: Should we add some post deploy task for this, so it is correctly updated on all projects?

packages/framework/src/Component/FileUpload/FileUpload.php
```diff
@@ -266,20 +279,36 @@ protected function getUploadEntityType(EntityFileUploadInterface $entity): strin
...
 
-        foreach ($this->directoriesByFileClass as $class => $dir) {
+        foreach ($this->filesConfigByFileClass as $class => $config) {
             if (is_subclass_of($fileClass, $class)) {
-                return $dir;
+                return $config;
             }
         }
 
         throw new MissingFileClassDirectoryMappingException(
```
TomasLudvik: This Exception should be renamed

packages/framework/src/Resources/config/services.yaml
```diff
@@ -293,10 +293,16 @@ services:
...
         arguments:
             $temporaryDir: '%shopsys.var_dir%'
-            $directoriesByFileClass:
-                Shopsys\FrameworkBundle\Component\Image\Image: '%shopsys.image_dir%'
-                Shopsys\FrameworkBundle\Component\UploadedFile\UploadedFile: '%shopsys.uploaded_file_dir%'
-                Shopsys\FrameworkBundle\Component\CustomerUploadedFile\CustomerUploadedFile: '%shopsys.customer_uploaded_file_dir%'
+            $filesConfigByFileClass:
+                Shopsys\FrameworkBundle\Component\Image\Image:
+                    directory: '%shopsys.image_dir%'
+                    visibility: !php/const League\Flysystem\Visibility::PUBLIC
```
TomasLudvik, line 303: Ops consideration for plain AWS S3. Explicit Visibility::PUBLIC makes the S3 adapter send ACL: public-read; buckets with ACLs disabled (the modern AWS default, BucketOwnerEnforced) reject such writes. There's precedent in the platform (DirectoryStructureCreator already sends ACLs), so managed deployments are presumably fine, but image/file uploads on ACL-disabled buckets will start failing where previously no ACL was sent. Might deserve a line in the upgrade note.

packages/framework/src/Component/FileUpload/FileUpload.php
TomasLudvik, line 62: Should we add private visibility here, to ensure everything is correct?

TomasLudvik (changes requested):

packages/framework/src/Component/FileUpload/FileUpload.php
grossmannmartin: renamed to `MissingFileConfigByFileClassException` in a484236f9b

packages/framework/src/Resources/config/services.yaml
grossmannmartin: The ACL-disabled S3 concern is real for plain AWS S3, but it is not newly caused by this PR.
The default Shopsys S3 bridge already uses public visibility for S3 writes. 
What this PR changes is that final uploaded files now explicitly set per-file-class visibility, instead of relying on retained/default visibility during the move.

packages/framework/src/Component/FileUpload/FileUpload.php
grossmannmartin: It seems reasonable - 1b2bdceb88

packages/framework/src/Resources/config/services.yaml
grossmannmartin: good idea! added in 9018b210ad

TomasLudvik: It is good, but we should also consider, that projects might have added some on their owns, so at least upgrade note should mention that they might want to use this task for theirs files as well

TomasLudvik (approved):

PR #4665, opened 2026-06-18
Count nested fields in GraphQL pagination complexity

grossmannmartin:

GraphQL pagination complexity now reflects the real cost of paginated Frontend API queries. The complexity guard includes the selected fields for each returned item and counts connection metadata, such as `totalCount` and `pageInfo`, only once. This prevents expensive nested paginated queries from being undercounted without artificially multiplying metadata fields by page size.

Standard connection fields no longer need `dynamicPaginationComplexity`. Projects can still define a custom whole-field complexity override when they need a specific cost model.

Storefront GraphQL queries were narrowed to request only data used by the rendered views. Complaint lists, complaint order-item selection, category bestsellers, product listing cards, GTM mapping, and related generated types now use smaller fragments so customer-facing flows stay within the complexity limit.

Upgrade notes are included for projects with custom Frontend API or storefront GraphQL queries, because custom paginated selections may need the same review after nested fields are counted accurately.

#### Fixes issues
N/A

----
  - https://mg-graphql-complexity-fix.odin.shopsys.cloud
  - https://cz.mg-graphql-complexity-fix.odin.shopsys.cloud
  - https://mg-graphql-complexity-fix.odin.shopsys.cloud/sk

----
  - https://mg-graphql-complexity-fix.odin.shopsys.cloud
  - https://cz.mg-graphql-complexity-fix.odin.shopsys.cloud
  - https://mg-graphql-complexity-fix.odin.shopsys.cloud/sk


TomasLudvik (approved):

project-base/app/config/packages/overblog_graphql.yaml
```diff
@@ -24,7 +24,7 @@ overblog_graphql:
     security:
         enable_introspection: '%kernel.debug%'
         handle_cors: false
-        query_max_complexity: 1110
+        query_max_complexity: 2000
```
chlebektomas: This change raises `query_max_complexity` from `1110` to `2000`, but SSP-4140 explicitly requires the existing limit to stay unchanged.

It also makes the Jira reproduction still pass on the PR preview: `products(first: 1)` with about 1200 aliases under `edges.node.name` returns data instead of a max-complexity validation error. With the original limit, this should be rejected.

Please keep the original `1110` limit and reduce/split the storefront queries that exceed it.

packages/frontend-api/tests/Unit/Component/ExpressionLanguage/ComplexityCalculatorTest.php
```diff
@@ -0,0 +1,98 @@
...
+        yield 'zero children complexity keeps original item calculation' => [
+            'arguments' => ['first' => 3],
+            'oneItemComplexity' => 2,
+            'defaultCount' => 10,
+            'childrenComplexity' => 0,
+            'expectedComplexity' => 6,
+        ];
+    }
+
+    public function testDynamicPaginationComplexityCompilerPassesChildrenComplexity(): void
```
chlebektomas: The unit test covers the calculator arithmetic and compiler string, but it does not protect the actual GraphQL validation behavior from SSP-4140.

Please add an API/GraphQL test that executes a paginated query with a wide nested selection, for example `products(first: 1)` with many aliases under `edges.node.name`, and asserts that it is rejected by the configured max-complexity validator.

chlebektomas (changes requested):

packages/frontend-api/tests/Unit/Component/ExpressionLanguage/ComplexityCalculatorTest.php
grossmannmartin: Good idea. added

project-base/app/config/packages/overblog_graphql.yaml
grossmannmartin: Yeah, I had to increase the complexity eventually, as with the newly counted children's complexity, the number was simply too low. 
I have optimized some queries, so overall the performance should be better.
We can (and probably should?) invest some more time to optimize them more and eventually lower the maximum complexity threshold.

chlebektomas (approved):

pk16011990 (approved):

PR #4667, opened 2026-06-20
Add make targets to switch the whole application between dev and production

vitek-rostislav:

Switching the whole application between development and production used to be a two-part, partly manual chore: on the backend we have `phing environment-change`, but the storefront had to be switched by hand — editing its run command in `docker-compose.yml` from `dev` to `build` and recreating the container.

This PR adds two make targets that handle both sides in a single command (available both in the monorepo `Makefile` and in `project-base/Makefile`):

- `make environment-prod` — switches the backend to `prod` (via `phing environment-change`) and recreates the storefront container running `build` (production build: `next build` + `next start`)
- `make environment-dev` — switches everything back to development (storefront running `next dev`)

How the storefront switch works: the storefront run command is now driven by a `STOREFRONT_RUN_COMMAND` environment variable in the (gitignored, generated) `docker-compose.yml`, read by the storefront entrypoint (defaulting to `dev`). The make target rewrites that value between `dev` and `build` and recreates the container to apply it. A self-documenting comment is added to all `docker-compose*.yml.dist` templates so freshly generated compose files work out of the box.

> Note: after `make environment-prod` the container comes up while `next build` runs in the background, so the storefront is served only once the build finishes (`docker compose logs -f storefront`).

----
  - https://rv-switch-app-env.odin.shopsys.cloud
  - https://cz.rv-switch-app-env.odin.shopsys.cloud
  - https://rv-switch-app-env.odin.shopsys.cloud/sk

----
  - https://rv-switch-app-env.odin.shopsys.cloud
  - https://cz.rv-switch-app-env.odin.shopsys.cloud
  - https://rv-switch-app-env.odin.shopsys.cloud/sk

### DIFF
--- a/docker/conf/docker-compose-mac.yml.dist
+++ b/docker/conf/docker-compose-mac.yml.dist
@@ -48,8 +48,9 @@ services:
             dockerfile: ./docker/dev.Dockerfile
             target: development
         container_name: shopsys-framework-storefront
-        command: # Optional
-            - dev # change to `- build` if you want to test application build
+        environment:
+            # toggled by `make environment-dev` / `make environment-prod`; recreate the container to apply (dev = dev server, build = production build)
+            - STOREFRONT_RUN_COMMAND=dev
         volumes:
             - storefront:/home/node/app
         ports:
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ help: ## Displays list of available commands
 	run-specific-test-regression run-specific-test-base \
 	open-acceptance-tests-base open-acceptance-tests-regression run-smoke-tests \
 	generate-snapshots-info-table prepare-data-for-acceptance-tests cypress-prepare cypress-cleanup \
-	check-licenses
+	check-licenses environment-prod environment-dev
 
 # ------------------------------------------------------------------------------
 # 🐳 Docker Compose (macOS with Mutagen)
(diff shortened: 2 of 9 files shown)


docker/conf/docker-compose-mac.yml.dist
```diff
@@ -48,8 +48,8 @@ services:
             dockerfile: ./docker/dev.Dockerfile
             target: development
         container_name: shopsys-framework-storefront
-        command: # Optional
-            - dev # change to `- build` if you want to test application build
+        command:
+            - dev # managed by make environment-dev / make environment-prod (dev = dev server, build = production build)
```
grossmannmartin: I have a bad feeling about it... It's really easy to break the automation as the proper line is found by the specific text 🤔

Also, I think the `sed` in the makefile returns success even when it replaces nothing, so it may be weird

grossmannmartin (changes requested):

docker/conf/docker-compose-mac.yml.dist
vitek-rostislav: ok, what about this approach? c5c565b9ae

vitek-rostislav: and once more and better - 0b6933daa3

grossmannmartin (approved):

grossmannmartin (approved):

pk16011990 (approved):

PR #4662, opened 2026-06-18
Add grouped transport selection to checkout

chlebektomas:
#### Description, the reason for the PR

This PR introduces transport groups that can be managed in administration and exposed through the Storefront API. The checkout transport step now presents related transport methods in grouped cards, making delivery-to-address and pickup-point options easier to compare and navigate with both mouse and keyboard.

The demo data was updated with transport, payment, and transport group images and a more representative ordering, so newly built demo projects show the improved checkout selection out of the box.

#### Fixes issues

N/A

#### License Agreement for contributions

- [x] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://tc-ssp-4082-transport-grouping.odin.shopsys.cloud
  - https://cz.tc-ssp-4082-transport-grouping.odin.shopsys.cloud
  - https://tc-ssp-4082-transport-grouping.odin.shopsys.cloud/sk

----
:globe_with_meridians: Live Preview:
  - https://tc-ssp-4082-transport-grouping.odin.shopsys.cloud
  - https://cz.tc-ssp-4082-transport-grouping.odin.shopsys.cloud
  - https://tc-ssp-4082-transport-grouping.odin.shopsys.cloud/sk

PR #4682, opened 2026-06-29
Add blog article authors and unify the admin single-value picker

TomasLudvik:

This PR introduces blog article authors so editorial content can be attributed to a person.

Administrators can manage a codebook of authors in the admin (name, plus optional photo, job title, and description) and assign one optional author to a blog article. On the storefront the author then appears on the article detail — a small chip with the photo and name in the header, and an "Author of the article" box below the content that omits any fields left blank — and the author is added to the page's `Article` structured data (JSON-LD) for search engines. The author's admin edit page lists the articles written by that author in a paginated grid.

The author is assigned using the same "open a window and pick from a list" experience that products and files already use. While wiring this up, the previously copy-pasted single-value pickers (product and single file) were consolidated onto one reusable base (`SinglePicker`), mirroring the existing `MultiplePicker`. This removes duplicated picker code, so future single-entity pickers need only a thin form type instead of a full copy; product selection (including adding products to an order) and file selection now go through this shared mechanism.

This PR introduces BC breaks (the picker consolidation); UPGRADE notes are included.

#### Fixes issues

N/A

----
  - https://tl-blog-authors.odin.shopsys.cloud
  - https://cz.tl-blog-authors.odin.shopsys.cloud
  - https://tl-blog-authors.odin.shopsys.cloud/sk


packages/administration/templates/form/type/SinglePickerType.html.twig
```diff
@@ -0,0 +1,59 @@
+{% block single_picker_row %}
+    <div class="mb-3 row">
+        {% if inline_label %}
```
grossmannmartin: I would remove this option entirely and let the picker label render the same way as other form fields are. 

Now it creates an unusual labeling

Image

packages/administration/templates/form/type/SinglePickerType.html.twig
```diff
@@ -0,0 +1,59 @@
...
+        {% endif %}
+    </div>
+{% endblock %}
+
+{% block single_picker_widget %}
+    {% if attr %}
+        <div {{ block('attributes') }}>
+    {% endif %}
+
+    <div class="input-group js-single-picker"
```
grossmannmartin: For a new component, I would rather use `data-js-...` attribute instead of `js-...` class. 

- It's slightly more readable as it's not mixing styling with behavior
- It won't be accidentally used for styling
- When added in row_attr on form type, it's not necessary to copy-paste styling classes (for example, in AdvertFormType)
- Lesser risk of accidental removal of the class (for example, when modyfing classes with JS)

packages/framework/assets/js/admin/components/SinglePicker.js
```diff
@@ -0,0 +1,91 @@
...
+        });
+
+        event.preventDefault();
+    }
+
+    select(id, name) {
+        if (this.onSelectCallback !== undefined) {
+            this.onSelectCallback(id, name);
+        } else {
+            this.$input.val(id);
```
grossmannmartin, line 55: Removing or selecting an author (item) does not trigger formchangeinfo, so it's possible to accidentally loose changes

packages/framework/assets/js/admin/components/SinglePicker.js
```diff
@@ -0,0 +1,91 @@
...
+        this.$picker = $picker;
+        this.onSelectCallback = onSelectCallback;
+
+        if (onSelectCallback === undefined) {
+            this.$input = $picker.find('.js-single-picker-input');
+            this.$label = $picker.find('.js-single-picker-label');
+            this.$addButton = $picker.find('.js-single-picker-button-add');
+            this.$removeButton = $picker.find('.js-single-picker-button-remove');
+
+            this.$removeButton.prop('disabled', this.$label.val() === $picker.data('placeholder'));
```
grossmannmartin: What about my favorite Spanish author, `Choose author`? He will have the remove button disabled.

(We should use the `data-js-single-picker-input` value for comparison instead)

packages/framework/src/Form/Admin/Blog/BlogArticleAuthorFormType.php
```diff
@@ -0,0 +1,118 @@
...
+        ]);
+        $builderImageGroup
+            ->add('image', ImageUploadType::class, [
+                'required' => false,
+                'image_entity_class' => BlogArticleAuthor::class,
+                'file_constraints' => [
+                    new Constraints\Image(
+                        maxSize: '2M',
+                        mimeTypes: ['image/png', 'image/jpg', 'image/jpeg', 'image/gif'],
+                        maxSizeMessage: 'Uploaded image is to large ({{ size }} {{ suffix }}). '
```
grossmannmartin: typo here `to` -> `too`. 
It's widely used in the codebase, and @vitek-rostislav already fixed the rest in #4662

packages/framework/src/Form/Transformers/BlogArticleAuthorIdToBlogArticleAuthorTransformer.php
```diff
@@ -0,0 +1,38 @@
...
+use Override;
+use Shopsys\FrameworkBundle\Model\Blog\Author\BlogArticleAuthor;
+use Shopsys\FrameworkBundle\Model\Blog\Author\BlogArticleAuthorRepository;
+use Shopsys\FrameworkBundle\Model\Blog\Author\Exception\BlogArticleAuthorNotFoundException;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+
+class BlogArticleAuthorIdToBlogArticleAuthorTransformer extends AbstractEntityIdToEntityTransformer
+{
+    public function __construct(
+        protected readonly BlogArticleAuthorRepository $blogArticleAuthorRepository,
```
grossmannmartin: Wouldn't the facade be better?

packages/framework/src/Model/Blog/Author/BlogArticleAuthorData.php
```diff
@@ -0,0 +1,40 @@
...
+    public $descriptions;
+
+    /**
+     * @var string|null
+     */
+    public $uuid;
+
+    public function __construct()
+    {
+        $this->name = '';
```
grossmannmartin: This default value seems a bit suspicious. I'm worried it could be error-prone. When no name is set, the app will behave just fine (with wrong data) instead of providing immediate feedback (with error)

packages/framework/src/Model/Blog/Author/BlogArticleAuthorData.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public $name;
+
+    /**
+     * @var \Shopsys\FrameworkBundle\Component\FileUpload\ImageUploadData
+     */
+    public $image;
+
+    /**
+     * @var string[]|null[]
```
grossmannmartin: Shouldn't this be rather `array<string, string|null>` as it's indexed by locale?
The same for the description below

packages/framework/src/Model/Blog/Author/BlogArticleAuthorRepository.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public function getAllBlogArticleAuthorsQueryBuilder(?string $searchText): QueryBuilder
+    {
+        $queryBuilder = $this->getBlogArticleAuthorRepository()->createQueryBuilder('ba');
+
+        if ($searchText !== null && $searchText !== '') {
+            $queryBuilder
+                ->andWhere('LOWER(ba.name) LIKE LOWER(:searchText)')
+                ->setParameter('searchText', '%' . DatabaseSearchingHelper::getLikeSearchString($searchText) . '%');
```
grossmannmartin: We can use `getFullTextLikeSearchString()` instead of manually composing the fulltext search placeholders

project-base/app/src/Resources/definition/blog_article/1.json
```diff
@@ -269,6 +269,25 @@
             },
             "mainBlogCategoryUuid": {
                 "type": "keyword"
+            },
+            "author": {
```
grossmannmartin, line 273: Shouldn't we specify `"type" : "nested"`?

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Blog/Author/BlogArticleAuthorDecorator.types.yaml
```diff
@@ -0,0 +1,28 @@
+BlogArticleAuthorDecorator:
+    type: object
+    decorator: true
+    config:
+        fields:
+            id:
+                type: "Int!"
+                description: "ID of the blog article author"
```
grossmannmartin: Id should not be exposed in the graphql unless there is a very good reason for it

grossmannmartin (changes requested):

packages/framework/src/Component/Grid/Grid.php
```diff
@@ -358,6 +375,15 @@ public function getAllowedLimits(): array
...
     }
 
+    /**
+     * @param int[] $allowedLimits
+     */
+    public function setAllowedLimits(array $allowedLimits): void
+    {
+        $this->allowedLimits = $allowedLimits;
+        $this->loadLimitFromRequest();
+    }
```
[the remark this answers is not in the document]
TomasLudvik: Fixed in 38304fda2c3cd3a71860354a992e4c39f9fe5d64 - `setAllowedLimits()` now resets `$limit` to the default and clears `$isLimitFromRequest` before reloading from the request, so a previously applied limit that is no longer in the allowed set cannot persist.

packages/framework/src/Component/Grid/Grid.php
```diff
@@ -449,6 +472,24 @@ protected function loadFromRequest(): void
...
+        }
+
+        $gridQueryData = $queryData[$this->id];
+
+        if (!array_key_exists('limit', $gridQueryData)) {
+            return;
+        }
+
+        $this->setLimit((int)trim($gridQueryData['limit']));
+        $this->isLimitFromRequest = true;
```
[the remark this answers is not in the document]
TomasLudvik: Fixed in b3c73836b4fb5ddee0f1ebb624172a170f2df10b - `loadLimitFromRequest()` now only marks the limit as request-sourced (and applies it) when the requested value is actually in `allowedLimits`.

packages/administration/templates/form/type/SinglePickerType.html.twig
TomasLudvik: Fixed in 9139be3fa95a3d9d0c16375a2a597284de1052fd - removed the `inline_label` option; the picker now renders its label the same way as other form fields.

packages/administration/templates/form/type/SinglePickerType.html.twig
TomasLudvik: Fixed in 09406fde12cf4fa215f63cb8b2fcb16653681007 - switched the single-picker hooks from `js-...` classes to `data-js-single-picker*` attributes.

packages/framework/assets/js/admin/components/SinglePicker.js
TomasLudvik: Fixed in fced079c89e13c429d5914d05dd0e3e45020f6dd - selecting/removing now triggers `FormChangeInfo.showInfo()` on the form-input path (the order-item callback path already handled it).

packages/framework/assets/js/admin/components/SinglePicker.js
TomasLudvik: Fixed in 0a0afa178dd7bdfc0d4abc91084a56ab49f50654 - the remove button disabled state now keys off the hidden input value instead of comparing the label to the placeholder, so an author literally named "Choose author" is still removable.

packages/framework/src/Form/Admin/Blog/BlogArticleAuthorFormType.php
TomasLudvik: Fixed in 407c5537a0bd85641570a9d55b2eb503c2e3b3ce.

packages/framework/src/Form/Transformers/BlogArticleAuthorIdToBlogArticleAuthorTransformer.php
TomasLudvik: Fixed in d2075dc81b28b032411d998d2d548b0bb03aaa50 - the transformer now depends on `BlogArticleAuthorFacade` instead of the repository.

packages/framework/src/Model/Blog/Author/BlogArticleAuthorData.php
TomasLudvik: Fixed in f199e9ee91fdbfd927084b8c5dcac2136520de25 - changed both `jobTitles` and `descriptions` to `array<string, string|null>`.

packages/framework/src/Model/Blog/Author/BlogArticleAuthorRepository.php
TomasLudvik: Fixed in 7df802012e09502983d0b2b2d94e6a29a46248cd - now uses the injected `DatabaseSearchingHelper::getFullTextLikeSearchString()`.

packages/frontend-api/src/Resources/config/graphql-types/ModelType/Blog/Author/BlogArticleAuthorDecorator.types.yaml
TomasLudvik: Fixed in 00c415fd3b3f7ec66a8db04ab01facd79805fe30 - removed the `id` field from the decorator and regenerated `schema.graphql` and the storefront types.

project-base/app/src/Resources/definition/blog_article/1.json
TomasLudvik: I kept the default `object` mapping instead of `nested`. In Elasticsearch, `nested` is needed for arrays of objects whose sub-fields must be queried independently (that is why `hreflang_links` is `nested`). Here `author` is a single to-one object that we only read back for display, so `object` is sufficient and avoids the extra nested-query overhead. Let me know if you had a specific query use case in mind that would need `nested`.

packages/administration/templates/form/type/SinglePickerType.html.twig
TomasLudvik: Follow-up: I reworked this in e4029741a46814b158bafd9d0d51b81130ae1931 rather than dropping the inline-labeled look. The picker now splits into `AbstractSinglePickerType` with two variants - `SinglePickerType` (the plain button, unchanged for product pickers) and `InlineSinglePickerType` (chip label + search icon). The blog author picker uses the inline variant, so its label renders like the neighbouring `InlineLabeledFieldType` fields (Visible on homepage, Date of creation) instead of restyling the shared picker, and the `inline_label` option stays removed.

packages/framework/src/Model/Blog/Author/BlogArticleAuthorData.php
TomasLudvik: On reflection you are right - addressed in bd897a28855b1db76208016b470ce3c33d9cb1db. Removed the empty-string default from the constructor, so `name` now defaults to `null`. A genuinely missing name then hits the `nullable: false` column constraint (an immediate error) instead of silently persisting an empty string, while the create form still renders an empty field.

grossmannmartin (approved):

chlebektomas (approved):

PR #4694, opened 2026-07-09
Multiple currencies per domain

pk16011990:

This PR adds support for **multiple currencies on a single domain**.

Until now, every domain was bound to exactly one currency selected in the administration. Merchants who wanted to sell in several currencies on one market (e.g., CZK and EUR in the Czech Republic) had to run a separate domain for each currency. This change lets a domain offer any number of currencies, defined in the committed configuration (`config/domains.yaml`, required `currencies` key, the first code being the domain default). The configured currencies are created automatically during the build and cannot be deleted in the administration, so the setup is always consistent with the code.

What the customer gets:

- a currency switcher in the storefront header (shown only when the domain has more than one currency) — the selection is persisted in a cookie, and all prices (listings, filters, product detail, cart, checkout) are displayed in the selected currency,
- orders are placed in the selected currency and permanently remember both the currency and the exchange rate valid at the time of purchase, so order history and emails always show the original amounts.

What the administrator gets:

- transport prices, payment prices, promo code limits, and the free transport & payment limit are entered explicitly per currency (bootstrap-converted from the default currency once by a migration),
- product prices in secondary currencies are, by default, calculated from the domain default currency by the exchange rate; a global config switch (`shopsys.pricing.product_prices_multicurrency_mode: manual`) enables entering product prices manually per currency instead (all-or-nothing per pricing group, so no product ever disappears when the customer switches currency),
- order statistics convert order totals using the exchange rate snapshotted in each order,
- product feeds can be generated per currency, configured per feed via the `currencies` attribute of the `shopsys.feed` tag (the default currency keeps the historical feed URL, so nothing changes for already registered aggregator feeds),
- GoPay payment methods are downloaded and paired per currency.

For developers and API consumers, the selected currency travels in the `X-Currency-Code` HTTP header (with a silent fallback to the domain default), so no GraphQL query or mutation needed a new argument, every `Price` in the API now carries its `currencyCode`, and the `Price` value object optionally carries a `Currency` and refuses to combine prices of different currencies. Projects that keep one currency per domain only need to add the one-line `currencies` key to their domain config — the behavior stays the same. See the included upgrade notes for the full list of BC breaks.

#### Fixes issues

N/A


pk16011990:
It's only a preview/draft of "how it could be solved".

PR #4587, opened 2026-04-29
Replace duplicated category tree forms with generic TreeSelectionType

vitek-rostislav:
#### Description, the reason for the PR

The administration form for selecting categories was implemented twice - once for product categories (`CategoriesType`) and once for blog categories (`BlogCategoriesType`). Both versions duplicated the same logic across PHP form types, custom checkbox types, data transformers, Twig templates and a dedicated JavaScript component. 

Moreover, both types originally extended `CollectionType`, which causes some bugs when an invalid form was submitted (non-checked checkboxes disappeared from the tree).

This PR replaces both implementations with a single generic `TreeSelectionType` that renders a lazily loaded checkbox tree for any entity that implements `TreeSelectionEntityInterface` and exposes its data through a `TreeSelectionDataProviderInterface`. 

`Category` / `BlogCategory` and their facades now implement these interfaces, so `CategoriesType` and `BlogCategoriesType` only configure the new shared type with their own branch-loading route and data provider. The shared Twig template, the renamed `TreeSelectionForm` JS component and a generic `IdsToEntitiesTransformer` cover the rendering, lazy expansion and data conversion for both forms.

#### Fixes issues

closes #207

#### License Agreement for contributions

- [x] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
:globe_with_meridians: Live Preview:
  - https://rv-fix-category-tree-2.odin.shopsys.cloud
  - https://cz.rv-fix-category-tree-2.odin.shopsys.cloud
  - https://rv-fix-category-tree-2.odin.shopsys.cloud/sk

----
:globe_with_meridians: Live Preview:
  - https://rv-fix-category-tree-2.odin.shopsys.cloud
  - https://cz.rv-fix-category-tree-2.odin.shopsys.cloud
  - https://rv-fix-category-tree-2.odin.shopsys.cloud/sk

PR #4701, opened 2026-07-13
Fix expanding lazily loaded branches in admin tree selection

vitek-rostislav:

In the administration, category tree branches that were loaded lazily (via AJAX) could not be expanded any further. For example on the product detail page, when a category had a third level of nesting and the product was not assigned in that branch, the deepest level could not be opened at all — clicking the expand icon loaded the data but nothing appeared.

The cause was a regression from #4587: the HTML template used by JavaScript to render lazily loaded items was missing the children container element, because the Twig condition compared the `__expandable__` placeholder against the literal `'true'`. Children of such items were then appended into an empty jQuery set and never reached the DOM. Branches pre-rendered on the server (root levels and paths to the product's assigned categories) were unaffected, which is why the bug only showed up from a certain nesting level.

The template now always includes the children container for expandable items, and the JavaScript removes it from non-expandable leaves so their markup stays identical to the server-rendered output. This affects all tree selection forms (product categories, blog categories, discount coupons, adverts, …).

#### Fixes issues

N/A

----
  - https://rv-fix-tree-selection-lazy-branch-expand.odin.shopsys.cloud
  - https://cz.rv-fix-tree-selection-lazy-branch-expand.odin.shopsys.cloud
  - https://rv-fix-tree-selection-lazy-branch-expand.odin.shopsys.cloud/sk

----
  - https://rv-fix-tree-selection-lazy-branch-expand.odin.shopsys.cloud
  - https://cz.rv-fix-tree-selection-lazy-branch-expand.odin.shopsys.cloud
  - https://rv-fix-tree-selection-lazy-branch-expand.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/templates/form/type/TreeSelectionType.html.twig
+++ b/packages/administration/templates/form/type/TreeSelectionType.html.twig
@@ -26,7 +26,7 @@
                 </div>
             </div>
         </div>
-        {% if childrenHtml is not empty or isExpandable == 'true' %}
+        {% if childrenHtml is not empty or isExpandable == 'true' or isExpandable == '__expandable__' %}
             <ul class="js-tree-selection-form-children-container tree-group">
                 {{ childrenHtml }}
             </ul>
--- a/packages/framework/assets/js/admin/components/TreeSelectionFormItem.js
+++ b/packages/framework/assets/js/admin/components/TreeSelectionFormItem.js
@@ -141,6 +141,9 @@ export default class TreeSelectionFormItem {
 
         $newItem.data('load-url', itemData.loadUrl);
         $newItem.data('expandable', itemData.isExpandable);
+        if (!itemData.isExpandable) {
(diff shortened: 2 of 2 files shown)


packages/administration/templates/form/type/TreeSelectionType.html.twig
```diff
@@ -25,7 +25,8 @@
                 </div>
             </div>
         </div>
-        {% if childrenHtml is not empty or isExpandable == 'true' %}
+        {# the condition must hold for the '__expandable__' placeholder so that the children container is part of the JS item template #}
+        {% if childrenHtml is not empty or isExpandable != 'false' %}
```
TomasLudvik:
```suggestion
        {% if childrenHtml is not empty or isExpandable == 'true' or isExpandable == '__expandable__' %}
```

TomasLudvik (approved):

PR #4732, opened 2026-07-27
links with target attribute get rel="noopener" automatically

pk16011990:

A page opened in a new tab can manipulate the opener window (reverse tabnabbing) unless the link carries `rel="noopener"`. The attribute is now added automatically everywhere a link with `target="_blank"` can be created or rendered, so an administrator cannot produce an unsafe link by accident and a developer cannot forget it in a template.

**Content authored in the administration** is fixed on save. Links written in CKEditor (both the classic editor and the one embedded in GrapesJS) get `rel="noopener"` merged into their `rel` attribute, and so do links whose target is switched on through the GrapesJS settings panel — those never pass through the editor content, so they needed handling of their own. The same rule is enforced again on the server by the new `HtmlContentProcessor`, so it holds even when the browser-side filter is bypassed. It now covers every field holding HTML: the wysiwyg fields, both GrapesJS editors and the advertisement code field, which had no protection at all. Fields declare that they hold HTML, so plain-text fields such as notes, meta descriptions or the CSP header are left untouched. The processor only ever rewrites the link tags and passes the rest of the markup through unchanged, because stored content legitimately holds advertisement scripts, GrapesJS inline styles and Outlook conditional comments.

**Hand-written markup** is covered by tooling. `rel="noopener"` was added to all links with `target="_blank"` in administration templates, and a new coding-standard fixer keeps it that way — it adds the attribute to any `<a>` tag with `target="_blank"` and can fix violations automatically. Wiring it up uncovered that coding standards never checked Twig templates at all, so the existing checks for a missing `<button type>` and for a forgotten `{{ dump() }}` had never run on a single template. Templates are checked from now on, which also added the missing `type` attribute to six administration buttons — without it a button placed inside a form submits it.

**The storefront** adds the attribute to every link rendered with `target="_blank"`, closing nine places where it was missing. The linter rule only checks plain `<a>` elements, so a target passed to a component was never reported.

Every layer deliberately looks at `target="_blank"` only, matching the keyword case-insensitively. The attribute is meaningless for `_self`, `_parent` and `_top`, which navigate an existing browsing context, and it is actively harmful for a named target: per the HTML specification a link carrying `rel="noopener"` ignores the target name, so `<a target="content-frame">` would stop navigating that frame and open a new tab instead. One link in the administration takes its target from a variable and therefore states the condition in the template itself, as the coding-standard fixer cannot resolve it.

Existing database content is not migrated automatically; the upgrade notes contain an optional SQL update.

#### Fixes issues

N/A

----
  - https://pk-ssp-3084-wysiwyg-rel-noopener.odin.shopsys.cloud
  - https://cz.pk-ssp-3084-wysiwyg-rel-noopener.odin.shopsys.cloud
  - https://pk-ssp-3084-wysiwyg-rel-noopener.odin.shopsys.cloud/sk

----
  - https://pk-ssp-3084-wysiwyg-rel-noopener.odin.shopsys.cloud
  - https://cz.pk-ssp-3084-wysiwyg-rel-noopener.odin.shopsys.cloud
  - https://pk-ssp-3084-wysiwyg-rel-noopener.odin.shopsys.cloud/sk

### DIFF
--- a/packages/coding-standards/src/CsFixer/MissingLinkRelNoopenerFixer.php
+++ b/packages/coding-standards/src/CsFixer/MissingLinkRelNoopenerFixer.php
@@ -1,0 +1,142 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\CodingStandards\CsFixer;
+
+use Override;
+use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\FixerDefinitionInterface;
+use PhpCsFixer\Tokenizer\Tokens;
+use SplFileInfo;
+
+final class MissingLinkRelNoopenerFixer implements FixerInterface
+{
+    use AppendsHtmlAttributeTrait;
+
+    private const string NOOPENER = 'noopener';
+
+    private const string LINK_OPENING_TAG_PATTERN = '@(<a\b)((?:"[^"]*"|\'[^\']*\'|[^>"\'])*?)(\s*/?>)@imsu';
+
+    private const string BLANK_TARGET_ATTRIBUTE_PATTERN
+        = '@(?:^|\s)target\s*=\s*(?:"_blank"|\'_blank\'|_blank(?=[\s/]|$))@isu';
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function getDefinition(): FixerDefinitionInterface
+    {
+        return new FixerDefinition(
+            'Adds mandatory rel="noopener" to <a> HTML tag with a target="_blank" attribute.',
+            [
+                new CodeSample('<a href="https://example.com" target="_blank">label</a>'),
+                new CodeSample('<a href="https://example.com" target="_blank" rel="nofollow">label</a>'),
+                new CodeSample("<a\n    href=\"https://example.com\"\n    target=\"_blank\"\n>label</a>"),
+            ],
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function isCandidate(Tokens $tokens): bool
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function isRisky(): bool
+    {
+        return false;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function fix(SplFileInfo $file, Tokens $tokens): void
+    {
+        $originalCode = $tokens->generateCode();
+        $code = preg_replace_callback(
+            self::LINK_OPENING_TAG_PATTERN,
+            function ($matches) {
+                $beginning = $matches[1];
+                $attributes = $matches[2];
+                $end = $matches[3];
+
+                if (!preg_match(self::BLANK_TARGET_ATTRIBUTE_PATTERN, $attributes)) {
+                    return $beginning . $attributes . $end;
+                }
+
+                return $beginning . $this->addNoopenerToRel($attributes) . $end;
+            },
+            $originalCode,
+        );
+
+        // the pattern is UTF-8 aware, so a template with broken encoding makes the match fail — leave it alone
+        $tokens->setCode($code ?? $originalCode);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function getName(): string
+    {
+        return 'Shopsys/missing_link_rel_noopener';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function getPriority(): int
+    {
+        return 0;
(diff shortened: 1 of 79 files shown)


packages/framework/src/Component/HtmlSanitizer/WysiwygHtmlSanitizer.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Shopsys\FrameworkBundle\Component\HtmlSanitizer;
+
+use Symfony\Component\DomCrawler\Crawler;
+
+class WysiwygHtmlSanitizer
+{
+    public function sanitize(?string $html): ?string
+    {
+        // quick check so unaffected content is returned as is, without being normalized by the DOM round-trip
+        if ($html === null || stripos($html, 'target=') === false) {
```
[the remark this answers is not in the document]
vitek-rostislav: this is resolved

packages/framework/src/Resources/config/services.yaml
```diff
@@ -13,7 +13,7 @@ services:
 
     Shopsys\FrameworkBundle\:
         exclude: '../../{Command,Controller,DependencyInjection,Form,Migrations,Resources,Twig}'
-        resource: '../../**/*{CategoryAutomatedFilter,Calculation,Context,Dispatcher,Enum,Executor,Extractor,Facade,Factory,Formatter,Freezer,Generator,Handler,Helper,InlineEdit,Listener,Loader,Mapper,Middleware,Parser,Provider,Recalculator,Registry,Repository,Resolver,Sender,Service,Scheduler,Subscriber,Task,Transformer,DataFetcher}.php'
+        resource: '../../**/*{CategoryAutomatedFilter,Calculation,Context,Dispatcher,Enum,Executor,Extractor,Facade,Factory,Formatter,Freezer,Generator,Handler,Helper,InlineEdit,Listener,Loader,Mapper,Middleware,Parser,Processor,Provider,Recalculator,Registry,Repository,Resolver,Sender,Service,Scheduler,Subscriber,Task,Transformer,DataFetcher}.php'
```
grossmannmartin, line 16: If we add `Processor` as an auto-registered suffix, we should: 
- remove custom `*Processor` definitions (ImageProcessor, OrderProcessor)
- add the same suffix to the `project-base/app/config/services.yaml` file

packages/coding-standards/src/CsFixer/MissingLinkRelNoopenerFixer.php
```diff
@@ -0,0 +1,142 @@
...
+        = '@(?:^|\s)target\s*=\s*(?:"_blank"|\'_blank\'|_blank(?=[\s/]|$))@isu';
+
+    /**
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function getDefinition(): FixerDefinitionInterface
+    {
+        return new FixerDefinition(
+            'Adds mandatory rel="noopener" to <a> HTML tag with a target attribute.',
```
grossmannmartin: `... with a target="blank" attribute.` ?

project-base/app/package.json
```diff
@@ -75,6 +75,9 @@
         "sortablejs": "^1.15.6"
     },
     "jest": {
-        "testEnvironment": "jsdom"
+        "testEnvironment": "jsdom",
+        "moduleNameMapper": {
+            "^framework/(.*)$": "@shopsys/framework/js/$1"
```
grossmannmartin: Why do we need this change?
I mean, I thought the bundling is handled by webpack

grossmannmartin (changes requested):

packages/framework/src/Resources/config/services.yaml
vitek-rostislav: yes, thank you! I meant to do that but forgot about it :facepalm:  dfcfa685c475b28943660e346e74624bbf5118ca

packages/coding-standards/src/CsFixer/MissingLinkRelNoopenerFixer.php
vitek-rostislav: 493cdac90205bc5ed49c9415b918e02ac7d44c41

project-base/app/package.json
vitek-rostislav: yes, thanks, reverted in c4803f8d96e20ac46159497ed934fe53ec8dadc2

vitek-rostislav: it was actually needed for the tests, not the bundling. Jest does not use webpack, so it could not resolve the framework/… alias in linkNoopener.test.js. However, the test was quite redundant, the behavior is tested in `packages/framework/assets/js/common/utils/addRelNoopenerWhenTargetIsBlank.test.js`

I will create a separate issue for the tests so that we are able to write similar tests in the future

vitek-rostislav: btw I tried to create a separate PR, but it feels like something I would rather not pursue, IMHO it is not worth it. We can talk about it if you want, but I am also ok if you just close this thread :smile:

grossmannmartin (approved):

PR #4745, opened 2026-07-30
Modernize and expand the demo storefront

chlebektomas:

This PR gives the demo storefront a broader and more modern visual foundation. Content makes better use of large screens, while the refreshed color palette, action hierarchy, iconography, and shared controls create a more consistent experience across the site.

The homepage, header, mobile and blog navigation, footer, cart, product listings, product detail, consent, and newsletter flows have been modernized together with richer demo content. Responsive images now request layout-appropriate sizes, long product details keep the purchase action available, and adding a product to comparison confirms the action with a toast instead of interrupting the customer with a popup.

Accessibility and interaction details are aligned across the storefront: drawers, tooltips, icon buttons, focus behavior, semantic markup, and fixed-header anchors now behave consistently on desktop and mobile.

#### Fixes issues

N/A

----
  - https://tc-ssp-4159-expand-storefront-content.odin.shopsys.cloud
  - https://cz.tc-ssp-4159-expand-storefront-content.odin.shopsys.cloud
  - https://tc-ssp-4159-expand-storefront-content.odin.shopsys.cloud/sk


project-base/storefront/components/Layout/Footer/NewsletterForm/newsletterFormMeta.tsx
```diff
@@ -11,23 +11,25 @@ import { yupResolver } from 'utils/forms/yupResolver';
...
 export const useNewsletterForm = (): [UseFormReturn<NewsletterFormType>, NewsletterFormType] => {
     const { t } = useTranslation();
     const resolver = yupResolver<NewsletterFormType>(
         Yup.object().shape<Record<keyof NewsletterFormType, any>>({
             email: validateEmail(t),
-            privacyPolicy: validatePrivacyPolicy(t),
         }),
     );
-    const defaultValues = { email: '', privacyPolicy: false };
+    const defaultValues = { email: '' };
```
JanMolcik, line 25: **Critical — GDPR consent downgraded from explicit opt-in to consent-by-submission.**

Base had:
```ts
validationSchema: { email: validateEmail(t), privacyPolicy: validatePrivacyPolicy(t) }
const defaultValues = { email: '', privacyPolicy: false };
```
plus a `CheckboxControlled` in the form. This PR validates `email` only, drops the default, removes the field from `NewsletterFormType`, and replaces the checkbox with non-interactive prose (`NewsletterForm.tsx:89`, `NewsletterSubscriptionAgreement`).

Marketing consent needs a freely-given affirmative act — an unticked box. Submit-implies-consent is the pattern EDPB Guidelines 05/2020 reject under GDPR Art. 4(11)/7.

Two things make this easy to miss: it fails **no** typecheck and **no** test (the mutation only ever accepted `$email`), and it is **inconsistent within this PR** — `Pages/Contact/contactFormMeta.tsx:25` still uses `validatePrivacyPolicy`. Upgrade note line 56 frames it as a mechanical field removal rather than a consent-model change.

**This one needs your answer rather than a patch from me:** is this deliberate and backed by server-side double opt-in (which would change the analysis), or incidental fallout from the footer redesign? If deliberate, please call it out prominently in the upgrade notes; if not, restore the required checkbox.

*Test:* assert submit is blocked until the consent box is checked.

project-base/storefront/components/Pages/UserConsent/UserConsentContent.tsx
```diff
@@ -17,7 +20,7 @@ export const UserConsentContent: FC = () => {
     return (
         <Webline width="lg">
             <VerticalStack gap="sm">
-                <h1>{t('User consent')}</h1>
+                <PageHero description={<UserConsentPolicyLink />} icon={LockCheckIcon} title={t('User consent')} />
```
JanMolcik, line 23: **Critical — static consent prose is rendered inside a `role="alert"` live region.**

`PageHero.tsx:65-73` renders `description` in:
```jsx
<p aria-atomic="true" aria-live="polite" role="alert">{description}</p>
```

Three compounding problems:

1. `role="alert"` implies **assertive** and overrides the `aria-live="polite"` attribute, so the whole policy sentence (plus the `sr-only` one) interrupts on page load. A page introduction is not a status message — WCAG 4.1.3.
2. The `alert` role exposes its children as a flat text alternative. `UserConsentPolicyLink` puts an interactive `<a>` inside it, so the "consent and tracking policy" link can be flattened to text and disappear from the screen-reader links rotor.
3. `role="alert"` only makes sense for `PageHero`'s error/success variants.

**Fix (also resolves the `UserConsentForm` comment below):** keep the policy link in the form and pass no `description` here —
```jsx
<PageHero icon={LockCheckIcon} title={t('User consent')} />
```
or add `isDescriptionLive={false}` to omit `role="alert"`/`aria-live`.

*Test:* `expect(screen.queryByRole('alert')).not.toBeInTheDocument()` and the policy link resolves via `getByRole('link', …)`.

project-base/storefront/components/Blocks/UserConsent/UserConsentForm.tsx
```diff
@@ -52,83 +51,103 @@ export const UserConsentForm: FC<UserConsentFormProps> = ({ onSetCallback }) =>
...
-                        link: userConsentPolicyArticleUrl ? (
-                            <ExtendedNextLink
-                                aria-describedby="user-consent-policy-link"
-                                href={userConsentPolicyArticleUrl}
-                                rel="noreferrer"
-                                tabIndex={0}
-                                target="_blank"
-                                title={t('Consent and tracking policy')}
+            <div className={twMergeCustom('flex flex-col gap-5', isFloating ? 'gap-4' : 'mx-auto w-full max-w-2xl')}>
+                {isFloating && (
```
JanMolcik: **Warning — the `page` variant lost its policy link entirely.**

The policy block is now behind `{isFloating && (…)}`; base rendered the `sr-only` paragraph + `Trans` link **unconditionally** for both variants.

It only looks fine because `UserConsentContent.tsx:23` re-adds it via `PageHero` — a hidden coupling between two components in different directories. `UserConsentForm` is a shared, downstream-extended component: any project rendering `<UserConsentForm />` (the `page` default) outside `UserConsentContent` silently loses the legally-relevant link, with no type error and no upgrade note.

`UserConsentForm.test.tsx:88-92` asserts the removal ("does not repeat the policy introduction in the page variant"), so this won't be caught.

**Fix:** keep the link in the form for `page` too and drop `description` from `PageHero` — resolves the `role="alert"` critical at the same time. Otherwise document the coupling on the `variant` prop and in the upgrade notes.

*Test:* `render(<UserConsentForm />); expect(screen.getByRole('link', { name: /consent and tracking policy/i })).toBeInTheDocument();`

project-base/storefront/utils/productLists/comparison/useComparison.tsx
```diff
@@ -35,7 +27,7 @@ export const useComparison = () => {
                 showErrorMessage(t('Unable to add product to comparison.'));
             },
             addProductSuccess: (result, productUuid) => {
-                updatePortalContent(<ProductComparePopup />);
+                showSuccessMessage(t('Product added to comparison.'));
```
JanMolcik, line 30: **Critical (two issues on this line).**

**A) The toast can steal focus to a stale element from an unrelated flow.**

This hook dropped `storeCurrentFocus()` (base line 66), but `showMessage`'s `onClose` still calls `restoreStoredFocus()` **unconditionally** (`utils/toasts/showMessage.tsx:72,82,91,100`), and `storedFocusElement` is global session state cleared only inside `restoreStoredFocus`/`clearStoredFocus`.

Traced repro: `utils/cart/useAddToCartHandler.tsx` calls `storeCurrentFocus()` *before* `await addToCart(...)`. If the mutation fails, no popup opens, so the stored value is never consumed and stays pointing at the add-to-cart button. Add a product to comparison afterwards → 6s later (`autoClose={6000}`) focus jumps back to that stale button, possibly mid-typing, on another product, or after a route change. WCAG 3.2.1 / 2.4.3.

*Fix:* make restoration opt-in — add `shouldRestoreFocus?: boolean` to `ShowMessageOptions` and only attach `onClose: () => restoreStoredFocus()` when set; or call `clearStoredFocus()` here before `showSuccessMessage`.

*Note:* `useComparison.test.tsx` seeds `storedFocusElement: null` in `beforeEach`, which is exactly what masks this. A case seeding a real element and asserting focus does **not** move would catch it.

**B) The toast gives no way to reach the comparison page — and the plumbing for it was added but never wired.**

The removed `ProductComparePopup` had a "Show products comparison" link. This PR **added** `action?: ReactNode` to `ShowMessageOptions` and renders it (`showMessage.tsx:23,28,36`), threading `options` through `showSuccessMessage` — but **no caller anywhere passes `action`**. Correspondingly `Show products comparison` was deleted from all three locales (present in base for cs/en/sk, now 0/0/0).

The goal was "confirm without interrupting", but the customer loses the only affordance to continue. Cypress was rewritten around it — `goToComparisonFromPopup()` → `visitComparisonPage()` (a direct `cy.visit`), so no test covers any user-reachable path from confirmation to comparison.

*Fix:*
```tsx
showSuccessMessage(t('Product added to comparison.'), {
    action: <Link href={productComparisonUrl} type="comparison">{t('Show products comparison')}</Link>,
});
```
and restore the translation key. Otherwise delete the unused `action` plumbing as dead code.

project-base/storefront/components/Layout/CommonLayout.tsx
```diff
@@ -82,7 +123,7 @@ export const CommonLayout: FC<CommonLayoutProps> = ({
 
                 <NotificationBars />
 
-                {isDesktop && showDesktopFixedHeader && (
+                {isDesktop && (
```
JanMolcik, line 126: **Critical — duplicate DOM `id`s on desktop (WCAG 4.1.1) and misdirected `aria-labelledby`/`aria-describedby`.**

This gate changed from `{isDesktop && showDesktopFixedHeader && (<FixedHeader/>)}` (base:85) to `{isDesktop && (…)}`, so `FixedHeader` — and its second `MenuIconic` (`FixedHeader.tsx:46`) — is now **permanently mounted** on desktop instead of only while scrolled.

`FixedHeader` does the right thing (`loginFormName="fixed-header-login-form"`, `shouldUseLocalUserMenuState`), but `MenuIconic.tsx:99` threads that flag **only to the authenticated branch** — line 103's `MenuIconicItemUserUnauthenticated` never receives it and holds `isActive` in local `useState`. So both account popovers open independently, and with both open three hardcoded IDs exist twice:

- `registration-benefits-title`, `registration-benefits-list` — `MenuIconicItemUserUnauthenticatedContent.tsx:35,39`
- `login-form-description` — `Blocks/Login/LoginForm.tsx:66`

IDREFs resolve to the first match in document order, so the fixed header's controls point at the regular header's content.

Worth noting the fix follows the file's own pattern: `LoginForm` **already** namespaces its *input* IDs via `formName` (`loginFormMeta.ts:33`) — only the description `<p>` was missed.

*Fix:* derive these IDs from the already-unique `loginFormName` (or `useId()`), and pass `loginFormName` down to `RegistrationBenefits` (it currently receives only `registrationUrl` and `onMenuClose`). Alternatively restore `showDesktopFixedHeader` in this gate if permanent mounting wasn't intended.

*Test:* desktop Cypress — scroll to reveal the fixed header, open both account popovers, then `cy.get('#registration-benefits-title').should('have.length', 1)` and likewise for the other two.

project-base/storefront/styles/utilities.css
```diff
@@ -35,9 +35,17 @@
...
 @utility animate-in {
     animation: animateIn 0.3s ease 0.05s both;
 }
+
+@utility scroll-mt-fixed-header {
+    scroll-margin-top: calc(var(--fixed-header-height, 0px) + var(--spacing-5));
+}
+
+@utility scroll-mt-fixed-header-with-navigation {
+    scroll-margin-top: calc(var(--fixed-header-height, 0px) + 5rem);
```
JanMolcik, line 50: **Warning — mobile scroll offset is bound to a desktop-only CSS variable.**

```css
scroll-margin-top: calc(var(--fixed-header-height, 0px) + 5rem);
```

`--fixed-header-height` is only ever set by `useDesktopFixedHeader.ts:71`, which bails out when `!isDesktop` (lines 61-63) and removes the property. On mobile the offset is therefore a hardcoded `5rem` with no relation to actual nav height.

This matters more than before, because `useHashNavigation.ts:71-75` was changed from a fixed `SCROLL_OFFSET = 150` to reading `scrollMarginTop` off each section — so both the landing position **and** active-section detection now derive from that possibly-wrong 80px. The mobile nav sits at `top: 0` with its own real height (`--sticky-navigation-offset` is also unset on mobile). If it renders taller or shorter than 80px, tapping a section tag lands the heading under the nav or over-scrolls, and the wrong tag highlights.

*Fix:* set a mobile equivalent of `--fixed-header-height`, or measure the nav height into a CSS variable in `ProductDetailSectionNavigation` and use it instead of the `5rem` literal.

*Test:* Cypress at mobile viewport — click each section tag and assert the target's `getBoundingClientRect().top >= navBottom`.

project-base/storefront/utils/productLists/comparison/useComparisonTable.tsx
```diff
@@ -7,19 +7,17 @@ export const useComparisonTable = (
...
     calcMaxMarginLeft: () => void;
     tableMarginLeft: number;
 } => {
     const [isArrowLeftActive, setArrowLeftActive] = useState(true);
     const [isArrowRightActive, setArrowRightActive] = useState(true);
-    const [isArrowLeftShowed, setArrowLeftShowed] = useState(true);
-    const [isArrowRightShowed, setArrowRightShowed] = useState(true);
     const [tableMarginLeft, setTableMarginLeft] = useState(0);
     const [tableMaxMarginLeft, setTableMaxMarginLeft] = useState(0);
+    const shouldShowArrows = tableMaxMarginLeft > 0;
```
JanMolcik, line 20: **Warning — the left arrow ships enabled but dead on first paint.**

`isArrowLeftActive` initialises to `useState(true)` (line 16) while `tableMarginLeft` is `0`. The only corrector is `setMargin`, reached from the slide handlers or the `useComponentUpdate` at line ~78 — and `useComponentUpdate` **deliberately skips mount**. So the left arrow is fully enabled at scroll position 0.

This regressed in severity here: `ProductComparisonContent.tsx` moved from `hasDisabledLook={!isActive}` (base:107) to a real `disabled={!isActive}` (:102). Previously it looked disabled but was a live button; now a user can tab to an **enabled** control, activate it, and get nothing — with no disabled affordance.

*Fix:* `useState(false)`, or derive it like the neighbouring line: `const isArrowLeftActive = tableMarginLeft > 0;`

*Test:* vitest on `useComparisonTable` — assert `isArrowLeftActive === false` on initial render.

While here: `calcMaxMarginLeft` is recreated every render and sits in a `useEffect` dep array, so every render triggers two `getBoundingClientRect()` layout reads; `getParametersData(comparedProducts)` is likewise unmemoised. Pre-existing, but this PR makes arrow visibility depend on `tableMaxMarginLeft`, so the churn matters more. `useCallback([])` / `useMemo([comparedProducts])` would settle it.

project-base/storefront/components/Forms/Button/IconButton.tsx
```diff
@@ -1,45 +1,101 @@
...
+        const button = (
+            <button
+                aria-label={ariaLabel ?? title}
+                className={twMergeCustom(getIconButtonClassName(variant, size, shape), className)}
+                data-tid={tid}
+                disabled={disabled}
+                ref={ref}
+                tabIndex={tabIndex}
+                title={tooltipLabel ? undefined : title}
+                type={type}
```
JanMolcik, line 49: **Warning — disabled icon buttons lose their hint entirely.**

`title={tooltipLabel ? undefined : title}` removes the native tooltip in favour of a Radix one whose trigger is the `<button>` itself. But the class list includes `disabled:pointer-events-none` (line 79), and a disabled button fires no focus events — so while disabled, **neither hover nor keyboard** can surface the tooltip, and the `title` fallback is gone.

This bites hardest in the cart, where `RemoveCartItemButton.tsx:21-34` and `CartList/CartListItem.tsx:321-337` pass `disabled={isRemovingFromCart}` — exactly the moment a user hovers to understand why the control is unresponsive. `DropzoneControlled.tsx:120-130` too.

`Spinbox.tsx:447` already solves this correctly by wrapping in a span, so the trigger isn't the element carrying `pointer-events-none`:

```tsx
return tooltipLabel ? (
    <Tooltip label={tooltipLabel} placement={tooltipPlacement}>
        <span className="inline-flex">{button}</span>
    </Tooltip>
) : button;
```

*Test:* render a `disabled` `IconButton` with `tooltipLabel`, fire `pointerMove` on the wrapper, advance timers, assert `getByRole('tooltip')` is present.

Minor, related: `aria-label={ariaLabel ?? title}` (line 42) means an accidental `ariaLabel=""` produces an unlabelled button, since `''` is not nullish. `||` would fall back safely. Same shape as `DrawerCloseButton.tsx:13`, where `title ?? t('Close') ?? 'Close'` has a dead trailing branch (`t()` returns `string`).

project-base/storefront/components/Layout/Footer/FooterExtras.tsx
```diff
@@ -23,56 +24,60 @@ export const FooterExtras = () => {
...
-                                        className="h-8 w-16 object-contain object-center"
-                                        height="64"
-                                        sizes="(max-width: 768px) 64px, 128px"
-                                        src={transport.mainImage?.url}
-                                        width="128"
-                                    />
-                                </div>
+                                <Tooltip key={transport.mainImage?.name || index} label={transport.name}>
+                                    <div className="flex items-center">
+                                        <Image
```
JanMolcik, line 33: **Warning — the transport name is available to mouse users only.**

`<Tooltip label={transport.name}>` wraps a `<div className="flex items-center">`. Radix `Trigger asChild` merges handlers onto that div but does **not** add `tabIndex`, so it isn't focusable and the tooltip opens on pointer hover only.

`transport.name` is new information introduced by this PR (base had no tooltip here). The only other text is `alt={transport.mainImage?.name || 'Transport'}` — the *image* name, not the transport name. So keyboard and screen-reader users get nothing while mouse users get the name. WCAG 2.1.1 / 1.3.1.

Contrast the social links just below (`:48-58`), which get this right with a focusable `<a tabIndex={0}>` — there the tooltip does open on focus.

*Fix:* put the meaningful label in the accessible tree — `alt={transport.name}` on the `Image` — and/or add `tabIndex={0}` to the wrapper so the tooltip is keyboard-reachable.

*Test:* mock a transport `{ name: 'DPD', mainImage: { name: 'dpd-logo', url: '/dpd.png' } }` and assert `getByAltText('DPD')`.

Same non-focusable-div-trigger pattern exists at `Pages/Customer/CustomerRecordElements.tsx:157-158`.

project-base/storefront/components/Pages/ProductDetail/ProductDetailSections/ProductDetailStickyAction.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    )}
+                </div>
+            </div>
+
+            <ProductPrice className="vl:flex hidden shrink-0" productPrice={product.price} textPriceSize="base" />
+
+            <div className="vl:w-40 w-36 shrink-0">
+                <DeferredProductDetailAddToCart
+                    buttonSize="large"
+                    buttonTid={TIDs.product_detail_sticky_addtocart_button}
```
JanMolcik, line 70: **Warning — two simultaneous add-to-cart buttons: duplicate accessible name + double-submit.**

When the desktop nav is stuck, the inline sticky action is visible while the hero add-to-cart is still mounted (scrolled out of view, nothing unmounts it).

**Duplicate name:** both render `ProductDetailAddToCart`, whose `Button` sets `aria-label={ariaLabel}` (`ProductDetailAddToCart.tsx:106`). Traced: `isWithSpinbox: false` (line 46) means `spinboxRef.current` is always `null`, so `useAddToCartAriaLabel`'s `onFocusHandler` returns early and `ariaLabel` stays `undefined` forever. Both buttons therefore expose the same generic "Add to cart" with no distinguishing context — WCAG 2.4.4 / 4.1.2 ambiguity when navigating by button list.

**Double-submit:** these are *separate* component instances, each with its own `useAddToCartHandler` → own `isAddingToCart` (line 41). `disabled={isAddingToCart}` (line 108) therefore guards each button independently — clicking the hero button then immediately the sticky one fires two `addToCart` mutations for the same product, because the second instance's local flag is still `false`. Whether the cart ends at quantity 1 or 2 depends on backend add-vs-set semantics; worth confirming, since `isAbsoluteQuantity` is only passed from `CartItemQuantityControls.tsx:77`.

*Fix:* include `product.fullName` in the sticky instance's `aria-label` (or label the sticky region), and gate the pending state on shared cart state rather than per-instance.

Credit where due: tab order is correctly excluded when hidden (`hidden` on the container, line 23), and there's no first-render duplication — `useMediaMin` returns `undefined` initially and both placements are guarded on the resolved breakpoint.

*Coverage note:* `TIDs.product_detail_sticky_action` and `product_detail_sticky_addtocart_button` are declared and applied but have **zero** usage in `cypress/` or `vitest/`. The "long product details keep the purchase action available" headline has no e2e coverage — the one related vitest case mocks `DeferredProductDetailAddToCart` out entirely.

project-base/storefront/components/Pages/ProductDetail/ProductDetailSections/ProductDetailSectionNavigation.tsx
```diff
@@ -88,31 +95,47 @@ export const ProductDetailSectionNavigation: FC<ProductDetailSectionNavigationPr
...
-                    />
+                <Webline className="flex items-center gap-6 py-4">
+                    <div className="min-w-0 flex-1">
+                        <HorizontalScrollHint
+                            render={(scrollContainerRef) => (
+                                <div ref={scrollContainerRef} className="flex gap-3 overflow-x-auto">
+                                    {sections.map((section) => (
+                                        <Tag
+                                            key={section.id}
+                                            buttonRef={(el) => {
```
JanMolcik, line 109: **Suggestion — `buttonRefs` entries are never evicted.**

The callback ref only ever `set`s:
```tsx
buttonRef={(el) => { if (el) { buttonRefs.current.set(section.id, el); } }}
```

Section visibility is data-driven (`ProductDetailSections.tsx:65` filters by `isVisible`), so client-side navigation between products with different section sets leaves detached `HTMLButtonElement`s retained for the component's lifetime. The scroll-into-view effect fails safe (a detached button has `parentElement === null`), so this is a retention issue rather than a crash.

```tsx
buttonRef={(el) => {
    if (el) { buttonRefs.current.set(section.id, el); }
    else { buttonRefs.current.delete(section.id); }
}}
```

Otherwise the listener hygiene in this file is good — scroll/resize/MutationObserver and the rAF handle are all cleaned up properly.

project-base/storefront/components/Blocks/BlogSignpost/BlogSignpostPanel.tsx
```diff
@@ -0,0 +1,118 @@
...
+    onPanelChange,
+}) => {
+    const { t } = useTranslation();
+    const panelRef = useRef<HTMLDivElement>(null);
+    const isActive = activePanelUuid === blogCategory.uuid;
+    const isAncestorOfActivePanel = activePanelPathUuids.includes(blogCategory.uuid);
+    const isCurrent = activeItem === blogCategory.uuid;
+    const isTransitioningPanel = previousPanelUuid === blogCategory.uuid;
+
+    useEffect(() => {
```
JanMolcik, line 46: **Warning — two panel-state issues.**

**A) Stale `previousPanelUuid` leaves an off-screen panel permanently painted.** It's set on every panel change but cleared only by the `initialPanelUuid` effect, so the previous panel keeps `isTransitioningPanel === true` and never regains `invisible`. *Fix:* clear it on `onTransitionEnd`.

**B) `panelUuidToFocus` is never cleared, so focus can be stolen.** It refires whenever `isActive` flips back to `true` for a panel still recorded as the focus target — a urql cache update to `blogCategories` can jump focus into the sidebar while the user is elsewhere. *Fix:* consume it as a one-shot, or hold it in a ref.

Unrelated but nearby, and **not** a defect — flagging so it isn't "fixed" by mistake: `BlogSignpost.tsx:23` using `blogCategoryItems?.[0]` instead of base's `.map()` over all roots is safe, because the domain enforces a single root (`BlogCategoryFacade.php:49` throws `RootLevelBlogCategoryAlreadyExistsException`, and migration `Version20241112100245.php:49` hard-fails on more than one). A short comment encoding that invariant would help the next reader.

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuSubItems.tsx
```diff
@@ -15,69 +16,78 @@ type SubMenuProps = {
...
-        <button className={subMenuItemTwClass} tabIndex={0} onClick={onClick}>
+        <ExtendedNextLink
+            passHref
+            className="flex min-h-12 items-center gap-3 rounded-md bg-background-more px-3 py-3 text-left font-secondary font-semibold text-base text-text-default no-underline transition-colors hover:bg-background-most hover:text-text-default hover:no-underline active:bg-background-most"
+            href={href}
+            type={type}
+            onClick={onClick}
+        >
+            <span className="relative shrink-0">
+                <Icon className="size-6" />
```
JanMolcik, line 83: **Warning — quick-link counts dropped out of the accessible name.**

Base rendered the count inside the link text ("Comparison (2)"); it is now an `aria-hidden` badge with no text alternative, so screen-reader users lose the count entirely.

`vitest/components/Layout/Header/MobileMenuSubItems.test.tsx:66-67` asserts the `aria-hidden`, so this reads as intentional — but it is a net accessibility loss versus base.

*Fix:* add an `sr-only` sibling with the count, or fold it into the link's `aria-label`.

project-base/storefront/utils/ui/useScrollRestoration.ts
```diff
@@ -23,26 +26,46 @@ export const useScrollRestoration = <E extends keyof MutuallyExcludedProps>({
...
-                if (scrollY && scrollY > 0) {
-                    window.scrollTo({
-                        top: scrollY,
-                        behavior: 'smooth',
-                    });
-                }
-            }
+            return;
+        }
+
```
JanMolcik, line 38: **Warning — this works, but only by accident.**

The desktop gate returns early while `--fixed-header-height` is still empty and relies on a `MutationObserver` to retry. But observers only fire on changes *after* `observe()` — if the property is already set when the effect runs, nothing ever fires.

Today that never happens, because `FixedHeader` is `dynamic({ssr: false})` and `isDesktop`-gated, which defers the write to a later commit. Any change that makes the property land earlier (an SSR'd inline style, a layout effect, a hydration script) silently kills scroll restoration: `scrollRestored.current` stays `false` and the timer has already fired.

*Fix:* re-check the property inside the effect body right after attaching the observer, or key off the `fixedHeaderHeight` state that `useDesktopFixedHeader` already returns.

SSR safety itself is fine — every `window` access is inside `useEffect`/`useEffectEvent`, so no hydration mismatch.

JanMolcik (changes requested):
## Request changes

Strong redesign overall — disciplined removals (24 deleted components with **zero** dangling references), honest behavioural tests (39 test files, +1589 lines, **no snapshot churn** — all `getByRole`/`toHaveAccessibleName`), accurate and thorough upgrade notes, and GraphQL in sync end-to-end. Vitest passes locally on this branch: **148/148 files, 1331/1331 tests**.

Five blocking items, all traced to source on `bee1264`. Line comments below.

### Criticals

1. **Newsletter consent** — the required opt-in checkbox became passive "by submitting" text. Needs a **human/legal decision**, not a patch. Inconsistent with `Pages/Contact/contactFormMeta.tsx:25` in this same PR, which still uses `validatePrivacyPolicy`.
2. **`role="alert"` around static consent prose** — assertive interruption on load + the policy link may drop out of the links rotor.
3. **Comparison toast can steal focus** to a stale element via global `storedFocusElement`.
4. **The toast has no route to the comparison page** — and the `action` plumbing added for it is never wired by any caller.
5. **Duplicate DOM `id`s on desktop** (WCAG 4.1.1) — `FixedHeader` is now permanently mounted.

Items 2 and 9 share one fix (keep the link in the form, drop `description` from `PageHero`). Item 5 is the cheapest high-value fix.

### Test coverage

The new tests are genuinely good. Two gaps worth noting:
- `vitest/utils/productLists/useComparison.test.tsx` seeds `storedFocusElement: null`, which **masks** critical 3.
- The new `product_detail_sticky_action` / `product_detail_sticky_addtocart_button` TIDs have **zero** usage in `cypress/` or `vitest/`, so the "keep the purchase action available" headline has no e2e coverage.

### Not verified

- **Jira SSP-4159 / 4075 / 4152 / 4243 could not be read** (my Atlassian access does not cover `shopsys.atlassian.net`). **No acceptance criteria were checked** — findings 1 and 9 in particular may be deliberate product decisions I have no way to confirm. Please correct me where that is the case.
- **Review env returned HTTP 403** for `tc-ssp-4159-expand-storefront-content.odin.shopsys.cloud` (and the `cz.`/`/sk` variants), so **no runtime/Playwright verification was performed**. All findings are static traces plus the local Vitest run. Visual/responsive regressions, real mobile nav heights, and screen-reader announcement order are unverified.

### Verified clean

All 24 removed components de-referenced (incl. `packages/` and Twig) · Cypress TIDs 124/124 consistent · `schema.graphql` → `SettingsQuery.graphql` → `.generated.tsx` → `types.ts` in sync, no `make generate-schema` drift · `--color-button-transparent-*` → `--color-button-inverted-*` rename has no dangling refs · Drawer focus trap unchanged · `articleHeadingAnchors` sanitises IDs to `[a-z0-9-]`, no injection via `dangerouslySetInnerHTML` · skip links resolve.

One doc nit outside the diff: `.agents/skills/cypress-tests/SKILL.md:874,882` still documents `TIDs.comparison_popup{,_link}`, both removed from `tids.ts` — agents following that skill will write broken tests.

project-base/storefront/components/Layout/Footer/NewsletterForm/newsletterFormMeta.tsx
chlebektomas: This is intentional. Since this is a dedicated newsletter form, submitting it is the affirmative action. EDPB does not specifically require a checkbox; it requires the action to be clear.

project-base/storefront/components/Pages/UserConsent/UserConsentContent.tsx
chlebektomas: Fixed. Static descriptions are no longer live regions. Dynamic messages use a persistent role="status" region so VoiceOver announces them reliably.

project-base/storefront/components/Blocks/UserConsent/UserConsentForm.tsx
chlebektomas: Refactored. UserConsentForm now only owns the controls and uses a presentational layout prop. The page and floating wrappers explicitly own their heading and policy link.

project-base/storefront/utils/productLists/comparison/useComparison.tsx
chlebektomas: Fixed the stale focus restoration—closing a toast no longer moves focus. The missing link is intentional: comparison is available in both the header and sticky header, keeping the behavior consistent with wishlist. Removed the unused action support.

project-base/storefront/components/Layout/CommonLayout.tsx
chlebektomas: Fixed. The IDs are now scoped by loginFormName, preventing duplicates and incorrect ARIA references. Added a desktop Cypress test covering both account popovers.

project-base/storefront/styles/utilities.css
chlebektomas: This is intentional. The mobile section navigation is 64px high, with the remaining 16px used as spacing, so the 5rem offset is correct.

project-base/storefront/utils/productLists/comparison/useComparisonTable.tsx
chlebektomas: The left arrow is disabled after the initial table measurement. I also added a dedicated disabled icon color to make the state visually clearer.

project-base/storefront/components/Forms/Button/IconButton.tsx
chlebektomas: Fixed. Disabled IconButtons now use a wrapper as the tooltip trigger, while enabled buttons remain direct triggers to preserve focus and ARIA behavior. Empty ariaLabel also falls back to the title.

project-base/storefront/components/Layout/Footer/FooterExtras.tsx
chlebektomas: Fixed. Transport images now use the transport name as alt text, making it available to screen readers without adding non-interactive logos to the tab order.

project-base/storefront/components/Pages/ProductDetail/ProductDetailSections/ProductDetailStickyAction.tsx
chlebektomas: Fixed the accessible name — it now includes the product name, quantity, and unit. I’ve kept the pending state unchanged, as the double-submit scenario is very unlikely and doesn’t justify introducing global cart state.

project-base/storefront/components/Pages/ProductDetail/ProductDetailSections/ProductDetailSectionNavigation.tsx
chlebektomas: Fixed. The callback ref now removes the entry when the section button unmounts, so detached elements are not retained.

project-base/storefront/components/Blocks/BlogSignpost/BlogSignpostPanel.tsx
chlebektomas: Fixed both. The previous panel is cleared after its transition ends, and the focus target is consumed immediately after focus is moved. Added coverage for both scenarios.

project-base/storefront/components/Layout/Header/MobileMenu/MobileMenuSubItems.tsx
chlebektomas: Fixed. Quick-link accessible names now include the item count, while the visual badge remains aria-hidden to avoid duplicate announcements.

project-base/storefront/utils/ui/useScrollRestoration.ts
chlebektomas: This is already covered by the 100ms timer. If the property is set before the effect, the timer sees it; if it is set afterwards, the MutationObserver triggers the retry. No change is needed.

JanMolcik (approved):
## Approved

All findings from my previous review are resolved. I re-verified each fix in the code at `e9d9f9d` rather than going by the replies, and additionally validated the accessibility-critical ones at runtime against the review environment.

### Verified in code

| Finding | What I confirmed |
|---|---|
| `role="alert"` on consent prose | `PageHero` now takes `descriptionRole?: 'alert' \| 'status'` with **no default**; `UserConsentContent` omits it, so no `role` attribute is emitted |
| Toast focus theft | `onClose`/`restoreStoredFocus` removed from `showMessage.tsx` entirely; unused `action` plumbing deleted |
| Duplicate DOM ids | `titleId`/`listId` derive from `idPrefix`, `descriptionId` from `formMeta.formName` |
| Disabled IconButton tooltip | `disabled ? <span className="inline-flex">{button}</span> : button` — wrapper only where needed, enabled buttons stay direct triggers. Cleaner than what I proposed |
| Footer transport name | `alt={transport.name}`, without adding non-interactive logos to the tab order |
| Page-variant policy link | `variant` → `layout` refactor; wrappers now own heading + policy link, hidden coupling removed |
| Sticky add-to-cart name | `addToCartAriaLabel` computed statically, so the label is always complete |
| `buttonRefs` retention | `else { buttonRefs.current.delete(section.id) }` |
| BlogSignpost panels | `onTransitionEnd` clear + focus target consumed once |
| Quick-link counts | `aria-label={count ? \`${label} (${count})\` : label}`, badge stays `aria-hidden` |

**Vitest on this head: 157/157 files, 1358/1358 tests pass** (up from 148/1331).

### Runtime verification

Browser-driven checks against `tc-ssp-4159-expand-storefront-content.odin.shopsys.cloud`:

- **`/user-consent`** — hero description renders with `role=null`, `aria-live=null`; policy link resolves as a real link. The two `role="alert"` nodes on the page are unrelated (outdated-browser notice + empty toast container). Confirms the fix behaves correctly for assistive tech, not just in source.
- **Desktop, fixed header mounted** — zero duplicate ids among the consent/login/benefits set; scoped scheme holds.
- **Footer** — alts are `["Packeta","PPL","Czech post","Personal collection","Drone delivery"]`, no generic "Transport" fallback.
- **Product detail** — 5 visible add-to-cart buttons, 4 distinct accessible names, each carrying product + quantity + unit (e.g. `Add to cart Apple iPhone 5S 64GB, gold, quantity 1 pcs`).
- **Comparison toast** — toast appears and focus is not yanked back to a stale element after it auto-closes.
- **Mobile** — comparison/wishlist controls expose descriptive accessible names.

### Withdrawing four of my own findings

On closer verification these were my false positives, not defects — flagging so nobody "fixes" working code:

1. **Comparison toast link.** You were right: comparison is reachable from the header (`MenuIconic.tsx:64`) and mobile menu, and `useWishlist` uses the same bare-toast pattern. Consistent by design.
2. **`useScrollRestoration` ordering.** No hole — `setTimeout` is registered *after* `observe()`, and `onScrollRestore` re-reads the property on every call, so both orderings are covered.
3. **Mobile anchor offset.** I measured it: `Tag` `py-1.5` + `text-sm` line-height = 32px, plus `Webline` `py-4` = exactly **64px**, with the remaining 16px as spacing. Your number was correct.
4. **Comparison left arrow.** `useState(true)` does survive, but `shouldShowArrows` is `false` until `tableMaxMarginLeft > 0`, and the same commit that first renders the arrows batches `setMargin(0)` → `setArrowLeftActive(false)`. The enabled-but-dead state is never observable.

### One request before merge (non-blocking)

The fixups introduced a **breaking prop change that the upgrade notes don't yet mention**. `UserConsentForm` went from `variant?: 'floating' | 'page'` to `layout?: 'compact' | 'default'`, and the heading + policy link moved out of the form into the wrappers. A downstream project rendering `<UserConsentForm variant="floating" />` will silently get the default layout and lose the policy link, with no type error pointing at the cause.

Suggested addition to `upgrade-notes/storefront_20260730_134830.md`:

```markdown
- update custom `components/Blocks/UserConsent/UserConsentForm` usages
    - replace `variant="floating"` with `layout="compact"` and `variant="page"` with `layout="default"` (the default)
    - `UserConsentForm` now renders only the consent controls — the heading and the policy link are owned by the
      wrapper, so custom wrappers must render `components/Blocks/UserConsent/UserConsentPolicyLink` themselves
      (see `components/Blocks/UserConsent/UserConsent` for the floating variant and
      `components/Pages/UserConsent/UserConsentContent` for the page variant)
- pass `descriptionRole="status"` (or `"alert"`) to `components/Layout/PageHero/PageHero` only for dynamic messages;
  static descriptions should omit it so they are not announced as live regions
```

Approving now — this is a docs line, not a code change, so it doesn't need another review round.

### Two things I'm consciously leaving to you

- **Newsletter consent.** Your reasoning holds for a dedicated newsletter form. What still reads as unintentional is the asymmetry: `Pages/Contact/contactFormMeta.tsx:25` keeps `validatePrivacyPolicy`, and submitting the contact form is equally affirmative. Worth aligning one way or the other, or noting why they differ.
- **Double-submit on the sticky/hero add-to-cart pair.** Accepted as unlikely; if you ever want it closed cheaply, gating on the already-shared `isCartFetchingOrUnavailable` avoids introducing global cart state.

Nice work on the `IconButton` solution in particular — narrower and more correct than my suggestion.

pk16011990 (approved):

PR #4747, opened 2026-07-31
Add product reviews

grossmannmartin:

This PR adds product reviews to the platform.

**Customers** can rate products with 1–5 stars and an optional text, either from the product detail or via a "Write a review" link on the order detail — the link works for registered and guest customers alike and opens the review form prefilled, including the purchased variant. Reviews linked to an order in an eligible status are marked as a verified purchase. A review can be published anonymously, and every new review waits for approval before it appears publicly. The product detail shows a new Reviews section with a rating summary, ordering, gradual loading, the shop's public response, and a link to the "How we work with reviews" article. Logged-in customers see their pending reviews greyed out, cannot review the same product twice, and find all their reviews in a new "My reviews" section of their account.

**Administrators** moderate reviews in a new multi-domain Product reviews section — approving, rejecting (with a mandatory reason), or editing inappropriate content. Any edit of the customer-entered content requires a reason that is stored in the review history for auditing purposes. The order status configuration decides which orders allow reviewing, and a per-domain "Product review policy" article can be selected under CMS > Legal conditions. The whole feature is enabled per domain by configuration; on domains without reviews none of the above appears.

**Under the hood**, approved reviews are exported into the product Elasticsearch document, so the public listing, the rating summary, and the `reviewsSummary` field for product listings are read from a single source and stay consistent. The frontend API exposes queries for public reviews and the customer's own reviews, and a `CreateProductReview` mutation. The entity log newly supports notes explaining why a change was made, which is reusable beyond reviews.

#### Fixes issues
N/A

----
  - https://mg-reviews.odin.shopsys.cloud
  - https://cz.mg-reviews.odin.shopsys.cloud
  - https://mg-reviews.odin.shopsys.cloud/sk


packages/framework/src/Model/ProductReview/ProductReviewFactory.php
```diff
@@ -0,0 +1,24 @@
...
+class ProductReviewFactory
+{
+    public function __construct(
+        protected readonly EntityNameResolver $entityNameResolver,
+    ) {
+    }
+
+    public function create(ProductReviewData $productReviewData): ProductReview
+    {
+        $productReviewData->status = ProductReviewStatusEnum::STATUS_PENDING;
```
vitek-rostislav: the default property value could/should be set in the data object itself, or?

packages/framework/tests/Unit/Model/ProductReview/ProductReviewFactoryTest.php
```diff
@@ -0,0 +1,113 @@
...
+use Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver;
+use Shopsys\FrameworkBundle\Model\Product\Product;
+use Shopsys\FrameworkBundle\Model\ProductReview\ProductReviewData;
+use Shopsys\FrameworkBundle\Model\ProductReview\ProductReviewDataFactory;
+use Shopsys\FrameworkBundle\Model\ProductReview\ProductReviewFactory;
+use Shopsys\FrameworkBundle\Model\ProductReview\ProductReviewStatusEnum;
+
+final class ProductReviewFactoryTest extends TestCase
+{
+    public function testCreateSetsPendingStatusAndPreservesReviewData(): void
```
vitek-rostislav: imho this test has only a little value (and even lower if my first comment is addressed)

packages/framework/tests/Unit/Model/ProductReview/ProductReviewFactoryTest.php
```diff
@@ -0,0 +1,113 @@
...
+            $this->createProductReviewData($this->createStub(Product::class), new DateTimeImmutable()),
+        );
+        $createdAt = new DateTimeImmutable('2026-07-27 10:00:00');
+        $productReviewData = (new ProductReviewDataFactory())->createFromProductReview($productReview);
+        $productReviewData->responseText = 'Thank you.';
+        $productReviewData->responseCreatedAt = $createdAt;
+
+        $productReview->edit($productReviewData);
+        $editedProductReviewData = (new ProductReviewDataFactory())->createFromProductReview($productReview);
+        $editedProductReviewData->responseText = 'Thank you for your review.';
```
vitek-rostislav: nitpick, I like using constants (or local variables) for the values that are then used in the assertion, it helps the readability IMHO.

packages/framework/src/Model/ProductReview/ProductReviewFactory.php
vitek-rostislav: btw does not the current state limit you in data fixtures from creating reviews with different statuses? :)

project-base/app/config/packages/shopsys_framework.yaml
```diff
@@ -18,3 +18,6 @@ shopsys_framework:
             - 'Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessorMiddleware\AddRoundingMiddleware'
             - 'App\Model\Order\Processing\OrderProcessorMiddleware\SetOrderDataAsAdministratorMiddleware'
             - 'Shopsys\FrameworkBundle\Model\Order\Processing\OrderProcessorMiddleware\SetDeliveryAddressByDeliveryAddressUuidMiddleware'
+    product:
+        reviews:
+            enabled_domain_ids: [1, 3]
```
vitek-rostislav: do we take this new config into account in the CLI bootstrapper? This way projects may forget to enable the feature for their domains setup

project-base/app/src/DataFixtures/Demo/ProductReviewDataFixture.php
```diff
@@ -0,0 +1,284 @@
...
+    private function createApprovedProductReviewsWithCustomer(
+        ObjectManager $manager,
+        int $domainId,
+        CustomerUser $customerUser,
+    ): void {
+        $order = $this->getReferenceForDomain(OrderDataFixture::ORDER_DELIVERED_YESTERDAY, $domainId, Order::class);
+        $productReviewData = $this->createProductReviewDataForProductReference('1', $domainId);
+        $this->setCustomerUserData($productReviewData, $customerUser);
+        $productReviewData->rating = 5;
+        $productReviewData->text = 'The product exceeded my expectations and was delivered quickly.';
```
vitek-rostislav: should not we use translator for the texts?

packages/administration/templates/content/productReview/grid/status.html.twig
```diff
@@ -0,0 +1,7 @@
+{% if row.status is same as(constant('Shopsys\\FrameworkBundle\\Model\\ProductReview\\ProductReviewStatusEnum::STATUS_PENDING')) %}
+    <span class="badge bg-yellow-lt text-yellow-lt-fg">{{ 'Pending'|trans }}</span>
+{% elseif row.status is same as(constant('Shopsys\\FrameworkBundle\\Model\\ProductReview\\ProductReviewStatusEnum::STATUS_APPROVED')) %}
+    <span class="badge bg-green-lt text-green-lt-fg">{{ 'Approved'|trans }}</span>
+{% else %}
```
vitek-rostislav: when a new status is introduced, this will break, maybe we can put this into `elseif` to ease future extending?

packages/administration/translations/messages.cs.po
```diff
@@ -187,6 +190,9 @@ msgstr "Použitelné proměnné pro obsah"
 msgid "Applicable variables for Subject"
 msgstr "Použitelné proměnné pro předmět"
 
+msgid "Approved"
+msgstr "Schválená"
```
vitek-rostislav: "Schváleno" would be more neutral - if the translation is used elsewhere in the future

packages/administration/translations/messages.cs.po
```diff
@@ -1846,6 +1867,9 @@ msgstr "Refundovaná částka"
 msgid "Refunded amount has been changed."
 msgstr "Refundovaná částka byla změněna."
 
+msgid "Rejected"
+msgstr "Zamítnutá"
```
vitek-rostislav: maybe rather more neutral "Zamítnuto"

packages/administration/translations/messages.sk.po
```diff
@@ -187,6 +190,9 @@ msgstr "Použiteľné premenné pre obsah"
 msgid "Applicable variables for Subject"
 msgstr "Použiteľné premenné pre predmet"
 
+msgid "Approved"
+msgstr "Schválená"
```
vitek-rostislav: same as for cs translation, I would use more neutral wording

packages/administration/translations/messages.sk.po
```diff
@@ -1846,6 +1867,9 @@ msgstr "Refundovaná čiastka"
 msgid "Refunded amount has been changed."
 msgstr "Refundovaná čiastka bola zmenená."
 
+msgid "Rejected"
+msgstr "Zamietnutá"
```
vitek-rostislav: same as for cs translation, I would use more neutral wording

packages/administration/templates/content/productReview/grid/rating.html.twig
```diff
@@ -0,0 +1 @@
+{{ include('@ShopsysAdministration/content/productReview/_rating.html.twig', {rating: value}) }}
```
vitek-rostislav, line 1: nitpick, the template is now just a proxy, maybe we could inline the "_rating" template directly in the grid config

packages/administration/templates/content/productReview/grid/status.html.twig
```diff
@@ -0,0 +1 @@
+{{ include('@ShopsysAdministration/content/productReview/_statusBadge.html.twig', {status: row.status}) }}
```
vitek-rostislav, line 1: same inline note here

vitek-rostislav: ah, but I am not sure whether we can pass variables there, so...

project-base/app/src/DataFixtures/Demo/OrderStatusDataFixture.php
```diff
@@ -59,6 +59,8 @@ private function createOrderStatusReference(
                 default => throw new UnknownNameTranslationForOrderStatusReferenceNameException($referenceName),
             };
         }
+        $orderStatusData->productReviewsAllowed = $referenceName === self::ORDER_STATUS_DONE;
```
vitek-rostislav: ORDER_STATUS_DONE is a system status, so IMHO we should set the `productReviewsAllowed` for it in a migration, or?

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
```diff
@@ -534,6 +548,26 @@ protected function isProductPromoted(Product $product, int $domainId): bool
...
 
+    /**
+     * Reviews of the whole variant family live on the main variant document only.
+     * Both review fields of a document read the same list, so it is cached for the current product.
+     *
+     * @return \Shopsys\FrameworkBundle\Model\ProductReview\ProductReview[]
+     */
+    protected function getProductReviewsForExport(Product $product, int $domainId): array
+    {
+        if ($product->isVariant() || !$this->productReviewEnabledChecker->isEnabledForDomain($domainId)) {
```
vitek-rostislav, line 537: why do we return an empty array for a variant? If they add a separate page for the variant on a project, they will need to adjust the export on a project. IMHO it would be harmless to have exported the data for variants as well

packages/framework/src/Model/ProductReview/ProductReviewFacade.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    protected function dispatchReviewsExport(ProductReview $productReview): void
+    {
+        $product = $productReview->getProduct();
+
+        if ($product === null) {
+            return;
+        }
+
+        $mainProduct = $product->isVariant() ? $product->getMainVariant() : $product;
```
vitek-rostislav: just a note - if we start exporting reviews for the variants as well, this will need to be also adjusted

packages/framework/src/Model/ProductReview/ProductReviewFacade.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $productReview->edit($productReviewData);
+
+        $this->em->flush();
+
+        $this->dispatchReviewsExport($productReview);
+    }
+
+    /**
+     * The document of the main variant carries the reviews of the whole family, so its product is the one to re-export
```
vitek-rostislav: the code is self-explanatory so the comment is redundant imho

packages/framework/src/Model/ProductReview/ProductReviewFacade.php
```diff
@@ -0,0 +1,108 @@
...
+    protected function dispatchReviewsExport(ProductReview $productReview): void
+    {
+        $productId = $productReview->getProduct()?->getId();
+
+        if ($productId === null) {
+            return;
+        }
+
+        $this->productRecalculationDispatcher->dispatchProductIds(
+            [$productId],
```
vitek-rostislav, line 68: maybe this deserves the high priority queue? Or maybe have the priority as an parameter of the method so that the high priority can be set when editing the review from admin UI?

packages/frontend-api/src/Model/ProductReview/ProductReviewApiFacade.php
```diff
@@ -0,0 +1,249 @@
...
+
+        return $productArray[ProductExportFieldProvider::REVIEW_SUMMARY];
+    }
+
+    /**
+     * @return array{average_rating: float|null, total_count: int, rating_counts: array<int, array{rating: int, count: int}>}|null
+     */
+    public function getReviewSummaryForProduct(Product $product): ?array
+    {
+        if ($product->isVariant()) {
```
vitek-rostislav, line 87: I see the variant restrictions are quite deep :)

packages/frontend-api/src/Model/ProductReview/ProductReviewApiFacade.php
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Maps the entity to the very same shape the Elasticsearch export uses, so both sources share one GraphQL resolver map
+     *
+     * @return array<string, mixed>
+     */
+    public function extractReviewToPublicArray(ProductReview $productReview): array
+    {
+        $reviewArray = $this->productReviewDocumentMapper->mapReview($productReview);
+        $reviewArray['status'] = $productReview->getStatus();
```
vitek-rostislav, line 158: why is not the status mapped directly in `mapReview`?

packages/frontend-api/src/Model/Resolver/ProductReview/ProductReviewResolverMap.php
```diff
@@ -0,0 +1,42 @@
...
+            'ProductReview' => [
+                'uuid' => static fn (array $reviewArray) => $reviewArray['uuid'],
+                'reviewerName' => static fn (array $reviewArray) => $reviewArray['reviewer_name'],
+                'rating' => static fn (array $reviewArray) => $reviewArray['rating'],
+                'text' => static fn (array $reviewArray) => $reviewArray['text'],
+                'createdAt' => static fn (array $reviewArray) => new DatePoint($reviewArray['created_at']),
+                'isVerifiedPurchase' => static fn (array $reviewArray) => $reviewArray['is_verified_purchase'],
+                'responseText' => static fn (array $reviewArray) => $reviewArray['response_text'],
+                'responseCreatedAt' => static fn (array $reviewArray) => $reviewArray['response_created_at'] !== null ? new DatePoint($reviewArray['response_created_at']) : null,
+                // the exported documents carry approved reviews only, so the status is stored just for the own reviews read from the database
```
vitek-rostislav, line 37: ok, this is the answer for my earlier question regarding the status :)

packages/frontend-api/src/Resources/config/graphql-types/Mutation/MutationDecorator.types.yaml
```diff
@@ -261,6 +261,15 @@ MutationDecorator:
...
                 access: "@=isGranted('create_complaint_voter', args)"
+            CreateProductReview:
+                type: ProductReview!
+                description: "Create a new product review that will be published after moderation"
+                args:
+                    input:
+                        type: ProductReviewInput!
+                        validation: cascade
+                resolve: "@=mutation('createProductReviewMutation', args, validator)"
+                access: "@=(isGranted('ROLE_API_ALL') || not isAuthenticated())"
```
vitek-rostislav: will this stay like this forever or do we plan a separate role?

packages/framework/src/Model/ProductReview/ProductReviewPolicyFacade.php
```diff
@@ -0,0 +1,51 @@
...
+        if ($productReviewPolicyArticleId === null) {
+            return null;
+        }
+
+        return $this->articleFacade->findById($productReviewPolicyArticleId);
+    }
+
+    public function setProductReviewPolicyArticleOnDomain(?Article $productReviewPolicyArticle, int $domainId): void
+    {
+        $this->setting->setForDomain(
```
vitek-rostislav, line 39: maybe we want to add `$this->cleanStorefrontCacheFacade->cleanStorefrontGraphqlQueryCache(CleanStorefrontCacheFacade::SETTINGS_QUERY_KEY_PART);` call after the setting? (We probably want that for all special articles but :shrug:)

packages/administration/src/Controller/ProductReviewPolicyController.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $selectedDomainId = $this->adminDomainTabsFacade->getSelectedDomainId();
+
+        if (!$this->productReviewEnabledChecker->isEnabledForDomain($selectedDomainId)) {
+            return $this->render('@ShopsysAdministration/content/productReviewPolicy/disabled.html.twig');
+        }
+
+        $productReviewPolicyArticle = $this->productReviewPolicyFacade->findProductReviewPolicyArticleByDomainId($selectedDomainId);
+
+        $form = $this->createForm(
```
vitek-rostislav, line 49: just an idea to discuss and possible create a separate issue - we now have 3 form types (privacy policy, user consent policy, product review policy) that serve basically the same purpose - set a single article from a selectbox. They differ just in the label and help text. It feels like this could be generalized. There is also a 4th use-case in `TermsAndConditionsSettingFormType` but that one has the article selection just a part of bigger form...

packages/framework/src/Model/AdminNavigation/SideMenuBuilder.php
```diff
@@ -631,6 +634,13 @@ protected function createMarketingMenu(): ItemInterface
         );
         $legalMenu->addChild(static::USER_CONSENT_POLICY, ['route' => 'admin_userconsentpolicy_setting', 'label' => t('User consent policy')]);
 
+        if ($this->productReviewEnabledChecker->isEnabledOnAnyDomain()) {
+            $legalMenu->addChild(
+                static::PRODUCT_REVIEW_POLICY,
+                ['route' => 'shopsys_administration_productreviewpolicy_setting', 'label' => t('Product review policy')],
```
vitek-rostislav, line 640: the `shopsys_administration` always triggers me in the `framework` package, I was thinking about some `SideMenuSubscriber` in the proper package, but then I noticed other routes here (e.g. `admin_order_list`) that are also declared in `administration` package and I stopped dreaming. Maybe one day, far far in the future...

docs/model/log-entity-changes.md
```diff
@@ -125,6 +125,80 @@ class OrderStatus extends AbstractTranslatableEntity
...
+    {
+        $productReview->edit($productReviewData);
+        $this->entityLogNoteRegistry->registerNote($productReview, $note);
+
+        $this->em->flush();
+    }
+}
+```
+
+### How a note is matched to a record
```
vitek-rostislav: it feels like this whole section in unnecessary. TBH, I m confused with that and sure what it should tell me (but maybe it is just my problem :smile:). Maybe it just needs some simplification?

docs/model/log-entity-changes.md
```diff
@@ -125,6 +125,80 @@ class OrderStatus extends AbstractTranslatableEntity
...
+
+class ProductReviewFacade
+{
+    public function __construct(
+        protected readonly EntityManagerInterface $em,
+        protected readonly EntityLogNoteRegistry $entityLogNoteRegistry,
+    ) {
+    }
+
+    public function editWithNote(ProductReview $productReview, ProductReviewData $productReviewData, string $note): void
```
vitek-rostislav: the example differs from the real use case. I can see it is probably for the simplification but when a real class is used, it might be confusing (e.g. it makes me wonder why the note is passed separately into the method when I see it is a part of the data object...) I would maybe use some generic names or something that does not exist in the codebase. We can discuss this of course.

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
```diff
@@ -34,11 +35,15 @@ public function __construct(
         protected readonly LoggableEntityConfigFactory $loggableEntityConfigFactory,
         protected readonly ChangeSetResolver $changeSetResolver,
         protected readonly EntityLogFacade $entityLogFacade,
+        protected readonly EntityLogNoteRegistry $entityLogNoteRegistry,
     ) {
     }
 
     public function postFlush(PostFlushEventArgs $args): void
     {
+        // notes are already read into the logs at this point, so they are dropped even when the flush produced no log at all
```
vitek-rostislav: I am a bit confused here, maybe it is because of the comment? I can imagine reading the method as "post flush, we reset the notes and do other stuff". But now, when I look at the "stuff", I am confused even more, lets talk about it, please :smile:

packages/framework/src/Model/ProductReview/ProductReview.php
```diff
@@ -0,0 +1,426 @@
...
+
+        $this->responseText = $responseText;
+    }
+
+    public function edit(ProductReviewData $productReviewData): void
+    {
+        $this->setData($productReviewData);
+    }
+
+    public function isContentEditedBy(ProductReviewData $productReviewData): bool
```
vitek-rostislav: it is rather a nitpick but I am struggling with that "By" in the method name, should it be read like "edited by the data"? I would rename the method to just simple "isContentEdited" TBH :)

packages/administration/src/Model/ProductReview/ProductReviewEditHandler.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * {@inheritdoc}
+     */
+    #[Override]
+    public function edit(object $entity, object $data): void
+    {
+        Assert::isInstanceOf($entity, ProductReview::class);
+        Assert::isInstanceOf($data, ProductReviewData::class);
+
+        if ($entity->isContentEditedBy($data) && $data->contentChangeReason !== null) {
```
vitek-rostislav: maybe we could delegate this responsibility into `AbstractCrudController::editAction` or somewhere else so that the developers do not have to keep this in mind for any entity that would add such a note? Something like this? 
```php
 if ($data instanceof EntityLogNoteProvidingDataInterface && $data->getEntityLogNote() !== null) {                                                 
      $this->entityLogNoteRegistry->registerNote($entity, $data->getEntityLogNote());                                                               
  } 
```

vitek-rostislav (changes requested):
Hi, nice job there, I have added my thoughts, let me know if you need to dicsuss anything

packages/framework/src/Model/ProductReview/ProductReviewFactory.php
grossmannmartin: Good idea; this beautifully simplifies the data fixtures. 
78f2932462 and 267af8bcf6

packages/framework/tests/Unit/Model/ProductReview/ProductReviewFactoryTest.php
grossmannmartin: I thought I deleted this one already 🤔 ... well  2f93e4a521

packages/framework/tests/Unit/Model/ProductReview/ProductReviewFactoryTest.php
grossmannmartin: test is removed

project-base/app/config/packages/shopsys_framework.yaml
grossmannmartin: added in 81ad5bf622

packages/administration/templates/content/productReview/grid/status.html.twig
grossmannmartin: like this 32f5694cb5?

and this d0505349af

project-base/app/src/DataFixtures/Demo/ProductReviewDataFixture.php
grossmannmartin: May be :) 23b0557521 and 6de3808388

packages/administration/translations/messages.cs.po
grossmannmartin: Fair point - dbceaaa3ab and 4480b7d4fc

packages/administration/translations/messages.sk.po
grossmannmartin: Fair point - dbceaaa3ab and 4480b7d4fc

packages/administration/translations/messages.cs.po
grossmannmartin: f49dcaeb34 and aba3610b34

packages/administration/translations/messages.sk.po
grossmannmartin: f49dcaeb34 and aba3610b34

packages/administration/templates/content/productReview/grid/rating.html.twig
grossmannmartin: This could probably be solved by passing `value` instead of `rating` in the basicInformation.html.twig file. But the status badge is problematic due to the use of the `row` variable there... 
At this point, I've decided to keep it unified with the proxy file at both locations... We can discuss further

packages/administration/templates/content/productReview/grid/status.html.twig
grossmannmartin: comment is above

project-base/app/src/DataFixtures/Demo/OrderStatusDataFixture.php
grossmannmartin: 🤔  Can we make this assumption (= reviews are enabled on the DONE status)? Should we?

packages/framework/src/Model/ProductReview/ProductReviewFacade.php
grossmannmartin: 8736f400f9

packages/framework/src/Model/ProductReview/ProductReviewFacade.php
grossmannmartin: Set to high priority in 6cf0343771

packages/frontend-api/src/Model/ProductReview/ProductReviewApiFacade.php
grossmannmartin: You've answered this in the next comment 😅

packages/frontend-api/src/Model/Resolver/ProductReview/ProductReviewResolverMap.php
grossmannmartin: Answer for the above

packages/frontend-api/src/Resources/config/graphql-types/Mutation/MutationDecorator.types.yaml
grossmannmartin: I have rewritten and somehow simplified it in 09a665f1c1, e6278ef4c2 (fixed loading demo data with reviews enabled on 2nd domain), and 018bda96ca
I found some non-ideal behavior for B2B customer accounts

packages/framework/src/Model/Product/Elasticsearch/ProductExportRepository.php
grossmannmartin: As we spoke, it's a more common scenario to show reviews from all variants, even for the variant card.

packages/framework/src/Model/ProductReview/ProductReviewPolicyFacade.php
grossmannmartin: That's a nice catch! I have added it to all special articles in 610bad644c

packages/administration/src/Controller/ProductReviewPolicyController.php
grossmannmartin: Created with the internal id SSP-4278

docs/model/log-entity-changes.md
grossmannmartin: That's a fair point. I have removed unnecessary info 925d5810e1

docs/model/log-entity-changes.md
grossmannmartin: Good point. Thanks to this, I realized that it's not ideal to register the note in the handler. I may want to edit and transfer reviews from the external system. So the appropriate place to check for the note is the `ProductReviewFacade::edit()` method. 

Changed in ebf65408a1

packages/framework/src/Model/ProductReview/ProductReview.php
grossmannmartin: Changed in https://github.com/shopsys/shopsys/commit/ebf65408a17665b749c2b52767ec78b7fdfc1683

packages/administration/src/Model/ProductReview/ProductReviewEditHandler.php
grossmannmartin: I'm not sure if I want to generalize it like this. We're starting to add some number of interfaces that somehow behave automatically. This may not be, in general, a bad thing. But we may end up with many interfaces that nobody will remember what they do.
And on top of that, I think the proper place for registering the note is the facade, so it's usable even for the transfered fields

packages/framework/src/Component/EntityLog/EventListener/EntityLogEventListener.php
grossmannmartin: I've tried to reword it in 8065aa0bdb

packages/administration/templates/content/productReview/grid/rating.html.twig
vitek-rostislav: I can live with that :slightly_smiling_face:

project-base/app/src/DataFixtures/Demo/OrderStatusDataFixture.php
vitek-rostislav: hm good question. I would say yes, and let's add the information to the upgrade notes (something like "migration XY sets the `product_reviews_allowed` column to `true` for the `DONE` order status, adjust the behavior in the administration or via a migration if you have different needs". The thing is, I assume most of the projects will be ok with this default behavior, so it should be one less thing to set up after the upgrading :shrug:

project-base/app/tests/FrontendApiBundle/Functional/ProductReview/CreateProductReviewMutationTest.php
```diff
@@ -0,0 +1,214 @@
...
+        $this->assertFalse($data['isVerifiedPurchase']);
+
+        $this->assertNull($this->getProductReviewByUuid($data['uuid'])->getOrderItem());
+    }
+
+    /**
+     * The purchase is proven by the order the url hash belongs to, so the reviewer may pick any of its products,
+     * for example another variant than the one the review was opened for
+     */
+    public function testGuestReviewOfAnotherProductOfTheSameOrderIsVerified(): void
```
vitek-rostislav: I do not understand this test, can we talk about it, please? It basically does the same as `testGuestReviewLinkedToOrderItemByUrlHashIsVerified`, or?

docs/model/log-entity-changes.md
vitek-rostislav: Yeah, that feels better in the code now :ok_hand: 
I would still adjust the docs here - we can now easily align them with the implementation, at least in the simplified form, what do you think?

```
public function edit(ProductReview $productReview, ProductReviewData $productReviewData): void
{
    $this->entityLogNoteRegistry->registerNote($productReview, $productReviewData->contentChangeReason);

    $productReview->edit($productReviewData);
    $this->em->flush();
}
```

packages/administration/src/Model/ProductReview/ProductReviewEditHandler.php
vitek-rostislav: I think moving to the facade really made it better, so ok :slightly_smiling_face:

packages/frontend-api/src/Model/ProductReview/ProductReviewApiFacade.php
```diff
@@ -0,0 +1,249 @@
...
+        $input = $argument['input'];
+
+        $product = $this->productRepository->getOneByUuid($input['productUuid']);
+
+        if ($product->isMainVariant()) {
+            throw new ProductReviewVariantRequiredUserError('A concrete variant of the product has to be reviewed.');
+        }
+
+        if ($customerUser !== null
+            && $this->productReviewFacade->existsByCustomerUserAndProductId($customerUser, $product->getId(), $this->domain->getId())
```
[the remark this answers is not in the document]
grossmannmartin: This is intentional. A review is always tied to a concrete variant — the customer reviews the exact variant they bought, and someone who buys two different variants of the same family is entitled to review both. The guard exists to stop the same customer re-reviewing the same variant, not to cap them at one review per family, so the per-variant key matches the intent. The family-wide aggregation in `getApprovedByMainProductForExport()` is deliberate as well — the family rating is meant to include every variant's reviews. Flooding stays gated by the `STATUS_PENDING` moderation and the `customerUserId` shown in the admin grid.

project-base/app/src/DataFixtures/Demo/OrderStatusDataFixture.php
grossmannmartin: Agreed, let's make it the default. Since `Version20260730120000` (the migration adding the column) is part of this PR and unreleased, I extended it directly instead of adding a second migration — it now runs `UPDATE order_statuses SET product_reviews_allowed = TRUE WHERE type = 'done'`. The data fixture line is gone, the migration is the single source of truth now.

Upgrade notes added as well, describing the default and how to change it (administration or an own migration).

https://github.com/shopsys/shopsys/commit/1a0137ebd730d0ea14bdae185dfd7ac127e39ccd
69071496de

docs/model/log-entity-changes.md
grossmannmartin: Good idea, docs aligned with the implementation. One deviation from your snippet: `registerNote()` takes a non-nullable `string` and `$contentChangeReason` is `?string`, so I kept the null guard from the real code — otherwise the documented example would not type-check.

```php
public function edit(ProductReview $productReview, ProductReviewData $productReviewData): void
{
    if ($productReviewData->contentChangeReason !== null) {
        $this->entityLogNoteRegistry->registerNote($productReview, $productReviewData->contentChangeReason);
    }

    $productReview->edit($productReviewData);

    $this->em->flush();
}
```

https://github.com/shopsys/shopsys/commit/044cc4813ccd5fc3435b7d2045e4e640ceb33476

project-base/app/tests/FrontendApiBundle/Functional/ProductReview/CreateProductReviewMutationTest.php
grossmannmartin: You are right, it is a duplicate. Products 15 and 18 are both ordinary items of the same `ORDER_DELIVERED_MONTH_AGO` order, neither of them is a variant, so the test asserts exactly what `testGuestReviewLinkedToOrderItemByUrlHashIsVerified()` already asserts — only the product number differs. The variant wording in the docblock is not exercised by anything, and the case it describes (a sibling variant that is not in the order) resolves to unverified, which `testGuestReviewOfProductMissingInTheOrderIsUnverified()` already covers.

Removed it.

https://github.com/shopsys/shopsys/commit/86b0ce54943f28accc732b891dbd2e3548e30d25

packages/framework/src/Model/ProductReview/ProductReview.php
```diff
@@ -0,0 +1,428 @@
...
+    #[AsMcpColumn]
+    #[ORM\Column(type: 'string', length: 100)]
+    protected $lastName;
+
+    /**
+     * @var string
+     */
+    #[AsMcpColumn]
+    #[ORM\Column(type: 'string', length: 255)]
+    protected $email;
```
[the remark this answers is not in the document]
grossmannmartin: @pk16011990  What is your opinion on this?

vitek-rostislav (approved):

packages/administration/src/Controller/ProductReviewController.php
```diff
@@ -0,0 +1,150 @@
...
+
+    #[Override]
+    public function configure(CrudConfig $config): void
+    {
+        $enabledDomainIds = $this->productReviewEnabledChecker->getEnabledDomainIds();
+
+        $config
+            ->setEntityNameSingular(t('Product review'))
+            ->setEntityNamePlural(t('Product reviews'))
+            ->setMenuTitle(t('Product reviews'))
```
henzigo: You don't need to call those setters.

project-base/storefront/components/Blocks/ProductReviews/ProductReviewsContent.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    const handleWriteReview = () => {
+        void openCreateProductReviewPopup({
+            productName,
+            productUuid: unreviewedVariants?.length ? null : productUuid,
+            variants: unreviewedVariants,
+        });
+    };
+
+    const remainingCount = Math.min(Math.max(totalCount - reviews.length, 0), PRODUCT_REVIEWS_LOAD_MORE_PAGE_SIZE);
```
fujta: We can use clamp function from utils

project-base/storefront/components/Blocks/ProductReviews/useCurrentCustomerUserProductFamilyReviews.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Reviews the logged in customer wrote for any product of the given product family
+ * (the backend resolves the uuid to the whole main variant family).
+ */
+export const useCurrentCustomerUserProductFamilyReviews = (
+    productUuid: string,
+): CurrentCustomerUserProductFamilyReviews => {
+    const [{ data: currentCustomerUserData, fetching: isCurrentCustomerUserFetching }] = useCurrentCustomerUserQuery();
+    const isUserLoggedIn = !!currentCustomerUserData?.currentCustomerUser;
+
+    const [{ data, fetching: areOwnReviewsFetching }] = useCurrentCustomerUserProductReviewsQuery({
```
fujta: I would use `data: currentCustomerUserProductReviewsData`

project-base/storefront/components/Blocks/ProductReviews/useCurrentCustomerUserReviewedProductUuids.ts
```diff
@@ -0,0 +1,30 @@
...
+const OWN_REVIEWED_PRODUCTS_LIMIT = 100;
+
+/**
+ * Uuids of products the logged in customer has already reviewed (regardless of the review status).
+ * Serves as a UX hint only — the duplicate review rule is enforced by the CreateProductReview mutation.
+ */
+export const useCurrentCustomerUserReviewedProductUuids = (): Set<string> => {
+    const isUserLoggedIn = useIsUserLoggedIn();
+
+    const [{ data }] = useCurrentCustomerUserProductReviewsQuery({
```
fujta: Here as well

fujta (dismissed):

packages/frontend-api/src/Model/ProductReview/ProductReviewApiFacade.php
```diff
@@ -0,0 +1,227 @@
...
+        $productReviewData = $this->productReviewDataFactory->create();
+        $productReviewData->domainId = $this->domain->getId();
+        $productReviewData->product = $product;
+        $productReviewData->catnum = $product->getCatnum();
+        $productReviewData->productName = (string)$product->getName($this->domain->getLocale());
+        $productReviewData->orderItem = $orderItem;
+        $productReviewData->customerUser = $customerUser;
+        $productReviewData->firstName = $input['firstName'];
+        $productReviewData->lastName = $input['lastName'];
+        $productReviewData->email = $customerUser?->getEmail() ?? $input['email'];
```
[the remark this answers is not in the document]
grossmannmartin: That's deliberate

PR #4752, opened 2026-08-05
Remove unused Heureka widget setting and fix disabling the API key

vitek-rostislav:

The Heureka widget code setting was removed from the administration. The widget code was only stored and edited, but never read anywhere — no template, storefront, or API ever used it, so administrators were maintaining a value with no effect. A migration cleans up the stored values.

This PR also fixes disabling the Heureka "Verified by Customers" integration. The integration is considered active whenever an API key is set for the domain, but clearing the field in the administration saved an empty string instead of removing the key. The integration therefore stayed active and kept sending requests with an invalid key. Now, clearing the field removes the key and the integration is properly deactivated.

#### Fixes issues

N/A

----
  - https://rv-heureka-cleanup.odin.shopsys.cloud
  - https://cz.rv-heureka-cleanup.odin.shopsys.cloud
  - https://rv-heureka-cleanup.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Controller/Admin/HeurekaController.php
+++ b/packages/framework/src/Controller/Admin/HeurekaController.php
@@ -41,7 +41,6 @@ class HeurekaController extends AdminBaseController
         if ($this->heurekaFacade->isDomainLocaleSupported($locale)) {
             $heurekaShopCertificationData = [
                 'heurekaApiKey' => $this->heurekaSetting->getApiKeyByDomainId($domainId),
-                'heurekaWidgetCode' => $this->heurekaSetting->getHeurekaShopCertificationWidgetByDomainId($domainId),
             ];
 
             $form = $this->createForm(HeurekaShopCertificationFormType::class, $heurekaShopCertificationData, [
@@ -53,10 +52,6 @@ class HeurekaController extends AdminBaseController
                 $heurekaShopCertificationData = $form->getData();
 
                 $this->heurekaSetting->setApiKeyForDomain($heurekaShopCertificationData['heurekaApiKey'] ?? '', $domainId);
-                $this->heurekaSetting->setHeurekaShopCertificationWidgetForDomain(
-                    $heurekaShopCertificationData['heurekaWidgetCode'] ?? '',
-                    $domainId,
-                );
 
                 $this->addSuccessFlash(t('Settings modified.'));
             }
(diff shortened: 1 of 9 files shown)


packages/framework/src/Model/Order/Messenger/HeurekaPlacedOrderMessageHandler.php
```diff
@@ -34,7 +36,24 @@ public function __invoke(PlacedOrderMessage $placedOrderMessage): void
...
+            }
+
             throw $exception;
         }
     }
+
+    protected function isCausedByClientError(RequesterException $exception): bool
+    {
+        // the library does not expose the HTTP status code, it is only part of the message, see \Heureka\ShopCertification\CurlRequester::request()
+        if (preg_match('~^Request resulted in HTTP code \'(\d+)\'~', $exception->getMessage(), $matches) !== 1) {
```
grossmannmartin: I'm not sure about this. What about 400 (bad request), 404 (not found), 429 (Too many requests)? Those should be re-delivered, imho. 

Also, I'm not a particular fan of regular expression matching. A simple typo fix or slight formatting change will break this code without even knowing it.

I'm ok with flooding. This means something is misconfigured and should be handled somehow.

packages/framework/src/Migrations/Version20160902145842.php
```diff
@@ -1,24 +0,0 @@
...
-
-declare(strict_types=1);
-
-namespace Shopsys\FrameworkBundle\Migrations;
-
-use Doctrine\DBAL\Schema\Schema;
-use Override;
-use Shopsys\MigrationBundle\Component\Doctrine\Migrations\AbstractMigration;
-
-final class Version20160902145842 extends AbstractMigration implements DomainAwareInterface
```
grossmannmartin, line 11: I'm not sure about deleting existing migrations. This particular case may be harmless. 
But if any future migration depends on the state after this migration is executed, the migration will fail. It's not an easy task to draw the line where it's safe to delete a migration and where not, imho. 

TLDR: I would keep the migration.

grossmannmartin (changes requested):

packages/framework/src/Model/Order/Messenger/HeurekaPlacedOrderMessageHandler.php
vitek-rostislav: fair point, I have dropped the commit entirely

packages/framework/src/Migrations/Version20160902145842.php
vitek-rostislav: true, I agree, the migration is kept now

vitek-rostislav:
HI @grossmannmartin, I have addressed your objections and rebased the branch (one commit was dropped, one new commit is there now regarding the API key storing in admin).

grossmannmartin (approved):

PR #4777, opened 2026-08-17
Protect CSV exports against formula injection

fujta:

CSV exports in the administration are now protected against formula injection. Values starting with `=`, `-`, `+`, `@`, tab, or CR are prefixed with an apostrophe, so spreadsheet software opens them as plain text instead of evaluating them as formulas. Until now, anyone could subscribe to the newsletter on the storefront with an address like `=1+1@evil.com` (`=` is a valid character in the local part of an email) and it was evaluated as a formula on the administrator's computer when the export was opened in a spreadsheet editor.

The escaping lives in the new `CsvFormulaEscaper` component, whose `escape()` method is meant for any user-influenced value written into a CSV export. The newsletter export uses it directly, CSV responses are newly created by the `CsvResponseFactory`, which escapes all exported values including the column names, and the price list import removes the escaping apostrophe again, so a re-imported price list export matches the original products. Along the way, `CsvResponse` newly respects the explicitly passed CSV headers, which were previously dropped because of an inverted condition. Upgrade notes describe how projects adopt the change.

#### Fixes issues

...

----
  - https://vk-ssp-4193-csv-formula-injection.odin.shopsys.cloud
  - https://cz.vk-ssp-4193-csv-formula-injection.odin.shopsys.cloud
  - https://vk-ssp-4193-csv-formula-injection.odin.shopsys.cloud/sk

----
  - https://vk-ssp-4193-csv-formula-injection.odin.shopsys.cloud
  - https://cz.vk-ssp-4193-csv-formula-injection.odin.shopsys.cloud
  - https://vk-ssp-4193-csv-formula-injection.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/HttpFoundation/CsvResponse.php
+++ b/packages/framework/src/Component/HttpFoundation/CsvResponse.php
@@ -16,7 +16,7 @@ class CsvResponse extends Response
     public function __construct(array $data, string $fileName, ?array $csvHeaders = null)
     {
         $csvEncoder = new CsvEncoder();
-        $context = [];
+        $context = [CsvEncoder::ESCAPE_FORMULAS_KEY => true];
 
         if ($csvHeaders !== null) {
             $context[CsvEncoder::HEADERS_KEY] = $csvHeaders;
--- a/packages/framework/src/Component/HttpFoundation/CsvResponse.php
+++ b/packages/framework/src/Component/HttpFoundation/CsvResponse.php
@@ -9,13 +9,17 @@ use Symfony\Component\Serializer\Encoder\CsvEncoder;
 
 class CsvResponse extends Response
 {
+    /**
+     * @param array<int, array<string, mixed>> $data
+     * @param array<array-key, string>|null $csvHeaders
+     */
     public function __construct(array $data, string $fileName, ?array $csvHeaders = null)
     {
         $csvEncoder = new CsvEncoder();
         $context = [];
 
-        if ($csvHeaders === null) {
-            $context = [CsvEncoder::HEADERS_KEY => $csvHeaders];
+        if ($csvHeaders !== null) {
+            $context[CsvEncoder::HEADERS_KEY] = $csvHeaders;
         }
 
         $content = $csvEncoder->encode($data, CsvEncoder::FORMAT, $context);
--- a/packages/framework/src/Model/PriceList/PriceListFacade.php
+++ b/packages/framework/src/Model/PriceList/PriceListFacade.php
@@ -7,6 +7,7 @@ namespace Shopsys\FrameworkBundle\Model\PriceList;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\QueryBuilder;
 use League\Flysystem\MountManager;
+use Shopsys\FrameworkBundle\Component\Csv\CsvHelper;
 use Shopsys\FrameworkBundle\Component\FileUpload\FileUpload;
 use Shopsys\FrameworkBundle\Component\Money\Money;
 use Shopsys\FrameworkBundle\Component\Translation\Translator;
@@ -264,8 +265,18 @@ class PriceListFacade
         return ',';
     }
 
+    /**
+     * @param array<string, mixed> $row
+     * @return array<string, mixed>
+     */
     protected function preProcessCsvRow(array $row): array
     {
+        foreach ($row as $columnName => $value) {
+            if (is_string($value)) {
+                $row[$columnName] = CsvHelper::unescapeFormula($value);
+            }
+        }
+
         if (!array_key_exists(PriceListCsvColumnsEnum::PRICE, $row)) {
             return $row;
         }
--- a/packages/framework/tests/Unit/Model/PriceList/PriceListFacadeTest.php
+++ b/packages/framework/tests/Unit/Model/PriceList/PriceListFacadeTest.php
@@ -1,0 +1,82 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\FrameworkBundle\Unit\Model\PriceList;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+use Shopsys\FrameworkBundle\Component\Csv\CsvHelper;
+use Symfony\Component\Serializer\Encoder\CsvEncoder;
+
+class PriceListFacadeTest extends TestCase
+{
+    /**
+     * The list of characters CsvEncoder escapes by is private and may change with any Symfony release,
+     * so this test derives the escaping from the encoder itself instead of repeating the list
+     */
+    #[DataProvider('exportedValueDataProvider')]
+    public function testUnescapeFormulaIsInverseToCsvEncoderEscaping(string $value): void
+    {
+        $this->assertSame($value, CsvHelper::unescapeFormula($this->exportValue($value)));
+    }
+
+    /**
+     * @return iterable<string, array{value: string}>
+     */
+    public static function exportedValueDataProvider(): iterable
+    {
+        foreach (range(1, 127) as $characterCode) {
+            yield sprintf('value starting with the character 0x%02X', $characterCode) => [
+                'value' => chr($characterCode) . '500A',
+            ];
+        }
+
+        yield 'formula hidden behind a leading space' => [
+            'value' => ' =1+1',
+        ];
+
+        yield 'formula hidden behind a leading non-breaking space' => [
+            'value' => "\u{00A0}=1+1",
+        ];
+
+        yield 'formula hidden behind a leading zero-width no-break space' => [
+            'value' => "\u{FEFF}=1+1",
+        ];
+
+        yield 'common value' => [
+            'value' => '500A',
+        ];
+
+        yield 'empty value' => [
+            'value' => '',
+        ];
+    }
+
+    /**
+     * A cell holding an apostrophe followed by a formula trigger character cannot be told apart from an escaped
+     * value, so it is read as escaped — a value that really starts with an apostrophe loses it on import
+     */
+    public function testApostropheIsRemovedFromValueThatCannotBeToldApartFromAnEscapedOne(): void
+    {
+        $this->assertSame('=1+1', CsvHelper::unescapeFormula("'=1+1"));
+    }
+
+    /**
+     * Returns the value as the CSV export writes it into the file
+     */
+    private function exportValue(string $value): string
+    {
+        $csvEncoder = new CsvEncoder();
+        $data = [['value' => $value]];
+        $context = [CsvEncoder::NO_HEADERS_KEY => true];
+
+        $escapedCsv = $csvEncoder->encode($data, CsvEncoder::FORMAT, $context + [CsvEncoder::ESCAPE_FORMULAS_KEY => true]);
+
+        if ($escapedCsv === $csvEncoder->encode($data, CsvEncoder::FORMAT, $context)) {
+            return $value;
+        }
+
+        return "'" . $value;
+    }
+}
--- a/upgrade-notes/backend_20260814_120000.md
+++ b/upgrade-notes/backend_20260814_120000.md
@@ -1,0 +1,10 @@
+#### Protect CSV exports against formula injection ([#4777](https://github.com/shopsys/shopsys/pull/4777))
+
+- `Shopsys\FrameworkBundle\Component\HttpFoundation\CsvResponse` now prefixes exported values starting with `=`, `-`, `+`, `@`, tab, CR or LF with an apostrophe
+    - if you re-import such an export, strip the apostrophe from the imported values with `Shopsys\FrameworkBundle\Component\Csv\CsvHelper::unescapeFormula()`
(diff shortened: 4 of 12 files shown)


packages/framework/src/Component/HttpFoundation/CsvResponse.php
```diff
@@ -4,25 +4,51 @@
...
+        $content = $csvEncoder->encode($this->escapeFormulasInRows($data), CsvEncoder::FORMAT, $context);
 
         parent::__construct($content);
 
         $this->headers->set('Content-Type', 'text/csv');
         $this->headers->set('Content-Disposition', 'attachment; filename="' . $fileName . '"');
     }
+
+    /**
+     * Escapes values so they cannot be evaluated as formulas by spreadsheet software, header names are left intact
```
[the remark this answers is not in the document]
fujta: Headers are left unescaped intentionally — the import matches columns by exact name (`areColumnsValid()`), so escaping them would break the export→import round trip. The only caller passes enum constants, and the upgrade notes now tell callers not to pass user-influenced values as headers or row keys.

packages/framework/src/Component/Csv/CsvFormulaEscaper.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        if ($value === '') {
+            return false;
+        }
+
+        if ($value[0] === "'" || in_array($value[0], static::FORMULA_TRIGGER_CHARACTERS, true)) {
+            return true;
+        }
+
+        $trimmedValue = ltrim($value, static::WHITESPACE_CHARACTERS);
```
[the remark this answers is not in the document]
fujta: Fixed in 867c2f18721b8acf072b5412966f54b194aa50c3 — the leading-whitespace check now strips Unicode whitespace too via a `preg_replace('~^[\s\p{Z}\x{FEFF}]+~u', '', $value)`, with a fallback to the ASCII-only `ltrim()` when the value is not valid UTF-8 (preg_replace returns null there). Added test cases for NBSP- and U+FEFF-hidden triggers.

packages/framework/src/Component/Csv/CsvFormulaEscaper.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        if ($value === '') {
+            return false;
+        }
+
+        if ($value[0] === "'" || in_array($value[0], static::FORMULA_TRIGGER_CHARACTERS, true)) {
+            return true;
+        }
+
+        $trimmedValue = preg_replace('~^[\s\p{Z}\x{FEFF}]+~u', '', $value);
```
vitek-rostislav: IMHO it is worth extracting into some helper, e.g. into the existing `TransformStringHelper`, or at least into a protected method

packages/framework/src/Component/HttpFoundation/CsvResponse.php
```diff
@@ -4,25 +4,75 @@
...
+     * @param array<int, array<string, mixed>> $data
+     * @param array<array-key, string>|null $csvHeaders
+     */
+    public function __construct(
+        array $data,
+        string $fileName,
+        ?array $csvHeaders = null,
+        ?CsvFormulaEscaper $csvFormulaEscaper = null,
+    ) {
+        $this->csvFormulaEscaper = $csvFormulaEscaper ?? new CsvFormulaEscaper();
```
vitek-rostislav: This is not a good practice. Lets introduce `CsvResponseFactory`, wire the `CsvFormulaEscaper` there so we do not have to deal with nullable value and direct creation of the escaper, and let's delegate all the new logic into the factory. Let's keep the response object just a simple value object  as it used to be before these changes :)

packages/framework/src/Component/HttpFoundation/CsvResponse.php
```diff
@@ -4,25 +4,75 @@
...
+     * @param array<int, array<string, mixed>> $data
+     * @return array<int, array<string, mixed>>
+     */
+    protected function escapeFormulasInRows(array $data): array
+    {
+        foreach ($data as $rowIndex => $row) {
+            $escapedRow = [];
+
+            foreach ($row as $columnKey => $value) {
+                $escapedRow[$this->csvFormulaEscaper->escape((string)$columnKey)] = $this->escapeFormulasInValue($value);
```
vitek-rostislav: based on the @param annotation, the `(string)` is not necessary

vitek-rostislav: btw please update the PR description that states the headers stay intact, but here we can see they are escaped as well - or do I understand it wrong?

packages/framework/src/Model/PriceList/PriceListFacade.php
```diff
@@ -199,6 +202,25 @@ public function importPriceList(
...
+     * @param array<int, array<string, mixed>> $data
+     * @return array<int, array<string, mixed>>
+     */
+    protected function unescapeColumnNames(array $data): array
+    {
+        foreach ($data as $rowIndex => $row) {
+            $unescapedRow = [];
+
+            foreach ($row as $columnKey => $value) {
+                $unescapedRow[$this->csvFormulaEscaper->unescape((string)$columnKey)] = $value;
```
vitek-rostislav: same here, the string cast should not be necessary (if the param annotation is correct)

packages/framework/tests/Unit/Model/PriceList/PriceListFacadeTest.php
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * @param array<string, string> $row
+     * @param array<string, string> $expectedRow
+     */
+    #[DataProvider('preProcessCsvRowDataProvider')]
+    public function testPreProcessCsvRow(array $row, array $expectedRow): void
+    {
+        $priceListFacade = $this->createPriceListFacade();
+
+        $reflectionMethod = new ReflectionMethod(PriceListFacade::class, 'preProcessCsvRow');
```
vitek-rostislav: The reflection usage here points us to the architecture itself. We should slighly re-think it. I suggest moving putting `escapeRow()`, `unescapeRow()` and `unescapeColumnNames()` int the `CsvFormulaEscaper` class. That way, you will have public api to test, and moreover, you will be to re-use these methods in `PriceListFacade` and the new factory I suggested in the earlier comment

vitek-rostislav: you could then completely remove this test and just add more cases into `CsvFormulaEscaperTest`

upgrade-notes/backend_20260814_120000.md
```diff
@@ -0,0 +1,7 @@
+#### Protect CSV exports against formula injection ([#4777](https://github.com/shopsys/shopsys/pull/4777))
+
+- new `\Shopsys\FrameworkBundle\Component\Csv\CsvFormulaEscaper` prefixes values whose first non-whitespace character is `=`, `-`, `+`, `@`, tab, or CR (and values starting with an apostrophe) with an apostrophe so spreadsheet software does not evaluate them as formulas — use its `escape()` method for any user-influenced value you write into a CSV export
+- newsletter CSV export now escapes subscriber emails — `NewsletterController` constructor has a new `$csvFormulaEscaper` parameter, if you extend the controller, add the parameter to your constructor
```
vitek-rostislav: the constructor change is unnecessary to mention in the upgrade notes (check the upgrade notes skill)

upgrade-notes/backend_20260814_120000.md
```diff
@@ -0,0 +1,7 @@
...
+- `CsvResponse` now escapes all string, `Stringable`, and nested array values including header names, and newly respects the `$csvHeaders` constructor parameter that was previously ignored because of an inverted condition — if you pass custom headers to `CsvResponse`, verify your export output
+    - `CsvResponse` constructor has a new optional `$csvFormulaEscaper` parameter — pass your `CsvFormulaEscaper` service when you have one at hand, otherwise the response creates a default instance; `PriceListController` constructor has a new `$csvFormulaEscaper` parameter, if you extend the controller, add the parameter to your constructor
```
vitek-rostislav: this will not be true after you create the factory :) + again unnecessary constructor change mention

upgrade-notes/backend_20260814_120000.md
```diff
@@ -0,0 +1,7 @@
...
+- price list import now removes the escaping apostrophe from imported values and column names (see `CsvFormulaEscaper::unescape()`), so a re-imported price list export matches the original products — `PriceListFacade` constructor has a new `$csvFormulaEscaper` parameter, if you extend the facade, add the parameter to your constructor
```
vitek-rostislav: Unnecesary. Check the skill, the upgrading notes should not describe the feature itself (the PR description serves that purpose), neither list changes that are easily spoted by static analysis. Just actionable instructions for the project developers :)

vitek-rostislav (changes requested):

packages/framework/src/Component/Csv/CsvFormulaEscaper.php
fujta: Separate helper is imo overkill. I created protected method

packages/framework/src/Component/HttpFoundation/CsvResponse.php
fujta: Good idea. I was wondering how to do it to get rid of manual instantiation :D

packages/framework/src/Component/Csv/CsvFormulaEscaper.php
```diff
@@ -0,0 +1,123 @@
...
+
+declare(strict_types=1);
+
+namespace Shopsys\FrameworkBundle\Component\Csv;
+
+use Stringable;
+
+class CsvFormulaEscaper
+{
+    protected const array FORMULA_TRIGGER_CHARACTERS = ['=', '-', '+', '@', "\t", "\r"];
```
grossmannmartin: This is an overall comment; I just want to make it "resolvable" so it's on a random line. 

This solution has a few drawbacks in general.

1. If you try to create the CSV file with "numbered" headers, like
```
1;2;3
a;b;c
d;e;f
```

the export fails with error 500 due to strict typing

2. I would rather reuse Symfony's CSV encoder `\Symfony\Component\Serializer\Encoder\CsvEncoder` instead of writing the encoder

3. A factory for a Response object is not a common pattern in the SSP, nor in Symfony itself. So it's easily missable, making the CSV vulnerable to injection again. 


I would recommend the following:

- CsvResponse keeps its current constructor signature and simply adds CsvEncoder::ESCAPE_FORMULAS_KEY => true to the context of the CsvEncoder it already instantiates. Anyone writing new CsvResponse($data, $file) gets escaping automatically.

- New `StreamedCsvResponse extends StreamedResponse` that accepts an iterable in the constructor (we pass the doctrine iterable in the newsletterController). Basically, all the heavy lifting from the NewsletterController would be moved into the new StreamedCsvResponse class. 
The class itself calls the Symfony encoder (per row – it's a deliberate drawback for the ease of usage). 

- CsvFormulaEscaper and CsvResponseFactory classes are no longer needed.

- PriceList import handles the unescape by itself, or a small helper is created

grossmannmartin (changes requested):

packages/framework/src/Component/Csv/CsvFormulaEscaper.php
fujta: Thanks, this makes sense to me — I'll rework it that way.

The numbered headers case is a real bug, good catch. `escapeRow()` gets `int` keys from PHP (`'1'` becomes `1`) and `escape(string $value)` then blows up on strict types. It goes away with the rework anyway.

One thing I'd like to agree on before I start. Symfony's encoder only looks at the first byte of a value (`\in_array(substr((string) $value, 0, 1), self::FORMULAS_START_CHARACTERS, true)` in `flatten()`) and never touches the keys, so we lose the whitespace/NBSP/BOM-hidden triggers I added in the earlier rounds, plus the header escaping — `" =1+1@evil.com"` would be exported as is. Are we ok with that? Personally I'd take the simpler code over that edge case, I just don't want it to happen silently.

The good news is their trigger set also covers `\n`, which ours doesn't, and they prefix with `'` too, so the price list round trip stays exactly as it is now.

_(edited — I first wrote they prefix with `\t`, that's wrong, it's `'`.)_

grossmannmartin: I'm ok with that. AFAIK, a whitespace-starting formula is not executed in Excel, but I cannot reliably verify this claim. 
All in all, yes we're ok with that

grossmannmartin (approved):

packages/framework/src/Model/PriceList/PriceListFacade.php
```diff
@@ -275,6 +288,18 @@ protected function preProcessCsvRow(array $row): array
         return $row;
     }
 
+    /**
+     * Removes the apostrophe that the CSV export prepends to values that spreadsheet software would evaluate as formulas
+     */
+    protected function unescapeFormula(string $value): string
```
pk16011990: In my opinion, it is not the responsibility of this class.

vitek-rostislav (approved):

PR #4778, opened 2026-08-17
Protect entity log change set against XSS via translation parameters

fujta:

The entity log in the administration composes the "from X to Y" change messages through the translator, and until now the whole change set — including the raw, unescaped old and new values entered by customers — was passed as translation parameters. A project translation that referenced `oldValue`/`newValue` instead of the readable placeholders (a realistic translator mistake, and project translations override the framework ones) would substitute unescaped customer input into the change set, which is rendered as raw HTML in the order detail. A payload stored e.g. in the order note would then execute in the administrator's session.

The change-set formatters now build the message through the new `AbstractChangeSetFormatter::formatFromToChanges()`, which passes only the two escaped readable values to the translator, so no translation — however broken — can pull raw customer input into the output. This is the load-bearing fix: with no raw values left in the parameter bag, there is nothing unescaped for a translation to substitute. The behavior is covered by a unit test that simulates such a broken project translation. Upgrade notes describe the only visible change for projects.

On top of that, the placeholders are now properly delimited as `%oldReadableValue%`/`%newReadableValue%` following the Symfony convention. This is a correctness/defense-in-depth improvement, not the security boundary on its own — delimiting prevents a placeholder from accidentally matching a substring inside a translated word, but it does not remove raw values from the parameter bag, so it would not close the hole without the escaped-values change above.

#### Fixes issues

...

----
  - https://vk-ssp-4115-entity-log-escaping.odin.shopsys.cloud
  - https://cz.vk-ssp-4115-entity-log-escaping.odin.shopsys.cloud
  - https://vk-ssp-4115-entity-log-escaping.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/EntityLog/ChangeSet/Formatter/AbstractChangeSetFormatter.php
+++ b/packages/framework/src/Component/EntityLog/ChangeSet/Formatter/AbstractChangeSetFormatter.php
@@ -17,9 +17,6 @@ abstract class AbstractChangeSetFormatter
     /**
      * Only escaped values may be passed to the translator - the result is rendered as raw HTML
      * and a translation referencing raw change values would inject unescaped customer input
-     *
-     * @param mixed $oldReadableValue
-     * @param mixed $newReadableValue
      */
     protected function formatFromToChanges(mixed $oldReadableValue, mixed $newReadableValue): string
     {
--- a/packages/framework/src/Component/EntityLog/ChangeSet/Formatter/AbstractChangeSetFormatter.php
+++ b/packages/framework/src/Component/EntityLog/ChangeSet/Formatter/AbstractChangeSetFormatter.php
@@ -20,9 +20,9 @@ abstract class AbstractChangeSetFormatter
      */
     protected function formatFromToChanges(mixed $oldReadableValue, mixed $newReadableValue): string
     {
-        return t('from oldReadableValue to newReadableValue', [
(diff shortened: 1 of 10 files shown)


packages/framework/src/Component/EntityLog/ChangeSet/Formatter/AbstractChangeSetFormatter.php
```diff
@@ -13,4 +13,19 @@ protected function formatCode(mixed $value): string
...
+    /**
+     * Only escaped values may be passed to the translator - the result is rendered as raw HTML
+     * and a translation referencing raw change values would inject unescaped customer input
+     *
+     * @param mixed $oldReadableValue
+     * @param mixed $newReadableValue
+     */
+    protected function formatFromToChanges(mixed $oldReadableValue, mixed $newReadableValue): string
+    {
+        return t('from oldReadableValue to newReadableValue', [
```
vitek-rostislav: I know it is a bit out of scope, but I would use proper placeholders here and dump the translations so it is obvious `oldReadableValue` and `newReadableValue` are really placeholders. If you agree, we can add it as a separate commit

So I suggest this (tests will need update as well):
```
        return t('from %oldReadableValue% to %newReadableValue%', [
```

vitek-rostislav (changes requested):

vitek-rostislav (approved):

pk16011990 (approved):

PR #4785, opened 2026-08-19
Rate limit password recovery requests by IP and e-mail

fujta:

Until now, the password recovery form on the storefront could be submitted without any limit. Anyone could use it to flood a chosen customer's inbox with reset e-mails, or to run through a list of addresses and find out which of them belong to a registered customer, because the form reveals whether the account exists.

This PR limits how often password recovery can be requested. Each visitor's connection gets 40 attempts per hour and each e-mail address gets 10 attempts per hour, counted across both steps of the flow — asking for the reset link and setting the new password. The limits are intentionally generous so a real customer who mistypes their address a few times is never blocked, while automated abuse is stopped. When a limit is reached, the customer sees a clear "too many attempts, try again later" message instead of a generic error, and the counters keep working across deployments so a release does not reset them.

#### Fixes issues

N/A

----
  - https://vk-ssp-4194-password-recovery-rate-limit.odin.shopsys.cloud
  - https://cz.vk-ssp-4194-password-recovery-rate-limit.odin.shopsys.cloud
  - https://vk-ssp-4194-password-recovery-rate-limit.odin.shopsys.cloud/sk

----
  - https://vk-ssp-4194-password-recovery-rate-limit.odin.shopsys.cloud
  - https://cz.vk-ssp-4194-password-recovery-rate-limit.odin.shopsys.cloud
  - https://vk-ssp-4194-password-recovery-rate-limit.odin.shopsys.cloud/sk


packages/frontend-api/src/Model/Mutation/Customer/User/CustomerUserPasswordRecoveryMutation.php
```diff
@@ -0,0 +1,90 @@
...
+
+        return $this->loginMutation->loginMutation($newArgument);
+    }
+
+    protected function checkPasswordRecoveryRateLimit(string $email): void
+    {
+        $clientIp = $this->requestStack->getCurrentRequest()?->getClientIp() ?? 'unknown';
+
+        $this->consumeRateLimit($this->passwordRecoveryIpRateLimiter, $clientIp);
+        $this->consumeRateLimit($this->passwordRecoveryEmailRateLimiter, $this->getEmailRateLimitKeyPart($email));
```
fujta, line 76: **[Medium · CWE-645] Targeted lockout of account recovery** — the per-e-mail budget is spent by *whoever* names the address, and exhausting it makes the endpoint reject the request outright. 10 unauthenticated `RequestPasswordRecovery` calls for `victim@example.com` (well inside the 40/hour IP budget, so a single IP locks out 4 accounts per hour, trivially parallelised across IPs) leave the real owner with `too-many-password-recovery-attempts` for the rest of the sliding window; repeating it hourly keeps them locked out indefinitely. Confidence: High.

The commit message is right that a composite `ip+email` key on its own is weak — but it is the *complement* here, not the replacement. The harm the low e-mail ceiling is really defending against is an attacker rotating `resetPasswordHash` and invalidating the link the victim just received. Address that directly instead:

1. In `CustomerUserPasswordFacade::resetPassword()`, reuse the existing hash while `resetPasswordHashValidThrough` is still in the future instead of generating a new one — attacker spam then cannot invalidate the victim's link.
2. With (1) in place the per-e-mail counter only has to cap mail volume to that address, so its ceiling can be raised well above what a legitimate user needs.
3. Hard-fail on per-IP and per-`(ip, email)` limiters — neither can be spent against a third party's request coming from a different IP.

packages/frontend-api/src/Model/Mutation/Customer/User/CustomerUserPasswordRecoveryMutation.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->checkPasswordRecoveryRateLimit($email);
+
+        $validator->validate();
+
+        $this->customerUserPasswordFacade->resetPassword($email, $this->domain->getId());
+
+        return 'success';
+    }
+
+    public function recoverPasswordMutation(Argument $argument, InputValidator $validator): LoginResultData
```
fujta, line 45: **[Low · CWE-770] `RecoverPassword` stays unthrottled** — `requestPasswordRecoveryMutation()` is now metered, but the mutation that *consumes* the hash is not. `setNewPassword()` distinguishes "no such customer" (`CustomerUserNotFoundByEmailAndDomainException`) from "wrong hash" (`InvalidResetPasswordHashUserException`), so this stays an unauthenticated, unmetered account-existence oracle and a free DB-lookup + password-hashing endpoint. Confidence: Medium.

Guessing the hash itself is *not* the concern — 50 characters from `random_int` over a 62-character alphabet, compared with `hash_equals` and expiring after 48 h, is out of reach.

Fix: consume `passwordRecoveryIpRateLimiter` (or a dedicated limiter) at the top of `recoverPasswordMutation()` as well, so failed hash attempts cost the caller something.

fujta:
## 🔒 AI security audit

_Advisory review of the PR diff against OWASP Top 10 / CWE — not a merge gate._

**Findings:** Critical 0 · High 0 · Medium 2 · Low 3 — line-specific ones are inline on the diff. **One previously reported Low is now fixed.**

Re-run over the diff as of `461a549e`. The change remains sound. The e-mail bucket is normalised (`mb_strtolower(trim(...))`) exactly the way `CustomerUserRepository::findCustomerUserByEmailAndDomain()` lowercases before lookup, so one account cannot be reached through two buckets; the key is SHA-256 hashed, so the pre-validation `consume()` cannot bloat Redis with 255-character keys; the two limiters live under distinct limiter ids, so they do not collide in the shared cache pool; aliasing the mutation N times in one GraphQL document consumes N tokens, so batching is no bypass; and leaving `%build-version%` out of the Redis prefix is the right call, confirmed against `RedisFacade::getCacheClients()`. No injection, authz, secret-handling or XSS issue is introduced.

### Fixed since the last run

- ✅ **[Low · CWE-770] `RecoverPassword` was unthrottled** — resolved by `00565bce`. `recoverPasswordMutation()` now consumes the IP limiter at the very top, before `$validator->validate()`, so the `ExistingEmail` oracle on `RecoverPasswordInput` is metered too and a made-up hash can no longer be used to sidestep the budget added for the request mutation. Leaving the e-mail limiter out of this mutation is the right trade-off — `RecoverPassword` sends no mail, and spending a victim's budget there would only widen the window in which their own recovery is blocked. The inline comment on that line is stale; treat it as closed.

### Still open (inline on the diff)

- **[Medium · CWE-645] Targeted lockout of account recovery** — `CustomerUserPasswordRecoveryMutation.php:76`. The per-e-mail budget is spent by whoever names the address, so 10 unauthenticated calls for `victim@example.com` leave the real owner rejected for the rest of the sliding window. Acknowledged in the PR description as a deliberate trade-off, mitigated by keeping the limit generous. Confidence: High.
- **[Medium · CWE-362] Both new sliding-window limiters run without a lock** — `rate_limiter.yaml:12`. `framework.rate_limiter.*.lock_factory` defaults to `auto`, which resolves to `lock.factory` only when `framework.lock` is enabled; `symfony/lock` is installed but never configured — re-verified this run: there is no `config/packages/lock.yaml`, no `framework.lock` key anywhere under `config/packages/`, and no `lock_factory` on any limiter. `SlidingWindowLimiter::consume()` is then a plain read-modify-write over `CacheStorage`, so a concurrent burst — precisely the traffic this PR exists to stop — consumes roughly one token instead of N. The sequential functional tests cannot catch it. Pre-existing for the other limiters in this file, but it directly weakens the two being added here. Confidence: Medium.

### Not tied to a single line

- **[Low · CWE-204] The endpoint remains an account-existence oracle.** `ExistingEmail` returns `USER_WITH_EMAIL_DOES_NOT_EXIST_ERROR` for unknown addresses on both mutations, so customers can still be enumerated — now at 40/hour per IP rather than unbounded, which is a clear improvement. Pre-existing and explicitly deferred in the PR description pending a product decision. Confidence: High on the behaviour.
- **[Low · CWE-16] The IP limiter is only as trustworthy as `TRUSTED_PROXIES`.** The upgrade notes cover the too-narrow direction (every customer collapses into the proxy's single 40/hour bucket). The other direction is worth a line too: a project that widens it to `REMOTE_ADDR` or `*` makes `getClientIp()` return an attacker-supplied `X-Forwarded-For`, and the IP limiter can then be bypassed with a fresh header per request, leaving only the per-e-mail budget. The repo default (`TRUSTED_PROXIES=127.0.0.1`) is fine. Confidence: Medium (depends on project deployment config).
- **[Low · CWE-203] The e-mail budget is a side channel on someone else's reset activity.** The limiter is keyed by e-mail alone and is deliberately never reset on success, so an unauthenticated attacker can count how many `RequestPasswordRecovery` calls for `victim@example.com` are accepted before `too-many-password-recovery-attempts` and learn how many resets that address requested in the last hour. It leaks activity, not existence (the token is spent pre-validation for any address), and the probe costs the attacker the same budget. Low impact; noted for completeness. Confidence: High on the mechanism.

### Checked and cleared

- **Recovery links are single-use.** `CustomerUserPasswordFacade::setNewPassword()` → `changePassword()` → `setPassword()` → `CustomerUser::setPasswordHash()`, which nulls `resetPasswordHash` and `resetPasswordHashValidThrough`. So the newly throttled `RecoverPassword` is not sitting on a replay vector, and the hash itself is not a realistic takeover path: `HashGenerator::generateHash(50)` draws 50 characters from a 62-character alphabet via `random_int()`, `isResetPasswordHashValid()` compares with `hash_equals()`, and the hash expires after 48 hours.
- Redis keys carry the limiter's own expiry via `CacheStorage`, so the pre-validation `consume()` cannot grow the keyspace without bound even though the client is now excluded from `shopsys:redis:clean-cache`.
- The test-env override to `cache.adapter.filesystem` is scoped to `config/packages/test/`, and the new Redis client uses `%env(REDIS_HOST)%` — no hardcoded credential anywhere in the diff.
- Storefront changes introduce no client-side secret exposure and no unescaped rendering. Dropping `customMessage` does not surface raw server messages: `getUserFriendlyErrors()` maps flash-message codes through the local translation table and everything unrecognised to the local “Unknown error.” string.

### Baseline

The built-in `/security-review` **ran successfully this time** over the same diff and returned **no findings** at its confidence bar. Everything above comes from the Shopsys-specific checklist, which deliberately covers what that pass excludes (rate limiting, enumeration, business logic).

project-base/app/src/FrontendApi/Mutation/Customer/User/CustomerUserPasswordRecoveryMutation.php
```diff
@@ -4,57 +4,13 @@
...
-
-class CustomerUserPasswordRecoveryMutation extends AbstractMutation
+use Shopsys\FrontendApiBundle\Model\Mutation\Customer\User\CustomerUserPasswordRecoveryMutation as BaseCustomerUserPasswordRecoveryMutation;
+
+/**
+ * @property \App\Model\Customer\User\CustomerUserPasswordFacade $customerUserPasswordFacade
+ * @property \App\FrontendApi\Mutation\Login\LoginMutation $loginMutation
+ * @method __construct(\App\Model\Customer\User\CustomerUserPasswordFacade $customerUserPasswordFacade, \Shopsys\FrameworkBundle\Component\Domain\Domain $domain, \App\FrontendApi\Mutation\Login\LoginMutation $loginMutation, \Overblog\GraphQLBundle\Definition\ArgumentFactory $argumentFactory, \Symfony\Component\RateLimiter\RateLimiterFactoryInterface $passwordRecoveryIpRateLimiter, \Symfony\Component\RateLimiter\RateLimiterFactoryInterface $passwordRecoveryEmailRateLimiter, \Symfony\Component\HttpFoundation\RequestStack $requestStack)
+ */
+class CustomerUserPasswordRecoveryMutation extends BaseCustomerUserPasswordRecoveryMutation
```
grossmannmartin, line 14: Nitpick:
It's not necessary to keep this empty extended mutation

grossmannmartin (approved):

PR #4788, opened 2026-08-20
Elasticsearch profiler now displays bulk (NDJSON) request bodies

grossmannmartin:

The Elasticsearch panel in the Symfony profiler was unusable for bulk requests. Bulk endpoints such as `_msearch` send their body as NDJSON (multiple JSON objects separated by newlines), which the tracer failed to parse as a single JSON document — so the profiler showed no request data at all for exactly the queries developers most often need to debug. The tracer now recognizes multiline request bodies and formats each NDJSON line separately, so bulk request bodies are displayed pretty-printed and stay usable for copy-pasting into Kibana.

Additionally, the Response and Json toggle buttons in the panel were swapped — each opened the other's data block. They now toggle their matching blocks, and the buttons are smaller so all three fit the table row.

#### Fixes issues
N/A

----
  - https://mg-profiler-es-fix.odin.shopsys.cloud
  - https://cz.mg-profiler-es-fix.odin.shopsys.cloud
  - https://mg-profiler-es-fix.odin.shopsys.cloud/sk

----
  - https://mg-profiler-es-fix.odin.shopsys.cloud
  - https://cz.mg-profiler-es-fix.odin.shopsys.cloud
  - https://mg-profiler-es-fix.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
+++ b/packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
@@ -4,8 +4,8 @@ declare(strict_types=1);
 
 namespace Shopsys\FrameworkBundle\Component\Elasticsearch\Debug;
 
-use InvalidArgumentException;
 use Nette\Utils\Json;
+use Nette\Utils\JsonException;
 use Override;
 use Psr\Log\AbstractLogger;
 use Psr\Log\LogLevel;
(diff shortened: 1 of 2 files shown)


packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
```diff
@@ -23,20 +23,40 @@ public function __construct(protected readonly ElasticsearchRequestCollection $e
...
-        return Json::decode($matches['json'], true);
+        return trim($matches['json']);
     }
 
-    protected function formatData(mixed $requestData): string
+    protected function extractData(string $requestBody): mixed
     {
-        return Json::encode($requestData, JSON_PRETTY_PRINT);
+        // bulk APIs (e.g. _msearch) send NDJSON - multiple JSON objects separated by newlines
+        if (str_contains($requestBody, "\n")) {
```
TomasLudvik, line 40: Is it not possible, that this will be triggered even for random newline in the data?

TomasLudvik (changes requested):

packages/framework/src/Component/Elasticsearch/Debug/ElasticsearchTracer.php
grossmannmartin: It can't, actually – the traced body is produced by the `SmartSerializer` class called in `Connection` class. JSON escapes newlines in strings, so a literal newline byte can only be the NDJSON separator of a bulk request.

TomasLudvik (approved):

PR #4794, opened 2026-08-21
Improve custom actions and custom routes in CRUD controllers

grossmannmartin:

Custom actions and custom routes in CRUD controllers were documented, but several gaps made them hard (or impossible) to use in practice. This PR closes them, so a developer adding a quick action (for example an "Approve" button on an edit page) can follow the actions reference and it just works:

- **Top actions on the edit page now receive the edited entity.** The `displayIf()` and `linkToRoute()` closures documented in the actions reference were always called with `null`, so the documented entity-typed closures failed. Now the edit page passes the edited entity, allowing actions to be shown conditionally and to build URLs with the entity id.
- **`linkToRoute()` now appends the CSRF token for protected routes.** An action linking to a route guarded by `#[CsrfProtection]` generated a URL without the token, so clicking it always failed with a 400 error. The token is now appended automatically, resolved from the cached route access-control data.
- **Permission attributes on custom routes no longer need an explicit role.** A plain `#[CanEdit]` now guards the route with the same role the built-in CRUD actions use — no more hand-written role constants like `#[CanEdit('ROLE_CRUD_PRODUCT_REVIEW')]`.
- **A custom role of a CRUD controller is now declared with the class-level `#[ForRole]` attribute**, replacing the removed `CrudConfig::setCustomRoleConstant()`. On a CRUD controller, `#[ForRole]` governs the whole controller consistently: the built-in actions, custom routes, and the administrator role matrix all follow one role, so they can no longer diverge. The role is resolved at compile time from class attributes (a CRUD controller extension can override it), so building the admin access control no longer instantiates controllers and runs their `configure()`.

The actions reference and admin rights documentation are extended accordingly, and an upgrade note covers the removed `setCustomRoleConstant()` method and the extended `#[ForRole]` semantics.

#### Fixes issues

N/A

----
  - https://mg-crud-edit-page-actions.odin.shopsys.cloud
  - https://cz.mg-crud-edit-page-actions.odin.shopsys.cloud
  - https://mg-crud-edit-page-actions.odin.shopsys.cloud/sk

----
  - https://mg-crud-edit-page-actions.odin.shopsys.cloud
  - https://cz.mg-crud-edit-page-actions.odin.shopsys.cloud
  - https://mg-crud-edit-page-actions.odin.shopsys.cloud/sk

### DIFF
--- a/docs/administration/crud-controller/reference/actions.md
+++ b/docs/administration/crud-controller/reference/actions.md
@@ -85,6 +85,13 @@ protected function configureActions(ActionsConfig $actions): void
 
     You can implement your own reusable actions by extending the `Shopsys\AdministrationBundle\Component\Action\AbstractAction` class.
 
+### What the closures receive
+
+The `displayIf()` and `linkToRoute()` closures receive the data of the page the action is displayed on:
+
+- `ActionType::EDIT` — the edited entity
+- `ActionType::LIST`, `ActionType::CREATE`, and `ActionType::DETAIL` — `null` (there is no single entity on these pages; for per-row actions in the list grid use row actions instead — their closures receive the data of the grid row they are rendered in)
+
 ### Configuration
 
 ```php
--- a/docs/administration/crud-controller/reference/actions.md
+++ b/docs/administration/crud-controller/reference/actions.md
@@ -96,6 +96,12 @@ The `displayIf()` and `linkToRoute()` closures receive the data of the page the 
 
     When the route linked via `linkToRoute()` is protected by the `#[CsrfProtection]` attribute, the CSRF token is appended to the generated URL automatically (unless you provide it yourself in the parameters).
 
+!!! note
+
+    Permission attributes (`#[CanView]`, `#[CanEdit]`, ...) on custom routes defined in a Crud Controller do not need an explicit role 
+    — they fall back to the controller's role constant (the generated one, or the one declared by the class-level `#[ForRole]` attribute), the same role the built-in CRUD actions use.
+    You only need to pass a role explicitly when the route should be guarded by a different role.
+
 ### Configuration
 
 ```php
--- a/docs/administration/crud-controller/reference/actions.md
+++ b/docs/administration/crud-controller/reference/actions.md
@@ -92,6 +92,10 @@ The `displayIf()` and `linkToRoute()` closures receive the data of the page the 
 - `ActionType::EDIT` — the edited entity
 - `ActionType::LIST`, `ActionType::CREATE`, and `ActionType::DETAIL` — `null` (there is no single entity on these pages; for per-row actions in the list grid use row actions instead — their closures receive the data of the grid row they are rendered in)
 
+!!! note
+
+    When the route linked via `linkToRoute()` is protected by the `#[CsrfProtection]` attribute, the CSRF token is appended to the generated URL automatically (unless you provide it yourself in the parameters).
+
 ### Configuration
 
 ```php
--- a/packages/administration/src/Component/Security/Attribute/AttributeProcessor.php
+++ b/packages/administration/src/Component/Security/Attribute/AttributeProcessor.php
@@ -7,7 +7,10 @@ namespace Shopsys\AdministrationBundle\Component\Security\Attribute;
 use InvalidArgumentException;
 use ReflectionClass;
 use ReflectionMethod;
+use Shopsys\AdministrationBundle\Component\Crud\CrudRoleConstantProvider;
+use Shopsys\AdministrationBundle\Component\Crud\Helper\CrudTransformationHelper;
 use Shopsys\AdministrationBundle\Component\Security\AccessControl\AccessControlRuleFactory;
+use Shopsys\AdministrationBundle\Controller\AbstractCrudController;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\CanCreate;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\CanDelete;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\CanEdit;
@@ -27,6 +30,7 @@ final class AttributeProcessor
 {
     public function __construct(
         private readonly AccessControlRuleFactory $accessControlRuleFactory,
+        private readonly CrudRoleConstantProvider $crudRoleConstantProvider,
     ) {
     }
 
@@ -75,8 +79,13 @@ final class AttributeProcessor
             $rules[] = $this->accessControlRuleFactory->create($roleWithPermission, $attribute->getMethods());
         }
 
-        // Get class-level role if ForRole attribute is present
-        $classRole = $this->getClassAttribute($class, ForRole::class)?->role;
+        // CRUD controllers use their role constant (generated, or set by ForRole — resolved at compile time including extension overrides) so custom routes are guarded by the same role as the built-in actions, other classes read the ForRole attribute directly
+        $classRole = $class->isSubclassOf(AbstractCrudController::class)
+            ? CrudTransformationHelper::generateRoleConstant(
+                $class->getShortName(),
+                $this->crudRoleConstantProvider->findCustomRoleConstant($class->getName()),
+            )
+            : $this->getClassAttribute($class, ForRole::class)?->role;
 
         // Process CRUD attributes
         $this->processPermissionAttributes($method, $rules, CanView::class, $classRole);
--- a/packages/administration/src/Controller/BlogArticleAuthorController.php
+++ b/packages/administration/src/Controller/BlogArticleAuthorController.php
@@ -15,6 +15,7 @@ use Shopsys\FrameworkBundle\Component\Grid\DataSourceInterface;
 use Shopsys\FrameworkBundle\Component\Grid\Grid;
 use Shopsys\FrameworkBundle\Component\Grid\GridFactory;
 use Shopsys\FrameworkBundle\Component\Grid\QueryBuilderDataSourceFactory;
+use Shopsys\FrameworkBundle\Component\Security\Attribute\ForRole;
 use Shopsys\FrameworkBundle\Component\Security\Role\AdminRoleConstant;
 use Shopsys\FrameworkBundle\Form\Admin\Blog\BlogArticleAuthorFormType;
 use Shopsys\FrameworkBundle\Model\AdminNavigation\SideMenuBuilder;
@@ -23,6 +24,7 @@ use Shopsys\FrameworkBundle\Model\Blog\Author\BlogArticleAuthor;
 use Shopsys\FrameworkBundle\Model\Localization\Localization;
 
 #[CrudController(BlogArticleAuthor::class)]
+#[ForRole(AdminRoleConstant::ROLE_BLOG_ARTICLE_AUTHOR)]
 class BlogArticleAuthorController extends AbstractCrudController
 {
     protected const int ARTICLES_GRID_DEFAULT_LIMIT = 10;
@@ -40,8 +42,7 @@ class BlogArticleAuthorController extends AbstractCrudController
     {
         $config
             ->registerHandler(BlogArticleAuthorCrudHandler::class)
-            ->setMenuSection(SideMenuBuilder::SECTION_BLOG)
-            ->setCustomRoleConstant(AdminRoleConstant::ROLE_BLOG_ARTICLE_AUTHOR);
+            ->setMenuSection(SideMenuBuilder::SECTION_BLOG);
     }
 
     #[Override]
--- a/packages/administration/src/DependencyInjection/Compiler/ResolveCrudRoleConstantsCompilerPass.php
+++ b/packages/administration/src/DependencyInjection/Compiler/ResolveCrudRoleConstantsCompilerPass.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shopsys\AdministrationBundle\DependencyInjection\Compiler;
+
+use Override;
+use ReflectionClass;
+use Shopsys\AdministrationBundle\Component\Crud\CrudControllerRegistry;
+use Shopsys\AdministrationBundle\Component\Crud\CrudRoleConstantProvider;
+use Shopsys\FrameworkBundle\Component\Security\Attribute\ForRole;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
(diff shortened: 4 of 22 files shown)


packages/administration/src/Twig/ActionExtension.php
```diff
@@ -65,12 +65,31 @@ protected function generateActionUrl(?ActionRouteInterface $actionRoute, mixed $
...
+
+            return $this->router->generate($actionRoute->getRouteName(), $parameters);
         }
 
         throw new InvalidArgumentException('Action has invalid route type');
     }
 
+    protected function isRouteCsrfProtected(string $routeName): bool
+    {
+        $controller = $this->router->getRouteCollection()->get($routeName)?->getDefault('_controller');
```
henzigo: `AccessControlDataProviderInterface` provides `findRouteByName` method that returns `RouteAccessControlData` object with all the data that you need 👍 

Also, you don't need to parse the controller, and that data is cached

packages/administration/src/Twig/ActionExtension.php
```diff
@@ -65,12 +65,31 @@ protected function generateActionUrl(?ActionRouteInterface $actionRoute, mixed $
...
+                $parameters[RouteCsrfProtector::CSRF_TOKEN_REQUEST_PARAMETER] ??= $this->csrfProtector->getCsrfTokenByRoute($actionRoute->getRouteName());
+            }
+
+            return $this->router->generate($actionRoute->getRouteName(), $parameters);
         }
 
         throw new InvalidArgumentException('Action has invalid route type');
     }
 
+    protected function isRouteCsrfProtected(string $routeName): bool
```
henzigo: I think this method should be moved directly into the `RouteCsrfProtector` class.

packages/administration/src/Component/Security/Attribute/AttributeProcessor.php
```diff
@@ -75,8 +77,9 @@ public function processMethod(ReflectionClass $class, ReflectionMethod $method):
             $rules[] = $this->accessControlRuleFactory->create($roleWithPermission, $attribute->getMethods());
         }
 
-        // Get class-level role if ForRole attribute is present
-        $classRole = $this->getClassAttribute($class, ForRole::class)?->role;
+        // Get class-level role if ForRole attribute is present, CRUD controllers fall back to their (generated or custom) role constant
+        $classRole = $this->getClassAttribute($class, ForRole::class)?->role
+            ?? $this->crudControllerRegistry->findRoleConstantByControllerClass($class->getName());
```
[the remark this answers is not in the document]
henzigo: Copilot is right, but I would prefer to go a totally different way, as we discussed. I already sent you the proposal on how custom actions could be fully integrated with CrudController.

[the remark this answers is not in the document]
grossmannmartin: I walked a different path entirely. The proposed solution has a few drawbacks in my opinion. 

1. It leaves the normal Symfony actions in some void state. They potentially accept #[ForRole] attribute, while default crud actions accept the `setCustomRoleConstant()`
2. Creates a subset of default Symfony routing and each necessary change (conditional routing, default values, etc.)
3. It's still string-coupled linkToCrudAction with the Action itself, just in a custom-made way.

Instead, I've removed the `setCustomRoleConstant()` method entirely, and for this purpose, the already established `#[ForRole]` attribute is now used. Extensions can re-declare the attribute with a custom role if necessary, and the priority is taken into account as before.

packages/administration/src/Twig/ActionExtension.php
grossmannmartin: fixed

packages/administration/src/Twig/ActionExtension.php
grossmannmartin: fixed

packages/administration/src/Component/Crud/CrudAdminRoleProvider.php
```diff
@@ -39,7 +39,7 @@ public function configureRoles(RoleCollection $roleCollection): void
         foreach ($this->crudControllerRegistry->getAll() as $crudControllerDefinition) {
             $config = $crudControllerDefinition->config;
 
-            if ($config->isFullDisabled() || $config->getCustomRoleConstant() !== null) {
+            if ($config->isFullDisabled() || $crudControllerDefinition->customRoleConstant !== null) {
```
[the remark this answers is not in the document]
grossmannmartin: Yes, it's a deliberate simplification of the whole concept. Previously, setting #[ForRole] gave a false impression that I'm setting the role for all actions, while only custom actions were covered.

docs/administration/crud-controller/reference/actions.md
```diff
@@ -85,6 +85,13 @@ protected function configureActions(ActionsConfig $actions): void
 
     You can implement your own reusable actions by extending the `Shopsys\AdministrationBundle\Component\Action\AbstractAction` class.
 
+### What the closures receive
+
+The `displayIf()` and `linkToRoute()` closures receive the data of the page the action is displayed on:
+
+- `ActionType::EDIT` — the edited entity
+- `ActionType::LIST`, `ActionType::CREATE`, and `ActionType::DETAIL` — `null` (there is no single entity on these pages; for per-row actions in the list grid use row actions instead — their closures receive the data of the grid row they are rendered in)
```
vitek-rostislav, line 93: TBH, I would expect the entity to be passed to the `detail` action as well, but the detail action is not implemented completely yet, right?

packages/administration/src/Component/Crud/Helper/CustomRoleConstantHelper.php
```diff
@@ -0,0 +1,26 @@
...
+
+final class CustomRoleConstantHelper
+{
+    /**
+     * Returns the custom role constant of a CRUD controller (or CRUD controller extension) class,
+     * declared by the class-level ForRole attribute, or null when the attribute is not present.
+     */
+    public static function findCustomRoleConstant(ReflectionClass $reflectionClass): ?string
+    {
+        $attributes = $reflectionClass->getAttributes(ForRole::class);
```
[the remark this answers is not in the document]
vitek-rostislav: interesting, do we need that? CRUD controllers are extended via extensions, but if they are not declared as `final`, developers still might want/need to extend them on a project, right? Do we support that?

packages/administration/src/Controller/BlogArticleAuthorController.php
```diff
@@ -23,6 +24,7 @@
 use Shopsys\FrameworkBundle\Model\Localization\Localization;
 
 #[CrudController(BlogArticleAuthor::class)]
+#[ForRole(AdminRoleConstant::ROLE_BLOG_ARTICLE_AUTHOR)]
```
vitek-rostislav, line 27: do we even need a custom role for this controller? I did some research and it seems like we could get rid of the custom constant completely - we can use `$this->definition->getRoleConstant()` within the controller itself, and `CrudControllerRegistry::findRoleConstantByControllerClass(BlogArticleAuthorController::class)` from outside. The only problem? `BlogArticleAuthorPickerController` is placed in the framework package. It would be nice if we moved it to the administration package and leveraged the generic CRUD powers, but I understand if you say this is out of scope of this task.

packages/administration/tests/Unit/DependencyInjection/Compiler/InitializeControllersCompilerPassTest.php
```diff
@@ -0,0 +1,74 @@
...
+        $this->assertSame([
+            [
+                'class' => CompilerPassFixtureCrudController::class,
+                'entityClass' => stdClass::class,
+                'customRoleConstant' => null,
+            ],
+            [
+                'class' => CompilerPassRoleFixtureCrudController::class,
+                'entityClass' => stdClass::class,
+                'customRoleConstant' => 'ROLE_COMPILER_PASS_FIXTURE',
```
vitek-rostislav: extracting the constant name into a constant would help the test readability a bit

packages/administration/tests/Unit/DependencyInjection/Compiler/LoadControllersExtensionCompilerPassTest.php
```diff
@@ -0,0 +1,74 @@
...
+                'extensionClass' => CompilerPassFixtureExtension::class,
+                'controllerClass' => CompilerPassExtendedCrudController::class,
+                'priority' => 0,
+                'customRoleConstant' => null,
+            ],
+            [
+                'extensionClass' => CompilerPassRoleFixtureExtension::class,
+                'controllerClass' => CompilerPassExtendedCrudController::class,
+                'priority' => 100,
+                'customRoleConstant' => 'ROLE_EXTENSION_FIXTURE',
```
vitek-rostislav: extracting the constant name into a constant would help the test readability a bit

packages/administration/tests/Unit/Component/Crud/CrudControllerRegistryTest.php
```diff
@@ -0,0 +1,128 @@
...
+
+namespace Tests\AdministrationBundle\Unit\Component\Crud;
+
+use PHPUnit\Framework\TestCase;
+use Shopsys\AdministrationBundle\Component\Crud\CrudControllerRegistry;
+use Shopsys\FrameworkBundle\Component\EntityExtension\EntityNameResolver;
+use stdClass;
+use Symfony\Component\DependencyInjection\ServiceLocator;
+
+final class CrudControllerRegistryTest extends TestCase
```
vitek-rostislav: extracting the constant names into constants would help the test readability a bit

vitek-rostislav (changes requested):

packages/administration/src/Component/Crud/Helper/CustomRoleConstantHelper.php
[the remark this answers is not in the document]
grossmannmartin: I don't think we want to support this. The proper way of extending it is via an extension. The CrudController attribute is not inherited either, imho.

docs/administration/crud-controller/reference/actions.md
grossmannmartin: Exactly, the detail action is deliberately excluded as it's not implemented properly and it requires at least partial implementation.

packages/administration/src/Controller/BlogArticleAuthorController.php
grossmannmartin: I don't mind the current solution as it's defined with the same logic as before.

packages/administration/tests/Unit/DependencyInjection/Compiler/InitializeControllersCompilerPassTest.php
grossmannmartin: fixed in https://github.com/shopsys/shopsys/commit/a6464651d78b2a95987b40b31ff1a8be5899e21f

packages/administration/tests/Unit/DependencyInjection/Compiler/LoadControllersExtensionCompilerPassTest.php
grossmannmartin: fixed in https://github.com/shopsys/shopsys/commit/a6464651d78b2a95987b40b31ff1a8be5899e21f

packages/administration/tests/Unit/Component/Crud/CrudControllerRegistryTest.php
grossmannmartin: fixed in https://github.com/shopsys/shopsys/commit/a6464651d78b2a95987b40b31ff1a8be5899e21f

packages/administration/src/Controller/BlogArticleAuthorController.php
vitek-rostislav: I mind it a bit as it gives us a misleading precedence/example for the future, but I will try to address this by an adhoc pr some time in the future :slightly_smiling_face:

vitek-rostislav (approved):

pk16011990 (approved):

henzigo (approved):

packages/administration/src/DependencyInjection/Compiler/ResolveCrudRoleConstantsCompilerPass.php
```diff
@@ -0,0 +1,77 @@
...
+            $reflectionClass = new ReflectionClass($controllerClass);
+            $customRoleConstant = $this->findForRole($reflectionClass);
+
+            // an extension may override the controller's role — the one whose configure() runs last (highest priority) wins
+            foreach ($extensionClassesByController[$controllerClass] ?? [] as $extensionClass) {
+                $customRoleConstant = $this->findForRole(new ReflectionClass($extensionClass)) ?? $customRoleConstant;
+            }
+
+            $roleConstants[$controllerClass] = [
+                'roleConstant' => CrudTransformationHelper::generateRoleConstant($reflectionClass->getShortName(), $customRoleConstant),
```
vitek-rostislav: The `roleConstant` key is derived data: it is always `customRoleConstant ?? generateRoleConstant(shortName)`, and `Definition::getRoleConstant()` / `CrudRegistryItem::getRoleConstant()` already compute exactly that from the config. The only thing this pass really resolves is the `ForRole` value (controller + extension override), so I would store just that and simplify the rest:

- **ResolveCrudRoleConstantsCompilerPass** – store only the `ForRole` value per controller (`class-string => ?string`), drop the `roleConstant` key.
- **CrudRoleConstantProvider** – remove `findRoleConstant()` entirely, keep the single `findCustomRoleConstant()` (both methods read the same compiled array, one accessor is enough).
- **AttributeProcessor** – the `findRoleConstant(...) ?? getClassAttribute(ForRole)` chain hides a "CRUD vs. other class" branch (for a registered CRUD controller the provider never returns `null`, so the fallback fires only for non-CRUD classes). Make it explicit and compute the effective role the same way the rest of the bundle does:
  ```php
  if ($class->isSubclassOf(AbstractCrudController::class)) {
      $classRole = CrudTransformationHelper::generateRoleConstant(
          $class->getShortName(),
          $this->crudRoleConstantProvider->findCustomRoleConstant($class->getName()),
      );
  } else {
      $classRole = $this->getClassAttribute($class, ForRole::class)?->role;
  }
  ```
  `getShortName()` matches the `controllerName` used by `CrudControllerRegistry`, so the result is identical to `Definition::getRoleConstant()`.
- **CrudControllerRegistry / CrudConfig / CrudAdminRoleProvider** – no change needed, they already work with the nullable custom constant only.

vitek-rostislav (approved):

PR #4795, opened 2026-08-21
Make templates and parameters of CRUD actions configurable

henzigo:

:warning: **Stacked PR** - merge order is #4842 (PHPStan support for `Webmozart\Assert` calls) → #4841 (`create-crud-controller` skill) → this PR; retarget this one to `20.0` once they are merged. The `configureTemplateParameters()` implementations and docs narrow the `$entity` parameter with `Assert::isInstanceOf()` (needs #4842), and the skill from #4841 is updated here to describe `setTemplate()` / `configureTemplateParameters()` instead of the removed `getEditTemplate()` / `getEditViewData()`.

This PR lets a CRUD controller render its own template on any action and pass extra data into it.

Until now, only the edit page could be customized — a controller overrode `getEditTemplate()` to swap the template and `getEditViewData()` to hand it additional data. The list, detail and create pages had no such option at all, so anything beyond the generic layout meant leaving the CRUD controller behind and writing a plain controller instead.

The template is now part of the CRUD configuration, so it is set the same way as every other option: `$config->setTemplate(ActionType::EDIT, '...')` in `configure()`, for any action that renders a page. Extra data for the template comes from a single `getAdditionalTemplateParameters()` method that receives the action, so one controller can feed different data to different pages. Extensions can do both as well — replace the template of a controller they extend and contribute their own parameters to it — which previously was not possible at all.

Because several sources now write into the same set of template parameters, a parameter whose name is already taken by the page itself or by another source raises an error naming the colliding parameter and where it came from, instead of quietly overwriting the page data and producing a broken page that is hard to trace back.

The blog article author administration, the only page that used the old edit-only mechanism, is migrated to the new one.

The five CRUD actions are also closed for overriding now. They looked like an extension point, but they are the part that dispatches the hooks of every registered extension and carries the CSRF protection of the delete action — so replacing one of them silently switched those guarantees off, including for extensions written by somebody else. Everything an action can reasonably need is available through the configuration methods and the hook interfaces, and a flow that fits none of them can be opted out of explicitly: disable the action and register your own with its own route. No CRUD controller in the platform overrode an action, so nothing had to be rewritten. **This is a BC break, upgrade notes are included.**

The CRUD handler and hook interfaces now receive the entity as `Presentable` everywhere. `EditHandlerInterface::createDataFromEntity()` / `edit()` and the entity parameter of the create, edit and delete hooks were typed as `object`, while `getById()`, `create()` and `delete()` already used `Presentable`. Existing implementations with `object $entity` keep working thanks to parameter contravariance; the shipped handlers, the docs and the `create-crud-controller` skill follow the interfaces.

#### Fixes issues

N/A

----
  - https://jg-crud-templates.odin.shopsys.cloud
  - https://cz.jg-crud-templates.odin.shopsys.cloud
  - https://jg-crud-templates.odin.shopsys.cloud/sk


docs/administration/crud-controller/getting-started/extending-existing-crud-controller.md
```diff
@@ -78,6 +79,37 @@ public function configureForm(CrudFormConfigurator $formConfigurator, ?object $e
...
+public function getAdditionalTemplateParameters(ActionType $actionType, ?Presentable $entity = null): array
+{
+    if ($actionType !== ActionType::EDIT) {
+        return [];
+    }
+
+    return [
+        'orderItemsGridView' => $this->createOrderItemsGrid($entity)->createView(),
+    ];
+}
```
grossmannmartin: I would switch the condition to avoid learning a bad habits in the documentation. 

This has to be rewritten when I need to pass some variable to the different action, while 

```php
if ($actionType === ActionType::EDIT) {
    return [
        'orderItemsGridView' => $this->createOrderItemsGrid($entity)->createView(),
    ];
}

return [];
```

is more open to adding new params and the reading flow feels more natural.

--- 

But I'm thinking whether this is a good way to go. Wouldn't be better to just implement 
```php
public function getEditTemplateParameters(Presentable $entity)
public function getCreateTemplateParameters()
public function getListTemplateParameters()
...
```

Now I have to check the action type and mind where the entity is and where is not, which forces me to write the inline annotation, otherwise the $entity is possibly null even when it's not.

```php
/** @var \Shopsys\FrameworkBundle\Model\Blog\Author\BlogArticleAuthor $blogArticleAuthor */
$blogArticleAuthor = $entity;
```

grossmannmartin (changes requested):

docs/administration/crud-controller/getting-started/extending-existing-crud-controller.md
henzigo: I prefer a more general approach. For example, if you want the same parameters for multiple actions, you’d have to duplicate that logic.

henzigo: However, I’ve reworked the implementation to make it look a bit nicer. Please have a look at the fixup commit.

upgrade-notes/backend_20260821_171914.md
```diff
@@ -0,0 +1,11 @@
+#### CRUD controller actions are final now ([#4795](https://github.com/shopsys/shopsys/pull/4795))
+
+- `listAction()`, `detailAction()`, `createAction()`, `editAction()` and `deleteAction()` of `Shopsys\AdministrationBundle\Controller\AbstractCrudController` are `final` and cannot be overridden anymore
+    - overriding them silently disabled all CRUD extension hooks (they are dispatched from inside the action bodies) and, in the case of `deleteAction()`, also dropped the CSRF protection, because PHP attributes are not inherited by an overriding method
```
grossmannmartin: This explanation has low value during the upgrade, and I would remove this list item.

docs/administration/crud-controller/getting-started/extending-existing-crud-controller.md
grossmannmartin: I don't know... I kinda don't like this

```php
protected function configureTemplateParameters(
    CrudTemplateParameters $templateParameters,
    ActionType $actionType,
    ?Presentable $entity = null,
): void {
    if ($actionType === ActionType::EDIT) {
        /** @var \Shopsys\FrameworkBundle\Model\Blog\Author\BlogArticleAuthor $blogArticleAuthor */
        $blogArticleAuthor = $entity;

        $templateParameters->set('gridView', $this->createBlogArticlesGrid($blogArticleAuthor)->createView());
    }
}
```

1. I find the comparison of the ActionType unnecessary and error-prone
2. Inside the `if`, ' $ blogArticleAuthor ' is non-nullable just due to inline annotation. The type itself is nullable, so strictly speaking, we should check for nullability. 
3. The same parameter for multiple actions is currently a very rare scenario
4. Even when some parameters are duplicated in methods, the intention is clearer

With separate methods, I could write simpler code

```php
/**
 * @param \Shopsys\FrameworkBundle\Model\Blog\Author\BlogArticleAuthor $entity
 */
#[Override]
protected function configureEditTemplateParameters(
    CrudTemplateParameters $templateParameters,
    Presentable $entity,
): void {
    $templateParameters->set('gridView', $this->createBlogArticlesGrid($entity)->createView());
}
```

upgrade-notes/backend_20260821_171914.md
henzigo: removed

docs/administration/crud-controller/getting-started/extending-existing-crud-controller.md
vitek-rostislav: Hi, if I had to choose, I would vote for the separate methods, it seems more starightforward. Now, when customizing any template, you always have to add that `if` condition

henzigo: Thanks for the feedback, both points about the `if` and the nullable entity are valid. I'd like to propose a third option that keeps a single method but moves the context into the parameters object, the same way `CrudFormConfigurator` already carries `getActionType()`:

```php
#[Override]
protected function configureTemplateParameters(CrudTemplateParameters $templateParameters): void
{
    if ($templateParameters->isAction(ActionType::EDIT)) {
        $templateParameters->set(
            'gridView',
            $this->createBlogArticlesGrid($templateParameters->getEntity(BlogArticleAuthor::class))->createView(),
        );
    }
}
```

```php
// CrudTemplateParameters

public function getActionType(): ActionType
{
    return $this->actionType;
}

public function isAction(ActionType ...$actionTypes): bool
{
    return in_array($this->actionType, $actionTypes, true);
}

public function hasEntity(): bool
{
    return $this->entity !== null;
}

/**
 * @template T of \Shopsys\FrameworkBundle\Component\Utils\Presentable
 * @param class-string<T> $entityClass
 * @return T
 */
public function getEntity(string $entityClass): Presentable
{
    Assert::true($this->hasEntity(), sprintf('The "%s" action has no entity.', $this->actionType->value));
    Assert::isInstanceOf($this->entity, $entityClass);

    return $this->entity;
}
```

`hasEntity()` is useful when the parameter doesn't depend on a specific action but on the displayed record, e.g. an extension adding a widget to every page showing an order:

```php
public function configureTemplateParameters(CrudTemplateParameters $templateParameters): void
{
    if ($templateParameters->hasEntity()) {
        $templateParameters->set(
            'orderHistoryWidget',
            $this->createOrderHistoryWidget($templateParameters->getEntity(Order::class)),
        );
    }
}
```

What it brings:

- **Runtime check of the entity**: `getEntity(Foo::class)` fails fast with a clear message when the entity is missing or of a wrong class, and PHPStan narrows it to `Foo` (via `phpstan-webmozart-assert`). No nullable parameter, no inline `@var`. Note that separate methods can't really give this: PHP doesn't allow narrowing `Presentable $entity` to the concrete class in the child, so the type would again be only a `@param` annotation without any runtime check.
- **Stable signature, easier to extend the API**: new context (the form, datagrid, selected domain, entity for the detail action, …) is just a new getter on the object, not a signature change and a BC break for every overriding controller and extension.
- **Future custom actions**: a custom action rendered via `renderAction()` reaches the same method with its own context, so controllers and extensions can contribute parameters to it without adding a new `configure*TemplateParameters()` method to the base classes for every action.
- **One extension point instead of 4 + 4** (controller + extension), and parameters shared by several or all actions (or by all actions with a record) are set in one place.
- **Consistency** with `configureForm()` / `CrudFormConfigurator`. As a follow-up, `CrudFormConfigurator::getEntity()` could be added too, so both extension points look the same.

vitek-rostislav: Developers would still have to put those `if`s into the `configureTemplateParameters` method often, however, I like the proposed solution the most, lets go this way :wink:

PR #4798, opened 2026-08-24
Stop reporting third-party browser script errors as storefront errors

TomasLudvik:

Production Sentry has been reporting a JavaScript crash as a storefront bug, and the alert rule has been auto-creating a Jira `FE bug` for each new variant of it (SSP-4169, and SSP-4234 as a duplicate). The crash is not in code we ship: it comes from the Smartsupp live-chat widget, which is added to the site through Google Tag Manager. Their script dereferences `window.parent.document` without checking that the parent frame still exists, so it throws once the widget's frame is detached.

The reason it looked like ours is that the Sentry SDK rewrites every stack frame's origin to `app:///`, whatever host the file actually came from. A third-party file therefore becomes indistinguishable from first-party code in the Sentry UI, and gets flagged as in-app. So the developer-facing problem is not the crash itself — we cannot fix someone else's CDN bundle — but that our error reporting cannot tell our code apart from other people's.

This is addressed in two layers. Errors originating in the Smartsupp widget are no longer sent to Sentry at all, so they stop creating issues and tickets while the widget keeps working exactly as before. Separately, our own bundles are now marked with a Sentry application key, so any *other* third-party script that starts throwing is tagged `third_party_code` instead of being mistaken for ours — which means the next unrecognised vendor does not have to be diagnosed from scratch before the noise can be stopped. That second layer only tags and never discards: a frame missing the key reads as third-party, so dropping on that basis would silently throw away every browser error if key injection ever regressed.

To complete the second layer, the production alert rule needs a filter so tagged events do not open Jira issues (`third_party_code` is not equal to `true`). That is Sentry-side configuration and is not part of this PR.

#### Fixes issues

N/A

----
  - https://tl-sentry-do-not-report-external-errors.odin.shopsys.cloud
  - https://cz.tl-sentry-do-not-report-external-errors.odin.shopsys.cloud
  - https://tl-sentry-do-not-report-external-errors.odin.shopsys.cloud/sk

### DIFF
--- a/project-base/storefront/instrumentation-client.ts
+++ b/project-base/storefront/instrumentation-client.ts
@@ -1,4 +1,5 @@
 import { getPublicConfigProperty } from 'envConfig';
+import { SENTRY_APPLICATION_KEY } from 'sentryApplicationKey';
 import * as Sentry from '@sentry/nextjs';
 import { getTracePropagationTargetsFromPublicGraphqlEndpoints as getPublicGraphqlTracePropagationTargets } from 'utils/sentry/tracePropagationTargets';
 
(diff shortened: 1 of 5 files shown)


project-base/storefront/instrumentation-client.ts
```diff
@@ -18,7 +19,14 @@ if (isSentryEnabled) {
         environment: environment,
         release: release,
         tracesSampleRate: 0.1,
-        integrations: [Sentry.browserTracingIntegration()],
+        denyUrls: [/smartsupp/],
```
grossmannmartin: Can't this silently swallow even the legitimate error we can trace, just because of some subpath-match?
Also, why isn't the marking of "our" code enough? Why do we need the denyUrls at all?

TomasLudvik: I have updated it to exact domain, are you ok with this solution?

grossmannmartin (approved):

pk16011990 (approved):

PR #4800, opened 2026-08-25
Protect public form mutations against spam bots

fujta:

Public form mutations in the Frontend API can be submitted by anyone, so they attract spam bots. The contact form sends an email to the shop operator on every submission and nothing stopped a bot from flooding that inbox.

Two layers now guard it. The first one is a honey pot — a field that is a part of the form but invisible to people, so only a bot fills it in. Such a submission is dropped without sending anything and the client gets exactly the same success response as a real one, so that the bot cannot learn which field gave it away. The second one is a per-IP rate limit: after 3 submissions from one address within 5 minutes the storefront shows *Too many submissions. Try again in a few minutes.* The limit is consumed even by a submission the honey pot has already caught, otherwise a bot filling it in would never use up its quota.

Both layers are reusable for any other public form. A project protects one by adding a case to `SpamProtectedFormEnum`, letting its GraphQL input inherit `HoneyPotInputObject` and calling the facade as the first thing in the mutation — there is nothing else to configure, all protected forms share one limiter and one error code. The whole pattern, including the reasoning behind its non-obvious parts, is described in the new [Protecting a Form Against Spam](https://docs.shopsys.com/en/latest/cookbook/protecting-a-form-against-spam/) cookbook page. Neither layer is a challenge-based solution such as reCAPTCHA — both only stop bots that submit the form blindly, and neither of them affects people filling the form in.

Three related changes come with it and are worth a look when reviewing:

- the contact form page shows the success screen only when the mutation confirms the email was sent and reports an error when it was not, while until now a failed send still looked like a sent message,
- `useErrorHandler` no longer lets its `customMessage` option hide the message of a known application error code, so that the new rate limit message actually reaches the visitor,
- functional tests get a random client IP address per test method, so that test methods no longer share the quota of an IP based rate limiter.

#### Fixes issues

N/A

----
  - https://vk-ssp-4281-form-spam-protection.odin.shopsys.cloud
  - https://cz.vk-ssp-4281-form-spam-protection.odin.shopsys.cloud
  - https://vk-ssp-4281-form-spam-protection.odin.shopsys.cloud/sk


packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
```diff
@@ -0,0 +1,80 @@
...
+        return true;
+    }
+
+    /**
+     * @throws \Shopsys\FrontendApiBundle\Model\SpamProtection\Exception\TooManyFormSubmissionsUserError
+     */
+    protected function checkRateLimit(string $formName): void
+    {
+        $rateLimit = $this->getRateLimiterFactory($formName)
+            ->create($formName . ':' . $this->clientIpProvider->getClientIp())
```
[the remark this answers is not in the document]
fujta: Not fixing this. Collapsing the key to a /64 prefix would put a whole company LAN segment into a single 3-per-5-minutes bucket, so a fourth colleague sending the contact form would be rejected. IPv4 clients behind one public address already share the bucket that way, so it is not a regression, but I do not want to introduce it deliberately — an attacker with a /48 delegation still rotates through 65k /64s and gets around it anyway. Keeping the full address as the key.

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
```diff
@@ -0,0 +1,80 @@
...
+namespace Shopsys\FrontendApiBundle\Model\SpamProtection;
+
+use Psr\Log\LoggerInterface;
+use Shopsys\FrontendApiBundle\Component\HttpFoundation\ClientIpProvider;
+use Shopsys\FrontendApiBundle\Model\SpamProtection\Exception\TooManyFormSubmissionsUserError;
+use Symfony\Component\RateLimiter\RateLimiterFactoryInterface;
+
+class FormSpamProtectionFacade
+{
+    public const string HONEY_POT_FIELD_NAME = 'subject';
```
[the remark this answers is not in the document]
fujta: We prototyped a per-form field name — the name is a part of the public schema, so it cannot come from an env variable, and making it configurable means keeping it in sync in three places for every protected form (its own GraphQL input object, a parameter the facade reads, and the map on the storefront), plus two extra tests to keep those from drifting apart. With a single protected form that buys nothing, so we judged it overkill and are not doing it for now. Worth revisiting once more forms are protected; until then the rate limiter is the layer that does not depend on the name.

packages/frontend-api/src/Component/HttpFoundation/ClientIpProvider.php
```diff
@@ -0,0 +1,22 @@
...
+    protected const string UNKNOWN_CLIENT_IP = 'unknown';
+
+    public function __construct(
+        protected readonly RequestStack $requestStack,
+    ) {
+    }
+
+    public function getClientIp(): string
+    {
+        return $this->requestStack->getCurrentRequest()?->getClientIp() ?? static::UNKNOWN_CLIENT_IP;
```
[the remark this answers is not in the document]
fujta: Checked this against the Symfony source and the fallback is not reachable from the current code. getClientIp() is getClientIps()[0], and getClientIps() returns [REMOTE_ADDR], so it is null only when REMOTE_ADDR is missing from $_SERVER. Request::create() sets REMOTE_ADDR to 127.0.0.1 (vendor/symfony/http-foundation/Request.php:340), so the sub-request example does not produce it, and the cookbook says nothing about Messenger or workers. Both consumers of ClientIpProvider — StoresQuery and FormSpamProtectionFacade — are GraphQL resolvers, so they always run inside an HTTP request where the SAPI has set REMOTE_ADDR.

So the 'unknown' bucket is a defensive fallback for the nullable return type rather than a path anything takes, and it fails closed. Leaving it as is.

project-base/app/config/packages/rate_limiter.yaml
```diff
@@ -1,5 +1,11 @@
 framework:
     rate_limiter:
+        # one limiter shared by every form protected by FormSpamProtectionFacade, each form counts on its own key
+        frontend_api_form_spam_protection:
+            policy: fixed_window
```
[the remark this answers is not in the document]
fujta: Not changing it. The boundary burst is real, but every other limiter in the project uses fixed_window, and we would rather keep the policy consistent across them than have this one form differ. If we switch, it should be all of them in one go, which is out of scope here.

project-base/storefront/components/Forms/HoneyPot/HoneyPotInput.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const HoneyPotInput: FC = () => {
+    const formProviderMethods = useFormContext();
+
+    if (!formProviderMethods) {
+        throw new Error('HoneyPotInput has to be rendered inside a FormProvider.');
+    }
+
+    return (
+        <input
+            {...formProviderMethods.register(HONEY_POT_FIELD_NAME, { value: '' })}
```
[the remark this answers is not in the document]
fujta: Not addressing this. Renaming the field was already turned down as overkill in the thread on FormSpamProtectionFacade, and the logging the comment asks for as a minimum is already there — the facade logs the discarded submission with its form name. The readOnly-until-first-interaction trick would make the trap easier for a bot to recognise than autofill is likely to trip it, so we are keeping the field as it is.

project-base/storefront/utils/forms/honeyPot.ts
```diff
@@ -0,0 +1,13 @@
+import { UseFormReturn } from 'react-hook-form';
+
+/**
+ * It has to match Shopsys\FrontendApiBundle\Model\SpamProtection\FormSpamProtectionFacade::HONEY_POT_FIELD_NAME.
+ */
+export const HONEY_POT_FIELD_NAME = 'subject';
+
+export const useHoneyPot = (formProviderMethods: UseFormReturn<any>) => ({
```
chlebektomas: Jen prosím ověřit od Codexe jinak LGTM
- reusable API nezaručuje použití obou částí honeypotu. Dokumentace tvrdí, že vynechání renderHoneyPot nebo getHoneyPotInput selže na lintu, ale const { renderHoneyPot } = useHoneyPot(...) je platný a lint-clean kód. Další formulář tak může vypadat chráněně, ale honeypot nikdy neposílat. Opravil bych tvrzení v dokumentaci a přidal test každého chráněného storefront formuláře, že vyplněná skrytá hodnota skutečně dorazí do mutation variables.

chlebektomas (changes requested):

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
vitek-rostislav: I do not think it is "universal" enough. You might either already have "subject" field on another form, or you might want to use a different honeypot field on a form where "subject" would not make sense (e.g. registration?). I suggest removing `HoneyPotInputObject` + its decorator completely. Instead, lets create some map in the enum itself, something like `SpamProtectedFormEnum::getHoneyPotFieldNameIndexedByFormName()` so that developers are able to configure the field themselves. The cookbook would then instruct them to add their custom honey pot field manually into their form's input object (they might even want to add a proper description for such a field...).Does it make sense? I know we will have to adjust the storefront part as well then, sorry :smile:

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
```diff
@@ -0,0 +1,80 @@
...
+        protected readonly LoggerInterface $logger,
+        protected readonly RateLimiterFactoryInterface $formSpamProtectionRateLimiter,
+        protected readonly SpamProtectedFormEnum $spamProtectedFormEnum,
+    ) {
+    }
+
+    /**
+     * @param array<string, mixed> $input
+     * @param string $formName one of \Shopsys\FrontendApiBundle\Model\SpamProtection\SpamProtectedFormEnum cases
+     * @throws \Shopsys\FrontendApiBundle\Model\SpamProtection\Exception\TooManyFormSubmissionsUserError
```
vitek-rostislav: we do not use `@throws` annotations

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->checkRateLimit($formName);
+
+        if (!$this->isHoneyPotFilled($input)) {
+            return false;
+        }
+
+        $this->logger->info(
+            'Form submission was discarded because the honey pot field was filled in.',
+            [
+                'formName' => $formName,
```
vitek-rostislav, line 40: I would consider adding the array input into the logger context as well, it might be handy IMHO. But I do not insist on that

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
```diff
@@ -0,0 +1,80 @@
...
+            ],
+        );
+
+        return true;
+    }
+
+    /**
+     * @throws \Shopsys\FrontendApiBundle\Model\SpamProtection\Exception\TooManyFormSubmissionsUserError
```
vitek-rostislav: we do not use `@throws` annotations

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
```diff
@@ -0,0 +1,80 @@
...
+            ->create($formName . ':' . $this->clientIpProvider->getClientIp())
+            ->consume();
+
+        if (!$rateLimit->isAccepted()) {
+            throw new TooManyFormSubmissionsUserError('Too many submissions of this form. Try again later.');
+        }
+    }
+
+    /**
+     * $formName is unused here, because all forms share one limiter — a project can override this method and branch on it.
```
vitek-rostislav: I would not say "one limiter" as it might be confusing (I will add a few more comments on this topic later in the docs...) - lets rather say "one configuration"

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
+# Protecting a Form Against Spam
```
vitek-rostislav: The docs and comments use the word "limiter" for two different things – the `RateLimiterFactory` (the configuration:  
policy, limit, interval, storage) and the per-key limiter with its own counter – so it is not clear what the forms 
actually share. They share the **configuration**, not the **quota**: the form name is a part of the key, so every form
and IP address counts on its own. The only sentence that says so correctly is line 49 ("every form counts its own quota"), and these places contradict it:

- `docs/cookbook/protecting-a-form-against-spam.md:22-23` – "all protected forms share one limiter and one error code"
- `docs/cookbook/protecting-a-form-against-spam.md:148-149` – "One limiter and one Redis-backed pool serve every protected form"
- `docs/cookbook/protecting-a-form-against-spam.md:180-181` – "declare a second limiter and override `getRateLimiterFactory()`" – also too terse: it does not say that the same `cache_pool` should be reused, nor that the
  second factory has to be injected through the constructor
- `project-base/app/config/packages/rate_limiter.yaml:3` – the comment "one limiter shared by every form …, each form
  counts on its own key"
- `FormSpamProtectionFacade::getRateLimiterFactory()` docblock – "all forms share one limiter"

I would suggest one consistent wording, e.g.: "All protected forms share one rate limiter configuration (limit, interval and storage), but the quota is counted separately for every form and IP address, because the form name is a 
part of the limiter key. Declare a second limiter in `rate_limiter.yaml` (reusing the same `cache_pool`), inject it
and override `getRateLimiterFactory()` only when one form needs different limits."

packages/frontend-api/tests/Unit/Model/SpamProtection/SpamProtectedMutationsTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+use PHPUnit\Framework\TestCase;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use ReflectionMethod;
+use Shopsys\FrontendApiBundle\Model\Mutation\AbstractMutation;
+use Shopsys\FrontendApiBundle\Model\SpamProtection\FormSpamProtectionFacade;
+use Symfony\Component\Yaml\Yaml;
+
+class SpamProtectedMutationsTest extends TestCase
```
vitek-rostislav: this test is really heavy weight, fragile, and it only protects front api package, not the project base. I think we should remove it and optionally create a phpstan rule for this. But I believe the rule is not necessary, IMHO we can live without it

project-base/app/config/packages/rate_limiter.yaml
```diff
@@ -1,5 +1,11 @@
 framework:
     rate_limiter:
+        # one limiter shared by every form protected by FormSpamProtectionFacade, each form counts on its own key
```
vitek-rostislav: as mentioned earlier, lets say rather "one configuration"

project-base/app/tests/FrontendApiBundle/Functional/Contact/ContactFormMutationTest.php
```diff
@@ -21,6 +23,19 @@ public function testValidMutation(): void
         self::assertTrue($response['data']['ContactForm'] ?? null);
     }
 
+    public function testFilledHoneyPotIsSilentlyDiscarded(): void
+    {
+        $response = $this->getResponseContentForGql(__DIR__ . '/graphql/ContactFormMutation.graphql', [
+            'name' => 'Name',
+            'email' => 'not-an-email',
```
vitek-rostislav, line 30: just a question - I know this was an architectural decision but should not we perform the validation on the request first? Maybe the bot get suspicious when invalid values are accepted? But I can live with that

docs/cookbook/index.md
```diff
@@ -18,3 +18,4 @@
 - [Create Asynchronous Notifier](./create-asynchronous-notifier.md)
 - [Adding a New Order Item Type](./adding-a-new-order-item-type.md)
 - [Extending Order Detail Page](./extending-order-detail-page.md)
+- [Protecting a Form Against Spam](./protecting-a-form-against-spam.md)
```
vitek-rostislav: I would maybe add "storefront form" wording so it is obvious we are talking about the SF

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
...
+would simply stop filling it in — which is also why the honey pot must **not** be a `Blank` validation constraint.
+
+!!! info "reCAPTCHA is not part of the platform"
+
+    Both layers only stop bots that submit the form blindly, and neither of them affects people filling the form in.
+    If they turn out not to be enough for your project, a challenge-based solution has to be integrated on top of them.
+
+## Protecting another form
+
+Everything below is the complete work needed for one more form. There is nothing to configure — all protected forms
```
vitek-rostislav: I would maybe add an extra section at the end of the cookbook that would explain step by step adding a separate configuration for a form (entry in rate_limiter, injecting the config into the facade, extending the protected method...)

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+```yaml
+# app/config/services.yaml
+Shopsys\FrontendApiBundle\Model\SpamProtection\SpamProtectedFormEnum:
+    alias: App\Model\SpamProtection\SpamProtectedFormEnum
+```
+
+Without the alias the facade keeps validating against the framework enum and throws `InvalidEnumCaseException` for your
+new name. The name becomes a part of the rate limiter key, so every form counts its own quota.
```
vitek-rostislav: this is imho an implementation detail, I do not think it is important to mention it here

vitek-rostislav: btw, moreover, the enum actually belongs to the frontend api package, so "framework" might be confusing. But as I say, I would remove these sentences completely

docs/cookbook/protecting-a-storefront-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
...
+        return true;
+    }
+
+    $validator->validate();
+
+    // ... the real work
+}
+```
+
+Call it **before** `$validator->validate()`, so that malformed submissions count towards the rate limit as well.
```
vitek-rostislav, line 112: ah, I see now why we check the spam before validation, ok :)

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
...
+`renderHoneyPot` alone renders the field without ever sending its value, so the form looks protected and is not.
+Leaving either half unused fails the lint instead. Spread `getHoneyPotInput()` wherever the mutation carries the honey
+pot — inside `input` for an input-object mutation, among the top level variables for a mutation with flat ones — and
+call it inside the submit handler, because filling a registered input in does not re-render the form.
+
+`renderHoneyPot` is opt-in on purpose: a form that sends its whole form model as the mutation input would otherwise
+start sending a field its GraphQL input does not know.
+
+The value is deliberately kept **out of the typed form model** — it is a trap, not a form field, so it does not belong
+in the `*FormType`, the Yup schema or the `*FormMeta` fields. It does reach the submit handler at runtime, because
```
vitek-rostislav: the `*FormType`s are used in BE only, or? I would not confuse the docs with it then :)

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
...
+Assert both layers, following `ContactFormMutationTest`: a filled honey pot returns success without any effect, and the
+submission after the last allowed one fails with the `too-many-form-submissions` user error.
+
+## One-time configuration
+
+Already in place, listed here because a downstream project has to have it as well. One limiter and one Redis-backed
+pool serve every protected form:
+
+```yaml
+# app/config/packages/rate_limiter.yaml
```
vitek-rostislav: I would maybe shorted this section and just say something like "see app/config/packages/rate_limiter.yaml" as the config might be updated in the future and the docs might become outdated

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
...
+    A limiter declared without `cache_pool` falls back to `cache.app`, which is not configured in this project and
+    therefore ends up on the local filesystem of a single application container. Every instance then counts on its own,
+    so the effective limit is multiplied by the number of running instances and reset on every deploy — the limiter
+    appears to do nothing.
+
+    The build version is deliberately left out of the Redis prefix, so that a deploy does not grant spammers a fresh
+    quota. In the test environment the pool must not be an `array` adapter either, because that one is thrown away on
+    every kernel reboot and a functional test would never reach the limit.
+
+To give one form a limiter of its own, declare a second limiter and override
```
vitek-rostislav: yes, this is the information that deserves a separate section imho

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,202 @@
...
+`FormSpamProtectionFacade::getRateLimiterFactory()`.
+
+## Prerequisite in production: trusted proxies
+
+The limiter depends on `Request::getClientIp()`, which is only as reliable as `framework.trusted_proxies` (the
+`TRUSTED_PROXIES` environment variable, `127.0.0.1` by default).
+
+Behind a CDN or a load balancer it must be set to the address or range of that proxy. Otherwise either every visitor
+collapses into a single IP bucket and legitimate people are blocked, or `X-Forwarded-For` is accepted from anyone and
+the limit can be bypassed by spoofing it.
```
vitek-rostislav: I would rephrase it, maybe like this: 
```
he limiter counts submissions per client IP address, so it needs to
  know which one it is. Behind a CDN or a load                        
  balancer the real address arrives in the `X-Forwarded-For` header,  
  which is only trusted from the proxies listed in the                
  `TRUSTED_PROXIES` environment variable (`127.0.0.1` by default) —   
  set it to the address or range of your proxy.                       
                                                                      
  With a wrong value the limiter fails in one of two ways: if the     
  proxy is not trusted, every visitor shares the proxy's              
  address and one bucket, so legitimate people get blocked; if        
  everyone is trusted, anyone can bypass the limit by sending         
  a forged `X-Forwarded-For`.

```

And maybe add link to `application-configuration.md` where `TRUSTED_PROXIES` is explained

upgrade-notes/backend_20260824_143517.md
```diff
@@ -0,0 +1,8 @@
+#### Public form mutations can be protected against spam bots by FormSpamProtectionFacade ([#4800](https://github.com/shopsys/shopsys/pull/4800))
+
+- stop sending the `subject` field of the `ContactFormInput` Frontend API input type from your own clients, it is a honey pot and a submission carrying it is discarded while the mutation still answers `true`
+- handle the new user error code `too-many-form-submissions` in your own clients of the `ContactForm` mutation, it is returned after 3 submissions from one IP address within 5 minutes
+- check that `framework.trusted_proxies` (the `TRUSTED_PROXIES` environment variable) matches your CDN or load balancer, otherwise the per-IP limit of the mutation either counts all visitors as one address or can be bypassed by spoofing `X-Forwarded-For`
+- remove your own service definition of any `*Enum.php` class of the `Shopsys\FrontendApiBundle` namespace, the service resource pattern of `ShopsysFrontendApiBundle` registers them now
```
vitek-rostislav, line 4: lets add the same enum autoconfigration to the services.yaml in project-base

upgrade-notes/backend_20260824_143517.md
```diff
@@ -0,0 +1,8 @@
+#### Public form mutations can be protected against spam bots by FormSpamProtectionFacade ([#4800](https://github.com/shopsys/shopsys/pull/4800))
+
+- stop sending the `subject` field of the `ContactFormInput` Frontend API input type from your own clients, it is a honey pot and a submission carrying it is discarded while the mutation still answers `true`
```
vitek-rostislav: i would remove this, it is rather confusing imho

upgrade-notes/backend_20260824_143517.md
```diff
@@ -0,0 +1,8 @@
...
+- handle the new user error code `too-many-form-submissions` in your own clients of the `ContactForm` mutation, it is returned after 3 submissions from one IP address within 5 minutes
```
vitek-rostislav: i would remove this as well, I do not think BE devs need to do this, or?

vitek-rostislav (changes requested):

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
fujta: Don't know if that's a good idea. We could log sensitive data filled in form potentionaly

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
fujta: Fixed. It's okay :D I did think about making it configurable, but I went for the foolproof option with the fewest steps — which I can see now wasn't the best call.
 
One deviation from your suggestion: the honey pot field is now declared in project-base's own `ContactFormInput`, not in the package decorator. Otherwise a project could add its own field name, but never get rid of subject — inherits merges the parent field in and nothing can remove it.

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
fujta: fixed

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
fujta: fixed

packages/frontend-api/src/Model/SpamProtection/FormSpamProtectionFacade.php
fujta: fixed

docs/cookbook/protecting-a-form-against-spam.md
fujta: fixed

docs/cookbook/protecting-a-form-against-spam.md
fujta: Fixed

upgrade-notes/backend_20260824_143517.md
fujta: ok

upgrade-notes/backend_20260824_143517.md
fujta: ok

upgrade-notes/backend_20260824_143517.md
fujta: Already is

packages/frontend-api/tests/Unit/Model/SpamProtection/FormSpamProtectionFacadeTest.php
```diff
@@ -0,0 +1,200 @@
...
+        yield 'whitespace only is not filled' => [['subject' => "  \t\n "], false];
+
+        yield 'non string value is not filled' => [['subject' => 42], false];
+
+        yield 'text is filled' => [['subject' => 'Cheap pills'], true];
+
+        yield 'text surrounded by whitespace is filled' => [['subject' => '  spam  '], true];
+    }
+
+    public function testEveryFormHasItsOwnHoneyPotFieldName(): void
```
vitek-rostislav: It was hard to understand this test, then I realized a few things that might help:
1) let's rename the test to something like `testHoneyPotFieldNameIsResolvedPerForm` or `testFormReactsOnlyToItsOwnHoneyPotField`
2) let's put 'subject' string into a constant `FIRST_FORM_HONEY_POT_FIELD_NAME`
3) this is a unit test so instead of extending `SpamProtectedFormEnum` class, let's create a stub for the class and mock its `getHoneyPotFieldNameIndexedByFormName` so that we have full control of it here in the test. Reader will then see directly, that there is an enum with these two form settings and in the end, they will easily read the test as "second form with filled field that is honey pot configured in the first form should not discard the submission" :)

docs/cookbook/index.md
```diff
@@ -18,3 +18,4 @@
 - [Create Asynchronous Notifier](./create-asynchronous-notifier.md)
 - [Adding a New Order Item Type](./adding-a-new-order-item-type.md)
 - [Extending Order Detail Page](./extending-order-detail-page.md)
+- [Protecting a Storefront Form Against Spam](./protecting-a-form-against-spam.md)
```
vitek-rostislav: the link in security headers article needs to be updated yet

vitek-rostislav (changes requested):

project-base/storefront/utils/forms/honeyPot.ts
fujta: Should be fixed

packages/frontend-api/tests/Unit/Model/SpamProtection/FormSpamProtectionFacadeTest.php
fujta: fixed

docs/cookbook/index.md
fujta: fixed

packages/frontend-api/tests/Unit/Model/SpamProtection/FormSpamProtectionFacadeTest.php
```diff
@@ -0,0 +1,200 @@
...
+            'limit' => $rateLimit,
+            'interval' => '5 minutes',
+        ], new InMemoryStorage());
+    }
+
+    private function createSpamProtectedFormEnumStub(): SpamProtectedFormEnum
+    {
+        $spamProtectedFormEnumStub = $this->createStub(SpamProtectedFormEnum::class);
+        $spamProtectedFormEnumStub->method('getHoneyPotFieldNameIndexedByFormName')->willReturn([
+            SpamProtectedFormEnum::CONTACT_FORM => self::FIRST_FORM_HONEY_POT_FIELD_NAME,
```
vitek-rostislav: lets not use `SpamProtectedFormEnum::CONTACT_FORM` at all in this test, create a separate constant (`FIRST_FORM_NAME`) here in the test class please

docs/security/security-headers.md
```diff
@@ -105,3 +105,9 @@ For implementation details, see:
 **In plain English:** nginx normally includes its version number in every response (for example, `Server: nginx/1.25.3`). Turning `server_tokens` off removes the version, so attackers cannot scan for servers running vulnerable nginx versions. Same idea as X-Powered-By above.
 
 - `server_tokens` is disabled to avoid exposing nginx version details in responses
+
+## Related
+
+Security headers protect the browser. Public form endpoints need protection of their own — see
+[Protecting a Storefront Form Against Spam](../cookbook/protecting-a-form-against-spam.md)
```
vitek-rostislav: sorry, one last thing - please rename the file itself as well (we are trying to keep the names unified). (`protecting-a-form-against-spam.md` -> `protecting-a-storefront-form-against-spam.md`)

project-base/app/config/graphql/types/ModelType/ContactForm/Input/ContactFormInput.types.yaml
```diff
@@ -2,3 +2,10 @@ ContactFormInput:
     type: input-object
     inherits:
         - 'ContactFormInputDecorator'
+    config:
+        fields:
+            # honey pot of SpamProtectedFormEnum::getHoneyPotFieldNameIndexedByFormName(),
+            # nullable and unvalidated on purpose and the mutation never reads it
+            subject:
+                type: "String"
+                description: "Subject of the message"
```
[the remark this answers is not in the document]
vitek-rostislav: seems like a valid point

docs/cookbook/protecting-a-form-against-spam.md
```diff
@@ -0,0 +1,285 @@
...
+
+Shopsys\FrontendApiBundle\Model\SpamProtection\FormSpamProtectionFacade:
+    alias: App\Model\SpamProtection\FormSpamProtectionFacade
+```
+
+## Prerequisite in production: trusted proxies
+
+The limiter counts submissions per client IP address, so it needs to know which one it is. Behind a CDN or a load
+balancer the real address arrives in the `X-Forwarded-For` header, which is only trusted from the proxies listed in the
+`TRUSTED_PROXIES` environment variable (`127.0.0.1` by default) — set it to the address or range of your proxy. The
```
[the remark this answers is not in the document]
vitek-rostislav: Maybe we could remove this whole section then?

vitek-rostislav (changes requested):

project-base/app/config/graphql/types/ModelType/ContactForm/Input/ContactFormInput.types.yaml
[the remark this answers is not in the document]
fujta: fixed

docs/cookbook/protecting-a-form-against-spam.md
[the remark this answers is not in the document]
fujta: fixed

packages/frontend-api/tests/Unit/Model/SpamProtection/FormSpamProtectionFacadeTest.php
fujta: fixed

docs/security/security-headers.md
fujta: fixed

vitek-rostislav (approved):

project-base/storefront/utils/errors/useErrorHandler.ts
```diff
@@ -49,7 +49,11 @@ export function useErrorHandler<TFormValues extends FieldValues = FieldValues>(
                 switch (decision.action) {
                     case 'toast':
                         showErrorMessage(
-                            decision.origin === 'application' && customMessage ? customMessage : decision.message,
+                            // customMessage is only a fallback for errors that have no message of their own,
+                            // otherwise it would hide the mapped message of a known error code
+                            decision.origin === 'application' && decision.errorType === 'default' && customMessage
```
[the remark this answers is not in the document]
chlebektomas: Keeping the current behavior intentionally: registered application errors should display their mapped messages, while customMessage remains a fallback for the default error type. The API already exposed these distinctions through extensions.userCode, so this change does not introduce a new information disclosure through the API. The storefront upgrade notes already document the changed precedence and the impact on project overrides. We accept the existing backend behavior and do not plan a code change for this finding.

packages/frontend-api/src/Model/Mutation/ContactForm/ContactFormMutation.php
```diff
@@ -11,17 +11,25 @@
...
         protected readonly LoggerInterface $logger,
+        protected readonly FormSpamProtectionFacade $formSpamProtectionFacade,
     ) {
     }
 
     public function contactFormMutation(Argument $argument, InputValidator $validator): bool
     {
+        if ($this->formSpamProtectionFacade->shouldDiscardSubmission($argument['input'], SpamProtectedFormEnum::CONTACT_FORM)) {
+            // the same result as for a successful submission is returned on purpose, so that a bot cannot tell it was detected
+            return true;
```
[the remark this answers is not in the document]
chlebektomas: We accept this limitation of the honeypot. Returning the same successful response body does not guarantee indistinguishable response timing; the current comment should be understood in that narrower sense. This is a supplementary defense against bots that submit forms blindly, and the per-IP rate limit applies to both accepted and discarded submissions. We do not consider adding delays or preparing emails that will be discarded proportionate here, so we are keeping the current implementation. This acknowledges the timing limitation rather than claiming it has been eliminated.

chlebektomas (approved):

PR #4802, opened 2026-08-26
Blog article breadcrumb now reflects a renamed blog category

fujta:

Renaming a blog category did not reach the breadcrumb on the blog article detail. The new name showed up everywhere else — in the administration, on the category detail and in the right-hand menu — but the article's breadcrumb kept the old one until every affected article was re-saved by hand.

The cause is that blog articles are served from Elasticsearch, and the breadcrumb was a copy baked into the indexed document at export time. Blog categories have no index and are read live, which is why every other place updated instantly.

**What changes for the user**

- Renaming a blog category is now reflected in the article breadcrumb immediately, with no manual re-saving and no waiting for a re-export.
- The same applies to the other ways the breadcrumb could go stale: changing a category slug, moving a category in the tree, or deleting one.

**How**

- `BlogArticle.breadcrumb` gets an explicit resolver calling the shared breadcrumb query that `BlogCategory.breadcrumb` already uses, so it is computed from the database instead of read from the frozen copy. This removes the whole class of bug rather than one case of it.
- Second, separable commit: `BlogCategoryFacade::scheduleArticlesToExportByCategory()` now schedules the category's whole subtree. It previously matched only articles assigned to the exact edited category, while the breadcrumb is built from the whole path to the root — so an article sitting in a grandchild category never heard that its grandparent had been renamed.

**Notes for the reviewer**

- The GraphQL schema is byte-identical with and without this change, so `schema.graphql` and the storefront generated types need no regeneration.
- The `breadcrumb` field is still written into the Elasticsearch document; nothing reads it any more. Removing it properly means an index-version bump plus upgrade notes, so it is left for a follow-up.
- The singular `getBlogArticleIdsByCategory()` is kept alongside the new plural one — it is public package API and removing it would be a BC break.
- Only the two blog article *detail* fragments request `breadcrumb`; no listing fragment does. So this is one breadcrumb computation per article detail view, not an N+1 on listings.
- Blog category visibility recalculation is not wired to the Elasticsearch export at all — a separate bug of the same family, deliberately left out and worth its own ticket.

#### Fixes issues

https://shopsys.atlassian.net/browse/SSP-4272

- [ ] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
  - https://vk-ssp-4272-blog-breadcrumb-live-resolver.odin.shopsys.cloud
  - https://cz.vk-ssp-4272-blog-breadcrumb-live-resolver.odin.shopsys.cloud
  - https://vk-ssp-4272-blog-breadcrumb-live-resolver.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Model/Blog/Article/BlogArticleFacade.php
+++ b/packages/framework/src/Model/Blog/Article/BlogArticleFacade.php
@@ -166,6 +166,19 @@ class BlogArticleFacade
     }
 
     /**
+     * @param \Shopsys\FrameworkBundle\Model\Blog\Category\BlogCategory[] $blogCategories
+     * @return int[]
+     */
+    public function getBlogArticleIdsByCategories(array $blogCategories, int $domainId, string $locale): array
+    {
+        return $this->blogArticleRepository->getBlogArticleIdsByCategories(
+            $blogCategories,
+            $domainId,
+            $locale,
+        );
+    }
+
+    /**
      * @return \Shopsys\FrameworkBundle\Model\Blog\Article\BlogArticle[]
      */
     public function getAllVisibleOnDomain(DomainConfig $domainConfig): array
--- a/packages/framework/src/Model/Blog/Article/BlogArticleRepository.php
+++ b/packages/framework/src/Model/Blog/Article/BlogArticleRepository.php
@@ -232,6 +232,27 @@ class BlogArticleRepository
     }
 
     /**
+     * @param \Shopsys\FrameworkBundle\Model\Blog\Category\BlogCategory[] $blogCategories
+     * @return int[]
+     */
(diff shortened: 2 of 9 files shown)


grossmannmartin:
Thanks for the work, but I'm closing this one as the approach isn't systematic.

Blog articles are served from Elasticsearch on purpose, and the export-changed mechanism (BlogArticleExportChangedCronModule + the export queue) exists precisely to keep derived data like the breadcrumb in sync.

This PR bypasses that for a single field: BlogArticle.breadcrumb is now resolved live from the database while the rest of the article still comes from the index, and the breadcrumb field keeps being exported while nothing reads it. That leaves the index in a half-trusted state and adds a per-article DB query that the index was meant to avoid. The same reasoning would apply to any other denormalized field, and we don't want to resolve them all live.

PR #4803, opened 2026-08-26
Quick order status filters on the admin order list

fujta:

To see only the orders in one state, an administrator had to open Advanced search and build a rule. This adds a row of one-click filters above the order list — the same control that already sits one row higher for the domain filter.

**What changes for the administrator**

- The order list shows links for every order status, plus "All" as the default.
- Clicking one narrows the listing to that status.
- The filter is respected and kept by **both** Quick search and Advanced search — narrowing by status and then searching gives the intersection, not one replacing the other.

**How, and two decisions worth reviewing**

- The condition is added to the query builder *shared* by the quick and advanced branches, right after the domain condition and after the branch — which is exactly how the existing Quick Domain Filter survives both searches.
- The choice lives in the **session**, not in the URL. Both search forms are GET forms without an `action`, so `?orderStatus=3` would be wiped from the URL on the first submit, and the advanced "Reset filter" link rebuilds its URL from route parameters only, dropping it a second time.
- The existing `OrderStatusFilter` advanced-search filter is deliberately **not** reused for these links: injecting an `as[…]` rule flips `isAdvancedSearchFormSubmitted()` to true, which makes the controller take the advanced branch and discard the quick-search data outright.

**Notes for the reviewer**

- Page number is stripped in the redirect rather than on the link, because it lives on the referer — a narrowed listing has fewer pages and would otherwise land on an empty one.
- The redirect targets a relative path rebuilt from the referer. As a side effect this avoids the open redirect via a forged `Referer` that `DomainFilterController` still has — worth its own ticket.
- The new facade has no namespace parameter (unlike `AdminDomainFilterTabsFacade`) because only one listing uses it.
- `AdminDomainFilterTabsFacade` calls `setSelectedDomainId()` from inside a `catch (SessionNotFoundException)`, which would rethrow. The new facade is structured to avoid that; the existing one was left untouched.
- Not verified: the click-through in a browser.

#### Fixes issues

https://shopsys.atlassian.net/browse/SSP-4104

- [ ] I have read and signed [License Agreement for contributions](https://www.shopsys.com/license-agreement)

----
  - https://vk-ssp-4104-quick-order-status-filter.odin.shopsys.cloud
  - https://cz.vk-ssp-4104-quick-order-status-filter.odin.shopsys.cloud
  - https://vk-ssp-4104-quick-order-status-filter.odin.shopsys.cloud/sk

----
  - https://vk-ssp-4104-quick-order-status-filter.odin.shopsys.cloud
  - https://cz.vk-ssp-4104-quick-order-status-filter.odin.shopsys.cloud
  - https://vk-ssp-4104-quick-order-status-filter.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/templates/content/order/list.html.twig
+++ b/packages/administration/templates/content/order/list.html.twig
@@ -6,49 +6,60 @@
 {% block h1 %}{{ 'Orders'|trans }}{% endblock %}
 
 {% block main_content %}
-    <div class="card">
-        <div class="card-header">
-            <ul class="nav nav-tabs card-header-tabs" data-bs-toggle="tabs">
-                <li class="nav-item">
-                    <a href="#quick"
-                       class="nav-link{{ not isAdvancedSearchFormSubmitted ? ' active' }}"
-                       data-bs-toggle="tab"
-                    >
-                        {{ 'Quick search'|trans }}
-                    </a>
-                </li>
-                <li class="nav-item">
-                    <a href="#advanced"
-                       class="nav-link{{ isAdvancedSearchFormSubmitted ? ' active' }}"
-                       data-bs-toggle="tab"
-                    >
-                        {{ 'Advanced search'|trans }}
-                    </a>
-                </li>
-            </ul>
-        </div>
-        <div class="card-body">
-            <div class="tab-content">
-                <div class="tab-pane{{ not isAdvancedSearchFormSubmitted ? ' active show' }}" id="quick">
-                    {{ form_start(quickSearchForm) }}
-                    {% include '@ShopsysAdministration/partial/_quick_search_form.html.twig' with {
-                        form_input: quickSearchForm.text,
-                        form_submit: quickSearchForm.submit,
-                        hide_title: true,
-                        placeholder: 'Search by order number, email, name, or UUID...'|trans,
-                        info_message: 'Searching searches in <ul> <li>UUID</li><li>order number</li><li>registered customer email</li><li>customer\'s email in order</li><li>customer\'s last name</li><li>company name.'|trans,
-                    } %}
-                    {{ form_end(quickSearchForm) }}
+    <div class="row row-deck row-cards mb-3">
+        <div class="col-12 col-xl-8">
+            <div class="card">
+                <div class="card-header">
+                    <ul class="nav nav-tabs card-header-tabs" data-bs-toggle="tabs">
+                        <li class="nav-item">
+                            <a href="#quick"
+                               class="nav-link{{ not isAdvancedSearchFormSubmitted ? ' active' }}"
+                               data-bs-toggle="tab"
+                            >
+                                {{ 'Quick search'|trans }}
+                            </a>
+                        </li>
+                        <li class="nav-item">
+                            <a href="#advanced"
+                               class="nav-link{{ isAdvancedSearchFormSubmitted ? ' active' }}"
+                               data-bs-toggle="tab"
+                            >
+                                {{ 'Advanced search'|trans }}
+                            </a>
+                        </li>
+                    </ul>
                 </div>
-                <div class="tab-pane{{ isAdvancedSearchFormSubmitted ? ' active show' }}" id="advanced">
-                    {% include '@ShopsysAdministration/content/order/advancedSearch/advancedSearch.html.twig'
-                        with {advancedSearchForm: advancedSearchForm} %}
+                <div class="card-body">
+                    <div class="tab-content">
+                        <div class="tab-pane{{ not isAdvancedSearchFormSubmitted ? ' active show' }}" id="quick">
+                            {{ form_start(quickSearchForm) }}
+                            {% include '@ShopsysAdministration/partial/_quick_search_form.html.twig' with {
+                                form_input: quickSearchForm.text,
+                                form_submit: quickSearchForm.submit,
+                                hide_title: true,
+                                placeholder: 'Search by order number, email, name, or UUID...'|trans,
+                                info_message: 'Searching searches in <ul> <li>UUID</li><li>order number</li><li>registered customer email</li><li>customer\'s email in order</li><li>customer\'s last name</li><li>company name.'|trans,
+                            } %}
+                            {{ form_end(quickSearchForm) }}
+                        </div>
+                        <div class="tab-pane{{ isAdvancedSearchFormSubmitted ? ' active show' }}" id="advanced">
+                            {% include '@ShopsysAdministration/content/order/advancedSearch/advancedSearch.html.twig'
+                                with {advancedSearchForm: advancedSearchForm} %}
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+        <div class="col-12 col-xl-4">
+            <div class="card">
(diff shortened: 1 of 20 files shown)


packages/administration/templates/content/order/list.html.twig
```diff
@@ -50,5 +50,7 @@
 
     {{ render(controller('Shopsys\\FrameworkBundle\\Controller\\Admin\\DomainFilterController::domainFilterTabsAction', { namespace: domainFilterNamespace })) }}
 
+    {{ render(controller('Shopsys\\FrameworkBundle\\Controller\\Admin\\OrderStatusFilterController::orderStatusFilterTabsAction')) }}
```
vitek-rostislav: maybe, I would consider finding a way how to pass additional filter directly into the domains tab filter (or better, create a new section that enables multiple quick filters that would be more compact) as now, it is kind of spread out IMHO. But this is rather a desing/business question, so I would check the current design with product owner and then we can discuss the alternative approach as that would require a wider refactoring

vitek-rostislav (approved):

packages/framework/src/Controller/Admin/OrderStatusFilterController.php
```diff
@@ -0,0 +1,76 @@
...
+        if (is_array($gridsParameters)) {
+            foreach ($gridsParameters as $gridId => $gridParameters) {
+                if (is_array($gridParameters)) {
+                    unset($gridParameters['page']);
+                    $queryParameters[Grid::GET_PARAMETER][$gridId] = $gridParameters;
+                }
+            }
+        }
+
+        $path = $urlParts['path'] ?? '/';
```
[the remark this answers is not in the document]
fujta: Good catch, fixed. The guard next to the host check now also rejects a referer path that is not single-slash-rooted, so both `//evil.com/x` and `/\evil.com` fall back to `admin_order_list`.

I kept it as a separate `isSafeRefererPath()` next to the host check rather than putting it inside `removeGridPagingFromUrl()` — that method only strips a query parameter, it shouldn't be the one deciding whether a redirect is safe.

`DomainFilterController::selectDomainAction()` has its own ticket, as noted in the PR description.

packages/framework/src/Controller/Admin/OrderStatusFilterController.php
```diff
@@ -0,0 +1,76 @@
...
+    #[CanView]
+    public function orderStatusFilterTabsAction(): Response
+    {
+        return $this->render('@ShopsysAdministration/partial/quick_order_status_filter.html.twig', [
+            'orderStatuses' => $this->orderStatusFacade->getAll(),
+            'selectedOrderStatusId' => $this->adminOrderStatusFilterFacade->getSelectedOrderStatusId(),
+        ]);
+    }
+
+    #[Route(path: '/order/filter-status/{orderStatusId}', requirements: ['orderStatusId' => '\d+'])]
```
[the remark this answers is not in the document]
fujta: Leaving this one as is, deliberately.

Our session cookie is `cookie_samesite: lax` (checked with `debug:config framework session`), so the `` PoC never sends the cookie — that request just hits the login page. It only works through a top-level navigation, and then the admin lands directly on the order list with the selected status visible in the dropdown. No data changes and nothing happens silently.

Adding a token only here would also make the two quick filters on the same page behave differently, since `DomainFilterController::selectDomainAction()` has none either. If we want to close it, it should be done for both at once, in a separate ticket.

packages/administration/templates/content/order/list.html.twig
fujta: Fixed

upgrade-notes/backend_20260826_100423.md
```diff
@@ -0,0 +1,12 @@
+#### Quick order status filters on the admin order list ([#4803](https://github.com/shopsys/shopsys/pull/4803))
+
+- added `Shopsys\FrameworkBundle\Model\Order\Status\AdminOrderStatusFilterFacade` keeping the selected order status in the session
+- added `Shopsys\FrameworkBundle\Controller\Admin\OrderStatusFilterController` with the `admin_orderstatusfilter_selectorderstatus` route
+    - if you customize routes in smoke tests, skip this route the same way as `admin_domainfilter_selectdomain`
+- `OrderListController::__construct()` takes a new `AdminOrderStatusFilterFacade` argument
+    - if you override this controller, add the argument to your constructor
+- added `Shopsys\FrameworkBundle\Controller\Admin\DomainFilterController::domainFilterSelectAction()` rendering the quick domain filter as a dropdown instead of a row of buttons
+    - `domainFilterTabsAction()` is unchanged, so all other admin lists keep the buttons
+- `@ShopsysAdministration/content/order/list.html.twig` renders both quick filters as dropdowns in one card next to the search card and no longer renders `DomainFilterController::domainFilterTabsAction`
```
vitek-rostislav, line 3: I would keep just this bullet point (plus the following "see #project-base-diff"), all the points above are unnecessary

packages/administration/templates/partial/quick_domain_filter_select.html.twig
```diff
@@ -0,0 +1,35 @@
...
+    <div class="text-nowrap">{{ 'Domain'|trans }}</div>
+    <div class="dropdown">
+        <a href="#"
+           class="btn dropdown-toggle w-100 justify-content-between"
+           data-bs-toggle="dropdown"
+           role="button"
+        >
+            <span>
+                {% if selectedDomainId is null %}
+                    {{ 'All domains'|trans }}
```
vitek-rostislav, line 12: It would be nice to add an icon for the "all" options here and in the status filter, IMHO that looks much better and helps the UX (without it, the "all" option looks rather like a section label than a valid option), what do you think?

I suggest using [this icon](https://tabler.io/icons?icon=world), just the spacing needs to be adjusted to match the other options 😇 

Image

To use it via `{{ ux_icon('world') }}` in the twig templates, you just need to put the following line into `ux_icons.yaml` file:
```
world: 'tabler:world'
```

and then run `./bin/console ux:icons:lock`

vitek-rostislav (changes requested):

packages/administration/templates/partial/quick_domain_filter_select.html.twig
fujta: Okay, I've used that globe icon for domains and for status I've created new icon (globe icon for statutes is not suitable imo and also it's confusing when it's used above for domains)

image

upgrade-notes/backend_20260826_100423.md
fujta: fixed

vitek-rostislav (approved):

packages/framework/src/Controller/Admin/CustomerController.php
```diff
@@ -229,6 +229,7 @@ public function listAction(Request $request): Response
             'isAdvancedSearchFormSubmitted' => $this->customerUserAdvancedSearchFacade->isAdvancedSearchFormSubmitted(
                 $request,
             ),
+            'showDomainSwitcher' => count($this->domain->getAdminEnabledDomains()) > 1,
```
vitek-rostislav: you do not have to pass the variable this way - you should use `isMultidomain()` twig function in the template directly.

Let's put aside the fact the function does not take the admin selected domains into account (we could target that separately). At least, it will be unified

vitek-rostislav (approved):

PR #4810, opened 2026-08-28
Main friendly URL is no longer changed when entity is renamed

TomasLudvik:

Until now, changing the name of anything with a friendly URL (product, category, brand, flag, article, blog article, blog category, SEO category, store) generated a new URL from the new name and made it the main one, demoting the previous URL to a redirect. From the SEO point of view this silently replaced established addresses and took control away from the administrator.

Now, once an entity has a main URL on a domain, renaming it leaves its URLs untouched — no new main URL and no new redirect record is created, so the main URL may intentionally differ from the current name. The only way to change the main URL is the manual URL management in administration. When an entity has no main URL for a domain yet (a newly created entity, a newly added domain, or a name filled in a language that was empty before), the URL is still generated from the name as before.

Downstream projects that rely on the previous automatic regeneration can restore it by overriding the new protected method `FriendlyUrlFacade::shouldCreateFriendlyUrl()`.

#### Fixes issues

[SSP-4275](https://shopsys.atlassian.net/browse/SSP-4275)

----
  - https://tl-entity-url-does-not-change.odin.shopsys.cloud
  - https://cz.tl-entity-url-does-not-change.odin.shopsys.cloud
  - https://tl-entity-url-does-not-change.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
+++ b/packages/framework/src/Component/Router/FriendlyUrl/FriendlyUrlFacade.php
@@ -65,6 +65,10 @@ class FriendlyUrlFacade
 
     protected function resolveUniquenessOfFriendlyUrl(FriendlyUrl $friendlyUrl, string $entityName): void
     {
+        if (!$this->shouldCreateFriendlyUrl($friendlyUrl)) {
+            return;
+        }
+
         $attempt = 0;
 
         do {
@@ -100,6 +104,17 @@ class FriendlyUrlFacade
 
         $this->em->persist($friendlyUrl);
         $this->setFriendlyUrlAsMain($friendlyUrl);
+    }
+
+    protected function shouldCreateFriendlyUrl(FriendlyUrl $friendlyUrl): bool
+    {
+        $mainFriendlyUrl = $this->findMainFriendlyUrl(
+            $friendlyUrl->getDomainId(),
+            $friendlyUrl->getRouteName(),
+            $friendlyUrl->getEntityId(),
+        );
+
+        return $mainFriendlyUrl === null;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


vitek-rostislav (changes requested):
As we discussed, let's rather go with explicit methods instead of the guard inside `resolveUniquenessOfFriendlyUrl()`:

- keep `createFriendlyUrls()` / `createFriendlyUrlForDomain()` unchanged (unconditional "create"),
- add new public methods named something like `createFriendlyUrlsIfMainMissing()` / `createFriendlyUrlForDomainIfMainMissing()` that decide directly from the scalars (`routeName`, `entityId`, `domainId`) and build the `FriendlyUrl` entity only for domains where the main URL is missing – no throw-away entities just for the lookup,
- keep `shouldCreateFriendlyUrl()` as the protected extension point, but let it take the scalars instead of a `FriendlyUrl` and call it from the new methods (for the all-domains variant consider `getMainFriendlyUrlsIndexedByDomains()` so it's a single query).

Please switch all `edit()` methods to the new variants, not only the four touched here, so the behaviour stays consistent across entities: `ProductFacade`, `CategoryFacade`, `BlogArticleFacade`, `BlogCategoryFacade`, `FlagFacade` (`createFriendlyUrls`) and `ArticleFacade`, `BrandFacade`, `StoreFacade` (`createFriendlyUrlForDomain`). `create()` methods, `FriendlyUrlGeneratorFacade` and fixtures stay on the original ones.

Upgrade notes then just need to mention the new methods and that projects with their own `edit()` should switch to them.

Btw I am ok with force-pushing the branch after the changes, as this would be completely reworked anyway :)

pk16011990 (approved):

PR #4812, opened 2026-08-28
Webalize uploaded file names to prevent issues with special characters

grossmannmartin:

Uploading a file whose name contained special characters (e.g. a customer complaint file named `Pozadí obrázku „generated-image“ bylo odstraněno.png`) caused problems, because the original file name is stored and later used as-is.

The name of an uploaded file is now webalized on assignment — diacritics and special characters (typographic quotes, etc.) are replaced with safe ASCII equivalents, while letter case is preserved. Customers can therefore attach files with any name to a complaint without breaking the upload.

#### Fixes issues
N/A

----
  - https://mg-webalize-uploaded-file-name.odin.shopsys.cloud
  - https://cz.mg-webalize-uploaded-file-name.odin.shopsys.cloud
  - https://mg-webalize-uploaded-file-name.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Component/AbstractUploadedFile/AbstractUploadedFile.php
+++ b/packages/framework/src/Component/AbstractUploadedFile/AbstractUploadedFile.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Component\AbstractUploadedFile;
 
 use Doctrine\ORM\Mapping as ORM;
+use Nette\Utils\Strings;
 use Override;
 use Shopsys\FrameworkBundle\Component\FileUpload\EntityFileUploadInterface;
 use Shopsys\FrameworkBundle\Component\FileUpload\Exception\InvalidFileKeyException;
@@ -40,7 +41,9 @@ abstract class AbstractUploadedFile implements EntityFileUploadInterface, Upload
      */
     #[AsMcpColumn]
     #[ORM\Column(type: 'string', length: 255)]
-    protected $name;
+    protected $name {
+        set => Strings::webalize($value, lower: false);
+    }
 
     /**
      * @var int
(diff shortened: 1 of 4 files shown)


packages/framework/src/Component/AbstractUploadedFile/AbstractUploadedFile.php
```diff
@@ -40,7 +41,9 @@ abstract class AbstractUploadedFile implements EntityFileUploadInterface, Upload
      */
     #[AsMcpColumn]
     #[ORM\Column(type: 'string', length: 255)]
-    protected $name;
+    protected $name {
+        set => Strings::webalize($value, lower: false);
+    }
```
TomasLudvik, line 49: Several things here:

1. We have setter for this property, would not be better to use that?
2. We alreeady have TransformStringHelper::safeFilename() would it not be better usage here?
3. Should we have some migration to fix already wrong existing files with this name?
4. Entity hooks seem to me like increased difficulty for projects.
5. This could possibly end with empty string - we should have some fallback for it.

TomasLudvik (changes requested):

packages/framework/src/Component/AbstractUploadedFile/AbstractUploadedFile.php
grossmannmartin: 1. Setter – both child classes assign $this->name in the constructor directly.  Switching just that one line to $this->setName() while other properties use direct assignment would be an unexpected, confusing pattern. On top of that, nothing would stop project code from assigning the property directly and unintentionally bypassing the sanitization. 
2. TransformStringHelper::safeFilename() is not static, so it cannot be used in the entity. Moreover, I don't think that an underscored filename is good for use on the web. 
3. Yes, that makes sense. Added in 277778fa55
4. Property hooks – for projects, nothing changes unless they redeclare the property; I'll mention it in the upgrade notes. 
5. Empty string – good catch! Fixed in 277778fa55

TomasLudvik (approved):

pk16011990 (approved):

PR #4813, opened 2026-08-28
VAT selects in admin now show the rate percentage

vitek-rostislav:

Administrators choose a VAT rate in several admin forms (product prices, transport prices, payment prices, default VAT setting), but the select only showed the rate name (e.g. "Standard rate"), so they had to remember or look up how much each rate actually is.

Every VAT select now shows the percentage next to the name, e.g. **Standard rate (21 %)**, formatted according to the admin locale.

To make this a single change, the four hand-written `ChoiceType` configurations were first unified into a new reusable `VatChoiceType` (`domain_id` option, `NotBlank` validation derived from `required`). Any future change to how VAT rates are offered in admin is done in one place.

Along the way we removed `VatFacade::getAllForDomainIncludingMarkedForDeletion()` (and its `VatRepository` counterpart): its implementation has been identical to `getAllForDomain()` since #1498 made the VAT rates domain-scoped (2019) — the VATs marked for deletion have not been filtered out since then, so the distinction only pretended to exist. This is the only BC break, covered by an upgrade note.

#### Fixes issues
...

----
  - https://rv-vat-choice-type.odin.shopsys.cloud
  - https://cz.rv-vat-choice-type.odin.shopsys.cloud
  - https://rv-vat-choice-type.odin.shopsys.cloud/sk

----
  - https://rv-vat-choice-type.odin.shopsys.cloud
  - https://cz.rv-vat-choice-type.odin.shopsys.cloud
  - https://rv-vat-choice-type.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Form/Admin/Product/Price/ProductPricesWithVatSelectType.php
+++ b/packages/framework/src/Form/Admin/Product/Price/ProductPricesWithVatSelectType.php
@@ -32,7 +32,7 @@ final class ProductPricesWithVatSelectType extends AbstractType
             ->add('vat', ChoiceType::class, [
                 'block_prefix' => 'prices_select_vat_input',
                 'required' => true,
-                'choices' => $this->vatFacade->getAllForDomainIncludingMarkedForDeletion($options['domain_id']),
+                'choices' => $this->vatFacade->getAllForDomain($options['domain_id']),
                 'choice_label' => 'name',
                 'choice_value' => 'id',
                 'constraints' => [
--- a/packages/framework/src/Form/Admin/Product/Price/ProductPricesWithVatSelectType.php
+++ b/packages/framework/src/Form/Admin/Product/Price/ProductPricesWithVatSelectType.php
@@ -5,23 +5,16 @@ declare(strict_types=1);
 namespace Shopsys\FrameworkBundle\Form\Admin\Product\Price;
 
 use Override;
-use Shopsys\FrameworkBundle\Model\Pricing\Vat\VatFacade;
+use Shopsys\FrameworkBundle\Form\VatChoiceType;
 use Shopsys\FrameworkBundle\Model\Product\ProductInputPriceData;
 use Symfony\Component\Form\AbstractType;
-use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
 use Symfony\Component\OptionsResolver\OptionsResolver;
-use Symfony\Component\Validator\Constraints;
 
 final class ProductPricesWithVatSelectType extends AbstractType
 {
-    public function __construct(
-        protected readonly VatFacade $vatFacade,
-    ) {
-    }
-
     /**
      * {@inheritdoc}
      */
@@ -29,16 +22,9 @@ final class ProductPricesWithVatSelectType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options): void
     {
         $builder
-            ->add('vat', ChoiceType::class, [
+            ->add('vat', VatChoiceType::class, [
                 'block_prefix' => 'prices_select_vat_input',
-                'required' => true,
-                'choices' => $this->vatFacade->getAllForDomain($options['domain_id']),
-                'choice_label' => 'name',
-                'choice_value' => 'id',
-                'constraints' => [
-                    new Constraints\NotBlank(message: 'Please enter VAT rate'),
-                ],
-                'label' => 'VAT',
+                'domain_id' => $options['domain_id'],
             ])
             ->add(
                 'manualInputPricesByPricingGroupId',
--- a/packages/framework/src/Form/Admin/Vat/VatSettingsFormType.php
+++ b/packages/framework/src/Form/Admin/Vat/VatSettingsFormType.php
@@ -6,9 +6,8 @@ namespace Shopsys\FrameworkBundle\Form\Admin\Vat;
 
 use Override;
 use Shopsys\FrameworkBundle\Component\Domain\AdminDomainTabsFacade;
-use Shopsys\FrameworkBundle\Model\Pricing\Vat\VatFacade;
+use Shopsys\FrameworkBundle\Form\VatChoiceType;
 use Symfony\Component\Form\AbstractType;
-use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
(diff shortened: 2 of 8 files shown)


packages/framework/src/Form/VatChoiceType.php
```diff
@@ -0,0 +1,61 @@
...
+                    $vat->getName(),
+                    $this->numberFormatterExtension->formatPercent($vat->getPercent()),
+                ),
+                'choice_value' => 'id',
+                'choice_translation_domain' => false,
+                'label' => 'VAT',
+                'required' => true,
+                'constraints' => fn (Options $options): array => $options['required']
+                    ? [new Constraints\NotBlank(message: 'Please enter VAT rate')]
+                    : [],
```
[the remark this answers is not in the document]
vitek-rostislav: this was an intention

packages/framework/src/Form/VatChoiceType.php
```diff
@@ -0,0 +1,61 @@
...
+            ->setAllowedTypes('domain_id', 'int')
+            ->setDefault('include_marked_for_deletion', false)
+            ->setAllowedTypes('include_marked_for_deletion', 'bool')
+            ->setDefaults([
+                'choices' => function (Options $options): array {
+                    if ($options['include_marked_for_deletion']) {
+                        return $this->vatFacade->getAllForDomainIncludingMarkedForDeletion($options['domain_id']);
+                    }
+
+                    return $this->vatFacade->getAllForDomain($options['domain_id']);
```
grossmannmartin: The implementation of those two methods (`getAllForDomainIncludingMarkedForDeletion` and `getAllForDomain`) is identical. I think this is a nice opportunity to fix it 😇

packages/framework/src/Form/Admin/Product/Price/ProductPricesWithVatSelectType.php
```diff
@@ -5,40 +5,27 @@
...
-                'required' => true,
-                'choices' => $this->vatFacade->getAllForDomainIncludingMarkedForDeletion($options['domain_id']),
-                'choice_label' => 'name',
-                'choice_value' => 'id',
-                'constraints' => [
-                    new Constraints\NotBlank(message: 'Please enter VAT rate'),
-                ],
-                'label' => 'VAT',
+                'domain_id' => $options['domain_id'],
+                'include_marked_for_deletion' => true,
```
grossmannmartin: Shouldn't we allow only those not marked for deletion here?

packages/framework/src/Form/Admin/Vat/VatSettingsFormType.php
```diff
@@ -26,11 +24,8 @@ public function __construct(
...
         $builder
-            ->add('defaultVat', ChoiceType::class, [
-                'required' => true,
-                'choices' => $this->vatFacade->getAllForDomain($this->adminDomainTabsFacade->getSelectedDomainId()),
-                'choice_label' => 'name',
-                'choice_value' => 'id',
+            ->add('defaultVat', VatChoiceType::class, [
+                'domain_id' => $this->adminDomainTabsFacade->getSelectedDomainId(),
                 'constraints' => [
                     new Constraints\NotBlank(message: 'Please enter default VAT'),
```
grossmannmartin, line 30: the constraint is not necessary anymore as the type already specify it, or?

packages/framework/src/Form/TransportInputPricesType.php
```diff
@@ -6,39 +6,25 @@
...
-                'required' => true,
-                'choices' => $this->vatFacade->getAllForDomainIncludingMarkedForDeletion($options['domain_id']),
-                'choice_label' => 'name',
-                'choice_value' => 'id',
-                'constraints' => [
-                    new NotBlank(message: 'Please enter VAT rate'),
-                ],
-                'label' => 'VAT',
+                'domain_id' => $options['domain_id'],
+                'include_marked_for_deletion' => true,
```
grossmannmartin: Does it make sense to allow selecting those that are marked for deletion? 

Hmm... when I think about it, the reason is probably that we have to display the currently set value because it hasn't yet changed, right? 
Then I don't understand the duality from the other side... We should always allow to display those that are marked for deletion as they may be set, or? 

Now the only place where it could make sense to NOT display the marked VAT is the default VAT setting... but due to the identical implementation of the repository method, it shows anyway...

packages/framework/src/Form/VatChoiceType.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $resolver
+            ->setRequired('domain_id')
+            ->setAllowedTypes('domain_id', 'int')
+            ->setDefault('include_marked_for_deletion', false)
+            ->setAllowedTypes('include_marked_for_deletion', 'bool')
+            ->setDefaults([
+                'choices' => function (Options $options): array {
+                    if ($options['include_marked_for_deletion']) {
+                        return $this->vatFacade->getAllForDomainIncludingMarkedForDeletion($options['domain_id']);
```
grossmannmartin: Oh god, this is another rabbit hole 🕳️🐇  

When you mark a standard rate for deletion, and then you run the vatDeletion cron module, it ends with the error 

```A new entity was found through the relationship 'App\Model\Product\ProductDomain#product' that was not configured to cascade persist operations for entity: App\Model\Product\Product@4665. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example #[ORM\ManyToOne(..., cascade: ['persist'])]. If you cannot find out which entity causes the problem implement 'App\Model\Product\Product#__toString()' to get a clue."```

I'm starting to think that the whole cron module replacing overhead could be easily solved by just a few simple database queries run during the request.

grossmannmartin (changes requested):

packages/framework/src/Form/Admin/Vat/VatSettingsFormType.php
vitek-rostislav: it must be there to change the message ("Please enter VAT rate" is a default value vs "Please enter default VAT" here.) We could get rid of the custom message and it would still make sense imho, but on the other hand, I do not mind redefining the constraint in order to change the message :shrug:

packages/framework/src/Form/VatChoiceType.php
vitek-rostislav: `getAllForDomainIncludingMarkedForDeletion` removed

packages/framework/src/Form/Admin/Product/Price/ProductPricesWithVatSelectType.php
vitek-rostislav: the option no longer exists

packages/framework/src/Form/TransportInputPricesType.php
vitek-rostislav: The option no longer exists :slightly_smiling_face:

packages/framework/src/Form/VatChoiceType.php
vitek-rostislav: https://github.com/shopsys/shopsys/pull/4815 :slightly_smiling_face:

grossmannmartin (approved):

PR #4816, opened 2026-09-01
Add custom image support to Extended SEO categories

TomasLudvik:

Administrators can now upload a custom image to an Extended SEO category on the last step of its form (the one with heading, description, title, and meta description). The image can be uploaded, replaced, and deleted, just like other entity images in the administration.

Until now, every SEO variant of a category always displayed the generic image of the category itself, so e.g. "Electronics without HDMI in sale" could not be visually distinguished from plain "Electronics". When a custom image is uploaded, the storefront now shows it on the SEO category page (both the visible header image and the og:image meta tag); SEO categories without their own image keep showing the category image as before.

Several demo SEO categories were given a demo image, and the fallback behavior is covered by Frontend API tests. No database migration and no GraphQL schema change were needed.

#### Fixes issues
N/A

----
  - https://tl-seo-category-image.odin.shopsys.cloud
  - https://cz.tl-seo-category-image.odin.shopsys.cloud
  - https://tl-seo-category-image.odin.shopsys.cloud/sk


packages/frontend-api/src/Component/Image/ImageApiRepository.php
```diff
@@ -116,4 +116,20 @@ public function getImageCountsIndexedByEntityId(array $entityIds, string $entity
...
     }
+
+    /**
+     * @return int[]
+     */
+    public function getEntityIdsWithImageByEntityName(string $entityName): array
+    {
+        $entityIds = $this->entityManager->createQueryBuilder()
+            ->select('DISTINCT i.entityId')
+            ->from($this->entityNameResolver->resolve(Image::class), 'i')
```
grossmannmartin: It's not necessary to call entitynameresolver manually.

packages/frontend-api/src/Component/Image/ImageApiRepository.php
```diff
@@ -116,4 +116,20 @@ public function getImageCountsIndexedByEntityId(array $entityIds, string $entity
...
+            ->where('i.entityName = :entityName')->setParameter('entityName', $entityName)
+            ->andWhere('i.type IS NULL')
```
grossmannmartin: Here we've hardcoded the state of the seomixes, thus making it hardly usable for any typed images. 
Wouldn't be better to pass the type as the parameter?

packages/frontend-api/src/Component/Image/ImageApiRepository.php
```diff
@@ -116,4 +116,20 @@ public function getImageCountsIndexedByEntityId(array $entityIds, string $entity
...
+            ->getQuery()
+            ->getSingleColumnResult();
+
+        return array_map(static fn ($entityId) => (int)$entityId, $entityIds);
```
grossmannmartin: The result should be mapped as an integer by default, so this array_map is redundant.

grossmannmartin (changes requested):

grossmannmartin (approved):

packages/frontend-api/src/Model/Resolver/Image/CategoryOrReadyCategorySeoMixImagesQuery.php
```diff
@@ -5,43 +5,72 @@
...
+            $this->getCategoryId($categoryOrReadyCategorySeoMix),
+            self::CATEGORY_ENTITY_NAME,
+            $type,
+        );
+    }
+
+    protected function hasReadyCategorySeoMixOwnImage(ReadyCategorySeoMix $readyCategorySeoMix, ?string $type): bool
+    {
+        $readyCategorySeoMixEntityName = $this->imageConfig->getImageEntityConfig($readyCategorySeoMix)->getEntityName();
+        $readyCategorySeoMixIdsWithImage = $this->imageApiFacade->getEntityIdsWithImageByEntityNameAndType(
```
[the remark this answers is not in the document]
TomasLudvik: Fixed in https://github.com/shopsys/shopsys/commit/190b162a7b0ae6f1280407744786c724e93451e2 — the own-image check now returns early unless `type` is one of the image types configured for ReadyCategorySeoMix (new `ImageEntityConfig::hasType()`), so the number of distinct queries is bounded by the configuration, not by client input.

packages/frontend-api/src/Component/Image/ImageApiFacade.php
```diff
@@ -44,4 +50,18 @@ public function getImageCountsIndexedByEntityId(array $entityIds, string $entity
...
+     * @return int[]
+     */
+    public function getEntityIdsWithImageByEntityNameAndType(string $entityName, ?string $type): array
+    {
+        return $this->inMemoryCache->getOrSaveValue(
+            static::IMAGE_API_CACHE_NAMESPACE,
+            fn () => $this->imageApiRepository->getEntityIdsWithImageByEntityNameAndType($entityName, $type),
+            static::ENTITY_IDS_WITH_IMAGE_CACHE_KEY,
+            $entityName,
+            $type,
```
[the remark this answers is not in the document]
TomasLudvik: Fixed in https://github.com/shopsys/shopsys/commit/190b162a7b0ae6f1280407744786c724e93451e2 — `type` is validated against the entity's configured image types before it reaches the facade, so only `null` or a configured type name can ever become part of the cache key; empty strings and values with reserved characters are rejected up front.

PR #4823, opened 2026-09-04
Send delayed messages of successful handlers even when a sibling handler fails

grossmannmartin:

Messages dispatched inside a handler (typically e-mails) are collected and sent only after the whole message has been processed by the worker. When a message has several handlers and one of them fails, Symfony Messenger retries the message but runs only the failed handler again; the handlers that already succeeded are skipped. Until now, everything collected during such a run stayed in the collector, so the e-mails of the successful handlers were never sent, and on top of that they leaked into the next processed message.

This PR fixes that: the messages collected by a handler are confirmed as soon as that handler finishes successfully. When the worker fails a message, only the confirmed messages are sent, while the messages of the failed handler are dropped and dispatched again on retry. The handled-successfully path stays unchanged.

The behavior is documented in the asynchronous processing docs, including the note that batch handlers are not covered by the confirmation.

#### Fixes issues
N/A

----
  - https://mg-mail-delayed.odin.shopsys.cloud
  - https://cz.mg-mail-delayed.odin.shopsys.cloud
  - https://mg-mail-delayed.odin.shopsys.cloud/sk


packages/framework/src/Component/Messenger/DelayedEnvelope/DispatchCollectedEnvelopesSubscriber.php
```diff
@@ -68,6 +79,11 @@ public static function getSubscribedEvents(): array
             KernelEvents::RESPONSE => ['handleCollectedMessageEnvelopes', -10],
             ConsoleEvents::TERMINATE => ['handleCollectedMessageEnvelopes', -10],
             WorkerMessageHandledEvent::class => ['handleCollectedMessageEnvelopes', 10],
+            // handlers that succeeded will not run again on retry, so their confirmed envelopes must be sent now,
+            // the unconfirmed envelopes belong to the failed handlers and will be dispatched again on retry.
+            // Runs after Symfony's retry (100) and failure transport (-100) listeners so a failure while sending
+            // cannot interrupt the hand-off of the failed message
+            WorkerMessageFailedEvent::class => ['handleConfirmedMessageEnvelopes', -110],
```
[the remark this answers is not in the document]
grossmannmartin: The hazard is real but pre-existing: the `WorkerMessageHandledEvent` listener has dispatched before the ack since the collector was introduced, and this listener mirrors it. Making the dispatch loop non-throwing changes the failure semantics of both paths and is out of scope here. Priority reverted to 10 to match the handled path.

packages/framework/src/Component/Messenger/DelayedEnvelope/DelayedEnvelopesCollector.php
```diff
@@ -9,21 +9,108 @@
...
+     * @var array<string, \Symfony\Component\Messenger\Envelope[]>
+     */
+    protected array $envelopesByRunningHandler = [];
+
+    /**
+     * Names of the running handlers, the last one receives the dispatched envelopes
+     *
+     * @var string[]
+     */
+    protected array $runningHandlerNames = [];
```
vitek-rostislav: It felt like the class has too many properties and could be simplified, I tried to remove  this one as the same information should be available in `$envelopesByRunningHandler` - what do you think? https://github.com/shopsys/shopsys/commit/817651bc62c7127546af992f37aab89744e5cea5

packages/framework/src/Component/Messenger/DelayedEnvelope/SegmentedHandlersLocator.php
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Reports the boundaries of every handler call to DelayedEnvelopesCollector, so the envelopes dispatched by a handler
+ * are confirmed when it succeeds and discarded when it fails, independently of the other handlers of the same message.
+ * Symfony retries only the failed handlers of a message (the successful ones are skipped thanks to HandledStamp),
+ * so the envelopes of a successful handler must be sent even when a sibling handler fails - nothing would dispatch them again.
+ *
+ * Symfony 7.4 offers no hook around a single handler call, hence the decoration of the handlers locator and wrapping
+ * of the handlers into closures. Symfony 8.2 dispatches HandlerStartingEvent, HandlerSuccessEvent and HandlerFailureEvent
+ * around every handler call (symfony/symfony#52425). After the upgrade, replace this class and its decoration
```
vitek-rostislav: really a nitpick, but lets put the direct link here, please, it will save us a few "clicks" in the future :smile: https://github.com/symfony/symfony/pull/52425

vitek-rostislav: btw, I have created an epic for SF8 upgrade and put new https://shopsys.atlassian.net/browse/SSP-4377 task there

vitek-rostislav (changes requested):

packages/framework/src/Component/Messenger/DelayedEnvelope/SegmentedHandlersLocator.php
grossmannmartin: fixed in 6d34152960

packages/framework/src/Component/Messenger/DelayedEnvelope/DelayedEnvelopesCollector.php
grossmannmartin: nice idea, changed in 6d34152960

vitek-rostislav (approved):

packages/framework/src/Component/Messenger/DelayedEnvelope/DispatchCollectedEnvelopesSubscriber.php
```diff
@@ -68,6 +79,9 @@ public static function getSubscribedEvents(): array
             KernelEvents::RESPONSE => ['handleCollectedMessageEnvelopes', -10],
             ConsoleEvents::TERMINATE => ['handleCollectedMessageEnvelopes', -10],
             WorkerMessageHandledEvent::class => ['handleCollectedMessageEnvelopes', 10],
+            // handlers that succeeded will not run again on retry, so their confirmed envelopes must be sent now,
+            // the unconfirmed envelopes belong to the failed handlers and will be dispatched again on retry
+            WorkerMessageFailedEvent::class => ['handleConfirmedMessageEnvelopes', 10],
```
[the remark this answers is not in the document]
grossmannmartin: Good catch. The order was in fact safe (path-sorted glob registers this subscriber first), but it was incidental. Priority raised to 20 so the flush is guaranteed to run before `ResetWorkerOnClosedEntityManagerSubscriber` can throw: https://github.com/shopsys/shopsys/commit/091ce911e9b9c49b06b4e2a56a9dfeea3f2a65be

PR #4828, opened 2026-09-08
Redesign product comparison and preserve product order

chlebektomas:

Product comparison now keeps product cards, parameter columns and the sticky header aligned while scrolling. The redesigned page includes clearer empty states, collapsible parameter groups, a differences-only filter with a count, and an Undo action for removed products. On mobile, customers can choose the two products to compare from full-width menus with product thumbnails.

Customers can reorder comparison products by dragging or using the keyboard, and their order is preserved after refreshing the page. Parameter groups remain stable while product columns move. Keyboard focus highlights the product name or review count without obscuring the rest of the card.

Product codes can be selected with one click wherever they are displayed. The inquiry popup and phone-prefix fields also have more room.

#### Fixes issues

N/A

----
  - https://tc-ssp-4231-product-comparison-redesign.odin.shopsys.cloud
  - https://cz.tc-ssp-4231-product-comparison-redesign.odin.shopsys.cloud
  - https://tc-ssp-4231-product-comparison-redesign.odin.shopsys.cloud/sk


project-base/storefront/utils/productLists/comparison/useComparisonProducts.ts
```diff
@@ -0,0 +1,83 @@
...
+            }
+        } finally {
+            savingOrder.current = false;
+        }
+    };
+    const orderedProducts = [
+        ...productOrder
+            .map((uuid) => comparedProducts.find((product) => product.uuid === uuid))
+            .filter((product): product is TypeProductInProductListFragment => !!product),
+        ...comparedProducts.filter((product) => !productOrder.includes(product.uuid)),
```
JanMolcik, line 56: [P2] Reconcile the local order after successful saves and Undo

After a successful reorder, `productOrder` permanently overrides later server responses. Reorder A,B,C to C,A,B, save, remove B, then Undo: `useComparisonPage` restores B through normal add, and `ProductList::addItem()` assigns `min(position)-1`, so the server returns B,C,A. This mapping still displays C,A,B; reloading clears the local override and changes the order to B,C,A. This is a source-traced persistence mismatch, not a live reproduction.

Limit the optimistic override to pending changes and reconcile successful responses while preserving queued reorders. If Undo should restore the original slot, persist that order too. Add a reorder → save → remove → Undo → reload integration test asserting identical displayed order before and after reload.

project-base/storefront/utils/toasts/showMessage.tsx
```diff
@@ -80,10 +88,17 @@ export const showMessage = (
             icon: <WarningIcon />,
         });
     } else {
-        toast.success(() => <ToastMessageContent message={message} tid={TIDs.toast_success} />, {
+        const content = <ToastMessageContent message={message} tid={TIDs.toast_success} action={options?.action} />;
+        const successOptions = {
             toastId,
-            closeOnClick: true,
+            closeOnClick: !options?.action,
```
JanMolcik, line 94: [P2] Update the Cypress toast helper for action toasts

Disabling body-click dismissal for Undo is intentional, but `checkAndHideToast` in `cypress/support/index.ts:161–169` still clicks `toast_success` and waits for removal. On this head, [comparison CI](https://github.com/shopsys/shopsys/actions/runs/34231152421/job/102079258697) times out after 20 seconds in `[Remove Single Product]` at helper line 169, identifying the new action-toast span.

Have the helper use the actual close control, or explicitly exercise Undo and dismissal in the comparison scenario, then rerun the comparison suite including remove/restore. Keep Undo's click behavior. This is a confirmed test/UI contract mismatch, not evidence that removing a product itself fails.

JanMolcik (commented):
Reviewed head: `4494560214b311418cf0c9d907376d24469c62ff`.

Two P2 findings below: a functional persistence mismatch and a Cypress integration failure. The recorded comparison job also has visual snapshot differences; inspect those separately rather than treating the toast failure as a product-removal defect.

Partial review. Local test suites were not run because Docker was unavailable. The initial preview HTTP 403 occurred while Shopsys VPN was off; subsequent limited browser attempts still returned 403. This is an access limitation, not evidence of a deployment defect. The relevant runtime scenarios and deployed SHA remain unverified.

JanMolcik (changes requested):
Requesting changes for the two confirmed issues in the existing inline discussions: [finding](https://github.com/shopsys/shopsys/pull/4828#discussion_r3972049121), [finding](https://github.com/shopsys/shopsys/pull/4828#discussion_r3972049123).

Decision follow-up to the [previous review](https://github.com/shopsys/shopsys/pull/4828#pullrequestreview-5158674971) on unchanged head `4494560214b311418cf0c9d907376d24469c62ff`.

Partial validation: the ordering mismatch is source-traced; applicable browser scenarios and deployed SHA remain unverified, and local suites were not run.

packages/framework/src/Model/Product/List/ProductListFacade.php
```diff
@@ -44,6 +46,43 @@ public function addProductToList(ProductList $productList, Product $product): Pr
...
+            $nextIndex = 0;
+
+            // Keep products absent from the storefront response in their existing slots.
+            foreach ($items as $position => $item) {
+                if (in_array($item->getProduct()->getUuid(), $productUuids, true)) {
+                    $orderedItems[$nextIndex++]->changePosition($position);
+                } else {
+                    $item->changePosition($position);
+                }
+            }
```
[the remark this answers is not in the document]
chlebektomas: Fixed in https://github.com/shopsys/shopsys/commit/82193ace6e90962765d68cbae7b5bdb7888f2039 — the input is capped at 100 items and membership validation is linear.

project-base/storefront/utils/productLists/comparison/useComparisonProducts.ts
chlebektomas: Fixed in https://github.com/shopsys/shopsys/commit/82193ace6e90962765d68cbae7b5bdb7888f2039 — confirmed saves now reconcile with server order, including Undo and reload.

project-base/storefront/utils/toasts/showMessage.tsx
chlebektomas: Fixed in https://github.com/shopsys/shopsys/commit/82193ace6e90962765d68cbae7b5bdb7888f2039 — the helper now uses the close control and the comparison scenario covers Undo.

JanMolcik (approved):

PR #4830, opened 2026-09-09
Quick approval of pending product reviews in administration

grossmannmartin:

Product reviews that are waiting for moderation can now be approved with a single click.

Until now the moderator had to open a pending review, change the status select in the form to "Approved" and save the form. For the most common moderation outcome that is needlessly long. The edit page of a pending review now offers an **Approve review** button in the header; clicking it approves the review, keeps any previously entered rejection reason out of the approved review, and returns the moderator to the review list with a confirmation message linking back to the review.

The approval triggers the same follow-up as saving the form (product export and storefront cache invalidation), so the approved review shows up on the storefront right away. The button is available only to administrators who may edit product reviews, and the link is CSRF-protected.

#### Fixes issues
N/A

----
  - https://mg-quick-review-approve.odin.shopsys.cloud
  - https://cz.mg-quick-review-approve.odin.shopsys.cloud
  - https://mg-quick-review-approve.odin.shopsys.cloud/sk

### DIFF
--- a/packages/administration/src/Controller/ProductReviewController.php
+++ b/packages/administration/src/Controller/ProductReviewController.php
@@ -6,7 +6,10 @@ namespace Shopsys\AdministrationBundle\Controller;
 
 use Doctrine\ORM\QueryBuilder;
 use Override;
+use Shopsys\AdministrationBundle\Component\Action\Action;
 use Shopsys\AdministrationBundle\Component\Attributes\CrudController;
+use Shopsys\AdministrationBundle\Component\Config\ActionsConfig;
+use Shopsys\AdministrationBundle\Component\Config\ActionType;
 use Shopsys\AdministrationBundle\Component\Config\CrudConfig;
 use Shopsys\AdministrationBundle\Component\Config\CrudListDomainControl;
 use Shopsys\AdministrationBundle\Component\Crud\Form\CrudFormConfigurator;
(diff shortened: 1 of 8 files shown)


packages/administration/src/Controller/ProductReviewController.php
```diff
@@ -143,4 +167,27 @@ protected function getEditViewData(object $entity): array
             'productReview' => $productReview,
         ];
     }
+
+    #[Route(path: '/product-review/approve/{id}', name: 'admin_crud_product_review_approve', requirements: ['id' => '\d+'], methods: ['GET'])]
+    #[CanEdit]
+    #[CsrfProtection]
+    public function approveAction(int $id): Response
```
henzigo, line 174: I’m fine with that as a temporary solution. 

The problem is that, in this case, the CrudController is completely unaware of this action – and vice versa. This means that if I wanted to add a routePrefix, it wouldn’t be applied to that particular action. 
Similarly, if the entire CrudController is disabled, the individual action will still exist.
But the main issue with those permissions remains. If I disable `ActionType::EDIT`, `CrudAdminRoleProvider` generates a `Role` without `Permission::EDIT`, and this stops working altogether

These are precisely some of the reasons why I proposed creating a new attribute and connecting these actions directly to Crud:

```php
#[CrudAction(
    controller: ProductReviewController::class, // probably needed only in case you define that action outside CrudController
    name: 'approve',
    path: '/approve/{id}',
    permissions: [
        new CanEdit(),
    ],
)]
public function approveAction(Request $request, int $id): Response
```

grossmannmartin: I don't like this and don't want this

TomasLudvik (approved):

packages/administration/src/Controller/ProductReviewController.php
```diff
@@ -143,4 +167,38 @@ protected function getEditViewData(object $entity): array
...
+    {
+        /** @var \Shopsys\AdministrationBundle\Model\ProductReview\ProductReviewEditHandler $handler */
+        $handler = $this->definition->getHandlerForAction(ActionType::EDIT);
+        $productReview = $handler->getById($id);
+
+        if ($productReview->getStatus() !== ProductReviewStatusEnum::STATUS_PENDING) {
```
henzigo, line 180: Just a small detail. This check should be moved to `ProductReview::approve()`. 

After moving that condition, the method may throw an exception, which you can catch in that action. 

At the very least, this will result in a better-handled method, preventing anyone from approving a review with the `STATUS_REJECTED` status.

grossmannmartin: Done, good idea

PR #4833, opened 2026-09-09
Default pricing group can no longer be deleted without a replacement

grossmannmartin:

Administrators could delete the default pricing group of a domain without the replacement ever becoming the new default. The confirmation dialog asked for a replacement group, but only customers were moved to it – the "default pricing group" setting of the domain kept pointing to the deleted group, which breaks pricing and product visibility for anonymous visitors on that domain. This was a regression from an earlier refactoring where the domain stopped being passed into the delete operation.

Now the replacement pricing group is also set as the domain default whenever the deleted group was the default. In addition, the delete operation validates its input on the server: a pricing group that is in use (default on its domain or assigned to customers) cannot be deleted without a replacement, and the replacement must be a different pricing group from the same domain. Previously these cases were only prevented by the browser form and could lead to a corrupted default setting or a database error.

#### Fixes issues
N/A

----
  - https://mg-fix-pricing-group.odin.shopsys.cloud
  - https://cz.mg-fix-pricing-group.odin.shopsys.cloud
  - https://mg-fix-pricing-group.odin.shopsys.cloud/sk

### DIFF
--- a/packages/framework/src/Controller/Admin/PricingGroupController.php
+++ b/packages/framework/src/Controller/Admin/PricingGroupController.php
@@ -6,6 +6,7 @@ namespace Shopsys\FrameworkBundle\Controller\Admin;
 
 use Shopsys\FrameworkBundle\Component\ConfirmDelete\ConfirmDeleteResponseFactory;
 use Shopsys\FrameworkBundle\Component\Domain\AdminDomainTabsFacade;
+use Shopsys\FrameworkBundle\Component\Domain\Domain;
 use Shopsys\FrameworkBundle\Component\HttpFoundation\HttpMethod;
 use Shopsys\FrameworkBundle\Component\Router\Security\Attribute\CsrfProtection;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\CanDelete;
@@ -14,6 +15,8 @@ use Shopsys\FrameworkBundle\Component\Security\Attribute\CanView;
 use Shopsys\FrameworkBundle\Component\Security\Attribute\ForRole;
 use Shopsys\FrameworkBundle\Component\Security\Role\AdminRoleConstant;
 use Shopsys\FrameworkBundle\Form\Admin\Pricing\Group\PricingGroupSettingsFormType;
+use Shopsys\FrameworkBundle\Model\Pricing\Group\Exception\InvalidPricingGroupReplacementException;
+use Shopsys\FrameworkBundle\Model\Pricing\Group\Exception\PricingGroupIsUsedException;
 use Shopsys\FrameworkBundle\Model\Pricing\Group\Exception\PricingGroupNotFoundException;
 use Shopsys\FrameworkBundle\Model\Pricing\Group\Grid\PricingGroupInlineEdit;
 use Shopsys\FrameworkBundle\Model\Pricing\Group\PricingGroupFacade;
@@ -31,6 +34,7 @@ class PricingGroupController extends AdminBaseController
         protected readonly PricingGroupInlineEdit $pricingGroupInlineEdit,
         protected readonly ConfirmDeleteResponseFactory $confirmDeleteResponseFactory,
         protected readonly AdminDomainTabsFacade $adminDomainTabsFacade,
+        protected readonly Domain $domain,
     ) {
     }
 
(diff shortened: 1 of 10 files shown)


packages/framework/src/Controller/Admin/PricingGroupController.php
```diff
@@ -90,16 +98,16 @@ public function deleteConfirmAction(int $id): Response
         try {
             $pricingGroup = $this->pricingGroupFacade->getById($id);
 
-            $selectedDomain = $this->adminDomainTabsFacade->getSelectedDomainConfig();
+            $domainConfig = $this->domain->getDomainConfigById($pricingGroup->getDomainId());
 
-            if ($this->pricingGroupSettingFacade->isPricingGroupUsedOnDomain($pricingGroup, $selectedDomain)) {
```
vitek-rostislav, line 95: Something to think about and possibly talk about - why do we even pass the `DomainConfig` into `isPricingGroupUsedOnDomain` and `isPricingGroupDefaultOnDomain`? The pricing group is tied to its domain, so it feels like we could simplify the interface and just ask, "is this pricing group used?" and then use its `domainId` to figure out whether it is set as the default on the given domain. But it would be a breaking change 🤷

grossmannmartin: great idea

project-base/app/tests/App/Functional/Model/Pricing/Group/PricingGroupFacadeTest.php
```diff
@@ -159,4 +161,20 @@ public function testDeleteWithItselfAsReplacementThrowsException(): void
 
         $this->pricingGroupFacade->delete($pricingGroup->getId(), $pricingGroup->getId());
     }
+
+    private function assignPricingGroupToCustomerUser(CustomerUser $customerUser, PricingGroup $pricingGroup): void
+    {
+        $customerUserData = $this->customerUserDataFactory->createFromCustomerUser($customerUser);
+        $customerUserData->pricingGroup = $pricingGroup;
+
+        /** @var \App\Model\Customer\BillingAddress $billingAddress */
```
vitek-rostislav, line 172: This annotation is unnecessary, let me quote my IDE:
> @var tag specifies the type already inferred from source code

vitek-rostislav: (btw, the same unnecessary annotation is actually twice in the test class)

project-base/app/tests/App/Functional/Model/Pricing/Group/PricingGroupFacadeTest.php
```diff
@@ -62,6 +62,7 @@ public function testDeleteAndReplace(): void
             Domain::FIRST_DOMAIN_ID,
             PricingGroup::class,
         );
+        /** @var \App\Model\Customer\User\CustomerUser $customerUser */
```
vitek-rostislav: ah, here are the new annotations, @grossmannmartin will not be happy with your fixes, @github-actions[bot]! 😄

phpstan-dead-code.neon
```diff
@@ -382,6 +382,10 @@ parameters:
             identifier: shipmonk.deadMethod
             message: '#^Unused\ Shopsys\\FrameworkBundle\\Model\\Pricing\\Group\\PricingGroupEvent::getPricingGroup$#'
             path: packages/framework/src/Model/Pricing/Group/PricingGroupEvent.php
+        -
+            identifier: shipmonk.deadMethod
```
vitek-rostislav: That naughty bot 😄 I think, if the method is used in tests only, it should actually be removed, and the tests should be adjusted 🤷

vitek-rostislav (changes requested):

phpstan-dead-code.neon
grossmannmartin: fixed

project-base/app/tests/App/Functional/Model/Pricing/Group/PricingGroupFacadeTest.php
grossmannmartin: fixed

project-base/app/tests/App/Functional/Model/Pricing/Group/PricingGroupFacadeTest.php
grossmannmartin: now it seems like it's still needed

vitek-rostislav: You are absolutely right, sir :smile:

vitek-rostislav (approved):

PR #4842, opened 2026-09-09
PHPStan now understands Webmozart Assert calls in CRUD handlers

henzigo:

PHPStan now understands `Webmozart\Assert\Assert` calls. CRUD handlers (and other code) narrow generic `object` parameters with `Assert::isInstanceOf()`, but PHPStan so far ignored those asserts, so the following calls were analysed against a plain `object` and typed facade calls were either reported as errors or had to be worked around. The `phpstan/phpstan-webmozart-assert` extension is added as a dev dependency and enabled in every `phpstan.neon` (monorepo, project-base, framework, administration, frontend-api), so the type after an assert is known to static analysis the same way it is at runtime.

`BlogArticleAuthorCrudHandler` now asserts the entity and data object types in all methods that receive them (`create()`, `createDataFromEntity()`, `edit()`), matching `TransportGroupCrudHandler`. A wrongly registered handler therefore fails fast with a clear message instead of somewhere deep in the facade.

The CRUD handler documentation (getting-started examples and the handlers reference) now shows and explains the `Assert::isInstanceOf()` narrowing, so new handlers do not forget it. The `coding-conventions` skill gets a new *Validation & guard clauses* section: prefer `Webmozart\Assert\Assert` over hand-written guards, narrow `object` parameters with `Assert::isInstanceOf()` instead of `/** @var */` docblocks, pass non-trivial messages as closures, and use a dedicated exception class when callers need to catch the failure. As a side effect of the new extension, `Assert::allNotNull()` on untyped data object properties in `OrderFacade` was reported as always true and was replaced by separate `Assert::notNull()` calls with the same behaviour.

#### Fixes issues

N/A

----
  - https://jg-phpstan-webmozart-assert.odin.shopsys.cloud
  - https://cz.jg-phpstan-webmozart-assert.odin.shopsys.cloud
  - https://jg-phpstan-webmozart-assert.odin.shopsys.cloud/sk

----
  - https://jg-phpstan-webmozart-assert.odin.shopsys.cloud
  - https://cz.jg-phpstan-webmozart-assert.odin.shopsys.cloud
  - https://jg-phpstan-webmozart-assert.odin.shopsys.cloud/sk

### DIFF
--- a/docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
+++ b/docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
@@ -16,6 +16,7 @@ declare(strict_types=1);
 namespace App\Model\Order;
 
 use Shopsys\AdministrationBundle\Component\Crud\Handler\DeleteHandlerInterface;
+use Webmozart\Assert\Assert;
 
 class OrderDeleteHandler implements DeleteHandlerInterface
 {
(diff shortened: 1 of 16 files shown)


docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
```diff
@@ -24,18 +26,28 @@ class OrderDeleteHandler implements DeleteHandlerInterface
     ) {
     }
 
-    public function getById(int $id): object
+    public function getById(int $id): Presentable
```
grossmannmartin, line 29: Thanks to covariance, we can type-hint the entity directly. 


```suggestion
    public function getById(int $id): Order
```

The same applies to multiple places in the documentation

grossmannmartin (approved):

pk16011990 (approved):

docs/administration/crud-controller/getting-started/adding-create-edit-and-delete-actions.md
henzigo: Good catch. I will look at it in separate PR

PR #4850, opened 2026-09-12
Show change history of loggable entities on CRUD edit pages

henzigo:

Administrators editing a record in a CRUD controller section now see its change history (who changed what and when) directly below the form, without the developer having to build it into the page. The history appears automatically for every entity that has entity logging enabled, so newly added CRUD sections get it for free and it looks the same everywhere.

Until now, each CRUD controller that wanted to show the history had to provide its own edit template and pass the entity log data to it by hand, as the product review administration did. That code is gone; the product review edit page shows the same history through the shared mechanism.

Where the history is not wanted, the CRUD controller turns it off with `showEntityLog(false)` in its configuration, or a custom edit template overrides the `entity_log` block. The entity log timeline component also accepts the logged entity directly, so other admin templates can render it without resolving the entity name and ID themselves.

#### Fixes issues

N/A

----
  - https://jg-crud-entity-log.odin.shopsys.cloud
  - https://cz.jg-crud-entity-log.odin.shopsys.cloud
  - https://jg-crud-entity-log.odin.shopsys.cloud/sk


docs/administration/crud-controller/reference/crud-controller.md
```diff
@@ -340,3 +340,17 @@ $config
     ->setMenuIcon('cart')
 ;
 ```
+
+#### `showEntityLog(bool $show)`
```
vitek-rostislav, line 344: I would consider removing this config method - when I want to log the entity changes, I have to configure it using the `Loggable` attribute. In such a case, I want to see the log in the admin. Logging the entity and not showing the log at the same time is a weird edge case IMHO. I do not think we need to support this using the config at all. It might be rather confusing as calling `showEntityLog(true)` for an entity without the attribute would do nothing. Removing the config method would nicely simplify the implementation to the concept - "is the entity loggable? -> then show its log on the edit page"

vitek-rostislav (changes requested):

docs/administration/crud-controller/reference/crud-controller.md
henzigo: Ok, why not. done

vitek-rostislav (changes requested):
During the testing, I have noticed one more thing we should address. Basically, you always want to show the entity log at the end of the detail page. I tried testing it on `BlogArticleAuthorController` and its edit page looks like this:
```
{% block main_content %}
    {{ parent() }}

    <div class="mt-5">
        <h2>{{ 'Articles by this author'|trans }}</h2>

        {{ gridView.render() }}
    </div>
{% endblock %}
```

As a result, the entity log is rendered "in the middle" of the page. 

I suggest adding an empty block into `@ShopsysAdministration/crud/edit.html.twig` template inside `main_content`, that would be the place to extend when you need to add something between the form and the changelog. I am also open to other creative solutions 😄

vitek-rostislav (approved):

PR #4854, opened 2026-09-14
Negated filters on nullable columns now keep rows with empty values

grossmannmartin:

Two queries negated a condition on a nullable column without handling `NULL`. In SQL, `NULL NOT IN (...)` and `NULL NOT LIKE ...` evaluate to unknown rather than true, so rows with an empty value were silently excluded even though they clearly belong to the "does not match" result.

**Admin product search:** the advanced search rule "Part number does not contain" skipped every product that had no part number at all. Administrators trying to find products outside a given numbering range never saw the products with the value left empty. The rule now keeps such products in the result. The sibling "Catalog number" rule is intentionally left untouched: project-base declares the catalog number as a required, unique column, so it can never be empty there.

**GoPay unpaid orders check:** the cron that re-checks unpaid GoPay orders excluded payment transactions whose external status was not stored yet. The external status column is nullable, so such transactions were never re-checked. They are now included in the lookup.

#### Fixes issues

N/A

----
  - https://mg-null-safe-negated-predicates.odin.shopsys.cloud
  - https://cz.mg-null-safe-negated-predicates.odin.shopsys.cloud
  - https://mg-null-safe-negated-predicates.odin.shopsys.cloud/sk


packages/framework/src/Model/Product/AdvancedSearch/Filter/ProductPartnoFilter.php
```diff
@@ -60,7 +60,14 @@ public function extendQueryBuilder(QueryBuilder $queryBuilder, array $rulesData)
                 $searchValue = $this->getSearchValue($ruleData);
                 $dqlOperator = $this->getDqlOperator($ruleData->operator);
                 $parameterName = 'productPartno_' . $index;
-                $queryBuilder->andWhere('NORMALIZED(p.partno) ' . $dqlOperator . ' NORMALIZED(:' . $parameterName . ')');
+                $condition = 'NORMALIZED(p.partno) ' . $dqlOperator . ' NORMALIZED(:' . $parameterName . ')';
+
+                if ($ruleData->operator === self::OPERATOR_NOT_CONTAINS) {
+                    // NULL NOT LIKE ... is UNKNOWN in SQL, so products without partno would be dropped
```
TomasLudvik, line 66: Do we really need these kind of comments?

TomasLudvik (approved):

packages/framework/src/Model/Product/AdvancedSearch/Filter/ProductPartnoFilter.php
grossmannmartin: 🤷‍♂️ This behavior was not familiar to me before... but removed in 90c1dd1189

pk16011990 (approved):

PR #4856, opened 2026-09-14
Admin templates now use protectedUrl() for CSRF-protected links

vitek-rostislav:

The new CSRF protection guide (#4619) tells developers to generate links to `#[CsrfProtection]` actions with `protectedUrl()`. The administration templates themselves did not follow that advice in several places — category, blog category, customer, customer user and delivery address lists composed the `routeCsrfToken` parameter by hand with `constant(...)` and `csrf_token(...)` around a plain `url()` call, five lines per link.

This PR replaces all eight such places with a one-line `protectedUrl()` call, so the templates match what the documentation recommends and there is one less pattern to copy when adding a new protected action. The generated links are identical (same route, same token id), only the way they are built in Twig changes. The only manual token left is the hidden input in the confirm-delete replacement form, which is the legitimate case the guide describes.

Because the manual-token example in the guide quoted exactly the category list snippet that no longer exists, it was rewritten around the remaining real case: the "set new and delete" form for deleting a unit that is still in use.

Stacked on top of #4619 — the diff shows only the template and docs changes once that PR is merged.

#### Fixes issues

N/A

----
  - https://rv-csrf-protected-url-in-templates.odin.shopsys.cloud
  - https://cz.rv-csrf-protected-url-in-templates.odin.shopsys.cloud
  - https://rv-csrf-protected-url-in-templates.odin.shopsys.cloud/sk


packages/administration/templates/content/blog/category/list.html.twig
```diff
@@ -54,11 +54,7 @@
                                class="link-secondary"
                             >{{ ux_icon('pencil') }}</a>
 
-                            {% set csrfTokenId = constant('Shopsys\\FrameworkBundle\\Component\\Router\\Security\\RouteCsrfProtector::CSRF_TOKEN_ID_PREFIX') ~ 'admin_blogcategory_delete' %}
-                        {% set blogCategoryDeleteUrl = url('admin_blogcategory_delete', {
-                            id: blogCategoryId,
-                            (constant('Shopsys\\FrameworkBundle\\Component\\Router\\Security\\RouteCsrfProtector::CSRF_TOKEN_REQUEST_PARAMETER')): csrf_token(csrfTokenId)
-                        }) %}
+                            {% set blogCategoryDeleteUrl = protectedUrl('admin_blogcategory_delete', {id: blogCategoryId}) %}
```
grossmannmartin: nitpick: I think we can inline each variable as call of protectedUrl is simple instruction now

grossmannmartin (approved):

PR #4866, opened 2026-09-16
Saving an entity from a stale browser tab no longer breaks images and files

grossmannmartin:

When an administrator had the same entity (product, blog article, brand, ...) open in two browser tabs and changed its images or files in one of them, saving the other tab ended badly. If the images had been added in the meantime, the save crashed with a 500 error (`Key "" for sequence/mapping ... does not exist in ImageUploadType.html.twig`). 
If they had been deleted in the meantime, the save failed with a meaningless "The collection is invalid" or "This form should not contain extra fields" error, and reloading the page only re-submitted the same stale form.

Now the stale tab saves without errors, and the images and files end up in a sensible state, following the same "last write wins" rule the rest of the form already has:

- images and files uploaded from another tab are kept, uploads from the stale tab are appended after them,
- images and files deleted from another tab stay deleted, and their names, filenames, and delete checkboxes submitted by the stale tab are ignored,
- ordering and renaming from the stale tab applies to the items that still exist,
- for single-image and single-file slots, a new upload or a file picked from the file picker replaces whatever is stored at that moment, so no orphaned image or file is left behind.

The re-rendered form also no longer crashes when it is shown after an unrelated validation error, and the stored images differ from the submitted ones.

Both behaviours are covered by functional tests for the image and file upload forms, including the two-tab scenarios for multiple and single items, reordering, renaming, deleting from both tabs, and the file picker.

#### Fixes issues

N/A

----
  - https://mg-fix-image-saving-stale-tab.odin.shopsys.cloud
  - https://cz.mg-fix-image-saving-stale-tab.odin.shopsys.cloud
  - https://mg-fix-image-saving-stale-tab.odin.shopsys.cloud/sk


project-base/app/tests/App/Functional/Form/ImageUploadTypeTest.php
```diff
@@ -0,0 +1,473 @@
...
+        $this->uploadImages($brand, 1);
+        $initialDataToSubmit = $this->createSubmittedDataFromStoredImages($brand);
+        [$replacementImageId] = $this->uploadImages($brand, 1);
+
+        $this->submitStaleFormAndSave($brand, Brand::class, $initialDataToSubmit);
+
+        $this->assertSame([$replacementImageId], $this->getStoredImageIds($brand));
+    }
+
+    public function testSingleImageStaleFormUploadReplacesReplacementUploadedMeanwhile(): void
```
vitek-rostislav, line 283: funny test name 😅

vitek-rostislav (approved):

pk16011990 (approved):

